PR #51, opened 2022-07-18
Add embedded variant of date range picker

zshakeel-amazon:
Changes Done
Adds a new variant of the date range picker that is not within a dropdown component.

Why?
This allows for this variant of the date range picker to be embedded within other dropdown components, avoiding dropdowns within dropdowns.

Testing
Manual testing. Added demo page.

### Documentation changes

Not needed

### Related Links

SIM: AWSUI-9912

#### Correctness

#### Security

#### Testing

</details>


src/date-range-picker/date-range-picker-embedded.tsx
```diff
@@ -0,0 +1,153 @@
...
+    | 'rangeSelectorMode'
+  > {
+  startOfWeek: DayIndex;
+  isSingleGrid: boolean;
+
+  ariaLabelledby?: string;
+  ariaDescribedby?: string;
+}
+
+export function DateRangePickerEmbeddedDropdown({
```
YueyingLu, line 37: We want to have it embedded in a dropdown, but it is not a dropdown or?

zshakeel-amazon: Agreed, I can definitely update the name here to exclude the dropdown part.

src/date-range-picker/date-range-picker-embedded.tsx
```diff
@@ -0,0 +1,153 @@
...
+        <SpaceBetween size="l">
+          <InternalBox padding={{ top: 'm', horizontal: 'l' }}>
+            <SpaceBetween direction="vertical" size="s">
+              {rangeSelectorMode === 'default' && (
+                <ModeSwitcher
+                  ref={focusRefs.default}
+                  mode={rangeSelectionMode}
+                  onChange={(mode: 'absolute' | 'relative') => {
+                    setRangeSelectionMode(mode);
+                    setApplyClicked(false);
```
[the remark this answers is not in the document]
zshakeel-amazon: Agreed, not required for the embedded variant

pages/date-range-picker/embedded.page.tsx
```diff
@@ -0,0 +1,57 @@
...
+// SPDX-License-Identifier: Apache-2.0
+import React, { useState } from 'react';
+import { Box, DateRangePickerProps, Link, Checkbox, SpaceBetween, SegmentedControl, FormField } from '~components';
+import { DateRangePickerEmbeddedDropdown } from '~components/date-range-picker/date-range-picker-embedded';
+import { i18nStrings, i18nStringsDateOnly, isValid, relativeOptions } from './common';
+
+export default function DatePickerScenario() {
+  const [showRelativeOptions, setShowRelativeOptions] = useState(true);
+  const [dateOnly, setDateOnly] = useState(false);
+  const [rangeSelectorMode, setRangeSelectorMode] = useState<DateRangePickerProps.RangeSelectorMode>('default');
```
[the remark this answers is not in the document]
zshakeel-amazon: Thanks, this seems to be an issue with the simple demo as well if you switch between the options without closing the dropdown. Will fix in revision

src/date-range-picker/date-range-picker-embedded.tsx
```diff
@@ -0,0 +1,149 @@
...
+    VALID_RANGE,
+  } = useDateRangePicker({
+    value,
+    relativeOptions,
+    rangeSelectorMode,
+    isValidRange,
+  });
+
+  return (
+    <div
```
pan-kot, line 73: Looks like a lot of the markup is duplicated between date-range-picker-embedded and the dropdown. It would be better if the dropdown can be shared instead:

```
// date-range-picker-embedded.tsx

function DateRangePickerEmbedded(props) {
  const drpProps = useDateRangePicker(props)
  ...
  return <Dropdown ... />
}

// internal.tsx

function DateRangePicker(props) {
  const drpProps = useDateRangePicker(props)
  ...
  return <>
    <Trigger ... />
    <Dropdown ... />
  <>
}
```

src/date-range-picker/range-picker.tsx
```diff
@@ -0,0 +1,159 @@
...
+
+    if (newValue) {
+      const validationResult = isValidRange(newValue);
+      setValidationResult(validationResult || VALID_RANGE);
+      if (validationResult?.valid === false) {
+        setValidationResult(validationResult);
+      }
+    }
+
+    fireNonCancelableEvent(onChange, { value: formatValue(newValue, { dateOnly, timeOffset }) });
```
just-boris, line 84: Looks like it is this anti-pattern: https://beta.reactjs.org/learn/you-might-not-need-an-effect#notifying-parent-components-about-state-changes

Is there any chance to follow their recommendations? 

P.S. highly recommend reading the full article

src/date-range-picker/use-date-range-picker.tsx
```diff
@@ -0,0 +1,106 @@
...
+  >(VALID_RANGE);
+
+  const focusRefs = {
+    default: useRef<Focusable>(null),
+    'absolute-only': useRef<Focusable>(null),
+    'relative-only': useRef<Focusable>(null),
+  };
+
+  return {
+    VALID_RANGE,
```
just-boris, line 87: If it is a constant (never changes), why do you need to return it here?

UPD: same about other constant values. Why are they coming from hook and not `import {} from './utils'` for example?

src/date-range-picker/use-date-range-picker.tsx
```diff
@@ -0,0 +1,106 @@
...
+  return relativeOptions.length > 0 ? 'relative' : 'absolute';
+}
+
+export function useDateRangePicker({ value, relativeOptions, rangeSelectorMode }: UseDateRangePickerProps) {
+  const [rangeSelectionMode, setRangeSelectionMode] = useState<'absolute' | 'relative'>(
+    getDefaultMode(value, relativeOptions, rangeSelectorMode)
+  );
+
+  useEffect(() => {
+    setRangeSelectionMode(getDefaultMode(value, relativeOptions, rangeSelectorMode));
```
just-boris, line 65: Another anti-pattern from the same article: https://beta.reactjs.org/learn/you-might-not-need-an-effect#updating-state-based-on-props-or-state

src/date-range-picker/use-date-range-picker.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+  const [validationResult, setValidationResult] = useState<
+    DateRangePickerProps.ValidRangeResult | DateRangePickerProps.InvalidRangeResult
+  >(VALID_RANGE);
+
+  const focusRefs = {
+    default: useRef<Focusable>(null),
+    'absolute-only': useRef<Focusable>(null),
+    'relative-only': useRef<Focusable>(null),
+  };
```
just-boris, line 84: What is the point of putting focus refs inside this hook, if they are not used by any callbacks?

The same applies to the hook in general. Make sure you extracted _logically connected_ things into a hook

src/date-range-picker/range-picker.tsx
```diff
@@ -0,0 +1,159 @@
...
+import ModeSwitcher from './mode-switcher';
+import { useDateRangePicker } from './use-date-range-picker';
+import { DateRangePickerBaseProps } from './interfaces';
+import { fireNonCancelableEvent } from '../internal/events/index.js';
+import InternalAlert from '../alert/internal';
+import LiveRegion from '../internal/components/live-region';
+import styles from './styles.css.js';
+
+export interface DateRangePickerDropdownProps
+  extends Pick<
```
just-boris, line 18: Same question as in the other PR: https://github.com/cloudscape-design/components/pull/75#discussion_r930921917

Instead of hand picking properties, it is easier to create and merge multiple base types. (Exception: it could be okay to `Pick` 1-2 props, but 11 is too much)

just-boris (approved):

just-boris (changes requested):
approved by misclick, see the comments above

src/date-range-picker/range-picker.tsx
zshakeel-amazon: Thank you for that article, very applicable to this PR. Will incorporate the recommendations

src/date-range-picker/range-picker.tsx
zshakeel-amazon: Noted. Refactored the interface for this component as per the suggestion

pan-kot:
To be re-opened from a fork.

PR #54, opened 2022-07-18
Enable Visual Regression Tests again

timogasda:
### Description

This reverts commit b5fc5b2c5e9a26851cbe64119c5c54ee5e488771 and enables visual regression tests again. They were originally disabled because they cannot easily run without the dependencies being published to npm.

### How has this been tested?
We have used this in the past, so it should work.

~~This initial run will fail, though, because there will be no reference images. It will work as normal with the PR after this one.~~ Because we kept the original testing files and just disabled the workflow, this should just work immediately after merging.

### Documentation changes

- [x] _No._

<details>

#### Correctness

#### Security

#### Testing

</details>

### DIFF
--- a/.github/workflows/visuals.yml
+++ b/.github/workflows/visuals.yml
@@ -1,13 +1,12 @@
 name: Visual Regression Tests
 
 on:
-  workflow_dispatch:
-  # pull_request:
-  #   branches:
-  #     - main
-  # push:
-  #   branches:
-  #     - gh-readonly-queue/main/**
+  pull_request:
+    branches:
+      - main
+  push:
+    branches:
+      - gh-readonly-queue/main/**
 
 defaults:
   run:


.github/workflows/visuals.yml
```diff
@@ -1,13 +1,12 @@
...
-  #     - main
-  # push:
-  #   branches:
-  #     - gh-readonly-queue/main/**
+  pull_request:
+    branches:
+      - main
+  push:
+    branches:
+      - gh-readonly-queue/main/**
```
timogasda, line 9: This will currently do nothing because we're not using the merge queue feature right now. But in case we do use it again in the future, this will just worked :)

just-boris (approved):

avinashbot (approved):
LGTM!

PR #67, opened 2022-07-22
Production version of the ContentLayout component.

DaemonCahill:
### Description

This PR submits the production version of the ContentLayout component. Previous usage was for internal purposes only.

### How has this been tested?

Unit and Integration tests have been added for various rendering states and dynamic disabling of the overlap.

### Documentation changes

### Related Links

<details>

#### Correctness

- [x] _Changes were manually tested for accessibility, see [accessibility guidelines (https://cloudscape.design/foundation/core-principles/accessibility/)._

#### Security

#### Testing

</details>


build-tools/integ/setup.integ.js
```diff
@@ -3,7 +3,7 @@
 const { configure } = require('@cloudscape-design/browser-test-tools/use-browser');
 
 configure({
-  browserName: 'ChromeHeadlessIntegration',
```
DaemonCahill, line 6: @connorlanigan Were these changes intended to be committed?

src/content-layout/interfaces.ts
```diff
@@ -1,6 +1,6 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
-import { BaseComponentProps } from '../../base-component';
```
DaemonCahill, line 3: @FlorianDr Wasn't there something here that you indicated needed to be changed?

FlorianDr: Yes, we were missing the usually plumbing with `getBaseProps`.

build-tools/integ/setup.integ.js
connorlanigan: They were intended to be committed for easier debugging while you write the tests (because you can actually see what the browser does), but they should _not_ be part of the commit that goes into the main branch.

src/content-layout/index.tsx
```diff
@@ -2,21 +2,33 @@
...
   const rootElement = useRef<HTMLDivElement>(null);
+  const { __internalRootRef } = useBaseComponent('ContentLayout');
+  const mergedRef = useMergeRefs(rootElement, __internalRootRef);
   const isVisualRefresh = useVisualRefresh(rootElement);
-  const isOverlapDisabled = !children || !header || disableOverlap;
 
-  // Documentation to be added.
+  /**
+   * Observe the height of the ContentLayout background element. This value will be used to
+   * setthe dynamicOverlapHeight property in the AppLayout which will extend the overlap
```
timogasda, line 28: typo: *set the

src/content-layout/index.tsx
```diff
@@ -28,13 +40,21 @@ export default function ContentLayout({ children, disableOverlap, header }: Cont
     [isVisualRefresh, overlapContainerQuery, setDynamicOverlapHeight]
   );
 
+  /**
+   * Disable the overlap if the component is missing either a header or child
+   * content. If the component is not using visual refresh then the overlap
+   * will not displayed at all. This is handled in the CSS not the JavaScript.
```
timogasda, line 46: *not be displayed

src/test-utils/dom/content-layout/index.ts
```diff
@@ -0,0 +1,16 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import { ComponentWrapper, ElementWrapper } from '@cloudscape-design/test-utils-core/dom';
+import styles from '../../../content-layout/styles.selectors.js';
+
+export default class ContentLayoutWrapper extends ComponentWrapper<HTMLDivElement> {
+  static rootSelector: string = styles.layout;
+
+  findHeader(): ElementWrapper | null {
+    return this.find(`.${styles.header}`);
```
timogasda, line 10: `this.findByClassName` would be more elegant here :)

src/content-layout/__integ__/content-layout.test.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  async toggleHeaderPresent() {
+    const headerPresentCheckbox = await this.browser.$('#header-present-checkbox');
+    await headerPresentCheckbox.click();
+  }
+}
+
+describe('ContentLayout component', () => {
+  test(
+    'It renders the overlap by default',
```
timogasda, line 22: very minor: I believe test names should start lower-case. Actually, you probably want to remove the "It" entirely, so it reads like a sentence. That's why the `it` synonym for `test` exists. For example, `it("renders the overlap by default")`

src/content-layout/__integ__/content-layout.test.ts
```diff
@@ -0,0 +1,55 @@
...
+// SPDX-License-Identifier: Apache-2.0
+import { BasePageObject } from '@cloudscape-design/browser-test-tools/page-objects';
+import useBrowser from '@cloudscape-design/browser-test-tools/use-browser';
+import styles from '../../../lib/components/content-layout/styles.selectors.js';
+
+class ContentLayoutPageObject extends BasePageObject {
+  async isOverlapEnabled() {
+    const backgroundElement = await this.browser.$(`.${styles.background}`);
+    const classes = await backgroundElement.getAttribute('class');
+    return !classes.includes(styles['is-overlap-disabled']);
```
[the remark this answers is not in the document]
DaemonCahill: Given the behavior of the component the above makes more sense to me, but I think @connorlanigan should weigh in on both of these test threads.

src/content-layout/__integ__/content-layout.test.ts
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  async toggleHeaderPresent() {
+    const headerPresentCheckbox = await this.browser.$('#header-present-checkbox');
+    await headerPresentCheckbox.click();
+  }
+}
+
+describe('ContentLayout component', () => {
```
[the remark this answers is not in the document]
DaemonCahill: Deferring to @connorlanigan on this one as well since I am new to testing.

src/content-layout/__integ__/content-layout.test.ts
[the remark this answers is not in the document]
timogasda: I think those two would be equivalent, right?

You currently look for an element with `.background`, and then check if the classes of that element include the class `is-overlap-disabled`. 

Shouldn't that be equivalent to just looking for the selector `.background.is-overlap-disabled`?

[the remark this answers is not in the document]
connorlanigan: They are not fully equivalent: if the background element does not render for some reason (e.g. an error on the page), a  merged selector would not detect that, and instead it would fail silently by returning that the overlap is enabled.

src/content-layout/__integ__/content-layout.test.ts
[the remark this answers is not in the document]
connorlanigan: Initially, my idea behind adding integ tests here was to make sure that the interaction with the AppLayout context works properly in a more "complete" environment than the unit test environment. However, I realised now that the interaction between these two components is mainly in one direction (from ContentLayout to AppLayout), and the effect of that will be covered in the screenshot tests.
I'd be okay with replacing these integ tests with unit tests.

src/content-layout/styles.scss
```diff
@@ -2,19 +2,24 @@
...
 */
-@use '../../styles/' as styles;
-@use '../../styles/tokens' as awsui;
+@use '../internal/styles/' as styles;
+@use '../internal/styles/tokens' as awsui;
 
+/*
+Pass through the header and child content if not rendering in 
+visual refresh mode. The padding is still added to the header 
+in this scenario because otherwise the the header and child
```
connorlanigan, line 11: typo: "the the"

src/content-layout/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+  const [overlapContainerQuery, overlapElement] = useContainerQuery(rect => rect.height);
+
+  useLayoutEffect(
+    function handleDynamicOverlapHeight() {
+      if (isVisualRefresh) {
+        setDynamicOverlapHeight(overlapContainerQuery ?? 0);
+      }
+    },
+    [isVisualRefresh, overlapContainerQuery, setDynamicOverlapHeight]
+  );
```
connorlanigan, line 41: Here we also need to reset the value before unmounting. We can use the new `useDynamicOverlap` hook introduced in https://github.com/cloudscape-design/components/pull/84

connorlanigan (changes requested):

src/content-layout/__integ__/content-layout.test.ts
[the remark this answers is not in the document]
timogasda: Ah, okay. So what you're saying is that you need the first line (`await this.browser.$`) to throw in case that element doesn't exist?

src/content-layout/index.tsx
DaemonCahill: Agreed I was just waiting on that PR to merge.

src/content-layout/__integ__/content-layout.test.ts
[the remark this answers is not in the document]
connorlanigan: Yes

src/content-layout/__integ__/content-layout.test.ts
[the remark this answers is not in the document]
connorlanigan: I've now added a commit that replaces the integ tests with unit tests

timogasda (approved):

src/content-layout/index.tsx
```diff
@@ -2,29 +2,43 @@
...
   const isVisualRefresh = useVisualRefresh(rootElement);
-  const isOverlapDisabled = !children || !header || disableOverlap;
 
-  const overlapElement = useDynamicOverlap();
+  /**
+   * Disable the overlap if the component is missing either a header or child
+   * content. If the component is not using visual refresh then the overlap
+   * will not be displayed at all. This is handled in the CSS not the JavaScript.
+   */
+  const isOverlapDisabled = !children || !header || disableOverlap;
```
FlorianDr, line 32: Should this also effect the `useDynamicOverlap` hook?

DaemonCahill: It shouldn't. Even if the overlap is disabled within the content layout component, it still needs to extend the background from the app layout component down to the overlap element.

DaemonCahill:
@connorlanigan @timogasda Can I get an approval on this assuming there are no outstanding changes?

src/app-layout/interfaces.ts
```diff
@@ -114,12 +114,14 @@ export interface AppLayoutProps extends BaseComponentProps {
 
   /**
    * Top area of the page content.
+   * **Deprecated**, replaced by the content layout `header` slot.
```
connorlanigan, line 117: "content layout" might not have a clear meaning for users who see this. I would suggest to explicitly mention that there is a component with that name. For example:
> **Deprecated**, replaced by the `header` slot of the ContentLayout component.

(same applies to the `disableContentHeaderOverlap` property)

connorlanigan (changes requested):

timogasda (approved):

connorlanigan (approved):

PR #75, opened 2022-07-26
Add embedded variant of date picker

zshakeel-amazon:
### Description

Adds a new variant of the date picker that is not within a dropdown component. This allows for this variant of the date picker to be embedded within other dropdown components, avoiding dropdowns within dropdowns.

### How has this been tested?

Manual testing. Added demo page for this variant

### Related Links

SIM: AWSUI-9912

<details>

#### Correctness

#### Security

#### Testing

</details>


src/date-picker/embedded.tsx
```diff
@@ -0,0 +1,43 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import React from 'react';
+import { DatePickerProps } from './interfaces';
+import Calendar, { DayIndex } from './calendar';
+import { memoizedDate } from './calendar/utils/date';
+import { CalendarTypes } from './calendar/definitions';
+
+export default function DatePickerEmbedded(
```
pan-kot, line 9: Why isn't it a React component?

zshakeel-amazon: Updated in revision

src/date-picker/embedded.tsx
```diff
@@ -0,0 +1,60 @@
...
+      previousMonthLabel={previousMonthAriaLabel}
+      todayAriaLabel={todayAriaLabel}
+      onChangeMonth={onChangeMonthHandler}
+      onSelectDate={onSelectDateHandler}
+      onFocusDate={onDateFocusHandler}
+    />
+  );
+};
+
+export interface DatePickerEmbeddedProps {
```
pan-kot, line 45: I think it does make sense to derive this props from the DatePicker like:

```
export type DatePickerEmbeddedProps = Exclude<DatePickerProps, 'name' | 'ariaLabel' | ...>
```

This would make it easier to update and to see what properties are unique, like calenderRef, which we'd need to turn to some internal one if decided to share the embedded component as a standalone.

pan-kot (approved):

src/date-picker/embedded.tsx
```diff
@@ -0,0 +1,70 @@
...
+      previousMonthLabel={previousMonthAriaLabel}
+      todayAriaLabel={todayAriaLabel}
+      onChangeMonth={onChangeMonthHandler}
+      onSelectDate={onSelectDateHandler}
+      onFocusDate={onDateFocusHandler}
+    />
+  );
+};
+
+export type DatePickerEmbeddedProps = Omit<
```
just-boris, line 55: How about flipping the hierarchy?

```js
interface BaseDatePickerProps {}

interface DatePickerProps extends BaseDatePickerProps {}

interface EmbeddedDatePickerProps extends BaseDatePickerProps {}
```

src/date-picker/embedded.tsx
```diff
@@ -0,0 +1,65 @@
...
+    locale,
+    value,
+  });
+
+  return (
+    <Calendar
+      selectedDate={memoizedDate('value', selectedDate)}
+      focusedDate={memoizedDate('focused', focusedDate)}
+      displayedDate={memoizedDate('displayed', displayedDate)}
+      locale={normalizedLocale}
```
pan-kot, line 33: It looks like the locale is normalized inside Calendar. Why the `normalizedLocale` is needed?

src/date-picker/embedded.tsx
```diff
@@ -0,0 +1,65 @@
...
+  | 'readOnly'
+  | 'autoFocus'
+  | 'ariaLabel'
+  | 'ariaRequired'
+  | 'onFocus'
+  | 'onBlur'
+  | 'onChange'
+> & {
+  calendarHasFocus: boolean;
+  focusedDate: string | null;
```
pan-kot, line 62: Can focusedDate/onDateFocusHandler be moved inside Calendar?

src/date-picker/embedded.tsx
```diff
@@ -0,0 +1,65 @@
...
+  | 'disabled'
+  | 'readOnly'
+  | 'autoFocus'
+  | 'ariaLabel'
+  | 'ariaRequired'
+  | 'onFocus'
+  | 'onBlur'
+  | 'onChange'
+> & {
+  calendarHasFocus: boolean;
```
pan-kot, line 61: What is the purpose of calendarHasFocus property, is it needed for the embedded component?

src/date-picker/embedded.tsx
zshakeel-amazon: Thanks, this should just be `locale`

src/date-picker/embedded.tsx
zshakeel-amazon: The calendar component requires it as a prop. Unless we can perhaps default it to true or false here?

src/date-picker/embedded.tsx
zshakeel-amazon: Yes it seems so. Updated in revision

src/date-picker/calendar/index.tsx
```diff
@@ -120,7 +133,7 @@ const Calendar = React.forwardRef(
     }, [calendarHasFocus, baseDate, isDateEnabled]);
 
     if (calendarHasFocus && !focusedDate) {
-      focusedDate = selectFocusedDate(selectedDate, baseDate);
+      selectFocusedDate(selectedDate, baseDate) && setFocusedDate(selectFocusedDate(selectedDate, baseDate));
```
pan-kot, line 136: This line is a bit confusing to read, would it make sense to change it to:

```
const nextFocusedDate = selectFocusedDate(selectedDate, baseDate)
if (nextFocusedDate) {
  setFocusedDate(nextFocusedDate)
}
```

Or maybe it will be easier to just call setFocusedDate inside selectFocusedDate:

```
if (calendarHasFocus && !focusedDate) {
  selectFocusedDate(selectedDate, baseDate)
}

function selectFocusedDate(selected: Date | null, baseDate: Date): void {
   ...setFocusedDate(...)
}
```

src/date-picker/embedded.tsx
```diff
@@ -0,0 +1,56 @@
...
+  value,
+  locale = '',
+  startOfWeek,
+  isDateEnabled,
+  nextMonthAriaLabel,
+  previousMonthAriaLabel,
+  todayAriaLabel,
+  calendarHasFocus,
+}: DatePickerEmbeddedProps) => {
+  const { displayedDate, selectedDate, onChangeMonthHandler, onSelectDateHandler } = useDatePicker({
```
pan-kot, line 19: According to its signature, the useDatePicker takes value and onChange. I bet this embedded component is now readonly.

src/date-picker/embedded.tsx
pan-kot: What is that logic? Do we need it for the embedded component?

src/date-picker/embedded.tsx
zshakeel-amazon: Right, I'll need to add the `onChange` here (and make it required)

src/date-picker/embedded.tsx
zshakeel-amazon: Great suggestion. Updating in revision

src/date-picker/embedded.tsx
zshakeel-amazon: Seems it is not needed based on offline discussion. Removing in revision

pan-kot:
The "Focusable element after the date picker" is the test page is not fucusable.

src/date-picker/index.tsx
```diff
@@ -233,24 +189,27 @@ const DatePicker = React.forwardRef(
         >
           {isDropDownOpen && (
             <>
-              {calendarHasFocus && <TabTrap focusNextCallback={focusCurrentDate} />}
+              {<TabTrap focusNextCallback={focusCurrentDate} />}
```
pan-kot, line 192: the braces {} are no longer needed

src/date-picker/calendar/index.tsx
```diff
@@ -109,18 +120,20 @@ const Calendar = React.forwardRef(
     }, [baseDate, focusedDate, gridHasFocus]);
 
     useEffect(() => {
-      const calendarShouldHaveFocus = calendarHasFocus;
       const calendarActuallyHasFocus = elementRef.current?.contains(document.activeElement);
 
-      if (calendarShouldHaveFocus && !calendarActuallyHasFocus) {
+      if (!calendarActuallyHasFocus) {
         elementRef.current?.focus();
```
pan-kot, line 126: The calendar is focused when the focus goes away from the grid. This is actually not needed. I removed this useEffect, onGridBlur and onGridFocus and everything works as expected in the default date picker. All tests are passing. Can you confirm?

I think this code was to provide a tab trapping logic, but then another tab trap solution was added w/o proper cleanup.

zshakeel-amazon: Can confirm behaviour is the same. Removing these also fixes the other issue with the focusable element not focusable after the date picker.

pages/date-picker/embedded.page.tsx
```diff
@@ -0,0 +1,28 @@
...
+    <Box padding="s">
+      <h1>Date picker embedded version</h1>
+      <Link id="focusable-element-before-date-picker">Focusable element before the date picker</Link>
+      <br />
+      <DatePickerEmbedded
+        value={value}
+        locale={'en-EN'}
+        previousMonthAriaLabel={'Previous month'}
+        nextMonthAriaLabel={'Next month'}
+        todayAriaLabel={'TEST TODAY'}
```
pan-kot, line 20:
```suggestion
        todayAriaLabel="Today"
```

src/date-picker/__integ__/date-picker-embedded.test.ts
```diff
@@ -0,0 +1,33 @@
...
+describe('Date picker embedded interactions', () => {
+  const setupTest = (testFn: (page: DatePickerPage) => Promise<void>) => {
+    return useBrowser(async browser => {
+      const page = new DatePickerPage(createWrapper().findDatePicker().getElement(), browser);
+      await browser.url('#/light/date-picker/embedded');
+      await testFn(page);
+    });
+  };
+
+  test(
```
pan-kot, line 16: I would make the tests a bit more general. The first one can start from focusable-element-before and tab until the focusable-element-after. The second one -- vice versa.

Also, there is certain focusing logic still inside the calendar which is focusing the date buttons when using keyboard navigation. It would be nice to replicate the following test here: https://github.com/cloudscape-design/components/blob/main/src/date-picker/__integ__/date-picker.test.ts#L107

pan-kot (approved):
LGTM

just-boris (approved):

PR #76, opened 2022-07-26
fix(icon): Improve default size when server rendering

avinashbot:
### Description

Icons don't have a contextual height when used with `size="inherit"` until TTI.

### How has this been tested?

Only locally. Do it yourself:

```
<div dangerouslySetInnerHTML={{
  __html: ReactDOMServer.renderToString(<Icon name="external" size="inherit" />)
}} />
```

### Documentation changes

- [x] _No._

### Related Links

AWSUI-18521

<details>

#### Correctness

- [x] _Changes were manually tested for accessibility, see [accessibility guidelines](https://cloudscape.design/foundation/core-principles/accessibility/).

#### Security

- [x] _If the code handles URLs: all URLs are validated through [the `checkSafeUrl` function](https://github.com/cloudscape-design/components/blob/main/src/internal/utils/check-safe-url.ts).

#### Testing

Only manual testing for now, we probably need to do SSR stuff in a more comprehensive way.

</details>

### DIFF
--- a/src/icon/__tests__/icon.test.tsx
+++ b/src/icon/__tests__/icon.test.tsx
@@ -39,10 +39,6 @@ describe('Icon Component', () => {
       const { container } = render(<Icon size="inherit" name="settings" />);
       expect(container.firstElementChild).toHaveClass(styles['icon-flex-height']);
     });
-    it('renders with inline style for height', () => {
-      const { container } = render(<Icon size="inherit" name="settings" />);
-      expect(container.firstElementChild).toHaveAttribute('style', 'height: 0px;');
-    });
   });
 
   describe('custom icons', () => {
--- a/src/icon/internal.tsx
+++ b/src/icon/internal.tsx
@@ -16,19 +16,22 @@ type InternalIconProps = IconProps &
     badge?: boolean;
   };
 
-function iconSizeMap(height: number) {
-  if (height) {
-    if (height >= 50) {
-      return 'large';
-    } else if (height >= 36) {
-      return 'big';
-    } else if (height >= 24) {
-      return 'medium';
-    } else if (height <= 16) {
-      return 'small';
-    } else {
-      return 'normal';
-    }
+function iconSizeMap(height: number | null) {
+  if (height === null) {
+    // This is the best guess for the contextual height while server rendering.
+    return 'normal';
+  }
+
+  if (height >= 50) {
+    return 'large';
+  } else if (height >= 36) {
+    return 'big';
+  } else if (height >= 24) {
+    return 'medium';
+  } else if (height <= 16) {
+    return 'small';
+  } else {
+    return 'normal';
   }
 }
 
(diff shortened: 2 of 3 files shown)


src/icon/internal.tsx
```diff
@@ -17,18 +17,18 @@ type InternalIconProps = IconProps &
...
-    }
+  if (height === 0) {
+    return 'normal';
+  } else if (height >= 50) {
+    return 'large';
+  } else if (height >= 36) {
+    return 'big';
+  } else if (height >= 24) {
+    return 'medium';
+  } else if (height <= 16) {
```
FlorianDr, line 28: minor: Shall we just extend this condition to have `height <= 16 && height > 0`. It would mean -1 branch + smaller change.

avinashbot: Just a matter of opinion; I wanted to have the "exceptional" situation to short-circuit the regular if statement cascade.

But it's pretty minor, -2 lines of code is also nice too :P

src/icon/internal.tsx
```diff
@@ -17,18 +17,16 @@ type InternalIconProps = IconProps &
...
-    } else if (height >= 36) {
-      return 'big';
-    } else if (height >= 24) {
-      return 'medium';
-    } else if (height <= 16) {
-      return 'small';
-    } else {
-      return 'normal';
-    }
+  if (height >= 50) {
```
FlorianDr, line 20: Do we want to add a test for using `renderToStaticMarkup` to confirm the behavior?

just-boris: +1. Please check how other `@jest-environment node` tests work and let's do the same here: https://github.com/cloudscape-design/components/search?q=jest-environment+node

src/icon/__tests__/icon.test.tsx
```diff
@@ -39,10 +39,6 @@ describe('Icon Component', () => {
       const { container } = render(<Icon size="inherit" name="settings" />);
       expect(container.firstElementChild).toHaveClass(styles['icon-flex-height']);
     });
-    it('renders with inline style for height', () => {
-      const { container } = render(<Icon size="inherit" name="settings" />);
-      expect(container.firstElementChild).toHaveAttribute('style', 'height: 0px;');
```
just-boris, line 44: Can we keep this test to confirm that the parent size is reflected properly (even if it is zero in jsdom)?

src/icon/internal.tsx
avinashbot: Ah, sweet. I didn't know we had a precedent for this. I wasn't a huge fan of "faking SSR" in a JSDOM environment but this is definitely cleaner.

src/icon/__tests__/icon.test.tsx
avinashbot: JSDOM doesn't do inheritance; so the inferred size is always going to be 0. The PR changes it slightly so that height is unset when getComputedStyle doesn't return a value (because we want the height to come from the CSS class, not the inline style).

just-boris: Fine with me, your call

avinashbot: Also, I'm testing that it does the opposite in the SSR test; actual testing of this logic can only reliably happen in screenshot or integ tests (where we do have tests for it).

just-boris (approved):

PR #77, opened 2022-07-26
Make more concise

daniijal:
Made a few changes that make the reading easier 😄


connorlanigan:
Thanks for your interest in the project. We prefer to keep the current Readme content.

PR #79, opened 2022-07-26
chore: Replace useLayoutEffect during server rendering to reduce excessive warnings

avinashbot:
- chore: Upgrade ESLint to 8.20 and unblock eslint plugins
- chore: Replace useLayoutEffect during server rendering to reduce excessive warnings

### Description

Very low-hanging SSR fruit. SSRing or SSGing any components with a `useLayoutEffect` causes React to print out a massive warning per instance that crowds out the render logs, leading to less than ideal DX and making real issues harder to debug. This isn't going to be fixed on React's side, so the best place to fix this is on our side.

A pretty common solution to this is to create a custom hook that runs useEffect on the server and useLayoutEffect on the client, which this PR introduces. Also had to upgrade ESLint because one of the config options wasn't supported until recently (`importNames`).

I'm sure this'll inspire some healthy debate. We can discuss this as a team tomorrow :)

### How has this been tested?

Added a unit test for the hook, as well as eslint checks to prevent regressions in this area.

### Documentation changes

- [x] _No._

### Related Links

<details>

#### Correctness

- [x] _Changes were manually tested for accessibility, see [accessibility guidelines](https://cloudscape.design/foundation/core-principles/accessibility/).

#### Security

- [x] _If the code handles URLs: all URLs are validated through [the `checkSafeUrl` function](https://github.com/cloudscape-design/components/blob/main/src/internal/utils/check-safe-url.ts).

#### Testing

</details>


src/property-filter/controller.ts
```diff
@@ -91,7 +91,7 @@ export const parseText = (
       value: negatedGlobalQuery[2],
     };
   }
-  const { property } = filteringProperties?.reduce<{
+  const { property } = (filteringProperties ?? []).reduce<{
```
avinashbot, line 94: This was a new bug uncovered during the dependency update (it's the only one).

.eslintrc
```diff
@@ -160,6 +160,11 @@
               {
                 "group": ["lodash", "lodash/*"],
                 "message": "lodash is a commonjs module, which breaks webpack esm optimizations."
+              },
+              {
+                "group": ["react"],
```
gethinwebster, line 165: I guess this would still not stop people from using `React.useLayoutEffect`? But hopefully that would get caught in PR as incorrect style anyway

gethinwebster (commented):
I'm not quite convinced about this - it feels like we're hiding away what are real warnings if we ever want to properly support SSR. It's potentially even useful for customers to see the warnings, to know that the components they're using aren't fully SSR-compliant?

As you say - a good topic for team discussion...

.eslintrc
avinashbot: Yeah, ESLint isn't super comprehensive here. I could probably throw in a `no-restricted-properties` for `React.useLayoutEffect`, but it's just linting, there are always ways around it. But this should catch it 99% of the time.

gethinwebster:
I think what I particularly dislike about the solution is that it's non-transparently assigning a no-op function in SSR world (i.e. unless you know that `useEffect` won't be run, you might think that the code will run in both places). So maybe for transparency, assigning an explicit no-op function when `window` is undefined could be clearer to those reading the `useIsomorphicLayoutEffect` source.

PR #84, opened 2022-07-27
refactor: Add useDynamicOverlap hook

### LINKED ISSUE #83: [Bug]: dynamicOverlapHeight is not reset to baseline after table component unmounts
### Browser

Chrome

### Package version

3.0.1

### React version

18.2.0

### Description

I am using a Table component with a variant set to "full-page". I am using react-router to route to a detail page and I notice the dynamicOverlapHeight  is retained when the detail page loads (--awsui-overlap-height-ujkcpl is present in <main>). When the Table (or any component that overrides these settings) unmounts, it should reset all of the default settings in the Layout to baseline.

Table uses an InternalContainer which appears to be the culprit: https://github.com/cloudscape-design/components/blob/main/src/container/internal.tsx#L59

I haven't tested but I assume to fix this we should change the layout callback as follows:

```ts
useLayoutEffect(
    function handleDynamicOverlapHeight() {
      if (hasDynamicHeight) {
        setDynamicOverlapHeight(overlapContainerQuery ?? 0);
      }
      return () => setDynamicOverlapHeight(0);
    },
    [hasDynamicHeight, overlapContainerQuery, setDynamicOverlapHeight]
  );
```

This should work as this code only overrides the offset if it is > 0: https://github.com/cloudscape-design/components/blob/7433543fcd2796d6398b67867d25b637c1a2b6de/src/app-layout/visual-refresh/layout.tsx#L125

### Source code

N/A

### Reproduction

- Create a table with `full-page` variant.
- Add a link from the table to a new page using something like react-router.
- The new page will preserve the offset set by the previous page which is not correct.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/cloudscape-design/components/blob/main/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/cloudscape-design/components/issues) for duplicate problems

FlorianDr:
### Description

Introduces new `useDynamicOverlap` hook, which centralizes the needed code into a common utility hook. Further, it also fixes issues, where the dynamic overlap will not be reset. 

closes #83 

### How has this been tested?

Unit tests

### Documentation changes

- [X] _No._

### Related Links

<details>

#### Correctness

#### Security

#### Testing

</details>

### DIFF
--- a/src/app-layout/visual-refresh/hooks/use-dynamic-overlap.ts
+++ b/src/app-layout/visual-refresh/hooks/use-dynamic-overlap.ts
@@ -1,11 +1,10 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-import { RefObject, useContext, useLayoutEffect } from 'react';
+import { useContext, useLayoutEffect } from 'react';
 
 import { AppLayoutContext } from '../context';
 import { useContainerQuery } from '../../../internal/hooks/container-queries';
-import { useVisualRefresh } from '../../../internal/hooks/use-visual-mode';
 
 export interface UseDynamicOverlapProps {
   /**
@@ -24,9 +23,8 @@ export function useDynamicOverlap(props?: UseDynamicOverlapProps) {
   const disabled = props?.disabled ?? false;
   const { setDynamicOverlapHeight } = useContext(AppLayoutContext);
   const [overlapContainerQuery, overlapElementRef] = useContainerQuery(rect => rect.height);
-  const isVisualRefresh = useVisualRefresh(overlapElementRef as RefObject<HTMLElement>);
 
-  const hasOverlap = isVisualRefresh && !disabled && overlapContainerQuery !== null;
+  const hasOverlap = !disabled && overlapContainerQuery !== null;
   useLayoutEffect(
     function handleDynamicOverlapHeight() {
       setDynamicOverlapHeight(hasOverlap ? overlapContainerQuery : 0);
--- a/src/app-layout/visual-refresh/hooks/use-dynamic-overlap.ts
+++ b/src/app-layout/visual-refresh/hooks/use-dynamic-overlap.ts
@@ -1,0 +1,42 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { RefObject, useContext, useLayoutEffect } from 'react';
+
+import { AppLayoutContext } from '../context';
+import { useContainerQuery } from '../../../internal/hooks/container-queries';
+import { useVisualRefresh } from '../../../internal/hooks/use-visual-mode';
+
+export interface UseDynamicOverlapProps {
+  /**
+   * Disables hook if not needed. By default, `false`.
+   */
+  disabled?: boolean;
+}
+
+/**
+ * Observes the height of an element referenced by the returning ref and sets the value as overlapping
+ * height for the surrounding AppLayout component.
+ * @param props.disabled disables hook if not applicable
+ * @returns ref to be measured as overlapping height
+ */
+export function useDynamicOverlap(props?: UseDynamicOverlapProps) {
+  const disabled = props?.disabled ?? false;
+  const { setDynamicOverlapHeight } = useContext(AppLayoutContext);
+  const [overlapContainerQuery, overlapElementRef] = useContainerQuery(rect => rect.height);
+  const isVisualRefresh = useVisualRefresh(overlapElementRef as RefObject<HTMLElement>);
+
+  const hasOverlap = isVisualRefresh && !disabled && overlapContainerQuery !== null;
+  useLayoutEffect(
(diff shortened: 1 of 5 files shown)


src/app-layout/visual-refresh/hooks/use-dynamic-overlap.ts
```diff
@@ -0,0 +1,40 @@
...
+ */
+export function useDynamicOverlap(props?: UseDynamicOverlapProps) {
+  const disabled = props?.disabled ?? false;
+  const { setDynamicOverlapHeight } = useContext(AppLayoutContext);
+  const [overlapContainerQuery, overlapElementRef] = useContainerQuery(rect => rect.height);
+
+  const hasOverlap = !disabled && overlapContainerQuery !== null;
+  useLayoutEffect(
+    function handleDynamicOverlapHeight() {
+      setDynamicOverlapHeight(hasOverlap ? overlapContainerQuery : 0);
```
DaemonCahill, line 30: Technically this will work because the `dynamicOverlapHeight` is instantiated as a zero value in the `AppLayout` context and the logic for overriding the custom property is:

`dynamicOverlapHeight > 0 && { [customCssProps.overlapHeight]: '${dynamicOverlapHeight}px' })`

So, if the `disabled` property in this hook is `true` then we will be calling the setter in the context to update the value to zero, which it likely already is. I'm wondering if we even need this `disabled` property. Ideally the component that implements the hook will only use it under circumstances where it _isn't_ disabled which would make this property irrelevant.

Is there a use case for the `disabled` property that I'm missing? Similarly, if internally the `overlapContainerQuery` is `null` we should not even call the setter, I think.

src/container/internal.tsx
```diff
@@ -45,25 +44,10 @@ export default function InternalContainer({
...
-
-  useLayoutEffect(
-    function handleDynamicOverlapHeight() {
-      if (hasDynamicHeight) {
-        setDynamicOverlapHeight(overlapContainerQuery ?? 0);
-      }
-    },
-    [hasDynamicHeight, overlapContainerQuery, setDynamicOverlapHeight]
-  );
+  const overlapElement = useDynamicOverlap({ disabled: !hasDynamicHeight });
```
DaemonCahill, line 50: Referencing the above comment on the disabled prop...in this instance we can lift the `!hasDynamicHeight` logic and either have a generic ref or use no ref at all. In which case we can remove the `disabled` property entirely from the hook.

DaemonCahill:
@FlorianDr Great work! I added a couple comments pertaining to the disabled property.

src/app-layout/visual-refresh/hooks/use-dynamic-overlap.ts
FlorianDr: If we don't call the setter when `disabled` or `overlapContainerQuery` flips, aren't we missing update to the "reactive" properties like we do right now with the lack of a cleanup function? Keep in mind that this function is not just called on first render of the overlap component but whenever state or properties change.

src/container/internal.tsx
FlorianDr: The optimization is nice, especially since `<Container />` is widely used. One comment though: If we move the decision about "disabled" to attaching or not attaching a ref, we start to spread logic across multiple places in the render function. I have seen it as a convention to have a `disabled` flag on hooks, since we can not just not call them. I will ask for more feedback.

just-boris: Tried to prototype this suggestion, it did not work in practice

```js
const elementRef = useRef();
const conditionalRef = hasOverlap ? elementRef : null;
useDynamicOverlap(conditionalRef);

<div ref={conditionalRef}>
```

It works when transitioning "disabled -> enabled", but in "enabled -> disabled" the container query does not get removed.

If we want to fix it, we need to change `useContainerQuery` signature, allow conditionally disabling this hook

DaemonCahill: This is an interesting problem. It seems like there are three scenarios:

1. Always add a container query to the internal container DOM and call the setter with a zero value if disabled.
2. Remove the `disabled` prop and have a conditional ref at implementation in this component. This will not _add_ unused container queries to the DOM but it will _leave_ unused container queries on the DOM once they are instantiated.
3. Make the adjustment to the `useContainerQuery` hook so we can remove conditional refs at implementation and unused container queries from the overlap hook.

I don't know enough about the `useContainerQuery` to make a judgement here, though it seems like a better long term option than adding a container query to every instance of the internal container.

What do you both think?

src/app-layout/visual-refresh/hooks/use-dynamic-overlap.ts
DaemonCahill: Continuing this conversation in the other thread...

src/container/internal.tsx
just-boris: I think, the current change does not introduce regressions. Because it is also a fix for a bug report, I suggest we proceed merging the current state.

Possible improvements to make the hook smart and conditional can be done as a follow-up

DaemonCahill: Sounds good to me. 👍

DaemonCahill (approved):

PR #109, opened 2022-08-02
update AbstractSwitch focus outline elements to spans rather than divs to ensure HTML validity

jorycunningham:
### Description

Changes instances in Toggle component where divs were nested in labels or spans to ensure valid HTML

### How has this been tested?

* Built and reviewed rendered HTML
* Ran axe-core against demo permutations page
* Checked rendered HTML using the W3C Nu validator

### Documentation changes

- [X] _No._

### Related Links

See AWSUI-18915

<details>

#### Correctness

- [ X] _Changes are backward-compatible if not indicated, see [`CONTRIBUTING.md`](CONTRIBUTING.md#public-apis)._
- [ X] _Changes do not include unsupported browser features, see [`CONTRIBUTING.md`](CONTRIBUTING.md#browsers-support)._
- [X ] _Changes were manually tested for accessibility, see [accessibility guidelines](https://cloudscape.design/foundation/core-principles/accessibility/)._

#### Security

#### Testing

</details>


pan-kot (commented):
It would be nice to setup some regression for that. If there is no built-in rule in axe, would it be possible to add a custom one?

In the very least, it would be beneficial to at-least add comments next to the updated lines with a link to the related accessibility guidelines.

pan-kot (approved):

pan-kot:
> @pan-kot I've added comments per your suggestion. In terms of automated testing, I think a dedicated HTML validator, run globally, would be more effective than AXE here. Might we consider adding something like [HTML-Validate](https://html-validate.org/frameworks/jest.html)?

Sure, that might be a nice follow-up exercise. I will discuss it with the team.

YueyingLu:
Hi @jorycunningham the focus style is gone with the change. If you open the for example checkbox/focus-test page, click focus button then tab, the focus ring doesn't appear anymore. It is the same for Toggle and RadioGroup. https://github.com/cloudscape-design/components/blob/main/src/checkbox/styles.scss#L25

jorycunningham:
@YueyingLu @avinashbot

I've updated this PR based on our conversations, specifically:
* Change css element selectors that rely on the HTML structure of AbstractSwitch to use spans, not divs
* Add a comment to the tsx file for AbstractSwitch
* Update the PR title to better reflect scope of this work

YueyingLu (approved):

src/internal/components/abstract-switch/index.tsx
```diff
@@ -80,7 +80,8 @@ export default function AbstractSwitch({
         className={styles['label-wrapper']}
         aria-disabled={disabled ? 'true' : undefined}
       >
-        <div className={clsx(styles.control, controlClassName)}>
+        {/*Using span, not div for HTML validity*/}
```
pan-kot, line 83: The comment is no longer needed, we added the html-validator to cover that, see: https://github.com/cloudscape-design/components/blob/main/src/toggle/__tests__/toggle.test.tsx#L95

src/internal/components/abstract-switch/index.tsx
```diff
@@ -90,27 +91,33 @@ export default function AbstractSwitch({
...
             'aria-label': ariaLabel,
           })}
-          {/*An empty div to display the outline, because the native control is invisible*/}
-          <div className={styles.outline} />
-        </div>
-        <div className={clsx(styles.content, !label && !description && styles['empty-content'])}>
+          {/*
+          An empty element to display the outline, because the native control is invisible.
+          Using span, not div for HTML validity.
+          Note: There is a CSS selector in src/toggle/styles.scss that relies on a span element selector to show focus.
```
pan-kot, line 97: Would it be possible to add an integration test for that to fail if the selector or layout would change?

src/checkbox/styles.scss
```diff
@@ -22,7 +22,7 @@ div.root {
 .input {
```
pan-kot, line 22: Can this class be moved from checkbox, radio and toggle to the abstract-switch itself?

jorycunningham: I agree that this seems repetitive, but If we put this in the abstract switch I think the CSS selector would become more fragile since we need the scope of the component to establish the dynamic class name (e.g.  styles.input).  We would have to rely on a static class or strictly on element selectors.

I think the CSS to establish the focus needs to be scoped to the component that "owns" the HTML that is delivering the style.

src/internal/components/abstract-switch/index.tsx
jorycunningham: I added this because I was concerned that the element selector was a bit fragile. I think what we want to assert is that there is a visible focus ring when the underlying control is focused or more to point that there is relationship between the CSS, HTML, and JS (for the data-focus-visible attribute) that results in a focus ring. As such is a visual regression appropriate here?

Can we orchestrate backstopJS to programmatically focus the inputs and take the visual regression snapshot?

YueyingLu: We have integration test of focus-test but it is not a part of visual regression test on github. Currently on github visual regression test only on permutation pages.

src/internal/components/abstract-switch/index.tsx
jorycunningham: @YueyingLu done!

src/checkbox/styles.scss
pan-kot: Ok, we can handle that as a follow-up.

pan-kot (approved):

PR #116, opened 2022-08-03
Date range picker refactoring (embedded)

zshakeel-amazon:
### Description

Adds a new variant of the date range picker that is not within a dropdown component. This allows for this variant of the date range picker to be embedded within other dropdown components, avoiding dropdowns within dropdowns.

### How has this been tested?

Existing tests. Added additional integration tests for embedded variant. Added demo page.

### Documentation changes

No

### Related Links

SIM: AWSUI-9912

<details>

#### Correctness

#### Security

#### Testing

</details>


src/date-range-picker/index.tsx
```diff
@@ -121,6 +105,12 @@ const DateRangePicker = React.forwardRef(
     }: DateRangePickerProps,
     ref: Ref<DateRangePickerProps.Ref>
   ) => {
+    const { formatValue } = useDateRangePicker({
```
michaeldowseza, line 108: Can this function not be exported separately. It seems weird to have a hook here that creates unnecessary and unused state.

pages/date-range-picker/embedded.page.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { getBrowserTimezoneOffset } from '~components/date-range-picker/time-offset';
+
+export default function DatePickerScenario() {
+  const [value, setValue] = useState<DateRangePickerProps['value']>({
+    type: 'absolute',
+    startDate: '2018-01-09T12:34:56Z',
+    endDate: '2018-01-19T15:30:00',
+  });
+
+  const timeOffset = getBrowserTimezoneOffset();
```
pan-kot, line 16: Why the timeOffset is defined explicitly for the embedded example?

src/date-range-picker/calendar/header/index.tsx
```diff
@@ -12,7 +12,6 @@ interface CalendarHeaderProps {
   onChangeMonth: (prev?: boolean) => void;
   previousMonthLabel: string;
   nextMonthLabel: string;
-  calendarHasFocus: boolean;
```
pan-kot, line 15: It would make sense to extract the refactorings that are applicable to the original components to separate PRs.

src/date-range-picker/index.tsx
```diff
@@ -266,7 +256,7 @@ const DateRangePicker = React.forwardRef(
           dropdownId={dropdownId}
         >
           {isDropDownOpen && (
-            <DateRangePickerDropdown
+            <DateRangePickerEditor
```
pan-kot, line 259: I thought the editor is being used by both embedded and original components. If that's not the case, I don't see a reason to make a dropdown -> editor rename.

src/date-range-picker/calendar/header/index.tsx
pan-kot: That includes calendarHasFocus and FocusLock changes

src/date-range-picker/embedded.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+import RelativeRangePicker from './relative-range';
+import ModeSwitcher from './mode-switcher';
+import { useDateRangePicker } from './use-date-range-picker';
+import { DateRangePickerBaseProps } from './interfaces';
+import { fireNonCancelableEvent } from '../internal/events/index.js';
+import { DateRangePickerProps, Focusable } from './interfaces';
+
+export interface DateRangePickerDropdownProps
+  extends Pick<
```
pan-kot, line 16: Since the interfaces are now split into BaseProps and Props -- it is still required to have a Pick? If a few props to be excluded that can probably be simplified with Omit.

src/date-range-picker/embedded.tsx
```diff
@@ -0,0 +1,129 @@
...
+    Required<DateRangePickerBaseProps>,
```
pan-kot, line 17: Why these props set as required? If some are optional in the original component I'd expect them to remain optional here.

zshakeel-amazon: In the original dropdown component, it seems we are making some of these optional props required anyway?: https://github.com/cloudscape-design/components/blob/main/src/date-range-picker/dropdown.tsx#L40

It seems many of these optional props are actually required but are defined as optional in the interfaces

src/date-range-picker/index.tsx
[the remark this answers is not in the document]
zshakeel-amazon: Yes it can. Will rather make it an export and remove the hook here

pages/date-range-picker/embedded.page.tsx
zshakeel-amazon: The embedded component does require a timeOffset. However we could instead set the offset within the component itself instead of here

src/date-range-picker/embedded.tsx
```diff
@@ -0,0 +1,129 @@
...
+              ref={focusRefs.default}
+              mode={rangeSelectionMode}
+              onChange={(mode: 'absolute' | 'relative') => {
+                setRangeSelectionMode(mode);
+              }}
+              i18nStrings={i18nStrings}
+            />
+          )}
+
+          {rangeSelectionMode === 'absolute' && (
```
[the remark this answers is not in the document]
zshakeel-amazon: Will look to simplify this in a follow up PR

src/date-range-picker/calendar/header/index.tsx
zshakeel-amazon: Separate PR: https://github.com/cloudscape-design/components/pull/121/checks

src/date-range-picker/embedded.tsx
zshakeel-amazon: We can definitely remove it

src/date-range-picker/calendar/header/index.tsx
zshakeel-amazon: Focus lock PR: https://github.com/cloudscape-design/components/pull/122

src/date-range-picker/embedded.tsx
```diff
@@ -0,0 +1,117 @@
...
+import RelativeRangePicker from './relative-range';
+import ModeSwitcher from './mode-switcher';
+import { useDateRangePicker } from './use-date-range-picker';
+import { DateRangePickerBaseProps } from './interfaces';
+import { fireNonCancelableEvent } from '../internal/events/index.js';
+import { DateRangePickerProps, Focusable } from './interfaces';
+import { getBrowserTimezoneOffset } from './time-offset';
+import { formatValue } from './use-date-range-picker.js';
+
+export interface DateRangePickerDropdownProps extends DateRangePickerBaseProps {
```
pan-kot, line 17: DateRangePickerEmbeddedProps

pan-kot (approved):

src/date-range-picker/interfaces.ts
```diff
@@ -69,68 +124,18 @@ export interface DateRangePickerProps extends BaseComponentProps, FormFieldValid
...
-   * The event `detail` contains the current value of the field.
-   */
-  onChange?: NonCancelableEventHandler<DateRangePickerProps.ChangeDetail>;
-
   /**
    * A function that defines whether a particular range is valid or not.
    *
    * Ensure that your function checks for missing fields in the value.
    */
   isValidRange: DateRangePickerProps.ValidationFunction;
```
pan-kot, line 132: Actually, why this is not supported for the embedded component?

pan-kot: And also startOfWeek

src/date-range-picker/embedded.tsx
```diff
@@ -0,0 +1,117 @@
...
+  startOfWeek: DayIndex;
+  isSingleGrid: boolean;
```
pan-kot, line 19: Why isSingleGrid is coming as an argument? It is calculated in the `index.tsx` as `useMobile()`, but the same can be done here.

src/date-range-picker/interfaces.ts
zshakeel-amazon: For the embedded component, we remove the validation section. `isValidRange` is used to determine the validation result which only gets used in the validation section. 

Do you think this should be added as a base prop? (Though it would not be used?)

src/date-range-picker/embedded.tsx
zshakeel-amazon: Agreed 👍

src/date-range-picker/interfaces.ts
zshakeel-amazon: For start of week, we can actually move the normalize start of week behaviour into the calendar itself (like we have done with date picker). Will create a separate PR for that

pan-kot (dismissed):

pan-kot (approved):

src/date-range-picker/index.tsx
```diff
@@ -27,6 +27,7 @@ import { warnOnce } from '../internal/logging.js';
 import { usePrevious } from '../internal/hooks/use-previous/index.js';
 import { useUniqueId } from '../internal/hooks/use-unique-id';
 import { formatTimezoneOffset, getBrowserTimezoneOffset, isIsoDateOnly } from '../internal/utils/date-time';
+import { formatValue } from './use-date-range-picker.js';
```
michaeldowseza, line 30: Any reason this function is in the hook file and not in the utils/date-time file like the other formatters on the line above?

michaeldowseza (commented):
Non blocking comment/question.

src/date-range-picker/index.tsx
zshakeel-amazon: This function is somewhat specific to the range picker since it formats the value depending on the range type so I would think it would make sense to keep it within the hook?. The above functions seem to all be quite generic which is why I think they're in the utils folder

pan-kot: I think this can be tackled as a follow-up

PR #117, opened 2022-08-03
Ensure last breadcrumb doesn't have an href so that it can have aria-disabled="true"

jorycunningham:
### Description

Removed href on last breadcrump in a group so that we can use aria-disabled on it while retaining validity of the HTML. 

### How has this been tested?

* Ran axe-core against build
* Validated rendered HTML in W3C Nu Validator
* Inspected rendered DOM

### Documentation changes

- [X] _No._

### Related Links

AWSUI-18939

<details>

#### Correctness

- [X]  _Changes do not include unsupported browser features, see [`CONTRIBUTING.md`](CONTRIBUTING.md#browsers-support)._
- [X]  _Changes were manually tested for accessibility, see [accessibility guidelines](https://cloudscape.design/foundation/core-principles/accessibility/)._

#### Security

#### Testing

</details>

### DIFF
--- a/src/breadcrumb-group/__tests__/breadcrumb-group.test.tsx
+++ b/src/breadcrumb-group/__tests__/breadcrumb-group.test.tsx
@@ -58,7 +58,12 @@ describe('BreadcrumbGroup Component', () => {
 
       links.forEach((link, i) => {
         expect(link.getElement()).toHaveTextContent(items[i].text);
-        expect(link.getElement()).toHaveAttribute('href', items[i].href);
+        if (i === links.length - 1) {
+          // last item should not have an href
+          expect(link.getElement()).not.toHaveAttribute('href', items[i].href);
+        } else {
+          expect(link.getElement()).toHaveAttribute('href', items[i].href);
+        }
       });
     });
 
@@ -136,7 +141,12 @@ describe('BreadcrumbGroup Component', () => {
     });
 
     test('does not throw an error when a safe javascript: URL is passed', () => {
-      const element = renderBreadcrumbGroup({ items: [{ text: '', href: 'javascript:void(0)' }] });
+      const element = renderBreadcrumbGroup({
+        items: [
+          { text: '1', href: 'javascript:void(0)' },
+          { text: '2', href: 'javascript:void(0)' },
+        ],
+      });
       expect((element.findBreadcrumbLink(1)!.getElement() as HTMLAnchorElement).href).toBe('javascript:void(0)');
       expect(console.warn).toHaveBeenCalledTimes(0);
     });
--- a/src/breadcrumb-group/item/item.tsx
+++ b/src/breadcrumb-group/item/item.tsx
@@ -28,11 +28,12 @@ export function BreadcrumbItem<T extends BreadcrumbGroupProps.Item>({
     <div className={clsx(styles.breadcrumb, isLast && styles.last)}>
       <a
         {...focusVisible}
-        href={item.href || '#'}
+        href={isLast ? undefined : item.href || '#'}
         className={clsx(styles.anchor, { [styles.compressed]: isCompressed })}
         aria-current={isLast ? 'page' : undefined} // Active breadcrumb item is implemented according to WAI-ARIA 1.1
         aria-disabled={isLast && 'true'}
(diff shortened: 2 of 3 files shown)


src/breadcrumb-group/item/item.tsx
```diff
@@ -28,11 +28,12 @@ export function BreadcrumbItem<T extends BreadcrumbGroupProps.Item>({
     <div className={clsx(styles.breadcrumb, isLast && styles.last)}>
       <a
         {...focusVisible}
-        href={item.href || '#'}
+        href={isLast ? undefined : item.href || '#'}
         className={clsx(styles.anchor, { [styles.compressed]: isCompressed })}
         aria-current={isLast ? 'page' : undefined} // Active breadcrumb item is implemented according to WAI-ARIA 1.1
         aria-disabled={isLast && 'true'}
         onClick={isLast ? preventDefault : onClickHandler}
+        tabIndex={isLast ? 0 : undefined} // tabIndex is added to the last crumb to keep it in the index without an href
```
michaeldowseza, line 36: Isn't this logically equivalent to this since the default tabIndex for a non-disabled anchor tag would be based on the dom order?

```
tabIndex={0}
```

michaeldowseza (commented):
PR title is a bit misleading and does not represent what the code is technically doing. The code is removing the `href` from the last breadcrumb which just happens to be the disabled.

src/breadcrumb-group/item/item.tsx
jorycunningham: You are asking why not just have tabIndex = 0 on all the links because it would be functionally equivalent? Is that correct?

It certainly would be functionally equivalent. I was favoring not adding unnecessary/redundant attributes and since isLast is already evaluated I don't believe it would have much if any performance impact to make it conditional, but I'm happy to take either approach, especially if one if more idiomatic to Cloudscape

jorycunningham:
> PR title is a bit misleading and does not represent what the code is technically doing. The code is removing the `href` from the last breadcrumb which just happens to be the disabled.

Apologies, I'm not sure I follow this. The reason we are removing the href IS that the last one has aria-disabled, it doesn't just happen to be disabled. The recommended resolution for AWSUI-18939 is to remove href on crumbs with aria-disabled="true" to prevent invalid HTML from the combination of the two attributes. Were aria-disabled not present there would be no need to remove the href.

michaeldowseza:
> Apologies, I'm not sure I follow this. The reason we are removing the href IS that the last one has aria-disabled, it doesn't just happen to be disabled. The recommended resolution for AWSUI-18939 is to remove href on crumbs with aria-disabled="true" to prevent invalid HTML from the combination of the two attributes. Were aria-disabled not present there would be no need to remove the href.

The misleading part is the title currently suggests that any breadcrumb that it will remove the href of any breadcrumb that has aria-disabled=true but that is not the case.

src/breadcrumb-group/item/item.tsx
michaeldowseza: I have no strong preference here.

jorycunningham:
I see, thank you for the explanation @michaeldowseza . I've updated the PR title and unit test logic.

michaeldowseza (approved):

src/breadcrumb-group/__tests__/breadcrumb-group.test.tsx
```diff
@@ -136,7 +141,12 @@ describe('BreadcrumbGroup Component', () => {
     });
 
     test('does not throw an error when a safe javascript: URL is passed', () => {
-      const element = renderBreadcrumbGroup({ items: [{ text: '', href: 'javascript:void(0)' }] });
+      const element = renderBreadcrumbGroup({
+        items: [
+          { text: '1', href: 'javascript:void(0)' },
+          { text: '2', href: 'javascript:void(0)' },
```
YueyingLu, line 147: Why add second item here?

YueyingLu: Hi @jorycunningham why do we change here?

jorycunningham: I did that because the behavior of the last in the series in different in regards to href and being disabled

PR #118, opened 2022-08-03
convert select option item divs into spans to preserve HTML validity

jorycunningham:
### Description

Converts divs that are parts of an option item into spans because they are sometimes the child of a button tag and buttons cannot have div children

### How has this been tested?

* Visual Diff
* Inspected rendered DOM
* Validated with W3C Nu Validator

### Documentation changes

- [X] _No._

### Related Links

AWSUI-18930

<details>

#### Correctness

- [X]_Changes do not include unsupported browser features, see [`CONTRIBUTING.md`](CONTRIBUTING.md#browsers-support)._

#### Security

#### Testing

</details>

### DIFF
--- a/src/internal/components/option/index.tsx
+++ b/src/internal/components/option/index.tsx
@@ -58,7 +58,7 @@ const Option = ({
   );
 
   return (
-    <div
+    <span
       title={option.label || option.value}
       data-value={option.value}
       className={className}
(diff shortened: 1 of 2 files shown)


src/internal/components/option/index.tsx
```diff
@@ -58,33 +58,33 @@ const Option = ({
   );
 
   return (
-    <div
+    <span
```
michaeldowseza, line 61: Does this cause any visual changes since these will now be in inline instead of block elements?

michaeldowseza (commented):
General question

src/internal/components/option/index.tsx
jorycunningham: Good question. All of the elements I changed had `display: flex `already so there was no change to the computed display value. Additionally, visual regression passed with no issues.

avinashbot: The one exception appears to be `.description`, but it's a flex child, so a display doesn't explicitly need to be set.

avinashbot (approved):

PR #134, opened 2022-08-09
Add dropdown focus trap

pan-kot:
### Description

Add dropdown focus trap as a generalised solution over recovery focus trap in autosuggest.

The change is needed to decouple autosuggest trigger and footer.

### How has this been tested?

Added new integration tests.

### Documentation changes

- [x] _No._

### Related Links

<details>

#### Correctness

#### Security

#### Testing

</details>

### DIFF
--- a/src/internal/components/dropdown/index.tsx
+++ b/src/internal/components/dropdown/index.tsx
@@ -13,6 +13,8 @@ import { useVisualRefresh } from '../../hooks/use-visual-mode';
 import { usePortalModeClasses } from '../../hooks/use-portal-mode-classes';
 import { DropdownContextProvider, DropdownContextProviderProps } from './context';
 import { useMobile } from '../../hooks/use-mobile';
+import TabTrap from '../tab-trap/index.js';
+import { getFirstFocusable, getLastFocusable } from '../focus-lock/utils.js';
 
 interface DropdownContainerProps {
   children?: React.ReactNode;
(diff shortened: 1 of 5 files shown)


src/internal/components/dropdown/index.tsx
```diff
@@ -333,28 +336,46 @@ const Dropdown = ({
       <div className={clsx(stretchTriggerHeight && styles['stretch-trigger-height'])} ref={triggerRef}>
         {trigger}
       </div>
+
+      <TabTrap
+        focusNextCallback={() => dropdownRef.current && getFirstFocusable(dropdownRef.current)?.focus()}
+        disabled={!open || !trapFocus}
+      />
```
just-boris, line 343: Why is this trap needed? Why does this code not use the `<FocusLock>` wrapper?

pan-kot: Thanks for asking :)

Ideally I'd love to wrap the dropdown in a focus-lock, like:

```
<FocusLock>
  <Trigger />
  <DropdownContainer>...</DropdownContainer>
</FocusLock>
```

However, this only works when expandToViewport=false. Otherwise, the dropdown container is rendered inside portal and the focus-lock utilities can't find the corresponding focus targets.

If you have some ideas on how to support that from the focus-lock I am happy to try it!

just-boris: This can be solvable by adopting `shards` prop from the original focus-lock: https://github.com/theKashey/react-focus-lock#portals

Fine for me to keep it this way for now

just-boris (approved):

src/internal/components/dropdown/index.tsx
pan-kot: Thanks, I mentioned the use case in the existing BRF ticket for generic focus-lock (18931).

PR #149, opened 2022-08-11
refactor(grid): Replace negative margins with CSS grid gap

avinashbot:
### Description

Proof of concept for replacing the negative margin approach in the grid component with the CSS native `gap` property. Safari 13.x/14.0 doesn't support `gap` on flex, so it's implemented as a "one-dimensional" CSS grid. Push/pull continues to work, but I've also marked it as deprecated in the interface - we mention this in the `gridDefinition` documentation, but don't mark it that way in the interface. This way, IDEs will mark those properties as deprecated inline.

The four biggest benefits to dropping negative margins are:
- no more scrollbars/overflow under the document or scroll containers without padding,
- no more pointer-events hacks because of negative margins extending outside the component,
- more precise JS space calculations (e.g. for ResizeObserver or `getBoundingClientRect`)

Unfortunately, due to column layout's design, the negative margins there will have to remain, at least for now, so I don't blow out the scope of this refactor into a redesign. The borders visually extend outside the borders of the component, and the only way to do that is negative margins. 

I've also made changes to other components. Check my comments in the file diff for explanations.

### How has this been tested?

Nothing is supposed to change, so these are tested through unit tests, but more importantly through visual regression tests.

### Documentation changes

  - Somewhat. Added `@deprecated` to push/pull but this doesn't show up on the website.

### Related Links

<details>

#### Correctness

  - Ha. Goodbye, IE11!

#### Security

#### Testing

</details>


src/grid/styles.scss
```diff
@@ -7,71 +7,67 @@
...
 
 .grid {
-  @include styles.styles-reset;
-  display: flex;
-  flex-wrap: wrap;
-  margin: calc(#{$grid-gutter-width} / -2);
-  pointer-events: none;
-
+  // The grid component mostly behaves like a flex layout, but gap with
+  // display flex isn't supported on Safari 13.x and 14.0.
```
DaemonCahill, line 16: I use one dimensional grids all the time and often even prefer them to flex so I don't consider this to be hack to be resolved in the future. That being said, as a policy we should consider broader use of @supports particularly for Safari scenarios because our support policy of -3 versions is excessive relative to other vendors.

src/attribute-editor/row.tsx
```diff
@@ -64,7 +64,7 @@ export const Row = React.memo(
       typeof slot === 'function' ? slot(item, itemIndex) : slot;
 
     return (
-      <InternalBox className={styles.row} margin={{ bottom: 's' }}>
+      <InternalBox className={styles.row} margin={{ bottom: 'l' }}>
```
avinashbot, line 67: Changed this. Previously, the gap between fields was `22px` (`10px` + `12px`) which isn't exactly a nice spacing value. This is now `24px`, so it's for the better.

.eslintrc
```diff
@@ -53,7 +53,7 @@
     "react-hooks/exhaustive-deps": [
       "error",
       {
-        "additionalHooks": "(useContainerQuery|useContainerBreakpoints|useEffectOnUpdate)"
+        "additionalHooks": "(useContainerQuery|useEffectOnUpdate)"
```
avinashbot, line 56: This is wrong. The array that `useContainerBreakpoints` takes isn't a dependencies array.

pages/popover/placement-test.page.tsx
```diff
@@ -65,6 +65,7 @@ export default function () {
                 <Popover
                   id={`popover-${row}-${col}`}
                   size="medium"
+                  fixedWidth={true}
```
avinashbot, line 68: The negative margins change seems to fix (well, more like "change") the width calculation that the popover makes. Now, it "correctly" detects the minimum width of the column layout and it allows it more positioning options on mobile.

| Before | After |
|---|---|
| Before | After |

src/grid/interfaces.tsx
```diff
@@ -57,11 +57,13 @@ export namespace GridProps {
 
     /**
      * The number of grid elements to pull the column to the left by.
+     * @deprecated The use of this property is not recommended for accessibility reasons.
```
michaeldowseza, line 60: I'd also add what they should use instead. 

"@deprecated The use of this property is not recommended for accessibility reasons. Use ... instead."

src/grid/styles.scss
```diff
@@ -7,71 +7,67 @@
...
-  &.push-0 {
-    left: auto;
+  .offset-#{$i} {
+    margin-left: $colspan-width;
   }
 
-  &.pull-0 {
-    right: auto;
+  .push-#{$i} {
+    left: $colspan-width;
```
connorlanigan, line 67: Could this potentially also be implemented using `grid-column-start`? What is the advantage of using `left`/`right`?

avinashbot: We don't know which grid-column each element is actually going to land in. We just use `span` and let the browser lay out the columns. push/pull is relative (i.e. push by 1 grid element), and there's no way to do that in CSS grid (thankfully).

PR #56, opened 2022-07-19
style: Improve focus indicator

YueyingLu:
### Description

Improve the focus indicator on dropdown and select. Because:

1. The visual indicator is very different from the blue outline and therefore users need to learn two distinct focus indicators
2. The light grey color is very subtle and on a large page it's easy to lose your focus
3. The light grey background is not a keyboard-specific focus, it's actually a hover effect that can also be triggered with keyboard focus
4. Nested dropdowns like button dropdown with expandable groups can have two options with the light grey background, making it ambiguous which option is actually focused
5. Selected options in select/multiselect already use blue as an indicator that they are selected. The visual indicator change when focusing these options is very minimal and can be missed easily.

Todo:
1. change indicator for selected options
2. check dark mode
3. add screenshot tests

### How has this been tested?

[_How did you test to verify your changes?_]

[_How can reviewers test these changes efficiently?_]

[_Check for unexpected visual regressions, see [`CONTRIBUTING.md`](CONTRIBUTING.md#run-visual-regression-tests) for details._]

### Documentation changes

[*Do the changes include any API documentation changes?*]
- [ ] _Yes, this change contains documentation changes._
- [ ] _No._

### Related Links

[*Attach any related links/pull request for this change*]

<details>
   <summary>Review checklist</summary>

_The following items are to be evaluated by the author(s) and the reviewer(s)._

#### Correctness

- [ ] _Changes are backward-compatible if not indicated, see [`CONTRIBUTING.md`](CONTRIBUTING.md#public-apis)._
- [ ] _Changes do not include unsupported browser features, see [`CONTRIBUTING.md`](CONTRIBUTING.md#browsers-support)._
- [ ] _Changes were manually tested for accessibility, see [accessibility guidelines](https://cloudscape.design/foundation/core-principles/accessibility/).

#### Security

- [ ] _If the code handles URLs: all URLs are validated through [the `checkSafeUrl` function](https://github.com/cloudscape-design/components/blob/main/src/internal/utils/check-safe-url.ts).

#### Testing

- [ ] _Changes are covered with new/existing unit tests?_
- [ ] _Changes are covered with new/existing integration tests?_
</details>

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #160, opened 2022-08-15
Improve focus indicator on buttondropdown

YueyingLu:
### Description
Split https://github.com/cloudscape-design/components/pull/56 into buttondropdown and select/multiselect/autosuggest. This is buttondropdown part.

Changes:

1. change to blue border on focused items
2. add more content for screenshot test
3. add unit test 

Why:

to meet a11y color contrast criteria
consistent with other focus indicator in the system
indicate hover focus and select state with more noticeable different style

### How has this been tested?

### Documentation changes

- [ ] _No._

### Related Links

AWSUI-15375

<details>

#### Correctness

#### Security

#### Testing

</details>

### DIFF
--- a/src/button-dropdown/internal.tsx
+++ b/src/button-dropdown/internal.tsx
@@ -51,6 +51,7 @@ const InternalButtonDropdown = React.forwardRef(
       isOpen,
       targetItem,
       isHighlighted,
+      isFocused,
       isExpanded,
       highlightItem,
       onKeyDown,
@@ -77,9 +78,6 @@ const InternalButtonDropdown = React.forwardRef(
     useForwardFocus(ref, dropdownRef);
 
     const clickHandler = () => {
-      if (!usingMouse.current) {
-        return;
-      }
       if (!loading && !disabled) {
         toggleDropdown();
         if (dropdownRef.current) {
@@ -140,6 +138,7 @@ const InternalButtonDropdown = React.forwardRef(
         onKeyDown={onKeyDown}
         onKeyUp={onKeyUp}
         onMouseDown={handleMouseEvent}
+        onMouseMove={handleMouseEvent}
         className={clsx(styles['button-dropdown'], styles[`variant-${variant}`], baseProps.className)}
         aria-owns={expandToViewport && isOpen ? dropdownId : undefined}
         ref={__internalRootRef}
@@ -184,6 +183,7 @@ const InternalButtonDropdown = React.forwardRef(
               hasExpandableGroups={expandableGroups}
               targetItem={targetItem}
               isHighlighted={isHighlighted}
+              isFocused={isFocused}
               isExpanded={isExpanded}
               highlightItem={highlightItem}
               expandToViewport={expandToViewport}
--- a/src/button-dropdown/internal.tsx
+++ b/src/button-dropdown/internal.tsx
@@ -41,7 +41,6 @@ const InternalButtonDropdown = React.forwardRef(
     ref: React.Ref<ButtonDropdownProps.Ref>
   ) => {
     const isInRestrictedView = useMobile();
-    const usingMouse = useRef(true);
     const dropdownId = useUniqueId('dropdown');
     for (const item of items) {
       checkSafeUrl('ButtonDropdown', item.href);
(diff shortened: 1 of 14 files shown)


src/button-dropdown/internal.tsx
```diff
@@ -77,9 +78,6 @@ const InternalButtonDropdown = React.forwardRef(
     useForwardFocus(ref, dropdownRef);
 
     const clickHandler = () => {
-      if (!usingMouse.current) {
```
pan-kot, line 80: why this was removed?

src/button-dropdown/items-list.tsx
```diff
@@ -34,6 +35,7 @@ export default function ItemsList({
           onItemActivate={onItemActivate}
           disabled={item.disabled ?? categoryDisabled}
           highlighted={isHighlighted(item)}
+          focused={isFocused(item)}
```
pan-kot, line 38: Does this work when an item was highlighted with the keyboard and then user highlights it with the mouse? As the highlighted index remains unchanged and the isFocused is calculated from a ref, the style might not change as expected as no re-render would occur.

src/button-dropdown/internal.tsx
YueyingLu: Remove it because it is for preventing firing onClick event when Space/Enter is pressed on IE11. CR-35629483

src/button-dropdown/items-list.tsx
rubencarvalho: This is a good point - it shouldn't work. At least for the first time this happens, afterwards the callback might be memoized. It would be good if we could add a test for this case.

YueyingLu: Good points. Changed the usingMouse as state to update isFocus so that the style would update when changing from keyboard to mouse. Also added a test for this case.

src/button-dropdown/item-element/index.tsx
```diff
@@ -23,6 +23,7 @@ const ItemElement = ({
   first = false,
   last,
   hasCategoryHeader,
+  focused = false,
```
pan-kot, line 26: Why not have the `highlightedType` property same to select/multiselect? The `focused` flag is not very clear in the context of the item-element component. If the item actually receives focus then in can be checked by comparing it against the document.activeElement. If the `focused` is only referring to the difference in visual highlight, I'd at least rename it to e.g. `isKeyboardHighlight`.

src/button-dropdown/utils/use-button-dropdown.ts
```diff
@@ -31,14 +31,23 @@ export function useButtonDropdown({
...
+    targetItem,
+    isHighlighted,
+    isFocused,
+    isExpanded,
+    highlightItem,
+    moveHighlight,
+    expandGroup,
+    collapseGroup,
+    reset,
+    setIsUsingMouse,
```
pan-kot, line 45: If the approach from select/multiselect can be taken, the highlightedType can be then set from inside `highlightItem` so no need in additional callback.

src/button-dropdown/item-element/index.tsx
YueyingLu: I think it is logically focused, because isFocus() checks if the item is last the one currently highlight (because we also highlight the parent item of the highlighted one) and if it is by keyboard (mouse is hover) -- it is what we expect from focus, or not? Also one of purposes of this change is to differentiate focus style and hover style. We can't use document.activeElement because highlighted item is always focus() but from user view mouse highlight is hover https://github.com/cloudscape-design/components/blob/main/src/button-dropdown/item-element/index.tsx#L78

src/button-dropdown/utils/use-button-dropdown.ts
YueyingLu: In buttondropdown the highlightItem() is called when mouseEnter on item-element, unlike in select it depends on the event on list. In order to change style when mousemove, we need change the event handler and pass usingMosue -- otherwise highlightItem is called before usingMouse gets updated as usingMouse is updated on the dropdown event handler. So it makes more sense to me to not make highlightedType dependent on highlightItem in this case.

src/button-dropdown/item-element/index.tsx
pan-kot: So if the item is always focused, having that `focused` property is a little confusing. Would it make sense to at least rename it or document with a comment?

YueyingLu: renamed to isKeyboardHighlight and isKeyboardHighlighted and also added comment to isKeyboardHighlight function.

pan-kot (approved):

pan-kot (approved):

PR #190, opened 2022-08-23
Introduce property-filter-autosuggest derived from internal autosuggest

pan-kot:
### Description

Property filter uses autosuggest internally but requires more features that the public autosuggest API has to offer. The coming changes to the property filter (see https://github.com/cloudscape-design/components/pull/166) would only worsen this problem.

In this PR a new internal property-filter-autosuggest component is introduced that is similar to autosuggest but includes the property filter-related changes. Those extra properties are removed from the internal autosuggest as no longer required.

### How has this been tested?

Existing tests.

### Documentation changes

- [x] _No._

### Related Links

<details>

#### Correctness

#### Security

#### Testing

</details>

### DIFF
--- a/pages/property-filter/property-filter-autosuggest.page.tsx
+++ b/pages/property-filter/property-filter-autosuggest.page.tsx
@@ -1,7 +1,7 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 import React, { useState } from 'react';
-import InternalAutosuggest from '~components/autosuggest/internal';
+import PropertyFilterAutosuggest from '~components/property-filter/property-filter-autosuggest';
 
 const options = [
   {
@@ -15,18 +15,15 @@ const options = [
   { value: 'Option', description: 'description2' },
 ];
 const enteredTextLabel = (value: string) => `Use: ${value}`;
-export default function AutosuggestPage() {
+
+export default function PropertyFilterAutosuggestPage() {
   const [value, setValue] = useState('Option 1');
   return (
     <div style={{ padding: 10 }}>
-      <h1>Internal autosuggest features</h1>
-      <InternalAutosuggest
-        __filterText={value.substr(2)}
-        __onOptionClick={e => {
-          e.preventDefault();
-        }}
-        __dropdownWidth={300}
-        __disableShowAll={true}
+      <h1>Property filter autosuggest features</h1>
+      <PropertyFilterAutosuggest
+        filterText={value.substr(2)}
+        onOptionClick={e => e.preventDefault()}
         value={value}
         options={options}
         onChange={event => setValue(event.detail.value)}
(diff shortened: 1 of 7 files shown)


gethinwebster:
I'm not totally sure about the tradeoff here. It reduces complexity in `autosuggest`, but in turn we get a lot of duplication of a relatively complex component, that we then have to update twice for any future improvements.
To me, it seems like the risk of duplication is larger overall, what is it about the alternative (i.e. the current approach) that you think makes it a worse solution?

pan-kot:
> I'm not totally sure about the tradeoff here. It reduces complexity in `autosuggest`, but in turn we get a lot of duplication of a relatively complex component, that we then have to update twice for any future improvements. To me, it seems like the risk of duplication is larger overall, what is it about the alternative (i.e. the current approach) that you think makes it a worse solution?

I think neither solution is ideal. I'd love to have some abstract autosuggest that takes a string input and offers some value representation be it a list of options or anything and also giving enough flexibility to control what is highlighted etc. If we create such a component, both autosuggest and property-filter can derive from there. However, that would take a decent amount of work (but hopefully we'll get there).

To me, the new property-filter-autosuggest component does not include that much of a duplicated logic - the majority of it is inside the utils that are imported from the autosuggest still. In case we need to update something in there, we'll have a choice to only do it for autosuggest, property-filter-autosuggest, or both components, in which case the logic can be still de-duplicated by introducing it as a util that belongs to the autosuggest scope.

The benefit of having a separate component though, is that we can update it separately and also change the signature. For example, one thing that wasn't done in scope of the current PR is handling of the autosuggest options. Instead of computing the options elsewhere (see [ref](https://github.com/cloudscape-design/components/blob/main/src/property-filter/index.tsx#L70)), this logic can be moved to the new component. The future extension of the property-filter would require at least two new internal properties to the autosuggest that are only needed by the property filter and break the autosuggest's boundaries even more.

Lastly, I think it is easier to transition to the ideal state from having multiple similar components (by reusing the common parts) rather than having the monstrous ones.

gethinwebster:
Ok, this seems reasonable. I think there are two things I'd suggest to minimize the risk:
1. Ensure that we have a ticket somewhere (perhaps BRF? although maybe it's too big for that) to address this technical debt and abstract further.
2. Ensure that people working on a11y are aware of this duplication, as a) a11y-related attributes and functionality seem to be some of the parts that are more duplicated rather than shared and b) this is likely to change in the near future with the a11y audit

gethinwebster (approved):

pan-kot:
> Ok, this seems reasonable. I think there are two things I'd suggest to minimize the risk:
> 
> 1. Ensure that we have a ticket somewhere (perhaps BRF? although maybe it's too big for that) to address this technical debt and abstract further.
> 2. Ensure that people working on a11y are aware of this duplication, as a) a11y-related attributes and functionality seem to be some of the parts that are more duplicated rather than shared and b) this is likely to change in the near future with the a11y audit

Thanks, I will create a follow-up BRF ticket and share the PR with the a11y people 👍

pan-kot:
Follow-up ticket: 19112

PR #201, opened 2022-08-25
chore: Add internal component to transparently proxy focus through portals

avinashbot:
#### The problem statement

There are a number of components that could render content in portals outside of the standard DOM order. Notably, these are popovers (with `renderWithPortal`) and dropdown-based components (with `expandToViewport`). When focus gets moved to these floating components, tabbing past the end causes focus to hit the end of the DOM instead of "resuming" past the trigger. Popovers normally render a focus trap, but the ones used in onboarding don't have a focus trap, so they're affected as well.

#### My proposal

A generic component, that behaves very much like a portal, but also moves focus through the portal'ed content as if it was rendered inside the DOM. It works by detecting tabbable elements before and after the portal and determining focus direction. Try it locally!

Maybe this logic could be integrated into the standard portal component, but I'm not sure yet.

### How has this been tested?

Only locally, and manually. An integration test is definitely in order, but let's validate the approach first.

### Documentation changes

- [x] _No._

### Related Links

<details>

#### Correctness

#### Security

#### Testing

</details>


pan-kot:
"In the future, if the portal is the final tabbable element, it would be nice to be able to disable the final tab trap so that the focus naturally moves into the address bar instead of back to the start of the DOM."

What is the problem with adding this already?

src/internal/components/focus-detector/index.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+export interface FocusDetectorProps {
+  onFocus: React.FocusEventHandler<HTMLDivElement>;
+  disabled?: boolean;
+}
+
+/**
+ * This component handles forwarding focus when navigating through the tab order.
+ *
+ * When the user focuses this component, the `next` callback function is called
```
pan-kot, line 13: A small mistake in the typedoc: the `onFocus` callback function is called

src/internal/components/focus-portal/index.tsx
```diff
@@ -0,0 +1,132 @@
...
+import FocusDetector from '../focus-detector';
+import { getFirstTabbable, getLastTabbable, getTabbables } from '../../utils/tabbables';
+
+export interface FocusPortalProps {
+  container?: Element;
+  children?: React.ReactNode;
+}
+
+export default function FocusPortal({ container, children }: FocusPortalProps) {
+  const proxyBeforeRef = useRef<HTMLDivElement | null>(null);
```
pan-kot, line 15: I'd call these refs `childrenBeforeRef` and `childrenAfterRef` to be better clear

src/internal/components/focus-portal/index.tsx
```diff
@@ -0,0 +1,132 @@
...
+  const proxyAfterRef = useRef<HTMLDivElement | null>(null);
+  const wrapperRef = useRef<HTMLDivElement | null>(null);
```
pan-kot, line 17: childrenWrapperRef ?

src/internal/components/focus-portal/index.tsx
```diff
@@ -0,0 +1,132 @@
...
+  const portalBeforeRef = useRef<HTMLDivElement | null>(null);
+  const portalAfterRef = useRef<HTMLDivElement | null>(null);
+
+  const focusBeforeProxy = () => {
+    const allTabbableElements = getTabbables(document.documentElement);
```
pan-kot, line 22: should you filter out all focus detectors rendered here?

Like `getTabbables(document.documentElement).filter(el => el !== proxyBeforeRef.current && el !== ...)`

pan-kot: That is likely to fix the potential of an infinite loop mentioned below

src/internal/components/focus-portal/index.tsx
```diff
@@ -0,0 +1,132 @@
...
+    return focusStartofDOM();
+  };
+
+  return (
+    <>
+      <FocusDetector
+        ref={proxyBeforeRef}
+        disabled={!children}
+        onFocus={() => {
+          focusInsideProxy() || focusAfterProxy();
```
pan-kot, line 95: if this is focused backwards, should it focus to before proxy?

src/internal/components/focus-portal/index.tsx
```diff
@@ -0,0 +1,132 @@
...
+      />
+      <Portal container={container}>
+        <FocusDetector
+          ref={portalBeforeRef}
+          disabled={!children}
+          onFocus={event => {
+            if (wrapperRef.current?.contains(event.relatedTarget)) {
+              focusBeforeProxy();
+            } else {
+              focusAfterPortal();
```
pan-kot, line 106: Why this is needed? I would imagine this focus to be only triggered with shift-tab from inside the portal.

src/internal/components/focus-portal/index.tsx
```diff
@@ -0,0 +1,132 @@
...
+        />
+        <div ref={wrapperRef}>{children}</div>
+        <FocusDetector
+          disabled={!children}
+          ref={portalAfterRef}
+          onFocus={event => {
+            if (wrapperRef.current?.contains(event.relatedTarget)) {
+              focusAfterProxy();
+            } else {
+              focusBeforePortal();
```
pan-kot, line 118: how this can be triggered?

src/internal/components/focus-portal/index.tsx
```diff
@@ -0,0 +1,132 @@
...
+            }
+          }}
+        />
+      </Portal>
+      <FocusDetector
```
pan-kot, line 123: Is this needed at all?

PR #202, opened 2022-08-25
add errorIconLabel prop to FormField to ensure icon has a text alternative

jorycunningham:
### Description

Our form field errors have an error icon that currently has no text equivalent. This PR adds a prop to provide a text alternative.

* add an errorIconLabel prop to FormField which creates an aria-label wrapper around the error icon
* update Form Field Permutations page to use new prop
* Add tests to ensure the prop renders an aria-label around the icon

### How has this been tested?

Tested manually, ran Axe Core against permutation page, visual regression test

To test, check the Form Fields permutation page with a screen reader and observe that there is now a text equivalent associated with the error icon

### Documentation changes

- [ ] _No._

### Related Links
fixes AWSUI-18945

<details>

#### Correctness

#### Security

#### Testing

</details>


src/test-utils/dom/form-field/index.ts
```diff
@@ -26,6 +26,10 @@ export default class FormFieldWrapper extends ComponentWrapper<HTMLElement> {
     return this.find(`:scope > .${styles.hints} .${styles.error__message}`);
   }
 
+  findErrorIconWrapper(): ElementWrapper | null {
```
timogasda, line 29: I'm not sure adding a test utility to find the error icon is necessary. I don't really see customers using this in their test code.

src/form-field/internal.tsx
```diff
@@ -20,13 +20,16 @@ import { joinStrings } from '../internal/utils/strings';
...
+  errorIconLabel?: string;
 }
 
-export const FormFieldError = ({ id, children }: FormFieldErrorProps) => (
+export const FormFieldError = ({ id, children, errorIconLabel }: FormFieldErrorProps) => (
   <div className={styles.error}>
     <div className={styles['error-icon-shake-wrapper']}>
       <div className={styles['error-icon-scale-wrapper']}>
-        <InternalIcon name="status-warning" size="small" />
+        <span role="img" aria-label={errorIconLabel}>
```
timogasda, line 30: Can we save ourselves one more HTML element and level of nesting and just add the `role` and `aria-label` to the wrapping `div`?

pages/form-field/permutations.page.tsx
```diff
@@ -36,6 +37,7 @@ const permutations = createPermutations<FormFieldProps>([
   },
   {
     label: ['Normal label'],
+    errorIconLabel: ['Error:'],
```
timogasda, line 40: Should it include the colon or not?

timogasda (changes requested):

src/form-field/internal.tsx
jorycunningham: I'm fine with that, I want to double check with @avinashbot since I was following his suggestion, is that agreeable @avinashbot?

src/test-utils/dom/form-field/index.ts
jorycunningham: Removed 👍

src/form-field/internal.tsx
jorycunningham: Good call. @avinashbot 👀  (Just making sure this aligns with the approach you suggested)

pages/form-field/permutations.page.tsx
jorycunningham: You caught me being indecisive! I like the colon if the whole error message was one string, but I don't think it belongs in the label of an icon. Removing it.

timogasda (approved):
Ship it!

timogasda:
To fix unit test errors with the documentation snapshots, you need to run `npm run build` and update the snapshots. Something like: `npx jest --watch -c jest.unit.config.js src/__tests__/documenter.test.ts` and then press `u` to update.

src/form-field/interfaces.ts
```diff
@@ -60,6 +60,11 @@ export interface FormFieldProps extends BaseComponentProps {
    */
   constraintText?: React.ReactNode;
 
+  /**
+   * Provides a text alternative for the error icon in the error message.
+   */
+  errorIconLabel?: string;
```
YueyingLu, line 66: I would recommend to name it `errorIconAriaLabel`, to keep consistent with other ariaLabel prop we have such as `externalIconAriaLabel`

jorycunningham: Are we sure we will always use aria-label? I can see a situation where this, for example, becomes visually hidden text in the future. I'm happy to change it if preferred, but I'm wary of baking in implementation assumptions.

YueyingLu: It is good point. However it is still good to differentiate with visible label by naming. In addition, using aria-label or hidden text make no difference to screenreader users. Even we change it in the future(which I don't yet see why), it is still not bad to keep naming it with aria-label to make the usage of the prop easier to understand. 
Usually when it comes to api change we want more team input to reach consensus. Let's ask @timogasda @avinashbot wdyt?

jorycunningham: A common approach to this in component libraries is to include an "a11y" prefix or suffix like `a11yErrorIconLabel` or `errorIconLabelA11y`.

timogasda: Generally, I would go for consistency. Unfortunately, we currently don't have consistency for this type of property name when we look at other components. For example, we have `dismissAriaLabel` in Alert, but `dismissLabel` in Flashbar. Personally, I prefer `errorIconLabel` because it does abstract from the implementation, which I think makes sense here.

jorycunningham:
I've renamed the prop errorIconAriaLabel and moved it into i18nStrings per @timogasda

src/__tests__/__snapshots__/documenter.test.ts.snap
```diff
@@ -4999,6 +4999,23 @@ This only works well if you're using a single control in the form field.
       "optional": true,
       "type": "string",
     },
+    Object {
+      "description": "An object containing all the necessary localized strings required by the component.",
```
stirache, line 5003: "An object containing all localized strings required by the component."

src/form-field/__tests__/form-field-rendering.test.tsx
```diff
@@ -91,6 +91,20 @@ describe('FormField component', () => {
     expect(labelElement?.getElement().tagName).toBe('LABEL');
   });
 
+  test('errorIcon has an accessible text alternative', () => {
```
stirache, line 94: Are these strings user-facing?

src/form-field/interfaces.ts
```diff
@@ -33,6 +33,11 @@ export interface FormFieldProps extends BaseComponentProps {
    */
   label?: React.ReactNode;
 
+  /**
+   * An object containing all the necessary localized strings required by the component.
```
stirache, line 37: "An object containing all localized strings required by the component."

src/form-field/interfaces.ts
```diff
@@ -67,6 +72,13 @@ export interface FormFieldProps extends BaseComponentProps {
   errorText?: React.ReactNode;
 }
 
+export interface I18nStrings {
+  /**
+   * Provides a text alternative for the error icon in the error message.
```
stirache, line 77: "Provides a text alternative for the error icon in an error message."

stirache (changes requested):

src/form-field/interfaces.ts
jorycunningham: This is copied verbatim from [other instances of the i18n prop](https://polaris.a2z.com/components/date-range-picker/?tabId=api). Are you suggesting a global change here? If so I think we would want to do that here then the rest in another PR

src/form-field/interfaces.ts
jorycunningham: Isn't the error message a specific error message, why would we use an indefinite article here? Also note that the definite "the" is used in similar descriptions in the same interface. What is the suggested style here?

src/form-field/__tests__/form-field-rendering.test.tsx
jorycunningham: These would appear in the console when tests are run during the development process.

src/__tests__/__snapshots__/documenter.test.ts.snap
jorycunningham: This is automatically generated by [this line of code.](https://github.com/cloudscape-design/components/pull/202/files/42aaff509366d7c123fe49fe66245198863f3454#diff-f94eadc24807d031eff51913e08cf8e23843e51183872ac76b0069e4b66561f8R37)

src/form-field/__tests__/form-field-rendering.test.tsx
YueyingLu: No it is not user-facing?

src/form-field/interfaces.ts
```diff
@@ -67,6 +72,13 @@ export interface FormFieldProps extends BaseComponentProps {
   errorText?: React.ReactNode;
 }
 
+export interface I18nStrings {
```
timogasda, line 75: I think this should be moved into a separate `FormFieldProps` namespace to make sure this is 100% compatible with our documenter and website. See for example wizard as a reference: https://github.com/cloudscape-design/components/blob/main/src/wizard/interfaces.ts#L103

timogasda (changes requested):

src/form-field/interfaces.ts
jorycunningham: Just pushed an update with this @timogasda

stirache (approved):

src/form-field/internal.tsx
```diff
@@ -115,7 +117,11 @@ export default function InternalFormField({
 
       {(constraintText || errorText) && (
         <div className={styles.hints}>
-          {errorText && <FormFieldError id={slotIds.error}>{errorText}</FormFieldError>}
+          {errorText && (
+            <FormFieldError id={slotIds.error} errorIconAriaLabel={i18nStrings?.errorIconAriaLabel}>
```
YueyingLu, line 121: Just notice one more thing: the id is attached to the text span. Does it make sense to move the id up to cover icon and the text?

YueyingLu:
@jorycunningham could you also check the test failures?

src/form-field/__tests__/form-field-rendering.test.tsx
```diff
@@ -91,6 +91,20 @@ describe('FormField component', () => {
...
+    const errorText = 'Yikes, that is just plan wrong';
+    const errorIconLabel = 'Error';
+    const wrapper = renderFormField({
+      errorText,
+      errorIconLabel,
```
YueyingLu, line 99: @jorycunningham should be `errorIconAriaLabel`?

YueyingLu:
@jorycunningham it is necessary to add errorIconAriaLabel to fix a11y failures, as you did for Flashbar.

jorycunningham:
> @jorycunningham it is necessary to add errorIconAriaLabel to fix a11y failures, as you did for Flashbar.

@YueyingLu yes. In this case it is an instance of the error icon in the tag editor page, I need to update the i18n interface for tag editor for the new prop, then provide it.

YueyingLu:
@jorycunningham thank you for updating. We still have test failures. My hunch is we moved the id. Could you check it?

timogasda (approved):

YueyingLu (approved):

PR #203, opened 2022-08-26
Fix browser timezone offset helper

pan-kot:
### Description

The new Date().getTimezoneOffset() might return different offset depending on the daylight savings, for example, in timezone Europe/Berlin:

```
new Date('2020-01-01').getTimezoneOffset() === -60
new Date('2020-06-01').getTimezoneOffset() === -120
```

That means that the browser timezone must be always taken from the target date, not the current date.

### How has this been tested?

Added more unit tests.

### Documentation changes

- [x] _No._

### Related Links

<details>

#### Correctness

#### Security

#### Testing

</details>


src/date-range-picker/embedded.tsx
```diff
@@ -31,7 +30,7 @@ export function DateRangePickerEmbedded({
   timeInputFormat = 'hh:mm:ss',
   rangeSelectorMode = 'default',
   onChange,
-  timeOffset = getBrowserTimezoneOffset(),
+  timeOffset,
```
timogasda, line 33: Would this be considered a breaking change?

pan-kot: Why? The default value is still applied but inside the utils.

timogasda (approved):

PR #211, opened 2022-08-26
add statusIconLabel to FlashBar to provide an accessible alternative …

jorycunningham:
### Description

Our form field errors have an error icon that currently has no text equivalent. This PR adds a prop to provide a text alternative.

* add statusIconLabel prop to FlashBar which creates an aria-label wrapper around the status icon
* update Permutations page to use new prop
* Add test to ensure the prop renders an aria-label around the icon

### How has this been tested?

Tested manually, ran Axe Core against permutation page, visual regression test, npm test

To test, check the Form Fields permutation page with a screen reader and observe that there is now a text equivalent associated with the error icon

### Documentation changes

- [ ] _No._

### Related Links
fixes AWSUI-18967

<details>

#### Correctness

#### Security

#### Testing

</details>


timogasda (changes requested):
LGTM, but documentation snapshots need to be updated.

src/flashbar/flash.tsx
```diff
@@ -58,6 +58,7 @@ export const Flash = React.forwardRef(
       content,
       dismissible,
       dismissLabel,
+      statusIconLabel,
```
YueyingLu, line 61: I would recommend to name it `statusIconAriaLabel`, to keep consistent with other ariaLabel prop we have such as `externalIconAriaLabel`

jorycunningham: See [my note here](https://github.com/cloudscape-design/components/pull/202#discussion_r957406356)

src/flashbar/interfaces.ts
```diff
@@ -35,6 +36,7 @@ export interface FlashbarProps extends BaseComponentProps {
    * * `dismissible` (boolean) - Determines whether the component includes a close button icon. By default, the close button is not included.
    * When a user clicks on this button the `onDismiss` handler is called.
    * * `dismissLabel` (string) - Specifies an `aria-label` for to the dismiss icon button for improved accessibility.
+   * * `statusIconLabel` (string) - Specifies an `aria-label` for to the status icon for improved accessibility.
```
stirache, line 39: "Specifies an 'aria-label' for the status icon, to improve accessibility."

stirache (changes requested):

src/flashbar/flash.tsx
just-boris: One more opinion there – we do not provide a label here to annotate the icon. Icon is our internal implementation detail (we may change it without a public API change).

Unlike FormField or Link components, here we annotate not the icon directly, but the `type` property. 

How about calling this label `typeLabel` or `typeAnnotationLabel` to clarify that this property has to be aligned with value of `type`?

YueyingLu: I am not sure if I fully understand the idea. Can we localize the type without public api? What is the pro to annotate type but not the icon?

src/flashbar/interfaces.ts
timogasda: Should we opt for consistency here? The current phrasing matches the text from `dismissLabel` above.

timogasda (changes requested):
Looks like the a11y tests are failing.

src/flashbar/flash.tsx
just-boris: This is only about property naming. I think, calling it “typeAriaLabel” would be semantically more correct

jorycunningham: I've updated this to 'statusIconAriaLabel' per @YueyingLu

YueyingLu:
Hi @jorycunningham . Could you check the a11y tests failure? Seems that some pages using flashbar are missing statusIconLabel.

YueyingLu:
Hi @jorycunningham  I think the a11y test failed because we have Flashbar in many pages in applayout. Could you check them? Also seems that we miss one in flashbar permutation. You can also check the a11y test locally `npm run test:a11y`.

src/flashbar/flash.tsx
```diff
@@ -112,7 +113,9 @@ export const Flash = React.forwardRef(
             : ''
         )}
       >
-        <div className={clsx(styles['flash-icon'], styles['flash-text'])}>{icon}</div>
+        <div className={clsx(styles['flash-icon'], styles['flash-text'])} role="img" aria-label={statusIconAriaLabel}>
```
timogasda, line 116: The previously failing a11y tests actually showcase an important detail about this implementation. Because we're adding `role="img"` all the time now, all existing flash bars will now raise a11y errors because of missing labels. It would now be impossible to mark this icon as decorative.

Are we absolutely sure that we want to implement it this way? Or are there cases where the icon of a flash bar message can be safely considered decorative, especially for the "info" type?

We could render the `role` conditionally, based on whether the ARIA label was provided or not. What do you think?

YueyingLu: In my opinion, it is good to have the a11y error as a reminder to always add the arialabel. But I am also fine to make it only together with arialabel, only if there are cases icons can be considered decorative.

jorycunningham: I don't think we have any situations where the icon is decorative, so I too think we should let the errors occur as an incentive to fix this. We might just want to make sure we encourage users to not to start their messages with "Warning/Error/Info/Success"

timogasda (approved):
Ship it! 🎉

YueyingLu (approved):

PR #212, opened 2022-08-26
change type from search to text for select and autosuggest because they have role combobox

jorycunningham:
### Description

Updates the inputs in select and autosuggest to use `type="text"` rather than `"search"` because role `combobox` (or any role) is not valid on type="search" inputs per [ARIA in HTML](https://www.w3.org/TR/html-aria/)

Also updates a test in `select `that was looking for `type="search"`

### How has this been tested?
Ran npm test and visual regression.

### Documentation changes

- [X] _No._

### Related Links

Fixes AWSUI-18954

<details>

#### Correctness

#### Security

#### Testing

</details>


YueyingLu:
Hey @jorycunningham. With the change the input loses search appearance, which I think is necessary to keep. I wonder if it makes sense to keep type="search" but add `list` attribute so that it has implicit role combobox, check [aria html](https://www.w3.org/TR/html-aria/#input-search-no-list:~:text=a%20list%20attribute-,role%3Dcombobox,-No%20role). What do you think?

YueyingLu:
Hi @jorycunningham . You are right, it is not a good approach. Can we move the role and related aria attribute up to the container div, and keep type=search in the input?

johannes-weber:
@jorycunningham I wanted to check if you had the opportunity to look in the previous question about moving the aria role up to the container to keep `type="search"` on the input?

jorycunningham:
> @jorycunningham I wanted to check if you had the opportunity to look in the previous question about moving the aria role up to the container to keep `type="search"` on the input?

@YueyingLu and I have been discussing this offline. It is my opinion that we cannot take this approach because of the changes to combobox in aria 1.2 and beyond. See the [1.2 spec ](https://w3c.github.io/aria/#combobox)and the[ change notes](https://github.com/w3c/aria/wiki/Resolving-ARIA-1.1-Combobox-Issues) around role placement for combobox.

src/autosuggest/internal.tsx
```diff
@@ -226,7 +226,7 @@ const InternalAutosuggest = React.forwardRef((props: InternalAutosuggestProps, r
         stretchWidth={!__dropdownWidth}
         trigger={
           <InternalInput
-            type="search"
+            type="text"
```
timogasda, line 229: If we really have no choice but to use the `type="text"` attribute, that's fine. However, we will need to adjust `InternalInput` to keep the current search-type visuals and behavior. Otherwise this is a major regression.

timogasda (changes requested):

src/autosuggest/internal.tsx
YueyingLu: @jorycunningham and I have discussed offline, the proposal is to add a new **internal** type to make the input visually search but use type="text".

PR #222, opened 2022-08-30
AWSUI-18936 : removed aria-label from div since its redundant

karmanya79:
### Description

[Issue]
The aria-label on the sorting button is unnecessary. The combination of the aria-sort attribute on the <th> and the internal text of the <div role="button"> are sufficient by themselves.

Remediation : Remove redundant aria-label from div 

### How has this been tested?

npm test

### Documentation changes

- [x] _No._

### Related Links

<details>

#### Correctness

#### Security

#### Testing

</details>


timogasda (changes requested):
I think this first needs a bit of internal discussion about the API here.

I'm fine with no longer relying on ARIA labels to communicate sorting states, but we cannot simply stop implementing parts of our API (in this case, `columnDefinitions.ariaLabel`). I would suggest keeping the implementation, deprecating the property and adjusting our examples according to the new recommendation. We should also update the API documentation of the property.

Let's talk offline :)

YueyingLu:
Hey @karmanya79 Thank you so much for opening the pr! We discussed with @jorycunningham today and we think it makes sense to keep the aria-label as aria-sorted is not fully supported by many screenreaders. So I resolved the ticket and close this PR. Appreciate your great work!

PR #224, opened 2022-08-31
Improve autosuggest and property-filter performance

pan-kot:
### Description

Fix autosuggest/property-filter limitation causing a crash when a having large number of nested options.

Reported error:
```
Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.
```

### How has this been tested?

Added new integration test to cover the improvement.

### Documentation changes

- [x] _No._

### Related Links

* Ticket AWSUI-19166

<details>

#### Correctness

#### Security

#### Testing

</details>


src/autosuggest/options-controller.ts
```diff
@@ -101,7 +101,9 @@ function createItems(options: Options): AutosuggestItem[] {
   const items: AutosuggestItem[] = [];
   for (const option of options) {
     if (isGroup(option)) {
-      items.push(...flattenGroup(option));
```
pan-kot, line 104: That line caused an exception when too many options. See [MDN article](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/apply): 

"But beware: by using apply this way, you run the risk of exceeding the JavaScript engine's argument length limit. The consequences of applying a function with too many arguments (that is, more than tens of thousands of arguments) varies across engines. (The JavaScriptCore engine has hard-coded argument limit of 65536.)"

An exception inside use-memo triggered a React.js exception saying an infinite update loop was detected, which is wrong.

pages/property-filter/performance.page.tsx
```diff
@@ -0,0 +1,56 @@
...
+            countText={`${items.length} matches`}
+            i18nStrings={i18nStrings}
+            expandToViewport={true}
+          />
+        }
+        columnDefinitions={columnDefinitions.slice(0, 2)}
+      />
+    </div>
+  );
+}
```
just-boris, line 56: Is it possible to reproduce the issue using autosuggest only?

src/autosuggest/options-controller.ts
just-boris: 1. This is not very obvious, how about leaving a comment saying why `.push(...arr)` does not work?
2. This does not seem like a root cause of the originally reported error (react rendered too many items). I am guessing we prevented creating copies of objects, but I do not see it anywhere. Could you help me to understand the core of the change here?

pages/property-filter/performance.page.tsx
pan-kot: Yes, should be possible.

src/autosuggest/options-controller.ts
pan-kot: Try this:

```
let items = []
for (let i = 0; i < 200_000; i++) {
  items.push(i)
}
[].push(...items) // Uncaught RangeError: Maximum call stack size exceeded
```

pan-kot: The uncaught exception inside useMemo callback triggers a React exception in turn

just-boris: Good, I did not know that! One more reason to avoid using memos without reasons, because they swallow the real exceptions.

pan-kot: I guess here we have a very clear performance gain from it, but need to be careful indeed.

just-boris (approved):

PR #229, opened 2022-09-01
fix(date-picker): Fix baseDate handling when navigating through the calendar

johannes-weber:
### Description
This change fixes two issues:

1) Scenario value set to a day in March, all dates <= April are disabled:
When opening the date picker, April was shown as month because and not march as the value stated.
With this change, March is now shown instead of April to reflect the current input value.

2) When attempting to use isDateEnabled given a set of available dates, there was an issue when the list of available dates was not continuous by month. E.g. if the following dates `['2022-04-14', '2022-03-14', '2022-01-14']` are enabled, the component was not allowing to go back to previous months: navigating from March 2022 into the past was not possible (making 2022-01-14 inaccessible).

### How has this been tested?

- tested manually.
- add additional unit tests to cover the fix.

### Documentation changes

- [x] _No._

### Related Links
AWSUI-19136

<details>

#### Correctness

#### Security

#### Testing

</details>

### DIFF
--- a/src/date-picker/calendar/__tests__/calendar.test.tsx
+++ b/src/date-picker/calendar/__tests__/calendar.test.tsx
@@ -39,8 +39,12 @@ describe('Date picker calendar', () => {
     };
   }
 
+  const findFocusedDay = (wrapper: DatePickerWrapper) => {
+    return wrapper.findCalendar()!.find(`.${styles['calendar-day-focusable']}`);
+  };
+
   const findFocusableDateText = (wrapper: DatePickerWrapper) => {
-    const focusedItem = wrapper.findCalendar()!.find(`.${styles['calendar-day-focusable']}`);
+    const focusedItem = findFocusedDay(wrapper);
     return focusedItem ? focusedItem.getElement().textContent!.trim() : null;
   };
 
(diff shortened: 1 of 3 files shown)


pan-kot (dismissed):
LGTM!

src/date-picker/calendar/index.tsx
```diff
@@ -76,12 +76,17 @@ const Calendar = ({
     return null;
   };
 
+  // Get the first enabled date of the month. If no day is enabled in the given month, return the first day of the month.
+  // This is needed because of baseDate is used as first focused date. E.g. when navigating to the calendar area.
```
connorlanigan, line 80: [language] I'd suggest:
> This is needed because `baseDate` is used as the first focusable date, for example when navigating to the calendar area.

src/date-picker/calendar/__tests__/calendar.test.tsx
```diff
@@ -392,14 +440,14 @@ describe('Date picker calendar', () => {
...
         const isDateEnabled = (date: Date) => date.getTime() > dateLimit;
-        const { wrapper } = renderDatePicker({ ...defaultProps, isDateEnabled });
+        const { wrapper } = renderDatePicker({ ...defaultProps, value: '2018-03-21', isDateEnabled });
         act(() => wrapper.findOpenCalendarButton().click());
-        act(() => wrapper.setInputValue('2018/03/21'));
-        expect(findCalendarHeaderText(wrapper)).toBe('April 2018');
-        expect(findFocusableDateText(wrapper)).toBe('11');
+        expect(findCalendarHeaderText(wrapper)).toBe('March 2018');
+        expect(wrapper.findCalendar()!.findSelectedDate().getElement().textContent).toBe('21');
+        expect(findFocusableDateText(wrapper)).toBeNull();
```
connorlanigan, line 450: In the `getBaseDate` function, we now return the first day of the month if no other day is selected, and it is my understanding from the comment there that this day would be used as the focus target. Doesn't this conflict with our `toBeNull` expectation here?

connorlanigan (changes requested):

src/date-picker/calendar/__tests__/calendar.test.tsx
pan-kot: In that case no day can be focused.

johannes-weber: What @pan-kot said. In this test case, March does not contain any focusable / enabled day. It marks 2018-03-21, the input's value, as selected but it's not enabled/focusable.

pan-kot (approved):

connorlanigan (approved):

PR #234, opened 2022-09-01
Property filter simplify types

pan-kot:
### Description

A refactoring to property filter that simplifies TS annotations, replaces unnecessary null-checks with default values and moves focus-no-open logic to the internal autosuggest.

### How has this been tested?

Small updates to existing unit tests. Manual checks.

### Documentation changes

- [ ] _No._

Minor changes: default values applied to filteringOptions and customGroupsText.

### Related Links

<details>

#### Correctness

#### Security

#### Testing

</details>


src/property-filter/controller.ts
```diff
@@ -1,21 +1,26 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
-import { PropertyFilterProps } from './interfaces';
-import { fireNonCancelableEvent } from '../internal/events';
+
+import { PropertyFilterProps as I } from './interfaces';
```
johannes-weber, line 4: Suggestion: use the full name instead the shortened one or a more descriptive alias.

I think shortening the interface name to a single character makes reading the code harder.

src/property-filter/token-editor.tsx
```diff
@@ -7,12 +7,11 @@ import { SelectProps } from '../select/interfaces';
 import InternalSelect from '../select/internal';
 import InternalSpaceBetween from '../space-between/internal';
 import InternalAutosuggest from '../autosuggest/internal';
-import { AutosuggestProps } from '../autosuggest/interfaces';
 import InternalPopover, { InternalPopoverRef } from '../popover/internal';
 import { InternalButton } from '../button/internal';
 import { useUniqueId } from '../internal/hooks/use-unique-id/index';
 
-import { PropertyFilterProps } from './interfaces';
+import { PropertyFilterProps as I } from './interfaces';
```
johannes-weber, line 14: Same as above.

johannes-weber (changes requested):

src/property-filter/controller.ts
pan-kot: I would prefer to omit the namespace name completely when using PropertyFilterProps in property filter code. However, I am not aware if that is possible at all. Btw, TS explicitly recommends using modules over namespaces ([see](https://www.typescriptlang.org/docs/handbook/namespaces-and-modules.html)).

Aliasing the namespace name to a single letter is a way to imitate the direct interface imports from a module.

I do agree that having the alias requires the code reader to acknowledge it first, but in the same time it reduces the clutter significantly. Wdyt?

pan-kot: Actually, I found a better way -- updated the PR, please take a look.

johannes-weber (approved):
LGTM.

pan-kot:
Synced with main.

johannes-weber (approved):

PR #240, opened 2022-09-02
Autosuggest input

pan-kot:
### Description

Introduced internal autosuggest input component which is an abstract autosuggest completely decoupled from the options list. This is needed to reuse code between autosuggest/internal and property-filter/property-filter-autosuggest.

### How has this been tested?

*Tests to be added*

### Documentation changes

- [x] _No._

### Related Links

<details>

#### Correctness

#### Security

#### Testing

</details>


src/autosuggest/__tests__/autosuggest-input.test.tsx
```diff
@@ -0,0 +1,264 @@
...
+
+  test('calls onDelayedInput after small interval following the input value change', () => {
+    jest.useFakeTimers();
+    const onDelayedInput = jest.fn();
+    const { wrapper } = render(
+      <AutosuggestInput value="1" onChange={() => undefined} onDelayedInput={onDelayedInput} />
+    );
+    act(() => {
+      Simulate.change(findInput(wrapper).getElement(), { target: { value: '2' } as unknown as EventTarget });
+    });
```
just-boris, line 259: Any reason why you are writing it by hand instead of our util, which does the same: https://github.com/cloudscape-design/components/blob/75a6fdb641d49818c94fd156788756c346694b4e/src/test-utils/dom/input/base-input.ts#L35-L37

If you could not use it, because component is private, you can instantiate an instance of `BaseInputWrapper` directly

src/autosuggest/__tests__/autosuggest-input.test.tsx
```diff
@@ -0,0 +1,264 @@
...
+// SPDX-License-Identifier: Apache-2.0
+import * as React from 'react';
+import { act, Simulate } from 'react-dom/test-utils';
+import { render as renderJsx } from '@testing-library/react';
+import dropdownStyles from '../../../lib/components/internal/components/dropdown/styles.selectors.js';
+import { getFocusables } from '../../../lib/components/internal/components/focus-lock/utils';
+import createWrapper, { ElementWrapper } from '../../../lib/components/test-utils/dom';
+import AutosuggestInput, { AutosuggestInputRef } from '../../../lib/components/autosuggest/autosuggest-input';
+import { KeyCode } from '@cloudscape-design/test-utils-core/utils';
+import '../../__a11y__/to-validate-a11y';
```
just-boris, line 11: Looks like this import does nothing in this file

src/autosuggest/autosuggest-input.tsx
```diff
@@ -0,0 +1,265 @@
...
+  getBlurEventRelatedTarget,
+  NonCancelableEventHandler,
+} from '../internal/events';
+import InternalInput from '../input/internal';
+import { BaseChangeDetail, BaseInputProps, InputKeyEvents, InputProps } from '../input/interfaces';
+import { getFocusables } from '../internal/components/focus-lock/utils';
+import { ExpandToViewport } from '../internal/components/dropdown/interfaces';
+import { InternalBaseComponentProps } from '../internal/hooks/use-base-component';
+import { KeyCode } from '../internal/keycode';
+import styles from './styles.css.js';
```
just-boris, line 24: Not super critical suggestion, but if you managed to extract this component, we can also split the styles file:

```
src/autosuggest/
├── autosuggest-input
│   ├── styles.scss
│   └── index.tsx
├── index.tsx
├── internal.tsx
└── styles.scss
```

src/autosuggest/autosuggest-input.tsx
```diff
@@ -0,0 +1,265 @@
...
+  dropdownWidth?: number;
+  onCloseDropdown?: NonCancelableEventHandler<null>;
+  onDelayedInput?: NonCancelableEventHandler<BaseChangeDetail>;
+  onPressArrowDown?: () => void;
+  onPressArrowUp?: () => void;
+  onPressEnter?: () => boolean;
+}
+
+export interface AutosuggestInputRef extends InputProps.Ref {
+  focusNoOpen(): void;
```
just-boris, line 47: Does not have to be in scope of this PR, but we might expose this feature to public API, because it is useful for our [validation pattern](https://cloudscape.design/patterns/general/validation/). When autosuggest dropdown opens on focus, it covers error text below the field.

If this happens, we probably want to make it an argument instead of a separate function:

```js
autosuggestRef.current.focus(); // opens dropdown
autosuggestRef.current.focus({ preventDropdown: true }); // dropdown stays closed
```

src/autosuggest/autosuggest-input.tsx
```diff
@@ -0,0 +1,265 @@
...
+      'aria-label': ariaLabel,
+      'aria-activedescendant': ariaActivedescendant,
+    };
+
+    const [trapDropdownFocus, setTrapDropdownFocus] = useState(false);
+
+    // Run this effect on every render to determine if necessary to trap focus around input and dropdown.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    useEffect(() => {
+      setTrapDropdownFocus(
```
just-boris, line 212: So, the new React docs explicitly flag this as an anti-pattern: https://beta.reactjs.org/learn/you-might-not-need-an-effect#updating-state-based-on-props-or-state

Can we statically compute this value? Perhaps, introducing this as a prop and compute on the parent component level

src/autosuggest/styles.scss
```diff
@@ -21,6 +21,10 @@
   flex-shrink: 0;
 }
 
+.dropdown-content {
+  display: contents;
```
just-boris, line 25: Looks like this property is not safari-friendly: https://caniuse.com/css-display-contents

Could you check if anything explodes there?

just-boris (commented):
All good, except two important comments
1. Use of `useEffect` to calculate focusable elements
2. Use of `display:contents` and browser support

src/autosuggest/styles.scss
pan-kot: Yes, I checked it in Safari already with VO and found no regression.

src/autosuggest/autosuggest-input.tsx
pan-kot: I tried to compute it w/o the state/effect but that does not trigger the dropdown to update. I will check again to see what can be done. It is possible to make it a prop but I'd like to avoid that.

src/autosuggest/autosuggest-input.tsx
pan-kot: I like the idea and can do it as a follow-up and with having an api-review doc.

src/autosuggest/autosuggest-input.tsx
pan-kot: Wouldn't it be better to move it to internal/components/autosuggest-input? In either case, I'd rather do it as a follow-up as otherwise it will produce a large diff hiding the actual changes I made here.

src/autosuggest/__tests__/autosuggest-input.test.tsx
pan-kot: Thanks, I will remove it from here. We check it on the autosuggest component which should be enough.

src/autosuggest/__tests__/autosuggest-input.test.tsx
pan-kot: Ah, I haven't thought about that -- thanks, will update 👍

src/autosuggest/styles.scss
just-boris: What versions did you check? We support down to Safari v13
https://github.com/cloudscape-design/components/blob/d4e70a6f3fa929b8c620f7f4b3ebd805277c2c3c/package.json#L156

We can of course merge as is and make our screenshot tests fail, but we can also validate this ahead of time

src/autosuggest/autosuggest-input.tsx
pan-kot: I am not sure the existing implementation is considered an anti-pattern according to the linked article. The reason of the useEffect is that we check the refs which only makes sense when the corresponding elements are rendered. If the computation is run inside the component's body, the refs would be null.

src/autosuggest/styles.scss
pan-kot: Just checked Safari 12.1.

src/autosuggest/autosuggest-input.tsx
just-boris: If we have to disable the eslint plugin, this is a definite incompatibility with React idiomatic patterns.

If you believe this is not a risk, I will have to trust you on that.

just-boris (approved):

src/autosuggest/autosuggest-input.tsx
pan-kot: Thanks 👍

PR #267, opened 2022-09-13
Allow autosuggest input content to be focusable

pan-kot:
### Description

Before this change clicking inside autosuggest input content caused the dropdown to close.

Required for: https://github.com/cloudscape-design/components/pull/166

### How has this been tested?

Added new unit- and integration tests.

### Documentation changes

- [x] _No._

### Related Links

<details>

#### Correctness

#### Security

#### Testing

</details>

### DIFF
--- a/src/internal/components/autosuggest-input/index.tsx
+++ b/src/internal/components/autosuggest-input/index.tsx
@@ -214,11 +214,7 @@ const AutosuggestInput = React.forwardRef(
       name,
       placeholder,
       autoFocus,
-      onClick: (event: React.MouseEvent) => {
-        // Prevents this event to be captured by the dropdown issuing close.
-        event.stopPropagation();
-        openDropdown();
-      },
+      onClick: openDropdown,
       role: 'combobox',
       'aria-autocomplete': 'list',
       'aria-expanded': expanded,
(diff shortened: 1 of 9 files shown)


src/internal/components/autosuggest-input/index.tsx
```diff
@@ -131,8 +132,10 @@ const AutosuggestInput = React.forwardRef(
       ) {
         return;
       }
-      closeDropdown();
-      fireNonCancelableEvent(onBlur, null);
+      if (!preventCloseOnBlurRef.current) {
```
gethinwebster, line 135: Some of this focus handling (not 100% sure it's this bit or elsewhere) seems to create a tab trap: if opening the `autosuggest-input` page and then tabbing through, it gets stuck cycling between the input and the button in the dropdown

src/internal/components/dropdown/index.tsx
```diff
@@ -283,12 +283,12 @@ const Dropdown = ({
      * finish bubbling, so that it is not immediately captured here.
      */
     const timeout = setTimeout(() => {
-      window.addEventListener('click', clickListener);
+      window.addEventListener('mousedown', clickListener);
```
gethinwebster, line 286: This means that when clicking on the input when the dropdown is open, the dropdown temporarily disappears and then reappears. I'm not sure if this is problematic as such, but it's definitely a change from the current behavior and seems potentially detrimental. Have you discussed with design?

gethinwebster (changes requested):

src/internal/components/autosuggest-input/index.tsx
pan-kot: The tab trap is intended. It is controlled by the `trapFocus` property of the dropdown.

src/internal/components/dropdown/index.tsx
pan-kot: Nice catch! I will check if it is possible to avoid having that effect.

src/internal/components/autosuggest-input/index.tsx
gethinwebster: OK yes, I see now that it didn't actually change. Still not quite sure I understand _why_ we'd ever want a tab trap here, but I guess that's outside the scope of this PR.

gethinwebster (approved):

PR #288, opened 2022-09-19
Align Checkbox Toggle implementation with description in Formfield co…

YueyingLu:
…ntext

### Description
To fix the misalignment between api description and implementatin:

**ariaDescribedby** and **ariaLabelledby**
> Adds aria-describedby to the component. If you're using this component within a form field, don't set this property because the form field component automatically sets it. 

> Adds aria-labelledby to the component. If you're using this component within a form field, don't set this property because the form field component automatically sets it. 

Toggle and Checkbox doesn't set description and label provided by parent Formfield(without this fix). They are necessary as the label and description of Formfield provides meaningful content to describe the usage of Toggle and Checkbox. For example, Formfield label and description act as group label or description when it wraps a group of checkbox.

**controlId**

> Specifies the ID of the native form element. You can use it to relate a label element's for attribute to this control. It defaults to an automatically generated ID that is provided by its parent form field component.

Toggle and Checkbox do not take controlId from parent Formfield in the implementation(with or without this fix). It makes sense because they usually have its own label, and no need to have Formfield label to control the form element in addition. Propose to change API description:

> Specifies the ID of the native form element. It uses an automatically generated ID by default.

Decision made in doc Checkbox-Toggle-and-Formfield-controlId-association

### How has this been tested?

### Documentation changes

- [ ] _No._

### Related Links

<details>

#### Correctness

#### Security

#### Testing

</details>

### DIFF
--- a/src/checkbox/__tests__/checkbox.test.tsx
+++ b/src/checkbox/__tests__/checkbox.test.tsx
@@ -7,7 +7,6 @@ import FormField from '../../../lib/components/form-field';
 import Checkbox, { CheckboxProps } from '../../../lib/components/checkbox';
 import styles from '../../../lib/components/internal/components/checkbox-icon/styles.selectors.js';
 import abstractSwitchStyles from '../../../lib/components/internal/components//abstract-switch/styles.css.js';
-import { joinStrings } from '../../../lib/components/internal/utils/strings/join-strings';
 import { createCommonTests } from './common-tests';
 
 function renderCheckbox(jsx: React.ReactElement) {
@@ -169,21 +168,16 @@ test('Should set aria-describedby and aria-labelledby from Formfield', () => {
   );
   const formFieldWrapper = createWrapper(container).findFormField();
   const checkboxWrapper = createWrapper(container).findCheckbox()!;
-  const checkboxAriaDescribedby = checkboxWrapper.findNativeInput().getElement().getAttribute('aria-describedby');
-  const checkboxAriaLabelledby = checkboxWrapper.findNativeInput().getElement().getAttribute('aria-labelledby');
+  const checkboxInputAriaDescribedby = checkboxWrapper.findNativeInput().getElement().getAttribute('aria-describedby');
+  const checkboxInputAriaLabelledby = checkboxWrapper.findNativeInput().getElement().getAttribute('aria-labelledby');
 
-  expect(checkboxAriaDescribedby).toBe(
-    joinStrings(
-      formFieldWrapper?.findDescription()?.getElement().id,
-      container?.querySelector(`.${abstractSwitchStyles.description}`)?.id
-    )
-  );
-  expect(checkboxAriaLabelledby).toBe(
-    joinStrings(
-      container?.querySelector(`.${abstractSwitchStyles.label}`)?.id,
-      formFieldWrapper?.findLabel()?.getElement().id
-    )
-  );
+  const formFieldLabelId = formFieldWrapper?.findLabel()?.getElement().id;
+  const formFieldDescriptionId = formFieldWrapper?.findDescription()?.getElement().id;
+  const checkboxLabelId = container?.querySelector(`.${abstractSwitchStyles.label}`)?.id;
+  const checkboxDescriptionId = container?.querySelector(`.${abstractSwitchStyles.description}`)?.id;
+
+  expect(checkboxInputAriaDescribedby).toBe(formFieldDescriptionId + ' ' + checkboxDescriptionId);
+  expect(checkboxInputAriaLabelledby).toBe(checkboxLabelId + ' ' + formFieldLabelId);
 });
 
 test('Should set aria-describedby and aria-labelledby from ariaLabelledby and ariaDescribedby', () => {
@@ -202,15 +196,14 @@ test('Should set aria-describedby and aria-labelledby from ariaLabelledby and ar
     </FormField>
   );
   const checkboxWrapper = createWrapper(container).findCheckbox()!;
-  const checkboxAriaDescribedby = checkboxWrapper.findNativeInput().getElement().getAttribute('aria-describedby');
-  const checkboxAriaLabelledby = checkboxWrapper.findNativeInput().getElement().getAttribute('aria-labelledby');
+  const checkboxInputAriaDescribedby = checkboxWrapper.findNativeInput().getElement().getAttribute('aria-describedby');
+  const checkboxInputAriaLabelledby = checkboxWrapper.findNativeInput().getElement().getAttribute('aria-labelledby');
 
-  expect(checkboxAriaDescribedby).toBe(
-    joinStrings('description-id', container?.querySelector(`.${abstractSwitchStyles.description}`)?.id)
-  );
-  expect(checkboxAriaLabelledby).toBe(
-    joinStrings(container?.querySelector(`.${abstractSwitchStyles.label}`)?.id, 'label-id')
-  );
+  const toggleLabelId = container?.querySelector(`.${abstractSwitchStyles.label}`)?.id;
+  const toggleDescriptionId = container?.querySelector(`.${abstractSwitchStyles.description}`)?.id;
+
+  expect(checkboxInputAriaDescribedby).toBe('description-id' + ' ' + toggleDescriptionId);
+  expect(checkboxInputAriaLabelledby).toBe(toggleLabelId + ' ' + 'label-id');
 });
 
 test('clicking on formfield label area wont call onchange', () => {
--- a/src/checkbox/__tests__/checkbox.test.tsx
+++ b/src/checkbox/__tests__/checkbox.test.tsx
@@ -205,17 +205,3 @@ test('Should set aria-describedby and aria-labelledby from ariaLabelledby and ar
   expect(checkboxInputAriaDescribedby).toBe('description-id' + ' ' + toggleDescriptionId);
   expect(checkboxInputAriaLabelledby).toBe(toggleLabelId + ' ' + 'label-id');
 });
-
-test('clicking on formfield label area wont call onchange', () => {
-  const onChange = jest.fn();
-  const { container } = render(
-    <FormField description="This is a formfield description." label="Form field label">
-      <Checkbox checked={false} description="This is description" onChange={onChange}>
-        Checkbox label
-      </Checkbox>
-    </FormField>
-  );
-  const formFieldWrapper = createWrapper(container).findFormField();
-  formFieldWrapper!.findLabel()!.click();
-  expect(onChange).not.toHaveBeenCalled();
-});
--- a/src/checkbox/__tests__/checkbox.test.tsx
+++ b/src/checkbox/__tests__/checkbox.test.tsx
@@ -6,7 +6,7 @@ import createWrapper, { CheckboxWrapper } from '../../../lib/components/test-uti
 import FormField from '../../../lib/components/form-field';
 import Checkbox, { CheckboxProps } from '../../../lib/components/checkbox';
 import styles from '../../../lib/components/internal/components/checkbox-icon/styles.selectors.js';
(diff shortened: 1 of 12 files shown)


src/checkbox/base-checkbox.tsx
```diff
@@ -25,11 +25,7 @@ export interface BaseCheckboxProps extends BaseComponentProps, FormFieldControlP
   disabled?: boolean;
 
   /**
-   * Specifies the ID of the native form element. You can use it to relate
-   * a label element's `for` attribute to this control.
-   *
-   * It defaults to an automatically generated ID that
-   * is provided by its parent form field component.
+   * Specifies the ID of the native form element. It uses an automatically generated ID by default.
```
pan-kot, line 28: Why the "You can use it to relate a label element's `for` attribute to this control" part was removed?

src/checkbox/internal.tsx
```diff
@@ -39,6 +38,7 @@ const InternalCheckbox = React.forwardRef<CheckboxProps.Ref, InternalProps>(
     },
     ref
   ) => {
+    const { ariaDescribedby, ariaLabelledby } = useFormFieldContext(rest);
```
pan-kot, line 41: Why using the context here and not inside AbstractSwitch?

If you do it inside the abstract switch it would also be possible to use the form-field controlId:

```
const formFieldContext = useFormFieldContext(rest);
const selfControlId = useUniqueId();
const controlId = propsControlId || formFieldContext.controlId || selfControlId;
```

pan-kot (changes requested):

src/checkbox/base-checkbox.tsx
YueyingLu: I think we don't need highlight it(it feels like recommendation), because unlike Input, Checkbox and Toggle usually have their own labels to control the form.

src/checkbox/internal.tsx
YueyingLu: Inside AbstractSwitch may not work for RadioGroup -- we want to add ariaDescribedby and ariaLabelledby to RadioGroup but AbstractSwitch is for individual RadioButton.

src/checkbox/base-checkbox.tsx
pan-kot: Sure, but as the controlId property exists I think it makes sense to describe its purpose fully. I'd imagine it is possible to provide an external label e.g. if you're building some selectable card or something alike.

If we want to discourage that use of the controlId for these elements I think it is better to deprecate this and provide an explanation.

src/checkbox/internal.tsx
pan-kot: Can we then make the `controlId` property of the abstract switch required and compute this outside to still have the precedence `propsControlId || formFieldContext.controlId || selfControlId `? (for the radio group it is going to be selfControlId always).

src/checkbox/base-checkbox.tsx
gethinwebster: can you include @stirache in this too, as it impacts public-facing documentation

src/checkbox/internal.tsx
YueyingLu: It is the same for controlId(miss it in the last reply), AbstractSwitch controlId is for the individual RadioButton `controlId={item.controlId}`  not for the RadioGroup.

pan-kot: Yes, so for the checkbox we can do:

```
function Checkbox(props) {
  const context = useFormFieldContext(props);
  const selfControlId = useUniqueId();
  const controlId = props.controlId || context.controlId || selfControlId;
  return <AbstractSwitch
    controlId={controlId}
    ariaLabelledby={context.ariaLabelledby}
    ariaDescribedby={context.ariaDescribedby}
    ...
  />
}
```

and for radio group:

```
function RadioGroup(props) {
  const context = useFormFieldContext(props);
  const selfControlId = useUniqueId();
  const controlId = props.controlId || context.controlId || selfControlId;
  return <div role="radio-group"
    ariaLabelledby={context.ariaLabelledby}
    ariaDescribedby={context.ariaDescribedby}
    ...
  >
    {items.map((...) => <RadioButton {...} />)}
  </div>;
}

function RadioButton(props) {
  const controlId = useUniqueId();
  return <AbstractSwitch controlId={controlId} ... />;
}
```

src/checkbox/base-checkbox.tsx
YueyingLu: The controlId can be used to access the input, not only to relate a label. So I don't discourage using it. But I am not keen to state this usage specifically, because it doesn't feel like a very common use case as they have their own label already. To build selectable cards, users should use card instead.

YueyingLu: @gethinwebster I will include Rachel when we have consensus what to include in the description.

pan-kot: I think the controlId is specifically meant for control/label association. Why would you need to assign an id on the native input otherwise? For testing we recommend using our test-utils that can be combined with date-testid.

YueyingLu: I dont have the data what else users may need the controlId. But for example we have controlId in Select, which is not meant (as I understand) for control/label association.

pan-kot: I think in select it works for the input/label association as well: https://github.com/cloudscape-design/components/blob/main/src/select/internal.tsx#L91.

```
// This line takes controlId from restProps or the form-field otherwise.
const formFieldContext = useFormFieldContext(restProps);
const selfControlId = useUniqueId('trigger');
const controlId = formFieldContext.controlId ?? selfControlId;
```

src/toggle/__tests__/toggle.test.tsx
```diff
@@ -100,3 +103,55 @@ test('check a11y', async () => {
...
+  );
+  const toggleWrapper = createWrapper(container).findToggle()!;
+  const toggleAriaDescribedby = toggleWrapper.findNativeInput().getElement().getAttribute('aria-describedby');
+  const toggleAriaLabelledby = toggleWrapper.findNativeInput().getElement().getAttribute('aria-labelledby');
+
+  expect(toggleAriaDescribedby).toBe(
+    joinStrings('description-id', container?.querySelector(`.${abstractSwitchStyles.description}`)?.id)
+  );
+  expect(toggleAriaLabelledby).toBe(
+    joinStrings(container?.querySelector(`.${abstractSwitchStyles.label}`)?.id, 'label-id')
```
just-boris, line 155: The implementation uses `joinStrings` already, so the test shouldn't, or it becomes `expect(add(2, 2)).toEqual(add(2, 2))` which is pointless.

Is there any reason not to put the full computed string here?

YueyingLu: The point of the test is to check if toggle input has aria-describedby identified with formfield description id **AND** its own description id. So I think it is testing `expect(add(attr1, attr2)).toEqual(add(attr3, atrr4))`. joinString is not the point.

src/checkbox/base-checkbox.tsx
YueyingLu: As I understand, the select takes the id from formfield, but not meant for input/label association? Also it is not stated in the description https://github.com/cloudscape-design/components/blob/main/src/select/interfaces.ts#L91

src/toggle/__tests__/toggle.test.tsx
just-boris: Having a static string in the right hand side is still preferred. If you cannot use a static string because there is a generated value, you can use `toMatch` ([example](https://github.com/cloudscape-design/components/blob/7433543fcd2796d6398b67867d25b637c1a2b6de/src/tabs/__tests__/tabs.test.tsx#L137))

```js
expect(toggleAriaLabelledby).toMatch(/^awsui-toggle-\d+ label-id/)
```

src/checkbox/base-checkbox.tsx
pan-kot: In that case I would rather challenge the select.controlId description to include that info.

src/checkbox/internal.tsx
pan-kot: Offline discussion: the control-id is not passed from form-field to the abstract-switch to avoid activating checkbox/radio/toggle when clicking on the form-field label. The current implementation addresses that.

src/checkbox/__tests__/checkbox.test.tsx
```diff
@@ -155,3 +157,65 @@ test('does not trigger any change events when value is changed through api', ()
...
+  const checkboxInputAriaLabelledby = checkboxWrapper.findNativeInput().getElement().getAttribute('aria-labelledby');
+
+  const toggleLabelId = container?.querySelector(`.${abstractSwitchStyles.label}`)?.id;
+  const toggleDescriptionId = container?.querySelector(`.${abstractSwitchStyles.description}`)?.id;
+
+  expect(checkboxInputAriaDescribedby).toBe('description-id' + ' ' + toggleDescriptionId);
+  expect(checkboxInputAriaLabelledby).toBe(toggleLabelId + ' ' + 'label-id');
+});
+
+test('clicking on formfield label area wont call onchange', () => {
```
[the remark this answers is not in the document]
YueyingLu: When adding the controlId from formfield to checkbox, this test failed. So it does do the job? And seems jsdom did fix for checkbox https://github.com/jsdom/jsdom/issues/2367. 
But as you said when testing onFocus in input, it does not work 🤔 .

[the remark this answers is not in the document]
just-boris: The challenge here is that passing test may have two reasons 1) checkbox has explicitly deactivated this 2) it does not work for any components (which is unexpected but all tests are green)

This is why we need to have at least 2 tests in place, one that demonstrates active behavior and another one to demonstrate inactive behavior (maybe more tests to cover other corner-case type of controls, like Select or RadioGroup)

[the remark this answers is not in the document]
just-boris: If you find a way to confirm it in JSDOM, this is fine for me too, but as far as I am aware, it cannot ensure this feature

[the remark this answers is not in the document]
YueyingLu: Added integration test for Input Checkbox and Toggle.

src/checkbox/base-checkbox.tsx
YueyingLu, line 28: @stirache please review this description change, thank you!

pan-kot (dismissed):

src/checkbox/__tests__/checkbox.test.tsx
```diff
@@ -3,8 +3,10 @@
 import React, { useState } from 'react';
 import { render } from '@testing-library/react';
 import createWrapper, { CheckboxWrapper } from '../../../lib/components/test-utils/dom';
+import FormField from '../../../lib/components/form-field';
 import Checkbox, { CheckboxProps } from '../../../lib/components/checkbox';
 import styles from '../../../lib/components/internal/components/checkbox-icon/styles.selectors.js';
+import abstractSwitchStyles from '../../../lib/components/internal/components//abstract-switch/styles.css.js';
```
just-boris, line 9: Double slash in this import. Even if this works, better to keep it clean

```suggestion
import abstractSwitchStyles from '../../../lib/components/internal/components/abstract-switch/styles.css.js';
```

just-boris (commented):
All good, except that comment about slashes

src/checkbox/base-checkbox.tsx
stirache: "Specifies the ID of the native form element. It uses an automatically generated ID by default." --> "Specifies the ID of the native form element. By default, it uses an automatically generated ID."

stirache (changes requested):

pan-kot (approved):

just-boris (approved):

PR #290, opened 2022-09-19
Fix autosuggest and property filter dropdown expanded state

pan-kot:
### Description

When no options found but the empty state or "use" option is present - show the dropdown but set the aria-expanded as "false". In property filter when disableFreeTextFiltering=true it is possible for the dropdown to be hidden completely.

### How has this been tested?

Manual tests.

### Documentation changes

- [x] _No._

### Related Links

<details>

#### Correctness

#### Security

#### Testing

</details>

### DIFF
--- a/src/internal/components/autosuggest-input/index.tsx
+++ b/src/internal/components/autosuggest-input/index.tsx
@@ -302,11 +302,10 @@ const AutosuggestInput = React.forwardRef(
             )
           }
           expandToViewport={expandToViewport}
-          hasContent={!!dropdownContent}
           trapFocus={trapDropdownFocus}
         >
           <div ref={dropdownContentRef} className={styles['dropdown-content']}>
-            {open && dropdownContent}
+            {open ? dropdownContent : null}
           </div>
         </Dropdown>
       </div>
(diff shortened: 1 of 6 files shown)


src/internal/components/autosuggest-input/index.tsx
```diff
@@ -209,7 +209,7 @@ const AutosuggestInput = React.forwardRef(
       }
     };
 
-    const expanded = open && (dropdownExpanded ?? !!dropdownContent);
+    const expanded = open && dropdownExpanded;
```
pan-kot, line 212: aria-expanded need to be set to false if there are no real options present (but there might be a special "Use" option still).

abdhalees (approved):
Approved as the code LGTM but what is the reason behind this change? what does this fix

pan-kot:
> Approved as the code LGTM but what is the reason behind this change? what does this fix

Without this there is a bug with autosuggest. When there are no actual options but only the `Use: "entered text"` option the dropdown was hidden with styles yet only the aria-expanded=false was expected.

Screenshot 2022-09-20 at 11 05 58

PR #294, opened 2022-09-21
Flashbar stacked items feature beta test

DaemonCahill:
This PR adds the ability for teams to beta test the notification stacking feature by enabling the option for users to toggle the behavior on and off. There are no changes to the interface in order to prevent widespread awareness and adoption before the feature is truly ready for production after usability testing.


src/flashbar/index.tsx
```diff
@@ -23,62 +25,142 @@ export default function Flashbar({ items, ...restProps }: FlashbarProps) {
...
+    }
+
     return (
-      <>
-        <TransitionGroup component={null}>
-          {items &&
-            items.map((item, index) => (
-              <Transition
-                transitionChangeDelay={{ entering: TIMEOUT_FOR_ENTERING_ANIMATION }}
+      <div className={clsx(styles['stacking-option'])}>
```
abdhalees, line 56: I don't see the toggle when opening `http://localhost:8080/#/light/flashbar/stacking/` , how do I activate it?

DaemonCahill: @abdhalees I just re-requested review. There have been significant changes from the first revision. For testing you can use any demo page that has notifications and turn on the secret API by adding `{...{ enableStackingOption: true }}` to the Flashbar component. Some good pages to test with are in the `AppLayout` and `Flashbar` sections.

src/flashbar/index.tsx
```diff
@@ -1,85 +1,177 @@
...
-        </TransitionGroup>
-      </>
+          </div>
+        )}
+
+        {isFlashbarStackExpanded && (
+          <div className={styles.expanded}>{items.map((item, index) => renderItem(item, item.id ?? index))}</div>
+        )}
+
+        <button
```
abdhalees, line 90: does this button need an aria label?

abdhalees (changes requested):

src/flashbar/index.tsx
DaemonCahill: Yes, it does. Great catch.

abdhalees (approved):

src/flashbar/index.tsx
```diff
@@ -1,85 +1,178 @@
...
-  // Motion will be disabled if any of the provided flash messages does not contain an `id`
-
-  const renderFlashItem = (item: FlashbarProps.MessageDefinition, index: number) => (
-    <Flash
-      key={item.id ?? index}
-      // eslint-disable-next-line react/forbid-component-props
-      className={clsx(isRefresh ? styles['flash-refresh'] : '')}
-      {...item}
-    />
+  useEffect(
```
just-boris, line 48: This should not be an effect: https://beta.reactjs.org/learn/you-might-not-need-an-effect#updating-state-based-on-props-or-state

src/flashbar/index.tsx
```diff
@@ -1,85 +1,178 @@
...
+   * flash messages does not contain an `id`.
+   */
+  const motionDisabled =
+    useReducedMotion(ref as any) || !isVisualRefresh || (items && !items.every(item => 'id' in item));
+
+  /**
+   * The `enableStackingOption` property is a hidden boolean that allows for teams
+   * to beta test the flashbar stacking feature.
+   */
+  const enableStackingOption = (restProps as any).enableStackingOption;
```
just-boris, line 44: Even if it is a beta, better to come up with a solid future-proof name for this property. I think, it should be `stackItems`, but open for other suggestions too.

src/flashbar/index.tsx
```diff
@@ -1,85 +1,178 @@
...
-      </>
+          </div>
+        )}
+
+        {isFlashbarStackExpanded && (
+          <div className={styles.expanded}>{items.map((item, index) => renderItem(item, item.id ?? index))}</div>
+        )}
+
+        <button
+          aria-label={!isFlashbarStackExpanded ? 'Expand stacked notifications' : 'Collapse stacked notifications'}
```
just-boris, line 91: Let's also expose new properties for this label right away? Even for beta releases, they have to be translatable

src/flashbar/index.tsx
```diff
@@ -1,85 +1,178 @@
...
+    );
+  }
+
   return (
     <div
       {...baseProps}
       className={clsx(baseProps.className, styles.flashbar, styles[`breakpoint-${breakpoint}`])}
       ref={mergedRef}
     >
-      <VisualContext contextName="flashbar">
```
just-boris, line 74: Why was this changed?

src/flashbar/index.tsx
DaemonCahill: @just-boris Is there an example somewhere in the code I should follow?

src/flashbar/index.tsx
DaemonCahill: The visual context function was inserting a `div` around the entire flashbar which applied to the toggle button as well. When I moved that `div` to wrap each flash individually it broke some existing style code relating to margins between flash items. Since the `div` does nothing but apply the visual context class, I added a new function to apply the appropriate class to each flash without the `div`.

src/flashbar/index.tsx
DaemonCahill: An important distinction that this property only turns on the option to stack for users, but the actual piece of state that controls whether or not the notifications are stacked is separate. I know @connorlanigan had some feedback on the name as well. I will start a Slack thread to see if we can achieve a quick consensus.

src/flashbar/index.tsx
just-boris: I think it should be `i18nStrings` property, similar to many other our components

src/flashbar/index.tsx
just-boris: not a big fan of removing the component here (because it will complicate future changes if visual contexts start having additional features), but fine to keep the change for now.

DaemonCahill: On this same point I (mistakenly) applied the visual context header class manually in the refactored app layout and content layout components. Inserting a div will absolutely break those components as well, so I think we will want to keep the div-less option. We can discuss in more detail if you want.

DaemonCahill: Okay! How would that work when applying a visual context class to an existing element, though?

just-boris (approved):
Looks good, thanks!

PR #306, opened 2022-09-23
POC: Components i18n

pan-kot:
### Description

Extract components translation strings that are reasonable to be injected by default.


pan-kot:
Closed as no longer relevant. The new solution is in work.

PR #319, opened 2022-09-30
chore: Add `aria-haspopup` to App layout toggle buttons

timogasda:
### Description

Add `aria-haspopup` to the App layout toggle buttons that open the drawers.

### How has this been tested?
Tested manually with a screen reader. Adjusted unit tests.

### Documentation changes

- [x] _No._

### Related Links

AWSUI-19119

<details>

#### Correctness

#### Security

#### Testing

</details>


src/app-layout/toggles/index.tsx
```diff
@@ -45,6 +45,7 @@ export const AppLayoutButton = React.forwardRef(
         iconName={iconName}
         disabled={disabled}
         ariaExpanded={ariaExpanded}
+        __nativeAttributes={ariaExpanded !== false ? {} : { 'aria-haspopup': true }}
```
connorlanigan, line 48: Could this be `ariaExpanded ? ...` instead of the explicit check against `false`? If not and this intentionally depends on specific behaviour for the value `undefined`, could we add a unit test that makes sure we don't accidentally break it in the future?

connorlanigan (changes requested):

src/app-layout/toggles/index.tsx
timogasda: Good point! I will have a look and either lock this down with a unit test or change the logic. But I think just the positive check didn't work, because it's actually either `false` or `undefined` (if I remember correctly).

timogasda: Okay, so this check is correct because we really only care about the `false` or `undefined` state here. And this is already covered by the unit test I modified in this PR. If I change this condition to something else, the test breaks.

pan-kot: The condition is a bit awkward. Why not change it to `{ 'aria-haspopup': !ariaExpanded }` and the test can be like `expect(...).toHaveAttribute('aria-haspopup', 'false')`?

src/app-layout/visual-refresh/app-bar.tsx
```diff
@@ -64,6 +64,7 @@ export default function AppBar() {
             className={testutilStyles['navigation-toggle']}
             ref={focusRefsNav.toggle}
             disabled={isAnyPanelOpen}
+            __nativeAttributes={{ 'aria-haspopup': true }}
```
pan-kot, line 67: Side note: why not expose the `ariaHaspopup` in the public API? If we have a good use case for it -- it is likely that our customers would have it as well.

src/app-layout/toggles/index.tsx
connorlanigan: So to summarise: `aria-haspopup` is set when `ariaExpanded` is `undefined` or when `ariaExpanded` is `true` - but I don't understand the meaning of why `undefined` should have this result. 
Why do we tie `haspopup` to `expanded` at all? Shouldn't `haspopup` _always_ be set for these buttons, since the menu is always available:
>The `aria-haspopup` attribute indicates the **availability** [of an] interactive popup element
> -- [MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-haspopup)

Maybe a new `ariaHaspopup` property on the `AppLayoutButtonProps` could make the API a bit clearer

pan-kot (approved):

connorlanigan (approved):

PR #323, opened 2022-10-03
Add visualSearch type to inputs for  adding search behavior to text inputs

jorycunningham:
### Description

Adds a `visualSearch` type to the InternalInput. This allow the inputs in select and autosuggest to use `type="text"` rather than `"search"` because role `combobox` (or any role) is not valid on type="search" inputs per [ARIA in HTML](https://www.w3.org/TR/html-aria/), while maintaining their search appearance.

Also updates a test in `select `that was looking for `type="search"`

### How has this been tested?
Ran npm test and visual regression.

### Documentation changes

- [ ] _No._

### Related Links

Fixes AWSUI-18954

<details>

#### Correctness

#### Security

#### Testing

- [X] _Changes are covered with new/existing unit tests?_
</details>


src/input/interfaces.ts
```diff
@@ -139,7 +139,7 @@ export interface InputProps
 }
 
 export namespace InputProps {
-  export type Type = 'text' | 'password' | 'search' | 'number' | 'email' | 'url';
```
avinashbot, line 142: It's probably better at this point to not make this a public value (i.e. by adding `variant: InputProps.Variant | 'visualSearch'` to `InternalInputProps`). The type is only needed for one niche internal combobox requirement, and we don't want to confuse people between using `search` and `visualSearch`.

avinashbot (changes requested):

src/__tests__/__snapshots__/documenter.test.ts.snap
```diff
@@ -5913,6 +5913,7 @@ be slightly different across browsers.",
           "number",
           "email",
           "url",
+          "visualSearch",
```
avinashbot, line 5916: Looks like this is still left over!

src/input/internal.tsx
```diff
@@ -155,6 +156,9 @@ function InternalInput(
 
   const mergedRef = useMergeRefs(ref, inputRef);
 
+  // type = "visualSearch" renders a type="text' input
+  attributes.type === 'visualSearch' ? (attributes.type = 'text') : attributes.type;
```
avinashbot, line 160: Looks like a slightly awkward way of writing the following, if I'm not mistaken:
```suggestion
if (attributes.type === 'visualSearch') {
  attributes.type = 'text';
}
```

---

Alternatively, you could make this change up in line 127:
```typescript
  type: type === 'visualSearch' ? 'text' : type,
```

avinashbot (changes requested):

avinashbot (approved):

PR #337, opened 2022-10-07
fix(date-range-picker): Get consistent label word breaks on narrow screens

johannes-weber:
### Description

Before this change, the word beak worked differently on Chrome,
FF and Safari on narrow screens (Safari and FF did not break the
label into two rows at all, Chrome did).

This change adds a work break right after the start or before the end time range throughout all browsers:

Screenshot 2022-10-07 at 16 07 29

### How has this been tested?

- I'm adopting existing screenshot tests to cover the narrow case.

- Checkout this PR, open the `/#/light/date-range-picker/with-default-date` example in Chrome, FF and Safari
 - When doing so, set the viewport width to 450px to test the narrow case.

### Documentation changes

- [x] _No._

### Related Links
- AWSUI-19387

<details>

#### Correctness

#### Security

#### Testing

</details>

### DIFF
--- a/src/date-range-picker/index.tsx
+++ b/src/date-range-picker/index.tsx
@@ -62,14 +62,14 @@ function renderDateRange(
 function BreakSpaces({ text }: { text: string }) {
   const tokens = text.split(/( )/);
   return (
-    <div style={{ whiteSpace: 'nowrap' }}>
+    <>
       {tokens.map((token, index) => (
         <React.Fragment key={index}>
-          {token}
+          {token.length > 1 ? <span className={styles['label-token-nowrap']}>{token}</span> : token}
           {token === ' ' && <wbr />}
         </React.Fragment>
       ))}
-    </div>
+    </>
   );
 }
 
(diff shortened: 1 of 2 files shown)


src/date-range-picker/index.tsx
```diff
@@ -62,14 +62,14 @@ function renderDateRange(
 function BreakSpaces({ text }: { text: string }) {
   const tokens = text.split(/( )/);
   return (
-    <div style={{ whiteSpace: 'nowrap' }}>
+    <>
       {tokens.map((token, index) => (
         <React.Fragment key={index}>
-          {token}
+          {token.length > 1 ? <span className={styles['label-token-nowrap']}>{token}</span> : token}
```
pan-kot, line 68: why this condition is needed?

pan-kot (approved):

src/date-range-picker/index.tsx
pan-kot: Got it, thanks

PR #359, opened 2022-10-12
fix(button-dropdown): Focus on first item of dropdown list when it's opened

abdhalees:
### Description

Focus on first menu item of dropdown list when button dropdown is opened so it will be announced in the screen reader

### How has this been tested?

- Added a unit test
- Tested manually, first menu item is announced by screen reader when the dropdown is opened

### Documentation changes

- [ ] _No._

### Related Links

AWSUI-18913

<details>

#### Correctness

#### Security

#### Testing

</details>

### DIFF
--- a/src/button-dropdown/__integ__/button-dropdown-items.test.ts
+++ b/src/button-dropdown/__integ__/button-dropdown-items.test.ts
@@ -17,3 +17,22 @@ test(
     expect(await browser.getWindowHandles()).toHaveLength(2);
   })
 );
+
+test(
+  'focus on first menu item when dropdown is opened',
+  useBrowser(async browser => {
+    const page = new ButtonDropdownPage('ButtonDropdown1', browser);
+    await browser.url('#/light/button-dropdown/simple');
+
+    await page.waitForVisible(page.getTrigger());
+    await page.openDropdown();
+
+    const firstTabbableMenuItemText = await page.getItemText('id1');
+
+    const focusedElement = await browser.execute(function () {
+      return document.activeElement!.textContent;
+    });
+
+    expect(focusedElement).toEqual(` ${firstTabbableMenuItemText} `);
+  })
+);
--- a/src/button-dropdown/internal.tsx
+++ b/src/button-dropdown/internal.tsx
@@ -17,6 +17,7 @@ import useForwardFocus from '../internal/hooks/forward-focus';
 import { usePrevious } from '../internal/hooks/use-previous';
 import InternalBox from '../box/internal';
 import { checkSafeUrl } from '../internal/utils/check-safe-url';
+import { useMergeRefs } from '../internal/hooks/use-merge-refs/index.js';
 
 const InternalButtonDropdown = React.forwardRef(
   (
@@ -74,6 +75,10 @@ const InternalButtonDropdown = React.forwardRef(
     const baseProps = getBaseProps(props);
 
     const dropdownRef = useRef<HTMLElement>(null);
+
+    const rootRef = useRef<HTMLDivElement>(null);
+    const mergedRootRef = useMergeRefs(rootRef, __internalRootRef);
+
     useForwardFocus(ref, dropdownRef);
 
     const clickHandler = () => {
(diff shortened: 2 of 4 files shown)


src/button-dropdown/category-elements/expandable-category-element.tsx
```diff
@@ -37,6 +38,12 @@ const ExpandableCategoryElement = ({
     }
   }, [expanded, highlighted]);
 
+  useEffect(() => {
```
pan-kot, line 41: Why do we need three useEffect calls?

I'd expect that to be performed only once in the component's root. Something like:

```
function InternalButtonDropdown() {
  // ...
  
  useEffect(() => {
    if (isOpen) {
      // focus first item
    }
  }, [isOpen])
  
  // ...
}
```

The "focus first item" part can be done in two ways:

1. Focus using selector: `buttonDropdownRef.current?.querySelector('.' + styles.item)?.focus()`.
2. Focus using reference:
```
function InternalButtonDropdown() {
  const firstItemRef = useRef(null)
  
  useEffect(() => {
    if (isOpen) {
      firstItemRef.current?.focus()
    }
  }, [isOpen])
  
  // ... <ItemsList ... firstItemRef={firstItemRef} />
}

function Item(props) {
  return <div ... ref={props.isFirst ? props.firstItemRef : undefined} />
}
```

What do you think?

pan-kot: The `useButtonDropdown` can probably return an extra property -- `isFirstItem` which, when taken an item would return true if and only if the item is the first visible item in the list. That can be used to assign the ref to it.

However, I don't mind using a selector as well.

pan-kot: Example using query selector: https://github.com/cloudscape-design/components/blob/dda400a1aec2bae02b199be0152423c9be4ac786/src/calendar/internal.tsx#L112

abdhalees: thanks for the suggestions andrei, I'll try to implement the query selector solution. 
the current solution is failing integ tests anyway so I have to change it

src/button-dropdown/internal.tsx
```diff
@@ -94,6 +94,12 @@ const InternalButtonDropdown = React.forwardRef(
       }
     }, [isOpen, wasOpen]);
 
+    useEffect(() => {
+      if (!wasOpen && isOpen && isInRestrictedView) {
```
pan-kot, line 98: Why checking `wasOpen` and `isInRestrictedView`?

If it serves some special purposes it would be nice to have dedicated tests for it.

abdhalees: removed `isInRestrictedView `

`wasOpen` is to make sure that the focus happens when the dropdown is previously closed. but after a bit of thinking it's seems trivial

src/button-dropdown/internal.tsx
```diff
@@ -94,6 +94,12 @@ const InternalButtonDropdown = React.forwardRef(
...
+        __internalRootRef?.current?.querySelector(`li:first-child [tabindex="-1"]`)?.focus();
```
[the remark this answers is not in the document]
just-boris: Instead of `!wasOpen && isOpen` combination, can you use `useEffectOnUpdate`?

src/button-dropdown/__integ__/button-dropdown-items.test.ts
```diff
@@ -17,3 +17,21 @@ test(
...
+    const page = new ButtonDropdownPage('ButtonDropdown1', browser);
+    await browser.url('#/light/button-dropdown/simple');
+    await page.waitForVisible(page.getTrigger());
+    await page.openDropdown();
+    const focusedElement = await browser.execute(function () {
+      return document.activeElement!.textContent;
+    });
+
+    const firstTabbableMenuItem = await browser.execute(
+      () => document.querySelector(`li [tabindex="-1"]`)!.textContent
```
pan-kot, line 33: I think it is better to avoid using the same implementation in the test and in the code. It is better to simply copy here the actual item's text content from the test page.

src/button-dropdown/internal.tsx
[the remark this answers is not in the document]
abdhalees: removed `wasOpen` it wasn't needed

pan-kot (approved):

pan-kot (approved):

PR #360, opened 2022-10-12
Use date cache in calendar and date range picker

pan-kot:
### Description

Use uniform dates memoization technique for calendar and date range picker.

### How has this been tested?

Existing tests.

### Documentation changes

- [x] _No._

### Related Links

<details>

#### Correctness

#### Security

#### Testing

</details>


pan-kot:
> What is the benefit of adding this caching mechanism for dates? will you add more information about why we are doing this?

The primary reason was to remove the `memoizedDate` function which not only performed the caching but also parsed the date causing confusion. Even more so, in one place we called it as `memoizedDate(formatDate(date))` which is formatting the date object to then parse it inside.

The second reason was to make the dates memoization aligned between calendar and date-range picker. In calendar, memoizeDate function was used while in the date-range picker we used the following:

```
const dateTime = date.getTime()

useEffect(() => {
  // ...
},
// mute eslint warnings as can't depend of the date itself as its reference might change
[dateTime])
```

src/calendar/internal.tsx
```diff
@@ -39,8 +39,12 @@ export default function Calendar({
   const gridWrapperRef = useRef<HTMLDivElement>(null);
   const [focusedDate, setFocusedDate] = useState<Date | null>(null);
 
+  const valueDateCache = useDateCache();
+  const focusedDateCache = useDateCache();
```
just-boris, line 43: Why do you have two separate caches here, but `src/date-range-picker/calendar/grids/index.tsx` uses a single cache for multiple values?

just-boris (approved):

src/calendar/internal.tsx
pan-kot: The values are passed to dependency arrays so using just one cache might cause a regression. I will check if the code can be further simplified as a follow-up.

PR #376, opened 2022-10-14
fix(charts): Make legend behave like an ARIA toolbar

avinashbot:
### Description

Change the role of the legend from a mysterious "application" to a more standard "toolbar". The interactions are actually mostly preserved here, with arrow keys for navigation. The focus is now on individual toolbar buttons, but the hover behavior is unchanged.

Though one choice I made was to detach the hover interactions and the focus interactions. It felt like a bad idea to throw the focus around without any explicit keyboard action from the user or with only a casual hover.

### How has this been tested?

The unit tests have been adapted to match the new behavior, but you're also welcome (and recommended!) to try with a screen reader. It should announce the individual toolbar buttons as you move around. On NVDA, it works with both browse and focus modes (though with browse mode, it only works on activation rather than focus, which is which).

### Documentation changes

- [x] _No._

### Related Links

<details>

#### Correctness

#### Security

#### Testing

</details>

### DIFF
--- a/src/internal/components/chart-legend/__tests__/chart-legend.test.tsx
+++ b/src/internal/components/chart-legend/__tests__/chart-legend.test.tsx
@@ -31,7 +31,13 @@ function renderChartLegend() {
 
   const focus = () =>
     act(() => {
-      wrapper.focus();
+      wrapper.find('[tabindex="0"]')?.focus();
+    });
+
+  const mouseover = (index: number) =>
+    act(() => {
+      wrapper.findAllByClassName(styles.marker)[index]!.fireEvent(new MouseEvent('mouseleave', { bubbles: true }));
+      wrapper.findAllByClassName(styles.marker)[index]!.fireEvent(new MouseEvent('mouseover', { bubbles: true }));
     });
 
   const pressLeft = () =>
@@ -46,23 +52,24 @@ function renderChartLegend() {
       wrapper.keydown(KeyCode.right);
     });
 
-  const findAllLabels = () => wrapper.findAll('li').map(getTextContent);
-
-  const findHighlightedLabels = () => wrapper.findAll(`li.${styles['marker--highlighted']}`).map(getTextContent);
-
-  const findDimmedLabels = () => wrapper.findAll(`li.${styles['marker--dimmed']}`).map(getTextContent);
-
-  const findFocusedLabels = () => wrapper.findAll(`li.${styles['marker--focused']}`).map(getTextContent);
+  const findAllLabels = () => wrapper.findAllByClassName(styles.marker).map(getTextContent);
+  const findHighlightedLabels = () => wrapper.findAllByClassName(styles['marker--highlighted']).map(getTextContent);
+  const findDimmedLabels = () => wrapper.findAllByClassName(styles['marker--dimmed']).map(getTextContent);
+  const findFocusedLabel = () =>
+    document.activeElement && document.activeElement !== document.body
+      ? document.activeElement.textContent?.trim()
+      : undefined;
 
   return {
     wrapper,
     focus,
+    mouseover,
     pressLeft,
     pressRight,
     findAllLabels,
     findHighlightedLabels,
     findDimmedLabels,
-    findFocusedLabels,
+    findFocusedLabel,
   };
 }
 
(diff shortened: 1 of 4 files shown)


src/internal/components/chart-legend/index.tsx
```diff
@@ -120,44 +125,45 @@ function ChartLegend<T>({
...
+                role="button"
                 key={index}
-                onMouseOver={handleMouseOver.bind(null, s.datum)}
-                onMouseLeave={handleMouseLeave}
                 className={clsx(styles.marker, {
                   [styles['marker--dimmed']]: isDimmed,
                   [styles['marker--highlighted']]: isHighlighted,
-                  [styles['marker--focused']]: isHighlighted && isFocused,
                 })}
-                aria-disabled={isDimmed ? true : undefined}
```
pan-kot, line 138: why aria-disabled was removed?

src/internal/components/chart-legend/index.tsx
```diff
@@ -120,44 +125,45 @@ function ChartLegend<T>({
...
+                ref={elem => {
+                  if (elem) {
```
pan-kot, line 143: would it be easier to assign the ref to the above element and use some selector e.g. `toolbarRef.current.querySelectorAll('div[role="button"]')[index]`?

src/internal/components/chart-legend/index.tsx
avinashbot: Cause they're not actually disabled since it's a toolbar and all options are selectable. If you use a screen reader cursor, you won't be able to navigate to the other options since they're marked as disabled.

src/internal/components/chart-legend/index.tsx
avinashbot: Either works! But I want to avoid using querySelector unless necessary, especially when refs are an option.

pan-kot:
I found a regression:

1. Open area chart test
2. Move focus to the chart and focus on the green series (p70)
3. Move focus to the legend - the p50 series is focused.

Expected the focused series to be in sync between chart plot and the legend.

src/internal/components/chart-legend/__tests__/chart-legend.test.tsx
```diff
@@ -73,35 +80,39 @@ describe('Chart legend', () => {
...
     expect(legend.findHighlightedLabels()).toEqual([series[1].label]);
     expect(legend.findDimmedLabels()).toEqual([series[0].label, series[2].label]);
-    expect(legend.findFocusedLabels()).toEqual([]);
+    expect(legend.findFocusedLabel()).toEqual(undefined);
+  });
+
+  test('should highlight on mouseover', () => {
+    const legend = renderChartLegend();
 
     legend.focus();
```
pan-kot, line 89: do we need to focus it first?

avinashbot: Nope, but the test is also checking that the focus doesn't move when one of the buttons is moused over. So highlight moves to 2, focus stays on 1 unless it's explicitly activated or moved to.

src/internal/components/chart-legend/index.tsx
```diff
@@ -120,44 +126,49 @@ function ChartLegend<T>({
...
+                    segmentsRef.current[index] = elem;
+                  } else {
+                    delete segmentsRef.current[index];
+                  }
+                }}
+                tabIndex={
+                  index === highlightedSeriesIndex || (highlightedSeriesIndex === undefined && index === focusedIndex)
```
pan-kot, line 151: Why the second check is needed? Can the legend be focused but not highlighted at the same time?

avinashbot: If there's no highlighted series, the last focused element is tracked. It's an implementation side effect basically, but also a recommend behavior for the ARIA toolbar (https://www.w3.org/WAI/ARIA/apg/patterns/toolbar/).

pan-kot:
Found another regression:

1. Open area chart test
2. Focus on purple (p80) series in the legend
3. Press TAB to focus on the next element
4. Press Shift+TAB to focus back

The p80 is focused. Expected the first item (p50) to be focused instead.

pan-kot:
I haven't checked the failing tests but it would be great to ensure all found regressions are covered with automation.

pan-kot:
When navigating over the legend with NVDA (Firefox and Chrome), the legend buttons are announced twice e.g. as "p80 button, p80".

pan-kot (approved):
LGTM

PR #381, opened 2022-10-14
Avoid reparenting in date picker that potentially causes an issue wit…

pan-kot:
…h focus handler

### Description

A customer reported an issue with the date picker focus behaviour. I couldn't reproduce it on my end but here is how it looks like (confirmed with a screen share):

1. The date picker is initially disabled;
2. The checkbox is clicked to enable the component;
3. User enters an invalid value to the date picker input and clicks outside, the onBlur handler is called and a validation error is shown;
4. User enters a valid value to the date picker input and clicks outside - the onBlur handler is not being called and the validation error is not cleared.

The issue is not reproducible with a different type of input e.g time input.
The issue is not reproducible if the date picker is initially enabled.

This is related to the use of the focus tracker. When the date picker input is disabled the trigger is rendered alone and the reference used by the focus tracker is not attached.

### How has this been tested?

Verified with the customer by sending an updated artefact.

### Documentation changes

- [x] _No._

### Related Links

* Ticket: AWSUI-19519

<details>

#### Correctness

#### Security

#### Testing

</details>


michaeldowseza (approved):

just-boris:
But this PR does not fix any reparenting. The hierarchy is still unequal between editable and readonly states

* div
   * Dropdown
        * trigger

* div
   * trigger

PR #386, opened 2022-10-18
fix(date-picker,date-range-picker): Fix scrolling on dropdown content

avinashbot:
### Description

This seemed to have been broken during some refactor along the way. 

### How has this been tested?

Visually. Reduce the screen height and you should see a scrollbar on the dropdown. If there's no height constraint, there should be no scrollbar.

There is a dev page for this but it's not being screenshot tested. I'll add that test separately.

### Documentation changes

- [x] _No._

### Related Links

AWSUI-19364

<details>

#### Correctness

#### Security

#### Testing

</details>

### DIFF
--- a/src/date-picker/index.tsx
+++ b/src/date-picker/index.tsx
@@ -169,7 +169,7 @@ const DatePicker = React.forwardRef(
             dropdownId={dropdownId}
           >
             {isDropDownOpen && (
-              <FocusLock autoFocus={true}>
+              <FocusLock className={styles['focus-lock']} autoFocus={true}>
                 <div {...focusVisible} tabIndex={0} className={styles.calendar} role="dialog" aria-modal="true">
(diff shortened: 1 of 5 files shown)


pan-kot:
Would it be possible to fix it in the dropdown component instead?

The same issue exists e.g. for property-filter editor and I couldn't fix it by adding similar styles.

pan-kot:
I think the solution does not cover small screens still. At some point the dropdown area is being cut-off. The scroll exists, but it is not possible to scroll to the very bottom.

Screenshot 2022-10-24 at 14 39 58

pan-kot:
I found the issue with property filter editor. There was an extra DIV that didn't have the necessary styles as well. I found and easy solution to ignore its styles by setting the display to `contents`: https://github.com/cloudscape-design/components/pull/405

PR #388, opened 2022-10-18
fix: Hotspot events inside a checkbox label unexpectedly toggle the checkbox

connorlanigan:
[A similar PR](https://github.com/cloudscape-design/components/pull/225) had previously been merged, but been reverted since it caused a regression.

#### The issue:
when a Hotspot is placed inside a checkbox label (or the label of a Toggle, Radio Button or Tile) and the user clicks on an element inside the Hotspot's Annotation popover, this click event also bubbles up to the checkbox's `<label>` element. This then triggers a duplicated/redirected event on the native `<input>` element.

While the bubbling of the first click event can be prevented with `e.stopPropagation`, this does not prevent the dispatching of the second event. According to [this StackOverflow answer](https://stackoverflow.com/a/67613046), this is known _"legacy-pre-activation behavior"_ and cannot be prevented since it is part of the WHATWG specification - the label->input event duplication/redirection happens before the actual event dispatching.

#### The change:

This change deactivates the legacy behaviour by replacing the `<label>` element with a `<div>`. To match the existing screenreader behaviour, it uses ARIA properties to connect the label to the checkbox. Event listeners are added to match the existing behaviour (but now the bubbling can be stopped with `e.stopPropagation()`).

The Hotspot component now has an additional event handler that prevents events from bubbling into the surrounding component.

### How has this been tested?

- Added a dev page where I performed manual testing
- Updated unit tests

- View the dev page locally and test its announcements and interactions with a screenreader

### Documentation changes

- [X] _No._

### Related Links

- AWSUI-19101

<details>

#### Correctness

#### Security

#### Testing

- [X] _Changes are covered with new/existing unit tests?_
</details>


src/toggle/internal.tsx
```diff
@@ -59,11 +60,16 @@ const InternalToggle = React.forwardRef<ToggleProps.Ref, InternalToggleProps>(
             type="checkbox"
             checked={checked}
             name={name}
-            onFocus={onFocus && (() => fireNonCancelableEvent(onFocus))}
-            onBlur={onBlur && (() => fireNonCancelableEvent(onBlur))}
-            onChange={onChange && (event => fireNonCancelableEvent(onChange, { checked: event.target.checked }))}
+            onFocus={() => fireNonCancelableEvent(onFocus)}
+            onBlur={() => fireNonCancelableEvent(onBlur)}
+            // empty handler to suppress React controllability warning
+            onChange={() => {}}
```
avinashbot, line 66: Same point as earlier. If this doesn't break anything.

src/internal/components/abstract-switch/index.tsx
```diff
@@ -77,10 +69,10 @@ export default function AbstractSwitch({
 
   return (
     <div {...rest} className={clsx(styles.wrapper, rest.className)} ref={__internalRootRef}>
-      <WrapperElement
-        {...wrapperAttributes}
+      <div
```
avinashbot, line 72: `<label>` also has one small styling difference - the cursor is `default` (instead of the caret you normally get when you hover over text). So we might want to change that in individual components too.

avinashbot (changes requested):

src/radio-group/radio-button.tsx
```diff
@@ -40,12 +36,21 @@ export default function RadioButton({
         <input
           {...nativeControlProps}
           type="radio"
+          ref={mergedRefs}
           name={name}
           value={value}
           checked={checked}
-          onChange={onChange && (() => fireNonCancelableEvent(onChange, { value }))}
+          // empty handler to suppress React controllability warning
+          onChange={() => {}}
```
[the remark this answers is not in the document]
connorlanigan: Events on the native element still work correctly, since they bubble up to the surrounding `.label-wrapper` element. We test this in our unit tests by using the `findInputByValue` test util.

src/internal/components/abstract-switch/index.tsx
connorlanigan: Oh, good catch, thanks! I'll update that

avinashbot (approved):

PR #389, opened 2022-10-18
fix(expandable-section): Reduce activation area

FlorianDr:
### Description

Reduces activation area of expandable section to icon + text. This also moves the conditional header `border-bottom` to the content. It is only disabled if there is content.  

### How has this been tested?

Screenshots

### Documentation changes

- [X] _No._

### Related Links

<details>

#### Correctness

#### Security

#### Testing

</details>

### DIFF
--- a/src/expandable-section/styles.scss
+++ b/src/expandable-section/styles.scss
@@ -34,9 +34,8 @@
 .trigger {
   cursor: pointer;
   box-sizing: border-box;
-  display: flex;
+  display: inline-flex;
   border: none;
-  width: 100%;
   line-height: awsui.$font-body-m-line-height;
   text-align: left;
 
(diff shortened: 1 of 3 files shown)


timogasda:
Why are we doing this? Why is it a "fix"?

And do we not think this is potentially a UX breaking change and/or regression?

src/expandable-section/styles.scss
```diff
@@ -124,6 +121,7 @@
 
   &-default {
     padding: awsui.$space-scaled-xs 0;
+    border-top: awsui.$border-divider-section-width solid awsui.$color-border-divider-default;
```
taheramr, line 124: The PR looks good but removing the `border-bottom` from the toggle and adding it to the content removes the border transitions

taheramr (changes requested):

FlorianDr:
> Could you explain a bit more why are we doing this? Why is it a "fix"?
> 
> And do we not think this is potentially a UX breaking change and/or regression?

Resolved in person.

PR #391, opened 2022-10-19
feat: Pagination w/o Numbers

32teeth:
### Description

I have a use case, where my pagination is **Prev / Next Token** based *(Doubly Linked List)* where i do not know how many items or pages exists. The service only provides ability to use **Prev / Next Token** and fetch the next batch of data

This is very similar to paginating through **_GitHub_** repositories if you have more than 50 repositories you get a next / back button

#### Presentation

`< >`

Thats it. That's all we need.

Screen Shot 2022-10-19 at 13 57 06

### DIFF
--- a/pages/onboarding/i18n.ts
+++ b/pages/onboarding/i18n.ts
@@ -10,9 +10,11 @@ export const tutorialPanelStrings: TutorialPanelProps.I18nStrings = {
   tutorialListDescription:
     'Use our walk-through tutorials to learn how to achieve your desired objectives within Amazon Transcribe.',
   tutorialListDownloadLinkText: 'Download PDF version',
+  labelTutorialListDownloadLink: 'Download PDF version of this tutorial',
   tutorialCompletedText: 'Tutorial completed',
   labelExitTutorial: 'dismiss tutorial',
   learnMoreLinkText: 'Learn more',
+  labelLearnMoreLink: 'Learn more about transcribe audio (opens new tab)',
   startTutorialButtonText: 'Start tutorial',
   restartTutorialButtonText: 'Restart tutorial',
   completionScreenTitle: 'Congratulations! You completed the tutorial',
--- a/src/__tests__/__snapshots__/documenter.test.ts.snap
+++ b/src/__tests__/__snapshots__/documenter.test.ts.snap
@@ -12772,9 +12772,19 @@ Object {
             "type": "string",
           },
           Object {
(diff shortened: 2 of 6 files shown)


arunim2405 (approved):

just-boris:
This feature can be achieved user-land with relatively low effort:

```
<SpaceBetween direction="horizontal" size="xs">
  <Button variant="link" iconName="angle-left">
    Previous
  </Button>
  <Button variant="link" iconName="angle-right" iconAlign="right">
    Next
  </Button>
</SpaceBetween>
```

https://codesandbox.io/s/cloudscape-page-navigation-buttons-0dqjvv?file=/src/App.js:198-422

Why does it have to be embedded into pagination component?

32teeth:
Updates on this?

just-boris:
We were waiting for an answer on this question

> Why does it have to be embedded into pagination component?

Before adding features to the common library we need to confirm if this use-case is common enough

32teeth:
> Why does it have to be embedded into pagination component?

Rather than leaving it up to the user to have 2 separate buttons which can create display shift or layout issues, manage it under one component under Pagination

just-boris:
Hello!

Bringing back notes from our internal team discussion

1. For handling token-based pagination we recommend using `<Pagination openEnd={true} />` pattern, as covered in https://github.com/cloudscape-design/components/issues/392 Any reason why this pattern did not work for you?
2. Could you give more details on "which can create display shift or layout issues"? How is putting the very same code inside our component helps here?

32teeth:
> Hello!
> 
> Bringing back notes from our internal team discussion
> 
> 1. For handling token-based pagination we recommend using `<Pagination openEnd={true} />` pattern, as covered in [[Bug]: Able to keep click next past end in Pagination with openEnd #392](https://github.com/cloudscape-design/components/issues/392) Any reason why this pattern did not work for you?
* Our specific design has prev and next only, NO numbers (see original post image)
> 2. Could you give more details on "which can create display shift or layout issues"? How is putting the very same code inside our component helps here?
* Quote is out of context.
  * 'Rather than leaving it up to the user to have 2 separate buttons which can create display shift or layout issues, manage it under one component under Pagination'
  * Instead of leaving it to the user to create two separate buttons, we can happily manage it under one roof called `<Pagination/>`

32teeth:
Hi, checking in?

32teeth:
Hoping this gets merged soon ;)

PR #399, opened 2022-10-24
chore: Integrate d3-scale as bundled dependency, upgrade d3-scale

johannes-weber:
### Description

With this change we add d3-scale as bundled dependency. This is because the recent `d3-scale` version bump (#371) the package and it's transitive dependency switched from CJS to ESM format. This format switch caused a many build  failures for consumers using jest (jest can't handle ESM by default, it needs to get transformed).
To avoid breaking consumer builds, we decided to ship `d3-scale` as bundled dependency.

Bundling is happening as part of the build process. First, TS compile the sources, in a second step, the dependency get's bundled with rollup.

Since we use the built artefacts in our unit tests, they will break in case of any changes in those bundled dependencies.

Additional bundle size: 
- 86kb (20kb gzipped)

### How has this been tested?

- tested manually.
- removed the jest-preset changes added in https://github.com/cloudscape-design/jest-preset/pull/11 - tests run green
- to test the changes locally, checkout this PR and remove the jest-preset changes as mentioned above and execute the unit tests.

### Documentation changes

- [x] _No._

### Related Links

- dependabot alert: https://github.com/cloudscape-design/components/security/dependabot/9

<details>

#### Correctness

#### Security

#### Testing

</details>


src/internal/components/cartesian-chart/scales.ts
```diff
@@ -1,6 +1,7 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
-import { scaleLinear, scaleLog, scaleTime, scaleBand, ScaleContinuousNumeric, ScaleTime, ScaleBand } from 'd3-scale';
+import { scaleLinear, scaleLog, scaleTime, scaleBand } from '../../vendor/d3-scale';
+import { ScaleContinuousNumeric, ScaleTime, ScaleBand } from 'd3-scale';
```
just-boris, line 4: Can we import types from the vendor folder too?

If possible, we should also update our eslint config ([this rule](https://github.com/cloudscape-design/components/blob/main/.eslintrc#L157)) to disallow importing d3-scale anywhere in the codebase.

just-boris:
This solution introduces concurrent writing to the `lib` folder. Both typescript and rollup write into `lib/components/internal/vendor` folder. In build it works, because tasks run in serial mode, but during development watch task may break this (typescript will overwrite expected rollup output). 

Can we add `"exclude": ["src/internal/vendor"]` to tsconfig?

Possibly, we can also run rollup and typescript in parallel

THIRD-PARTY-LICENSES
```diff
@@ -21,4 +21,228 @@ FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
 AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
 LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
-SOFTWARE.
\ No newline at end of file
+SOFTWARE.
+
```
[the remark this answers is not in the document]
johannes-weber: I've included the plugin it in the recent changes and added tests about the generated content to get aware of licence changes.

I decided to merge the d3-* related licences manually because I could not find a simple approach to fully automated that merge automatically in a reasonable amount of time without adding much extra build complexity. I suggest to start that automation when we are in a situation where we have to ship more 3rd party dependencies.

src/internal/components/cartesian-chart/scales.ts
johannes-weber: Good point! - I haven't thought about that approach to prevent importing d3 directly! Updated that, it's now part of the PR and d3-scale imports are highlighted by eslint.

src/internal/vendor/d3-scale-third-party-licenses.txt
```diff
@@ -0,0 +1,220 @@
+Name: d3-array
```
just-boris, line 1: Is this file expected to be in git? It is generated by rollup, isn't it?

gulpfile.js
```diff
@@ -22,12 +22,14 @@ const {
...
   themeableSource,
+  bundleVendorFiles,
 } = require('./build-tools/tasks');
 
 const quickBuild = series(
   clean,
   parallel(packageJSON, generateEnvironment, generateIcons, generateIndexFile, licenses),
-  parallel(generateCustomCssPropertiesMap, styles, typescript, testUtils)
+  parallel(generateCustomCssPropertiesMap, styles, typescript, testUtils),
+  bundleVendorFiles
```
pan-kot, line 32: Why do we need to bundle those in quick-build?

gulpfile.js
```diff
@@ -40,9 +42,17 @@ exports['test:a11y'] = a11y;
...
+      '!**/__tests__/**',
+      '!**/__integ__/**',
+      '!**/__a11y__/**',
+      '!src/internal/vendor/**/*.ts',
+    ],
     typescript
   );
   watch(['src/test-utils/dom/**/*.ts', '!src/test-utils/dom/index.ts'], testUtils);
   watch(['style-dictionary/**/*.ts', 'src/**/*.scss'], styles);
+  watch(['src/internal/vendor/**/*.ts', '!src/internal/vendor/**/__tests__/**'], series(typescript, bundleVendorFiles));
```
pan-kot, line 57: why do we watch the vendor files at all?

src/internal/vendor/d3-scale-third-party-licenses.txt
johannes-weber: Yes, it is generated by rollup. I added it to the repository where I added a snapshot test which fails incase of any changes in the licences content. This mechanism is intended to make us aware of any licence changes in the bundled 3rd party packages.

gulpfile.js
johannes-weber: In order to have the bundled dependency already inside the lib folder. We use the source in our lib folder and run our tests against, build the demo pages, etc. With that we realize quickly that some of the compiled/bundled code does not work as expected.

gulpfile.js
johannes-weber: Because we want to compile and bundle the changes and have them reflected in the lib folder (related to the answer on your comment about quick-build above).

The alternative would be to generate the bundled dependency in a separate step which is not part of watch or quick build. With that we'd get feedback about any issues quiet late in our development workflow.

pan-kot: My expectation was that the vendor deps can be only bundled once. Would it be possible for example to create a custom build step that only runs as an npm post-install hook?

pan-kot: Is there value in running the bundling with every build?

src/internal/vendor/d3-scale-third-party-licenses.txt
just-boris: Yes, I get why there is snapshot in git. But why `.txt` file too?

gulpfile.js
just-boris: Quick build includes `clean` task which removes `lib` folder. This is why running vendor task is necessary.

We can look into ways of refactoring it further, but how long is this task? If under 10 seconds, does it worth our time?

src/internal/vendor/d3-scale-third-party-licenses.txt
johannes-weber: That's because the licence plugin first attaches the banner followed by extracting the licence information of the bundled packages. Because of this order it would fail if the pre-generated licence file is not present. 

Assuming there are changes in the 3p licences happening during a build, the tests would fail because the generated file would contain the latest version. I'd love to have the licence extraction before the banner get's attached. I'm looking into the plugin to see why this is the case to understand better and consider about opening a PR for the licence plugin to get that solved.

johannes-weber: That's because the licence plugin first attaches the banner followed by extracting the licence information of the bundled packages. Because of this order it would fail if the pre-generated licence file is not present. 

Assuming there are changes in the 3p licences happening during a build, the tests would fail because the generated file would contain the latest version. I'd love to have the licence extraction before the banner get's attached. I'm looking into the plugin to see why this is the case to understand better and consider about opening a PR for the licence plugin to get that solved.

gulpfile.js
johannes-weber: I agree that there are ways to refactor that tasks and this may includes more modifications in the way we build the components.
The `internal/vendor` related tasks take around 5-7 sec (TS compile 5-6sec, bundling 1 sec).

gulpfile.js
johannes-weber: Yes, the bundled artifacts are stored in the `lib` folder which gets recreated every build.
I was exploring the option for having a one time bundling in place, but decided to not move in that direction.

Having a one time bundle, would mean we need to store the bundled JS files inside our src folder (or somewhere outside the lib folder) and we need to move them into the lib folder when building the components as part of the build process. I think the current approach is simpler compared to one-time bundling.

With this approach, we ensure to bundle the latest version into /lib, and consume the latest updates licence information.

What alternative would you suggest to not bundle with every build?

gulpfile.js
pan-kot: How about we build the vendor files at npm preinstall hook and on build/quick-build we copy from e.g. `vendor` to `lib/components/internal/vendor`?

pan-kot: For example: https://github.com/cloudscape-design/components/pull/429

just-boris (approved):

gulpfile.js
johannes-weber: I like this approach. There is one problem about transpiling the bundled d3-scale dependency to es15 ([see related comment in the PR](https://github.com/cloudscape-design/components/pull/429#discussion_r1009240687)).

I think with the latest update the PR is in a good shape. Let me know if this is a hard blocker for you.

just-boris:
Can we already merge this one to start things going and review that refactoring later (because it has pros and cons)

pan-kot (approved):

PR #404, opened 2022-10-24
Fix accessible labeling of External Link Icon in Button component

jorycunningham:
### Description
This PR resolves an issue where the "external link" icon did not have an accessible name in the External Link form of our Button component.

#### This commit establishes the following logic:

1. If there is an `externalIconAriaLabel` prop but no `ariaLabel` prop, the icon gets wrapped in a `role="img"` with an aria-label of "{`externalIconAriaLabel`}" so that it becomes part of the button's accessible name from contents.
2. If there is BOTH an `externalIconAriaLabel`  AND `ariaLabel` prop, the anchor tag itself receives an aria-label of "{`ariaLabel` }, {`externalIconAriaLabel`}"

#### These changes were made to enable this:

- Added `externalIconAriaLabel` prop to the button's interface
- Updated permutations page to demonstrate this functionality
- Added logic to build the anchor's aria-label from multiple sources when needed
- Added logic to pass the `externalIconAriaLabel` value down into the Internal Icon as `alt`
- Added logic to InternalIcon that adds a role="img" and and aria-label with the `alt` value if there is alt present.

### How has this been tested?

Built local permutations, tested manually and with automated a11y checks

To test, create two buttons like these

```
<!-- this renders a link with accessible name 'Report a bug, opens in a new window'
<Button
      href="https://example.com"
      iconAlign="right"
      iconName="external"
      target="_blank"
      externalIconAriaLabel = 'opens in a new window'
    >
      Report a bug
    </Button>

<!-- this renders a link with accessible name 'Report a bug accessible label, opens in a new window'
<Button
      href="https://example.com"
      iconAlign="right"
      iconName="external"
      target="_blank"
     ariaLabel="Report a bug accessible label"
      externalIconAriaLabel = 'opens in a new window'
    >
      Report a bug
    </Button>
```

NOTE: The visual regression test for the Button/permutations page has new buttons added to it to test external link icon

### Documentation changes

- [ ] _No._

### Related Links

AWSUI-19578

<details>

#### Correctness

#### Security

#### Testing

</details>


src/button/icon-helper.tsx
```diff
@@ -27,7 +27,7 @@ function getIconAlign(props: ButtonIconProps) {
 }
 
 function IconWrapper({ iconName, iconUrl, iconAlt, iconSvg, iconSize, ...props }: ButtonIconProps) {
-  if (!iconName && !iconUrl && !iconSvg) {
+  if (!iconName && !iconUrl && !iconSvg && !iconAlt) {
```
jorycunningham, line 30: I want to especially call out this change. It makes it so that icons that have iconAlt will be labeled regardless of other factors. This seems safe to me but I want folks who are more familiar to confirm

timogasda: If I understand this change correctly, it will cause all Icons to disappear if they don't have `iconAlt` set, right? That would be a breaking change.

timogasda (changes requested):
Left one specific comment.

I'm also not sure that we actually need a specific property to label one specific icon here. You can just use `ariaLabel` to do that.

I understand that we are trying to mimic the `Link` component here (and I'm usually always a big fan of API consistency), but I think this is a little different. The Link component only has that one type of icon but the Button component can have a wide range of icons.

jorycunningham:
Closing per @YueyingLu

PR #406, opened 2022-10-25
Update popover placement priorities to keep axis if possible

pan-kot:
### Description

Updated top/bottom popover placement priorities to make popover maintain its axis (vertical/horizontal) when possible.

### How has this been tested?

Screenshot tests

### Documentation changes

- [x] _No._

### Related Links

AWSUI-19491

<details>

#### Correctness

#### Security

#### Testing

</details>


rubencarvalho (dismissed):

src/popover/utils/positions.ts
```diff
@@ -21,26 +21,26 @@ export const PRIORITY_MAPPING: Record<PopoverProps.Position, InternalPosition[]>
   top: [
```
avinashbot, line 21: Woah, we can't remove priorities here. I'm all for re-ordering, but we should specify every priority.

avinashbot (changes requested):

src/popover/utils/positions.ts
pan-kot: I think that might be a diff issue.

Compare:

main: https://github.com/cloudscape-design/components/blob/main/src/popover/utils/positions.ts#L20
proposed:  https://github.com/cloudscape-design/components/blob/0785aff77069837c28737bac61976bf237b9c861/src/popover/utils/positions.ts#L20

pan-kot: The idea is to maintain the axis. So if the intended placement was `top` or `bottom` - those non-center top/bottom placements can have higher prio than side placements.

avinashbot: > The idea is to maintain the axis.

I was referring to this segment in the edit. Why make `bottom-center` the second item rather than the fourth?

```
  top: [
    'top-center',
    'bottom-center',
    'right-top',
    'right-bottom',
    ...
```

pan-kot: Ah, I see.

So we consider center placement as more important than top/bottom. Do you think we need to challenge that as well?

avinashbot: Going back to the original design rationale for the ordering (that people read top to bottom, and in the languages we support, left to right), maybe. But that's a conversation for later that should include designers, because it's make a huge change in the system, especially for patterns like copy to clipboard.

pan-kot: Ok, I agree with that.

rubencarvalho (approved):

rubencarvalho (approved):
I like we now prioritise all `top` options (center, right, left) first before going to `bottom` and vice-versa 👍

avinashbot (approved):

PR #407, opened 2022-10-25
feat(table): New header visuals for full-page table columns

connorlanigan:
### Description

This change updates the column headers of the full-page variant of the Table component. They are now using the same colour scheme as the body of the table, instead of being part of the dark page header.

Screenshot of the full-page table after this proposed change

### How has this been tested?

Local testing in the dev pages

### Documentation changes

- [X] _No._

### Related Links

- AWSUI-19627

<details>

#### Correctness

#### Security

#### Testing

</details>

### DIFF
--- a/src/container/styles.scss
+++ b/src/container/styles.scss
@@ -99,7 +99,21 @@
   }
 
   &-variant-full-page.header-stuck {
-    box-shadow: awsui.$shadow-sticky;
+    box-shadow: none;
+
+    &::after {
+      content: '';
+
+      position: absolute;
+      right: 0;
+      left: 0;
+      bottom: 0;
+      top: 0;
+
+      box-shadow: awsui.$shadow-sticky;
+      // This polygon only shows the part of the shadow that is lower than the element itself.
+      clip-path: polygon(-999% 100%, 999% 100%, 999% 999%, -999% 999%);
+    }
   }
 }
 
--- a/src/split-panel/icons/bottom-icon-refresh.tsx
+++ b/src/split-panel/icons/bottom-icon-refresh.tsx
@@ -3,13 +3,16 @@
 import React from 'react';
 import { getClassName, SVGTableRowProps } from './side-position-refresh';
 
-const TableRow = ({ offset }: SVGTableRowProps) => {
+const TableRow = ({ offset, isHeader }: SVGTableRowProps) => {
   const offsetTop = 0.4482;
   const offsetBottom = 3.4482;
   return (
-    <g transform={`translate(0, ${offset})`} className={getClassName('disabled')}>
+    <g transform={`translate(0, ${offset})`} className={getClassName(isHeader ? 'column-header' : 'disabled')}>
       <path d={`M31 ${offsetTop}H34V${offsetBottom}H31V${offsetTop}Z`} />
-      <path d={`M39 ${offsetTop}H63V${offsetBottom}H39V${offsetTop}Z`} className={getClassName('secondary')} />
+      <path
+        d={`M39 ${offsetTop}H63V${offsetBottom}H39V${offsetTop}Z`}
+        className={!isHeader ? getClassName('secondary') : undefined}
+      />
       <path d={`M135 ${offsetTop}H155V${offsetBottom}H135V${offsetTop}Z`} />
       <path d={`M158 ${offsetTop}H202V${offsetBottom}H158V${offsetTop}Z`} />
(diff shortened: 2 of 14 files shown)


src/split-panel/icons/side-position-refresh.tsx
```diff
@@ -6,17 +6,24 @@ import styles from '../styles.css.js';
...
 
-const TableRow = ({ offset, separator = true }: SVGTableRowProps) => {
+const TableRow = ({ offset, separator = true, isHeader }: SVGTableRowProps) => {
   const offsetTop = 0.4482;
   const offsetBottom = 3.4482;
   return (
-    <g transform={`translate(0, ${offset})`} className={getClassName('disabled')}>
+    <g
+      transform={`translate(0, ${offset})`}
+      className={isHeader ? getClassName('column-header') : getClassName('disabled')}
```
fa7ad, line 20:
```suggestion
      className={getClassName(isHeader ? 'column-header': 'disabled')}
```

I think this variation is slightly easier to read, only apply if you agree.

src/split-panel/icons/bottom-icon-refresh.tsx
```diff
@@ -3,13 +3,19 @@
...
 
-const TableRow = ({ offset }: SVGTableRowProps) => {
+const TableRow = ({ offset, isHeader }: SVGTableRowProps) => {
   const offsetTop = 0.4482;
   const offsetBottom = 3.4482;
   return (
-    <g transform={`translate(0, ${offset})`} className={getClassName('disabled')}>
+    <g
+      transform={`translate(0, ${offset})`}
+      className={isHeader ? getClassName('column-header') : getClassName('disabled')}
```
fa7ad, line 12: same suggestion as the other comment

src/container/styles.scss
```diff
@@ -99,7 +99,21 @@
...
+
+      position: absolute;
+      right: 0;
+      left: 0;
+      bottom: 0;
+      top: 0;
+
+      box-shadow: awsui.$shadow-sticky;
+      // This polygon only shows the part of the shadow that is lower than the element itself.
+      clip-path: polygon(-999% 100%, 999% 100%, 999% 999%, -999% 999%);
```
fa7ad, line 115: Does the indicated browser support cover our browserlist?

https://caniuse.com/css-clip-path

src/table/header-cell/index.tsx
```diff
@@ -70,6 +72,7 @@ export function TableHeaderCell({
         [styles['header-cell-disabled']]: sortingDisabled,
         [styles['header-cell-ascending']]: sortingStatus === 'ascending',
         [styles['header-cell-descending']]: sortingStatus === 'descending',
+        [styles['header-cell-hidden']]: hidden,
```
fa7ad, line 75: Does introducing a hidden state change the experience for screen-reader/assistive-technology?

fa7ad:
Looks like codecov is complaining about coverage, does it make sense to add some unit tests? Maybe tests for the new props in some of the components?

src/container/styles.scss
connorlanigan: Yes, `clip-path` is supported in all browsers that we support. I've also verified this manually in Chrome, Firefox & Safari (one version each).

src/table/header-cell/index.tsx
connorlanigan: The "hidden" state already exists in the code, this change just modifies the border colour of those hidden cells.
When the sticky table header is enabled, there are actually _two_ headers: [one that scrolls with the table and keeps the layout correct](https://github.com/cloudscape-design/components/blob/main/src/table/internal.tsx#L231-L233), and [a duplicated header that stays stuck to the screen](https://github.com/cloudscape-design/components/blob/main/src/table/internal.tsx#L182-L193). From a visual standpoint, only one header is visible. The duplicated header [is also hidden from assistive technologies](https://github.com/cloudscape-design/components/blob/main/src/table/sticky-header.tsx#L53).
There is a short moment where technically both headers are visible, which happens when you scroll just a bit so that the font size of the page heading changes - there is an (intentionally) empty space between the now-stuck column headers and the first row. This empty space is made up of these "hidden" cells, and they should not have any border (in contrast to the non-hidden cells of the now-stuck header).

src/split-panel/icons/side-position-refresh.tsx
connorlanigan: Good idea, will update in the next revision

fa7ad (approved):

PR #417, opened 2022-10-27
Revert "Decouple popover and annotation context"

just-boris:
Reverts cloudscape-design/components#413

This broke visual regression tests


FlorianDr (approved):
LGTM

PR #437, opened 2022-11-01
fix: Ensure that popover focus lock works as expected

gethinwebster:
### Description

Prevents some unexpected side-effects from popover focus lock, specifically when the popover
is placed inside a focusable element (i.e. if any parent has `tabindex` set).

Recreation of the problem before fix: https://codesandbox.io/s/recursing-tharp-7ybx7d?file=/src/App.js

### How has this been tested?

Local testing & new tests.

### Documentation changes

- [X] _No._

### Related Links

AWSUI-19360

<details>

#### Correctness

#### Security

#### Testing

</details>


fa7ad (commented):
> [popover-placement-test-ts-L66](https://github.com/cloudscape-design/components/pull/437/files#file-src-popover-__integ__-popover-placement-test-ts-L66)

Is FocusLock affecting the CSS somehow? the integration test for placement seems to be failing?

gethinwebster:
The popover dismiss button was being focused while the popover was still repositioning itself, therefore it was offscreen & caused the browser to reset scroll. That in turn affected the positioning logic.

fa7ad (dismissed):

src/popover/body.tsx
```diff
@@ -50,6 +49,10 @@ export default function PopoverBody({
     [onDismiss]
   );
 
+  // used to ensure that we only autoFocus the dismiss button
+  // once the popover has been repositioned
+  useEffect(() => setRendered(true), []);
```
pan-kot, line 54: I guess the rendered is set to true when the popover is opened, not repositioned?

pan-kot: Why is it necessary to defer the auto-focus?

src/popover/internal.tsx
```diff
@@ -57,6 +57,8 @@ function InternalPopover(
   const clickFrameId = useRef<number | null>(null);
 
   const [visible, setVisible] = useState(false);
+  const [wasVisible, setWasVisible] = useState(false);
+  const [dismissedViaDocumentClick, setDismissedViaDocumentClick] = useState(false);
```
pan-kot, line 61: it is possible to simplify the state by replacing `visible` with some `displayState` union, like:

```
type DisplayState = { type: 'hidden' } | { type: 'visible' } | { type: 'dismissed', withClick: boolean }
```

src/popover/body.tsx
gethinwebster: Yes, but that allows the parent PopoverContainer to complete its initial positioning logic before we try to focus the popover content. Otherwise, the PopoverBody may be outside the visible screen (that's how some of the positioning logic is done), and therefore can cause unwanted scrolling when it's focused.

src/internal/components/chart-popover/index.tsx
```diff
@@ -77,6 +78,12 @@ export default function ChartPopover({
     };
   }, [container, onDismiss]);
 
+  const dismissButtonRefCallback = useCallback((element: ButtonProps.Ref) => {
+    if (element) {
+      element.focus();
```
pan-kot, line 83: why is this necessary?

src/popover/internal.tsx
```diff
@@ -99,6 +102,19 @@ function InternalPopover(
...
   }, []);
 
+  useEffect(() => {
+    if (
+      visible === false &&
+      wasVisible === true &&
+      document.activeElement === document.body &&
+      !dismissedViaDocumentClick
+    ) {
+      triggerRef.current?.focus();
```
pan-kot, line 112: It is possible to do it the moment dismiss happens? E.g. when the dismiss button is pressed or Esc keypress is intercepted?

Having this as a side-effect often leads to tricky issues.

pan-kot: Or maybe it is possible to change onDismiss:

```
const onDismiss = useCallback(() => {
    setVisible(false);
    if (shouldFocusTrigger()) {
      triggerRef.current?.focus();
    }
}, []);
```

The shouldFocusTrigger part can check the last interaction -- if it was mouse-down or TAB keypress -- ignore, otherwise -- focus.

src/popover/body.tsx
pan-kot: Is there some deferring logic in the library focus-lock? We can maybe update our implementation with e.g. adding a 0-timeout here: https://github.com/cloudscape-design/components/blob/main/src/internal/components/focus-lock/index.tsx#L31

src/internal/components/chart-popover/index.tsx
gethinwebster: To focus the dismiss button when the popover is already open, but is changed into "locked" mode by clicking/pressing enter on a specific point in the chart. The focus-lock library seemed to handle this magically somehow, but doing it more explicitly seems to make sense. Perhaps it would make sense to move this logic elsewhere though, I'll look into this

gethinwebster: I've looked, and don't think it's reasonable/sensible to move this logic elsewhere: it would require passing refs through multiple layers of components, and adding the logic separately for each chart type.

src/popover/internal.tsx
gethinwebster: Seems a good simplification, taking a look

src/popover/body.tsx
gethinwebster: I've not been able to get to the bottom of why it worked before, but it seems like this shouldn't be the responsibility of the FocusLock component, but rather of the component that uses it, to ensure it's ready before requesting focus lock.

src/popover/internal.tsx
```diff
@@ -58,20 +58,29 @@ function InternalPopover(
...
-  const onTriggerKeyDown = useCallback((event: React.KeyboardEvent) => {
-    if (event.keyCode === KeyCode.tab || event.keyCode === KeyCode.escape) {
-      setVisible(false);
-    }
-  }, []);
+  const onTriggerKeyDown = useCallback(
+    (event: React.KeyboardEvent) => {
+      if (event.keyCode === KeyCode.tab || event.keyCode === KeyCode.escape) {
+        setVisible(false);
+        focusTrigger();
```
pan-kot, line 79: I think the trigger should not be focused when received TAB keypress. The only exception is when expandToViewport is true -- in that case we can't focus elsewhere so focusing on the trigger is reasonable.

gethinwebster: good catch, this handler doesn't need to refocus ever in fact, as the trigger is by definition already focused...

src/mixed-line-bar-chart/__tests__/mixed-chart.test.tsx
```diff
@@ -928,8 +928,6 @@ describe('Details popover', () => {
 
     wrapper.findChart()!.fireEvent(new MouseEvent('mousedown', { bubbles: true }));
 
-    expect(wrapper.findDetailPopover()!.findDismissButton()!.getElement()).toHaveFocus();
```
gethinwebster, line 931: this is now tested in integ tests (see above) as I was unable to get it to work stably here (and testing such interactions in unit tests anyway can be quite difficult/problematic)

src/internal/components/chart-popover/index.tsx
gethinwebster: refactored to make this hopefully all a lot clearer

src/popover/body.tsx
gethinwebster: went with an approach of focusing without scrolling instead, so that the exact timing is not important

src/annotation-context/annotation/annotation-popover.tsx
```diff
@@ -78,16 +77,9 @@ export function AnnotationPopover({
...
-      if (element) {
-        element.focus({ preventScroll: true });
-        // Falls back to alignTop on IE11
-        scrollElementIntoView(trackRef.current ?? undefined);
-      }
-    },
-    [trackRef]
-  );
+  useEffect(() => {
+    scrollElementIntoView(trackRef.current ?? undefined);
```
pan-kot, line 81: When this is fired?

From the code it looks like the scroll is made only once - when the annotation popover is rendered (but not when it is opened or reopened)?

src/mixed-line-bar-chart/chart-container.tsx
```diff
@@ -284,12 +284,13 @@ export default function ChartContainer<T extends ChartDataTypes>({
...
   }, [highlightedX, showPopover]);
 
-  const onSVGMouseDown = () => {
+  const onSVGMouseDown = (e: React.MouseEvent<SVGSVGElement, MouseEvent>) => {
     if (isPopoverOpen) {
       if (isPopoverPinned) {
         dismissPopover();
       } else {
         pinPopover();
+        e.preventDefault();
```
pan-kot, line 293: why prevent default is needed?

gethinwebster: Otherwise the click event continues, causing focus to be moved to the clicked element (the chart itself)

src/popover/body.tsx
```diff
@@ -34,12 +31,12 @@ export default function PopoverBody({
   children,
   onDismiss,
   variant,
-  returnFocus = true,
   overflowVisible,
-  dismissButtonRef,
   className,
 }: PopoverBodyProps) {
   const labelledById = useUniqueId('awsui-popover-');
+  const [dismissButtonFocused, setDismissButtonFocused] = useState(false);
```
pan-kot, line 38: As this state is only used inside an effect it can be implemented with a ref to avoid unnecessary re-render:

```
const dismissButtonFocusedRef = useRef(false)

useEffect(() => {
  if (showDismissButton && !dismissButtonFocusedRef.current) {
    dismissButtonRef.current?.focus({ preventScroll: true });
  }
  dismissButtonFocusedRef.current = showDismissButton
}, [showDismissButton])
```

src/annotation-context/annotation/annotation-popover.tsx
pan-kot: Oh, I see - thanks. I think it would make sense to explain this in the comments above the use-effect.

src/mixed-line-bar-chart/chart-container.tsx
pan-kot: Do we need to apply similar change to area-chart and pie-chart then?

gethinwebster: yes actually, seems these weren't covered by any tests, added the change & tests for these two as well

pan-kot (approved):
LGTM!

PR #440, opened 2022-11-01
feat: Focus on a step header after wizard content updated

Al-Dani:
### Description

Programmatically set focus on the  wizard step header when the content is updated via navigation.

Provides visibility on content updates for screen reader users.

### How has this been tested?

New integration test added.

### Documentation changes

- [x] _No._

### Related Links

<details>

#### Correctness

#### Security

#### Testing

</details>


Al-Dani:
Will update unit tests

just-boris:
What is happening in this PR? Are you sure it has to be a focus change and not live announcement (see https://github.com/cloudscape-design/components/pull/377 for example). Where can I read more about the request?

src/wizard/wizard-form-header.tsx
```diff
@@ -1,17 +1,18 @@
...
-import React from 'react';
+import React, { MutableRefObject } from 'react';
 import { useDynamicOverlap } from '../app-layout/visual-refresh/hooks/use-dynamic-overlap.js';
 import styles from './styles.css.js';
 
 interface WizardFormHeaderProps {
   children: React.ReactNode;
   isMobile: boolean;
   isVisualRefresh: boolean;
+  __internalHeaderRef?: MutableRefObject<any> | null;
```
fa7ad, line 12: Can we not use `any`?

If a better type can't be found, maybe we can go with `MutableRefObject<HTMLDivElement>` or `MutableRefObject<HTMLElement>`

src/wizard/wizard-form.tsx
```diff
@@ -43,6 +44,14 @@ export default function WizardForm({
   const isLastStep = activeStepIndex >= steps.length - 1;
   const skipToTargetIndex = findSkipToTargetIndex(steps, activeStepIndex);
   const isMobile = useMobile();
+  const stepHeaderRef = useRef<HTMLDivElement>(null);
+
+  useEffectOnUpdate(() => {
+    if (stepHeaderRef && stepHeaderRef.current) {
+      const headers = stepHeaderRef.current.querySelectorAll('h1');
```
fa7ad, line 51: Can't we use ref to target `InternalHeader` directly (rather than traverse through the DOM)?

Al-Dani: good catch, thank you! I missed `InternalHeader` ref

fa7ad (approved):

pan-kot:
> What is happening in this PR? Are you sure it has to be a focus change and not live announcement (see #377 for example). Where can I read more about the request?

I also wonder if it is a good thing to focus on something that is only focusable programmatically? I would imagine that might cause certain confusion. Typically we use programmatic focus for keyboard navigation.

src/header/internal.tsx
```diff
@@ -60,7 +60,7 @@ export default function InternalHeader({
             isRefresh && styles[`title-variant-${variantOverride}-refresh`]
           )}
         >
-          <HeadingTag className={clsx(styles.heading, styles[`heading-variant-${variantOverride}`])}>
+          <HeadingTag className={clsx(styles.heading, styles[`heading-variant-${variantOverride}`])} tabIndex={-1}>
```
pan-kot, line 63: I am curious if it is safe to allow all headers to accept focus. Some tools might rely on the presence of tab-index -- would make sense to check how screen-readers react to that.

src/wizard/wizard-form.tsx
pan-kot: I would change the selector to be `stepHeaderRef.current.querySelector('[tab-index]')` to not rely on tag name to which the tab-index is attached. Also, no need to query all as we only use one anyways.

src/wizard/wizard-form.tsx
```diff
@@ -88,7 +103,9 @@ export default function WizardForm({
         errorText={errorText}
         errorIconAriaLabel={i18nStrings.errorIconAriaLabel}
       >
-        {content}
+        <div role="region" aria-label={title}>
```
pan-kot, line 106: The wizard test-util wrapper inherits from the form wrapper which has method `findContent`. We might want to override the method to ensure it returns the client `content` ignoring the new region wrapper. A dry-run to live would also make sense.

pan-kot (changes requested):
Added comments to the code.

Additionally, I'd love to discuss if alternative solutions are possible as focusing on the header is questionable: the users loose the ability to navigate fast by keep pressing "Enter" and the header is only focusable programmatically which might cause confusion for both keyboard and screen-reader user groups.

src/header/styles.scss
```diff
@@ -45,7 +46,7 @@
 }
 
 .main {
-  @include styles.text-flex-wrapping;
```
Al-Dani, line 48: changed to general `text-wrapping`, as `flex-wrapping` has `overflow: hidden`, which hides focus outline

Al-Dani:
> Additionally, I'd love to discuss if alternative solutions are possible as focusing on the header is questionable

Andrei, thank you for the input! Focusable header is a solution that we landed during a11y meetings. Screen reader users should be aware of the content changes and also understand the context of presented content.

pages/wizard/simple.page.tsx
```diff
@@ -19,9 +20,14 @@ const steps: WizardProps.Step[] = [
   },
   {
     title: 'Step 2',
+    info: <Link variant="info">Info</Link>,
     content: (
       <div className={styles['step-content']}>
-        <div id="content-text">Content 2</div>
```
pan-kot, line 24: The `#content-text` id is used by wizard integ tests.

src/header/internal.tsx
```diff
@@ -60,7 +64,11 @@ export default function InternalHeader({
             isRefresh && styles[`title-variant-${variantOverride}-refresh`]
           )}
         >
-          <HeadingTag className={clsx(styles.heading, styles[`heading-variant-${variantOverride}`])}>
+          <HeadingTag
```
pan-kot, line 67: Would it be possible to avoid modifying our header component? You can maybe create some `src/header/focusable-header.tsx` and see how much code is possible to reuse between that and `src/header/internal.tsx`.

just-boris: +1

This header will never have actions or counter prop. Can be simplified to `<Box variant="h2">` with some additions

src/header/styles.scss
pan-kot: I would imagine the style was there for a reason. These styles also have different min/max widths. You can run the screenshot tests to see if there are some regressions. Ideally is to test the before and after with some large contents and see if all good.

src/wizard/wizard-form.tsx
```diff
@@ -43,6 +44,14 @@ export default function WizardForm({
   const isLastStep = activeStepIndex >= steps.length - 1;
   const skipToTargetIndex = findSkipToTargetIndex(steps, activeStepIndex);
   const isMobile = useMobile();
+  const stepHeaderRef = useRef<HTMLDivElement | null>(null);
+
+  useEffectOnUpdate(() => {
+    if (stepHeaderRef && stepHeaderRef.current) {
+      const header = stepHeaderRef.current.querySelector('[tabindex = "-1"]') as HTMLElement;
```
just-boris, line 51: Can we avoid `querySelector` and expose the correct element to focus right away?

src/header/styles.scss
```diff
@@ -171,6 +172,10 @@
     font-weight: awsui.$font-heading-m-weight;
     line-height: awsui.$font-heading-m-line-height;
   }
+
+  @include focus-visible.when-visible {
+    @include styles.link-focus;
```
pan-kot, line 177: Why is it link-focus and not, say, container focus? I think this one might be better applicable but I don't like the name as it sounds off. Should we maybe introduce a separate one like `header-focus` to ensure those are properly de-coupled?

src/wizard/__integ__/wizard.test.ts
```diff
@@ -63,6 +63,15 @@ describe('Wizard keyboard navigation', () => {
         await expect(page.getText('#content-text')).resolves.toBe('Content 1');
       })
     );
+
+    test(
+      'should focus on header after navigation',
+      setupTest(async page => {
+        await page.resetFocus();
+        await page.keys(['Tab', 'Tab', 'Space']);
```
pan-kot, line 71: I would add separate tests for Next and Prev buttons as both focus the header

src/wizard/wizard-form.tsx
```diff
@@ -43,6 +44,14 @@ export default function WizardForm({
...
+      header?.focus();
+    }
+  }, [activeStepIndex]);
```
pan-kot, line 54: With that the header will also be focused when the navigation happens programmatically. That might not be safe.

Imagine the page loads and the active step index is determined after some delay - the navigation in that case might or might not be expected. We might want to let customers control that by offering a function like `wizardRef.current.focusStepHeader()`.

pan-kot (changes requested):
I agree with the overall direction but there are some comments on the code still.

Btw, love the fact we can remove scroll to top!

src/header/styles.scss
Al-Dani: I discussed that with Kashi. Basically, the idea is that we want similar focus behaviour for all text-styled components.

pan-kot: ok, sounds good

fa7ad (approved):

avinashbot (changes requested):

avinashbot (approved):

src/wizard/index.tsx
```diff
@@ -61,17 +48,18 @@ export default function Wizard({
...
 
+  const stepHeaderRef = useRef<HTMLDivElement | null>(null);
+  const delegateFocusToHeader = () => {
+    stepHeaderRef.current?.focus();
+  };
+
   const navigationEvent = (requestedStepIndex: number, reason: WizardProps.NavigationReason) => {
     setActiveStepIndex(requestedStepIndex);
     fireNonCancelableEvent(onNavigate, { requestedStepIndex, reason });
+    delegateFocusToHeader();
```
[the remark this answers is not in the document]
Al-Dani: we will follow up on that if needed

PR #445, opened 2022-11-03
feat: Improve tabbing behavior for select, multiselect, and autosuggest with expandToViewport enabled

avinashbot:
### Description

Hey, look at me. Take a deep breath. We're going to get through this. I'll be overexplaining this but it's not a particularly complicated PR.

So, what's the problem? Basically, when you have `expandToViewport` set to `true`, the dropdown is positioned at the end of the DOM. With select and multiselect, the focus gets moved into the options list inside the dropdown. Then, you hit `Tab`, and you're sent off the page instead of to the element after the select trigger. Not good UX.

```
<!-- We want this! -->
<select>
  <trigger /> [1. start here]
  <dropdown>
    <options-list /> [2. focus automatically moved here]
  <dropdown />
</select>

<some-input /> [3. hitting tab gets you here]
```

```
<!-- But this is what happens -->
<select>
  <trigger /> [1. start here]
</select>

<some-input /> [you skipped me :(]

<portal>
  <dropdown>
    <options-list /> [2. focus automatically moved here]
  <dropdown />
</portal>

</body> [3. hitting tab gets you here, i guess???]
```

Okay, so why not do some browser magic to figure out which element is next after the tab order, detect when we tab out of the dropdown, and then move the focus into the next element transparently? Sounds like fun, but it's strewn with edge cases and potential focus traps (but if you want to try, you're welcome to pick up from where I gave up in #201).

There is an alternate solution. With the autosuggest, the focus stays with the input, but when there's an async error, a `Tab` sends you to the "Retry" link (`recoveryText`). The next `Tab` sends you back to the input in a kind of infinite `Tab` loop, but one that can be broken by just dismissing the suggestions (hitting `Esc`). We want this for select and multiselect too (except when we tab back, the dropdown closes). So let's move this behavior into the shared dropdown component.

```
<!-- And this is the fix! -->
<select>
  <trigger /> [1. start here] [3. tabbing forward gets you here again, but closes the dropdown]
</select>

<some-input /> [4. yay!]

<portal>
  <dropdown>
    <options-list /> [2. focus automatically moved here]
  <dropdown />
</portal>
```

### Changes

One: Add a new property to the dropdown called `loopFocus`, which sends the focus back to the trigger when attempting to tab past the dropdown content. It's enabled by default when `expandToViewport` is enabled.

Two: Let's add focus/blur handlers too to simplify upstream implementations. Check out how much manual focus-tracking complexity that removed in `use-select.ts` and `autosuggest-input/index.tsx`! We shipped a completely new tabbing behavior for select and multiselect, with almost no diff in line count.

### How has this been tested?

A whole day of fighting with integration tests, basically! The dropdown loop flag itself doesn't have any new tests, but it's tested through select, multiselect, and autosuggest. I'd also encourage playing around with the behavior yourself manually.

### Documentation changes

- [x] _No._

### Related Links

<details>

#### Correctness

#### Security

#### Testing

</details>

### DIFF
--- a/src/internal/components/autosuggest-input/index.tsx
+++ b/src/internal/components/autosuggest-input/index.tsx
@@ -16,7 +16,6 @@ import {
   InputKeyEvents,
   InputProps,
 } from '../../../input/interfaces';
-import { getFocusables } from '../focus-lock/utils';
 import { ExpandToViewport } from '../dropdown/interfaces';
 import { InternalBaseComponentProps } from '../../hooks/use-base-component';
 import { KeyCode } from '../../keycode';
@@ -39,6 +38,7 @@ export interface AutosuggestInputProps
   dropdownContent?: React.ReactNode;
   dropdownFooter?: React.ReactNode;
   dropdownWidth?: number;
+  loopFocus?: boolean;
   onCloseDropdown?: NonCancelableEventHandler<null>;
   onDelayedInput?: NonCancelableEventHandler<BaseChangeDetail>;
   onPressArrowDown?: () => void;
@@ -82,6 +82,7 @@ const AutosuggestInput = React.forwardRef(
       dropdownContent = null,
       dropdownFooter = null,
       dropdownWidth,
+      loopFocus,
       onCloseDropdown,
       onDelayedInput,
       onPressArrowDown,
@@ -124,14 +125,7 @@ const AutosuggestInput = React.forwardRef(
       close: closeDropdown,
     }));
 
-    const handleBlur: React.FocusEventHandler = event => {
-      if (
-        event.currentTarget.contains(event.relatedTarget) ||
-        dropdownContentRef.current?.contains(event.relatedTarget) ||
-        dropdownFooterRef.current?.contains(event.relatedTarget)
-      ) {
-        return;
-      }
+    const handleBlur = () => {
       if (!preventCloseOnBlurRef.current) {
         closeDropdown();
         fireNonCancelableEvent(onBlur, null);
@@ -225,17 +219,6 @@ const AutosuggestInput = React.forwardRef(
       'aria-activedescendant': ariaActivedescendant,
     };
 
-    const [trapDropdownFocus, setTrapDropdownFocus] = useState(false);
-
-    // Run this effect on every render to determine if necessary to trap focus around input and dropdown.
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-    useEffect(() => {
-      setTrapDropdownFocus(
-        (dropdownFooterRef.current ? getFocusables(dropdownFooterRef.current).length > 0 : false) ||
-          (dropdownContentRef.current ? getFocusables(dropdownContentRef.current).length > 0 : false)
-      );
-    });
-
     // Closes dropdown when outside click is detected.
     // Similar to the internal dropdown implementation but includes the target as well.
     useEffect(() => {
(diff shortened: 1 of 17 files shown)


src/internal/components/dropdown/index.tsx
```diff
@@ -121,11 +121,15 @@ const Dropdown = ({
...
   scrollable = true,
-  trapFocus = false,
+  loopFocus = expandToViewport,
+  onFocus,
+  onBlur,
   contentKey,
 }: DropdownProps) => {
-  const triggerRef = useRef<HTMLDivElement>(null);
-  const dropdownRef = useRef<HTMLDivElement>(null);
+  const wrapperRef = useRef<HTMLDivElement | null>(null);
```
avinashbot, line 129: We need two more refs that also cover the tab traps, because we don't want to call focus/blur if the tab traps (below) are focused into.

src/internal/components/autosuggest-input/index.tsx
```diff
@@ -225,17 +219,6 @@ const AutosuggestInput = React.forwardRef(
       'aria-activedescendant': ariaActivedescendant,
     };
 
-    const [trapDropdownFocus, setTrapDropdownFocus] = useState(false);
```
avinashbot, line 228: We can avoid having to do focus tracking here. If there's a recovery link, it has a tab element. If there's no recovery link, it doesn't.

src/internal/components/dropdown/index.tsx
```diff
@@ -217,6 +221,26 @@ const Dropdown = ({
     }
   };
 
+  const focusHandler = (event: React.FocusEvent) => {
+    const previouslyFocusedElement = event.relatedTarget;
+    if (
+      (!wrapperRef.current || !wrapperRef.current.contains(previouslyFocusedElement)) &&
+      (!dropdownContainerRef.current || !dropdownContainerRef.current.contains(previouslyFocusedElement))
```
abdhalees, line 228: could this condition be extracted to a function with a name that describes what it checks for?

```suggestion
      const isFocusTrapped = (wrapper, dropdown, focusedElement) => 
       (!wrapper.current || !wrapper.current.contains(focusedElement)) &&
      (!dropdown.current || !dropdown.current.contains(focusedElement))
```

src/internal/components/dropdown/interfaces.ts
```diff
@@ -117,10 +117,21 @@ export interface DropdownProps extends ExpandToViewport {
    * Whether the dropdown will have a scrollbar or not
    */
   scrollable?: boolean;
+
   /**
    * Whether the dropdown will have a focus trap including trigger, header, content and footer.
    */
-  trapFocus?: boolean;
+  loopFocus?: boolean;
```
abdhalees, line 124: what's the reason behind renaming? isn't a trap focus the same as a loop focus?

src/internal/components/dropdown/index.tsx
avinashbot: Sure! Updated.

src/internal/components/dropdown/interfaces.ts
avinashbot: I wanted to make it clear. All the dropdown does is send focus back to the trigger. In select and multiselect, this also closes the dropdown. So this doesn't really form a tab trap in the classical sense (like the one date picker has, for example).

abdhalees (dismissed):

abdhalees (approved):

src/select/utils/use-select.ts
```diff
@@ -14,9 +15,8 @@ import { OptionsListProps } from '../../internal/components/options-list';
 import { FilterProps } from '../parts/filter';
 import { ItemProps } from '../parts/item';
 import { usePrevious } from '../../internal/hooks/use-previous';
-import { BaseKeyDetail } from '../../internal/events';
-import { CancelableEventHandler, fireCancelableEvent, NonCancelableEventHandler } from '../../internal/events/index';
-import { containsOrEqual } from '../../internal/utils/dom';
+import { BaseKeyDetail, NonCancelableEventHandler } from '../../internal/events';
+import { CancelableEventHandler, fireNonCancelableEvent } from '../../internal/events/index';
```
Al-Dani, line 19: I wonder why do we have imports from `internal/events/index` separately? Everything could be imported from  `internal/events`

avinashbot: \*shakes fist as vscode\*

Al-Dani (approved):

PR #447, opened 2022-11-04
Table column resizer keyboard access

YueyingLu:
### Description

AWSUI-18932
1. Make table column resizer keyboard accessible
3. Use role `separator` and aria attributes https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/separator_role#focusable_separator

### How has this been tested?

### Documentation changes

- [ ] _No._

### Related Links

<details>

#### Correctness

#### Security

#### Testing

</details>

### DIFF
--- a/pages/table/resizable-columns.page.tsx
+++ b/pages/table/resizable-columns.page.tsx
@@ -28,7 +28,6 @@ const columnsConfig: TableProps.ColumnDefinition<Item>[] = [
     header: 'Name',
     cell: item => <Link href={`#${item.id}`}>{item.text}</Link>,
     width: 200,
-    resizerAriaLabel: 'Resize name column',
   },
   {
     id: 'region',
@@ -37,27 +36,23 @@ const columnsConfig: TableProps.ColumnDefinition<Item>[] = [
     minWidth: 130,
     width: 130,
     sortingField: 'region',
-    resizerAriaLabel: 'Resize region ',
   },
   {
     id: 'description',
     header: 'Description',
     minWidth: 100,
     cell: item => item.description,
-    resizerAriaLabel: 'Resize description column',
   },
   {
     id: 'state',
     header: 'State',
     maxWidth: 150,
     cell: item => item.state,
-    resizerAriaLabel: 'Resize state column',
   },
   {
     id: 'extra',
     header: 'Extra column',
     cell: () => '-',
-    resizerAriaLabel: 'Resize extra column',
   },
 ];
 
@@ -173,6 +168,9 @@ export default function App() {
           sortingDescending={sorting?.isDescending}
           onSortingChange={event => setSorting(event.detail)}
           onColumnWidthsChange={handleWidthChange}
+          ariaLabels={{
+            resizerAriaLabel: 'Resize',
+          }}
         />
       </ScreenshotArea>
     </SpaceBetween>
--- a/pages/table/resizable-columns.page.tsx
+++ b/pages/table/resizable-columns.page.tsx
@@ -28,6 +28,7 @@ const columnsConfig: TableProps.ColumnDefinition<Item>[] = [
     header: 'Name',
     cell: item => <Link href={`#${item.id}`}>{item.text}</Link>,
     width: 200,
+    resizerAriaLabel: 'Resize name column',
   },
   {
     id: 'region',
@@ -36,23 +37,27 @@ const columnsConfig: TableProps.ColumnDefinition<Item>[] = [
     minWidth: 130,
     width: 130,
     sortingField: 'region',
+    resizerAriaLabel: 'Resize region ',
   },
   {
     id: 'description',
     header: 'Description',
     minWidth: 100,
     cell: item => item.description,
+    resizerAriaLabel: 'Resize description column',
   },
   {
     id: 'state',
     header: 'State',
     maxWidth: 150,
     cell: item => item.state,
+    resizerAriaLabel: 'Resize state column',
   },
   {
     id: 'extra',
     header: 'Extra column',
     cell: () => '-',
+    resizerAriaLabel: 'Resize extra column',
   },
 ];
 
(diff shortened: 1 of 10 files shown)


pages/table/resizable-columns.page.tsx
```diff
@@ -168,6 +168,9 @@ export default function App() {
           sortingDescending={sorting?.isDescending}
           onSortingChange={event => setSorting(event.detail)}
           onColumnWidthsChange={handleWidthChange}
+          ariaLabels={{
+            resizerAriaLabel: 'Resize',
```
gethinwebster, line 172: have we confirmed with the a11y team that this is an acceptable approach? (as opposed to having individual labels for each column)

src/table/resizer/index.tsx
```diff
@@ -42,10 +49,12 @@ export function Resizer({ onDragMove, onFinish }: ResizerProps) {
...
 
     const updateColumnWidth = (newWidth: number) => {
+      const { right, width } = headerCell.getBoundingClientRect();
+      const updatedWidth = newWidth < minWidth ? minWidth : newWidth;
+      updateTrackerPosition(right + updatedWidth - width);
+      setHeaderCellWidth(newWidth);
       // callbacks must be the last calls in the handler, because they may cause an extra update
       onDragStable(newWidth);
-      // we read the element size again because the previous callback changes it
-      updateTrackerPosition(headerCell.getBoundingClientRect().right);
```
YueyingLu, line 48: onDragStable update the column width with useState. At this moment the width can be not yet updated. It should be updated by known width value change, which is more reliable.

src/table/resizer/index.tsx
```diff
@@ -42,10 +49,12 @@ export function Resizer({ onDragMove, onFinish }: ResizerProps) {
     };
 
     const updateColumnWidth = (newWidth: number) => {
+      const { right, width } = headerCell.getBoundingClientRect();
+      const updatedWidth = newWidth < minWidth ? minWidth : newWidth;
```
YueyingLu, line 53: We don't take maxWidth into consideration, because
> Note that when the resizableColumns property is set to true this property is ignored.

pages/table/resizable-columns.page.tsx
YueyingLu: We are using ariaLabeledby, referring to this ariaLabel and header. So that they should be unique. Does it make sense?

src/table/resizer/index.tsx
```diff
@@ -79,20 +87,43 @@ export function Resizer({ onDragMove, onFinish }: ResizerProps) {
...
-    document.addEventListener('mousemove', onMouseMove);
-    document.addEventListener('mouseup', onMouseUp);
+    document.body.classList.remove(styles['resize-active-with-focus']);
+    if (isDragging) {
+      document.addEventListener('mousemove', onMouseMove);
+      document.addEventListener('mouseup', onMouseUp);
+    }
+    if (resizerHasFocus) {
+      document.body.classList.add(styles['resize-active-with-focus']);
+      headerCell.addEventListener('keydown', onKeyDown);
```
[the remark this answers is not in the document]
YueyingLu: It seems keypress is deprecated https://developer.mozilla.org/en-US/docs/Web/API/Element/keypress_event. Tested with keydown on Safari FF and Chrome, it also repeats when holding down.

pages/table/resizable-columns.page.tsx
gethinwebster: Ah, I hadn't noticed that. I think that will be problematic for languages other than English, because it implies an ordering of the two "parts" of the overall label. For example in German you'd expect the word order to be opposite (verb comes at end of the sentence)

src/table/resizer/index.tsx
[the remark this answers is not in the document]
gethinwebster: ah interesting, sounds good then!

pages/table/resizable-columns.page.tsx
YueyingLu: I think the resizerAriaLabel not necessary being verb in this case, it can be like "resize handler". They are two descriptive parts.

gethinwebster: I think we need to start from what the desired announcement is, and then work backwards to how best to technically achieve this. What should the user hear? What will we add to our "Writing guidelines" (or Accessibility guidelines, I forget where it usually goes) for this new property?

avinashbot (changes requested):
**Looks good! Unfortunately, there's _one_ change I think we can't ship without.** (see bold section below)

Manual testing:

- Works great with NVDA on focus mode. Given that the expected controls for splitters are the arrow keys, I assume NVDA users are well aware that they need to switch out of browse mode to use it (cause arrow keys control browse mode navigation).
- VoiceOver is a different story. It doesn't seem to announce changes in Safari. But it does on Chrome, so I assume this is a screen reader bug, and I don't think we should add an aria-live region to the possible detriment of other assistive tech.
- VoiceOver also looks like it's providing the current width as percentages, rather than raw values. So when aria-valuemax isn't provided, it defaults to 100, and declares every column as "collapsed" (because it's out of range, I guess?). **Fix: provide a value aria-valuemax** (my suggestion is providing the table width, maybe; cause at least the percentages serve _some_ use).

YueyingLu:
Hi @avinashbot 
VO+safari doesnt announce aria-label/aria-labelledby. It only happens on safari, I agree it is on Safari or VO side.
We dont have max width of columns, according to [WAI spec](https://w3c.github.io/aria/#aria-valuemax) and also from @jorycunningham we dont need to add aria-valuemax if the value not exist. Providing a wrong value to fix VO, might make it confusing for other screenreaders, which are more commonly used.

avinashbot (approved):
`aria-valuetext` fixes the Safari issue, as well as making it announce changes. Great fix!

PR #450, opened 2022-11-04
Remove the AppLayout contentType dependency for stickyHeader behavior.

DaemonCahill:
Currently the `AppLayout` component is using the `contentType` property to determine whether or not a child component is implementing `stickyHeader` behavior. The assumption was that the `cards` and `table` content types would always be accompanied by the use of a sticky header. This was not the case in classic thus creating a feature dependency in visual refresh that is considered a regression.

This PR adds a new property to the visual refresh `AppLayout` component called `hasStickyBackground`. A child component will communicate up via the `AppLayout` context to determine whether or not a sticky header is implemented independent of the `contentType` property.

### DIFF
--- a/src/app-layout/visual-refresh/app-bar.scss
+++ b/src/app-layout/visual-refresh/app-bar.scss
@@ -89,13 +89,22 @@ section.appbar {
     }
 
     /*
-    If the content type supports a dark sticky header then 
-    minimal bottom padding should be on the breadcrumbs to
+    Table and cards content types implement a custom dark header. 
+    Minimal bottom padding should be on the breadcrumbs to
     reduce the the vertical distance from the main content.
     */
     > .breadcrumbs.content-type-cards,
     > .breadcrumbs.content-type-table {
       padding-bottom: awsui.$space-xxs;
+    }
+
+    /*
+    If a child component tells the AppLayout to have a sticky 
+    background (typically because of a stickyHeader with a 
+    full-page variant) the z-index of the breadcrumbs needs to 
+    be lifted so it doesn't render behind the background element.
+    */
+    > .breadcrumbs.has-sticky-background {
       z-index: 799;
     }
   }
--- a/src/container/internal.tsx
+++ b/src/container/internal.tsx
@@ -50,7 +50,7 @@ export default function InternalContainer({
   const rootRef = useRef<HTMLDivElement>(null);
   const headerRef = useRef<HTMLDivElement>(null);
   const { isSticky, isStuck, stickyStyles } = useStickyHeader(rootRef, headerRef, __stickyHeader, __stickyOffset);
-  const { setHasStickyOverlap } = useContext(AppLayoutContext);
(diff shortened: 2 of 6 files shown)


src/app-layout/visual-refresh/app-bar.scss
```diff
@@ -89,13 +89,22 @@ section.appbar {
     }
 
     /*
-    If the content type supports a dark sticky header then 
-    minimal bottom padding should be on the breadcrumbs to
+    Table and cards content types implement a custom dark header.
```
jperals, line 92: The header is actually not dark for cards nor for table; this must have been true before Visual Refresh where this element was light gray, but now it is white like the rest of the content for both components.

src/container/internal.tsx
```diff
@@ -49,15 +50,34 @@ export default function InternalContainer({
...
   const headerIdProp = __headerId ? { id: __headerId } : {};
 
+  /**
+   * The visual refresh AppLayout component needs to know if a child component
+   * has a high constrast sticky header. This is to make sure the background element
+   * stays in the same vertical position as the header content.
+   */
+  useLayoutEffect(
+    function handleHasStickyBackground() {
+      if (isRefresh && isSticky && variant === 'full-page') {
```
jperals, line 70: Is there a chance that something changes so that we want to change `hasStickyBackground` back from `true` to `false`? For example, an element with sticky header being no longer present.

abdhalees (approved):
LGTM

src/app-layout/visual-refresh/app-bar.scss
DaemonCahill: > ...since the header is actually not dark for cards nor for table...

The content type property is coupled with the full page variants of both components which have a persistent dark header. Unless I am missing something? If so could you elaborate...

src/container/internal.tsx
DaemonCahill: Yes! If the internal container un-mounts due to a route change (or some other reason) that should be handled by the `cleanup` function which resets `hasStickyBackground` back to `false`.

src/app-layout/visual-refresh/app-bar.scss
DaemonCahill: Here is the full page table demo: https://cloudscape.design/examples/react/table.html
And the full page cards demo: https://cloudscape.design/examples/react/cards.html

src/container/internal.tsx
jperals: Oh I see, this `InternalContainer` is a child of both `Table` and `Card`, so every time the Table or Card unmounts, this will also unmount, do I understand correctly? (I thought it was sort of the other way around, this component being a sort of global container which would eventually contain Tables or Cards)

src/app-layout/visual-refresh/app-bar.scss
jperals: Oh OK, it's about the full page versions, this is what I was missing.

PR #454, opened 2022-11-08
feat: Improve screen reader behavior for flashbar

avinashbot:
### Documentation changes:

>  **ariaRole** (boolean) - For flash messages added after page load, specifies how this message is communicated to assistive technology. Use "status" for status updates or informational content. Use "alert" for important messages that need the user's attention.

### Description

Improved focus behavior for newly added flashbars. This is done by either (a) announcing flash messages using aria-live, or (b) moving focus to important flash messages. This is controlled by a new property, `ariaRole`, which is provided to flash messages added after page load.

Flash messages are also now inside an unordered list for a more semantic structure. There is also throttling for focus moves, to prevent abrupt jumps when multiple messages are added sequentially. For that, I stole the throttle code from area chart and moved it into the shared utils.

### How has this been tested?

New unit tests and integration tests. Also manually tested with testable screen readers.

Oh yeah, there's also one failing permutation, but I think that's fine. If anything, the current state feels more right.

<details>
<summary>Open screenshot</summary>
image
</details>

### Documentation changes

- _Yes, this change contains documentation changes._
   - It introduces one new property, `ariaRole`.

### Related Links

AWSUI-19116

<details>

#### Correctness

#### Security

#### Testing

</details>


src/flashbar/flash.tsx
```diff
@@ -46,14 +51,24 @@ function dismissButton(
   );
 }
 
-interface Flash extends FlashbarProps.MessageDefinition {
+export const focusFlashById = throttle(
```
fa7ad, line 54: Is there a unit test for this function?

src/flashbar/index.tsx
```diff
@@ -17,32 +17,59 @@ import useFocusVisible from '../internal/hooks/focus-visible';
...
+  // Track new or removed item IDs in state to only trigger focus changes for newly added items.
+  // https://reactjs.org/docs/hooks-faq.html#how-do-i-implement-getderivedstatefromprops
+  if (items) {
+    const newItems = items.filter(({ id }) => id && !previousItems.some(item => item.id === id));
+    const removedItems = previousItems.filter(({ id }) => id && !items.some(item => item.id === id));
+    if (newItems.length > 0 || removedItems.length > 0) {
+      setPreviousItems(items);
+      const newFocusItems = newItems.filter(({ ariaRole }) => ariaRole === 'alert');
+      if (newFocusItems.length > 0) {
+        setNextFocusId(newFocusItems[0].id!);
```
fa7ad, line 48: What do you think about adding a test to check if the focus order (next/prev) is working as expected?

src/flashbar/flash.tsx
avinashbot: No, but this behavior is integration tested.

src/flashbar/index.tsx
avinashbot: All focus moves are tested only in integration tests. Makes sense to add another test to make sure focus doesn't move back to previously focused elements, and I'll add that. But I think Codecov only does unit test coverage, so it's always going to complain about these lines.

Focus moves, especially when combined with React rerenders are extremely flaky with JSDOM, for some reason.

fa7ad (approved):

src/flashbar/flash.tsx
```diff
@@ -94,36 +112,52 @@ export const Flash = React.forwardRef(
...
 
+    const announcement = [statusIconAriaLabel, header, content].filter(Boolean).join(' ');
+
     return (
+      // We're not using "polite" or "assertive" here, just turning default behavior off.
+      // eslint-disable-next-line @cloudscape-design/prefer-live-region
       <div
         ref={ref}
+        role={ariaRole}
+        aria-live={ariaRole ? 'off' : undefined}
```
YueyingLu, line 123: what would be the difference to set 'off' or 'undefined' ?

avinashbot: `role="status"` or `role="alert"` announce message contents by default. Here, we're just using it to provide context to assistive technology, so we need to turn off the default aria-live behavior.

YueyingLu (approved):

stirache (approved):

PR #461, opened 2022-11-10
fix: Prevent app layout from obscuring clicks when disableContentPaddings=true

avinashbot:
### Description

<details>
   <summary><strong>Demo recording</strong></summary>

https://user-images.githubusercontent.com/983897/200972832-efdfda31-85bf-4ea0-8972-24828b83e48b.mov

</details>

The refresh app layout has scaffolding that blocks mouse clicks from passing through it to the content below. Normally, that wouldn't be a problem, because the only thing under that area are content paddings. But... if you disable content paddings and want to have interactive controls from edge to edge, then you're going have issues.

To resolve those without tweaking around the app layout structure and breaking a million other things, I just set `pointer-events: none` on the wrapper elements, and reactivate pointer events on the child elements (i.e. the panels and toggle buttons).

### How has this been tested?

Check out that dev page! Check out those integration tests! So cool!

### Documentation changes

- [x] _No._

### Related Links

<details>

#### Correctness

  - Keyboard behavior is unchanged

#### Security

#### Testing

</details>

### DIFF
--- a/src/app-layout/visual-refresh/navigation.scss
+++ b/src/app-layout/visual-refresh/navigation.scss
@@ -17,6 +17,18 @@
   top: var(#{custom-props.$offsetTop});
   z-index: 850;
 
+  /*
+  The navigation and tools containers (that contain the toggle buttons)
+  stretch the full height of the app layout. Normally, this wouldn't be an
+  issue because they sit above the app layout's content padding.
+  
+  But if disableContentPaddings is set to true and there are buttons on the
+  left/right edges of the screen, they will be covered by the containers. So
+  we need to disable pointer events in the container and re-enable them in
+  the panels and toggle buttons.
+  */
+  pointer-events: none;
+
   @include styles.media-breakpoint-up(styles.$breakpoint-xx-large) {
     #{custom-props.$navigationWidth}: 320px;
   }
@@ -73,6 +85,7 @@ nav.navigation {
   overscroll-behavior-y: contain;
   position: relative;
   word-wrap: break-word;
+  pointer-events: auto;
 
   // Animation for the Navigation opacity and width when it is added to the DOM
   @keyframes openNavigation {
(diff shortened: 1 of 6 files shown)


src/app-layout/visual-refresh/navigation.scss
```diff
@@ -17,6 +17,18 @@
...
+  The navigation and tools containers (that contain the toggle buttons)
+  stretch the full height of the app layout. Normally, this wouldn't be an
+  issue because they sit above the app layout's content padding.
+  
+  But if disableContentPaddings is set to true and there are buttons on the
+  left/right edges of the screen, they will be covered by the containers. So
+  we need to disable pointer events in the container and re-enable them in
+  the panels and toggle buttons.
+  */
+  pointer-events: none;
```
DaemonCahill, line 30: Should we pull in `disableContentPaddings` to the navigation and tools components from the context? If so, we could bind the pointer events removal to the use of this property exclusively:

```
.navigation-container {
  &.disable-content-paddings {
    pointer-events: none;
  }
}

.tools-container {
  &.disable-content-paddings {
    pointer-events: none;
  }
}
```

It's not technically necessary, though it does contain the impact of the change to only these edge case circumstances.

DaemonCahill:
Looks great! Thank you (as always) for the lightning fast turn around. I added one comment about potentially constraining the scope of the change. It's up to you whether or not you think it is worth implementing.

src/app-layout/visual-refresh/navigation.scss
DaemonCahill: From @avinashbot transferring from Slack for the record:

> My worry is that this could mask any pointer-events issues we might introduce in the future. If we do it only for a small set of cases, it might be less likely that we detect it during integration or screenshot tests.

Agreed on this point. The broader the use the more likely we are to catch missing pointer events from future interactive elements in development.

pan-kot (approved):
LGTM

PR #495, opened 2022-11-17
fix: Focus resize handler when split panel is opened

avinashbot:
### Description

When the split panel is opened, we previously moved focus to the close button. But this isn't great because it's the third (and often last) item in the tab order, after the resizer and the preferences button. While it would be nice to keep this, making the close button the first thing in the tab order (without changing the visuals) is unintuitive, especially because it's the right-most item in the header area. So we just move focus to the resize handler, which seems like the most sensible thing for the time being. And we're probably going to revisit focus moves at some point.

Related links, issue #, if available: AWSUI-19409

### How has this been tested?

~~Too lazy to run all the tests. If a test yells at me, I'll update it. Or else, I'll write a test.~~ Updated unit and integration tests.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/__integ__/app-layout-focus-delegation.test.ts
+++ b/src/app-layout/__integ__/app-layout-focus-delegation.test.ts
@@ -26,13 +26,13 @@ function setupTest(
 [false, true].forEach(visualRefresh =>
   describe(`visualRefresh=${visualRefresh}`, () => {
     test(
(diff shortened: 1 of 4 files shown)


rubencarvalho (approved):
I think these changes make sense. LGTM!

PR #501, opened 2022-11-18
fix: Prevent resetting StickyBackground from elements using InternalContainer

johannes-weber:
### Description

Before this change, the sticky background got reset when mounting and unmounting another component which uses the `InternalContainer`. This resulted in a header background which did not had the full width after unmounting. See #505 for details.
 

Related links, issue #, if available: AWSUI-19824, CR-80122674

### How has this been tested?

- manually tested in all browsers we currently support.
- added additional screenshot tests for verification.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/container/internal.tsx
+++ b/src/container/internal.tsx
@@ -1,7 +1,7 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 import clsx from 'clsx';
-import React, { useContext, useLayoutEffect, useRef } from 'react';
+import React, { useContext, useEffect, useRef } from 'react';
 import { AppLayoutContext } from '../app-layout/visual-refresh/context';
 import { ContainerProps } from './interfaces';
 import { getBaseProps } from '../internal/base-component';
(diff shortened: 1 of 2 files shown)


src/container/internal.tsx
```diff
@@ -69,12 +69,15 @@ export default function InternalContainer({
    */
   useLayoutEffect(
```
just-boris, line 70: Can this issue be fixed by replacing `useLayoutEffect` with `useEffect`?

johannes-weber: Without having the condition it would behave the same.
I'm not 100% sure why `useLayoutEffect` have been used initially because .  I think changing that to `useEffect` does not impact the behaviour because we don't do any DOM readings or direct DOM manipulations inside the hook.

just-boris: Understood. More context: https://github.com/facebook/react/issues/21400

The issue still can be fixed by applying `useEffect` somewhere, but not this one.

In general, when you are using any effect, you are making a compromise causing expensive investigations and difficult bug fixes. Please read [this article](https://beta.reactjs.org/learn/you-might-not-need-an-effect) and avoid.

just-boris (approved):

Issue #483, opened 2022-11-16
[Bug]: TagEditor is stucking with "Loading key values" when using React.StrictMode


fr0stf0x:
### Browser

Chrome

### Package version

3.0.116

### React version

18.2.0

#### Actual behavior:
`<TagEditor />` is stucking with "Loading key values" text when using `<React.StrictMode />`

Turn off <StrictMode /> then it will be fine

#### Expected behavior:
`<TagEditor />` should load keys successfully when using `<React.StrictMode />`

### Source code

https://codesandbox.io/s/confident-frog-zxflx8

### Reproduction

https://codesandbox.io/s/confident-frog-zxflx8

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/cloudscape-design/components/blob/main/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/cloudscape-design/components/issues) for duplicate problems

taheramr:
Hello, 
Thanks for reaching out! We are working on resolving this issue and will provide updates on this issue.

fr0stf0x:
It seems like the bug is fixed in 3.0.127, isn't it @taheramr?

just-boris:
Yes, it is. Resolving

PR #502, opened 2022-11-18
fix: Fix tag editor key suggestions getting stuck in React 18's strict mode

avinashbot:
#### React 18's strict mode

So, here's what's happening. React 18 has a [new strict mode](https://reactjs.org/docs/strict-mode.html#ensuring-reusable-state). Essentially, the big change is that when new components are added, they're (1) rendered, (2) destroyed, and (3) recreated, and only then are they added to the DOM. As with other strict mode quirks, this is to ensure components are more resilient against future changes. You can read more about their reasoning in the article.

But importantly, refs aren't recreated when the component is recreated, because refs are outside React's state by definition. This means that when the component is first destroyed (2), our cleanup effect runs, which cancels the fetcher and doesn't let another one start.

#### The fix

But do we need to clean up the in-progress fetch function at all? We don't (but we needed to). Calling `setState` after a component already unmounted used to print a loud error in the console to encourage users to clean up any memory leaks inside effects. But it has since been removed. Why is that? See facebook/react#22114: 

> However, there is no actual problem [for fetch requests]. There's no "memory leak" here: the POST comes back, we try to set state, the component is unmounted (so nothing happens), and then we're done. Nothing keeps holding onto the component indefinitely. [...] So the warning against a problematic pattern ends up pushing people towards more problematic patterns though the original code was actually fine.

Sooo....

```diff
-    // Cancel any active promises on unmount.
-    useEffect(() => () => requestCancelFnRef.current.cancel(), []);
```

All done! All that explanation to delete two lines of code.

Related links, issue #, if available: #483

### How has this been tested?

"Ensures load function isn't cancelled on unmount" isn't a very testable concept. If you disagree, let's have a chat to see how else we can test this.

But do we need React 18 tests in general? Absolutely; this isn't our first React 18 only issue, and probably won't be our last. Probably my next little side improvement, but I don't think this issue needs to wait for those to happen. Open to change my mind on that though.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/tag-editor/internal.tsx
+++ b/src/tag-editor/internal.tsx
@@ -1,6 +1,6 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
-import React, { useEffect, useRef, useState } from 'react';
+import React, { useRef, useState } from 'react';
 
 import { AutosuggestProps } from '../autosuggest/interfaces';
 import InternalAutosuggest from '../autosuggest/internal';
@@ -65,12 +65,10 @@ export const TagControl = React.forwardRef(
       cancel: () => {},
       isCancelled: () => false,
     });
+
     const latestFilteringQuery = useRef<FilteringParams>({ key: undefined, value: undefined });
     const isSameQuery = (key: string | undefined, value: string) =>
       latestFilteringQuery.current.key === key && latestFilteringQuery.current.value === value;
-
-    // Cancel any active promises on unmount.
-    useEffect(() => () => requestCancelFnRef.current.cancel(), []);
 
     const onLoadItems = (filteringText: string) => {
       if (!onRequest || isSameQuery(filteringKey, filteringText) || requestCancelFnRef.current.isCancelled()) {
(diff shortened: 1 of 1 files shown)


src/tag-editor/internal.tsx
```diff
@@ -65,28 +65,26 @@ export const TagControl = React.forwardRef(
...
-    // Cancel any active promises on unmount.
-    useEffect(() => () => requestCancelFnRef.current.cancel(), []);
-
     const onLoadItems = (filteringText: string) => {
       if (!onRequest || isSameQuery(filteringKey, filteringText) || requestCancelFnRef.current.isCancelled()) {
         return;
       }
       requestCancelFnRef.current.cancel();
 
-      if (filteringText === '' && initialOptionsRef?.current && initialOptionsRef.current.length > 0) {
```
avinashbot, line 81: This is all refactoring. Moved a setter from multiple `if` statements to a single `if` statement. Felt easier to read for me.

fa7ad (approved):
LGTM! Good job on the inline comments, helped me understand what's going on 😄

PR #507, opened 2022-11-21
feat: Add steps information to annotation trigger aria label

YueyingLu:
### Description

Current aria-label on annotation trigger has only _open/close_ state, the aria-labels are "open/close annotation". However, one page can have multiple annotations with same accessible name, screenreader users cant distinguish them. The PR is to add steps information to the aria-label, being for example "open annotation Step 3/6".

Related links, issue #AWSUI-19245

### How has this been tested?

 Added unit test. 

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/annotation-context/annotation/annotation-trigger.tsx
+++ b/src/annotation-context/annotation/annotation-trigger.tsx
@@ -5,6 +5,7 @@ import styles from './styles.css.js';
 import { AnnotationIcon } from './annotation-icon';
 import useFocusVisible from '../../internal/hooks/focus-visible/index.js';
 import { AnnotationContextProps } from '../interfaces';
+import { joinStrings } from '../../internal/utils/strings/join-strings.js';
 
 export interface AnnotationTriggerProps {
   open: boolean;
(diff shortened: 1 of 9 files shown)


src/annotation-context/annotation/annotation-trigger.tsx
```diff
@@ -33,7 +36,10 @@ export default React.forwardRef<HTMLButtonElement, AnnotationTriggerProps>(funct
       ref={ref}
       className={styles.hotspot}
       aria-haspopup="dialog"
-      aria-label={i18nStrings.labelHotspot(open)}
+      aria-label={joinStrings(
```
avinashbot, line 39: Joining two aria labels with a space feels a little... English-centric. Would be ideal if we could just make this one internationalizable aria-label.

Maybe we could provide this extra information as additional parameters for the existing hotspot label function so developers can generate the whole label at once in a more flexible way? It should be backwards compatible ([see playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBA6glgEwgJwMIEMA2mBG6DGA1lALxQAU6AXFAM7DJwB2A5gDRQ432Osf7cGzFgEpSAPigA3APaIA3AFgAUPhlN6UJumTIZAdwzY8RGvCRosuAsTKVBvURKgBvAL5LlQA)), but with TypeScript, who knows...

What do you think?

YueyingLu: I think it makes sense, especially "open annotation" is a verb, maybe weird in some languages. Updated in the new revision.

avinashbot (approved):

PR #509, opened 2022-11-21
chore: Add analytics beta hook

michaeldowseza:
### Description

Adds an internal hook that will check for the presence of a beta data attribute to enable auto tagging of the Wizard component for better analytics integration.

### How has this been tested?

- Added a simple unit test for the hook
- Manual validation of added tags

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/hooks/use-analytics/index.ts
+++ b/src/internal/hooks/use-analytics/index.ts
@@ -1,0 +1,17 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// We expect analytics to be set only once and before the application is rendered.
+let analyticsEnabled: boolean | undefined = undefined;
+
+export function useAnalytics() {
+  if (analyticsEnabled === undefined) {
+    if (typeof window === 'undefined' || typeof document === 'undefined') {
+      analyticsEnabled = false;
+    } else {
+      analyticsEnabled = !!document.querySelector('.awsui-analytics-enabled');
+    }
+  }
+
+  return analyticsEnabled;
+}
--- a/src/wizard/wizard-form.tsx
+++ b/src/wizard/wizard-form.tsx
@@ -11,6 +11,7 @@ import WizardFormHeader from './wizard-form-header';
 import styles from './styles.css.js';
 import useFocusVisible from '../internal/hooks/focus-visible';
 import { useEffectOnUpdate } from '../internal/hooks/use-effect-on-update';
+import { useAnalytics } from '../internal/hooks/use-analytics';
 
 interface WizardFormProps {
   steps: ReadonlyArray<WizardProps.Step>;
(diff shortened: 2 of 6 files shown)


src/internal/hooks/use-analytics/index.ts
```diff
@@ -0,0 +1,17 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { useEffect } from 'react';
+
+// We expect analytics is to be set only once and before the application is rendered.
+let analyticsEnabled: boolean | undefined = undefined;
+
+export function useAnalytics() {
```
just-boris, line 9: This is different than `useVisualRefresh` implementation: https://github.com/cloudscape-design/components/blob/main/src/internal/hooks/use-visual-mode/index.ts#L36

Should these two be unified to have the same activation requirements

michaeldowseza: Comment can actually be removed. It started off similar but there's no expectation that this attribute should appear and apply to the entire page.

michaeldowseza: Adapted based on feedback. Kept the comment as it now operates in a similar fashion as the `useVisualRefresh` hook.

pages/app/index.tsx
```diff
@@ -65,6 +65,15 @@ function App() {
     }
   }, [visualRefresh]);
 
+  useEffect(() => {
+    // NB: do not use classList.toggle, it does not work as expected in IE
```
just-boris, line 69: IE is gone, this is not needed

pages/app/app-context.tsx
```diff
@@ -28,6 +29,7 @@ const appContextDefaults: AppContextType = {
     density: Density.Comfortable,
     visualRefresh: THEME === 'default',
     motionDisabled: false,
+    analyticsEnabled: false,
```
just-boris, line 32: Do we really need this conditionality? Can we keep it always on in our development (bleeding edge mode on)?

src/internal/hooks/use-analytics/index.ts
```diff
@@ -0,0 +1,17 @@
...
+
+// We expect analytics to be set only once and before the application is rendered.
+let analyticsEnabled: boolean | undefined = undefined;
+
+export function useAnalytics() {
+  if (analyticsEnabled === undefined) {
+    if (typeof window === 'undefined' || typeof document === 'undefined') {
+      analyticsEnabled = false;
+    } else {
+      analyticsEnabled = !!document.querySelector('.awsui-analytics');
```
just-boris, line 12: Can we change to something more explicit, for example `awsui-analytics-enabled` or `awsui-use-analytics`

src/wizard/wizard-form.tsx
```diff
@@ -13,6 +13,7 @@ import useFocusVisible from '../internal/hooks/focus-visible';
 import { useEffectOnUpdate } from '../internal/hooks/use-effect-on-update';
 
 interface WizardFormProps {
+  analyticsEnabled: boolean;
```
just-boris, line 16: Since our hook is computed once and reuses cached value, we can call it as many times as we want.

Should we call it in every component instead of polluting the API with passing this prop down

just-boris (approved):

src/internal/hooks/use-analytics/index.ts
michaeldowseza: `awsui-analytics-enabled` seems fine.

src/wizard/wizard-form.tsx
michaeldowseza: Makes sense 👍🏻

just-boris (approved):

PR #517, opened 2022-11-22
fix: Add aria live announcer to code editor for errors and warnings 

jorycunningham:
### Description

This PR adds an aria-live region to the code editor so that updates to the error and warning counts are announced to assistive technology.

The live region receives the same text that is being used to label the error and warning buttons.

Note: aria-live="assertive" has been used to ensure that the "polite" announcement of the characters typed does not cancel out the warning and to allow assistive technology to treat the announcement with importance

Resolves AWSUI-19219

### How has this been tested?

Verified functionality in NVDA, JAWS, and VoiceOver

To test:
* use the Code Editor "simple" page and cause an error to fire (for example add a "!" to the end of a line).
* observe output from screen readers when the error fires.

<details>

#### Correctness

#### Security

#### Testing

YES
</details>

---


avinashbot (commented):
I'll defer to you here, but doesn't `aria-live="assertive"` feel a little _too_ assertive? Ace tries to check errors on every keystroke and will yell about errors even as you're in the middle of writing a single line.

The noise of the error status turning red and grey back and forth is easy to ignore visually, but not easy to ignore if aria-live is interrupting in the middle of every keystroke. `aria-live="polite"` should wait a short moment _after_ the user is presumably done typing before announcing errors or warnings, so it feels like it makes more sense, right?

We can discuss this later today :)

avinashbot (approved):
Had a chat yesterday and tried it out myself; aria-live="assertive" is the way to go, especially since there since to be other live announcements in the ace editor fighting each other.

PR #521, opened 2022-11-23
feat: Add live region to Table to announce page changes

abdhalees:
### Description

- Add `totalItemsCount` property and use it in `aria-rowcount`
- Add `firstIndex` property and use it to calculate `aria-rowindex`
- Add `liveAnnouncment` function as a live region to announce to screen readers page changes

### How has this been tested?

unit test and with voice over and nvda

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/table/interfaces.tsx
```diff
@@ -228,6 +228,25 @@ export interface TableProps<T = any> extends BaseComponentProps {
...
+   */
+  totalItemsCount?: number;
+  /**
+   * Index of the first item in the table, necessary for screen readers to be aware
+   * of which range of items is currently displayed in the table,
+   * if there is no pagination in the table leave this property undefined.
+   */
+  firstIndex?: number;
+  /**
+   * Returns a text that is only read by screen readers,
```
stirache, line 245:
```suggestion
Use this function to announce page changes for screen reader users. It returns a text when the firstIndex or totalItemsCount change, but only when firstIndex is available. 
```

src/table/interfaces.tsx
```diff
@@ -228,6 +228,25 @@ export interface TableProps<T = any> extends BaseComponentProps {
...
+   * of how many items there are in a table, if the total items in the table is unknown
+   * leave this property undefined.
+   */
+  totalItemsCount?: number;
+  /**
+   * Index of the first item in the table, necessary for screen readers to be aware
+   * of which range of items is currently displayed in the table,
+   * if there is no pagination in the table leave this property undefined.
+   */
+  firstIndex?: number;
```
stirache, line 243:
```suggestion
Use this function to inform screen readers which range of items is currently displayed in the table. It returns an index of the first item in the table. If the table has no pagination, leave this property undefined.
```

src/table/interfaces.tsx
```diff
@@ -228,6 +228,25 @@ export interface TableProps<T = any> extends BaseComponentProps {
    * @visualrefresh `embedded`, `stacked`, and `full-page` variants
    */
   variant?: TableProps.Variant;
+
+  /**
+   * Total count of items in the table, necessary for screen readers to be aware
```
stirache, line 233:
```suggestion
Use this function to inform screen readers how many items there are in a table.  It returns a total count of all items in a table. If there is an unknown total of items in a table, leave this property undefined.
```

src/table/internal.tsx
```diff
@@ -232,13 +235,19 @@ const InternalTable = React.forwardRef(
...
+              </LiveRegion>
+            )}
             <table
               ref={tableRef}
               className={clsx(styles.table, resizableColumns && styles['table-layout-fixed'])}
               // Browsers have weird mechanism to guess whether it's a data table or a layout table.
               // If we state explicitly, they get it always correctly even with low number of rows.
               role="table"
               aria-label={ariaLabels?.tableLabel}
+              aria-rowcount={totalItemsCount ? totalItemsCount + 1 : -1}
```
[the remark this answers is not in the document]
abdhalees: If `aria-rowcount` is undefined the assistive technology will count the number of visible rows and announce it (the current behaviour)

so if a table has 400 rows and 20 rows per page the screen reader will announce "row 1 of 20" although the table has 400 rows. 

I think it's better not to announce the row count then announcing the wrong count
you can check the current behaviour in our demos https://cloudscape.design/examples/react/table.html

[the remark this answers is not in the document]
avinashbot: > Authors MUST set the value of [aria-rowcount](https://www.w3.org/TR/wai-aria-1.1/#aria-rowcount) to an integer equal to the number of rows in the full table. If the total number of rows is unknown, authors MUST set the value of aria-rowcount to -1 to indicate that the value should not be calculated by the user agent.

completely checks out to me, thanks!

avinashbot (dismissed):

src/table/interfaces.tsx
```diff
@@ -228,6 +228,22 @@ export interface TableProps<T = any> extends BaseComponentProps {
...
+   * If there is an unknown total of items in a table, leave this property undefined.   */
+  totalItemsCount?: number;
+  /**
+   *  Use this property to inform screen readers which range of items is currently displayed in the table.
+   *  It specifies the index (1-based) of the first item in the table.
+   *  If the table has no pagination, leave this property undefined.   */
+  firstIndex?: number;
+  /**
+   * Use this function to announce page changes for screen reader users.
+   * When firstIndex is available, it returns a text when the firstIndex or totalItemsCount change.
```
timogasda, line 244: The wording is a bit confusing to me. You use "when" twice.

> it returns a text when the firstIndex or totalItemsCount change

I would think that the function *always* returns a string, not just when the properties change.

src/table/interfaces.tsx
```diff
@@ -228,6 +228,22 @@ export interface TableProps<T = any> extends BaseComponentProps {
...
+   */
+  liveAnnouncement?: (data: TableProps.LiveAnnouncement) => string;
```
timogasda, line 246: How do we name ARIA live related properties in other components? I think we usually include `aria` in there somewhere.

src/table/interfaces.tsx
abdhalees: reading it again I agree, what about: 

> If firstIndex is available, it announces a text when the firstIndex or totalItemsCount change.

src/table/interfaces.tsx
abdhalees: I didn't find another place where we have a property that is only used in aria-live

timogasda: We have `renderHighlightedAriaLive` for Autosuggest, Select, Multiselect, and `renderSelectedAbsoluteRangeAriaLive` for Date range picker.

src/table/interfaces.tsx
timogasda: A lot better! I'd be fine with it, but I feel like we are still explaining the internals in too much detail. Our customers should already trust us to announce the text at the right time, but I do think mentioning the required properties is useful.

How about:

> Use this function to announce page changes to screen reader users.
> Requires the properties `firstIndex` and `totalItemsCount` to be set correctly.

timogasda (approved):

PR #522, opened 2022-11-23
feat: Inline Cell Editing

fa7ad:
### Description
This PR implements support for inline cell editing in Table.

You should be able to:
1. See an indication of which columns are editable, in the column headers
2. Hover over cells to determine if they are editable (visual indicator)
3. Keyboard users should be able to focus an editable cell by tabbing
    1. While focused, pressing enter should enter edit mode
4. Error text should be displayed if validation is not successful
5. Loading Indicator should be displayed when submitting an edit
6. Editable cells should be navigable using arrow keys (only when selection is disabled)

Demo link posted on slack

### How has this been tested?

I've added some integration tests to verify the behavior

<details>

#### Correctness

#### Security

#### Testing

</details>

---


package.json
```diff
@@ -58,6 +58,7 @@
     "@rollup/plugin-node-resolve": "^15.0.1",
     "@testing-library/jest-dom": "^5.16.4",
     "@testing-library/react": "^12.1.5",
+    "@testing-library/user-event": "^14.4.3",
```
just-boris, line 61: We do not use this library in our test. We use our own test-utils to manipulate with the elements

src/table/interfaces.tsx
```diff
@@ -17,6 +17,13 @@ import { NonCancelableEventHandler, CancelableEventHandler } from '../internal/e
 export interface TableForwardRefType {
   <T>(props: TableProps<T> & { ref?: React.Ref<TableProps.Ref> }): JSX.Element;
 }
+
+interface TableDirtyState {
+  /**
+   * There were unsaved changes in the table.
+   */
+  dirty?: boolean;
```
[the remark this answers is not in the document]
fa7ad: Removed Completely, because the value in this context would have always been wrong

src/table/body-cell/index.tsx
```diff
@@ -2,56 +2,104 @@
...
+  column,
+  isEditActive,
+  onEditStart,
+  onEditEnd,
+  submitEdit,
+  ariaLabels,
+  ...rest
+}: TableBodyCellProps<ItemType>) {
+  const editActivateRef = useRef<ButtonProps.Ref>(null);
+  const tableTdRef = useRef<HTMLTableCellElement>(null);
```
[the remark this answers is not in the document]
fa7ad: Removed

src/table/header-cell/index.tsx
```diff
@@ -49,6 +50,7 @@ export function TableHeaderCell({
   updateColumn,
   resizableColumns,
   onResizeFinish,
+  __isEditable,
```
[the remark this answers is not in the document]
fa7ad: I thought it was a good idea to make it internal to discourage users from directly using the component. But, since this prop only renders an icon, I've made it public now

src/table/__tests__/table.test.tsx
```diff
@@ -133,6 +137,14 @@ test('should render table with accessible headers', () => {
   });
 });
 
+test('should render table with icons', () => {
```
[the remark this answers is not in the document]
fa7ad: Good catch, I'll update the name.

src/table/use-table-focus-navigation.ts
```diff
@@ -0,0 +1,144 @@
...
+      const [rowIndex, columnIndex] = currentFocusCell.current.slice();
+      let newRowIndex = rowIndex;
+      let newColumnIndex = columnIndex;
+
+      if (vertical !== 0) {
+        newRowIndex = Math.min(numRows, Math.max(rowIndex + vertical, 0));
+      }
+
+      if (horizontal !== 0) {
+        newColumnIndex = columnIndex;
```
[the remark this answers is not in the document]
fa7ad: Good catch, thanks

src/table/use-table-focus-navigation.ts
```diff
@@ -0,0 +1,143 @@
...
+        currentFocusCell.current &&
+        tableRoot?.current
+      ) {
+        focusCell(newRowIndex, newColumnIndex);
+      }
+    },
+    [columnDefinitions.length, focusCell, focusableColumns, numRows, tableRoot]
+  );
+
+  const handleArrowKeyEvents = useCallback(
```
avinashbot, line 84: Should this be calling `event.preventDefault()` if focus is moved? Because it looks like it's currently both scrolling the page and moving focus.

avinashbot, line 84: Should we be calling `event.preventDefault()` if focus is moved? Because it looks like it's currently both scrolling the page and moving focus.

src/table/use-table-focus-navigation.ts
```diff
@@ -0,0 +1,143 @@
...
+ * All the functionality is implemented in the hook, so the table component does not
+ * need to implement any keyboard event handlers.
+ * @param enable - Toggle functionality of the hook
+ * @param tableRoot - A ref to a table container. Ideally the root element of the table (<table>); tbody is also acceptable.
+ * @param columnDefinitions - The column definitions for the table.
+ * @param numRows - The number of rows in the table.
+ */
+function useTableFocusNavigation<T extends { editConfig?: TableProps.EditConfig<any, any> }>(
+  enable: boolean,
+  tableRoot: MutableRefObject<any> | null,
```
avinashbot, line 29: Why is this nullable? From what I can tell, it looks like you can get away with:
```suggestion
  tableRoot: RefObject<HTMLTableElement | null>,
```

src/table/header-cell/index.tsx
```diff
@@ -115,6 +117,11 @@ export function TableHeaderCell({
       >
         <div className={clsx(styles['header-cell-text'], wrapLines && styles['header-cell-text-wrap'])} id={headerId}>
           {column.header}
+          {isEditable ? (
```
avinashbot, line 120: Are we going to have a recommendation for the alternative text to be provided in the `ColumnDefinition`'s `ariaLabel`?

src/table/use-table-focus-navigation.ts
```diff
@@ -0,0 +1,143 @@
...
+        default:
+          return;
+      }
+    },
+    [shiftFocus, tableRoot]
+  );
+
+  useEffect(() => {
+    const eventListeners = new Map<[number, number], { focusin(evt: any): void }>();
+    if (tableRoot?.current) {
```
avinashbot, line 115: It feels weird that we're assigning event handlers on the DOM for elements that we control. Why not send `rowIndex` and `cellIndex` into the table cell component and let it attach an event listener?

src/table/interfaces.tsx
```diff
@@ -292,5 +353,12 @@ export namespace TableProps {
...
      */
     scrollToTop(): void;
+
+    /**
+     * Dismiss an inline edit if currently active.
+     */
+    cancelEdit?(): void;
   }
+
+  export type SubmitEditFunction<T> = (item: T, column: TableProps.ColumnDefinition<T>, newValue: any) => Promise<void>;
```
avinashbot, line 363: Why _must_ it return a promise? Currently, I am forced to make my handler an async function even if all I do is synchronous. What about `Promise<void> | void`? To compensate for that, you just need to wrap it with `Promise.resolve`, like  `Promise.resolve(submitEdit(...)).then(() => { ... })`.

src/table/body-cell/index.tsx
```diff
@@ -2,56 +2,98 @@
...
-  wrapLines: boolean | undefined;
-  isFirstRow: boolean;
-  isLastRow: boolean;
-  isSelected: boolean;
-  isNextSelected: boolean;
-  isPrevSelected: boolean;
-  children?: React.ReactNode;
+export { TableTdElement } from './td-element';
+
+const readonlyState = { isEditActive: false, currentValue: '', setValue: () => {} };
```
avinashbot, line 16: `Object.freeze()`? Just to be safe?

src/table/interfaces.tsx
avinashbot: Also, I'm torn about this suggestion, but I'd like to talk it out. What about allowing a second type parameter for the value type (i.e. `SubmitEditFunction<T, V = any>`)? This lets people narrow down the value type if they want to and have a nicer experience with typescript. But because this is only implemented on the edit function, this could also get out of sync with the real values, which would cause more harm than good.

At the very least, `unknown` might be a better fit than `any` so that they're forced to verify the return value type in runtime (or just typecast, I guess).

src/table/body-cell/inline-editor.tsx
```diff
@@ -0,0 +1,118 @@
...
+import { useClickAway } from './click-away';
+import { TableProps } from '../interfaces';
+import styles from './styles.css.js';
+
+// A function that does nothing
+const noop = () => undefined;
+
+interface InlineEditorProps<ItemType> {
+  ariaLabels: TableProps['ariaLabels'];
+  column: TableProps.ColumnDefinition<ItemType>;
```
avinashbot, line 16: A little trick:

```
  column: Extract<TableProps.ColumnDefinition<ItemType>, { editConfig: EditConfig<ItemType, any> }>;
```

src/table/header-cell/index.tsx
fa7ad: Ah I guess we missed this one, but yeah since we do have `ariaLabels` I'm adding an extra prop there

src/table/interfaces.tsx
fa7ad: My assumption was that edits would always be asynchronous in nature, but now that I think about it more I can see why someone would want a synchronous callback. I'll update the type to support sync functions.

src/table/use-table-focus-navigation.ts
fa7ad: I think the logic for this reads better as imperative code rather than following a more descriptive pattern like passing state around components.

src/internal/types.d.ts
```diff
@@ -0,0 +1,32 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+declare global {
```
avinashbot, line 3: Wouldn't this pollute our users' global typings?

src/table/interfaces.tsx
```diff
@@ -246,19 +246,18 @@ export interface TableProps<T = any> extends BaseComponentProps {
 export namespace TableProps {
   export type TrackBy<T> = string | ((item: T) => string);
 
-  export interface CellContext<ValueType = string> {
+  export type EditableColumn<T, V = any> = Extract<ColumnDefinition<T, V>, ColumnEditConfig<T, V>>;
```
avinashbot, line 265: `EditableColumnDefinition` maybe? Just to keep the names matching?

src/table/header-cell/index.tsx
```diff
@@ -119,7 +119,7 @@ export function TableHeaderCell({
           {column.header}
           {isEditable ? (
             <span className={styles['edit-icon']}>
-              <InternalIcon name="edit" />
+              <InternalIcon name="edit" alt={column.editConfig?.editIconAltText} />
```
avinashbot, line 122: This won't work; `alt` is only used for `iconUrl`. You need to wrap it with a span, give it role="img" and provide an aria-label. I think we do this in some other components (like alert).

Our typical convention is probably something like `editIconAriaLabel` for the naming.

src/table/interfaces.tsx
```diff
@@ -246,19 +246,18 @@ export interface TableProps<T = any> extends BaseComponentProps {
...
+
+  export interface CellContext<V> {
     isEditing?: boolean;
-    currentValue: ValueType;
-    setValue: React.Dispatch<ValueType>;
+    currentValue: Optional<V>;
+    setValue: React.Dispatch<Optional<V>>;
```
avinashbot, line 270: Best not to rely on React's typings for something simple like this.

```suggestion
    setValue: (value: V | undefined) => void;
```

avinashbot (changes requested):

src/table/use-table-focus-navigation.ts
avinashbot: Can't say I'm exactly a fan of breaking out of React like this, but I'm not going to block you on it

avinashbot (changes requested):

avinashbot (approved):
Phew!

Al-Dani (dismissed):

src/table/header-cell/index.tsx
```diff
@@ -115,6 +117,15 @@ export function TableHeaderCell({
       >
         <div className={clsx(styles['header-cell-text'], wrapLines && styles['header-cell-text-wrap'])} id={headerId}>
           {column.header}
+          {isEditable ? (
+            <span
+              className={styles['edit-icon']}
+              role="img"
+              aria-label={column.editConfig?.editIconAriaLabel ?? 'editable'}
```
avinashbot, line 124: I wouldn't recommend this. If implementers forget, accessibility checkers could easily find images without alternative text. If we provide a backup value, it's easy to miss, especially for internationalization.

We don't do this anywhere else either.

avinashbot (changes requested):

src/table/interfaces.tsx
```diff
@@ -249,15 +263,63 @@ export interface TableProps<T = any> extends BaseComponentProps {
...
+     */
+    ariaLabel?: string;
+    /**
+     * Specifies an ariaLabel for error icon that is displayed when the validation fails.
+     */
+    errorIconAriaLabel?: string;
+    /**
+     * alternate text for edit icon used in table header
+     */
+    editIconAriaLabel?: string;
```
avinashbot, line 289: Just a thought: wouldn't it be easier if they were included in the table-level `ariaLabels` property? I assume error aria label is always going to be "Error" and edit icon aria label is always going to be "editable".

Not going to go against any team API agreement though, so it's not blocking.

avinashbot (approved):
One non-blocking comment :)

src/test-utils/dom/table/index.ts
```diff
@@ -132,4 +133,26 @@ export default class TableWrapper extends ComponentWrapper {
   findPagination(): PaginationWrapper | null {
     return this.findComponent(`.${styles['tools-pagination']}`, PaginationWrapper);
   }
+
+  findEditingCell(): ElementWrapper | null {
+    return this.findNativeTable().findByClassName(bodyCellStyles['body-cell-edit-active']);
+  }
+
+  findEditingCellInputSlot(): ElementWrapper | null {
+    return (
```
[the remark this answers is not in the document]
fa7ad: Its used in `src/table/__integ__/inline-editing.test.ts`

src/test-utils/dom/table/index.ts
```diff
@@ -132,4 +133,26 @@ export default class TableWrapper extends ComponentWrapper {
...
+      this.findEditingCell()?.findByClassName(bodyCellStyles['body-cell-editor-row'])?.find(':first-child') ?? null
+    );
+  }
+
+  findEditingCellControls(): ElementWrapper | null {
```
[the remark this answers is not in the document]
fa7ad: Making it private

src/test-utils/dom/table/index.ts
```diff
@@ -132,4 +133,26 @@ export default class TableWrapper extends ComponentWrapper {
   findPagination(): PaginationWrapper | null {
     return this.findComponent(`.${styles['tools-pagination']}`, PaginationWrapper);
   }
+
+  findEditingCell(): ElementWrapper | null {
+    return this.findNativeTable().findByClassName(bodyCellStyles['body-cell-edit-active']);
+  }
+
+  findEditingCellInputSlot(): ElementWrapper | null {
```
just-boris, line 141: This should not be a public util.

When testing this interaction, you could do the following:

```
tableWrapper.findEditingCell().findSelect(); // or findInput() or whatever you have inside
```

This selector is only for your internal assertions. You can write this selector inline in the tests which needed it: https://github.com/cloudscape-design/components/search?q=useBrowser+selectors.js&type=code

No need to make it public

src/multiselect/interfaces.ts
```diff
@@ -39,6 +39,11 @@ export interface MultiselectProps extends BaseSelectProps {
    * The event `detail` contains the current `selectedOptions`.
    */
   onChange?: NonCancelableEventHandler<MultiselectProps.MultiselectChangeDetail>;
+
+  /**
+   * Automatically focuses the trigger when component is mounted.
+   */
+  autoFocus?: boolean;
```
just-boris, line 46: To make this change to appear in the changelog, it has to be a separate PR

src/input/internal.tsx
```diff
@@ -168,6 +176,12 @@ function InternalInput(
     attributes.type = 'text';
   }
 
+  useEffect(() => {
```
just-boris, line 179: What was not working with native `autoFocus` attribute, so you had to override its behavior?

src/form-field/internal.tsx
```diff
@@ -91,7 +94,7 @@ export default function InternalFormField({
       )}
 
       <div className={clsx(styles.controls, __hideLabel && styles['label-hidden'])}>
-        <InternalGrid gridDefinition={gridDefinition}>
+        <InternalGrid gridDefinition={gridDefinition} disableGutters={__disableGutters}>
```
just-boris, line 97: Why is this needed? After hiding the label and disabling gutters, do you have anything else left from the form field functionality? If you need validation message, it can be rendered using `<FormFieldError>` sub component directly

src/select/parts/styles.scss
```diff
@@ -36,6 +37,14 @@ $checkbox-size: awsui.$size-control;
 
 .trigger {
   @include styles.text-overflow-ellipsis;
+  @include focus-visible.when-visible {
```
just-boris, line 40: What do these styles do? Cannot see them used anywhere

src/table/body-cell/index.tsx
```diff
@@ -2,72 +2,128 @@
...
+    ...(focusVisible as Record<string, string>),
+    onFocus: scheduleRestoreScrollPosition,
+    'data-inline-editing-active': isEditing.toString(),
+  };
+
+  useEffectOnUpdate(() => {
+    if (!isEditing && editActivateRef.current) {
+      editActivateRef.current.focus({ preventScroll: true });
+    }
+    const timer = scheduleRestoreScrollPosition();
```
just-boris, line 74: What happens if these two events occur same time?

```js
> 66 onFocus: scheduleRestoreScrollPosition,
> 74 const timer = scheduleRestoreScrollPosition();
```

Why does the first one not store the timer for cancellation?

src/table/body-cell/index.tsx
```diff
@@ -2,72 +2,128 @@
...
+
+  const tdNativeAttributes = {
+    ...(focusVisible as Record<string, string>),
+    onFocus: scheduleRestoreScrollPosition,
+    'data-inline-editing-active': isEditing.toString(),
+  };
+
+  useEffectOnUpdate(() => {
+    if (!isEditing && editActivateRef.current) {
+      editActivateRef.current.focus({ preventScroll: true });
```
just-boris, line 72: Why do you prevent native scroll and schedule manual scroll right after?

just-boris: P.S. This entire scroll management story produced at a lot of non-idiomatic code. Let's talk in DM more about reasons for it and how to convert into more idiomatic code

src/table/use-table-focus-navigation.ts
```diff
@@ -0,0 +1,166 @@
...
+  const focusableColumns = useMemo(() => {
+    const cols = columnDefinitions.map(column => !!column.editConfig);
+    if (selectionType !== 'none') {
+      cols.unshift(false);
+    }
+    return cols;
+  }, [columnDefinitions, selectionType]);
+
+  const maxColumnIndex = useMemo(() => focusableColumns.length - 1, [focusableColumns]);
+  const minColumnIndex = useMemo(() => (selectionType === 'none' ? 1 : 0), [selectionType]);
```
just-boris, line 44: No point in memoizing primitive values. They have no reference equality

src/table/use-table-focus-navigation.ts
```diff
@@ -0,0 +1,166 @@
...
+  useEffect(() => {
+    const eventListeners = new Map<[number, number], { focusin(evt: any): void }>();
+    // istanbul ignore if
+    if (!tableRoot.current) {
+      return;
+    }
+
+    const tableElement = tableRoot.current;
+
+    // istanbul ignore next (tested in use-focus-navigation.test.tsx#L210)
```
just-boris, line 137: If it is tested, then ignore is not needed, because there is coverage. If not tested, then this comment is misleading

src/table/use-table-focus-navigation.ts
```diff
@@ -0,0 +1,166 @@
...
+        return;
+      }
+      const listenerFns = {
+        focusin: () => {
+          currentFocusCell.current = [rowIndex, cellIndex];
+        },
+      };
+      eventListeners.set([rowIndex, cellIndex], listenerFns);
+      cell.addEventListener('focusin', listenerFns.focusin, { passive: true });
+    });
```
just-boris, line 159: `focusin` event can bubble, so this can become a single listener

```js
function focusinCell(event) {
  const current = event.currentTarget;
  const td = findUpUntil(node => node.tagName === 'TD');
  if(!td) return;
  
  const tr = td.parentNode;
  const cellIndex = tr.children.indexOf(td);
  const rowIndex = tr.parentNode.indexOf(tr);

  currentFocusCell.current = [rowIndex, cellIndex];
}
```

src/table/use-table-focus-navigation.ts
```diff
@@ -0,0 +1,166 @@
...
+ * This hook is used to navigate between table cells using the keyboard arrow keys.
+ * All the functionality is implemented in the hook, so the table component does not
+ * need to implement any keyboard event handlers.
+ * @param enable - Toggle functionality of the hook
+ * @param tableRoot - A ref to a table container. Ideally the root element of the table (<table>); tbody is also acceptable.
+ * @param columnDefinitions - The column definitions for the table.
+ * @param numRows - The number of rows in the table.
+ */
+function useTableFocusNavigation<T extends { editConfig?: TableProps.EditConfig<any> }>(
+  selectionType: 'single' | 'multi' | 'none' = 'none',
```
just-boris, line 28: This argument should not have an own default value, because it always must come from the table

src/table/use-table-focus-navigation.ts
fa7ad: Its tested in integration, but coverage doesn't seem to pick that up

just-boris: [The linked file](https://github.com/fa7ad/components/blob/feat/table-cell-editing/src/table/__tests__/use-focus-navigation.test.tsx) is a unit test, should work fine. Did you mean to link a different file?

src/select/parts/styles.scss
fa7ad: This adds a focus ring around select when you activate it through keyboard. Other inputs do this already, but this was missing from all the inputs based on select

src/input/internal.tsx
fa7ad: this is related to the scroll management, native autofocus doesn't maintain user's scroll position. it tries to scroll the element into view

src/select/parts/styles.scss
just-boris: First of all, there is already an existing focus outline on select component. What in the existing styles was not working?

Second, the `.trigger` styles are defined here: https://github.com/cloudscape-design/components/blob/015b28e11a7bb424412e61305a0bb4b9b0a151da/src/select/parts/trigger.tsx#L57

It does not receive `focusVisible` attributes, so these CSS selector never matches

src/internal/context/form-field-context.ts
```diff
@@ -47,7 +47,11 @@ export interface FormFieldValidationControlProps extends FormFieldControlProps {
   invalid?: boolean;
 }
 
-export const FormFieldContext = createContext<FormFieldValidationControlProps>({});
+interface InternalFormFieldControlProps extends FormFieldValidationControlProps {
+  __useReactAutofocus?: boolean;
```
just-boris, line 51: One more call out, for the record. We need to find another way to pass it down, or possibly make it mandatory for consumers to use `autoFocus` on their end

src/table/use-table-focus-navigation.ts
fa7ad: Yeah I think I wrote the wrong file name

src/select/parts/styles.scss
fa7ad: It matches on ButtonTrigger

https://github.com/cloudscape-design/components/blob/015b28e11a7bb424412e61305a0bb4b9b0a151da/src/internal/components/button-trigger/index.tsx#L9

just-boris: That line generates the following css-selector in the end

```
.awsui_trigger_dwuol_1dfho_116[data-awsui-focus-visible=true]:not(#\9):focus::before
```

In order to work, the attribute has to be on that node directly, not anywhere else. Button trigger has is own scoped styles, so selectors from this file have no effect there either

just-boris (approved):

src/form-field/internal.tsx
timogasda: It's also needed for rendering constraint text and connecting ARIA properties.

just-boris: The point remains the same. Instead of conditionally turning off stuff we should use composition. Extract the form field basics and use them here instead

src/select/parts/styles.scss
timogasda: Addressed in https://github.com/cloudscape-design/components/pull/891

src/table/body-cell/use-stable-scroll-position.tsx
```diff
@@ -0,0 +1,62 @@
...
+ * `restoreScrollPosition` is called.
+ * @param activeElementRef Ref to an active element in the table. This is used to find the nearest scrollable parent.
+ */
+export function useStableScrollPosition<T extends HTMLElement>(
+  activeElementRef: React.RefObject<T>
+): UseStableScrollPositionResult {
+  const scrollRef = useRef<Parameters<HTMLBodyElement['scroll']>>();
+
+  const storeScrollPosition = useCallback(() => {
+    const scrollableParent = getScrollableParent(activeElementRef.current ?? document.body);
```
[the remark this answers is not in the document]
timogasda: Addressed in https://github.com/cloudscape-design/components/pull/891

src/internal/context/form-field-context.ts
timogasda: Addressed (removed) in https://github.com/cloudscape-design/components/pull/916

src/table/body-cell/index.tsx
timogasda: Addressed (removed) in https://github.com/cloudscape-design/components/pull/916

src/table/body-cell/use-stable-scroll-position.tsx
```diff
@@ -0,0 +1,62 @@
...
+    if (scrollableParent) {
+      scrollRef.current = [scrollableParent.scrollLeft, scrollableParent.scrollTop];
+    }
+  }, [activeElementRef]);
+
+  const restoreScrollPosition = useCallback(() => {
+    const scrollableParent = getScrollableParent(activeElementRef.current ?? document.body);
+    if (
+      scrollRef.current &&
+      scrollRef.current.toString() !== '0,0' &&
```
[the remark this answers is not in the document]
timogasda: Addressed (removed) in https://github.com/cloudscape-design/components/pull/916

PR #526, opened 2022-11-24
refactor: Make range calendar selection controllable

just-boris:
### Description

Move state into a single place in `DateRangePickerDropdown`  component. This will allow to expose programmatic API to control this value outside the component

### How has this been tested?

PR build, dev-pipeline (link in DM)

<details>

#### Correctness

#### Security

#### Testing

</details>

---


pages/date-range-picker/range-calendar.page.tsx
```diff
@@ -23,7 +25,7 @@ export default function RangeCalendarScenario() {
 
         <RangeCalendar
           value={value}
-          onChange={setValue}
+          setValue={setValue}
```
just-boris, line 28: onChange has become setValue to support full React's setState signature, including update in callback

src/date-range-picker/calendar/grids/index.tsx
```diff
@@ -97,7 +80,7 @@ export const Grids = ({
       const newMonth = !isSingleGrid && direction === -1 ? addMonths(baseDate, -1) : baseDate;
       const nearestBaseDate = getBaseDate(newMonth, isDateEnabled);
 
-      const newFocusedDate = selectFocusedDate(focusedDate, nearestBaseDate, isDateEnabled);
+      const newFocusedDate = findDateToFocus(focusedDate, nearestBaseDate, isDateEnabled);
```
just-boris, line 83: If function does not make any side effects (actually selecting something), then it is find..., not select...

src/date-range-picker/calendar/index.tsx
```diff
@@ -42,62 +42,24 @@ export default function DateRangePickerCalendar({
...
+      if (!isSingleGrid && isSameMonth(startDate, addMonths(currentMonth, -1))) {
+        return startDate;
       }
     }
-    return selectFocusedDate(rangeStart.date, currentMonth, isDateEnabled);
+    return findDateToFocus(parseDate(value.start.date), currentMonth, isDateEnabled);
   });
 
-  // This effect "synchronizes" the local state update back up to the parent component.
-  useEffect(() => {
```
just-boris, line 84: React docs explicitly call this an anti-pattern: https://beta.reactjs.org/learn/you-might-not-need-an-effect#notifying-parent-components-about-state-changes

Avoid code patterns like this

src/internal/utils/date-time/format-date-time.ts
```diff
@@ -0,0 +1,8 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import { formatDate } from './format-date';
+import { formatTime } from './format-time';
+
+export function formatDateTime(date: Date) {
```
just-boris, line 6: Very surprised we did not have this function yet. Did I miss something?

src/internal/utils/date-time/join-date-time.ts
```diff
@@ -4,3 +4,8 @@
 export function joinDateTime(dateString: string, timeString: string) {
   return `${dateString}T${timeString}`;
 }
+
+export function splitDateTime(dateStr: string) {
```
just-boris, line 8: Taking suggestions how to rename the file, since now it includes both ways, split and join

michaeldowseza: Since we seem to be splitting every function into its own file, this could also go into `slit-date-time.ts`

just-boris: I would prefer having symmetric functions in a single file (because they are always changed together)

michaeldowseza (approved):

PR #528, opened 2022-11-25
fix(attribute-editor): Misaligned 'Remove' button when no label is present

connorlanigan:
When there is no visible label on any of the columns (e.g. because the label is inherited from a parent form field), the 'Remove' button had an incorrect vertical offset applied.

### Description

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


jperals (commented):
There is also a layout issue when one label is provided but not the other. Do we want to support that use case as well?

Screenshot 2022-11-28 at 10 43 10

connorlanigan:
When there are multiple columns, they should all have labels, in order to differentiate what each column is for. If one of the columns doesn't have a label, it would be very difficult to users to know what that column is and what they should put there, so it's not really a supported or expected use case.

This fix is primarily for the special case where there is only a single column. In that case, having a label on the column is redundant if there is already a visible label provided by a FormField (and this label is also used for announcing the column's input fields to screenreaders), so it is an expected use case to have no label in single-column situations.

jperals (approved):

Issue #503, opened 2022-11-20
[Bug]: Embedded `Table` with `PropertyFilter` styles issue


nichita-pasecinic:
### Package version

^3.0.126

### React version

^17.0.2

### Description

Hi,

 An `Table` component with prop `variant={'embedded'}` and filter component like so:
```jsx
filter={
  <PropertyFilter
    i18nStrings={PROPERTY_FILTERING_I18N_CONSTANTS}
    countText={filterCountText}
    {...propertyFilterProps}
    expandToViewport
  />
}
```

has some weird styles:

![image](https://user-images.githubusercontent.com/98735377/202920129-5b841d0a-265a-4160-a13f-643f8c6f91c3.png)

Is this expected ?

### Source code

```jsx
<Table
      {...collectionProps}
      variant={'embedded'}
      items={items}
      columnDefinitions={columnDefinitions}
      resizableColumns
      filter={
        <PropertyFilter
          i18nStrings={PROPERTY_FILTERING_I18N_CONSTANTS}
          countText={filterCountText}
          {...propertyFilterProps}
          expandToViewport
        />
      }
    />
```

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/cloudscape-design/components/blob/main/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/cloudscape-design/components/issues) for duplicate problems

connorlanigan:
Hello,

the Table in your screenshot looks similar to [the "Embedded table" playground example on our website](https://cloudscape.design/components/table/?tabId=playground&example=embedded-table). Can you describe how the table in your screenshot differs from what you expect?

nichita-pasecinic:
Hi @connorlanigan 
this part (border radiuses/corners) looks odd to me 
![image](https://user-images.githubusercontent.com/98735377/203054850-24051a7d-f075-4ca7-a159-aabd79d1edac.png)

connorlanigan:
Ah, I see, thank you. I will bring this up with the team.

timogasda:
Hello @nichita-pasecinic !

Sorry for the late response. This should have been fixed with #529. Please update your Cloudscape version and try again.

PR #529, opened 2022-11-25
fix: Remove table round corner under header

YueyingLu:
### Description

Table rounded corner causes gap between header. The fix is to remove the border-radius when header exists.  

Related links, issue AWSUI-19854

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/table/styles.scss
+++ b/src/table/styles.scss
@@ -61,8 +61,6 @@
   padding-top: awsui.$space-table-content-top;
   padding-bottom: awsui.$space-table-content-bottom;
   overflow-x: auto;
-  border-top-right-radius: awsui.$border-radius-container;
-  border-top-left-radius: awsui.$border-radius-container;
   &.variant-stacked,
   &.variant-container {
     & > .table {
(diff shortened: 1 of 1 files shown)


abdhalees (commented):
can you update the commit message and PR title to something that describes what the code does? e.g. `fix: Remove round corner from embedded table header`

src/table/styles.scss
```diff
@@ -79,6 +77,10 @@
   &.has-footer {
     padding-bottom: 0px;
   }
+  &:not(.has-header) {
```
abdhalees, line 80: the issue says it only happens in embedded tables, why do we need to remove the border from every table?

YueyingLu: It actually happens to not only embedded. But in general, when there is already a header. For example `container`:
Screenshot 2022-11-28 at 11 14 54

abdhalees (approved):

timogasda:
This fixes https://github.com/cloudscape-design/components/issues/503.

PR #536, opened 2022-11-29
AWSUI-19252: Internalink has modified ariaLabel

karmanya79:
### Description

Modifying aria-label of InternalLink so it has more precise value

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/top-navigation/parts/utility.tsx
+++ b/src/top-navigation/parts/utility.tsx
@@ -12,6 +12,7 @@ import { TopNavigationProps } from '../interfaces';
 
 import styles from '../styles.css.js';
 import { checkSafeUrl } from '../../internal/utils/check-safe-url';
+import { joinStrings } from '../../internal/utils/strings';
 
 export interface UtilityProps {
   hideText: boolean;
@@ -22,7 +23,9 @@ export interface UtilityProps {
 export default function Utility({ hideText, definition, offsetRight }: UtilityProps) {
   const hasIcon = !!definition.iconName || !!definition.iconUrl || !!definition.iconAlt || !!definition.iconSvg;
   const shouldHideText = hideText && !definition.disableTextCollapse && hasIcon;
-  const ariaLabel = definition.ariaLabel ?? definition.text;
+  const ariaLabel = definition.ariaLabel
+    ? definition.ariaLabel
+    : joinStrings(definition.text, definition.externalIconAriaLabel);
 
   if (definition.type === 'button') {
     checkSafeUrl('TopNavigation', definition.href);
@@ -67,7 +70,7 @@ export default function Utility({ hideText, definition, offsetRight }: UtilityPr
           <InternalLink
             variant="top-navigation"
             href={definition.href}
-            target={definition.external ? '_blank' : undefined}
+            external={true}
             onFollow={definition.onClick}
             ariaLabel={ariaLabel}
           >


src/top-navigation/parts/utility.tsx
```diff
@@ -22,7 +23,9 @@ export interface UtilityProps {
 export default function Utility({ hideText, definition, offsetRight }: UtilityProps) {
   const hasIcon = !!definition.iconName || !!definition.iconUrl || !!definition.iconAlt || !!definition.iconSvg;
   const shouldHideText = hideText && !definition.disableTextCollapse && hasIcon;
-  const ariaLabel = definition.ariaLabel ?? definition.text;
+  const ariaLabel = definition.ariaLabel
+    ? definition.ariaLabel
+    : joinStrings(definition.text, definition.externalIconAriaLabel);
```
avinashbot, line 28: `externalIconAriaLabel ` only exists for `type === "button"`. It looks like you'll need to redefine this inside the if condition (i.e. after line 30).

avinashbot (changes requested):

src/top-navigation/parts/utility.tsx
karmanya79: yes, just noticed this, will update it

src/top-navigation/parts/utility.tsx
```diff
@@ -67,7 +71,7 @@ export default function Utility({ hideText, definition, offsetRight }: UtilityPr
           <InternalLink
             variant="top-navigation"
             href={definition.href}
-            target={definition.external ? '_blank' : undefined}
+            external={true}
```
avinashbot, line 74: Also make sure to delete lines 90-101 (the icon hack). Pretty sure this renders two external icons now :)

avinashbot (changes requested):

src/top-navigation/parts/utility.tsx
```diff
@@ -22,9 +23,12 @@ export interface UtilityProps {
 export default function Utility({ hideText, definition, offsetRight }: UtilityProps) {
   const hasIcon = !!definition.iconName || !!definition.iconUrl || !!definition.iconAlt || !!definition.iconSvg;
   const shouldHideText = hideText && !definition.disableTextCollapse && hasIcon;
-  const ariaLabel = definition.ariaLabel ?? definition.text;
+  let ariaLabel = definition.ariaLabel ?? definition.text;
 
   if (definition.type === 'button') {
+    ariaLabel = definition.ariaLabel
+      ? definition.ariaLabel
+      : joinStrings(definition.text, definition.externalIconAriaLabel);
```
timogasda, line 31: I think we should avoid "building" any strings for alternative text because we don't know the implications in some localizations.

We have the `externalIconAriaLabel` property on `Link`/`InternalLink`, right? I think it would make more sense to pass that down. What do you think?

timogasda (changes requested):

src/top-navigation/parts/utility.tsx
avinashbot: Ideally we'd avoid it, yes. The issue we have in this case is that the ariaLabel is always provided for utilities, and always overrides the internal externalIconAriaLabel. If we only want to provide ariaLabel if the text isn't visible, we'd need to conditionally provide that based on (1) if text is provided, (2) if text is responsively collapsed, (3) if it's overridden to always stay open. Additionally, people might need to provide aria labels even if text is provided because the icon could have additional visual meaning.

So this approach seemed simpler.

src/top-navigation/parts/utility.tsx
timogasda, line 74: I think this needs to be `external={definition.external}` ;)

timogasda (changes requested):

src/top-navigation/parts/utility.tsx
timogasda: I see. The whole responsiveness logic makes this a little unintuitive either way we go. I'm fine with this approach then.

src/top-navigation/parts/utility.tsx
avinashbot: For your convenience:
```suggestion
            external={definition.external}
```

timogasda (approved):

avinashbot (approved):

PR #538, opened 2022-11-30
fix: let line chart highlight all points at a given X coordinate usin…

jperals:
…g keyboard navigation

### Description

Issue: AWSUI-19201

Context: mouse hover on line chart allows to highlight not only one single point but also all points (from all series) at a given X coordinate, but this behavior was missing for keyboard users. This PR adds it.

The visual appearance is the same as when the points are highlighted with the mouse, but with the focus indicator around the points and the  vertical marker:
Screenshot 2022-11-30 at 09 00 09

### How has this been tested?

- Added unit tests and integration tests
- Manually tested on local server
- Validated new aria-label with VoiceOver

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/components/cartesian-chart/vertical-marker.tsx
+++ b/src/internal/components/cartesian-chart/vertical-marker.tsx
@@ -16,12 +16,13 @@ export interface VerticalMarkerProps {
   showPoints?: boolean;
   showLine?: boolean;
   points: null | readonly SeriesPoint[];
+  ariaLabel?: string;
 }
 
 export default memo(forwardRef(VerticalMarker));
 
 function VerticalMarker(
-  { height, showPoints = true, showLine = true, points }: VerticalMarkerProps,
+  { height, showPoints = true, showLine = true, points, ariaLabel }: VerticalMarkerProps,
   ref: React.Ref<SVGLineElement>
 ) {
   const [firstPoint] = points || [];
@@ -31,6 +32,7 @@ function VerticalMarker(
       <line
         ref={ref}
         aria-hidden="true"
+        aria-label={ariaLabel}
         className={styles['vertical-marker']}
         style={{ visibility: showLine && firstPoint ? 'visible' : 'hidden' }}
         x1={firstPoint?.x}
--- a/src/mixed-line-bar-chart/hooks/use-navigation.ts
+++ b/src/mixed-line-bar-chart/hooks/use-navigation.ts
@@ -3,7 +3,7 @@
 import { useCallback, useMemo, useState } from 'react';
 import { KeyCode } from '../../internal/keycode';
 import { ChartContainerProps } from '../chart-container';
-import { ChartDataTypes, MixedLineBarChartProps } from '../interfaces';
+import { ChartDataTypes, MixedLineBarChartProps, VerticalMarkerLeft } from '../interfaces';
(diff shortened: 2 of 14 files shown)


tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "lib": ["es5", "es2015.collection", "dom"],
-    "types": [],
+    "types": ["jest"],
```
jperals, line 4: This was necessary for WebStorm to interpret test files correctly. Opened the project with VSCode as well to make sure nothing was broken.

jperals: Removed in latest commit after offline conversation, since this would affect non-test files as well.

src/internal/components/cartesian-chart/vertical-marker.tsx
```diff
@@ -16,12 +16,13 @@ export interface VerticalMarkerProps {
...
   showLine?: boolean;
   points: null | readonly SeriesPoint[];
+  ariaLabel?: string;
 }
 
 export default memo(forwardRef(VerticalMarker));
 
 function VerticalMarker(
-  { height, showPoints = true, showLine = true, points }: VerticalMarkerProps,
+  { height, showPoints = true, showLine = true, points, ariaLabel }: VerticalMarkerProps,
```
avinashbot, line 25: Actually, it looks like this does nothing, because it's placed on an element with `aria-hidden`. 

Inside chart, it looks like all announcements are being made with a live region inside ChartPlot, so I don't think we don't need to pass an aria label.

avinashbot (changes requested):

src/internal/components/cartesian-chart/vertical-marker.tsx
jperals: As discussed, the way this works is that the `aria-label` is copied over to another element which is the one that is actually focused. But on the other hand, we decided that we don't want this extra information to be announced so I've removed this prop in the latest commit.

Al-Dani (dismissed):

src/mixed-line-bar-chart/hooks/use-navigation.ts
```diff
@@ -80,10 +81,21 @@ export function useNavigation<T extends ChartDataTypes>({
...
-      onLineGroupFocus();
+      onLineFocus();
     }
   };
 
+  const allUniqueX = useMemo(() => {
+    const result = [];
+    for (let i = 0; i < scaledSeries.length; i += 1) {
+      const point = scaledSeries[i];
+      if (point !== undefined && (!result.length || result[result.length - 1].scaledX !== point.x)) {
```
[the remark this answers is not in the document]
jperals: Yes, we are assuming that points with the same `x` will be consecutive. I think we can assume that since `scaledSeries` is sorted by `x` as it is generated by the parent component `ChartContainer` by calling `makeScaledSeries`, which performs the sorting here:  https://github.com/cloudscape-design/components/blob/main/src/mixed-line-bar-chart/make-scaled-series.ts#L96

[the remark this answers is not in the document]
jperals: Added a comment in a new commit

avinashbot (approved):

PR #546, opened 2022-12-02
chore: Upgrade dependencies

michaeldowseza:
### Description

Regenerate package-lock file to receive latest updates. This is done here to make it easier to separate the upgrade to Node16 from the fixing of dependencies caused by regenerating the lock file.

- Regenerate package-lock
- Upgrade eslint + stylelint deps. This is done because the updated patch versions received from just regenerating the package-lock file break certain files in unexpected ways. This was fixed in newer versions of these dependencies
- Fix eslint and stylelint issues

n/a

### How has this been tested?

- No new tests added

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/app-layout/visual-refresh/main.scss
```diff
@@ -121,7 +129,7 @@
 }
 
 // Prevent content that is visually hidden behind drawers and content area in mobile view from receiving focus
-/* stylelint-disable-next-line selector-combinator-disallowed-list, selector-max-universal */
```
michaeldowseza, line 124: Apparently the order matters now. So it needs to ignore the universal selector first and then the combinator

src/internal/components/cartesian-chart/bottom-labels.tsx
```diff
@@ -82,7 +82,6 @@ function BottomLabels({
     <g
       transform={`translate(0,${height})`}
       className={styles['labels-bottom']}
-      focusable={false}
```
michaeldowseza, line 85: I can't find this attribute on the spec https://developer.mozilla.org/en-US/docs/Web/SVG/Element/g . Looking for input here, should this be `tabIndex={-1}`?

tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "lib": ["es5", "es2015.collection", "dom"],
```
michaeldowseza, line 3: I don't know what changed here that this is suddenly required 🤷🏻‍♂️

build-tools/stylelint/__tests__/no-motion-outside-of-mixin.test.js
```diff
@@ -64,29 +64,29 @@ describe('No motion outside of mixin', () => {
         }
       `);
       expect(result.errored).toBe(true);
-      expect(result.results[0].warnings).toEqual([
```
michaeldowseza, line 67: Updated deps brought in new props to this returned object but the returned object is different for each of the test permutations. Made these slightly more generic by not checking the line and column

build-tools/stylelint/__tests__/no-motion-outside-of-mixin.test.js
```diff
@@ -64,29 +64,29 @@ describe('No motion outside of mixin', () => {
...
-        {
-          column: 15,
-          line: 5,
-          rule: '@cloudscape-design/no-motion-outside-of-mixin',
-          severity: 'error',
-          text: `Property "${property}" should be directly under 'with-motion' helper (@cloudscape-design/no-motion-outside-of-mixin)`,
-        },
-      ]);
+      expect(result.results[0].warnings.length).toBe(1);
```
just-boris, line 67: What changed here? New fields in the output?

1. Can we add them to the assertion? I vote for this if they are deterministic
2. Can we wrap the former assertion into `expect.objectContaining` to keep the former "single expect" form?

just-boris: P.S. everything below looks good and makes sense, thanks for upgrading

src/internal/components/cartesian-chart/bottom-labels.tsx
avinashbot: https://allyjs.io/tutorials/focusing-in-svg.html  
Looks like it's IE only, so I'm fine with dropping this.

tsconfig.json
avinashbot: I'm also curious what caused this, but I'm okay with pulling in es2015 libs.

avinashbot (approved):

build-tools/stylelint/__tests__/no-motion-outside-of-mixin.test.js
just-boris: How about 

```js
expect(result.results[0].warnings).toEqual([
   expect.objectContaining({
      rule: '@cloudscape-design/no-motion-outside-of-mixin',
      severity: 'error',
      text: `Property "${property}" should be directly under 'with-motion' helper (@cloudscape-design/no-motion-outside-of-mixin)`
   })
]);
```

The benefit of this pattern is that it will print you all object difference, not only one field. Every failed `expect` aborts the test

just-boris (dismissed):
thanks

build-tools/stylelint/__tests__/no-motion-outside-of-mixin.test.js
michaeldowseza: Changed 👍🏻

just-boris (approved):

PR #548, opened 2022-12-02
chore: set `"sideEffects": false` flag to reduce bundle size

ryparker:
## Description

This PR adds the "sideEffects" flag to package.json so that webpack >= v4 will skip bundling unused modules/files. Because this library uses index.ts to mass export directory modules, this change should reduce bundle size dramatically.

See: https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
   },
   "homepage": "https://cloudscape.design",
   "files": [],
+  "sideEffects": false,
   "scripts": {
     "quick-build": "gulp quick-build",
     "build": "NODE_ENV=production gulp build",


ryparker:
After doing some testing with webpack it looks like the flag made no difference in the bundle size.

Without "sideEffects" flag
without-sideeffects-flag

With "sideEffects": false flag
with-sideeffects-flag

just-boris:
Hello!

Just in case, `sideEffects` field _is_ already included in our package.json. It is dynamically added in the build process: https://github.com/cloudscape-design/components/blob/59a17971fb3a9644335f446c5fadf5d97f62e95e/build-tools/tasks/package-json.js#L86

Here you can also see it in the final artifact: https://unpkg.com/browse/@cloudscape-design/components@3.0.130/package.json

If you did not see it in your local install, it must be something on your end

ryparker:
👍🏻 Thanks for clearing that up Boris.

PR #557, opened 2022-12-05
feat: Table spacing adjustments for Visual Refresh.

DaemonCahill:
### Description

This PR implements a subset of planned adjustments to the table spacing in Visual Refresh:

- Tables within containers are only indented once
- Tables within containers with horizontal scroll and a footer extend the divider to the edge
- Tables without selection remove left padding from the first cell
- Tables with selection align the selection control with the filter search icon
- Tables with sorting align the last cell caret icon with the pagination preferences icon


src/table/body-cell/styles.scss
```diff
@@ -96,3 +105,19 @@ $border-placeholder: awsui.$border-item-width solid transparent;
...
+*/
+.body-cell:not(.is-visual-refresh) {
+  &:first-child {
+    padding-left: $cell-edge-horizontal-padding;
+  }
+}
+
+.body-cell.is-visual-refresh {
+  &:first-child {
+    padding-left: 0;
```
jperals, line 121: This seems to affect also the full page variant. Should this one keep a left padding? See `/#/light/table/full-page-variant/`

Screenshot 2022-12-08 at 11 07 01

DaemonCahill: Good find! This is intentional as the full page variant has no internal padding and is assumed to be used exclusively within the `AppLayout` component (with `contentType='table'`). The same behavior can be seen with the `Wizard` component. We are having some internal discussions on how to better handle our full page patterns but for the purposes of this PR this is correct.

jperals:
> Tables within containers with horizontal scroll and a footer extend the divider to the edge

Do we have any visual test for this?

src/table/body-cell/index.tsx
```diff
@@ -17,6 +18,7 @@ interface TableBodyCellProps {
   isNextSelected: boolean;
   isPrevSelected: boolean;
   children?: React.ReactNode;
+  hasFooter: boolean | undefined;
```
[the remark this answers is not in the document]
DaemonCahill: Sure, no problem!

src/table/styles.scss
```diff
@@ -113,12 +120,18 @@
   /* used in test-utils */
 }
 
+/*
+The min/max/width token values in in Visual Refresh should align
```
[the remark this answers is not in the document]
DaemonCahill: Good catch!

style-dictionary/visual-refresh/spacing.ts
```diff
@@ -20,7 +20,7 @@ const tokens: StyleDictionary.SpacingDictionary = {
   spaceCodeEditorStatusFocusOutlineGutter: '-7px',
   spaceContainerContentTop: '0px',
   spaceContainerHeaderVertical: '{spaceS}',
-  spaceContainerHorizontal: '{spaceXl}',
+  spaceContainerHorizontal: '{spaceL}',
```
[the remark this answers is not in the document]
DaemonCahill: Yes, this is intentional! The design team decided collectively to adjusted the container padding as well. Supposedly this was an oversight with the initial release of GA.

src/table/internal.tsx
```diff
@@ -243,7 +243,11 @@ const InternalTable = React.forwardRef(
             )}
             <table
               ref={tableRef}
-              className={clsx(styles.table, resizableColumns && styles['table-layout-fixed'])}
+              className={clsx(
+                styles.table,
+                resizableColumns && styles['table-layout-fixed'],
+                hasFooter && styles['has-footer']
```
jperals, line 249: Does the `has-footer` class do anything on the `table` element? I couldn't find any selector in the styles that would match this.

jperals (approved):

src/table/internal.tsx
DaemonCahill: Good catch. It looks like it is unused.

src/table/styles.scss
```diff
@@ -113,9 +113,18 @@
...
 
+/*
+The min/max/width token values in in Visual Refresh should align 
+the table header and body cells selection control with the table
+filter search icon.
+*/
 .selection-control {
-  position: relative;
   box-sizing: border-box;
+  max-width: awsui.$size-table-selection-horizontal;
```
jperals, line 130: Does it make sense to define `max-width` and `min-width` to the same value that is already used as `width`?

src/table/header-cell/styles.scss
```diff
@@ -120,6 +120,11 @@ in the pagination slot.
     padding-left: awsui.$space-xxxs;
   }
 
+  /*
+  Striped rows requires additional left padding because the 
+  shaded background makes the child content appear too close 
+  the the table edge.
```
jperals: Nitpick: "the the" instead of "to the" :)

jperals: Oh I see this was fixed later on

jperals (approved):

src/table/styles.scss
DaemonCahill: I'm not sure, to be honest. This is the way the original code was written so I just stuck with it lest I create unintended consequences.

jperals: Right, I see

PR #566, opened 2022-12-08
fix: Prevent empty table from flickering

abdhalees:
### Description

bring back this reverted commit https://github.com/cloudscape-design/components/pull/563 
with fixing the regression it caused, sticky header column widths were not synced when items change by pagination or after loading finishes

### How has this been tested?

added an integration test to check if the header widths are synced when items change

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/table/use-sticky-header.ts
```diff
@@ -60,7 +60,7 @@ export const useStickyHeader = (
         primaryTable.style.marginTop = '';
       }
     };
-  });
+  }, [secondaryTableRef, tableWrapperRef, syncColumnHeaderWidths, items]);
```
just-boris, line 63: This is not future-proof, because `items` array can change without changing the actual content (for example, refreshing same data).

The same flickering is still possible under some conditions. What will happen if we remove the return value from this effect completely? Does it cause any new issues?

abdhalees: nope, that solves it because the cleanup is what actually causes the flickering
that was my first solution but didn't want to remove the clean up.

just-boris: I cannot see the scenario where not having this cleanup breaks anything. When table unmounts, all of these nodes are destroyed anyway. And during table lifetime `syncColumnHeaderWidths` is always called, as we expect.

abdhalees: removed the cleanup and hook dependancies array and the angular hack.
but kept the test in case we attempt to break this again

just-boris (approved):

just-boris (approved):

PR #574, opened 2022-12-13
chore: remove unused code

just-boris:
### Description

Remove unused code. Most of it is a leftover when we extracted visual refresh app-layout into `src/app-layout/visual-refresh`

### How has this been tested?

Locally and with PR build

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/app-layout/tools-and-split-panel.tsx
```diff
@@ -28,8 +26,7 @@ interface ToolsAndSplitPanelProps {
   contentHeightStyle: ReturnType<typeof useContentHeight>['contentHeightStyle'];
 
   tools: React.ReactNode;
-  splitPanel?: React.ReactNode;
```
just-boris, line 31: Do not use `?` on any internal interfaces, it is only for public. 

This is important to not forget to pass the prop down

When using `?`
```
<ToolsAndSplitPanel
   // forgot to pass splitPanel, typescript is quiet about it
   // splitPanel={splitPanel}
/>
```

When using  `value | undefined` (already included in `React.ReactNode` in our case)

```
<ToolsAndSplitPanel
   // forgot to pass splitPanel, typescript complains
   // splitPanel={splitPanel}
/>

<ToolsAndSplitPanel
   // technically it is the same as missing prop, but explicit that we did not forget to pass it down
   splitPanel={undefined}
/>
```

fa7ad (approved):

src/app-layout/tools-and-split-panel.tsx
fa7ad: I like the idea of making the props explicit 👍

PR #575, opened 2022-12-14
fix: Ensure property filter prefers case-sensitive match

gethinwebster:
### Description

Ensure that property filter prefers an exact case-sensitive match of property key
over a non-exact match.

AWSUI-19954

### How has this been tested?

New unit test

<details>

#### Correctness

#### Security

#### Testing

</details>

---


just-boris:
Your call. A lot of things in property filter is revolving around assumption that data can be searched case-insensitive way.

If you are starting fixing this, it will not be the last one and I am not sure if you will hit a limit of non-breaking changes somewhere

timogasda (approved):
Agree with Genthin that this is the right direction.

PR #576, opened 2022-12-14
chore: Add flow telemetry for wizard

michaeldowseza:
### Description

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/internal/metrics.ts
```diff
@@ -102,6 +128,38 @@ export const Metrics = {
...
+    }
+
+    if (typeof metric.eventValue === 'object') {
+      metric.eventValue = JSON.stringify(metric.eventValue);
+    }
+
+    const panorama = findPanorama(window);
+    // TODO: Add console name to event context
+    if (typeof panorama === 'function') {
+      panorama(metricName, {
```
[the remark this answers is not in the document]
michaeldowseza: Thanks for the call out. Will bring it up with the team 👍🏻

src/internal/metrics.ts
```diff
@@ -102,6 +128,37 @@ export const Metrics = {
     }
   },
 
+  /**
+   * Calls Console Platform's client v2 logging JS API with provided metric name and detail.
+   * Does nothing if Console Platform client logging JS is not present in page.
+   */
+  sendMetricV2(metricName: string, metric: MetricsV2EventItem): void {
```
michaeldowseza, line 135: Thoughts on appropriate error handling for this? Wrapping the entire function in a try catch to silently fail as a catch all scenario vs letting the errors bubble up.

just-boris: Depends on guarantees from receiving end. `AWSC.Clog.log` API is asynchronous inside, so it always succeeded (because it only put message in a queue, sent with intervals).

If their sync logic is also rather small and they do not throw any validation errors, we can keep it without try/catch. 

But this is only true if we replace `window.panorama` with something else. We do not want our components crash via `window.panorama = {}`, throwing `TypeError: panorama is not a function`

just-boris: P.S. another comment to the same line – can we call it `sendPanoramaMetric`? Nouns are easier to recall when reading unfamiliar code, than version numbers

just-boris: Event detail does not contain any user input (and never should, you can ask me in DM why). If stringily fails, it is a bug on our end, should be captured with tests

just-boris (approved):

just-boris (approved):

PR #582, opened 2022-12-17
refactor: Move size observers into split panel code

just-boris:
### Description

Move resize observers down from app layout to split panel code

### How has this been tested?

Locally, PR build, dev pipeline

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/utils/use-observed-element.ts
+++ b/src/app-layout/utils/use-observed-element.ts
@@ -3,18 +3,10 @@
 import { useResizeObserver } from '../../internal/hooks/container-queries';
 import { useCallback, useState } from 'react';
 
-export function useObservedElement(
-  selectorOrRef: string | React.MutableRefObject<HTMLElement> | (() => HTMLElement | null) | undefined
-) {
+export function useObservedElement(selector: string) {
   const getElement = useCallback(() => {
-    if (typeof selectorOrRef === 'string') {
-      return document.querySelector(selectorOrRef);
-    } else if (typeof selectorOrRef === 'function') {
-      return selectorOrRef() ?? null;
-    } else {
-      return selectorOrRef?.current ?? null;
-    }
-  }, [selectorOrRef]);
+    return document.querySelector(selector);
+  }, [selector]);
 
   const [height, setHeight] = useState(0);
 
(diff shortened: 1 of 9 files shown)


src/app-layout/visual-refresh/layout.tsx
```diff
@@ -63,37 +56,6 @@ export default function Layout({ children }: LayoutProps) {
...
-    },
-    [
-      footerHeight,
-      getSplitPanelHeader,
-      isSplitPanelOpen,
-      setOffsetBottom,
-      splitPanelPosition,
-      splitPanel,
-      splitPanelSize,
-    ]
```
just-boris, line 94: This effect exclusively depends on values coming from app layout context. Why calculating it here with an effect, if the same value can be pre-computed on `context.tsx` file already?

src/internal/context/split-panel-context.ts
```diff
@@ -45,17 +43,15 @@ export const SplitPanelContext = createContext<SplitPanelContextProps>({
...
   isMobile: false,
   isForcedPosition: false,
   lastInteraction: undefined,
-  splitPanelRef: undefined,
-  splitPanelHeaderRef: undefined,
   onResize: () => {},
   onToggle: () => {},
   onPreferencesChange: () => {},
   reportSize: () => {},
+  reportHeaderHeight: () => {},
```
just-boris, line 54: Duplication was here. `reportSize` provided into about split panel size, but also refs to do the same.

Let's use only the `reportSize` pattern, plus add `reportHeaderHeight` to do the same for header

src/app-layout/utils/use-observed-element.ts
```diff
@@ -3,18 +3,10 @@
...
-) {
+export function useObservedElement(selector: string) {
   const getElement = useCallback(() => {
-    if (typeof selectorOrRef === 'string') {
-      return document.querySelector(selectorOrRef);
-    } else if (typeof selectorOrRef === 'function') {
-      return selectorOrRef() ?? null;
-    } else {
-      return selectorOrRef?.current ?? null;
-    }
```
just-boris, line 16: These features are not used anymore. `useContainerQuery` is preferred API, so no point to keep hacky API here

gethinwebster (approved):
nice simplification

PR #587, opened 2022-12-19
feat: Update Flashbar stacking feature to latest requirements

jperals:
### Description

Updates stacked behavior as well as internal API as reviewed and documented.

https://user-images.githubusercontent.com/1257272/208540145-80959a5a-5234-4b96-9979-f07cdaad31ed.mov

### How has this been tested?

- Added unit tests for `getStackedItems`
- Manually tested

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/__tests__/required-props-for-components.ts
```diff
@@ -133,6 +133,9 @@ const defaultProps: Record<string, Record<string, any>> = {
   'area-chart': {
     series: [],
   },
+  flashbar: {
```
jperals, line 136: This is necessary to feed the component with dummy a `items` array in unit tests which started failing after I added logic which relies on the prop existing. Since the prop is mandatory, it's better to update this here than to make the component code permissive.

src/app-layout/notifications/styles.scss
```diff
@@ -7,6 +7,6 @@
 
 .notifications-sticky {
   top: 0;
-  z-index: 825;
+  z-index: 900;
```
jperals, line 10: I looked into all existing z-indices in the repo as far as I can tell. 900 sets the notifications above the split panel (which has 851) but still below the navigation elements (1000 and beyond).

src/flashbar/motion.scss
```diff
@@ -5,7 +5,7 @@
 
 @use '../internal/styles' as styles;
 @use '../internal/styles/tokens' as tokens;
-.flash-refresh {
+.flash-with-motion {
```
jperals, line 8: Renaming because animated notifications now re not exclusive to Visual Refresh: also Classic gets animations if using the stacking feature.

src/app-layout/notifications/styles.scss
jperals: Reverted these z-index changes because they break the (very subtle) shadows casted by the side nav and toolbars on the flashbar in Classic. We'll handle later separately how to show the flashbar on top of the split panel, which is not something specific of the stacking feature.

PR #605, opened 2023-01-03
chore: use weaker types for inline editing

fa7ad:
### Description

This PR changes some the types used by Inline Editing to be weaker. The stronger types seem to break some consuming packages. 

Includes changes proposed by @just-boris in #595 

### How has this been tested?
Dry run build 6282595099 and 6284939920

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/table/tall-rows.page.tsx
+++ b/pages/table/tall-rows.page.tsx
@@ -16,13 +16,13 @@ export default function () {
 
   const tallItemsConfig: TableProps.ColumnDefinition<Instance>[] = columnsConfig.map((config, index) => ({
     ...config,
-    cell: (item: Instance) =>
+    cell: (item, ...rest) =>
       index === 0 ? (
         <Link onFollow={() => setClicks(prev => prev + 1)}>{item.id}</Link>
       ) : (
         <ul>
           {range(0, 20).map(index => {
-            return <li key={index}>{(config.cell as TableProps.CellRenderer<Instance>)(item)}</li>;
+            return <li key={index}>{config.cell(item, ...rest)}</li>;
           })}
         </ul>
       ),
--- a/src/table/__tests__/inline-editing-test-utils.test.tsx
+++ b/src/table/__tests__/inline-editing-test-utils.test.tsx
@@ -16,10 +16,10 @@ const defaultItems: Item[] = [
   { id: 3, name: 'Bananas' },
 ];
 
-const editableColumns: TableProps.ColumnDefinition<Item, string | number>[] = [
+const editableColumns: TableProps.ColumnDefinition<Item>[] = [
   {
     header: 'id',
-    cell: (item: Item, { isEditing }) => {
+    cell: (item, { isEditing }) => {
       if (isEditing) {
         return <input type="number" value={item.id} readOnly={true} data-testid={`id-editing-${item.id}`} />;
       }
@@ -33,7 +33,7 @@ const editableColumns: TableProps.ColumnDefinition<Item, string | number>[] = [
   },
   {
     header: 'name',
-    cell: (item: Item, { isEditing }) => {
+    cell: (item, { isEditing }) => {
       if (isEditing) {
         return <input type="number" value={item.name} readOnly={true} data-testid={`name-editing-${item.name}`} />;
       }
@@ -46,6 +46,10 @@ const editableColumns: TableProps.ColumnDefinition<Item, string | number>[] = [
     },
   },
 ];
+
+const defaultAriaLabels: TableProps.AriaLabels<{ id: number }> = {
+  tableLabel: 'Distributions',
+};
 
 function renderTable(jsx: React.ReactElement) {
   const { container, rerender, getByTestId, queryByTestId } = render(jsx);
@@ -71,5 +75,19 @@ describe('Editable Table TestUtils', () => {
     const buttons = getByTestId('id-editing-1').closest('form')!.querySelectorAll('button')!;
     expect(wrapper.findEditingCellCancelButton()!.getElement()!).toBe(buttons[0]);
     expect(wrapper.findEditingCellSaveButton()!.getElement()!).toBe(buttons[1]);
+  });
+
+  test('renders when items and labels types do not match', () => {
+    const ariaLabels: TableProps.AriaLabels<Item> = {
+      ...defaultAriaLabels,
+      submitEditLabel: column => `Submit edit for ${column.header}`,
+      cancelEditLabel: column => `Cancel edit for ${column.header}`,
+      activateEditLabel: column => `Activate edit for ${column.header}`,
+    };
+
+    const { wrapper } = renderTable(
+      <Table<Item> columnDefinitions={editableColumns} items={defaultItems} ariaLabels={ariaLabels} />
+    );
+    expect(wrapper.findBodyCell(2, 2)!.getElement()!).toBeInTheDocument();
   });
 });
--- a/src/table/interfaces.tsx
+++ b/src/table/interfaces.tsx
@@ -164,7 +164,7 @@ export interface TableProps<T = any, V = any> extends BaseComponentProps {
    * * `submitEditLabel` (EditableColumnDefinition) => string -
    *                      Specifies an alternative text for the submit button in editable cells.
    */
-  ariaLabels?: TableProps.AriaLabels<T, V>;
+  ariaLabels?: TableProps.AriaLabels<T>;
 
   /**
    * Specifies the definition object of the currently sorted column. Make sure you pass an object that's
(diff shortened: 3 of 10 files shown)


pages/table/tall-rows.page.tsx
```diff
@@ -22,7 +22,7 @@ export default function () {
       ) : (
         <ul>
           {range(0, 20).map(index => {
-            return <li key={index}>{(config.cell as TableProps.CellRenderer<Instance>)(item)}</li>;
+            return <li key={index}>{config.cell(item)}</li>;
```
fa7ad, line 25: Typecast no longer needed for ambiguous calls (previously typescript wasn't be able to infer the number of arguments)

src/table/__tests__/inline-editing-test-utils.test.tsx
```diff
@@ -16,24 +16,24 @@ const defaultItems: Item[] = [
...
+      if (isEditing) {
+        return <input type="number" value={item.id} readOnly={true} data-testid={`id-editing-${item.id}`} />;
+      }
+      return item.id;
+    },
   },
   {
     header: 'name',
-    cell: (item: Item, { isEditing }) => {
+    cell: (item: Item, { isEditing }: TableProps.CellContext<string>) => {
```
just-boris, line 36: Why is this explicit type needed? Is it possible to make it coming by default?

If needed, we can provide an empty `context` object for non-editable cells, if we need fixed arity between to modes

just-boris:
This also needs to have a test (something like rendering an example from description in my PR #595) and ensuring it compiles with typescript. Easiest option, add an extra unit test [to this file](https://github.com/cloudscape-design/components/blob/inline-editing-rerevert/src/table/__tests__/inline-editing-test-utils.test.tsx)

connorlanigan (changes requested):

src/table/__tests__/inline-editing-test-utils.test.tsx
fa7ad: Strictly speaking, the explicit type is not needed in this particular usage. I'll remove it.

src/table/interfaces.tsx
```diff
@@ -321,25 +321,20 @@ export namespace TableProps {
...
-      }
-    | {
-        editConfig: EditConfig<ItemType, ValueType>;
-        cell: EditableCellRenderer<ItemType, ValueType>;
-      };
-
-  export type ColumnDefinition<ItemType, ValueType = any> = {
+  export interface ColumnEditConfig<ItemType> {
+    editConfig?: EditConfig<ItemType>;
+    cell(item: ItemType, context?: CellContext<any>): React.ReactNode;
```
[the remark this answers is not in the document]
fa7ad: I'll update the example (editable-with-app-layout) to use correct types, but I'm not sure there's much we can do to ensure type consistency across the two fields.
I'm open to suggestions if you think theres a way to prevent the scenario you described from happening.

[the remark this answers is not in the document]
fa7ad: Following our discussion offline, I've updated the types for the submit function to use `any` for the type for `newValue`.

connorlanigan (approved):

src/table/interfaces.tsx
```diff
@@ -321,25 +318,20 @@ export namespace TableProps {
...
```
just-boris, line 323: I do not like that now there is optional chaining like `context?.isEditing` all over the place. Should we provide an empty context object for all cells?

src/table/interfaces.tsx
```diff
@@ -281,17 +281,14 @@ export interface TableProps<T = any, V = any> extends BaseComponentProps {
 export namespace TableProps {
   export type TrackBy<T> = string | ((item: T) => string);
 
-  export type EditableColumnDefinition<T, V = any> = Extract<ColumnDefinition<T, V>, ColumnEditConfig<T, V>>;
+  export type EditableColumnDefinition<T> = Extract<ColumnDefinition<T>, ColumnEditConfig<T>>;
```
just-boris, line 284: Is this one used anywhere in public API? If not, should be removed from `TableProps` namespace

src/table/interfaces.tsx
```diff
@@ -321,25 +318,20 @@ export namespace TableProps {
...
-        editConfig?: undefined;
-        cell: CellRenderer<ItemType>;
-      }
-    | {
-        editConfig: EditConfig<ItemType, ValueType>;
-        cell: EditableCellRenderer<ItemType, ValueType>;
-      };
-
-  export type ColumnDefinition<ItemType, ValueType = any> = {
+  export interface ColumnEditConfig<ItemType> {
```
just-boris, line 321: Any reason to keep it as a separate interface and not merge into `ColumnDefinition`?

src/table/interfaces.tsx
```diff
@@ -406,9 +400,9 @@ export namespace TableProps {
     cancelEdit?(): void;
   }
 
-  export type SubmitEditFunction<ItemType, ValueType = any> = (
+  export type SubmitEditFunction<ItemType> = (
```
just-boris, line 403: Having an extra generic parameter allows nicer code on consumer's end

```js
const handleSubmit: TableProps.SubmitEditFunction<Item, string> = (item, column, newValue) => {
  // ... code
}
```

Which is nicer than

```js
const handleSubmit: TableProps.SubmitEditFunction<Item> = (item, column, newValue: string) => {
  // ... code
}
```

Looks odd when you annotate one argument and not others

just-boris: P.S. we might make the default generic type `unknown` which will enforce to give the annotation

fa7ad: @connorlanigan suggested going with `unknown` as well, and I agree that its better than `any` in this case. I'll publish a new revision soon

src/table/interfaces.tsx
fa7ad: No reason, other than the extracted type. But if we remove that we can combine this with ColumnDefinition

src/table/interfaces.tsx
fa7ad: Its used internally, I'll removing it from the namespace; maybe even completely

src/table/interfaces.tsx
fa7ad: We do provide a readOnly context value even for non-editable cells. But the problem with requiring this value is packages downstream might not be providing this value. I guess the question is, Do we want to break those implementations?

connorlanigan: Since this is a parameter (i.e. a value _we_ provide rather than a value _consumers_ provide), I think it won't cause errors for consumers who don't use the parameter.

See https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgJKQLYBUCeAHFAbwF8BYAKFElkRQGEIAbRuge3AgA8wAeLAPhIUKYfPSaMAYgFcQCMMHYB1YGAAWAeTwL2cRsgC8yABSqIGAFxpMuAgBpkCdpG4B+Kw2ZsO3HnBA4-ACUhvzIAG6swAAmANwiYsieUrLyiiAq6gBKEACO0sBQENGGJmaW1ua2EA5OPmAeEt4uvP6BIQZhkTHx5BR1AM5gyOwQjcwycjoZqpra6XqlppgdYbLREDCgxb2Dw2AA7qzjKVPpmWo5+YXFS+WryOub23FAA

fa7ad: My comment wasn't clear, sorry about that. What I meant to say is that some of our consumers intercept this call like this; this then breaks the type signature.

image

just-boris: Namely, this code: https://github.com/awslabs/iot-app-kit/blob/main/packages/related-table/src/RelatedTable/RelatedTable.tsx#L22

Discussed in DM, this use is not matching our API contract. `cell` is only supposed to be called by us, and not anyone outside. As a quick fix, we will change their code to `columnToExpand?.cell(node, ...rest)`

src/table/interfaces.tsx
connorlanigan: >Having an extra generic parameter allows nicer code on consumer's end

It does, but we currently can't enforce that the value we provide will match the generic parameter customers specify, see the comment thread above: https://github.com/cloudscape-design/components/pull/605#discussion_r1060726179

So `unknown` will force consumers to be aware that they have to perform either a type assertion or a type check before using it.

connorlanigan (approved):

src/table/interfaces.tsx
just-boris: This is why I proposed `type SubmitEditFunction<ItemType, ValueType= unknown> = (newValue: ValueType) => void;` and not `(newValue: unknown)` directly. Moving it into generic allows to specify the new type

[TS playground demo](https://www.typescriptlang.org/play?#code/C4TwDgpgBAygrgIwLYEtgFEAmaBicB2AxsCgPb4A8AagIYA2cEAKuBALxQEDW+pA7vgB8UDgAp8EPrQYQAXFGmMWkAJQjhAN1IpMAbgCwAKCNHC5AM7Ao5xKmABGefGRosuAsTKVLAJxT4Ac2ExCSl6RjU2YQBvIyh4qFDFCAA6YFIYYD9A0RVdKAB6Aqg+Uh8ucyMAXxNDM3xLa1s0ACYnZoxsYDwiEnIRKHFJZMiYuISk8NT0zOyA3PyiqB8IVHxMCB8oGgRSOCtUc3N-AKhQSGrapcxSCHNE0itS8qgECEIaOHNoUgAzbaOKAC+B2KDoaBAyzgdDupgsVhsLmAAGZ2ki3N0PH18AMhmEZPJfCdRlBYoYEolhlM0hksicFtUgA)

just-boris: ^ @fa7ad this is the last comment/request. Can approve after this change

just-boris (approved):

PR #607, opened 2023-01-03
chore: Support Visual Refresh theme in metrics

timogasda:
### Description
We want to identify the "runtime" Visual Refresh theme in the metrics, which is currently taken from the `THEME` environment variable and ignores the VR context. This change re-introduces the forgotten `initMetrics` function that overrides the metrics theme for VR.

Related links, issue #, if available: AWSUI-20010

### How has this been tested?
I have tested this locally.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/app/index.tsx
+++ b/pages/app/index.tsx
@@ -3,8 +3,8 @@
 import React, { Suspense, useContext, useEffect } from 'react';
 import { render } from 'react-dom';
 import { HashRouter, Redirect } from 'react-router-dom';
+import { createHashHistory } from 'history';
 import { applyMode, applyDensity, disableMotion } from '@cloudscape-design/global-styles';
-import { useEffectOnUpdate } from '~components/internal/hooks/use-effect-on-update';
 import './polyfills';
 
 // import font-size reset and Ember font
@@ -16,18 +16,14 @@ import PageView from './components/page-view';
 import IndexPage from './components/index-page';
 import Header from './components/header';
 import StrictModeWrapper from './components/strict-mode-wrapper';
-import AppContext, { AppContextProvider } from './app-context';
+import AppContext, { AppContextProvider, parseQuery } from './app-context';
 
 function App() {
   const {
     mode,
     pageId,
-    urlParams: { density, motionDisabled, visualRefresh },
+    urlParams: { density, motionDisabled },
   } = useContext(AppContext);
-
-  useEffectOnUpdate(() => {
-    window.location.reload();
-  }, [visualRefresh]);
 
   const isAppLayout =
     pageId !== undefined && (pageId.indexOf('app-layout') > -1 || pageId.indexOf('content-layout') > -1);
@@ -71,9 +67,11 @@ function App() {
   );
 }
 
-// The VR class needs to be set before any React rendering takes occurs.
-const vrDisabled = window.location.hash.includes('visualRefresh=false');
-document.getElementById('app')?.classList.toggle('awsui-visual-refresh', !vrDisabled);
+const history = createHashHistory();
+const { visualRefresh } = parseQuery(history.location.search);
+
+// The VR class needs to be set before any React rendering occurs.
+document.body.classList.toggle('awsui-visual-refresh', visualRefresh);
 
 render(
   <HashRouter>
--- a/pages/app/index.tsx
+++ b/pages/app/index.tsx
@@ -56,15 +56,6 @@ function App() {
     }
   }, [isMacOS]);
 
-  useEffect(() => {
-    // NB: do not use classList.toggle, it does not work as expected in IE
-    if (visualRefresh) {
-      document.body.classList.add('awsui-visual-refresh');
-    } else {
-      document.body.classList.remove('awsui-visual-refresh');
-    }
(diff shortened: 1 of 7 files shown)


src/internal/utils/__tests__/metrics.test.ts
```diff
@@ -325,11 +332,11 @@ describe('Client Metrics support', () => {
     test('logs the component loaded metric', () => {
       defineClog();
       Metrics.logComponentLoaded();
-      checkMetric(`awsui_components_d30`, ['main', 'components', 'default', 'loaded', 'DummyFrameWork', '3.0(HEAD)']);
+      checkMetric(`awsui_components_d30`, ['main', 'components', 'default', 'loaded', 'react', '3.0(HEAD)']);
     });
   });
 
-  describe.only('sendPanoramaMetric', () => {
```
timogasda, line 332: oops, somebody forgot this 🙈

src/internal/metrics.ts
```diff
@@ -98,22 +107,21 @@ const findAWSC = (currentWindow?: MetricsWindow): AWSC | undefined => {
...
   },
 
   /**
    * Calls Console Platform's client logging JS API with provided metric name, value, and detail.
    * Does nothing if Console Platform client logging JS is not present in page.
    */
   sendMetric(metricName: string, value: number, detail?: string): void {
+    if (!theme) {
+      // Metrics need to be initialized first (initMetrics)
+      return;
```
just-boris, line 122: The codecov is right, there needs to be a test for it

just-boris (commented):
all good except missing tests

1.  No metrics before initialization
2. Metric can send a custom theme

just-boris (approved):
There is `MetricsTestHelper` export intended to help testing one-time things. 

But the current 2-file approach is fine too

pages/app/index.tsx
```diff
@@ -80,6 +71,10 @@ function App() {
   );
 }
 
+// The VR class needs to be set before any React rendering takes occurs.
```
timogasda, line 74: Something very interesting happens when the `useVisualRefresh` hook gets called much earlier, i.e., during metrics. VR is only applied half-way through rendering and is causing inconsistent results. That's why the visual regression tests failed.

Curious how you feel about this change, @just-boris @rubencarvalho

just-boris: This is expected, because we do not follow our own visual refresh activation guide. Specifically, we run it inside `useEffect`, where the guideline has not to do so:

https://github.com/cloudscape-design/components/blob/07e7e371d5bb6560f5f0d24f4e7d1664fb26f856/pages/app/index.tsx#L59-L66

> The theme activates only on the first render. **Do NOT use React hooks** or lifecycle methods or JSX content for this class name, because it will not work correctly

Your addition makes sense, but some suggestions

1. Replace `document.getElementById('app')?` with `document.getElementById('app')!`, because if this code is missing we want to fail loud, not quiet 
2. Remove `useEffect` code, because it is not needed
3. Use hash history instead of manual parsing

```js
import { createHashHistory } from "history";

// use history.location.search to parse the parameters, reusing with this code for consistency https://github.com/cloudscape-design/components/blob/07e7e371d5bb6560f5f0d24f4e7d1664fb26f856/pages/app/app-context.tsx#L40

<Router history={history}>...</Router>
```

just-boris: There is no such thing as automatic referesh, because the page reloads anyway:  https://github.com/cloudscape-design/components/blob/07e7e371d5bb6560f5f0d24f4e7d1664fb26f856/pages/app/index.tsx#L28-L30

This code could get some refactoring and simplification

pages/app/index.tsx
```diff
@@ -80,6 +71,10 @@ function App() {
...
+const vrDisabled = window.location.hash.includes('visualRefresh=false');
```
rubencarvalho, line 75: Maybe I'm missing something, but do we need to parse the hash instead of using the `visualRefresh` value from the `AppContext`?

pages/app/index.tsx
```diff
@@ -80,6 +71,10 @@ function App() {
...
+document.getElementById('app')?.classList.toggle('awsui-visual-refresh', !vrDisabled);
```
rubencarvalho, line 76: I also see the benefit of not using the optional chaining here.

src/internal/metrics.ts
rubencarvalho, line 122: We could also `console.error` this.

just-boris: If our code works correctly, this condition should never be visited. If any reporting, it should be throwing an error (fine as a future PR)

timogasda: All the other error cases in this function (and file, actually) use `console.error` and don't throw. So I would be in favor of another `console.error`.

pages/app/index.tsx
timogasda: Yup, will update this.

src/internal/metrics.ts
just-boris: ok

rubencarvalho (approved):

PR #616, opened 2023-01-05
chore: Remove react-resizable dependency

just-boris:
### Description

This library is unmaintained and not compatible with React 18

### How has this been tested?

* Locally, compared behaviors against stable release
* Added tests for most important behaviors

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/code-editor/resizable-box/index.tsx
+++ b/src/code-editor/resizable-box/index.tsx
@@ -1,0 +1,57 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import React, { useEffect, useRef, useState } from 'react';
+import { useStableEventHandler } from '../../internal/hooks/use-stable-event-handler';
+import styles from './styles.css.js';
+
+export interface ResizeBoxProps {
+  children: React.ReactNode;
+  height: number;
+  minHeight: number;
+  onResize: (newHeight: number) => void;
+}
+
+export function ResizableBox({ children, height, minHeight, onResize }: ResizeBoxProps) {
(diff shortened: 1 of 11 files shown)


src/code-editor/resizable-box/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+      const { top } = container.getBoundingClientRect();
+      const cursor = event.clientY;
+      onResizeStable(Math.max(cursor + dragOffset - top, minHeight));
+    };
+    const onMouseUp = () => {
+      setDragOffset(null);
+    };
+    document.body.classList.add(styles['resize-active']);
+    document.addEventListener('mousemove', onMouseMove);
+    document.addEventListener('mouseup', onMouseUp);
```
connorlanigan, line 43: Is this `mouseup` event also fired when the mouse button is released outside of the browser window / document? I've encountered implementations where the resizing becomes "sticky" to the mouse if you release the button outside. We might need to listen on `window` instead or use something like [`setPointerCapture()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/setPointerCapture)

just-boris: 1. I have copied this code from [resizable table columns implementation](https://github.com/cloudscape-design/components/blob/main/src/table/resizer/index.tsx). We have not received any related bug reports there. (Why not reusing the code between these two? Because resizing tables is much different than divs, very few lines to share)
2. Regardless of the above, tested locally, it stops the resize as expected

connorlanigan: Great, thank you for verifying!

connorlanigan (approved):

PR #625, opened 2023-01-09
refactor: Use app layout internals only inside app layout

just-boris:
### Description

There are two app layout contexts. One for use inside other components to calculate offsets (mostly for sticky elements). Second is intended to be used only inside app layout sub-components and only in visual refresh. Let's make it more strict

### How has this been tested?

* PR build
* Dev pipeline build

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/visual-refresh/context.tsx
+++ b/src/app-layout/visual-refresh/context.tsx
@@ -2,13 +2,13 @@
 // SPDX-License-Identifier: Apache-2.0
 import React, {
   createContext,
-  createRef,
   useCallback,
   useEffect,
   useLayoutEffect,
   useImperativeHandle,
   useRef,
   useState,
+  useContext,
 } from 'react';
 import { AppLayoutProps } from '../interfaces';
 import { fireNonCancelableEvent } from '../../internal/events';
@@ -23,8 +23,9 @@ import { warnOnce } from '../../internal/logging';
 import { applyDefaults } from '../defaults';
 import { FocusControlState, useFocusControl } from '../utils/use-focus-control';
 import { useObservedElement } from '../utils/use-observed-element';
+import { AppLayoutContext } from '../../internal/context/app-layout-context';
 
-interface AppLayoutContextProps extends AppLayoutProps {
+interface AppLayoutInternals extends AppLayoutProps {
   dynamicOverlapHeight: number;
   handleSplitPanelClick: () => void;
   handleNavigationClick: (isOpen: boolean) => void;
@@ -49,8 +50,6 @@ interface AppLayoutContextProps extends AppLayoutProps {
   offsetBottom: number;
   setDynamicOverlapHeight: (value: number) => void;
   setHasStickyBackground: (value: boolean) => void;
-  setIsNavigationOpen: (value: boolean) => void;
-  setIsToolsOpen: (value: boolean) => void;
   setSplitPanelReportedSize: (value: number) => void;
   setSplitPanelReportedHeaderHeight: (value: number) => void;
   headerHeight: number;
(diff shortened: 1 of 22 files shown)


src/app-layout/visual-refresh/context.tsx
```diff
@@ -63,81 +62,26 @@ interface AppLayoutContextProps extends AppLayoutProps {
   toolsFocusControl: FocusControlState;
 }
 
-// TODO simplify default params + typings
```
just-boris, line 66: Resolving this todo is the primary goal of this PR. Hooray

src/container/internal.tsx
```diff
@@ -65,24 +65,21 @@ export default function InternalContainer({
...
    * The visual refresh AppLayout component needs to know if a child component
-   * has a high constrast sticky header. This is to make sure the background element
+   * has a high contrast sticky header. This is to make sure the background element
    * stays in the same vertical position as the header content.
    */
-  useEffect(
-    function handleHasStickyBackground() {
-      const shouldUpdateStickyBackground = isRefresh && isSticky && variant === 'full-page';
+  useEffect(() => {
+    const shouldUpdateStickyBackground = isSticky && variant === 'full-page' && setHasStickyBackground;
```
just-boris, line 72: Instead of checking `isRefresh`, we can check if `setHasStickyBackground` was provided by the context (only exists in visual refresh). Same outcome, but more specific to the use-case

src/table/sticky-scrollbar.tsx
```diff
@@ -19,18 +17,15 @@ export default forwardRef(StickyScrollbar);
...
   const mergedRef = useMergeRefs(ref, scrollbarRef);
 
   /**
    * Use the appropriate AppLayout context (Classic or Visual Refresh) to determine
    * the offsetBottom value to be used in the useStickyScrollbar hook.
    */
-  const { stickyOffsetBottom: offsetBottomClassic } = useAppLayoutContext();
-  const { offsetBottom: offsetBottomVisualRefresh } = useContext(AppLayoutContext);
-  const offsetBottom = isRefresh ? offsetBottomVisualRefresh : offsetBottomClassic;
+  const { stickyOffsetBottom } = useAppLayoutContext();
```
just-boris, line 26: Good simplification here. Instead of reading two contexts and reconciling the value, we expect the correct value to be provided by the respective context

src/internal/hooks/use-dynamic-overlap/__tests__/use-dynamic-overlap.test.tsx
```diff
@@ -1,10 +1,10 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
just-boris, line 1: Moved this file from `src/app-layout/hooks` into `src/internal/hooks` because it is in fact a cross-component API

src/internal/hooks/use-dynamic-overlap/__tests__/use-dynamic-overlap.test.tsx
```diff
@@ -1,10 +1,10 @@
...
 // SPDX-License-Identifier: Apache-2.0
-import React from 'react';
+import React, { useState } from 'react';
 import { render, screen } from '@testing-library/react';
 
-import { useDynamicOverlap } from '../../../../../lib/components/app-layout/visual-refresh/hooks/use-dynamic-overlap';
-import { AppLayoutContext, AppLayoutProvider } from '../../../../../lib/components/app-layout/visual-refresh/context';
```
just-boris, line 7: Moved this file from `src/app-layout/hooks` into `src/internal/hooks` because it is in fact a cross-component API

Also look how it helped to remove dependency on app layout internals. It now uses only the package-public context

src/split-panel/index.tsx
```diff
@@ -147,6 +147,7 @@ export default function SplitPanel({
       value={{
         stickyOffsetTop: topOffset,
         stickyOffsetBottom: bottomOffset,
+        hasBreadcrumbs: false,
```
Al-Dani, line 150: I'm not sure I understand how this works. Do we set `hasBreadcrumbs` to false because split panel doesn't care about this property?

just-boris: Practically any value does not make sense here, because it is used only by ContentLayout component, which is not expected to be inside split panel.

I put `false` value here as a safer option (no value is better than assuming there is something)

PR #629, opened 2023-01-11
chore: Fix wizard rerendering causing multiple start step events

michaeldowseza:
### Description

Removed `trackStartStep` as a dependency of the effect hook. As the functions inside the `useWizardAnalytics` hook are recreated on render it causes this hook to trigger multiple times. The intention of the effect is to send an event only when the step index changes.

### How has this been tested?

A quick test in the dev page while logging the events while triggering a re-render in the parent component. As seen in the before multiple events are fire but in the after only a single event is fired.

Before:
Screenshot 2023-01-11 at 14 13 45

After:
Screenshot 2023-01-11 at 14 14 30

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/wizard/index.tsx
+++ b/src/wizard/index.tsx
@@ -1,6 +1,6 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
-import React, { useRef } from 'react';
+import React, { useEffect, useRef } from 'react';
 import clsx from 'clsx';
 import { getBaseProps } from '../internal/base-component';
 import { fireNonCancelableEvent } from '../internal/events';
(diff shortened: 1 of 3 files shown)


src/wizard/index.tsx
```diff
@@ -89,9 +88,10 @@ export default function Wizard({
     );
   }
 
-  useEffectOnUpdate(() => {
+  useEffect(() => {
```
pan-kot, line 91: Why replacing useEffectOnUpdate with useEffect?

just-boris:
How about adding a unit test for it?

src/wizard/index.tsx
michaeldowseza: useEffectOnUpdate skips the initial render as intended and so  then I never got a metric on the 1st render 😅

src/wizard/index.tsx
```diff
@@ -89,9 +88,10 @@ export default function Wizard({
...
     trackStartStep(actualActiveStepIndex);
-  }, [actualActiveStepIndex, trackStartStep]);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
just-boris, line 93: This disable can be avoided. `useWizardAnalytics` is not a hook, because it does not contain any first-class react hooks. These functions can be imported directly

```
import { trackStartStep } from './internal/analytics';
```

and this will make linter happy without workarounds

michaeldowseza: Agreed. I think I had bigger plans to have it in a neat single hook but you're right that it's better to just import the functions. Will change in next commit.

michaeldowseza:
> How about adding a unit test for it?

Added.

just-boris (approved):

PR #635, opened 2023-01-12
feat: Make split panel header scrollable when page height is small

YueyingLu:
### Description

Split panel header should not be fixed when page height is constraint.
Split panel resizer should still be fixed at the top of the panel.

Related links, issue AWSUI-20054

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/index.tsx
+++ b/src/app-layout/index.tsx
@@ -326,6 +326,13 @@ const OldAppLayout = React.forwardRef(
       onPreferencesChange: onSplitPanelPreferencesSet,
       reportSize: setSplitPanelReportedSize,
       reportHeaderHeight: setSplitPanelReportedHeaderHeight,
+      isStickyHeader: () => {
+        const height =
+          disableBodyScroll && legacyScrollRootRef.current
+            ? legacyScrollRootRef.current.clientHeight
+            : document.documentElement.clientHeight - headerHeight - footerHeight;
+        return height > CONSTRAINED_PAGE_HEIGHT;
+      },
     };
     const splitPanelWrapped = splitPanel && (
       <SplitPanelContextProvider value={splitPanelContext}>{splitPanel}</SplitPanelContextProvider>
--- a/src/split-panel/styles.scss
+++ b/src/split-panel/styles.scss
@@ -160,6 +160,7 @@ $app-layout-drawer-width: calc(#{awsui.$space-layout-toggle-diameter} + 2 * #{aw
   display: flex;
   justify-content: center;
   z-index: 2;
+  background-color: awsui.$color-background-layout-panel-content;
 }
 
 .slider-wrapper-side {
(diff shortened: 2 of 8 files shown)


avinashbot (changes requested):
Screen Shot 2023-01-12 at 5 20 36 PM

Code looks good to me, but the approach might need a slight refactoring because the slider background is overlapping the scrollbar...

YueyingLu:
Expected visual change of the position the resizer when scrollbar is shown. Because before the change, the width of the resizer line includes the width of the scrollbar, but now the width is the actual width of content without scrollbar. 
Screenshot 2023-01-16 at 16 00 40

avinashbot (approved):
LGTM (looks good to me), LGTM (let's get this merged)

src/split-panel/styles.scss
```diff
@@ -198,14 +199,16 @@ $app-layout-drawer-width: calc(#{awsui.$space-layout-toggle-diameter} + 2 * #{aw
...
-  /* stylelint-disable-next-line plugin/no-unsupported-browser-features */
-  position: sticky;
-  top: 0;
   display: flex;
   align-items: center;
   flex-direction: column;
   z-index: 1; // should be above .content-bottom
   padding: 0 awsui.$space-scaled-2x-xxxl;
+  &.sticky-header {
+    /* stylelint-disable-next-line plugin/no-unsupported-browser-features */
```
just-boris, line 208: This now works in the entire range of all browsers we support. Might need a follow-up to upgrade the rule

src/split-panel/styles.scss
```diff
@@ -215,6 +218,7 @@ $app-layout-drawer-width: calc(#{awsui.$space-layout-toggle-diameter} + 2 * #{aw
   :not(.drawer-closed) > .drawer-content-bottom > & {
     background-color: awsui.$color-background-layout-panel-content;
     border-bottom: awsui.$border-panel-header-width solid awsui.$color-border-divider-default;
+    margin-top: -$slider-width - 2px; // the height of the slider
```
just-boris, line 221: > $slider-width
> // the height of the slider

Something does not match up here

src/app-layout/index.tsx
```diff
@@ -326,6 +326,9 @@ const OldAppLayout = React.forwardRef(
       onPreferencesChange: onSplitPanelPreferencesSet,
       reportSize: setSplitPanelReportedSize,
       reportHeaderHeight: setSplitPanelReportedHeaderHeight,
+      isStickyHeader: () => {
```
just-boris, line 329: `isStickyHeader` name means that there is a sticky header in app layout somewhere and split panel should adapt to that. In fact this is the split panel's header, so more correct name would be `headerShouldStick`

just-boris (commented):
saw that you added and removed support for `disableBodyScroll`. Even if it is deprecated, it is still used for some customers. What happens when it is set to true? Does it cause any regression or only does not activate the improvement?

YueyingLu:
Hi @just-boris It was added and removed because I didn't notice `disableBodyScroll` is deprecated at beginning. This change is an improvement on small screen. I think it is nicer to avoid using deprecated prop as possible and the worst thing can happen to users still using `disableBodyScroll` is the sticky header of split panel is not de-activated at the expected page height.

src/split-panel/styles.scss
YueyingLu: $slider-width is used to define the height of the slide when it is bottom split panel. I will update the comment to make it more clear. 
https://github.com/cloudscape-design/components/blob/6fd6fd1a8f92fc6a941ef9aba79bdc380eda12f4/src/split-panel/styles.scss#L159

just-boris:
ok, sounds good for `disableBodyScroll`. No improvement, no regression. Let's do the other comments and I will approve

src/split-panel/styles.scss
just-boris: I think, there is something wrong with `$slider-width` if it is used in both vertical and horizontal directions, it should be called `$slider-size`, but ok, it is for another time.

just-boris (approved):

PR #640, opened 2023-01-13
fix: Dropdown status not announced by screenreader using expandToViewport

YueyingLu:
### Description
It is a follow up fix of https://github.com/cloudscape-design/components/pull/608. 
When using `expandToViewport` the dropdown status rendered when dropdown opens. The aria-live div and its content comes in the same time, so no real content change to get announced. The delay in LiveRegion helps the content in aria-live does get changed and announced. 

Related links, issue AWSUI-19995

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/components/dropdown-footer/__tests__/dropdown-footer.test.tsx
+++ b/src/internal/components/dropdown-footer/__tests__/dropdown-footer.test.tsx
@@ -31,8 +31,8 @@ describe('Dropdown footer', () => {
 
   test('adds correct aria attributes', () => {
     const { wrapper } = renderComponent(<DropdownFooter />);
-    const element = wrapper.find('div')!.getElement();
-    expect(element).toHaveAttribute('aria-live', 'polite');
-    expect(element).toHaveAttribute('aria-atomic', 'true');
+    const element = wrapper.find('span')!.getElement();
+    expect(element.firstElementChild).toHaveAttribute('aria-live', 'polite');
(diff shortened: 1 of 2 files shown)


src/internal/components/dropdown-footer/__tests__/dropdown-footer.test.tsx
```diff
@@ -31,8 +31,8 @@ describe('Dropdown footer', () => {
 
   test('adds correct aria attributes', () => {
     const { wrapper } = renderComponent(<DropdownFooter />);
-    const element = wrapper.find('div')!.getElement();
-    expect(element).toHaveAttribute('aria-live', 'polite');
-    expect(element).toHaveAttribute('aria-atomic', 'true');
+    const element = wrapper.find('span')!.getElement();
+    expect(element.firstElementChild).toHaveAttribute('aria-live', 'polite');
+    expect(element.firstElementChild).toHaveAttribute('aria-atomic', 'true');
```
abdhalees, line 36: I don't mind having these tests, but how useful are they since we're using the shared component? we don't have this test for other components that uses `LiveRegion`
however we have test to check if the text is being updated

YueyingLu: The purpose of this test is not testing the structure of `LiveRegion` but to check if `aria-live` being used on the footer. No matter it is from `LiveRegion` or not.

abdhalees (approved):

PR #642, opened 2023-01-13
feat: Update collapsible Flashbar feature to latest design

jperals:
### Description

Updates the collapsible Flashbar according to latest design specs:
- Changes display order of notifications (newest on top)
- Adds animations
- Introduces various changes to the expand/collapse button
- Adresses a11y requirements
- Updates the API

Also fixes a small bug with the Flash focus ring not looking as expected.

This does not make the API for this feature publicly discoverable yet.

Quick video demo below:

https://user-images.githubusercontent.com/1257272/213673490-9774c7c6-5df5-4b26-8799-2ee28c116d72.mov

### How has this been tested?

- Added a good number of unit tests
- Added integration tests for focus management and keyboard navigation
- Added visual regression tests
- Manually tested with VoiceOver (Mac) and NVDA (Windows)

_Note:_ the _Visual Regression Tests_ [task](https://github.com/cloudscape-design/components/actions/runs/3966812661/jobs/6798131492) yields two failures because the added visual tests do not exist in `main`:

```
compare | Reference image not found permutations_flashbar_collapsiblepermutations_0_document_0_phone.png
compare | Reference image not found permutations_flashbar_collapsiblepermutations_0_document_1_desktop.png
```

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/flashbar/collapsible-flashbar.tsx
```diff
@@ -0,0 +1,392 @@
...
+            ref={toggleElementRef}
+          >
+            <span className={styles.status} role="status">
+              {toggleButtonText && <h2 className={styles.text}>{toggleButtonText}</h2>}
+              <span className={styles['types-count']} id={itemCountElementId}>
+                <LiveRegion>
+                  {counterTypes
+                    .map(
+                      ({ type, labelName }: { type: FlashType; labelName: LabelName }) =>
+                        `${countByType[type]} ${(i18nStrings && i18nStrings[labelName]) ?? ''}`
```
timogasda, line 322: I think using this format makes assumptions about localization, i.e. that all locales put the number first and then the subject. There's two alternatives I can think of:

1) Format it with a colon like: `${countByType[type]}: ${i18nStrings[labelName]}`
2) Turn `labelName` into a function that customers can localize properly.

src/flashbar/collapsible-flashbar.tsx
```diff
@@ -0,0 +1,392 @@
...
+                      key={type}
+                      iconName={iconName}
+                      label={i18nStrings ? i18nStrings[labelName] : undefined}
+                      count={countByType[type]}
+                    />
+                  ))}
+                </span>
+              </span>
+            </span>
+            <button
```
timogasda, line 338: Can we use `<Button variant="icon">` here?

src/flashbar/collapsible-flashbar.tsx
```diff
@@ -0,0 +1,392 @@
...
+              <span className={styles['types-count']} id={itemCountElementId}>
+                <LiveRegion>
+                  {counterTypes
+                    .map(
+                      ({ type, labelName }: { type: FlashType; labelName: LabelName }) =>
+                        `${countByType[type]} ${(i18nStrings && i18nStrings[labelName]) ?? ''}`
+                    )
+                    .join(', ')}
+                </LiveRegion>
+                <span aria-hidden="true" className={styles['types-count-visible']}>
```
timogasda, line 326: Why are we hiding these?

src/flashbar/collapsible-flashbar.tsx
```diff
@@ -0,0 +1,392 @@
...
+            </button>
+          </div>
+        )}
+        {!isFlashbarStackExpanded && renderList()}
+      </>
+    </div>
+  );
+}
+
+type LabelName =
```
timogasda, line 356: I would consider moving this code below into separate files.

src/flashbar/collapsible.scss
```diff
@@ -0,0 +1,229 @@
...
+      #{awsui.$font-body-m-line-height} + (
+          #{awsui.$space-scaled-xs} + #{awsui.$border-field-width} + #{awsui.$space-scaled-xxs}
+        ) * 2
+    );
+  }
+
+  &.visual-refresh {
+    > .item:not(:last-child) > .flash,
+    > .item.flash {
+      box-shadow: 0px 4px 4px rgba(0, 0, 0, 0.25);
```
timogasda, line 78: We usually want to turn these magic numbers into proper design tokens. Applies for other occurrences below as well.

src/flashbar/collapsible.scss
```diff
@@ -0,0 +1,229 @@
...
+
+    > .types-count {
+      margin-left: awsui.$space-xxs;
+
+      > .types-count-visible {
+        > .type-count {
+          margin-right: awsui.$space-s;
+
+          > .count-number {
+            margin-left: awsui.$space-xxs;
```
timogasda, line 163: I'm seeing a lot of spacing tokens in the CSS. Is there a reason why we want this way instead using `<SpaceBetween>` and `<Box>`?

src/flashbar/collapsible.scss
```diff
@@ -0,0 +1,229 @@
...
+    background: awsui.$color-background-layout-toggle-hover;
+  }
+
+  &:active {
+    background: awsui.$color-background-layout-toggle-active;
+  }
+
+  &.visual-refresh {
+    $border-width: 2px;
+    @include toggle-button-layout(-27px, $border-width);
```
timogasda, line 180: What are these magic numbers representing? Can we calculate them or derive them from something?

src/flashbar/utils.ts
```diff
@@ -0,0 +1,103 @@
...
+function getItemColor(item: FlashbarProps.MessageDefinition) {
+  return getColorFromType(getItemType(item));
+}
+
+/*
+ Returns a selection of notifications, preserving the order when possible but making sure that all different colors in
+ the stack are represented in the returned array.
+ The order corresponds to how they are represented when using the collapsible feature (newest first, oldest last).
+ */
+export function getItemSelection(
```
timogasda, line 43: I find the name of this function confusing. "Selecting" has a very specific meaning in our system and code base. I would suggest something like `groupItemsByColor`.

src/flashbar/__tests__/utils.test.ts
```diff
@@ -0,0 +1,89 @@
...
+      expectedOutput: ['info', 'success', 'error'],
+    },
+    {
+      description: '3 items of different color followed by more items',
+      input: ['success', 'info', 'error', 'info', 'warning', 'progress'],
+      expectedOutput: ['success', 'info', 'error'],
+    },
+  ];
+
+  for (const testCase of cases) {
```
timogasda, line 65: minor comment: You can use `it.each` instead :)

src/flashbar/collapsible-flashbar.tsx
```diff
@@ -0,0 +1,392 @@
...
+    if (listTop === undefined) {
+      const listElement = listElementRef?.current;
+      if (listElement) {
+        setListTop(listElement.getBoundingClientRect().y);
+      }
+    }
+
+    if (initialAnimationState) {
+      updateBottomSpacing();
+      animate({
```
timogasda, line 156: Why do we need to use these special FLIP animations? Can we not use CSS and React Transitions?

timogasda (changes requested):
Huge PR, but overall it's looking really well! The elaborate code comments definitely helped.

src/internal/animate.ts
```diff
@@ -0,0 +1,110 @@
...
+      if (element) {
+        const oldRect = oldState[id];
+        if (oldRect) {
+          // Animate from here on
+          element.style.transitionProperty = `transform`;
+          // Unset inline CSS transforms so that the final state is applied
+          element.style.transform = '';
+        } else {
+          // If the element didn't exist previously, fade in as well
+          element.style.transitionProperty = `transform, opacity`;
```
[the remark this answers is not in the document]
jperals: The reason for this helper to exist is for animations in cases where the initial or final position might be in arbitrary places, so you need some JavaScript magic to calculate the value of the `transform` property. In general cases, for animating opacity you don't need all this because you have the static initial and end state so you can do it in a simpler way, e.g, with just CSS. The only reason we animate opacity it here is for the special case when the element didn't exist in the initial state, in which case I considered it makes sense as a general case to fade it in (but the animation properties such as duration and curve are still defined in CSS).

I guess it's unlikely that opacity conflicts with something else (CSS), but I prefer to keep these JS DOM manipulations minimal just in case.

src/internal/animate.ts
```diff
@@ -0,0 +1,110 @@
...
+  This can be useful when the initial dimensions or position of the element is not known,
+  so the initial offset or scaling needs to be retrieved via JS.
+
+  Caveat: this currently does not support elements having CSS transforms in the end state.
+  These would be overridden in the animation instead of combined.
+ */
+export function animate({
+  oldState,
+  elements,
+  onTransitionsEnd,
```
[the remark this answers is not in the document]
jperals: Exactly, I couldn't figure out a way to trigger it. I tried dispatching a `transitionend` event on the passed-in element but that wasn't caught by the listener...

src/flashbar/collapsible-flashbar.tsx
```diff
@@ -0,0 +1,392 @@
...
+
+  function toggleCollapseExpand() {
+    sendToggleMetric(items.length, !isFlashbarStackExpanded);
+    if (!isReducedMotion) {
+      prepareAnimations();
+    }
+    setIsFlashbarStackExpanded(prev => !prev);
+  }
+
+  const updateBottomSpacing = useCallback(() => {
```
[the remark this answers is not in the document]
jperals: What is specific to the collapsible Flashbar is the toggle button which overflows it at the bottom. Without this extra padding, it gets cut out when the Flashbar reaches the bottom of the page.

src/flashbar/utils.ts
jperals: OK, though not sure about the `group` part because it returns a single flat array and they are not grouped in a key-value object or anything like that. But I'll think of something else.

src/internal/animate.ts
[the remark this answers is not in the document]
timogasda: Ah okay, that's fine.

src/flashbar/collapsible-flashbar.tsx
```diff
@@ -0,0 +1,392 @@
...
+  const toggleElementRef = useRef<HTMLDivElement | null>(null);
+  const [transitioning, setTransitioning] = useState(false);
+  const [listTop, setListTop] = useState<number>();
+
+  const flashbarElementId = useUniqueId('flashbar');
+  const itemCountElementId = useUniqueId('item-count');
+
+  useEffect(() => {
+    if (items.length <= maxNonCollapsibleItems) {
+      setIsFlashbarStackExpanded(false);
```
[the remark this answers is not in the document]
jperals: Thanks, remove this effect

src/flashbar/collapsible.scss
jperals: This is the amount of space that the expand/collapse button is pushed up to overlap the Flashbar. I just based this on the given UI design specification. I know it's a very arbitrary number but it's also a very specific case and the number doesn't really come from anything else. I guess I could approximate it to use a combination of tokens if needed —would that be preferable?

src/flashbar/utils.ts
jperals: Went for `getVisibleCollapsedItems`

src/flashbar/__tests__/utils.test.ts
jperals: Thanks, I didn't know about it. But this doesn't seem to offer a way to let each test have a distinct description, which would make it a bit harder to debug if one or some of them fails.

src/flashbar/collapsible-flashbar.tsx
jperals: Because the Flash could be in different places (screen coordinates) both in the initial and the final state, so we would not know the amount of translation that we should apply.

For enter and leave animations we keep using CSS and React Transitions —this keeps being the preferred method when possible.

src/flashbar/interfaces.ts
```diff
@@ -58,15 +58,23 @@ export interface FlashbarProps extends BaseComponentProps {
...
+
 export interface StackedFlashbarProps {
-  stackItems: true;
-  ariaLabels?: StackedFlashbarProps.AriaLabels;
   items: ReadonlyArray<FlashbarProps.MessageDefinition>;
+  collapsible: boolean;
+  i18nStrings?: StackedFlashbarProps.I18nStrings;
 }
 
 export namespace StackedFlashbarProps {
```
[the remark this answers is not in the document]
jperals: True, thanks! Done

src/flashbar/collapsible.scss
jperals: I guess there is no big reason. I might have thought this would involve a tiny tiny bit Iess JavaScript at runtime. Would using such components be preferable?

src/flashbar/collapsible.scss
timogasda: If it can't be calculated properly, that's fine. But then you could at least put it in a SASS variable and/or add a comment.

jperals: OK, done both things (SASS variable and comment) in latest revision

taheramr (commented):
Can you update the commit message to something that we can go back to if needed? latest design does not imply how the old design looked or how this changes the flashbar's look and feel.

src/flashbar/collapsible-flashbar.tsx
jperals: This span is equivalent of what is in the LiveRegion.

Initially I put this whole span _inside_ the LiveRegion, with ARIA labels for the icons, but the LiveRegion would not read the labels, so went for doing two separate blocks, one for screen readers and one excluded from screen readers.

src/flashbar/collapsible-flashbar.tsx
jperals: Option 1 is more realistic at this point —but wouldn't it be rather `${i18nStrings[labelName]}: ${countByType[type]}`? (for example: "Error: 1", otherwise it would be "1: Error")

src/flashbar/collapsible-flashbar.tsx
```diff
@@ -0,0 +1,392 @@
...
+      }
+    },
+    [i18nStrings]
+  );
+
+  useEffect(() => {
+    if (!i18nStrings) {
+      warnOnce(componentName, 'Using the `collapsible` option requires passing an `i18nStrings` object.');
+      return;
+    }
```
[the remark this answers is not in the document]
jperals: This came from discussing the fact that if using the collapsible feature, some strings should be present, even though technically they are not mandatory.

src/flashbar/collapsible-flashbar.tsx
```diff
@@ -0,0 +1,392 @@
...
+              styles.toggle,
+              isVisualRefresh && styles['visual-refresh'],
+              isFlashbarStackExpanded ? styles.expanded : styles.collapsed,
+              transitioning && styles['animation-running']
+            )}
+            onClick={toggleCollapseExpand}
+            ref={toggleElementRef}
+          >
+            <span className={styles.status} role="status">
+              {toggleButtonText && <h2 className={styles.text}>{toggleButtonText}</h2>}
```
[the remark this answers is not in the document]
jperals: This came explicitly form the a11y specifications.

src/flashbar/collapsible-flashbar.tsx
```diff
@@ -0,0 +1,392 @@
...
+    // Allow vertical space between Flashbar and page bottom only when the Flashbar is reaching the end of the page,
+    // otherwise avoid spacing with eventual sticky elements below.
+    const listElement = listElementRef?.current;
+    const flashbar = listElement?.parentElement;
+    if (listElement && flashbar && listTop !== undefined) {
+      const bottom = listTop + listElement.clientHeight;
+      const windowHeight = window.innerHeight;
+      flashbar.style.paddingBottom = '';
+      if (isFlashbarStackExpanded && bottom + parseInt(getComputedStyle(flashbar).paddingBottom) <= windowHeight) {
+        flashbar.style.paddingBottom = '0';
```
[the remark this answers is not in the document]
jperals: We can, but still using direct DOM manipulation (e.g, with `classList.add` and `classList.remove` because when expanding, we want this change to be already applied before rendering, so that the animation goes directly to the desired place instead of first animating and then adding the bottom spacing in the case that the Flashbar reaches the end of the page when it expands.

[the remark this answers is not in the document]
jperals: > instead of first animating and then adding the bottom spacing

Well, to be exact this is not what would happen. It would render at once but the initial position of the animation would be a bit off.

src/flashbar/collapsible-flashbar.tsx
```diff
@@ -0,0 +1,392 @@
...
+
+export default function CollapsibleFlashbar({ items, ...restProps }: FlashbarProps & StackedFlashbarProps) {
+  const [enteringItems, setEnteringItems] = useState<ReadonlyArray<FlashbarProps.MessageDefinition>>([]);
+  const [exitingItems, setExitingItems] = useState<ReadonlyArray<FlashbarProps.MessageDefinition>>([]);
+  const [isFlashbarStackExpanded, setIsFlashbarStackExpanded] = useState(false);
+
+  const getElementsToAnimate = useCallback(() => {
+    const flashElements = isFlashbarStackExpanded ? expandedItemRefs.current : collapsedItemRefs.current;
+    return { ...flashElements, toggleButton: toggleElementRef.current };
+  }, [isFlashbarStackExpanded]);
```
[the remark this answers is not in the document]
jperals: This was hinted at me by my IDE. The reasoning is that there are memoized functions that used to depend on `elementsToAnimate` which would be recreated on every render. And the reason why I use `useCallback` and not `useMemo` (I tried this too) is because it would depend on the arrays `expandedItemRefs.current` and `collapsedItemRefs.current`, which don't trigger updates when they mutate. The latter is solved by invoking the function every time I need it.

So it all has to do with the subtleties of object identity JS: if not memoizing at all, I would get trouble from generating too many objects; if memoizing the values, I would not update every time I should; only memoizing as a function does the trick. I don't think it's a big overhead —it should be an unproblematic function and it will be recreated only when the expand/collapse state is toggled.

src/flashbar/collapsible-flashbar.tsx
jperals: The arrow rotates 180 degrees instead of being a different arrow per each state, even though I know the animation is not very noticeable (especially because it happens at the same time than a more prominent animation). This was part of the UI spec.

But I based my implementation on the Button's and tried to reuse what I could, to maximize consistency, like the `default-text-style` mixin. This button is basically made of these 4 rules:

```
@include typography.default-text-style;
cursor: pointer;
margin-top: 0;
margin-bottom: 0;
```

src/flashbar/collapsible-flashbar.tsx
```diff
@@ -0,0 +1,392 @@
...
+      const bottom = listTop + listElement.clientHeight;
+      const windowHeight = window.innerHeight;
+      flashbar.style.paddingBottom = '';
+      if (isFlashbarStackExpanded && bottom + parseInt(getComputedStyle(flashbar).paddingBottom) <= windowHeight) {
+        flashbar.style.paddingBottom = '0';
+      }
+    }
+  }, [isFlashbarStackExpanded, listTop]);
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
jperals: Thanks, changed.

src/flashbar/collapsible-flashbar.tsx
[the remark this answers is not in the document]
jperals: Deleted as per offline discussion

src/flashbar/collapsible-flashbar.tsx
```diff
@@ -0,0 +1,392 @@
...
+            className={clsx(
+              styles.toggle,
+              isVisualRefresh && styles['visual-refresh'],
+              isFlashbarStackExpanded ? styles.expanded : styles.collapsed,
+              transitioning && styles['animation-running']
+            )}
+            onClick={toggleCollapseExpand}
+            ref={toggleElementRef}
+          >
+            <span className={styles.status} role="status">
```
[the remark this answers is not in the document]
jperals: Thanks, deleted the attribute

src/flashbar/collapsible-flashbar.tsx
```diff
@@ -0,0 +1,392 @@
...
+              {toggleButtonText && <h2 className={styles.text}>{toggleButtonText}</h2>}
+              <span className={styles['types-count']} id={itemCountElementId}>
```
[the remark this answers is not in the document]
jperals: This code is now outdated. `aria-describedby` now references the immediate common ancestor of the LiveRegion and the content with `aria-hidden="true"`.

src/flashbar/collapsible-flashbar.tsx
jperals: Done as in previous comment.

src/flashbar/collapsible-flashbar.tsx
jperals: Done.

src/flashbar/collapsible.scss
jperals: Added a design token for the actual whole shadow, as I saw there were other shadows defined as well. This also allowed to de-duplicate these two selector groups into one.

src/flashbar/__integ__/collapsible.test.ts
```diff
@@ -1,6 +1,7 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
timogasda, line 1: _(Putting a random comment here so we can have a proper conversation thread)_

Here's my findings from testing the current implementation locally. I tested with VoiceOver together with Chrome and Safari.

1. The expanded flashbar items are not navigable as a list. Chrome+VO announces it as a list ("Notifications, 6 items") and Safari as a "group". But neither lets me use the VO commands to navigate the list item by item.
2. The tab order with the expand/collapse button is confusing. When collapsed, the button comes before the first (and only visible) flash item. When I expand, I would expect the the full list of items to appear after the button, but they are now "behind" the button.
3. I hear the Notifications counter quite often. I think it might be connected to too many things. Right now you will hear it via: 1) Live region updates, 2) `aria-describedby` on the expand/collapse button, 3) `aria-describedby` on the notifications group, 4) when navigating to the `span` that contains the counter text. Maybe one of the `aria-describedby` props is unnecessary?

src/flashbar/collapsible.scss
timogasda: It's an interesting question. I think in the past we definitely preferred components, but personally I prefer CSS as well. So... I'm okay with keeping it like this.

src/internal/animate.ts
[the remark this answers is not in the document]
timogasda: Got it, thank you for the explanation!

src/flashbar/collapsible-flashbar.tsx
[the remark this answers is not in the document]
timogasda: Got it!

src/flashbar/collapsible-flashbar.tsx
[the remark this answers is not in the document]
timogasda: Thanks for adding the class!

src/flashbar/collapsible-flashbar.tsx
timogasda: Yeah I get it now. This might actually be a good case for a "native" ARIA live region. But this works, too, so no need to change it.

timogasda (changes requested):
Again, really impressive work. The testing is super detailed! Kudos. 🎉

Just a few more code things to follow-up. I also did some manual testing and posted the findings.

src/flashbar/collapsible-flashbar.tsx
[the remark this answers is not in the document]
jperals: Not really intentional, but not sure there was another way if using the live region and if we want it to include this text, which is something I confirmed with a11y...

I'm asking back about the reasoning about the H2 element in the meantime.

src/flashbar/__tests__/collapsible.test.tsx
```diff
@@ -0,0 +1,323 @@
...
+      expect(findOuterToggleElement(wrapper.findFlashbar()!)).toBeFalsy();
+
+      rerender(<Flashbar items={[item1, item2]} {...{ collapsible: true, i18nStrings: defaultStrings }} />);
+      expect(wrapper.findFlashbar()!.findItems()).toHaveLength(1);
+      expect(findOuterToggleElement(wrapper.findFlashbar()!)).toBeTruthy();
+    });
+  });
+
+  describe('Accessibility', () => {
+    it('renders items in an unordered list', () => {
```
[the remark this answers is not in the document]
jperals: What is meant is that it is rendered as an UL element (versus ordered list, OL). The order of the items however is kept of course, and this seems to work as desired with screen readers. This was also an explicit a11y requirement. Also, the non-collapsible Flashbar also uses UL, so I think if we want to reconsider it, we should do it also for the non-collapsible Flashbar, which means: 1) IMHO it would be better to do it separately; 2) I can imagine this breaking unexpected things?

src/flashbar/__integ__/collapsible.test.ts
jperals: > 1. The expanded flashbar items are not navigable as a list. Chrome+VO announces it as a list ("Notifications, 6 items") and Safari as a "group". But neither lets me use the VO commands to navigate the list item by item.

I wonder if the same is true for the non-collapsible Flashbar, since AFAIK, except or the toggle button the markup for the list strictly speaking should be essentially the same. If so, I think anything affecting both versions can be handled separately for the sake of keeping the scope more focused. This PR is already very big and the API is not public yet.

jperals: > 2. The tab order with the expand/collapse button is confusing. When collapsed, the button comes before the first (and only visible) flash item. When I expand, I would expect the the full list of items to appear after the button, but they are now "behind" the button.

This came from the a11y spec as well. I can share more details with you

src/flashbar/__tests__/collapsible.test.tsx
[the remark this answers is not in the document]
timogasda: Yeah, this is completely outside of the scope for this PR. It just got me wondering why we don't use `ol` in general for flashbar.

src/flashbar/__integ__/collapsible.test.ts
timogasda: > I wonder if the same is true for the non-collapsible Flashbar

Yes, sorry, I should have clarified. This is happening for both types of flashbar. So this can be looked at outside of this PR.

pages/flashbar/interactive.page.tsx
```diff
@@ -78,7 +94,7 @@ export default function FlashbarPermutations() {
           </Button>
           <Button onClick={() => removeLastAndAdd('error')}>Add And Remove</Button>
         </SpaceBetween>
-        <Flashbar items={items} {...({ stackItems } as any)} />
+        <Flashbar items={items} {...(privateProps as any)} />
```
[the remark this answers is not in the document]
jperals: I just removed the `as any`

src/flashbar/__tests__/collapsible.test.tsx
```diff
@@ -0,0 +1,323 @@
...
+import { FlashbarProps, FlashType, CollapsibleFlashbarProps } from '../interfaces';
+import { render } from '@testing-library/react';
+import { warnOnce } from '../../../lib/components/internal/logging';
+
+jest.mock('../../../lib/components/internal/logging', () => ({
+  warnOnce: jest.fn(),
+}));
+
+afterEach(() => {
+  (warnOnce as jest.Mock).mockReset();
```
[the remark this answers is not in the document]
jperals: Right, thanks! Just removed them

timogasda (commented):
Getting really close! :)

Given the importance of the H2, we should not ARIA-hide it.

src/flashbar/__integ__/collapsible.test.ts
```diff
@@ -30,6 +35,7 @@ describe('Collapsible Flashbar', () => {
         setupTest(async page => {
           await page.toggleCollapsibleFeature();
           await page.addInfoFlash();
+          await wait(FOCUS_THROTTLE_DELAY);
```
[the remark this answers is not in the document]
jperals: If it's for saving time, it wouldn't work unfortunately because we need to wait exactly this amount of time, because of the focus throttling. The test also a bit more exact this way. What if the expected state is satisfied too early, but one the throttle delay passes it's in an undesired state? If using `waitFor`, the test would succeed but it should actually fail.

src/flashbar/flash.tsx
```diff
@@ -148,7 +150,7 @@ export const Flash = React.forwardRef(
           <div
             {...focusVisible}
             className={styles['flash-focus-container']}
-            tabIndex={ariaRole === 'alert' ? -1 : undefined}
+            tabIndex={focusable || ariaRole === 'alert' ? -1 : undefined}
```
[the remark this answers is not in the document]
jperals: Done

pages/flashbar/collapsible.permutations.page.tsx
```diff
@@ -0,0 +1,91 @@
...
+export default function CollapsibleFlashbarPermutations() {
+  return (
+    <>
+      <h1>Collapsible Flashbar permutations</h1>
+      <ScreenshotArea>
+        <PermutationsView
+          permutations={permutations}
+          render={permutation => (
+            <Flashbar
+              {...{
```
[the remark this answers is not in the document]
jperals: Done

src/flashbar/collapsible-flashbar.tsx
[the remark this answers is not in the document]
jperals: Went back to the simpler aria-role approach instead of LiveRegion component, it turned out to be less problematic.

src/flashbar/collapsible-flashbar.tsx
```diff
@@ -0,0 +1,351 @@
...
+  count,
+}: {
+  iconName: IconProps.Name;
+  label?: string;
+  count: number;
+}) => {
+  return (
+    <span className={styles['type-count']}>
+      <span aria-label={label} role="img">
+        <span title={label} aria-hidden="true">
```
jperals, line 344: The `title` attribute is applied to an extra nested element with `aria-hidden` instead of the same element that has `aria-label` in order to prevent NVDA from reading the same text twice.

src/flashbar/styles.scss
```diff
@@ -66,22 +66,21 @@
   display: flex;
   flex-grow: 1;
   min-width: 0;
+}
```
jperals, line 69: This fixes an existing bug I noticed: the focus ring in the Flash element was not being displayed correctly because the corresponding CSS rules were inside the selector for `.flash-body` instead of `.flash-focus-container`

src/flashbar/__tests__/collapsible.test.tsx
```diff
@@ -203,44 +205,26 @@ describe('Collapsible Flashbar', () => {
...
-      expect(liveRegionElement).toHaveTextContent(`${customLabels.successCountAriaLabel}: 2`);
-      expect(liveRegionElement).toHaveTextContent(`${customLabels.warningCountAriaLabel}: 3`);
-      const ariaHiddenElement = liveRegionElement!.nextElementSibling;
-      expect(ariaHiddenElement).toHaveAttribute('aria-hidden', 'true');
+    it('announces updates to the item counter with aria-live', () => {
+      const flashbar = renderFlashbar();
+      const counter = findOuterCounter(flashbar);
+      expect(counter).toHaveAttribute('aria-live', 'polite');
+      // We add `role="status"` as well, to maximize compatibility
+      // https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions#roles_with_implicit_live_region_attributes
```
timogasda, line 226: (minor) This is good to know, but I would expect this comment in the component code, not the test code.

src/flashbar/collapsible-flashbar.tsx
timogasda: I suspect that NVDA isn't reading it twice now because this element is `aria-hidden`, not because it's a separate element. But either way, it's the result that we want.

timogasda (dismissed):
No blocking comments. Very well done! Thanks for persevering through this long review.

src/flashbar/__tests__/flashbar.test.tsx
```diff
@@ -374,7 +374,7 @@ describe('Analytics', () => {
   it('sends a render metric when items are provided', () => {
     createFlashbarWrapper(
       <Flashbar
-        {...{ collapsible: true }}
+        {...{ stackItems: true }}
```
timogasda, line 377: (for later): Unnecessary spread. There's a few more like it in this file.

timogasda (approved):

PR #644, opened 2023-01-17
fix: remove operator chars from aria announcement

Cooperbuilt:
### Description
This PR removes "=", "!=", "<", ">", "<=", ">=", ":" from the announcement returned from use-announcement to fix an a11y bug in which operator characters were read multiple times. This change was tested via unit test and using the mac VoiceOver tool.

This fixes two a11y issues in the property filter (the autocomplete popup and edit filter dialog). See below for more info.

### Change Motivation
Some options in the search form's autocomplete popup and edit filter dialog have confusing accessible names announced by screen readers. This is a result of how special characters are used within the option text. For example, the option "Status != Does not equal" is announced by screen readers as "Status equals does not equal". This is because special characters (with the exception of some punctuation characters) are read literally by screen readers. The search filter options may be difficult to understand using a screen reader

### How has this been tested?
I added a simple unit test to show the behavior of removing an operator character. I can expand on this if this team determines it's appropriate. Just let me know! 

I tested this with VoiceOver in as many configurations as I could and it reliably removes the operator character from voice over reading. Instead of reading "Instance Equals Equals" and other similar redundancies it now skips the operator character and instead reads the operator description. 

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/select/__tests__/use-announcement.test.ts
+++ b/src/select/__tests__/use-announcement.test.ts
@@ -38,6 +38,15 @@ const options = [
       },
     ],
   },
+  {
+    label: 'Group 3',
+    options: [
+      {
+        label: 'Child =',
+        value: 'should-not-be-announced-5',
+      },
+    ],
+  },
 ];
 
 const { flatOptions, parentMap } = flattenOptions(options);
@@ -91,6 +100,18 @@ describe('useAnnouncement', () => {
       },
     });
     expect(hook.result.current).toEqual('Selected Group 1 Child 1');
+  });
+
+  test('announcements with selected highlighted item should not include operator characters', () => {
+    const hook = renderHook(useAnnouncement, {
+      initialProps: {
+        getParent,
+        selectedAriaLabel: 'Selected',
+        highlightedOption: flatOptions[8],
+        announceSelected: true,
+      },
+    });
+    expect(hook.result.current).toEqual('Selected Group 3 Child');
   });
 
   test('should return custom announcement string when renderHighlightedAriaLive is provided', () => {
--- a/src/select/utils/use-announcement.ts
+++ b/src/select/utils/use-announcement.ts
@@ -55,7 +55,16 @@ export function useAnnouncement<Option extends OptionHolder>({
   }
 
   // Use default renderer with selected ARIA label if defined and relevant.
+  /**
+   * The replace method on the end will remove all characters "=", "!=", "<", ">", "<=", ">=", ":"
+   * from the announcement, not just specific instances of them. This was added to avoid screen
+   * readers reading "equals equals" and similarly confusing combinations of operators
+   */
   const selectedAnnouncement = announceSelected && selectedAriaLabel ? selectedAriaLabel : '';
   const defaultDescription = defaultOptionDescription(option, group);
-  return [selectedAnnouncement, defaultDescription].filter(Boolean).join(' ');
+  return [selectedAnnouncement, defaultDescription]
+    .filter(Boolean)
+    .join(' ')
+    .replace(/[=!<>:]/g, '')
+    .trim();
 }


Cooperbuilt:
I wanted to call out that when running tests locally I saw some failures but they seem completely unrelated. Looking for any guidance. Some examples:
```
Documenter › definition for tabs matches the snapshot

    Cannot find module '/Users/evancooper/projects/cloudscape-components/lib/components-definitions/components/tabs' from 'src/__tests__/utils.tsx'
```

and 
```html
FAIL  src/area-chart/__tests__/area-chart-initial-state.test.tsx (24.161 s)
  ● filter selected aria label is assigned

    expect(element).toContainHTML()
    Expected:
      Series:
    Received:
      <div class="awsui_root_f8g6q_wid5c_93 awsui_chart-filter_1px7g_alf5y_104"><div class="awsui_root_qwoo0_1jaxg_143"><div class=""><button aria-expanded="true" aria-haspopup="listbox" aria-labelledby="multiselect-arialabel-177-1673918774007-2025 trigger-content-185-1673918774224-2297" class="awsui_button-trigger_18eso_xvimv_97 awsui_pressed_18eso_xvimv_151 awsui_has-caret_18eso_xvimv_135" id="formField175-1673918774007-6790" type="button"><span aria-disabled="true" class="awsui_placeholder_dwuol_1dfho_93 awsui_trigger_dwuol_1dfho_116" id="trigger-content-185-1673918774224-2297" /><span class="awsui_arrow_18eso_xvimv_97"><span class="awsui_icon_h11ix_31bp4_98 awsui_size-normal-mapped-height_h11ix_31bp4_151 awsui_size-normal_h11ix_31bp4_147 awsui_variant-normal_h11ix_31bp4_219"><svg aria-hidden="true" focusable="false" viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg"><path class="filled stroke-linejoin-round" d="M4 5h8l-4 6-4-6z" /></svg></span></span></button></div><div tabindex="-1" /><div><div tabindex="-1" /><div aria-hidden="false" class="awsui_dropdown_qwoo0_1jaxg_93 awsui_open_qwoo0_1jaxg_234 awsui_hide-upper-border_qwoo0_1jaxg_227 awsui_occupy-entire-width_qwoo0_1jaxg_224 awsui_dropdown-drop-right_qwoo0_1jaxg_221" data-animating="true" data-open="true"><div class="awsui_dropdown-content-wrapper_qwoo0_1jaxg_93"><div class="awsui_ie11-wrapper_qwoo0_1jaxg_257"><div class="awsui_dropdown-content_qwoo0_1jaxg_93" style="max-height: -46px;"><ul aria-activedescendant="option-list178-1673918774007-6403-option-0" aria-labelledby="multiselect-arialabel-177-1673918774007-2025 formField175-1673918774007-6790" class="awsui_options-list_19gcf_1wdy7_93" id="option-list178-1673918774007-6403" role="listbox" style="position: relative;" tabindex="-1"><li aria-checked="true" class="awsui_selectable-item_15o6u_1ndqp_93 awsui_selected_15o6u_1ndqp_125 awsui_highlighted_15o6u_1ndqp_125 awsui_next-item-selected_15o6u_1ndqp_147 awsui_interactiveGroups_15o6u_1ndqp_163" data-mouse-target="0" data-test-index="1" id="option-list178-1673918774007-6403-option-0" role="option"><div aria-hidden="true" class="awsui_option-content_15o6u_1ndqp_220"><div … /></div><div class="awsui_measure-strut_15o6u_1ndqp_201" /><div class="awsui_screenreader-content_15o6u_1ndqp_214">Series Area Series 1</div></li><li aria-checked="true" class="awsui_selectable-item_15o6u_1ndqp_93 awsui_selected_15o6u_1ndqp_125 awsui_pad-bottom_15o6u_1ndqp_113 awsui_interactiveGroups_15o6u_1ndqp_163" data-mouse-target="1" data-test-index="2" id="option-list178-1673918774007-6403-option-1" role="option"><div class="awsui_option-content_15o6u_1ndqp_220"><div … /></div><div class="awsui_measure-strut_15o6u_1ndqp_201" /><div class="awsui_screenreader-content_15o6u_1ndqp_214" /></li></ul><div aria-atomic="true" aria-live="polite" class="awsui_root_1afi9_1ehx9_93 awsui_hidden_1afi9_1ehx9_103" /></div></div></div></div><div tabindex="-1" /></div></div><span class="awsui_root_xttbq_1yl3y_93" id="multiselect-arialabel-177-1673918774007-2025" /></div>
```

rubencarvalho:
Hello Evan,

Thanks for opening the PR!

I understand the motivation of your changes but the `useAnnouncement` hook is used by multiple components to produce the live region string to be announced when an option is highlighted.

The changes you proposed would therefore also impact, for example, the `MultiSelect`, `AutoSuggest` and `Select` components, in which case the solution proposed might not be the most optimal.

I'm happy to bring this up within our team and/or continue the discussion towards an alternative here.

Let me know what you think.
Thanks!

Cooperbuilt:
Hey! Thanks for the reply here. I appreciate your thoughtful response. I do know this hook is used in multiple places and I made that change deliberately. I don't believe we want a screen reader recording operator characters but I see why that would be risky. I could scope this filter to the announcement text of just the Property filter if that makes more sense. Let me know what you'd like to do, and thanks for the help with the contribution.

Cooperbuilt:
Following up here. I believe making this change at the use-announcement level is the right call because of inconsistencies in the screen-reader handling of mathematical characters and punctuation characters. [Here is a table](https://www.deque.com/blog/dont-screen-readers-read-whats-screen-part-1-punctuation-typographic-symbols/#math) from a leading a11y resource on the differences between screen readers in how they read mathematical characters (like the ones I am filtering)

Note that many screenreaders fail to even announce "=", "<", etc. Because of this inconsistency, I think it makes sense to enforce the removal of operator characters to enforce engineers using descriptive i18n labels for their descriptions that include the full word like "equals" and "does not equal".

Cooperbuilt:
Closing due to a change in direction

PR #647, opened 2023-01-18
fix: Avoid similar announcement of menu button on buttons in toggle panel.

YueyingLu:
### Description

> The expanded state of side navigation region is not clear to users of screen reader. It is recommended to remove the aria-expanded and aria-haspopup attributes from the Close (X) control of the side nav. Presently both the close control as well as the hamburger menu control have these attributes, which result in similar announcement of ‘menu’ on both, which in turn can lead to confusion. Keeping the aria-haspopup along with aria-expanded on the hamburger menu control would be sufficient to provide the information

Related links, issue AWSUI-20160

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/visual-refresh/app-bar.tsx
+++ b/src/app-layout/visual-refresh/app-bar.tsx
@@ -57,7 +57,7 @@ export default function AppBar() {
         >
           <InternalButton
             ariaLabel={ariaLabels?.navigationToggle ?? undefined}
-            ariaExpanded={isNavigationOpen}
+            ariaExpanded={isNavigationOpen ? undefined : false}
             iconName="menu"
             formAction="none"
             onClick={() => handleNavigationClick(true)}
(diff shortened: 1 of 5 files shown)


src/app-layout/visual-refresh/app-bar.tsx
```diff
@@ -57,15 +57,15 @@ export default function AppBar() {
         >
           <InternalButton
             ariaLabel={ariaLabels?.navigationToggle ?? undefined}
-            ariaExpanded={isNavigationOpen}
+            ariaExpanded={isNavigationOpen ? undefined : false}
```
just-boris, line 60: Is this basically an "always false" value?

YueyingLu: Yes, it is false (on hamburger menu button) or unset (close/collapse button).

just-boris: How about removing it, if never activated?

YueyingLu: It needs _**false**_ value to announce something like "collapsed menu button"

just-boris: I see, I confused the API of `AppLayoutButton` and `InternalButton`. 

Very error-prone to have two components with same property name behaving differently

just-boris (approved):

PR #659, opened 2023-01-23
feat: Add ariaControls prop to RadioGroup

fa7ad:
### Description
This PR adds the `ariaControls` prop to RadioGroup component.
This should allow users to implement a progressive disclosure pattern AND make it accessible.

~~NOTE: The API documentation needs to reviewed and potentially changed; hold off on merging just yet.~~ _approved_

### How has this been tested?
* Updated existing Unit Tests
* Created a dev page that uses the prop to implement this pattern.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/radio-group/progressive-disclosure.page.tsx
+++ b/pages/radio-group/progressive-disclosure.page.tsx
@@ -1,0 +1,57 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import React, { useState } from 'react';
+import Box from '~components/box';
+import FormField from '~components/form-field';
+import RadioGroup from '~components/radio-group';
+import Select, { SelectProps } from '~components/select';
+
+const langOptions: Array<SelectProps.Option> = [
+  { value: 'en-US', label: 'English (United States)' },
+  { value: 'en-GB', label: 'English (United Kingdom)' },
+  { value: 'de-DE', label: 'Deutsch (Deutschland)' },
+  { value: 'de-CH', label: 'Deutsch (Schweiz)' },
+];
+
+export default function RadiosPage() {
+  const [radioSelection, setRadioSelection] = useState<string>('');
+  const [lang, setLang] = useState<SelectProps.Option>(langOptions[0]);
+
+  return (
+    <Box padding="l">
+      <h1>Radio group progressive disclosure demo</h1>
+      <FormField
+        label="Language settings"
+        description="You can select a specific language or have this service automatically identify the language in your media."
+      >
+        <RadioGroup
+          value={radioSelection}
+          onChange={event => setRadioSelection(event.detail.value)}
+          ariaControls="language-settings"
+          items={[
+            {
+              label: 'Specific language',
+              value: 'specific',
+              description:
+                'If you know the language spoken in the source media, choose this option for optimal results.',
+            },
+            {
+              label: 'Automatic language detection',
+              value: 'automatic',
+              description: 'If you do not know the language spoken in the source media, choose this option.',
+            },
+          ]}
+        />
+      </FormField>
+      <Box id="language-settings" display={radioSelection !== 'specific' ? 'none' : 'block'}>
+        <FormField label="Language" description="Select the language you want to use.">
+          <Select
+            selectedOption={lang}
+            onChange={event => setLang(event.detail.selectedOption)}
+            options={langOptions}
+          />
+        </FormField>
+      </Box>
+    </Box>
+  );
+}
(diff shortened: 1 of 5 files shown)


pages/radio-group/progressive-disclosure.page.tsx
```diff
@@ -0,0 +1,113 @@
...
+            },
+            {
+              label: 'Automatic language detection',
+              value: 'automatic',
+              description: 'If you do not know the language spoken in the source media, choose this option.',
+            },
+          ]}
+        />
+      </FormField>
+      <Box id="dialekt-settings" display={!Object.keys(lang).includes(radioSelection) ? 'none' : 'block'}>
```
just-boris, line 88: Should it be `display:none`, `aria-hidden=true` or just a conditional rendering? What happens if aria controls resolves to a non-existing id?

pages/radio-group/progressive-disclosure.page.tsx
```diff
@@ -0,0 +1,113 @@
...
+    case 'update/en':
+      return { ...state, en: action.payload };
+    case 'update/de':
+      return { ...state, de: action.payload };
+    default:
+      return state;
+  }
+};
+
+export default function RadiosPage() {
```
just-boris, line 49: Impressive demo, but a little bit over-engineered for this case. It is extra complexity when somebody tries to read this demo and apply it for their case

1. Remove `useReducer`, plain state will do
2. Remove extra options from radio group, we only need to show case how to show/hide an extra control


Basically, this demo should show a minimal reproducible example of `ariaControls` use-case.

src/radio-group/interfaces.ts
```diff
@@ -40,6 +40,11 @@ export interface RadioGroupProps extends BaseComponentProps, FormFieldControlPro
    */
   ariaRequired?: boolean;
 
+  /**
+   * Adds `aria-controls` attribute to the radio group. Use this property if the radio group controls the visibility of other elements on the page.
```
just-boris, line 44: How about this

```suggestion
   * Adds `aria-controls` attribute to the radio group. If radio group selection controls presence of another form field, use this property to provide an id the controlled field
```

just-boris (approved):

pages/radio-group/progressive-disclosure.page.tsx
just-boris: Can we do conditional rendering? (I think this should be recommended, for example to allow lazy-loading options for conditional select). If not, what is the alternative?

timogasda: > What happens if aria controls resolves to a non-existing id?

Certain a11y checkers will throw a violation 👀

fa7ad: Exactly, the referenced ID is not in the document; some checkers will throw a violation. <sup>[example](https://rocketvalidator.com/html-validation/the-aria-controls-attribute-must-point-to-an-element-in-the-same-document)</sup>

just-boris: How is it different if hidden elements are not in the a11y tree anyway?

just-boris: P.S. if you have an expert approval on this pattern, I can approve the PR after addressing other comments.

But my general software development experience suggests me that something is does not match up here

fa7ad: `aria-controls` is not limited to visible elements; at least as far as the spec is concerned.

https://www.w3.org/TR/wai-aria-1.1/#aria-controls
For example, Tab content visibility is given as a valid use of aria-controls

In the discussion for this issue, they make the distinction between non-existent and non-visible elements.
https://github.com/w3c/aria/issues/531

src/radio-group/interfaces.ts
just-boris: Still would like to hear what you think about this suggestion

just-boris (dismissed):

src/radio-group/interfaces.ts
fa7ad: I'm generally open to the change, but not so sure about the suggestion.

I'm not sure if this part is syntactically correct: `to provide an id the controlled field`.

How about this instead?

```js
   /**
    * Adds `aria-controls` attribute to the radio group. 
    * If the radio group controls the visibility of any secondary content (like another FormField), provide an ID that refers to the secondary content.
    */

```

fa7ad: Updated API doc following offline discussion

just-boris (approved):

PR #663, opened 2023-01-24
fix: Container with empty content has incorrect header border radius

YueyingLu:
### Description

When container has no content and disableContentPaddings, the bottom border radius are overlapped by header. The change adds border radius when no content in the container, as we did for collapsed ExpandableSection. 
For test, add one more instance in simple.page, it will show in the screenshot tests. 

Related links, issue AWSUI-20171

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/container/simple.page.tsx
+++ b/pages/container/simple.page.tsx
@@ -102,6 +102,10 @@ export default function SimpleContainers() {
               Content area takes the available vertical space
             </Container>
           </div>
+          <Container
+            disableContentPaddings={true}
+            header={<Header variant="h2">Container Without Content</Header>}
+          ></Container>
         </SpaceBetween>
       </ScreenshotArea>
     </article>
--- a/src/container/index.tsx
+++ b/src/container/index.tsx
@@ -24,7 +24,7 @@ export default function Container({
       disableContentPaddings={disableContentPaddings}
       {...externalProps}
       {...baseComponentProps}
-      __hiddenContent={disableContentPaddings && !props.children}
+      __hiddenContent={!props.children}
     />
   );
 }
--- a/src/container/index.tsx
+++ b/src/container/index.tsx
@@ -24,6 +24,7 @@ export default function Container({
       disableContentPaddings={disableContentPaddings}
       {...externalProps}
       {...baseComponentProps}
+      __hiddenContent={disableContentPaddings && !props.children}
     />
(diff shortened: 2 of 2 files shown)


src/container/index.tsx
```diff
@@ -24,6 +24,7 @@ export default function Container({
       disableContentPaddings={disableContentPaddings}
       {...externalProps}
       {...baseComponentProps}
+      __hiddenContent={disableContentPaddings && !props.children}
```
abdhalees, line 27: is `disableContentPaddings` necessary? 
this should work regardless of padding, customers used this as a hack but I think it should work without it

YueyingLu: Makes sense! Will update it.

pages/container/simple.page.tsx
```diff
@@ -102,6 +102,10 @@ export default function SimpleContainers() {
               Content area takes the available vertical space
             </Container>
           </div>
+          <Container
+            disableContentPaddings={true}
```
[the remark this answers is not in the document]
YueyingLu: I would rather leave it here because without disableContentPaddings the border looks correct. It is to make sure the fix still works even with disableContentPaddings.

abdhalees (approved):

src/container/index.tsx
```diff
@@ -24,6 +24,7 @@ export default function Container({
       disableContentPaddings={disableContentPaddings}
       {...externalProps}
       {...baseComponentProps}
+      __hiddenContent={!props.children}
```
just-boris, line 27: Does this mean that all internal use of container remain unfixed? Why is that needed?

This fix should be applied to internal instances too

just-boris: Basically, [this line](https://github.com/cloudscape-design/components/blob/main/src/container/internal.tsx#L104) should become

```
[styles['with-hidden-content']]: !children || __hiddenContent
```

YueyingLu: Hi Boris, thanks for checking it. I was not sure we should not use container without content internally at all, It should be either ExpandableSection or Header? But I agree it makes sense the fix should be for all cases.

YueyingLu: fix in https://github.com/cloudscape-design/components/pull/669

PR #674, opened 2023-01-27
fix: Make current page breadcrumb item not a link

gethinwebster:
### Description

Previously the current page item was kinda a link and kinda not, which was causing some accessibility issues. After internal discussion and reviewing reference implementations, the best approach is to make it just text.

Related links, issue #, if available: AWSUI-20172

### How has this been tested?

Updated unit tests, deployed locally & tested with VO & Chrome. See internal ticket for more details.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/test-utils/dom/breadcrumb-group/index.ts
+++ b/src/test-utils/dom/breadcrumb-group/index.ts
@@ -15,7 +15,7 @@ export default class BreadcrumbGroupWrapper extends ComponentWrapper {
    * @see findBreadcrumbLink
    */
   findBreadcrumbLinks(): Array<ElementWrapper> {
-    return this.findAll(`.${itemStyles.breadcrumb} a`);
+    return this.findAll(`.${itemStyles.breadcrumb}>*`);
   }
   /**
    * Returns a link for a given index.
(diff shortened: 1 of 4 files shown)


src/test-utils/dom/breadcrumb-group/index.ts
```diff
@@ -15,7 +15,7 @@ export default class BreadcrumbGroupWrapper extends ComponentWrapper {
    * @see findBreadcrumbLink
    */
   findBreadcrumbLinks(): Array<ElementWrapper> {
```
jperals, line 17: Not sure `findBreadcrumbLinks` should return an item that is not a link, but I assume this would be a breaking change we don't want to make?

Maybe it would be good to add a note to the component's [Testing page](https://cloudscape.design/components/breadcrumb-group/?tabId=testing) to clarify this (the fact that the function will return the last item even if it's not a link).

gethinwebster: Yeah, it would be a breaking change (and from a purely semantic point of view it didn't really return a link before, as the last item was an `a` tag with no `href` or click handlers). I wonder whether deprecating this, and replacing it with `findBreadcrumbItems` (with the same implementation) could be clearer, what do you think?

jperals: Sounds good to me (but we should probably seek further team consensus about this since it's an API change)

gethinwebster: discussed further in team, preference is to keep the naming but make the documentation clearer

jperals (approved):

PR #694, opened 2023-02-01
chore: Let Flashbar render the items in the same order as in the array, also when using the Stacked notifications feature

jperals:
### Description

After this change, the behavior of the Flashbar when using the Stacking feature matches the current behavior when it comes to item order: the first item in the `items` array is displayed visually at the top of the stack and the last one of the array is displayed at the bottom.

### How has this been tested?

- Covered by existing unit tests and integration tests
- Added one integration test
- Manually tested

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/flashbar/interactive.page.tsx
+++ b/pages/flashbar/interactive.page.tsx
@@ -36,7 +36,7 @@ export default function FlashbarPermutations() {
   };
 
   const add = (type: FlashbarProps.Type, hasHeader = false) => {
-    setItems(items => [...items, generateItem(type, dismiss, hasHeader)]);
+    setItems(items => [generateItem(type, dismiss, hasHeader), ...items]);
   };
 
   const addMultiple = (type: FlashbarProps.Type, hasHeader = false) => {
@@ -45,12 +45,12 @@ export default function FlashbarPermutations() {
     setTimeout(() => add(type, hasHeader), 200);
   };
 
-  const addTop = (type: FlashbarProps.Type, hasHeader = false) => {
-    setItems(items => [generateItem(type, dismiss, hasHeader), ...items]);
+  const addToBottom = (type: FlashbarProps.Type, hasHeader = false) => {
+    setItems(items => [...items, generateItem(type, dismiss, hasHeader)]);
   };
 
-  const removeLastAndAdd = (type: FlashbarProps.Type, hasHeader = false) => {
-    setItems(items => [...items.slice(0, items.length - 1), generateItem(type, dismiss, hasHeader)]);
+  const removeAndAddToBottom = (type: FlashbarProps.Type, hasHeader = false) => {
+    setItems(items => [generateItem(type, dismiss, hasHeader), ...items.slice(1, items.length)]);
   };
 
   const [collapsible, setCollapsible] = useState(false);
@@ -95,10 +95,10 @@ export default function FlashbarPermutations() {
             Add Multiple Error Flashes
           </Button>
           <Button onClick={() => add('warning')}>Add Warning Flash</Button>
-          <Button data-id="add-to-top" onClick={() => addTop('error')}>
-            Add To Top
+          <Button data-id="add-error-to-bottom" onClick={() => addToBottom('error')}>
+            Add To Bottom
           </Button>
-          <Button onClick={() => removeLastAndAdd('error')}>Add And Remove</Button>
+          <Button onClick={() => removeAndAddToBottom('error')}>Add And Remove</Button>
         </SpaceBetween>
         <Flashbar items={items} {...restProps} />
       </SpaceBetween>
--- a/src/flashbar/__integ__/collapsible.test.ts
+++ b/src/flashbar/__integ__/collapsible.test.ts
@@ -8,22 +8,22 @@ describe('Collapsible Flashbar', () => {
     it(
       'can expand and collapse stacked notifications with the keyboard',
       setupTest(async page => {
-        await page.toggleCollapsibleFeature();
+        await page.toggleStackingFeature();
 
         // Navigate past all buttons to add and remove flashes
         await page.keys(new Array(8).fill('Tab'));
 
-        await expect(page.findFlashes()).resolves.toBe(1);
+        await expect(page.countFlashes()).resolves.toBe(1);
         await page.keys('Space');
 
-        await expect(page.findFlashes()).resolves.toBe(5);
+        await expect(page.countFlashes()).resolves.toBe(5);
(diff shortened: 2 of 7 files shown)


pages/flashbar/interactive.page.tsx
```diff
@@ -34,7 +34,7 @@ export default function FlashbarPermutations() {
   };
 
   const add = (type: FlashbarProps.Type, hasHeader = false) => {
-    setItems(items => [...items, generateItem(type, dismiss, hasHeader)]);
+    setItems(items => [generateItem(type, dismiss, hasHeader), ...items]);
```
jperals, line 37: Updated the test page to follow the new guidelines, so clicking on the buttons that add notifications adds them to the top, not the bottom.

(True that the new guidelines are not official yet, but it seems reasonable to avoid double work and change them again when releasing. And after all these are not reference demos, just tests)

src/flashbar/__integ__/collapsible.test.ts
```diff
@@ -2,13 +2,14 @@
...
 import { FOCUS_THROTTLE_DELAY } from '../utils';
+import { wait } from '../../internal/utils/promises';
 
 describe('Collapsible Flashbar', () => {
   describe('Keyboard navigation', () => {
     it(
       'can expand and collapse stacked notifications with the keyboard',
       setupTest(async page => {
-        await page.toggleCollapsibleFeature();
+        await page.toggleStackingFeature();
```
jperals, line 12: Updated according to new naming convention. There a still places using the term "collapsible", but not sure I want to change it everywhere. The feature is about "stacking", and when using it the notifications become "stacked", but it might still make sense to say that the Flashbar becomes "collapsible". In any case further changes are out of scope here

src/flashbar/collapsible-flashbar.tsx
```diff
@@ -88,9 +88,9 @@ export default function CollapsibleFlashbar({ items, ...restProps }: FlashbarPro
 
   useLayoutEffect(() => {
     if (isFlashbarStackExpanded && items?.length) {
-      const lastItem = items[items.length - 1];
-      if (lastItem.id !== undefined) {
-        focusFlashById(ref.current, lastItem.id);
+      const mostRecentItem = items[0];
```
jperals, line 91: Actual component changes are in this file, all other diffs are about updating tests.

src/flashbar/__integ__/collapsible.test.ts
```diff
@@ -2,28 +2,29 @@
 // SPDX-License-Identifier: Apache-2.0
 import { setupTest } from './pages/interactive-page';
 import { FOCUS_THROTTLE_DELAY } from '../utils';
+import { wait } from '../../internal/utils/promises';
```
just-boris, line 5: This utility already exists, `page.pause()`

Same may also argue that `page.waitForAssertion()` may work better for your case

src/flashbar/__integ__/collapsible.test.ts
```diff
@@ -78,7 +79,7 @@ describe('Collapsible Flashbar', () => {
       test(
         'adding flash with ariaRole="alert" moves focus to the new flash',
         setupTest(async page => {
-          await page.toggleCollapsibleFeature();
+          await page.toggleStackingFeature();
           await page.toggleCollapsedState();
           await wait(FOCUS_THROTTLE_DELAY);
```
just-boris, line 84: What happens if user clicks two buttons too quick?

What happens if our consumers do not add this timeout in their integration tests?

Does the stacking feature respect our disable-motion config? (which is activated in tests and makes all transitions instant)

jperals: Note that the delay is for the focus behavior, not for items to be added/removed or animations. The focus is throttled with a delay in order to give time for screen readers to read an item before focusing on another one. This was implemented as part of [#454](https://github.com/cloudscape-design/components/pull/454). As explained in that PR's description:

> There is also throttling for focus moves, to prevent abrupt jumps when multiple messages are added sequentially

So to answer your questions:

> What happens if user clicks two buttons too quick?

The items will be added or removed as expected. The focus would be debounced so that screen reader messages do not overlap too much, as explained before.

> What happens if our consumers do not add this timeout in their integration tests?

If they use utility functions like `getItems`, they should return the expected results instantaneously as usual. If they are asserting focus behavior, they should be aware of the throttling behavior —but I doubt they want to do that.

> Does the stacking feature respect our disable-motion config? (which is activated in tests and makes all transitions instant)

Yes, just manually tested to double-check. These delays have nothing to do with animations.

src/flashbar/__integ__/collapsible.test.ts
jperals: Had a similar discussion here: https://github.com/cloudscape-design/components/pull/642#discussion_r1090366256

I prefer to wait exactly this amount of time because I actually _know_ I have to wait for the focus throttle delay (related to the other discussion below in this PR).

But I didn't know about `page.pause`, makes sense.

src/flashbar/__integ__/collapsible.test.ts
just-boris: Thanks for linking that PR, because that one is problematic. 

For this test my recommendation would be to wrap the last line into `waitForAssertion`. No any arbitrary delays in between (because real users wouldn't wait the same)

jperals: They _would_ wait the same, because it's the exact delay that is applied to the focus change.

This is not the same situation as when waiting for the page to e.g, react to user input or submitting a form, where you don't know how much it will take and you want to continue as soon as possible, with a timeout to fail when the duration would be considered too long. I understand that for these typical cases something like `useForAssertion` is preferable.

just-boris: Even if the delay is deterministic, by definition of e2e test it should not rely in implementation detail (so the test passes before and after any non-breaking-change refactorings)

jperals: But a non-breaking refactoring should not change that delay. And the other way around, the way I think of this, changing the delay should be considered a breaking change, so if the focus change takes shorter or longer unintentionally, that should be caught.

just-boris: Not a big deal. You may keep it this way if you want. But remove the custom `wait`, please

jperals: Thanks. Updated to use `page.pause` and removed the custom `wait` function.

just-boris (approved):

PR #699, opened 2023-02-02
fix: Add ARIA label to Flashbar

jperals:
### Description

A requirement for an ARIA label was defined while developing the Stacked notifications feature. For that reason the ARIA label was added to the "collapsible" version of the Flashbar only, but in reality it should always be there, also when not using the new feature.

This change backports the ARIA label to the "regular Flashbar, making sure that once the stacking feature is released, the new `i18nStrings.ariaLabel` prop has the desired effect regardless of the stacking feature being used (i.e, `stackItems` set to `true`) or not.

(The prop itself will become public once the stacking feature is public)

### How has this been tested?

- Added 2 unit tests, which are run twice each (with and without animations)
- Manually tested by inspecting the list element with browser tools

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/flashbar/__tests__/flashbar.test.tsx
+++ b/src/flashbar/__tests__/flashbar.test.tsx
@@ -6,7 +6,7 @@ import Flashbar, { FlashbarProps } from '../../../lib/components/flashbar';
 import Button from '../../../lib/components/button';
 import createWrapper from '../../../lib/components/test-utils/dom';
 import styles from '../../../lib/components/flashbar/styles.css.js';
-import { createFlashbarWrapper } from './common';
+import { createFlashbarWrapper, findList } from './common';
 
 let mockUseAnimations = false;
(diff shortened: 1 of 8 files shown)


src/flashbar/__tests__/flashbar.test.tsx
```diff
@@ -334,6 +334,44 @@ describe('Flashbar component', () => {
...
+                },
+              ]}
+              i18nStrings={{
+                ariaLabel: customAriaLabel,
+              }}
+            />
+          );
+          const list = findList(flashbar)!;
+          expect(list).toBeTruthy();
+          expect(list.getElement().getAttribute('aria-label')).toEqual(customAriaLabel);
```
timogasda, line 372: minor: You can use `expect(...).toHaveAttribute`

timogasda (approved):

PR #702, opened 2023-02-03
feat: Reduce blurriness of container shadows against dark header background in VR

jkuelz:
### Description

The main purpose of this commit is to reduce the "blurriness" or "halo" effect when a container shadow overlaps a dark header  in VR (AWSUI-19630). See screenshot below for before/after (**note:** the previous is on the _right_, the updated is on the _left_):

github-vr-shadows-light-screenshot

The way it accomplishes this is by adding `mix-blend-mode: multiply` to container shadows. In order to ensure the `mix-blend-mode` only effects the container shadow and not the content inside, the shadow has been moved to a pseudo element (`::after`) so that it can be treated as it’s own layer.

With this change, the container border has also been moved to a separate pseudo element (`::before`), so it can receive a different stacking context from the shadow and not blend with the background. This largely removes the need to account for the border in sticky header offset calculations, as well as padding values in selected cards.

As a result, this commit also removes some outdated code that was originally needed for IE11 support, as well as some border calculations that are no longer needed.

Lastly, this change slightly increases the contrast of VR shadows in dark mode to make the container perimeter more noticeable, which has also been repeated feedback (AWSUI-16528). The updates have been reviewed and approved by visual design. **Note:** the previous is on the _right_, the updated is on the _left_:

Screen Shot 2023-02-02 at 11 59 37 PM

Related links, issue #, if available: AWSUI-19630, AWSUI-16528

### How has this been tested?

Ran through the pipeline; checked screenshots in Chrome, Firefox, Safari. Also reviewed with visual designer to approve shadow changes.

See changes in pipeline: AwsUi-v3-jkuelz (approval attempt: 314060285, 314765186)

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/cards/__integ__/sticky-header.test.ts
```diff
@@ -47,7 +47,7 @@ describe('Cards Sticky Header', () => {
   const toggleVerticalOffsetBtn = '#toggle-vertical-offset-btn';
   const overflowParentPageHeight = 300;
   const verticalOffset = 50;
-  const containerBorder = 1;
+  const containerBorder = CONTAINER_ROOT_BORDER;
```
michaeldowseza, line 50: This const seems a bit redundant. What's the different between border and root border?

src/app-layout/__integ__/app-layout-sticky-elements.test.ts
```diff
@@ -67,7 +68,7 @@ test(
       await page.windowScrollTo({ top: 200 });
       const { bottom: pageHeaderBottom } = await page.getBoundingBox('header');
       const { top: tableHeaderTop } = await page.getBoundingBox(page.findStickyTableHeader().toSelector());
-      expect(tableHeaderTop).toEqual(pageHeaderBottom - 1);
+      expect(tableHeaderTop).toEqual(pageHeaderBottom - CONTAINER_ROOT_BORDER);
```
michaeldowseza, line 71: Can't we just remove this since we subtract 0?

src/cards/__integ__/sticky-header.test.ts
michaeldowseza: The way this variable is used is also a bit odd since we add and subtract 0?

src/container/__integ__/awsui-applayout-sticky.test.ts
```diff
@@ -16,7 +16,7 @@ const stickyToggleSelector = createWrapper().findFlashbar().findItems().get(1).f
 const headerSelector = '#b #h';
 const scrollableDivSelector = '#scrollable-div';
 
-const CONTAINER_ROOT_BORDER = 1;
+const CONTAINER_ROOT_BORDER = 0;
```
michaeldowseza, line 19: Sane as other comments of this could be removed.

michaeldowseza (commented):
Added some comments . A lot of the consts now change to a 0px offset and can probably be removed.

src/cards/__integ__/sticky-header.test.ts
jkuelz: I think one was for setting the default position in the overflow container, and the other was an offset correction for the table itself, but I agree it seems a bit redundant. I removed both, as well as the code that required the correction in the next revision.

src/app-layout/__integ__/app-layout-sticky-elements.test.ts
jkuelz: We can now. Originally was there because we had [this code](https://github.com/cloudscape-design/components/blob/40a405bd591436179a2e1e9a4287664bc7f6582c/src/container/use-sticky-header.ts#L25-L33) that checks for the container and header borders and accounts for it in sticky header. Removed both in next revision because borders no longer add to container height.

src/container/__integ__/awsui-applayout-sticky.test.ts
jkuelz: Repasting same response for visibility - Originally was there because we had [this code](https://github.com/cloudscape-design/components/blob/40a405bd591436179a2e1e9a4287664bc7f6582c/src/container/use-sticky-header.ts#L25-L33) that checks for the container and header borders and accounts for it in sticky header. Removed both in next revision because borders no longer contribute to container height.

michaeldowseza (approved):

PR #705, opened 2023-02-03
chore: Decouple metrics utils from package settings

pan-kot:
### Description

Reorganised metrics and base-component code to remove dependencies on package source and version. That is to move the corresponding code to the component-toolkit (as /internal/base-component and /internal/metrics) to be shared with other projects.

### How has this been tested?

Existing tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/internal/metrics/index.ts
```diff
@@ -0,0 +1,25 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
pan-kot, line 1: In index.ts we inject PACKAGE_VERSION and THEME to metrics utils. This allows all code in internal/metrics except for index.ts to be shared via the component-toolkit.

src/internal/metrics/init-awsui-versions.ts
```diff
@@ -0,0 +1,21 @@
...
+}
+declare const window: CustomWindow | undefined;
+
+// expose version info, so it can be checked using the browser devtools
+export function initAwsUiVersions(packageVersion: string) {
+  if (typeof window !== 'undefined') {
+    if (!window.awsuiVersions) {
+      window.awsuiVersions = {};
+    }
+    if (!window.awsuiVersions.components) {
```
just-boris, line 16: You probably want to make `components` part configurable too. To allow `{components: [], boards: []}` structure

src/internal/metrics/index.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export const metrics = new Metrics(PACKAGE_VERSION);
+export const metricsTestHelper = new MetricsTestHelper();
+
+export function initAwsUiVersions() {
+  awsuiVersions.initAwsUiVersions(PACKAGE_VERSION);
+}
+
+export function useComponentMetrics<T = any>(componentName: string) {
+  const theme = useVisualRefresh() ? 'vr' : THEME;
```
just-boris, line 18: This condition is a trait of all components in our system, including new the package. 

Actually, I would extract the original `useBaseComponent` hook into the shared utils, because it contains all features we need for all components. You may keep the hooks separately in the implementation, but the utility API should expose `useBaseComponent`

src/internal/base-component/index.ts
```diff
@@ -3,24 +3,9 @@
 
 // these styles needed to be imported for every public component
 import './styles.css.js';
-import { PACKAGE_VERSION } from '../environment';
+import { initAwsUiVersions } from '../metrics';
```
just-boris, line 6: Just in case, this object is not really used in any metrics, so putting it into that folder is misleading

src/internal/metrics/metrics.ts
```diff
@@ -0,0 +1,127 @@
...
+
+  /*
+   * Reports a metric value 1 to Console Platform's client logging service to indicate that the
+   * component was loaded. The component load event will only be reported as used to client logging
+   * service once per page view.
+   */
+  logComponentLoaded() {
+    this.sendMetricObjectOnce(
+      {
+        source: 'components',
```
just-boris, line 90: This needs to be parametrised, "components" or "boards"

src/__tests__/use-telemetry-usage.test.tsx
```diff
@@ -24,7 +24,7 @@ const verifyMetricsAreLoggedOnlyOnce = () => {
 };
 
 const getExpectedMetricName = (lowerCaseComponentName: string) => {
-  return `awsui_${lowerCaseComponentName}_${MetricsTestHelper.formatVersionForMetricName(THEME, PACKAGE_VERSION)}`;
+  return `awsui_${lowerCaseComponentName}_${metricsTestHelper.formatVersionForMetricName(THEME, PACKAGE_VERSION)}`;
```
just-boris, line 27: Since it is now a separate module, it can be imported directly, no need to transfer it via test helper.

Metrics test helper need only maintain `resetOneTimeMetricsCache`, because it an internal state. Everything else are pure functions, can be imported/exported directly

src/internal/metrics/init-awsui-versions.ts
pan-kot: Oh, I kind of thought we'd like to keep components for both. But makes sense as the package versions can be different - will update 👍

src/internal/metrics/index.ts
pan-kot: Ok, the only difference between `useComponentMetrics` and `useBaseComponent` is the theme and package version injection. The package injection will still happen from outside the toolkit but the theme can be handled of course. The downside is that we'd have to move or copy at least the `useVisualRefresh` hook to the toolkit to support that.

src/internal/base-component/index.ts
pan-kot: Makes sense, I'll create a separate folder for that.

src/__tests__/use-telemetry-usage.test.tsx
pan-kot: Makes sense

src/alert/index.tsx
```diff
@@ -4,7 +4,7 @@ import React from 'react';
 import { applyDisplayName } from '../internal/utils/apply-display-name';
 import { AlertProps } from './interfaces';
 import InternalAlert from './internal';
-import useBaseComponent from '../internal/hooks/use-base-component';
+import { useBaseComponent } from '../internal/base-component';
```
just-boris, line 7: Any chances of not changing this in the current PR? The import path will change to `@cloudscape-design/component-toolkit` soon, so let's keep the current ones for now

src/internal/base-component/telemetry.ts
```diff
@@ -0,0 +1,19 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { useEffect } from 'react';
+import { Metrics } from '../metrics/metrics';
+
+export function useTelemetry(metrics: Metrics, componentName: string, theme: string) {
```
just-boris, line 7: Why is this injected? In all use-cases, we use the same metrics provider, never changes dynamicaly

src/internal/base-component/index.ts
```diff
@@ -1,49 +1,34 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+import { ALWAYS_VISUAL_REFRESH, PACKAGE_VERSION, THEME } from '../environment';
+import { metrics } from '../metrics';
+import { initAwsUiVersions } from './init-awsui-versions';
+import { useVisualRefresh } from './use-visual-refresh.js';
```
just-boris, line 7: I have lost tracking what is happening with all file moves. Let's get back to the task

1. `useBaseComponent` changes its location from this package to `@cloudscape-design/component-toolkit/internal`
2. Because it has dependency on `useVisualRefresh` hook, that hook moves to `@cloudscape-design/component-toolkit/internal` too.

After the move, we replace all imports in this in this package with their equivalents from `@cloudscape-design/component-toolkit/internal`.

Why is it needed to move that many files around? I am afraid that if we merge it the current way, the very next PR would require to undo some of the changes

pan-kot: Thanks for the feedback - I will revert some changes to reduce the amount of diff and annotate the code that can be moved to the toolkit.

src/internal/base-component/telemetry.ts
pan-kot: The metrics object is instantiated with package source and package version. If the metrics will be instantiated here the function will no longer be suitable to move to the toolkit.

src/alert/index.tsx
pan-kot: Yes, will update.

src/internal/base-component/telemetry.ts
just-boris: Discussed in DM, proposed solution is to provide all variables via arguments

```
function useTelemetry(package: string, version: string, componentName: string, theme: string)
```

src/internal/utils/__tests__/metrics.test.ts
```diff
@@ -1,8 +1,9 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
-import { Metrics } from '../../metrics';
 
-jest.mock('../../environment', () => ({ PACKAGE_VERSION: '3.0 (HEAD)' }), { virtual: true });
```
just-boris, line 5: Removing module mocks is always good. 👍

just-boris (approved):
Approved, but I would wait for @michaeldowseza approval too, in case if there are conflicts with his work in progress

PR #707, opened 2023-02-03
chore: Split linting and unlocking of package-lock script

michaeldowseza:
### Description

Npm8 changed the preinstall script behaviour to run after the npm install event but before the package install script. This differs from < Npm8 where the preinstall script would also run before the npm install event. 

The changes here split our regular package-lock check and our unlocking of the package-lock dependencies. This is to cover the 2 separate use cases:

Happy cases
1. A developer checks out the package where the package-lock file does not contain any cloudscape-design dependencies. They run `npm install` and receive the latest versions which will also now update the package-lock to include these dependencies. The postinstall script will run the unlock-dependencies script to remove the dependencies. This allows the developer to run npm install again without any issues.
2. When committing a change the precommit hook will run and check if the developer has mistakenly committed a package-lock that contains a codeartifact url and then will run the unlock-dependencies script to remove any cloudscape-design dependencies that still somehow ended up there. This produces the first happy case above for the next developer.

Edge cases
1. A developer bypasses the precommit hook and manages to get their PR merged with the dependency. The above change to the Github action will allow them to continue to work as expected. 

To handle the Github workflows this has been moved to the shared .github repository to run as a node script before the install. See [pull-request](https://github.com/cloudscape-design/.github/pull/34) for more details.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


package.json
```diff
@@ -18,7 +18,7 @@
     "lint": "eslint --ignore-path .gitignore --ext ts,tsx,js . && stylelint --ignore-path .gitignore '{src,pages}/**/*.{css,scss}'",
     "start": "gulp watch & NODE_ENV=development webpack serve --config pages/webpack.config.js",
     "start:integ": "NODE_ENV=development webpack serve --config pages/webpack.config.integ.js",
-    "preinstall": "./scripts/prepare-package-lock.js",
+    "postinstall": "node ./scripts/unlock-package-lock.js",
```
michaeldowseza, line 21: There is no longer a true preinstall script. Using postinstall here to be more clear about when this is running.

just-boris: Why do we need to run it in CI at all? Wouldn't https://github.com/cloudscape-design/.github/pull/34 change be enough to solve the task?

scripts/lint-package-lock.js
```diff
@@ -0,0 +1,38 @@
...
+}
+
+function checkDependencies(dependencyName, dependencies) {
+  const dependency = dependencies[dependencyName];
+
+  if (dependencyName.includes('@cloudscape-design/')) {
+    throw Error('package-lock.json file contains @cloudscape-design dependency');
+  }
+
+  if (dependency && dependency.resolved && dependency.resolved.includes('codeartifact.us-west-2.amazonaws.com')) {
```
just-boris, line 27: How about removing this script completely (keep only `unlock-package-lock.js`)

if lockfile contains any invalid registries, the build will fail regardless

package.json
michaeldowseza: This is for developers for when they run install and if they don't do a commit where the package-lock can get linted then the package versions would be locked the next time they try to pull the latest updates. Unless I'm just misunderstanding npm here?

just-boris: I am not sure about lifecycle order here. I recall npm wrote the lockfile after our script runs.

I can test it again after merging this

michaeldowseza: Hmm, you're right here actually. In < npm8 we used to hook into post shrinkwrap to be able to modify the package-lock 🤔 So then yes this line becomes useless and can be removed.

just-boris (approved):

PR #715, opened 2023-02-06
fix: Flashbar bottom spacing management logic

jperals:
#### Background

The Stacked notifications feature adds an element called "notifications bar" to the Flashbar, which partially overlaps the Flashbar and partially overlaps the content below. We don't want it to create extra space with the content below (e.g, sticky table headers should still collapse without any space between the flashbar and the table), but at the same time, there should be some space with the window bottom in order to prevent the notifications bar from getting cropped (there is a function for that called `updateBottomSpacing`).

Right:
Screenshot 2023-02-06 at 14 38 40
Wrong:
Screenshot 2023-02-06 at 14 41 20

#### The fix

This change fixes a few issues that were making this behavior not totally reliable:
- Account for extra margin of the parent notifications region when using App Layout with Visual Refresh
- The extra padding was being applied to non-sticky Flashbar, but it should be applied only to the sticky one
- Rounding issues

### How has this been tested?

- Added integration test. A new testing page for the sticky behavior had to be added for this
- Manually tested with: Chrome, Firefox and Safari: sticky and non-sticky; Visual Refresh and Classic

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/flashbar/common.ts
+++ b/pages/flashbar/common.ts
@@ -1,0 +1,38 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { FlashbarProps } from '~components';
+
+export const i18nStrings = {
+  ariaLabel: 'Notifications',
+  notificationBarText: 'Notifications',
+  notificationBarAriaLabel: 'View all notifications',
+  errorIconAriaLabel: 'Error',
+  successIconAriaLabel: 'Success',
+  warningIconAriaLabel: 'Warning',
+  infoIconAriaLabel: 'Information',
+  inProgressIconAriaLabel: 'In progress',
+};
+
+const RANDOM_NUMBER_RANGE = 1000;
+
+export function generateItem(
+  type: FlashbarProps.Type,
+  dismiss: (index: string) => void,
+  hasHeader = false,
+  initial = false
+): FlashbarProps.MessageDefinition {
+  const randomKey = `key_${Math.floor(Math.random() * RANDOM_NUMBER_RANGE)}`;
+  return {
+    type,
+    id: randomKey,
+    dismissible: true,
+    dismissLabel: 'Dismiss',
+    onDismiss: () => dismiss(randomKey),
+    buttonText: 'Do Action',
+    statusIconAriaLabel: 'Info',
+    content: `This is a flash item with key ${randomKey.split('_').join(' ')}`,
+    ariaRole: initial ? undefined : type === 'error' ? 'alert' : 'status',
+    ...(hasHeader && { header: 'Has Header Content' }),
+  };
+}
--- a/src/app-layout/notifications/styles.scss
+++ b/src/app-layout/notifications/styles.scss
@@ -4,9 +4,11 @@
 */
 
 @use '../../internal/styles/tokens' as awsui;
+@use '../../internal/generated/custom-css-properties/index.scss' as custom-props;
 
 .notifications-sticky {
   top: 0;
   z-index: 825;
   position: sticky;
+  #{custom-props.$flashbarStickyBottomMargin}: #{awsui.$space-xxl};
 }
--- a/src/app-layout/notifications/styles.scss
+++ b/src/app-layout/notifications/styles.scss
@@ -4,7 +4,7 @@
 */
 
 @use '../../internal/styles/tokens' as awsui;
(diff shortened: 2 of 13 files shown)


src/app-layout/notifications/styles.scss
```diff
@@ -4,9 +4,11 @@
 */
 
 @use '../../internal/styles/tokens' as awsui;
+@use '../../internal/generated/custom-css-properties/index.scss' as custom-props;
```
fa7ad, line 7: We generally don't use file extensions in imports, is this necessary?

src/flashbar/collapsible-flashbar.tsx
```diff
@@ -105,16 +105,17 @@ export default function CollapsibleFlashbar({ items, ...restProps }: FlashbarPro
...
           // then we might decide to remove it or not.
-          flashbar.classList.add(styles['spaced-bottom']);
+          flashbar.classList.remove(styles.floating);
+          const windowHeight = window.innerHeight;
+          // Take the parent region into account if using the App Layout, because it might have additional margins.
+          // Otherwise we use the Flashbar component for this calculation.
+          const outerElement = flashbar.parentElement?.parentElement || flashbar;
           const applySpacing =
-            isFlashbarStackExpanded && bottom + parseInt(getComputedStyle(flashbar).paddingBottom) >= windowHeight;
+            isFlashbarStackExpanded && Math.ceil(outerElement.getBoundingClientRect().bottom) >= windowHeight;
```
fa7ad, line 116: Not a big deal, but `parseInt` behaves more similar to `Math.floor` than `Math.ceil`

src/flashbar/__integ__/collapsible.test.ts
```diff
@@ -129,4 +130,22 @@ describe('Collapsible Flashbar', () => {
       );
     });
   });
+
+  describe('Sticky Flashbar', () => {
+    test(
+      'keeps a space to the screen bottom to prevent the notification bar from getting cropped',
+      setupStickyFlashbarTest(async page => {
```
fa7ad, line 137: [non-blocker] Would be nice to have tests on multiple screen sizes (mobile, desktop, tablet, etc).

src/app-layout/notifications/styles.scss
```diff
@@ -4,9 +4,11 @@
...
 
 .notifications-sticky {
   top: 0;
   z-index: 825;
   position: sticky;
+  #{custom-props.$flashbarStickyBottomMargin}: #{awsui.$space-xxl};
```
fa7ad, line 13: How does this work? 
Does the left side get replaced with the name for a CSS custom property/variable (`--something-like-this`)?

fa7ad (commented):
Left some minor comments, no major/blockers issues.

src/flashbar/collapsible-flashbar.tsx
jperals: I use `Math.ceil` on purpose because the behavior will still be fine if the calculation yields a too high value, but it will flicker if it's too low.

src/app-layout/notifications/styles.scss
jperals: Yes. Having a custom property directly written here is not allowed by our linting, so I added a new one `flashbarStickyBottomMargin` in `internal/generated/custom-css-properties/list.js` (this file itself doesn't seem to be generated, but it is used for the generation of custom properties)

src/app-layout/notifications/styles.scss
jperals: Removed

fa7ad (approved):

pages/flashbar/common.ts
```diff
@@ -0,0 +1,38 @@
...
+
+const RANDOM_NUMBER_RANGE = 1000;
+
+export function generateItem(
+  type: FlashbarProps.Type,
+  dismiss: (index: string) => void,
+  hasHeader = false,
+  initial = false
+): FlashbarProps.MessageDefinition {
+  const randomKey = `key_${Math.floor(Math.random() * RANDOM_NUMBER_RANGE)}`;
```
pan-kot, line 25: We have a util you can use: https://github.com/cloudscape-design/components/tree/main/src/internal/hooks/use-unique-id

jperals: Follow-up micro-PR: https://github.com/cloudscape-design/components/pull/724

PR #709, opened 2023-02-03
refactor: Remove unused code in app layout

just-boris:
### Description

Remove unused code. See comments inline with more details

Related links, issue #, if available: n/a

### How has this been tested?

Locally and my dev-pipeline 

<details>
   <summary>Review checklist</summary>

_The following items are to be evaluated by the author(s) and the reviewer(s)._

#### Correctness

- _Changes include appropriate documentation updates._
- _Changes are backward-compatible if not indicated, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#public-apis)._
- _Changes do not include unsupported browser features, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#browsers-support)._
- _Changes were manually tested for accessibility, see [accessibility guidelines](https://cloudscape.design/foundation/core-principles/accessibility/)._

#### Security

- _If the code handles URLs: all URLs are validated through [the `checkSafeUrl` function](https://github.com/cloudscape-design/components/blob/main/src/internal/utils/check-safe-url.ts)._

#### Testing

- _Changes are covered with new/existing unit tests?_
- _Changes are covered with new/existing integration tests?_
</details>

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #719, opened 2023-02-07
fix: Keep focused button in viewport after collapsing Flashbar

jperals:
### Description

When collapsing a non-sticky notifications stack, the entire Flashbar might go up beyond the viewport, causing the focused button to be lost from sight. This change catches that case and scrolls it into view when that happens.

### How has this been tested?

- Unit tests
- Manual tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/flashbar/collapsible-flashbar.tsx
```diff
@@ -139,16 +146,29 @@ export default function CollapsibleFlashbar({ items, ...restProps }: FlashbarPro
...
+
       animate({
         elements: getElementsToAnimate(),
         oldState: initialAnimationState,
         newElementInitialState: ({ top }) => ({ scale: 0.9, y: -0.2 * top }),
         onTransitionsEnd: () => setTransitioning(false),
       });
+
+      if (shouldScrollUp) {
+        window.scrollTo({ top: 0 });
```
just-boris, line 165: If your task is to scroll button into viewport, use `button.scrollIntoViewport` ([our wrapper with better defaults](https://github.com/cloudscape-design/components/blob/657d369dc08468d2066cd64b12ad93804d6cc911/src/internal/utils/scrollable-containers.ts#L76-L81)) and not manual scroll on the window. There could be multiple scrollable containers in a way, this code is bug-prone.


-- 

Also, why is there a conditionality on keyboard/mouse interactions? They both should receive the same scroll treatment

jperals: Consulting with a11y and did both changes:
- Use `scrollIntoViewport` with existing function
- Apply the same behavior with mouse as with keyboard

Manually tested again and updated unit tests accordingly.

src/flashbar/collapsible-flashbar.tsx
```diff
@@ -139,16 +146,28 @@ export default function CollapsibleFlashbar({ items, ...restProps }: FlashbarPro
...
+        isElementTopBeyondViewport(notificationBarRef.current);
+
       animate({
         elements: getElementsToAnimate(),
         oldState: initialAnimationState,
         newElementInitialState: ({ top }) => ({ scale: 0.9, y: -0.2 * top }),
         onTransitionsEnd: () => setTransitioning(false),
       });
+
+      if (notificationBarRef.current && shouldScrollUp) {
```
just-boris, line 163: Is "shouldScrollUp" check really needed? If it is outside the viewport, where else could this be?

Why can you not simplify the code down to only essential part?

jperals: Calling `scrollElementIntoViewport` on an element will scroll it to the top even if it is already inside the viewport (just verified this behavior manually on Chrome). I prefer to scroll only if it is not visible mainly because this is more similar to the behavior of the native keyboard navigation. Also, there are already quite a lot of things moving when you collapse, so I would try to not add any more visual change that is not really necessary.

just-boris: Did you test native `scrollIntoView` implementation or our helper, `scrollElementIntoView`? As I mentioned in the past comment: https://github.com/cloudscape-design/components/pull/719#discussion_r1100118357, our version has better defaults.

It was specifically created to allow you to write simpler code

src/flashbar/collapsible-flashbar.tsx
```diff
@@ -140,16 +147,28 @@ export default function CollapsibleFlashbar({ items, ...restProps }: FlashbarPro
...
+        scrollElementIntoView(notificationBarRef.current);
+      }
+
       setTransitioning(true);
       setInitialAnimationState(null);
     }
-  }, [updateBottomSpacing, getElementsToAnimate, initialAnimationState, isFlashbarStackExpanded]);
+  }, [updateBottomSpacing, getElementsToAnimate, initialAnimationState, isFlashbarStackExpanded, isFocusVisible]);
```
just-boris, line 171: Dependency on `isFocusVisible` is no longer needed

src/flashbar/collapsible-flashbar.tsx
jperals: I tested the native implementation on the browser first, and then our function by unconditionally calling the it within the component. In both cases the element was scrolled to the top even if it was already present. And I couldn't see any option in the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) that suggests it would change the behavior to not scroll if the element is already visible.

just-boris: `block: 'nearest'` is supposed to do exactly what you expect. It did it in all other places where this helper is used.

If you do not want to investigate it further, fine I can approve.

But you are also giving me a chance of creating PRs removing hundreds lines of unused code (like this one https://github.com/cloudscape-design/components/pull/709) when I actually do proper investigation instead of you

src/flashbar/collapsible-flashbar.tsx
jperals: Right, removed

just-boris (approved):

src/flashbar/collapsible-flashbar.tsx
jperals: I must have done something wrong while checking this (maybe it scrolled just because the page became shorter?):
> and then our function by unconditionally calling the it within the component

I checked this again and it didn't scroll. So I made a follow-up PR: [and then our function by unconditionally calling the it within the component](https://github.com/cloudscape-design/components/pull/734)

just-boris: Great! Appreciate a deeper dive :)

PR #724, opened 2023-02-08
chore: reuse unique ID function to de-duplicate logic

jperals:
### Description

Small follow-up from #715 (see [comment](https://github.com/cloudscape-design/components/pull/715#discussion_r1099963151)): reuse existing `generateUniqueId` logic instead of creating a new one.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/flashbar/common.ts
+++ b/pages/flashbar/common.ts
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { FlashbarProps } from '~components';
+import { generateUniqueId } from '~components/internal/hooks/use-unique-id';
 
 export const i18nStrings = {
   ariaLabel: 'Notifications',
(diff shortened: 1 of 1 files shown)


Al-Dani:
not sure, how much cense is in the following using of the random key on the line 33: 

> content: `This is a flash item with key ${randomKey.split('_').join(' ')}`,

But if it's ok from the context, then there is nothing more from my side

Al-Dani (approved):

jperals:
@Al-Dani Sorry I saw your message too late. Yes, not sure it makes a lot of sense to do that transformation for rendering, but I don't think it really matters (in the sense that it doesn't break anything) and I preferred to keep the changes minimal. The original logic was not added in #715 but just moved around.

just-boris:
If this value ends up in screenshot tests (where we use props to identify "old" and "new" images, it is going to be a problem).

For safe screenshot tests, we have this helper: https://github.com/cloudscape-design/components/blob/main/pages/utils/pseudo-random.ts

It gives you non-consequent, but deterministic numbers on every run.

--

Also, is there any reason why you cannot use indexes, 1, 2, 3, 4, 5, etc?

jperals:
I am actually considering using this page for screenshot tests right now, so I might do just that (most likely using the simpler index approach). Thanks for the suggestions!

PR #725, opened 2023-02-08
chore: Use wizard id for funnel id in metrics

michaeldowseza:
### Description

Pass the wizard id, that is optional and deprecated, to the custom metric event as the funnel id. This is a best effort attempt for grouping metrics by an id as there is currently no other identifier available. 

### How has this been tested?

Added a base unit test.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/wizard/index.tsx
+++ b/src/wizard/index.tsx
@@ -34,6 +34,7 @@ export default function Wizard({
 }: WizardProps) {
   const { __internalRootRef } = useBaseComponent('Wizard');
   const baseProps = getBaseProps(rest);
+  const funnelId = baseProps.id;
 
   const [breakpoint, breakpointsRef] = useContainerBreakpoints(['xs']);
   const ref = useMergeRefs(breakpointsRef, __internalRootRef);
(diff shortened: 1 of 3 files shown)


src/wizard/index.tsx
```diff
@@ -54,7 +55,7 @@ export default function Wizard({
   const isLastStep = actualActiveStepIndex >= steps.length - 1;
 
   const navigationEvent = (requestedStepIndex: number, reason: WizardProps.NavigationReason) => {
-    trackNavigate(actualActiveStepIndex, requestedStepIndex, reason);
+    trackNavigate(actualActiveStepIndex, requestedStepIndex, reason, funnelId);
```
jperals, line 58: Not blocking but I would consider using named parameters vs positional. Makes it easier to read functions calls IMO.

jperals (approved):

src/wizard/index.tsx
michaeldowseza: This is likely to change in the future anyway so I can re evaluate it again then :)

PR #729, opened 2023-02-08
chore: Prepare interactive Flashbar page for visual regression testing

jperals:
### Description

This change makes the following changes to the `flashbar/interactive` page to make it suitable for visual regression testing:
- Deterministic content (notification IDs were being generated randomly)
- Wrap the Flashbar in a screenshot area and adequate padding

### How has this been tested?

Tested by running visual regression tests and getting the expected results, including the screenshots.

<details>

#### Correctness

#### Security

#### Testing

</details>

---


pages/flashbar/common.ts
```diff
@@ -15,22 +15,29 @@ export const i18nStrings = {
...
   return {
     type,
-    id: randomKey,
+    id,
     dismissible: true,
     dismissLabel: 'Dismiss',
-    onDismiss: () => dismiss(randomKey),
+    onDismiss: () => dismiss(id),
     buttonText: 'Do Action',
+    onButtonClick: () => null,
```
jperals, line 38: Adding `onButtonClick` gets rid of a console warning about passing `buttonText` but not this other property.

pages/flashbar/interactive.page.tsx
```diff
@@ -3,34 +3,60 @@
...
   const dismiss = (index: string) => {
     setItems(items => items.filter(item => item.id !== index));
   };
 
   const add = (type: FlashbarProps.Type, hasHeader = false) => {
-    setItems(items => [generateItem(type, dismiss, hasHeader), ...items]);
+    setNextId(id => {
+      setItems(items => [generateItem({ type, id: id.toString(), dismiss, hasHeader }), ...items]);
+      return id + 1;
+    });
```
michaeldowseza, line 19: I might need an opinion from someone with a lil' more React knowledge but this looks a bit odd. Is it fine to update state from within here?

jperals: This was to make sure `items` are updated using the current `nextId`. In other words, to make the state update operation more atomic.

Otherwise I was running into a situation where, if for example `add` was being called from another function, a JS closure was being created that led to calling `add` with an old value of `nextId`.

jperals: Here is an example using this approach: https://stackoverflow.com/a/62806870

jperals: Here is the page (thanks for the deploy feature!!) if you want to interact with it: https://d21d5uik3ws71m.cloudfront.net/components/aed5a5b9f5cedae2a51004d37b3958864b45f10a/index.html#/light/flashbar/interactive

jperals: It seems a more conventional approach, with two updates, also works fine as long as I don't wrap both calls in a `setTimeout`:

```
setItems(items => [generateItem({ type, id: nextId.toString(), dismiss, hasHeader }), ...items]);
setNextId(id => id + 1);
```

I can change that if you think it's more straightforward/conventional, I have no strong preference.

just-boris: `nextId` never used in rendering, it is only a counter. You should use ref for such values.

This way you will stay away from possible undocumented behaviors

jperals: Thanks, this works well. Updated

pages/flashbar/common.ts
just-boris: `buttonText`/`onButtonClick` API is deprecated, as you can see in API docs

If you want to fix it, replace `buttonText` with `action: <Button>Do action</Button>`

pages/flashbar/interactive.page.tsx
```diff
@@ -67,7 +85,11 @@ export default function InteractiveFlashbar() {
           </Button>
           <Button onClick={() => removeAndAddToBottom('error')}>Add And Remove</Button>
         </SpaceBetween>
-        <Flashbar items={items} {...restProps} />
+        <ScreenshotArea>
```
just-boris, line 88: Best practice is to separate functional and screenshot testing pages

Future people will appreciate that adding an extra example to functional tests will not require manual approval in pipeline, because a screenshot test got affected

just-boris (commented):
The comment about `buttonText` is optional, but the request about separate page is important, because merging this page structure will cause friction on our pipeline

pages/flashbar/collapsible.visual-tests.page.tsx
```diff
@@ -0,0 +1,43 @@
...
+    ...sampleNotifications.success,
+    id: '2',
+  },
+];
+
+export default function CollapsibleFlashbarVisualTests() {
+  return (
+    <>
+      <h1>Collapsible Flashbar visual tests</h1>
+      <FocusTarget />
```
jperals, line 22: This element allows to expand the Flashbar with the keyboard so that the focus ring is also covered in the tests.

pages/flashbar/interactive.page.tsx
jperals: Done, thanks for the suggestion. I added `collapsible.visual-tests.page.tsx`.

just-boris (approved):

PR #735, opened 2023-02-09
fix: Always set role="dialog" on popover with dismiss button

avinashbot:
### Description

The current state is: if a popover has no header, it's not marked as a dialog, even if it traps focus. Not great.

This was originally changed to ensure popover dialogs always have valid titles, but an unlabeled dialog is better (well, less terrible) than a non-dialog focus trap. We may need to allow users to provide assistive tech only aria-labels for dialog headings in the future, but I'll wait for customer data before I do that; an explicit heading still provides an equal experience for everyone, and screen readers still read out unlabeled dialog contents.

So we have popovers acting as:
- Announcements: (e.g. "Text copied", no dismiss button, just aria-live)
- Modal dialogs: (i.e. has a dismiss button, should have a header)
- Non-modal dialogs: (i.e. annotation popovers, you can tab straight through them)

Related links, issue #, if available: doesn't fix, but comes from AWSUI-20299

### How has this been tested?

Changed unit tests.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/popover/__tests__/popover.test.tsx
+++ b/src/popover/__tests__/popover.test.tsx
@@ -203,17 +203,6 @@ describe('ARIA labels', () => {
     expect(wrapper.findBody()!.getElement()).not.toHaveAttribute('aria-labelledby');
   });
 
-  it('sets aria-labelledby when dismiss button is not present', () => {
-    const wrapper = renderPopover({
-      children: 'Trigger',
-      content: 'Popover',
-      header: 'Hello!',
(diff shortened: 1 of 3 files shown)


gethinwebster:
if we don't think that this pattern should actually be used, would it be better to discourage people more clearly, e.g. have some dev-time warnings?

gethinwebster (approved):

PR #744, opened 2023-02-10
fix: Ensure autosuggest `enteredTextLabel` is accessible

gethinwebster:
### Description

The `enteredTextLabel` (e.g. "Use ...") item previously got a screenreader override that didn't contain the appropriate message. This fixes that.

Related links, issue #, if available: AWSUI-20334

### How has this been tested?

Tested DOM output, and checked for expected announcements in VO & Chrome.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/autosuggest/__tests__/autosuggest.test.tsx
+++ b/src/autosuggest/__tests__/autosuggest.test.tsx
@@ -5,6 +5,7 @@ import { render } from '@testing-library/react';
 import createWrapper from '../../../lib/components/test-utils/dom';
 import Autosuggest, { AutosuggestProps } from '../../../lib/components/autosuggest';
 import styles from '../../../lib/components/autosuggest/styles.css.js';
+import itemStyles from '../../../lib/components/internal/components/selectable-item/styles.css.js';
 import { KeyCode } from '@cloudscape-design/test-utils-core/utils';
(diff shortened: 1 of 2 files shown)


just-boris:
Can we resolve it somehow on [useAnnouncement hook](https://github.com/cloudscape-design/components/blob/main/src/select/utils/use-announcement.ts#L43) level and not emit incorrect suggestion in the first place?

I am afraid that this value could be mistakenly used in other places (maybe not in the current code, but with another accidental refactoring)

gethinwebster:
> Can we resolve it somehow on [useAnnouncement hook](https://github.com/cloudscape-design/components/blob/main/src/select/utils/use-announcement.ts#L43) level and not emit incorrect suggestion in the first place?

I'm not sure that would be the right place, as the hook doesn't know about the different Autosuggest 'types' of item. I chose here because it's the same place that the `'use-entered'` item content is overridden, and it seems like these two bits of logic belong together (it's _because_ we override the content to not just display the `value` that we want to also override the screen reader behavior).

just-boris:
I am not a big fan that incorrect `screenReaderContent` value is hanging around our code, but if you think it is a better solution let it be this way

just-boris (approved):

PR #745, opened 2023-02-10
chore: Emit metric when wizard is unmounted

michaeldowseza:
### Description

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/wizard/index.tsx
+++ b/src/wizard/index.tsx
@@ -16,7 +16,14 @@ import useBaseComponent from '../internal/hooks/use-base-component';
 import { useMergeRefs } from '../internal/hooks/use-merge-refs';
 import { useVisualRefresh } from '../internal/hooks/use-visual-mode';
 
-import { trackStartStep, trackNavigate, trackSubmit, trackStartWizard } from './internal/analytics';
+import {
+  trackStartStep,
+  trackNavigate,
+  trackSubmit,
+  trackStartWizard,
+  trackEndWizard,
+  trackCancel,
+} from './internal/analytics';
 
 export { WizardProps };
 
@@ -61,7 +68,10 @@ export default function Wizard({
   };
   const onStepClick = (stepIndex: number) => navigationEvent(stepIndex, 'step');
   const onSkipToClick = (stepIndex: number) => navigationEvent(stepIndex, 'skip');
-  const onCancelClick = () => fireNonCancelableEvent(onCancel);
+  const onCancelClick = () => {
+    trackCancel(actualActiveStepIndex, funnelId);
+    fireNonCancelableEvent(onCancel);
+  };
   const onPreviousClick = () => navigationEvent(actualActiveStepIndex - 1, 'previous');
   const onPrimaryClick = () => {
     if (isLastStep) {
@@ -90,6 +100,11 @@ export default function Wizard({
 
   useEffect(() => {
     trackStartWizard();
+
+    return () => {
+      trackEndWizard(funnelId);
+    };
+    // eslint-disable-next-line react-hooks/exhaustive-deps -- this is intentional
   }, []);
 
   useEffect(() => {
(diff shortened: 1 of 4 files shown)


pages/app/index.html
```diff
@@ -1,38 +1,42 @@
...
 
-<body>
-  <!-- #b element should be there for compatibility with global navigation -->
-  <div id="b">
-    <div id="app"></div>
-  </div>
-</body>
+      window.panorama = function (functionName, opts) {
+        console.log(functionName, opts);
+      };
```
michaeldowseza, line 32: The only real change in this file. The rest was auto formatted.

just-boris: Can we avoid changing this file? This script has to be inline, because it is the only supported way to configure CSP header.

For everything else it is better to use separate files.

michaeldowseza: We can avoid it if it can possibly cause problems. Will update PR.

just-boris: P.S. I am also not sure if about losing test coverage for our components behavior when `window.panorama` is undefined (which is a very common case).

Can we conditionally apply panorama mock only in particular scenarios?

src/wizard/index.tsx
```diff
@@ -90,6 +100,11 @@ export default function Wizard({
 
   useEffect(() => {
     trackStartWizard();
+
+    return () => {
+      trackEndWizard(funnelId);
```
just-boris, line 105: Why does "start" not receive an id but "end" does?

src/wizard/internal/analytics.tsx
```diff
@@ -36,10 +39,29 @@ const timeEnd = (key = 'current', clear = false) => {
...
 export const trackStartWizard = () => {
+  timeCache = {};
+  currentStepIndex = 0;
+  success = false;
+
   timeStart(prefix);
 };
 
-export const trackStartStep = (stepIndex?: number, funnelId?: string) => {
+export const trackEndWizard = (funnelId?: string) => {
```
just-boris, line 49:
```suggestion
export const trackEndWizard = (funnelId: string | undefined) => {
```

This way in stays optional in public API, but typescript will warn you if you forgot to pass it down from one internal function to another

src/wizard/internal/analytics.tsx
```diff
@@ -16,7 +16,10 @@ const createEventDetail = (stepIndex = 0) => `step${stepIndex + 1}`;
 // A custom time cache is used to not clear the timer between navigation attempts
 // This allows us the ability to track time to attempt each step as well as the time to complete
 // each step
-const timeCache: Record<string, number> = {};
+let timeCache: Record<string, number>;
+let currentStepIndex: number;
+let success: boolean;
```
just-boris, line 21: Not a big fan of these global vars.

Fine as a temporary solution, but probably should be scoped to  unique `funnelId`

src/wizard/index.tsx
michaeldowseza: Maybe trackStartWizard is incorrectly named as it is not sending a metric and only creates a timestamp for component mount and then trackStep will send the metric. Since there is no trackEndStep, trackEndWizard will create an end timestamp as well as send a metric.

src/wizard/internal/analytics.tsx
michaeldowseza: Agreed that it's not great and this will hopefully change in the long term.

src/wizard/internal/analytics.tsx
michaeldowseza: Updated 👍🏻

src/wizard/index.tsx
just-boris: This aligns with my another comment: https://github.com/cloudscape-design/components/pull/745#discussion_r1112894071

Let's keep the current state for now

just-boris (approved):

PR #750, opened 2023-02-13
feat: New truncate-heading prop for header

pan-kot:
### Description

Default heading wrapping behavior might not be applicable when used inside small containers. When new property is set, the heading is truncated with ellipsis instead.

When the heading is truncated some part of its information is lost (exception - screen-reader users). That means the feature can only be used when there is an alternative way to obtain it.

### How has this been tested?

Screenshot tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/header/interfaces.ts
+++ b/src/header/interfaces.ts
@@ -41,6 +41,11 @@ export interface HeaderProps extends BaseComponentProps {
    * Area next to the heading to display an Info link.
    */
   info?: React.ReactNode;
+  /**
+   * Determines if the heading (specified with the `children` property) overflows with ellipsis when there is not enough space.
+   * Defaults to `false`.
+   */
+  overflowHeading?: boolean;
 }
 
 export namespace HeaderProps {
--- a/src/header/interfaces.ts
+++ b/src/header/interfaces.ts
@@ -42,10 +42,10 @@ export interface HeaderProps extends BaseComponentProps {
    */
   info?: React.ReactNode;
   /**
-   * Determines if the heading (specified with the `children` property) overflows with ellipsis when there is not enough space.
+   * Determines if the heading (specified with the `children` property) should be truncated with ellipsis when there is not enough space.
    * Defaults to `false`.
    */
-  overflowHeading?: boolean;
+  truncateHeading?: boolean;
 }
 
 export namespace HeaderProps {
--- a/src/header/internal.tsx
+++ b/src/header/internal.tsx
@@ -9,6 +9,7 @@ import { useVisualRefresh } from '../internal/hooks/use-visual-mode';
 import { HeaderProps } from './interfaces';
 import styles from './styles.css.js';
 import { SomeRequired } from '../internal/types';
+import ScreenreaderOnly from '../internal/components/screenreader-only';
 
 interface InternalHeaderProps extends SomeRequired<HeaderProps, 'variant'>, InternalBaseComponentProps {
   __disableActionsWrapping?: boolean;
@@ -16,6 +17,7 @@ interface InternalHeaderProps extends SomeRequired<HeaderProps, 'variant'>, Inte
 
 export default function InternalHeader({
   variant,
+  overflowHeading,
   headingTagOverride,
   children,
   actions,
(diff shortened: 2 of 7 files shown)


src/header/interfaces.ts
```diff
@@ -41,6 +41,11 @@ export interface HeaderProps extends BaseComponentProps {
    * Area next to the heading to display an Info link.
    */
   info?: React.ReactNode;
+  /**
+   * Determines if the heading (specified with the `children` property) should be truncated with ellipsis when there is not enough space.
+   * Defaults to `false`.
+   */
+  truncateHeading?: boolean;
```
just-boris, line 48: Naming. 

1. We have similar props on button component (called `wrapText`) and table (`wrapLines`). So the new property should follow this pattern
2. The content of default slot is called `title` (see display name in the [API docs](https://github.com/cloudscape-design/components/blob/1b69f03545f9f0aaf88fb685e28efabf1ca05140/src/header/interfaces.ts#L10) and also [usage guidelines](https://cloudscape.design/components/header/?tabId=usage)).

Based on these two, I suggest `wrapTitle` for the name

just-boris: P.S. you may also look at API docs for respective components and reuse some of the wording, but I do not have much opinion there

src/header/internal.tsx
```diff
@@ -60,10 +62,20 @@ export default function InternalHeader({
...
+              truncateHeading && styles['heading-overflow']
+            )}
+          >
+            <span
+              className={clsx(styles['heading-text'], styles[`heading-text-variant-${variantOverride}`])}
+              aria-hidden={truncateHeading}
+            >
               {children}
             </span>
+            {truncateHeading && <ScreenreaderOnly>{children}</ScreenreaderOnly>}
```
just-boris, line 78: Why is this needed, what is wrong with the visible text?

`ScreenreaderOnly` component works as maximum truncation down to 1px, so what is wrong with relying on half-truncated content?

src/header/styles.scss
```diff
@@ -103,6 +103,10 @@
 }
 
 .title {
+  display: flex;
```
just-boris, line 106: Remembering the story with flexible containers (https://github.com/cloudscape-design/components/pull/603) can we make extra flex-styles opt-in for the new mode to reduce possible impact?

src/header/internal.tsx
DaemonCahill: I just bumped into this as well when I noticed that there are two DOM nodes with the same text when aria-hidden='true'.

Screen Shot 2023-02-13 at 4 05 34 PM

I don't think I understand what problem is solved by putting aria-hidden on the truncated node and then replicating the same text and rendering it off the canvas...

DaemonCahill:
It might be worthwhile to expand the screenshot testing of this property beyond just rendering headers.

qa

src/header/interfaces.ts
pan-kot: Thanks. 

The reason I used `heading` and not `title` is because title also contains counter and info link that are not truncated.

The reason I used `truncate` and not `wrap` is for default value for that property to be `false`.

I am happy to challenge both decisions - I'll bring that to the team review.

src/header/internal.tsx
pan-kot: That allows screen-reader users to always read the full contents no matter if the actual header is truncated or not.

I don't mind removing it for now and adding back in case we'll have the feedback.

src/header/styles.scss
pan-kot: Yes, I think it is better to do that.

Ideally is to have fewer conditions in the code but I agree it is better to decrease the blast radius. I am still waiting for the screenshot test results to see if there are noticeable diffs.

pan-kot:
> It might be worthwhile to expand the screenshot testing of this property to more diverse use cases than just rendering standalone headers.
> 
> qa

The feature is specifically introduced for dashboard and I agree it does make total sense to test it there as well. As for other possible use cases I'd hesitate. The problem with truncation is that it hides certain information and there must be a way to recover it. Otherwise the feature is not applicable.

src/header/internal.tsx
DaemonCahill: Pardon my ignorance but isn't the opposite true? As long as aria hidden is not set, the presence of hidden overflow will still announce the entire content to a screenreader?

DaemonCahill:
> The feature is specifically introduced for dashboard and I agree it does make total sense to test it there as well. As for other possible use cases I'd hesitate.

@pan-kot Once we release this it will be a public API on the header component that can be used anywhere, right? So, the intended use behind the API is somewhat irrelevant and what matters is where it _could_ be used?

pan-kot:
Closed due to a better mitigation found for board components.

PR #757, opened 2023-02-14
feat: Add 'rel' attribute to Button

rubencarvalho:
### Description

Added a `rel` attribute to the `Button` component. 

Without this change, when we add an `href` to the `Button` component with a `target="_blank"`, the `Button` component automatically adds the `rel` value of `"noreferrer noopener"`, and doesn't offer the option of overwriting this value. Some customers need to alter this default behaviour, hence this change.

Our `Link` component already has a way to override the `rel` attribute but the `Button` does not have this prop, so this change will also play into consistency across our components.

Related issue: AWSUI-20328

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/__tests__/__snapshots__/documenter.test.ts.snap
+++ b/src/__tests__/__snapshots__/documenter.test.ts.snap
@@ -2758,8 +2758,8 @@ It prevents users from clicking the button.",
       "type": "string",
     },
     Object {
-      "description": "Adds a \`rel\` attribute to the link. If the \`rel\` property is provided, it overrides the default behaviour.
-By default, the component sets the \`rel\` attribute to \\"noopener noreferrer\\" when \`target\` is \`\\"_blank\\"\`.",
+      "description": "Adds a \`rel\` attribute to the link. By default, the component sets the \`rel\` attribute to \\"noopener noreferrer\\" when \`target\` is \`\\"_blank\\"\`.
+If the \`rel\` property is provided, it overrides the default behavior.",
       "name": "rel",
       "optional": true,
       "type": "string",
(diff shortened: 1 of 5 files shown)


src/button/interfaces.ts
```diff
@@ -81,6 +81,12 @@ export interface ButtonProps extends BaseComponentProps {
    */
   target?: string;
 
+  /**
```
rubencarvalho, line 84: The description to the property is pending content review, will update once it gets finalised.

rubencarvalho: Reviewed and approved by content team.

src/__tests__/__snapshots__/documenter.test.ts.snap
```diff
@@ -2757,6 +2757,13 @@ It prevents users from clicking the button.",
       "optional": true,
       "type": "string",
     },
+    Object {
+      "description": "Adds a \`rel\` attribute to the link. By default, the component sets the \`rel\` attribute to \\"noopener noreferrer\\" when \`target\` is \`\\"_blank\\"\`.
```
johannes-weber, line 2761: I think we should set the `defaultValue` to "noopener noreferrer". We state it in the description but not in the API table's "Default" column.
(This should be done for the link component in a follow up PR as well).

johannes-weber (changes requested):

src/__tests__/__snapshots__/documenter.test.ts.snap
rubencarvalho: This `defaultValue` is only "default" if `target="_blank"`, otherwise no value will be set. That's why it's only on the description and not on the API "Default" column.

So, if I have a `<Button href="#">`, no "rel" attribute will be rendered. 

I think that's why we didn't do it for the Link as well. What do you think?

johannes-weber (approved):
LGTM!

johannes-weber (approved):
LGTM!

PR #759, opened 2023-02-14
feat: Improve collection preferences modal responsiveness when only using some features

timogasda:
### Description

This change makes sure that the collection preferences modal switches to a one-column layout (instead of two columns) whenever possible. For example, when there is only content for the left-hand side (page sizes, custom preferences, ...), or when there's only content for the right-hand side (visible column preferences).

Additionally, the size of the modal is reduces from "large" to "medium" if only one column is rendered.

These changes removes some unused space and makes sure that we use the space more efficiently.

Related links, issue #, if available: AWSDesignSystem-369

Internal document: mp8XA9mFCRQI

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/collection-preferences/__integ__/collection-preferences.test.ts
+++ b/src/collection-preferences/__integ__/collection-preferences.test.ts
@@ -48,7 +48,7 @@ describe('Collection preferences', () => {
   );
 
   test(
-    'renders 2 columns',
+    'renders 2 columns if all preferences are present',
     setupTest(async page => {
       const wrapper = createWrapper().findCollectionPreferences('.cp-1');
       await page.waitForVisible(wrapper.findTriggerButton().toSelector());
--- a/src/collection-preferences/__integ__/collection-preferences.test.ts
+++ b/src/collection-preferences/__integ__/collection-preferences.test.ts
@@ -4,34 +4,46 @@ import { BasePageObject } from '@cloudscape-design/browser-test-tools/page-objec
 import useBrowser from '@cloudscape-design/browser-test-tools/use-browser';
 import createWrapper from '../../../lib/components/test-utils/selectors';
 
-class PageObject extends BasePageObject {}
+const setupTest = (testFn: (page: BasePageObject) => Promise<void>) => {
+  return useBrowser(async browser => {
+    const page = new BasePageObject(browser);
+    await browser.url('#/light/collection-preferences/simple');
+    await page.setWindowSize({ width: 1200, height: 1200 });
+    await testFn(page);
+  });
+};
 
 describe('Collection preferences', () => {
   test(
-    'renders 1 column',
-    useBrowser(async browser => {
-      await browser.url('#/light/collection-preferences/simple');
-      const page = new PageObject(browser);
-      await page.setWindowSize({ width: 1200, height: 1200 });
-
+    'renders no columns if there is only custom content',
+    setupTest(async page => {
       const wrapper = createWrapper().findCollectionPreferences('.cp-3');
       await page.waitForVisible(wrapper.findTriggerButton().toSelector());
       await page.click(wrapper.findTriggerButton().toSelector());
       await expect(page.isExisting(wrapper.findModal().toSelector())).resolves.toBe(true);
 
+      // The content is small enough so that it doesn't need column layout
       const columnLayout = wrapper.findModal().findContent().findColumnLayout();
-      await expect(page.isExisting(columnLayout.findColumn(1).toSelector())).resolves.toBe(true);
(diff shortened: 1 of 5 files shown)


src/collection-preferences/__integ__/collection-preferences.test.ts
```diff
@@ -4,34 +4,46 @@ import { BasePageObject } from '@cloudscape-design/browser-test-tools/page-objec
...
       const wrapper = createWrapper().findCollectionPreferences('.cp-3');
       await page.waitForVisible(wrapper.findTriggerButton().toSelector());
       await page.click(wrapper.findTriggerButton().toSelector());
       await expect(page.isExisting(wrapper.findModal().toSelector())).resolves.toBe(true);
 
+      // The content is small enough so that it doesn't need column layout
       const columnLayout = wrapper.findModal().findContent().findColumnLayout();
-      await expect(page.isExisting(columnLayout.findColumn(1).toSelector())).resolves.toBe(true);
-      await expect(page.isExisting(columnLayout.findColumn(2).toSelector())).resolves.toBe(false);
+      await expect(page.isExisting(columnLayout.toSelector())).resolves.toBe(false);
```
michaeldowseza, line 27: Is there other content we should be expecting here? The test asserts there is no column layout but it could also pass if there is no content at all that is rendered

timogasda: This is really just checking for the responsiveness behavior. Everything else is already covered in the unit tests. So we could check for other content to be there but it won't bring any additional value.

src/collection-preferences/index.tsx
```diff
@@ -65,39 +65,43 @@ const ModalContent = ({
     );
   }
 
+  const hasLeftContent = !!(pageSizePreference || wrapLinesPreference || stripedRowsPreference || customPreference);
```
michaeldowseza, line 68: This is also defined below in CollectionPreferences. Can it not be accepted as a prop to avoid duplicated logic?

src/collection-preferences/utils.tsx
```diff
@@ -46,17 +47,17 @@ export const ModalContentLayout = ({ left, right }: ModalContentLayoutProps) =>
   if (smallContainer) {
     return (
       <div ref={ref}>
-        <div>{left}</div>
-        {right && <div className={styles['second-column-small']}>{right}</div>}
+        {left && <div>{left}</div>}
+        {right && <div className={clsx(left && styles['second-column-small'])}>{right}</div>}
```
michaeldowseza, line 51: Minor: I don't think we gain anything from using clsx here if we're not combining class names. 

So maybe readability vs. unnecessary use of a library to avoid a ternary operator?
```
className={left ? styles['second-column-small'] : null}
```

I don't think I have a strong preference since its not a new dependency and we use it elsewhere.

michaeldowseza (commented):
In general looks good, left mostly minor comments.

src/collection-preferences/__integ__/collection-preferences.test.ts
michaeldowseza: But could the responsiveness result in all the content being hidden?

src/collection-preferences/utils.tsx
timogasda: Yeah I chose those for better readability. A ternary operator just seemed a little unusual here. And since we generally use `clsx` across the code base, I thought this was nicer.

src/collection-preferences/index.tsx
timogasda: We can.. I just felt like it's too trivial to add a prop for, since it's something you can calculate based on the existing props.

src/collection-preferences/__integ__/collection-preferences.test.ts
timogasda: fair point :)

michaeldowseza (approved):

src/collection-preferences/__integ__/collection-preferences.test.ts
```diff
@@ -4,34 +4,52 @@ import { BasePageObject } from '@cloudscape-design/browser-test-tools/page-objec
...
+
+      await expect(page.isExisting(wrapper.findModal().findVisibleContentPreference().toSelector())).resolves.toBe(
+        true
+      );
+    })
+  );
+
+  test(
+    'renders 2 columns',
```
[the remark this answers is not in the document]
timogasda: sure :)

jperals (approved):

PR #760, opened 2023-02-14
fix: Make sure Notification bar is not covered by page elements below

jperals:
### Description

In Classic, the Notifications region gets a z-index of 825 if it is sticky, to let the page content go below it when scrolling. When it's not sticky, this was not necessary because it was being displayed above the content without overlaying it. But since the release of the [Stacked notifications](https://cloudscape.design/components/flashbar/?tabId=playground&example=stacked-notifications) feature, a new element (Notifications bar) partially overflows the Flashbar, spreading into the content area in some amount. If the page content below has very little or no top spacing with the Flashbar (such as when using `disableContentPadding`), this can lead to the main content being rendered on top of the Notifications bar.

This change fixes this by applying the same z-index of 825 to the Notifications region in all cases, also when it is not sticky.

Before:
Screenshot 2023-02-15 at 08 13 51

After:
Screenshot 2023-02-15 at 08 12 57

_Note:_ There is no need for such fix in Visual Refresh, because in the new App Layout the z-index is already always the same (also 825), see https://github.com/cloudscape-design/components/blob/main/src/app-layout/visual-refresh/notifications.scss#L13.

### How has this been tested?

- Manually tested
- Added new page for visual regression tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---


jperals:
Oops, removed. Thanks for spotting it!

just-boris (approved):

PR #764, opened 2023-02-15
fix: Ensure correct selection and resize controls are focused when sticky header is enabled

avinashbot:
### Description

Basically, this implementation manually tracks the three kinds of focusable elements in the header: selection checkbox, sort button, resize control — and passes it to the sticky header, which draws fake focus rings around them. This means keyboard users always interact with the "real" controls, but the focus visuals are always visible. Mouse interaction still means that focus is placed inside an aria-hidden region, but... baby steps. This is the clearest approach for now, since we're sorta stuck with using the fake floating header (can't use `position: sticky` because of an inner scroll container).

After shipping the immediate fix, it might be nice to pass the focus to the "real" control on click so that you can switch between keyboard/mouse, but that might introduce weird unintended consequences. So I'll play it safe and leave that out of this one.

It's a shame that we have to have a fake header, but it's the best way to maintain the current design and keep scroll lag-free: I did work on a quick test that always used a real header (see [table-sticky-refactor](https://github.com/cloudscape-design/components/tree/table-sticky-refactor)), but it means sticky needs to be implemented with JS scroll handlers.

Related links, issue #, if available: AWSUI-20270

### How has this been tested?

Fixing some existing tests where we check that the tab order is as expected. For the actual simulated focus outlines, visual tests are probably the least brittle way of checking this, especially since the change spans multiple components.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/table/header-cell/index.tsx
+++ b/src/table/header-cell/index.tsx
@@ -10,7 +10,7 @@ import { getAriaSort, getSortingIconName, getSortingStatus, isSorted } from './u
 import styles from './styles.css.js';
 import { Resizer } from '../resizer';
 import { useUniqueId } from '../../internal/hooks/use-unique-id';
-import { InteractiveElement } from '../thead';
+import { InteractiveComponent } from '../thead';
 
 interface TableHeaderCellProps<ItemType> {
   className?: string;
@@ -31,8 +31,8 @@ interface TableHeaderCellProps<ItemType> {
   resizableColumns?: boolean;
   isEditable?: boolean;
 
-  focusedElement?: InteractiveElement | null;
-  onFocusedElementChange?: (element: InteractiveElement | null) => void;
+  focusedComponent?: InteractiveComponent | null;
+  onFocusedComponentChange?: (element: InteractiveComponent | null) => void;
 }
 
 export function TableHeaderCell<ItemType>({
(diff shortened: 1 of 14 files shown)


src/table/header-cell/index.tsx
```diff
@@ -132,9 +137,17 @@ export function TableHeaderCell<ItemType>({
       {resizableColumns && (
         <>
           <Resizer
+            tabIndex={tabIndex}
+            showFocusRing={
+              focusedComponent?.type === 'resizer' &&
+              focusedComponent.col === colIndex &&
+              focusVisible['data-awsui-focus-visible']
```
timogasda, line 144: I wish we had a nicer way to do this. But probably not?

timogasda (commented):
Looks good overall, but some tests are failing.

src/table/header-cell/index.tsx
avinashbot: If only... unless you mean the way it's implemented? I could implement this by having three state variables (i.e. `focusedColumn`, `focusedResizer`, `isSelectionFocused`), but since they're all mutually exclusive and I don't want to split those apart and give this code a bigger footprint than it needs, I went with a single state variable.

timogasda (approved):

PR #769, opened 2023-02-16
fix: Remove update outside of act() for components using useReducedMotion in JSDOM

avinashbot:
### Description

`window.matchMedia` isn't defined in JSDOM. We do have a check for this in TypeScript, but the way we constructed that check means that the function could return undefined. And setState doesn't "dedupe" `undefined` and `false`, so MutationObserver ends up updating the state after the React render.

This feels like debugging incorrect type issues in JavaScript again, but JSDOM quirks are hopefully few and far between...

Related links, issue #, if available: AWSUI-20146 (partial fix)

### How has this been tested?

Manually, in a demo app. But the semantics shouldn't have changed :)

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/motion.ts
+++ b/src/internal/motion.ts
@@ -4,4 +4,4 @@ import { findUpUntil } from './utils/dom';
 
 export const isMotionDisabled = (element: HTMLElement): boolean =>
   !!findUpUntil(element, node => node.classList.contains('awsui-motion-disabled')) ||
-  (window.matchMedia && window.matchMedia('(prefers-reduced-motion: reduce)')?.matches);
+  (window.matchMedia?.('(prefers-reduced-motion: reduce)').matches ?? false);


src/internal/motion.ts
```diff
@@ -4,4 +4,4 @@ import { findUpUntil } from './utils/dom';
 
 export const isMotionDisabled = (element: HTMLElement): boolean =>
   !!findUpUntil(element, node => node.classList.contains('awsui-motion-disabled')) ||
-  (window.matchMedia && window.matchMedia('(prefers-reduced-motion: reduce)')?.matches);
+  (!!window.matchMedia && window.matchMedia('(prefers-reduced-motion: reduce)')?.matches);
```
just-boris, line 7: Should this be `||`?

just-boris:
```suggestion
!window.matchMedia || window.matchMedia('(prefers-reduced-motion: reduce)').matches;
```

avinashbot: But that would mean `isMotionDisabled` would return `true` for JSDOM where it used to return `false`. It's probably better, but it has the potential to break some unit tests.

just-boris: ok, another one

```suggestion
  window.matchMedia?.('(prefers-reduced-motion: reduce)').matches) ?? false;
```

Maybe also break down the function into two, for readability

```js
const isMotionDisabled = (element: HTMLElement): boolean => hasOptOutClass(element) || hasMediaQuery() ?? false;

avinashbot: Sure, the suggestion looks good - no need to overengineer the utility function :)

just-boris (approved):

PR #770, opened 2023-02-16
fix: Remove mutation observer from useVisualContext

avinashbot:
### Description

With React 18, `useMutationObserver` seems to be firing its callback _after_ the rendering is complete, which causes it to throw "an update was not wrapped in act(...)" errors in the top navigation component, since it renders a dropdown inside a custom visual context. So the visual context is initially `""`, but is updated after render to be `"top-navigation"`.

Unlike other "settings" (mode, motion, and density), visual contexts are used internally within the components in pretty close proximity to the children and aren't really updated outside of the React loop, so we could just get away with a simple `useLayoutEffect` instead of a mutation observer.

Related links, issue #, if available: AWSUI-20146 (partial fix)

### How has this been tested?

Existing tests should cover this. Also tested this in a demo application, which stops it from throwing "an update was not wrapped in act(...)" errors. 

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/components/visual-context/index.tsx
+++ b/src/internal/components/visual-context/index.tsx
@@ -1,24 +1,28 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 import clsx from 'clsx';
-import React, { useState } from 'react';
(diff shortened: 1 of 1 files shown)


johannes-weber (approved):
LGTM!

PR #782, opened 2023-02-22
fix: Debounce countText announcement and link it to text filter input

avinashbot:
### Description

1. Debounce countText using `LiveRegion` instead of using aria-live directly, so that announcements wait for a little bit after text input is done being announced. Went for 2 seconds after manual testing, but there's no rule here.
2. Attach countText to the input using `aria-describedby` as a "fallback", so it's always announced at the end when the user focuses into the text field.

Related links, issue #, if available: AWSUI-20166

### How has this been tested?

Unit tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/text-filter/internal.tsx
+++ b/src/text-filter/internal.tsx
@@ -9,6 +9,11 @@ import { fireNonCancelableEvent } from '../internal/events';
 import styles from './styles.css.js';
 import { InternalBaseComponentProps } from '../internal/hooks/use-base-component';
 import { TextFilterProps } from './interfaces';
+import LiveRegion from '../internal/components/live-region';
+import { useUniqueId } from '../internal/hooks/use-unique-id';
+
+// Debounce delay for live region (based on testing with VoiceOver)
+const LIVE_REGION_DELAY = 2000;
 
 type InternalTextFilterProps = TextFilterProps & InternalBaseComponentProps;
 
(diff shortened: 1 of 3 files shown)


src/text-filter/internal.tsx
```diff
@@ -44,16 +51,15 @@ const InternalTextFilter = React.forwardRef(
...
         />
-        <span
-          aria-live="polite"
-          aria-atomic="true"
-          className={clsx(styles.results, showResults && styles['results-visible'])}
-        >
-          {showResults ? countText : ''}
+        <span className={clsx(styles.results, showResults && styles['results-visible'])}>
+          <LiveRegion delay={LIVE_REGION_DELAY} visible={true}>
+            <span id={liveContentId}>{showResults ? countText : ''}</span>
```
jperals, line 61: In case there are no results, the `aria-describedby` is `undefined`, but the inner span is still rendering empty, as caught by this test: https://github.com/cloudscape-design/components/actions/runs/4242559394/jobs/7374220883

```
FAIL src/s3-resource-selector/s3-modal/__tests__/buckets-table.test.tsx
  ● filtering buckets

    expect(element).toBeEmptyDOMElement()

    Received:
      "<span><span id=\"text-filter79-1677068421820-9802\"></span></span><span class=\"awsui_root_xttbq_1i4dh_93 awsui_root_3bgfn_im8v7_5\"><span aria-atomic=\"true\" aria-live=\"polite\"></span></span>"
```

Is the inner span with the `id` necessary when there are no results? If so, `aria-describedby` should probably still be defined; if not, we can apply the conditional render to the entire span.

avinashbot: Let's just update the test? I don't like to conditionally render LiveRegion, especially because it has timing state built in.

jperals: Sounds good, but then if the span is always rendered, does it make sense to make the `aria-describedby` attribute conditional or should line 54 better be just like

```
ariaDescribedby={liveContentId}
```

avinashbot: I just wanted to play it safe, didn't want to set aria-describedby to an empty element 😅

jperals (approved):

PR #800, opened 2023-03-01
fix: Add role 'separator' to SideNavigation divider

Al-Dani:
### Description

Follow-up for #613 
Moving divider to the previous list element caused issues with testing, because `findItemByIndex(index)?.findDivider()` no longer worked as expected.
It also made dividers purely decorative, screen-reader users were not aware of existing grouping provided by dividers.
This change set a  role `separator` for the `listitem` that wraps the divider. 

Related links, issue #, if available: AWSUI-20463

### How has this been tested?

Unit test for the divider now includes more elements, which makes the divider not the first one in the list.

<details>
   <summary>Review checklist</summary>

_The following items are to be evaluated by the author(s) and the reviewer(s)._

#### Correctness

- _Changes include appropriate documentation updates._
- _Changes are backward-compatible if not indicated, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#public-apis)._
- _Changes do not include unsupported browser features, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#browsers-support)._
- _Changes were manually tested for accessibility, see [accessibility guidelines](https://cloudscape.design/foundation/core-principles/accessibility/)._

#### Security

- _If the code handles URLs: all URLs are validated through [the `checkSafeUrl` function](https://github.com/cloudscape-design/components/blob/main/src/internal/utils/check-safe-url.ts)._

#### Testing

- _Changes are covered with new/existing unit tests?_
- _Changes are covered with new/existing integration tests?_
</details>

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #824, opened 2023-03-13
feat: Split Side navigation by dividers into separate lists

Al-Dani:
### Description

Follow-up for https://github.com/cloudscape-design/components/pull/613, https://github.com/cloudscape-design/components/pull/800

Dividers are used to split the list of items in side navigation into different `<ul>`. This will improve the perception of the side navigation for screenreader users.

Related links, issue #, if available: AWSUI-20463, AWSUI-20493

### How has this been tested?

New unit test added to cover more complex structure of the side nav

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/side-navigation/internal.tsx
+++ b/src/side-navigation/internal.tsx
@@ -1,6 +1,6 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
-import React, { useCallback, useState, useEffect, useMemo } from 'react';
+import React, { useCallback, useState, useEffect, useMemo, ReactNode } from 'react';
 import clsx from 'clsx';
 import { ExpandableSectionProps } from '../expandable-section/interfaces';
 import InternalExpandableSection from '../expandable-section/internal';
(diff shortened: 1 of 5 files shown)


src/side-navigation/internal.tsx
```diff
@@ -64,63 +64,147 @@ export function Header({ definition, activeHref, fireFollow }: HeaderProps) {
...
+      {lists.map((list, index) => {
+        if (!list.items || list.items.length === 0) {
+          return (
+            <div key={`hr-${index}`} className={clsx(styles.list, styles[`list-variant-${variant}`])}>
+              {list.element}
+            </div>
+          );
+        } else {
+          return (
+            <ul key={`list-${index}`} className={clsx(styles.list, styles[`list-variant-${list.listVariant}`])}>
```
timogasda, line 186: I'm not sure if multiple lists after one another are the correct semantic structure here. Maybe nested lists instead? Alternatively, I'm wondering if `menu` would be even better for side navigation, see for example: https://www.w3.org/WAI/ARIA/apg/patterns/menubar/examples/menubar-editor/.

Al-Dani: Nested lists won't work, as it'll result in the same problem we tried to solve (`<ul>` that has non-`li` direct children).

Discussed the overall approach with A11y team, they approved it

Al-Dani:
Failed Visual regression test is a known flakiness in chart component

Issue #704, opened 2023-02-03
TopNavigation title gets truncated when logo svg is scaled in percentage


YueyingLu:
### Discussed in https://github.com/cloudscape-design/components/discussions/606

<div type='discussions-op-text'>

<sup>Originally posted by **suvihaara** January  3, 2023</sup>
Hi, 

I had an issue where when an SVG is scaled in % rather than a value, the TopNavigation title is truncated. I fixed it by changing the scaling of the SVG to be set in values rather than %. Is there a way that would allows the TopNavigation Logo SVGs to be scaled in % without truncating the title?</div>

taheramr:
Hello, thanks for reaching out.

This looks like a valid bug and I was able to reproduce it, we are currently looking at different solutions to this and we'll keep you updated.

taheramr:
.

taheramr:
Sorry for the late replies on this bug, this [PR](https://github.com/cloudscape-design/components/pull/826) should fix the issue with the top navigation and full width/height svg logos.

Will followup when this is merged.

taheramr:
Hello again,

We are updating our components docs to guide developers to use fixed width and height for SVGs so they get rendered correctly without making the title text truncated in this [PR](https://github.com/cloudscape-design/components/pull/1069).

So the path forward for this would be editing your SVGs width and height or letting the [`viewBox`](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/viewBox) decide on the dimensions of the SVG.

I'm closing this as no further action is needed!
Amr

PR #826, opened 2023-03-13
fix: Fix top navigation text overflowing when provided with 100% w-h svgs

taheramr:
### Description
https://github.com/cloudscape-design/components/issues/704
If there is a provided svg to the top navigation component logo that has `100%` for height and width in the svg definition the top navigation title will always overflow and be cut.

To reproduce this, edit one of the sample svg logos in the test pages and make it have 100% height and width and check the page.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/top-navigation/styles.scss
+++ b/src/top-navigation/styles.scss
@@ -84,7 +84,8 @@
   display: block;
   max-height: awsui.$space-xxl;
   margin-right: awsui.$space-s;
-  width: auto;
+  width: 100%;
+  max-width: fit-content;
   flex-shrink: 0;
 
   // Setting an arbitrary min-width here discourages browser from lazy rendering
@@ -180,6 +181,7 @@
   // Hide dividers around primary buttons.
   left: -1px;
   border-left: 1px solid awsui.$color-background-container-content;
+
   &::after {
     display: none;
   }
--- a/pages/top-navigation/logos/full-w-h-logo.svg
+++ b/pages/top-navigation/logos/full-w-h-logo.svg
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg width="100%" height="100%" viewBox="0 0 43 31" version="1.1" xmlns="http://www.w3.org/2000/svg">
+    <g stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
+        <g transform="translate(-31.000000, -12.000000)">
+            <g transform="translate(31.000000, 12.000000)">
+                <rect id="Rectangle" fill="#232f3e" stroke="#d5dbdb" x="0.5" y="0.5" width="42" height="30" rx="2"></rect>
+                <text font-family="AmazonEmber-Regular, Amazon Ember" font-size="12" fill="#FFFFFF">
+                    <tspan x="9" y="19">Logo</tspan>
+                </text>
+            </g>
+        </g>
+    </g>
+</svg>
--- a/pages/top-navigation/simple.page.tsx
+++ b/pages/top-navigation/simple.page.tsx
@@ -7,6 +7,7 @@ import logo from './logos/simple-logo.svg';
 import longLogo from './logos/long-logo.svg';
 import narrowLogo from './logos/narrow-logo.svg';
 import tallLogo from './logos/tall-logo.svg';
+import fullHeightWidthLogo from './logos/full-w-h-logo.svg';
 import { I18N_STRINGS } from './common';
 
 export default function TopNavigationPage() {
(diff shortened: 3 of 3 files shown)


src/top-navigation/styles.scss
```diff
@@ -60,9 +60,11 @@
   min-width: 0;
 
   > .identity-link {
-    display: flex;
+    display: grid;
+    grid-template-columns: max-content auto;
     align-items: center;
     text-decoration: none;
+    gap: awsui.$space-s;
```
just-boris, line 67: Is there a way to do the same without changing `display`?

taheramr: I tried fixing this with the current `display: flex` but there is too many magic to maintain like for example the `min-width: 1px` we currently have and the fact that if a customer provided an `svg` with 100% height and width the flex container will not know how much width/height should it give to the svg and the text always overflows.

just-boris: I got it working by applying `flex-grow: 1` to the `.identity` selector.

I think your solution with `display: grid` does something similar, but in more magical way, because this is how nested `display: grid` and `display: flex` interact.

IMO, having two blocks (logo and the text) in this placement is not a valid use-case for grid display

taheramr: that does not solve it tho, if I applied the same logic with an svg image of 100% height and width it'll still show the same issue.

src/top-navigation/styles.scss
```diff
@@ -58,6 +58,7 @@
 
 .identity {
   min-width: 0;
+  flex-grow: 1;
```
taheramr, line 61: @just-boris so you are right about this, but the problem is that we have magical calculations happens in the `useTopNavigation` that figures out the identity width, given the fact that flex-grow would make its width bigger than the top navigation container itself so I had to re-calculate the identity width below using its inner content ( logo width + title width ) in the virtual top navigation.

Kudos for the suggestion tho.

just-boris: Thanks for the extra dive. This approach is better!

just-boris (dismissed):

jperals (dismissed):

src/top-navigation/use-top-navigation.ts
```diff
@@ -78,37 +78,45 @@ export function useTopNavigation({ identity, search, utilities }: UseTopNavigati
...
 
     // Get available width from the visible top navigation.
+    /* istanbul ignore else  */
     const availableWidth = getContentBoxWidth(mainRef.current.querySelector(`.${styles['padding-box']}`)!);
     if (availableWidth === 0) {
       // Likely in an SSR or Jest situation.
       setResponsiveState(responsiveStates[0]);
-      return;
+    } else {
+      /* istanbul ignore else  */
```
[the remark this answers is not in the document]
taheramr: I tried the setup you mentioned and for some reason it did not work, will check if this passed and let you know

jperals (approved):

src/top-navigation/use-top-navigation.ts
```diff
@@ -79,36 +79,42 @@ export function useTopNavigation({ identity, search, utilities }: UseTopNavigati
...
+    /* istanbul ignore else  */
     if (availableWidth === 0) {
       // Likely in an SSR or Jest situation.
       setResponsiveState(responsiveStates[0]);
-      return;
+    } else {
+      const titleWidth = virtualRef.current.querySelector(`.${styles.title}`)?.getBoundingClientRect().width ?? 0;
+      const logoWidth = virtualRef.current.querySelector(`.${styles.logo}`)?.getBoundingClientRect().width ?? 0;
+
+      // the identity div has a flex-grow: 1 css, this makes calculating its content width requiring to calcualte both
```
jperals, line 90: This comment is outdated now because the identity div does not have `flex-grow: 1` anymore, right?

src/top-navigation/use-top-navigation.ts
```diff
@@ -79,36 +79,42 @@ export function useTopNavigation({ identity, search, utilities }: UseTopNavigati
...
+      // the logo and title, if we instead calculate the width of the container it'll be larger than its contnet becuase of the flex-grow.
```
jperals, line 91: Typo: "contnet"

src/top-navigation/styles.scss
```diff
@@ -84,7 +84,8 @@
   display: block;
   max-height: awsui.$space-xxl;
   margin-right: awsui.$space-s;
-  width: auto;
+  width: 100%;
```
jperals, line 87: Why is this preferable to the previous approach of using `flex-grow: 1` for `.identity`?

taheramr: I found out the flex grow will be problematic cause it makes the identity ( which is an anchor ) takes the full width, meaning the user can click on an empty section in the top navigation and still be able to trigger the anchor because it has full width.
With the above facts I removed the logic in the use top navigation, no need for it with this fix.

src/top-navigation/use-top-navigation.ts
taheramr: removed

src/top-navigation/styles.scss
jperals: Unfortunately it doesn't look good on Firefox... The logo shows up very small.

src/top-navigation/use-top-navigation.ts
taheramr: removed

jperals:
Is this PR still relevant after https://github.com/cloudscape-design/components/pull/1069 or should it be closed?

PR #829, opened 2023-03-13
feat: File upload component

pan-kot:
### Description

Introduced new component - File upload.

### How has this been tested?

Unit- and integration tests, form-field integration tests, permutations test page

<details>

#### Correctness

#### Security

#### Testing

</details>

---


pages/file-upload/permutations.page.tsx
```diff
@@ -0,0 +1,55 @@
...
+        new File([new Blob(['demo content 2'], { type: 'text/plain' })], 'demo file 2'),
+      ],
+    ],
+  },
+  {
+    value: [[new File([new Blob(['demo content'], { type: 'text/plain' })], 'demo file')]],
+    showFileType: [true],
+    showFileSize: [true],
+    showFileLastModified: [true],
+    showFileThumbnail: [true],
```
connorlanigan, line 31: Can we include the `false` variants here as well? That will allow us to see all different combinations of these values on the page, and to catch any CSS issues in the screenshots.

pages/file-upload/with-app-layout.page.tsx
```diff
@@ -0,0 +1,243 @@
...
+  header: 'File upload',
+  content: <Box>File upload test scenario page</Box>,
+};
+const profilePictureToolsContent = {
+  header: 'Profile picture',
+  content: (
+    <SpaceBetween size="s">
+      <Box>The profile picture must satisfy the below requirements:</Box>
+      <ul>
+        <li>Maximum size: 1 MB; Density: at least 300 x 300 pixels</li>
```
connorlanigan, line 24: minor: "Density" would more accurately be "Resolution"

src/file-upload/default-formatters.ts
```diff
@@ -0,0 +1,14 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { formatDateTime } from '../internal/utils/date-time';
+
+export function formatFileSize(size: number): string {
+  const KB = 1000;
+  const MB = 1000 ** 2;
+  return size < MB ? `${(size / KB).toFixed(2)} KB` : `${(size / MB).toFixed(2)} MB`;
```
connorlanigan, line 9: Can we include GB and TB as well? That way we can cover the common file sizes

src/file-upload/file-option.tsx
```diff
@@ -0,0 +1,171 @@
...
+}
+
+export function FileOption({ file, metadata, nameEditing, i18nStrings }: FileOptionProps) {
+  const isImage = !!file.type && file.type.split('/')[0] === 'image';
+  const [isNameEditFocused, setNameEditFocused] = useState(false);
+  const isFocusVisible = !!useFocusVisible()['data-awsui-focus-visible'];
+  const fileOptionNameRef = useRef<HTMLDivElement>(null);
+  const fileNameEditActivateButtonRef = useRef<ButtonProps.Ref>(null);
+
+  // Focus on name editing activate button when existing name editing mode.
```
connorlanigan, line 45: typo: "existing" -> "exiting"

src/file-upload/file-option-thumbnail.tsx
```diff
@@ -0,0 +1,29 @@
...
+}
+
+export function FileOptionThumbnail({ file }: FileOptionThumbnailProps) {
+  const thumbnailRef = useRef<HTMLImageElement>(null);
+
+  useEffect(() => {
+    const reader = new FileReader();
+    reader.onloadend = () => {
+      if (thumbnailRef.current && thumbnailRef.current.src) {
+        thumbnailRef.current.src = reader.result as string;
```
connorlanigan, line 18: I would suggest limiting this by size, e.g. 50MB. Users might upload e.g. satellite images, which can be multi-gigabyte TIFF files. Loading these into a DOM attribute will likely slow down the page or device.

----

Can we also use React state for the `src` value instead of imperatively updating the DOM? Otherwise that state might get lost if/when React rerenders.

----

Potentially it might also be useful to use `URL.createObjectURL` instead of a FileReader, which will return a blob URL as a kind of "handle" to that file content. The advantage is that the (potentially large) file content will not be loaded into Javascript. The disadvantage is that this URL is not automatically garbage collected, so we'll have to call `revokeObjectURL` when the component unmounts or the file changes, to prevent a memory leak.

src/file-upload/styles.scss
```diff
@@ -0,0 +1,84 @@
...
+  width: 100%;
+  overflow: hidden;
+}
+
+.file-option-name-edit-activate {
+  flex-shrink: 0;
+}
+
+.file-option-name-edit-submit {
+  margin-right: 1px;
```
connorlanigan, line 48: Can we use a token like `awsui.$space-scaled-xs`?

connorlanigan (dismissed):

src/file-upload/default-formatters.ts
connorlanigan: I think GB is not unlikely at all. For example, if you want to [upload an existing virtual machine into EC2](https://docs.aws.amazon.com/vm-import/latest/userguide/vmimport-image-import.html#import-vm-image) or large data sets for machine learning, this will be multiple GB in many cases.
While TB sizes are currently unlikely because of the long upload time on most consumer connections, I think it is good to be on the safe side here by supporting "one step larger" than we expect, especially since it requires very little code. We will then not have to think about this function again for many years.

just-boris: Is it possible to have a fake file object? Can we use this component not for uploads from the local system directly, but from some remote endpoint to another?

In any way, adding extra units does not cost us much, so why not to add it regardless

pages/file-upload/permutations.page.tsx
pan-kot: I am not sure if this is really needed those, adding true, false everywhere would produce lots of permutations. However, I can add [false, true] for the thumbnail because it renders differently compared to other atrributes.

src/file-upload/file-option.tsx
```diff
@@ -0,0 +1,171 @@
...
+                      variant="inline-icon"
+                      className={styles['file-option-name-edit-submit']}
+                      onClick={nameEditing.onSubmit}
+                      ariaLabel={i18nStrings.submitFileNameEditAriaLabel}
+                    />
+                  </InternalSpaceBetween>
+                </>
+              ) : (
+                <>
+                  <div className={styles['file-option-name-label']} title={file.name}>
```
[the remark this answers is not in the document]
pan-kot: We decided to de-scope the file name editing for now. The title is going to be removed. Instead, the text is allowed to wrap.

src/file-upload/file-option.tsx
```diff
@@ -0,0 +1,171 @@
...
+                    {file.name}
+                  </div>
+
+                  <InternalButton
+                    ref={fileNameEditActivateButtonRef}
+                    __hideFocusOutline={true}
```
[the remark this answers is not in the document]
pan-kot: It was communicated with a separate outline, however, that is no longer a concern because file name editing is de-scoped.

src/file-upload/internal.tsx
```diff
@@ -0,0 +1,178 @@
...
+  const onFileRemove = (removeFileIndex: number) => {
+    fireNonCancelableEvent(onChange, { value: value.filter((_, fileIndex) => fileIndex !== removeFileIndex) });
+  };
+
+  const onFileNameUpdate = (fileIndex: number, fileName: string) => {
+    if (value[fileIndex].name === fileName) {
+      return;
+    }
+    const files = [...value];
+    const updated = new File([files[fileIndex]], fileName);
```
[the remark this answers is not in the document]
pan-kot: Thanks, I think synchronising the attributes is probably the right call. We will come back to this questions if re-introducing the name editing.

src/file-upload/styles.scss
pan-kot: No longer applicable.

src/file-upload/dropzone.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+export function useDropzoneVisible() {
+  const [isDropzoneVisible, setDropzoneVisible] = useState(false);
+
+  // Registering global drag events listeners.
+  useEffect(() => {
+    // The timer helps avoiding dropzone blinking.
+    let dragTimer: null | ReturnType<typeof setTimeout> = null;
+
+    // The file-upload dropzone is shown when the user drags files over the to the browser.
```
connorlanigan, line 23: typo: "the to the" -> "to the"

src/file-upload/dropzone.tsx
```diff
@@ -0,0 +1,113 @@
...
+    const onDocumentDragOver = (event: DragEvent) => {
+      event.preventDefault();
```
connorlanigan, line 25: Why is this line needed, what is the default behaviour? If multiple dropzones are on a page, does this prevent the others from working? Should we move this line inside the conditional block below?

src/file-upload/dropzone.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+      if (event.dataTransfer && event.dataTransfer.types && event.dataTransfer.types.indexOf('Files') !== -1) {
```
connorlanigan, line 27: I think this can be simplified to `if(event.dataTransfer?.types?.includes('Files')) {`

src/file-upload/dropzone.tsx
```diff
@@ -0,0 +1,113 @@
...
+  onChange: (files: File[]) => void;
+  children: React.ReactNode;
+}
+
+export function useDropzoneVisible() {
+  const [isDropzoneVisible, setDropzoneVisible] = useState(false);
+
+  // Registering global drag events listeners.
+  useEffect(() => {
+    // The timer helps avoiding dropzone blinking.
```
connorlanigan, line 20: Do we know why the blinking happens?

src/file-upload/dropzone.tsx
```diff
@@ -0,0 +1,113 @@
...
+  return isDropzoneVisible;
+}
+
+export function Dropzone({ onChange, children }: DropzoneProps) {
+  const dropzoneRef = useRef<HTMLDivElement>(null);
+  const [isDropzoneHovered, setDropzoneHovered] = useState(false);
+
+  const onDragOver = (event: React.DragEvent) => {
+    event.preventDefault();
+    event.stopPropagation();
```
connorlanigan, line 68: Why do we need to stop propagation?

src/file-upload/dropzone.tsx
```diff
@@ -0,0 +1,113 @@
...
+      event.dataTransfer.dropEffect = 'none';
+    }
+  };
+
+  const onDrop = (event: React.DragEvent) => {
+    event.preventDefault();
+
+    const newFiles: File[] = [];
+    const nullableFiles = event.dataTransfer.items
+      ? Array.from(event.dataTransfer.items).map(item => (item.kind === 'file' && item.getAsFile()) || null)
```
connorlanigan, line 91: the filter condition could be a bit more idiomatic like this: 
`item => (item.kind === 'file' ? item.getAsFile()) : null`

src/file-upload/dropzone.tsx
```diff
@@ -0,0 +1,113 @@
...
+      : Array.from(event.dataTransfer.files);
+    nullableFiles.forEach(fileOrNull => fileOrNull && newFiles.push(fileOrNull));
```
connorlanigan, line 93: this filtering can be implemented directly on the array with `.filter(Boolean)`

src/file-upload/dropzone.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+    onChange(newFiles);
+    setDropzoneHovered(false);
```
connorlanigan, line 96: minor: I would call `setDropzoneHovered` before firing the event. Otherwise, if an error happens in the event handler, the state would not get updated

src/file-upload/dropzone.tsx
```diff
@@ -0,0 +1,113 @@
...
+  };
+
+  return (
+    <div
+      ref={dropzoneRef}
```
connorlanigan, line 101: It looks like this ref is unused

src/file-upload/dropzone.tsx
pan-kot: This prevents browser's standard treatment, see example: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dragover_event

This line does not prevent the event from being received by the dropzones.

That said, there might be a case where the global listener that we register does cause trouble for the pages that do expect some default behaviour of the dragged files. Once reported we can think of moving the line somewhere or adding a flag. For example, react-dropzone has a flag to allow document drop on the page: https://github.com/react-dropzone/react-dropzone/blob/master/src/index.js#L64.

src/file-upload/dropzone.tsx
pan-kot: Will change to `event.dataTransfer?.types?.indexOf('Files') !== -1`. The `includes` is not possible due to our current transpilation settings.

src/file-upload/interfaces.ts
```diff
@@ -0,0 +1,110 @@
...
+   */
+  constraintText?: React.ReactNode;
+  /**
+   * Text that displays as a validation message.
+   */
+  errorText?: React.ReactNode;
+  /**
+   * An array of file errors corresponding to the files in the `value`.
+   */
+  fileErrors?: null | ReadonlyArray<null | string>;
```
just-boris, line 62: When would someone want to use `null` instead of `undefined` here?

src/file-upload/internal.tsx
```diff
@@ -0,0 +1,181 @@
...
+  const handleFilesChange = (newFiles: File[]) => {
+    const newValue = multiple ? [...value, ...newFiles] : newFiles[0] ? newFiles : [...value];
+    fireNonCancelableEvent(onChange, { value: newValue });
+  };
+
+  const onFileRemove = (removeFileIndex: number) => {
+    const newValue = value.filter((_, fileIndex) => fileIndex !== removeFileIndex);
+    fireNonCancelableEvent(onChange, { value: newValue });
+    setRemovedFileIndex(removeFileIndex);
+    setRemovedLastFile(value.length === 1);
```
just-boris, line 83: This seems like "you might not need an effect" case: https://react.dev/learn/you-might-not-need-an-effect#sharing-logic-between-event-handlers

Is there anything against focusing it right in the handler?

src/file-upload/dropzone.tsx
pan-kot: I added it when experimenting with dataTransfer but it is not really needed. I'll remove it.

src/file-upload/internal.tsx
just-boris: P.S. I have questions (simplification options) for `removedFileIndex` too, but since it requires changing current `TokenList` implementation, let's skip this

src/file-upload/dropzone.tsx
```diff
@@ -0,0 +1,113 @@
...
+  children: React.ReactNode;
+}
+
+export function useDropzoneVisible() {
+  const [isDropzoneVisible, setDropzoneVisible] = useState(false);
+
+  // Registering global drag events listeners.
+  useEffect(() => {
+    // The timer helps avoiding dropzone blinking.
+    let dragTimer: null | ReturnType<typeof setTimeout> = null;
```
just-boris, line 20: What kind of blinking is happening? Could you give more details?

src/file-upload/internal.tsx
```diff
@@ -0,0 +1,181 @@
...
+          dismissLabel={i18nStrings.removeFileAriaLabel(value[0], 0)}
+          onDismiss={() => onFileRemove(0)}
+          errorText={fileErrors?.[0]}
+          errorIconAriaLabel={i18nStrings.errorIconAriaLabel}
+        >
+          <FileOption file={value[0]} metadata={metadata} i18nStrings={i18nStrings} />
+        </Token>
+      ) : null}
+
+      {(multiple && value.length > 0) || value.length > 1 ? (
```
just-boris, line 156: When is the `value.length > 1` condition needed?

src/token-group/token.tsx
```diff
@@ -6,25 +6,48 @@ import clsx from 'clsx';
...
-      {onDismiss && <DismissButton disabled={disabled} dismissLabel={dismissLabel} onDismiss={onDismiss} />}
+      <div className={clsx(styles['token-box'], disabled && styles['token-box-disabled'])}>
+        {children}
+        {onDismiss && <DismissButton disabled={disabled} dismissLabel={dismissLabel} onDismiss={onDismiss} />}
+      </div>
+      {errorText && (
+        <FormFieldError id={errorId} errorIconAriaLabel={errorIconAriaLabel}>
+          {errorText}
+        </FormFieldError>
+      )}
```
just-boris, line 50: How about making it a slot, to the consumer of the token will render `FormFieldError` when needed.

pages/file-upload/page-helpers.tsx
```diff
@@ -0,0 +1,134 @@
...
+        <Flashbar
+          items={[{ type: 'success', header: 'Files uploaded', statusIconAriaLabel: 'uploaded', ariaRole: 'status' }]}
+        />
+      );
+    case 'submitted':
+      return (
+        <Flashbar
+          items={[
+            { type: 'success', header: 'Files uploaded', statusIconAriaLabel: 'uploaded', ariaRole: 'status' },
+            { type: 'success', header: 'Form submitted', statusIconAriaLabel: 'submitted', ariaRole: 'status' },
```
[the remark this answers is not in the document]
pan-kot: This was done for visibility to acknowledge both events succeeded. I can only keep the submitted one though.

pages/file-upload/permutations.page.tsx
```diff
@@ -0,0 +1,54 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import React from 'react';
+import createPermutations from '../utils/permutations';
+import PermutationsView from '../utils/permutations-view';
+import ScreenshotArea from '../utils/screenshot-area';
+import FileUpload, { FileUploadProps } from '~components/file-upload';
+import { i18nStrings } from './shared';
+
+const permutations = createPermutations<Omit<FileUploadProps, 'dismissAriaLabel' | 'i18nStrings'>>([
```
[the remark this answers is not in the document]
pan-kot: Good catch. I will add errorText, constraintText and fileErrors as now these properties are available.

pages/file-upload/permutations.page.tsx
```diff
@@ -0,0 +1,54 @@
...
+        new File([new Blob(['demo content 1'], { type: 'text/plain' })], 'demo file 1'),
+        new File([new Blob(['demo content 2'], { type: 'text/plain' })], 'demo file 2'),
+      ],
+    ],
+  },
+  {
+    value: [[new File([new Blob(['demo content'], { type: 'text/plain' })], 'demo file')]],
+    showFileSize: [true],
+    showFileLastModified: [true],
+    showFileThumbnail: [true],
```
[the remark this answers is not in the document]
pan-kot: I don't think we really need all combinations but I will add some variability for the thumbnail.

pages/app-layout/utils/content-blocks.tsx
```diff
@@ -45,8 +45,8 @@ export function Containers() {
   );
 }
 
-export function Tools({ children }: { children: React.ReactNode }) {
-  return <HelpPanel header={<h2>Overview</h2>}>{children}</HelpPanel>;
+export function Tools({ header = 'Overview', children }: { header?: string; children: React.ReactNode }) {
```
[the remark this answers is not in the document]
pan-kot: Thanks, I will revert.

src/file-upload/__tests__/file-upload-dropzone.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+}
+
+function TestDropzoneVisible() {
+  const isDropzoneVisible = useDropzoneVisible();
+  return <div>{isDropzoneVisible ? 'visible' : 'hidden'}</div>;
+}
+
+describe('File upload dropzone', () => {
+  beforeEach(() => jest.useFakeTimers());
+  afterEach(() => jest.useRealTimers());
```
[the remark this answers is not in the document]
pan-kot: Thanks, I'll try this out

src/file-upload/dropzone.tsx
```diff
@@ -0,0 +1,113 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import React, { useEffect, useRef, useState } from 'react';
+import styles from './styles.css.js';
```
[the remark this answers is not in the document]
pan-kot: Alright, will do

src/file-upload/file-input.tsx
```diff
@@ -0,0 +1,103 @@
...
+  if (formFieldContext.invalid) {
+    nativeAttributes['aria-invalid'] = true;
+  }
+  if (ariaRequired) {
+    nativeAttributes['aria-required'] = true;
+  }
+
+  // Synchronizing component's value with the native file input state.
+  useEffect(() => {
+    if (window.DataTransfer) {
```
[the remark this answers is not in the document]
pan-kot: This is for jsdom

src/file-upload/file-input.tsx
```diff
@@ -0,0 +1,103 @@
...
+        formAction="none"
+        onClick={onUploadButtonClick}
+        className={styles['upload-button']}
+        __nativeAttributes={{ tabIndex: -1, 'aria-hidden': true }}
+        __forceFocusOutline={isFocused}
+      >
+        {children}
+      </InternalButton>
+
+      <ScreenreaderOnly id={uploadButtonLabelId}>{children}</ScreenreaderOnly>
```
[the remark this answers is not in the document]
pan-kot: Because the button is aria-hidden

src/file-upload/file-option-thumbnail.tsx
```diff
@@ -0,0 +1,34 @@
...
+      setImageSrc(src);
+
+      return () => {
+        URL.revokeObjectURL(src);
+      };
+    }
+  }, [file]);
+
+  const onImageLoad = () => {
+    URL.revokeObjectURL(imageSrc);
```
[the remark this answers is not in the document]
pan-kot: I used this example as a reference: https://developer.mozilla.org/en-US/docs/Web/API/File_API/Using_files_from_web_applications#example_using_object_urls_to_display_images

src/file-upload/dropzone.tsx
just-boris: Did you consider using `dragenter` event instead? 

https://html.spec.whatwg.org/multipage/dnd.html#event-dnd-dragenter

That one seem to be strictly attached to the document body

src/file-upload/file-input.tsx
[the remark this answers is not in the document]
just-boris: Let's add a comment about this

src/file-upload/file-input.tsx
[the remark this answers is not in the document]
just-boris: Why is it aria-hidden?

src/file-upload/file-option-thumbnail.tsx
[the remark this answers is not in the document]
just-boris: > Set up the image's load event handler to release the object URL since it's no longer needed once the image has been loaded. This is done by calling the [URL.revokeObjectURL()](https://developer.mozilla.org/en-US/docs/Web/API/URL/revokeObjectURL) method and passing in the object URL string as specified by img.src.

This breaks "open in a new tab" and "save image" features. Not sure why would anyone save their own local file one more time, but also unclear what is the harm of preserving the value.

I think they do it because they have no "unmount" lifecycle, so they want to prevent memory leaks. In our case it is not needed

src/file-upload/dropzone.tsx
pan-kot: I think the main problem is not with dragover but with dragleave. If there was a reliable event to be used instead that could have helped. The dragleave is issued whenever cursor leaves UI elements and it needs to be cancelled by the dragover.

src/file-upload/internal.tsx
pan-kot: That is because technically one can set `multiple=false` and still provide multiple files or even change `multiple` dynamically. In that case we issue a dev. warning but also there needs to be some representation. One option would be to show the first file only but I think a better fair one is to show all files.

src/file-upload/dropzone.tsx
just-boris: You can check the event target to figure which element was left. Try to test locally, maybe `event.target.tagName === 'BODY'` will work

src/file-upload/styles.scss
```diff
@@ -0,0 +1,62 @@
...
+  }
+}
+
+.file-input-container {
+  position: relative;
+}
+
+.upload-input {
+  position: absolute;
+  clip: rect(0, 0, 0, 0) !important;
```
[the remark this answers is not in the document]
pan-kot: The purpose of this line is to reliably hide the input visually. We also use it in our utility mixin:
```
@mixin awsui-util-hide {
  position: absolute !important;
  top: -9999px !important;
  left: -9999px !important;
}
```

For the input I don't want to apply the mixin to keep its actual position within the screen. I know at least one case when using absolute position breaks screen-readers navigation.

src/file-upload/__tests__/file-upload-dropzone.test.tsx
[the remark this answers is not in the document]
just-boris: Let's resolve the other thread first: https://github.com/cloudscape-design/components/pull/829#discussion_r1164323904

I believe timeout is not needed at all

src/token-group/token.tsx
pan-kot: Generally I don't mind it but it comes with a downside. The error association would have to be done on the outside:

```
function Demo({ tokens }) {
  return <>{tokens.map((token, index) => (
    <Token ... ariaDescription={`${errorId}-${index}`} error={<FormFieldError id={`${errorId}-${index}`}>{token.error}</FormFieldError>}>{token.content}</Token>
  ))}</>
}
```

src/token-group/token.tsx
```diff
@@ -6,25 +6,48 @@ import clsx from 'clsx';
...
+  className,
+}: TokenProps) {
+  const errorId = useUniqueId('error');
   return (
     <div
-      className={clsx(styles.token, disabled && styles['token-disabled'])}
+      className={clsx(styles.token, className)}
       role="group"
       aria-label={ariaLabel}
+      aria-describedby={errorText ? errorId : undefined}
```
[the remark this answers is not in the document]
pan-kot: The tokens were just recently made groups. They have aria-labels but don't have descriptions. The description is only needed to associate it with an error if it is defined.

style-dictionary/visual-refresh/metadata/colors.ts
```diff
@@ -500,6 +500,26 @@ const metadata: StyleDictionary.MetadataIndex = {
     themeable: false,
     public: true,
   },
+  colorDropzoneBackgroundActive: {
+    description: 'The color of file upload dropzone background in active state.',
+    themeable: false,
+    public: true,
```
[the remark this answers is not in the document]
pan-kot: Great catch 👍 I din't mean making the new tokens public - will update.

src/file-upload/internal.tsx
just-boris: In table single selection mode we simply ignore extra items: https://github.com/cloudscape-design/components/blob/main/src/table/use-selection.ts#L97

I suggest to not over-engineer here too, and not support invalid use-cases

src/token-group/token.tsx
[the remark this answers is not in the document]
just-boris: The text is rendered on the page and wrapped into the same role element. I do not understand why extra effort is needed. Do you have any resource where I can read about this requirement?

src/file-upload/dropzone.tsx
pan-kot: Good idea - I'll try it out 👍

src/file-upload/internal.tsx
pan-kot: Alright, I am not against that

src/file-upload/file-input.tsx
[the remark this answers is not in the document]
pan-kot: The actual interactive element is the `<input type="file" />`. The button is a decoration. It is never supposed to be focused or navigated to with assistive tech native navigation, that's why it has tabIndex=-1 and aria-hidden=true.

I will add a code comment as well.

src/file-upload/file-option-thumbnail.tsx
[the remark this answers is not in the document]
pan-kot: Agreed - will only keep the revoke in the effect cleanup.

src/file-upload/interfaces.ts
pan-kot: I would use `null` when setting this value explicitly. It can also be set to null by default if used with some form helper library.

src/file-upload/internal.tsx
pan-kot: Thanks - I'd love to improve that logic and remove the effects if possible.

The challenge so far is that there must be a render in between remove and focus update. Imagine 5 items are present and 2 are hidden behind "show more". When removing the 3rd item the focus must go to the item which is not yet present in the DOM at that point.

src/file-upload/dropzone.tsx
pan-kot: The dragleave event's target is not reliable. It can be html, main or some div when the element is dragged away from the page.

pages/file-upload/permutations.page.tsx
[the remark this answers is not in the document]
pan-kot: Ideally for that to load actual images but I couldn't figure out how to do that. I can load an image from sources using a webpack loader but not sure which format does that sting have and how to convert it to a file.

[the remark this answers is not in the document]
pan-kot: That should be solvable but I'd prefer to add that as a follow-up.

src/file-upload/dropzone.tsx
```diff
@@ -0,0 +1,113 @@
...
+    }
+  };
+
+  const onDrop = (event: React.DragEvent) => {
+    event.preventDefault();
+
+    const newFiles: File[] = [];
+    const nullableFiles = event.dataTransfer.items
+      ? Array.from(event.dataTransfer.items).map(item => (item.kind === 'file' && item.getAsFile()) || null)
+      : Array.from(event.dataTransfer.files);
```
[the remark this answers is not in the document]
pan-kot: There are two alternative APIs available but after diving deeper it looks like we can rely just on files. The items is there to cover d&d scenarios when the user drops mixed data e.g. a file and a raw text.

src/file-upload/file-option-thumbnail.tsx
```diff
@@ -0,0 +1,34 @@
...
+  };
+
+  return (
+    <div className={styles['file-option-thumbnail']} aria-hidden={true}>
```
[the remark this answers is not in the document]
pan-kot: The image has alt text set to the file name. Making it available to the screen readers won't provide any additional info.

src/file-upload/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+export { FileUploadProps };
+
+const FileUpload = React.forwardRef(
+  (
+    {
+      multiple = false,
+      showFileSize = false,
+      showFileLastModified = false,
+      showFileThumbnail = false,
```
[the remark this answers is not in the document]
just-boris: The current `InternalFileUploadProps` type definition if very far from my understanding of "explicit"

```js
type InternalFileUploadProps = SomeRequired<
  FileUploadProps,
  'multiple' | 'showFileSize' | 'showFileLastModified' | 'showFileThumbnail'
> &
  InternalBaseComponentProps;
```

This could be just  `type InternalFileUploadProps = FileUploadProps & InternalBaseComponentProps`

But I will leave it up to you. Anyone can change it later when they stumble upon this

src/file-upload/interfaces.ts
just-boris: Do you have a specific example? Because it is 1-way decision, if we find an example in the future, it will be too late, the value is already used.


For example, we use `null` in Select and it is clearly documented how it works: https://github.com/cloudscape-design/components/blob/471735b67df75c4cb69b1b7b201ef05ca6ddf2df/src/select/interfaces.ts#L143-L147

(it cannot use `undefined`, because it causes controllability warnings, but this explanation does not work for file upload)

src/file-upload/internal.tsx
just-boris: `setRemovedLastFile` only needed to move focus from 1 last item to the button. What does it have to do with 5 items use-case?

```suggestion
    if(value.length === 1) {
       // when removing last item, the focus should move to the input
       fileInputRef.current?.focus();
    }
```

src/test-utils/dom/file-upload/index.ts
```diff
@@ -0,0 +1,84 @@
...
+   * @param tokenIndex 1-based index of the file token to return.
+   */
+  findFileToken(fileTokenIndex: number): null | FileTokenWrapper {
+    if (fileTokenIndex === 1) {
+      return this.findComponent(`.${tokenGroupSelectors.token}`, FileTokenWrapper);
+    }
+    return this.findComponent(
+      `.${tokenListSelectors['list-item']}:nth-child(${fileTokenIndex}) .${tokenGroupSelectors.token}`,
+      FileTokenWrapper
+    );
```
just-boris, line 41: What is happening here, why is there a conditon?

src/token-group/token.tsx
just-boris: See the comment above https://github.com/cloudscape-design/components/pull/829/files?w=1#r1164379793. You can avoid any complexity in token group at all.

Not everything in this world needs to be `describedby`

src/file-upload/dropzone.tsx
just-boris: Timeouts are even more unreliable. We had similar issue in dropdown component: https://github.com/cloudscape-design/components/pull/541

Let's try harder and avoid timeouts. DM me if you are very stuck

src/file-upload/dropzone.tsx
just-boris: And we have a helper for this: https://github.com/cloudscape-design/components/blob/main/src/internal/utils/has-value.ts

```js
files = nullableFiles.filter(hasValue);
```

src/file-upload/file-input.tsx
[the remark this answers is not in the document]
just-boris: You can reference the button with your `aria-labelledby` just as well. It does not matter from which source it pulls text content

I will leave it up to you, but I still call it unnecessary complexity. As well as a minor bug that when copying this text, you will end up with "Upload Upload" in the clipboard

src/file-upload/styles.scss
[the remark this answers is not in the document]
just-boris: I would say that mixin should not have these `!important` too, but it is a big task to refactor. 

For the new code, let's not introduce any `!important`

src/file-upload/dropzone.tsx
pan-kot: Thanks, that is no longer applicable as I simplified the code.

src/file-upload/file-input.tsx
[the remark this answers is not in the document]
pan-kot: It does not work in Safari when aria-labellledby points to a content which is ARIA-hidden.

src/file-upload/index.tsx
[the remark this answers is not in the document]
pan-kot: Alright, I will change it now.

src/file-upload/internal.tsx
pan-kot: Ah, this one. Will update that

src/file-upload/interfaces.ts
pan-kot: I will remove the `null` option then. It is easy to add it later if requested.

src/token-group/token.tsx
pan-kot: Apart from that, what is the benefit of having it as a slot? Do we expect different error designs for the token?

src/test-utils/dom/file-upload/index.ts
pan-kot: When multiple=false at most one token is rendered and we do not use the token-list. When multiple=false we need to use token-list styles to locate the token by index.

just-boris: This implementation is based on fragile assumption then when there are multiple elements are found, it quietly clicks the first.

For example, this line returns multiple matches: 

https://github.com/cloudscape-design/components/blob/a6cc465c20ae4491e438a087c71f6bf80fe8c13f/src/file-upload/__integ__/file-upload.test.ts#L23

It works in Webdriver, but will fail in Cypress: 
https://docs.cypress.io/api/commands/click#Click-all-elements-with-id-starting-with-btn

just-boris (dismissed):

src/file-upload/styles.scss
```diff
@@ -9,3 +9,7 @@
 .hints {
   /* used in test-utils */
 }
```
just-boris: This is now unnecessary

pan-kot: It is not but I like mentioning it still. Like in other file I mention more class names:

```
.file-option-name,
.file-option-size,
.file-option-last-modified,
.file-option-thumbnail {
  /* used in test-utils */
}
```

It does not matter if there are styles defined for these or not, by looking at the file's header you know which classes are used as test selectors.

just-boris: I can accept this explanation of all utils are grouped together. But when there are only two selectors `.root, .hints` it is not useful

PR #832, opened 2023-03-14
fix: Remove spacing around null SpaceBetween children

gethinwebster:
### Description

Ensure that conditionally-rendered (and therefore potentially null/empty) children are handled by SpaceBetween. Also take the opportunity to modernize the approach (as we no longer need to support IE11).

Related links, issue #, if available: AWSUI-20570

### How has this been tested?

Screenshot tests run in dev pipeline

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/space-between/permutations.page.tsx
+++ b/pages/space-between/permutations.page.tsx
@@ -8,35 +8,55 @@ import ScreenshotArea from '../utils/screenshot-area';
 import styles from './styles.scss';
 import Container from '~components/container';
 
-const ExampleContent = () => <div className={styles.box}>This is some text.</div>;
+const ExampleContent = ({ renderNull }: { renderNull?: boolean }) =>
+  renderNull ? null : <div className={styles.box}>This is some text.</div>;
 
 /* eslint-disable react/jsx-key */
 const permutations = createPermutations<SpaceBetweenProps & { note?: string }>([
   {
     size: ['xxxs', 'xs', 's', 'm', 'l', 'xl', 'xxl'],
     direction: ['vertical', 'horizontal'],
-    children: [[<ExampleContent />, <ExampleContent />, <ExampleContent />]],
+    children: [
+      [
+        <ExampleContent renderNull={true} />,
+        <ExampleContent />,
+        <ExampleContent renderNull={true} />,
+        <ExampleContent />,
+        <ExampleContent />,
+        <ExampleContent renderNull={true} />,
+      ],
+    ],
   },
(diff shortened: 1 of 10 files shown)


src/area-chart/internal.tsx
```diff
@@ -111,7 +111,7 @@ export default function InternalAreaChart<T extends AreaChartProps.DataTypes>({
     visibleData: visibleSeries,
     statusType,
   });
-  const showFilters = statusType === 'finished' && (!isEmpty || isNoMatch);
+  const showFilters = statusType === 'finished' && (!isEmpty || isNoMatch) && (additionalFilters || !hideFilter);
```
gethinwebster, line 114: charts used to rely on negative margin around empty horizontal SpaceBetween to 'hide' the filter container

pages/space-between/permutations.page.tsx
```diff
@@ -8,35 +8,55 @@ import ScreenshotArea from '../utils/screenshot-area';
...
+        <ExampleContent renderNull={true} />,
+      ],
+    ],
   },
   {
     size: ['xs', 'xl'],
     direction: ['horizontal', 'vertical'],
     children: [
       [
+        () => null,
```
just-boris, line 35: This element gets filtered out by `react-keyed-flatten-children` library used in space between component (same will happen even if it returns a non-empty value) 

Is this the behavior you wanted to test here?

src/area-chart/internal.tsx
just-boris: Very interesting how come that we have this repeated for each chart instead of a reusable component, but this is a story for another time

src/space-between/styles.scss
```diff
@@ -50,33 +50,22 @@ $sizes-vertical: (
...
   flex-direction: column;
-}
 
-@each $name, $size in $sizes-vertical {
-  .child-vertical-#{$name}:not(:last-child) {
-    margin-bottom: $size;
+  @each $name, $size in $sizes-vertical {
+    &-#{$name} {
+      // This cancels out the margins of the children in the first row / in the first column.
+      gap: $size;
```
just-boris, line 68: should we use `row-gap` here for safety? (because we only need vertical spacing)

pages/space-between/permutations.page.tsx
gethinwebster: No, good callout, will adjust

just-boris (approved):

PR #839, opened 2023-03-15
fix: FindSplitPanel selector returns true without splitpanel

YueyingLu:
### Description

findSplitPanel() always return true in VR because `splitPanelStyles.root` is available in `aside` for tool and splitpanel toggle button. 
The fix is to remove this class from `aside`.
Also to change findOpenButton() not rely on SplitPanelWrapper because it is not always inside a split panel. 

Related links, issue #AWSUI-20594

### How has this been tested?

 Add unit test. Build succeed 6382342675

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/visual-refresh/tools.scss
+++ b/src/app-layout/visual-refresh/tools.scss
@@ -170,4 +170,6 @@ handleSplitPanelMaxWidth function in the context.
       z-index: 1;
     }
   }
+
+  @include styles.styles-reset;
 }
--- a/src/test-utils/dom/split-panel/index.ts
+++ b/src/test-utils/dom/split-panel/index.ts
@@ -1,6 +1,6 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
-import { ComponentWrapper, ElementWrapper } from '@cloudscape-design/test-utils-core/dom';
+import { ComponentWrapper, createWrapper, ElementWrapper } from '@cloudscape-design/test-utils-core/dom';
 import ButtonWrapper from '../button';
 import styles from '../../../split-panel/styles.selectors.js';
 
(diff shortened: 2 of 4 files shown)


src/test-utils/dom/split-panel/index.ts
```diff
@@ -20,7 +20,8 @@ export default class SplitPanelWrapper extends ComponentWrapper {
   }
 
   findOpenButton(): ButtonWrapper | null {
-    return this.findComponent(`.${styles['open-button']}`, ButtonWrapper);
+    const wrapper = createWrapper();
+    return wrapper.findComponent(`.${styles['open-button']}`, ButtonWrapper);
```
Al-Dani, line 24: do we have a test for the case when it's not inside Split Panel?

YueyingLu: Yes 
https://github.com/cloudscape-design/components/blob/29aa65765f15ac20a060ee17a6a5f7139159e6fa/src/app-layout/__integ__/app-layout-refresh-edge-controls.test.ts#L87-L89
https://github.com/cloudscape-design/components/blob/8f604c9f1d91153478199064b377b6d8184b4cf0/src/app-layout/__integ__/app-layout-focus-delegation.test.ts#L106-L109

Al-Dani (approved):

src/app-layout/visual-refresh/tools.scss
```diff
@@ -170,4 +170,6 @@ handleSplitPanelMaxWidth function in the context.
       z-index: 1;
     }
   }
+
+  @include styles.styles-reset;
```
just-boris, line 174: The order of `@include` matters, this way it will cancel all declarations above.

it is recommended to always write in this order, to not have any accidental overrides

```css
.show-tools {
  @include styles.styles-reset; 
  box-sizing: border-box;
  padding: awsui.$space-scaled-s awsui.$space-layout-toggle-padding;
}
```

YueyingLu: This come from SplitPanelStyles, which was the last one of the classname list https://github.com/cloudscape-design/components/pull/839/files#diff-21e6d2c3699425bb1b04025db0219b97b57bf68f7df16d23eccd4737791ee397L124. So I guess there was such overrides before? I am thinking running screenshot tests to see if this style reset is needed at all.

just-boris: Order of classes in HTML does not matter, only order in CSS files matters. If you do not believe - try by yourself

PR #843, opened 2023-03-16
feat: Make Modal footers sticky so that they are always visible

timogasda:
### Description

This change makes all Modal footers sticky, which means that they are visible regardless of the user's scrolling position.

This is useful for modals where the discoverability of the actions is crucial, for example in Collection Preferences.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/modal/styles.scss
```diff
@@ -106,6 +106,12 @@ $modal-z-index: 5000;
 .footer {
   @include container.divider;
   padding: container.$footer-padding;
+  background-color: awsui.$color-background-container-content;
+  border-radius: 0 0 awsui.$border-radius-container awsui.$border-radius-container;
```
connorlanigan, line 110: When the Modal is used with `disableContentPaddings`, this border radius lets the content shine through. Can we either apply the border radius only when the Modal has been scrolled all the way, or remove the border radius completely and add `overflow:hidden` to the Modal?

Screenshot of a testing page of the Modal component, showing a bit of text in the corner of the footer.

connorlanigan (changes requested):

src/modal/styles.scss
timogasda: good catch! I will look into that.

src/modal/internal.tsx
```diff
@@ -95,6 +96,10 @@ export default function InternalModal({
     }
   };
 
+  // We use an empty div element at the end of the content slot as a sentinel
+  // to detect when the user has scrolled to be bottom.
+  const { ref: stickySentinelRef, isIntersecting: footerStuck } = useIntersectionObserver();
```
timogasda, line 101: Sadly there is no native way to detect if a sticky element is stickied or not. Intersection Observer seems to be the most common pattern to work around this, and it seems to work quite well. Browser support also looks good: https://caniuse.com/intersectionobserver.

src/internal/hooks/use-intersection-observer/__tests__/use-intersection-observer.test.tsx
```diff
@@ -0,0 +1,43 @@
...
+describe('useIntersectionObserver', () => {
+  beforeEach(() => {
+    window.IntersectionObserver = mockIntersectionObserver;
+  });
+
+  it('should create an observer with the defined target element', async () => {
+    const { findByTestId } = render(<TestComponent />);
+    const target = await findByTestId('test');
+
+    expect(mockIntersectionObserver).toHaveBeenCalled();
```
connorlanigan, line 33: As far as I can tell this relies on the test order, because `mockIntersectionObserver` and `mockObserve` are reused across these test cases. Currently this should work fine since the second test case does not check the usage of the mocks, but to make this test file more future-proof, I'd recommend to call `jest.clearAllMocks()` ([docs](https://jestjs.io/docs/jest-object#jestclearallmocks)) in the `beforeEach` handler.

connorlanigan (changes requested):

src/internal/hooks/use-intersection-observer/__tests__/use-intersection-observer.test.tsx
timogasda: good catch!

connorlanigan (approved):

PR #845, opened 2023-03-16
feat: Internal generic token group component

pan-kot:
### Description

The generic token group is an internal component to reuse token group for arbitrary content. It is needed to support the addition of file upload component, see https://github.com/cloudscape-design/components/pull/829.

### How has this been tested?

Relying on the existing tests because the new component is an excerpt from the internal token group.

Added new test page to be used as dev. reference/playground.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/token-group/generic-token-group.tsx
+++ b/src/token-group/generic-token-group.tsx
@@ -1,0 +1,91 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import React, { forwardRef, useState } from 'react';
+import clsx from 'clsx';
+
+import { useUniqueId } from '../internal/hooks/use-unique-id';
+import { BaseComponentProps, getBaseProps } from '../internal/base-component';
+
+import SpaceBetween from '../space-between/internal';
+
+import { TokenGroupProps } from './interfaces';
+import DismissButton from './dismiss-button';
+import SelectToggle from './toggle';
+
+import styles from './styles.css.js';
+
+interface ItemAttributes {
+  disabled?: boolean;
+  dismissLabel?: string;
+  onDismiss?: () => void;
+}
+
+interface GenericTokenGroupProps<Item> extends BaseComponentProps {
+  items: readonly Item[];
+  limit?: number;
+  alignment: TokenGroupProps.Alignment;
+  renderItem: (item: Item, itemIndex: number) => React.ReactNode;
+  getItemAttributes: (item: Item, itemIndex: number) => ItemAttributes;
+  i18nStrings?: TokenGroupProps.I18nStrings;
+}
+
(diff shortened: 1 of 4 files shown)


src/token-group/generic-token-group.tsx
```diff
@@ -0,0 +1,91 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
gethinwebster, line 1: Do we have rules/guidelines on when reusable sub-components should go into `./internal/components/[sub-component]` vs `./[component]/[sub-component]`? Unfortunately it seems we have a bit of a mix of both currently :/

src/token-group/internal.tsx
```diff
@@ -1,86 +1,33 @@
...
-      )}
-    </div>
-  );
-}
-
-interface TokenProps extends TokenGroupProps.Item {
-  onDismiss?: () => void;
-}
-
-export function Token({ disabled, dismissLabel, onDismiss, ...props }: TokenProps) {
```
gethinwebster, line 71: do we know why this was exported, was it just a mistake? seems strange, but I can't find any usage of it anywhere in our codebase

src/token-group/generic-token-group.tsx
```diff
@@ -0,0 +1,91 @@
...
+}
+
+function GenericToken({ disabled, dismissLabel, onDismiss, children }: GenericTokenProps) {
+  return (
+    <div
+      className={clsx(styles.token, disabled && styles['token-disabled'])}
+      aria-disabled={disabled ? 'true' : undefined}
+    >
+      {children}
+      {onDismiss && <DismissButton disabled={disabled} dismissLabel={dismissLabel} onDismiss={onDismiss} />}
```
gethinwebster, line 88: conditional rendering based on presence of an event handler seems a bit strange, but I guess it's only a minor concern as this is internal

src/token-group/internal.tsx
pan-kot: It could have been a small mistake or maybe the Token has been separately tested at some point. Currently there are no references to it.

src/token-group/generic-token-group.tsx
pan-kot: I'll check if this can be made better explicit

src/token-group/generic-token-group.tsx
pan-kot: Typically we put internal components to under `/src/internal/components` but in this case it is a little controversial. In the file-upload we need to use sort of part of the `token-group` component, very similar how we use the popover internally. Instead of using the public popover in the annotation-context, breadcrumb-group, button-dropdown and chart-popover we use `./src/popover/container`. Both generic-token-group and popover-container are very tightly coupled to their public variants so I think keeping those closer is a good call.

gethinwebster (approved):

PR #847, opened 2023-03-17
chore: Add I18nProvider to pass localized strings to components (beta)

avinashbot:
### Description

As part of testing internationalization with a select few users, I've created a provider component (that currently lives in `internal` to give people a hint not to rely on it) that passes a formatter function down to any components that use those strings. The actual strings themselves are part of the next PR.

The format function "pattern" means that the "intl-messageformat" dependency and all of the ICU parsing logic (as well as the Typescript requirements it demands) are not included unless the top-level provider is actually imported, which is a nice bonus.

### How has this been tested?

Check those unit tests, baby!

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/internal/i18n/provider.tsx
```diff
@@ -0,0 +1,147 @@
...
+      // A general rule in the library is that undefined is basically
+      // treated as "not provided". So even if a user explicitly provides an
+      // undefined value, it will default to i18n values.
+      if (provided !== undefined) {
+        return provided;
+      }
+
+      // For the given locale and cache key, the computed message should
+      // always be the same. So we can safely return the cached value.
+      const cacheKey = `${namespace}.${component}.${key}`;
```
just-boris, line 63: Is the caching really necessary? Did you see any difference on lighthouse before adding this?

src/internal/i18n/provider.tsx
```diff
@@ -0,0 +1,147 @@
...
+  // Keep a cache per render for improved performance. Generally, it's expected
+  // to have an I18nProvider pretty high up the render tree and for it to
+  // rarely rerender, so it should stick around and offer performance gains.
+  const cache = useMemo(() => new Map<string, unknown>(), []);
+  cache.clear();
+
+  // The provider accepts an array of configs. We merge parent messages and
+  // flatten the tree early on so that accesses by key are simpler and faster.
+  const parentMessages = useContext(I18nMessagesContext);
+  const messages = useMemo(() => mergeMessages([parentMessages, ...messagesArray]), [parentMessages, messagesArray]);
```
just-boris, line 45: This memo is useless, as in all usage examples before, you pass a new array on every call

```
<I18nProvider messages={[MESSAGES]} locale="en"></I18nProvider>
```

tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "lib": ["ES2015", "DOM"],
+    "lib": ["ES2015", "ES2017.Intl", "ES2018.Intl", "ES2020.Intl", "DOM"],
```
just-boris, line 3: Is this a breaking change? Did you try a dry-run?

src/internal/i18n/provider.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+  // If a locale isn't provided, we can try and guess from the html lang,
+  // and lastly default to English. Locales have a recommended case, but are
+  // matched case-insensitively, so we lowercase it internally.
+  const locale = (providedLocale || document?.documentElement.lang || 'en').toLowerCase();
```
just-boris, line 50: Should it reuse the stuff from this file? https://github.com/cloudscape-design/components/blob/main/src/internal/utils/locale/normalize-locale.ts

tsconfig.json
avinashbot: Yep, it passed in dry-run.

src/internal/i18n/__tests__/i18n.test.tsx
```diff
@@ -0,0 +1,173 @@
...
+          props.nested?.nestedFunction,
+          format => props => format(props)
+        )?.({ type: 'function' })}
+      </li>
+    </ul>
+  );
+}
+
+it('detects locale from the html tag', () => {
+  document.documentElement.lang = 'es';
```
timogasda, line 54: I think this would affect other test cases, as well. Probably needs to be reset.

src/internal/i18n/provider.tsx
avinashbot: Dealing with caching for strings is wasted effort, yes - this was mostly intended for ensuring stability on custom override functions. I added this out of caution given the number of components that will adopt the format function. It's definitely not _required_, so I'm open to removing it.

src/internal/i18n/provider.tsx
avinashbot: I would prefer to keep this scoped to string handling as much as possible to make sure it's predictable (e.g. no attempting to get browser locale) — especially since it'll be used for object keys.

src/internal/i18n/provider.tsx
avinashbot: Not a strong requirement in this case either (since in tests or dev pages, we don't render that much stuff) But this leaves the option of tracking placing the array as a constant and providing it in to get performance benefits. There are two cases where this could be used:

1. At the top level of the application (maybe outside applayout, even outside router) where re-rendering isn't much of a concern, and the function will stay static no matter what.
2. As an override somewhere inside the application, where stability _does_ matter. And in this case, a `format` function recalculation would cascade down into every single Cloudscape component.

src/internal/i18n/provider.tsx
just-boris: How about risking date picker using one locale for the months and another for all provided strings?

src/internal/i18n/provider.tsx
avinashbot: Fair enough - we can rely on the beta process to watch for any complaints about performance.

src/internal/i18n/provider.tsx
avinashbot: This is already the risk right now in most places (react-intl, for instance, falls back to ["en"](https://github.com/formatjs/formatjs/blob/ff21b2a531997eba2ef1f3965f169e4e71654923/packages/intl/src/utils.ts#L59)). Either changing this to default to the OS locale breaks consistency between Cloudscape components, or changing date picker is a breaking change. I think the best course of action is to accept this.

In the long term, as more consumers adopt this provider, we could consider using that to determine the locale instead of using browser settings.

src/internal/i18n/__tests__/i18n.test.tsx
```diff
@@ -0,0 +1,177 @@
...
+    const { container } = render(
+      <I18nProvider messages={[MESSAGES, spanishMessages]}>
+        <TestComponent />
+      </I18nProvider>
+    );
+
+    expect(container.querySelector('#top-level-string')).toHaveTextContent('Custom Spanish string');
+    // Shouldn't default to English for non-existent strings.
+    expect(container.querySelector('#nested-string')).toHaveTextContent('');
+  } finally {
```
timogasda, line 76: I think this is what `beforeEach`/`afterEach` is for. I wouldn't mess with `expect` and how jest expects exceptions to be thrown.

src/internal/i18n/provider.tsx
timogasda: I'm in favor of aligning the i18n provider with whatever Date picker uses. But I'm fine with shipping this as-is for the Beta. Please make sure that we revisit this for the "proper" launch, though.

avinashbot: Makes sense - once we have every component supported by the provider, having the date picker follow the new Cloudscape recommended practice for internationalization instead of doing its own thing is the best solution to me.

src/internal/i18n/__tests__/i18n.test.tsx
avinashbot: It's a standard exception, but I just didn't want to mess with describe blocks. Not a strong opinion, updated :)

PR #849, opened 2023-03-17
chore: Add i18n provider support to beta components

avinashbot:
- experiment: Translation strings using messageformat
- Switch to intl-messageformat and implement POC for alert.
- Remove "overrides" as a concept and add flashbar.
- Add strings to rest of beta components (needs cleaning up).
- Allow language tags to match strings defined for wider languages
- Allow provider nesting.
- Remove basic dev page for now.
- Add dynamic formatter to property filter remove button aria label.
- Make internationalized properties optional.
- Add all languages and provide temporary fallback for property filter enteredTextLabel.
- Handle newly nullable values in property filter.
- Fix merge logic to not mutate and correctly downcase merged locales.
- Split locale files by language.
- Add i18n imports to package.json.
- Prevent mocking useContext from breaking useInternalI18n.
- Add test for locale fallbacks.
- Update documenter and remove unnecessary dependencies.
- chore: Add I18nProvider to pass localized strings to components (beta)
- chore: Add messages and convert them to ASTs during build time

### Description

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/property-filter/controller.ts
```diff
@@ -346,7 +346,7 @@ export const getAutosuggestOptions = (
 };
 
 export const operatorToDescription = (operator: ComparisonOperator, i18nStrings: I18nStrings) => {
-  const mapping: { [K in ComparisonOperator]: string } = {
+  const mapping: { [K in ComparisonOperator]: string | undefined } = {
```
jperals, line 349: Why this change?

src/form-field/internal.tsx
```diff
@@ -115,7 +117,10 @@ export default function InternalFormField({
       {(constraintText || errorText) && (
         <div className={styles.hints}>
           {errorText && (
-            <FormFieldError id={slotIds.error} errorIconAriaLabel={i18nStrings?.errorIconAriaLabel}>
+            <FormFieldError
```
gethinwebster, line 120: could we move this inside <FormFieldError> instead, that way we don't have to provide a separate string in TagEditor as well (where this internal component is reused)

src/property-filter/index.tsx
```diff
@@ -61,8 +73,58 @@ const PropertyFilter = React.forwardRef(
...
+      operatorGreaterText: format('i18nStrings.operatorGreaterText', rest.i18nStrings.operatorGreaterText),
+      operatorLessOrEqualText: format('i18nStrings.operatorLessOrEqualText', rest.i18nStrings.operatorLessOrEqualText),
+      operatorLessText: format('i18nStrings.operatorLessText', rest.i18nStrings.operatorLessText),
+      operatorText: format('i18nStrings.operatorText', rest.i18nStrings.operatorText),
+      operatorsText: format('i18nStrings.operatorsText', rest.i18nStrings.operatorsText),
+      propertyText: format('i18nStrings.propertyText', rest.i18nStrings.propertyText),
+      tokenLimitShowFewer: format('i18nStrings.tokenLimitShowFewer', rest.i18nStrings.tokenLimitShowFewer),
+      tokenLimitShowMore: format('i18nStrings.tokenLimitShowMore', rest.i18nStrings.tokenLimitShowMore),
+      valueText: format('i18nStrings.valueText', rest.i18nStrings.valueText),
+      removeTokenButtonAriaLabel: format(
```
gethinwebster, line 116: from seeing it here side by side, I wonder if tweaking naming might help readability. should the main `format` function be better named `i18n` or similar (as I don't think it's actually doing any formatting?), leaving only the _actual_ ICU formatter function that's passed to callbacks to be named `format`?

avinashbot: Good point. I've become used to the name, but `i18n` probably makes much more sense when used inside components. Will update.

src/form-field/internal.tsx
gethinwebster: yes (and that can stay), but moving the context usage into the subcomponent (and I guess making `FormFieldError.errorIconAriaLabel` optional) means that _we_ don't have to provide a new default for TagEditor

src/pagination/internal.tsx
```diff
@@ -87,9 +88,18 @@ export default function InternalPagination({
   ...rest
 }: InternalPaginationProps) {
   const baseProps = getBaseProps(rest);
-  const pageNumberLabelFn = ariaLabels?.pageLabel ?? defaultAriaLabels.pageLabel;
   const { leftDots, leftIndex, rightIndex, rightDots } = getPaginationState(currentPageIndex, pagesCount, openEnd);
 
+  const format = useInternalI18n('pagination');
+
+  const nextPageLabel = format('nextPageLabel', ariaLabels?.nextPageLabel) ?? defaultAriaLabels.nextPageLabel;
```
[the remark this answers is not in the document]
avinashbot: Our own i18n strings aren't always guaranteed to be present either. They only exist when the provider is present.

src/property-filter/controller.ts
avinashbot: Because the i18n strings that the function returns are now optional in Typescript.

src/flashbar/collapsible-flashbar.tsx
```diff
@@ -45,7 +46,7 @@ export default function CollapsibleFlashbar({ items, ...restProps }: FlashbarPro
     setInitialAnimationState(rects);
   }, [getElementsToAnimate]);
 
-  const { ariaLabel, baseProps, breakpoint, isReducedMotion, isVisualRefresh, mergedRef, ref } = useFlashbar({
```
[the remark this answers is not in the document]
avinashbot: I already did option A :)

I went with that to keep all the format stuff in one place rather than have `useFlashbar` call `useInternalI18n` just for _one_ string.

pages/alert/simple.page.tsx
```diff
@@ -7,40 +7,47 @@ import ScreenshotArea from '../utils/screenshot-area';
 import SpaceBetween from '~components/space-between';
 import styles from './styles.scss';
 
+import { I18nProvider } from '~components/internal/i18n';
+import messages from '~components/internal/i18n/messages/all.all';
+
+const MESSAGES = [messages];
```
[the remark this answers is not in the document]
avinashbot: No reason, both work :)
I'll move it back in for simplicity.

src/flashbar/collapsible-flashbar.tsx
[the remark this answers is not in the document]
jperals: Ah right, sorry I missed that part.

src/property-filter/index.tsx
jperals: Doesn't it format things like messages with interpolated values? Isn't it actually an alias to `intl.formatMessage` assuming that it is provided by the context?

gethinwebster (approved):

src/alert/internal.tsx
```diff
@@ -42,7 +42,7 @@ export default function InternalAlert({
   ...rest
 }: InternalAlertProps) {
   const baseProps = getBaseProps(rest);
-  const format = useInternalI18n('alert');
+  const i18n = useInternalI18n('alert');
```
jperals, line 45: Not sure about this change; doesn't `useInternalI18n` return `intl.formatMessage` or a similar function which will have to format the message (as in: find the translated version, interpolate values)?

Actually, would a name like `useFormatter` or `useI18nFormatter` make more sense for the hook?

Or am I misunderstanding how this works?

avinashbot: I think Gethin makes a good point - I think calling it a "formatter" at all isn't quite correct in the context of the components. Because in most cases, you aren't thinking about it in terms of formatting anything, you're just providing a key and a fallback, and letting the hook take care of picking the right string.

avinashbot: Adding to that: for dynamic values, you provide a custom handler, which _does_ include formatting as a concept (`format => props => format({ ... })`)

avinashbot: I started with that idea initially. But in the end, I chose this API for flexibility, since `customHandler: format => T` could return anything (e.g. objects, react nodes), and not just be limited to string functions.

PR #861, opened 2023-03-20
fix: Fix SSR support in I18nProvider

avinashbot:
### Description

Whoops. This one's on me. Thought optional chaining on `document` would save me here, but I would still need to explicitly check that the variable is even defined in SSR. 

### How has this been tested?

Wrote an SSR test to make sure this doesn't break.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/i18n/provider.tsx
+++ b/src/internal/i18n/provider.tsx
@@ -6,6 +6,7 @@ import IntlMessageFormat from 'intl-messageformat';
 import { MessageFormatElement } from '@formatjs/icu-messageformat-parser';
 
 import { InternalI18nContext, FormatFunction, CustomHandler } from './context';
+import { warnOnce } from '../logging';
 
 export interface I18nProviderProps {
   messages: ReadonlyArray<I18nProviderProps.Messages>;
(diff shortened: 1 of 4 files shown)


jperals (dismissed):

src/internal/i18n/provider.tsx
```diff
@@ -38,10 +38,19 @@ export function I18nProvider({ messages: messagesArray, locale: providedLocale,
   const parentMessages = useContext(I18nMessagesContext);
   const messages = mergeMessages([parentMessages, ...messagesArray]);
 
-  // If a locale isn't provided, we can try and guess from the html lang,
-  // and lastly default to English. Locales have a recommended case, but are
-  // matched case-insensitively, so we lowercase it internally.
-  const locale = (providedLocale || document?.documentElement.lang || 'en').toLowerCase();
+  let locale: string;
```
just-boris, line 41: Are you still sure you want to go your own hard way instead of reusing the existing code?
https://github.com/cloudscape-design/components/blob/main/src/internal/utils/locale/normalize-locale.ts

If not reusing, why not?

avinashbot: I'm a little cautious of _partially_ using the file, because the date component logic is wrong for this.

I mentioned in the earlier PR that I think the best course of action would be to make _this_ the definitive source of truth for internationalization and have the date components follow the provider instead. But you're right, maybe I can rename `normalizeLocale` to `normalizeLocaleForDateComponents` and have a nice global locale file for everything.

avinashbot: To expand further: there would be very little shared code - just two similar use cases bunched up into a single file just making things confusing to understand. Philosophically, it's more important for the locales to be used to match the correct strings in the messages JSON than to be valid or accurate BCP47 locale strings.

The context:
- shouldn't "fix" underscore delimited locale strings (because they're used for key accesses)
- shouldn't "auto-expand" the provided locale with the current browser locale or html locale
- shouldn't fall back to the browser locale (cause that makes SSR mismatches _even worse_)

avinashbot: For posterity: this rabbit hole is deeper than first expected. Writing a proposal for that, but in the mean time, there's very little opportunity for code sharing. Merging this fix for now.

PR #874, opened 2023-03-23
fix: Focus dispatching when token is removed

pan-kot:
### Description

Added explicit focus transitions when token group tokens are removed. The focus is transitioned to the next/prev available non-disabled items or, in a rare case, to the "show more" toggle.

Waiting for https://github.com/cloudscape-design/components/pull/876 to be merged first.

### How has this been tested?

New unit tests.

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/token-group/generic-token-group.tsx
```diff
@@ -50,19 +53,69 @@ function GenericTokenGroup<Item>(
   const baseProps = getBaseProps(props);
   const className = clsx(baseProps.className, styles.root, hasItems && styles['has-items']);
 
+  useEffect(() => {
```
gethinwebster, line 56: Presumably this is in an effect so that it's delayed until after the surrounding application has updated the state as a result of `onDismiss`? Can we reasonably rely on that? It seems like it's making quite a few assumptions about React internals and the surrounding application.

pan-kot: It makes the following assumptions:

1. onDismiss removes the target item and and doesn't add, remove or change position of other items;
2. onDismiss is happening with a state update w/o an extra async delay.

The first assumption is something we don't expect to be violated. The second can be violated e.g. if the update would be wrapped in a `setTimeout`. In that case the focus would remain on the current item and then transition to the next focusable element - same as it was before. If we ever have this situation the only way to fix that would be to expose the token refs. However, the current solution would still work as a reasonable default.

gethinwebster: ok, I think this makes sense. Might even steal this approach for some applayout focus handling fixes...

src/token-group/__tests__/token-group.test.tsx
```diff
@@ -169,4 +185,64 @@ describe('TokenGroup', () => {
...
+    test('Focus is dispatched to the next active token when non-last token is removed', () => {
+      const wrapper = renderStatefulTokenGroup({
+        items: [
+          { label: '1', dismissLabel: 'Remove 1' },
+          { label: '2', dismissLabel: 'Remove 2' },
+          { label: '3', dismissLabel: 'Remove 3', disabled: true },
+          { label: '4', dismissLabel: 'Remove 4' },
+        ],
+      });
+      act(() => wrapper.findToken(2)!.findDismiss().click());
```
just-boris, line 199: This `act` does nothing, after https://github.com/cloudscape-design/test-utils/pull/25

Why did you add it?

src/multiselect/__integ__/multiselect.test.ts
```diff
@@ -108,7 +108,7 @@ const createSetupTest = (wrapper: MultiselectWrapper) => (testFn: (page: Multise
 });
 
 describe('focus handling', () => {
-  test('focuses trigger after dismissing a token', () => {
+  test('focuses trigger after dismissing all tokens', () => {
```
gethinwebster, line 111: do we really want to change this behavior? do we know why we had the old behavior in the first place?

src/token-group/__tests__/token-group.test.tsx
pan-kot: Thanks, Boris - I'll check.

src/multiselect/__integ__/multiselect.test.ts
pan-kot: That looks reasonable to me that the tokens focusing behaviour is consistent no matter where the token group is used. However, I'll check if we do want a different treatment for multiselect.

gethinwebster: so long as someone on the design team has seen it it should be fine

just-boris (dismissed):

src/token-group/__tests__/token-group.test.tsx
```diff
@@ -169,4 +185,64 @@ describe('TokenGroup', () => {
...
+        items: [
+          { label: '1', dismissLabel: 'Remove 1' },
+          { label: '2', dismissLabel: 'Remove 2' },
+          { label: '3', dismissLabel: 'Remove 3', disabled: true },
+          { label: '4', dismissLabel: 'Remove 4' },
+        ],
+      });
+      wrapper.findToken(2)!.findDismiss().click();
+
+      expect(wrapper.findToken(3)!.findDismiss().getElement()).toHaveFocus();
```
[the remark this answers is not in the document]
pan-kot: I am preferring to use indices here specifically because the purpose of the test is to check if the correct next/prev element is referenced so the indices convey it better I think.

But I will consider having both assertions for better regression.

src/token-group/generic-token-group.tsx
```diff
@@ -19,8 +19,8 @@ interface ItemAttributes {
   disabled?: boolean;
   dismiss?: {
```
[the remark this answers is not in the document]
just-boris: I would suggest this API

```js
<GenericToken
   {...itemAttributes}
   dissmiss={itemAttributes.dismiss && <DismissButton ref={ref} {...itemAttributes.dismiss} />}
/>
```

This way you also skip one level of ref forwarding.

src/multiselect/__integ__/multiselect.test.ts
ethangeyer: UX approved :shipit:

gethinwebster (approved):

gethinwebster (approved):

PR #875, opened 2023-03-23
chore: Add i18n to autosuggest, date picker, and wizard

avinashbot:
### Description

Because the messages came back after the implementation, I missed a few components that should've been included in the latest batch of translation strings.

The translation files were also wrong because they swapped wizard and autosuggest. So I switched them out, sorted them alphabetically, and spaced them out so they looked nice and maintainable.

### How has this been tested?

Existing unit tests.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/calendar/grid/index.tsx
+++ b/src/calendar/grid/index.tsx
@@ -39,7 +39,7 @@ export interface GridProps {
   onFocusDate: (date: null | Date) => void;
   onChangeMonth: (date: Date) => void;
   startOfWeek: DayIndex;
-  todayAriaLabel: string;
+  todayAriaLabel?: string;
   selectedDate: Date | null;
   ariaLabelledby: string;
 }
--- a/src/calendar/interfaces.ts
+++ b/src/calendar/interfaces.ts
@@ -51,17 +51,17 @@ export interface CalendarProps extends BaseComponentProps {
   /**
    * Used as part of the `aria-label` for today's date in the calendar.
    */
-  todayAriaLabel: string;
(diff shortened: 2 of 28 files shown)


src/calendar/grid/index.tsx
```diff
@@ -39,7 +39,7 @@ export interface GridProps {
   onFocusDate: (date: null | Date) => void;
   onChangeMonth: (date: Date) => void;
   startOfWeek: DayIndex;
-  todayAriaLabel: string;
+  todayAriaLabel?: string;
```
gethinwebster, line 42: if we're marking this optional, presumably we should use the hook here https://github.com/cloudscape-design/components/pull/875/files#diff-c717613d99792450f8a0070b5630d6bbf3934cbe387130bc8861e5b20b87a946R159 rather than just in the date-picker?

src/calendar/interfaces.ts
```diff
@@ -51,17 +51,17 @@ export interface CalendarProps extends BaseComponentProps {
...
    * Used as part of the `aria-label` for today's date in the calendar.
    */
-  todayAriaLabel: string;
+  todayAriaLabel?: string;
 
   /**
    * Specifies an `aria-label` for the 'next month' button.
    */
-  nextMonthAriaLabel: string;
+  nextMonthAriaLabel?: string;
```
gethinwebster, line 59: ditto on these, maybe we should also update where the strings are stored in the message packs for clarity?

gethinwebster (approved):

PR #876, opened 2023-03-23
feat: Token group tokens use role group and can support list rendering

pan-kot:
### Description

Internal generic token group supports rendering as list (required for file upload and can potentially be made public).

When rendering as list is not used all token group tokens receive role="group" for improved navaigation.

### How has this been tested?

New unit tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/space-between/internal.tsx
+++ b/src/space-between/internal.tsx
@@ -8,12 +8,16 @@ import flattenChildren from 'react-keyed-flatten-children';
 import { SpaceBetweenProps } from './interfaces';
 import { InternalBaseComponentProps } from '../internal/hooks/use-base-component';
 
-type InternalSpaceBetweenProps = SpaceBetweenProps & InternalBaseComponentProps;
+type InternalSpaceBetweenProps = SpaceBetweenProps &
+  InternalBaseComponentProps & {
+    asList?: boolean;
+  };
 
 export default function InternalSpaceBetween({
   direction = 'vertical',
   size,
   children,
+  asList,
   __internalRootRef,
   ...props
 }: InternalSpaceBetweenProps) {
@@ -24,8 +28,11 @@ export default function InternalSpaceBetween({
    */
   const flattenedChildren = flattenChildren(children);
 
+  const ParentTag = asList ? 'ul' : 'div';
+  const ChildTag = asList ? 'li' : 'div';
+
   return (
-    <div
+    <ParentTag
       {...baseProps}
       className={clsx(baseProps.className, styles.root, styles[direction], styles[`${direction}-${size}`])}
       ref={__internalRootRef}
@@ -36,11 +43,11 @@ export default function InternalSpaceBetween({
         const key = (child as any).key;
 
         return (
-          <div key={key} className={clsx(styles.child, styles[`child-${direction}-${size}`])}>
+          <ChildTag key={key} className={clsx(styles.child, styles[`child-${direction}-${size}`])}>
             {child}
-          </div>
+          </ChildTag>
         );
       })}
-    </div>
+    </ParentTag>
   );
 }
--- a/src/token-group/__tests__/generic-token-group.test.tsx
+++ b/src/token-group/__tests__/generic-token-group.test.tsx
@@ -1,0 +1,40 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import GenericTokenGroup, { GenericTokenGroupProps } from '../../../lib/components/token-group/generic-token-group';
+import createWrapper, { TokenGroupWrapper } from '../../../lib/components/test-utils/dom';
+
+interface Item {
+  name: string;
+}
+
+const defaultProps: GenericTokenGroupProps<Item> = {
+  items: generateItems(5),
+  alignment: 'horizontal',
+  renderItem: item => item.name,
+  getItemAttributes: item => ({ name: item.name }),
+};
+
+function renderTokenGroup(props: Partial<GenericTokenGroupProps<Item>> = {}): TokenGroupWrapper {
+  const { container } = render(<GenericTokenGroup {...defaultProps} {...props} />);
+  return createWrapper(container).findTokenGroup()!;
+}
+
+function generateItems(count: number) {
+  return [...new Array(count)].map((_, index) => ({ name: `name-${index}` }));
+}
+
+describe('GenericTokenGroup', () => {
+  test('tokens have role="group" and aria-label set', async () => {
+    renderTokenGroup();
+
+    await expect(screen.findByRole('group', { name: 'name-0' })).resolves.toBeDefined();
+  });
+
+  test('tokens do not have role and aria-label set when `asList` is used', async () => {
+    renderTokenGroup({ asList: true });
+
+    await expect(screen.findByRole('group', { name: 'name-0' })).rejects.toBeDefined();
+  });
+});
--- a/src/token-group/generic-token-group.tsx
+++ b/src/token-group/generic-token-group.tsx
@@ -7,7 +7,7 @@ import clsx from 'clsx';
 import { useUniqueId } from '../internal/hooks/use-unique-id';
 import { BaseComponentProps, getBaseProps } from '../internal/base-component';
 
-import SpaceBetween from '../space-between/internal';
+import InternalSpaceBetween from '../space-between/internal';
 
 import { TokenGroupProps } from './interfaces';
 import DismissButton from './dismiss-button';
@@ -16,6 +16,7 @@ import SelectToggle from './toggle';
 import styles from './styles.css.js';
 
 interface ItemAttributes {
+  name: string;
   disabled?: boolean;
   dismiss?: {
     label?: string;
@@ -23,8 +24,9 @@ interface ItemAttributes {
   };
 }
 
-interface GenericTokenGroupProps<Item> extends BaseComponentProps {
+export interface GenericTokenGroupProps<Item> extends BaseComponentProps {
   items: readonly Item[];
+  asList?: boolean;
   limit?: number;
   alignment: TokenGroupProps.Alignment;
   renderItem: (item: Item, itemIndex: number) => React.ReactNode;
@@ -37,7 +39,7 @@ export default forwardRef(GenericTokenGroup) as <T>(
 ) => ReturnType<typeof GenericTokenGroup>;
 
 function GenericTokenGroup<Item>(
-  { items, alignment, renderItem, getItemAttributes, limit, ...props }: GenericTokenGroupProps<Item>,
+  { items, alignment, renderItem, getItemAttributes, asList, limit, ...props }: GenericTokenGroupProps<Item>,
   ref: React.Ref<HTMLDivElement>
 ) {
   const [expanded, setExpanded] = useState(false);
@@ -53,13 +55,13 @@ function GenericTokenGroup<Item>(
   return (
     <div {...baseProps} className={className} ref={ref}>
       {hasItems && (
-        <SpaceBetween id={controlId} direction={alignment} size="xs">
+        <InternalSpaceBetween id={controlId} direction={alignment} asList={asList} size="xs">
           {slicedItems.map((item, itemIndex) => (
-            <GenericToken key={itemIndex} {...getItemAttributes(item, itemIndex)}>
+            <GenericToken asGroup={!asList} key={itemIndex} {...getItemAttributes(item, itemIndex)}>
               {renderItem(item, itemIndex)}
             </GenericToken>
           ))}
-        </SpaceBetween>
+        </InternalSpaceBetween>
       )}
       {hasHiddenItems && (
         <SelectToggle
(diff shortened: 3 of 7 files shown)


src/token-group/__tests__/generic-token-group.test.tsx
```diff
@@ -0,0 +1,40 @@
...
+}
+
+describe('GenericTokenGroup', () => {
+  test('tokens have role="group" and aria-label set', async () => {
+    renderTokenGroup();
+
+    await expect(screen.findByRole('group', { name: 'name-0' })).resolves.toBeDefined();
+  });
+
+  test('tokens do not have role and aria-label set when `asList` is used', async () => {
```
gethinwebster, line 35: I'm not convinced by this logic - I don't think that `listitem` gives the same semantics/benefits as `group`, so I think we should use both

pan-kot: What is the difference? From testing it seems that it works just the same: https://d21d5uik3ws71m.cloudfront.net/components/7b802919d9f9e3d708cb24f9dab54e63bd408a67/index.html#/light/token-group/generic

We also do not use explicit groups in e.g. multiselect.

pan-kot: Keeping the group adds another level of nesting

pan-kot: When navigating the token it enters and exits the group twice if we keep the role="group" when also using a list.

https://user-images.githubusercontent.com/20790937/227233539-62898fac-00e6-480d-8c16-55ef24e6a1b5.mov

gethinwebster: got it, it sounds like the issue here is more VoiceOver

src/space-between/internal.tsx
```diff
@@ -24,8 +28,11 @@ export default function InternalSpaceBetween({
    */
   const flattenedChildren = flattenChildren(children);
 
+  const ParentTag = asList ? 'ul' : 'div';
```
gethinwebster, line 31: allowing `aria-setsize` to be set here (e.g. when tokens are hidden) might be needed, based on discussions

src/token-group/generic-token-group.tsx
gethinwebster, line 1: should we start using this in multiselect/property filter straight away?

pan-kot: Sounds good - I'll add this

src/space-between/internal.tsx
pan-kot: Right, my plan is to change `asList` to `listProps` going forward as soon any list-specific property is required.

src/space-between/internal.tsx
```diff
@@ -8,12 +8,16 @@ import flattenChildren from 'react-keyed-flatten-children';
 import { SpaceBetweenProps } from './interfaces';
 import { InternalBaseComponentProps } from '../internal/hooks/use-base-component';
 
-type InternalSpaceBetweenProps = SpaceBetweenProps & InternalBaseComponentProps;
+type InternalSpaceBetweenProps = SpaceBetweenProps &
+  InternalBaseComponentProps & {
+    asList?: boolean;
```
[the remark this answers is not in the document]
pan-kot: The challenge is that space-between supports two different directions and also responsiveness. We want to keep that no matter if list is used or not. Using Box won't be enough, we can use tokens but that would require replicating complex space-between CSS everywhere. I rather like making it a public API but it is not necessary to fast-forward it. It would be great to use it internally at first to better understand what list props we'd like to support.

src/space-between/internal.tsx
```diff
@@ -25,22 +29,26 @@ export default function InternalSpaceBetween({
...
       {flattenedChildren.map(child => {
-        // If this react child is a primitive value, the key will be undefined
-        // eslint-disable-next-line @typescript-eslint/no-explicit-any
-        const key = (child as any).key;
+        const key = typeof child === 'object' ? child.key : undefined;
+
+        // When using "ul" variant we avoid extra element wrappers to keep list semantics.
+        if (Variant === 'ul' && typeof child === 'object') {
+          const className = clsx(child.props.className, styles.child, styles[`child-${direction}-${size}`]);
+          return React.cloneElement(child, { className });
```
pan-kot, line 43: I kind of like the way it works - instead of creating additional div elements this applies styles directly to passed children. However, it requires complex assumptions: children must propagate className; children styles must not define margins.

I wonder if there is a better solution for this. E.g. we can create an internal List element and use it interchangeable. However, that would require copying space-between styles. Wdyt?

pan-kot: In case we create a separate list component it would require a slightly different API from space-between if we want to avoid complex assumptions on React children. In that case using space-between and list-space-between interchangeably could be difficult. However, I am not sure if we need to - how about making the token group a list by default?

gethinwebster: I agree TokenGroup could be a list by default

src/internal/components/token-list/styles.scss
```diff
@@ -0,0 +1,74 @@
...
+}
+
+.list {
+  display: flex;
+  list-style: none;
+  padding: 0;
+  margin: 0;
+}
+
+.list-horizontal {
```
pan-kot, line 21: Here we use same styles as in space-between but simplified

src/internal/components/option/index.tsx
```diff
@@ -61,6 +62,7 @@ const Option = ({
 
   return (
     <span
+      role={isPresentational ? 'presentation' : undefined}
```
pan-kot, line 65: The presence of title and aria-disabled makes screen-readers consider the element as group which is unwanted when the group role is defined above. For token groups the group must include the dismiss button.

src/property-filter/__tests__/property-filter.test.tsx
```diff
@@ -718,11 +718,11 @@ describe('property filter parts', () => {
       expect(wrapper.findTokens()!).toHaveLength(1);
       expect(wrapper.findTokens()![0].findLabel().getElement()).toHaveTextContent('string : first');
       // show more
-      act(() => wrapper.findTokenToggle()!.find('button')!.click());
```
pan-kot, line 721: I updated the test-utils to return the actual button instead of a wrapper. I'll make a dry-run to ensure the change is safe.

src/internal/components/token-list/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+  return (
+    <div className={clsx(styles.root, styles[`root-${toggleAlignment}`])}>
+      <ul id={controlId} className={clsx(styles.list, styles[`list-${alignment}`])}>
+        {slicedItems.map((item, itemIndex) => {
+          const { ariaLabel, disabled } = getItemAttributes(item, itemIndex);
+          return (
+            <li
+              key={itemIndex}
+              aria-label={ariaLabel}
```
just-boris, line 45: If this label is required it means that token content is never used. Is this expected?

just-boris: Checked in Chrome devtools, the content of token group is marked as "Ignored" in AOM. This does not look good at all

image

src/internal/components/option/index.tsx
gethinwebster: I think having explicitly contradictory attributes like this is not a good thing, assuming they're already provided on the parent can't we remove them entirely?

src/internal/components/token-list/index.tsx
gethinwebster: (this is likely due to the `role="presentation"` I think https://github.com/cloudscape-design/components/pull/876/files#r1149456787)

pan-kot: The role="presentation" makes sure the internally used option is not treated as a group. Otherwise there are two groups - the token itself and the option with the last one being not complete as it does not include the dismiss button.

The testing with screen-readers does not reveal any problems yet - all content is announced and properly grouped. Could you suggest a testing scenario that confirms the change causes content not being accessible?

just-boris: Extra content in tokens (description and tags) is concealed when aria-label is used: https://d21d5uik3ws71m.cloudfront.net/components/df9ba4bdca99f9addd3f836354a0b24238cad39c/index.html#/light/token-group/permutations

> The testing with screen-readers does not reveal any problems yet

Did the missing labels revealed any problems? Why not to start with a simple option?

src/internal/components/option/index.tsx
pan-kot: If we do not set role="presentation" the assumed role is "group" (at least in VO), probably because there is a title and aria-disabled available. That makes token include two groups - one is the LI item which includes the dismiss control and one is the option which does not include the dismiss.

pan-kot: Sounds reasonable. How about I remove the role="presentation" part for now which can be justified by the fact the change improves a11y nevertheless? Then I can make the related changes as a follow-up. Wdyt?

gethinwebster: works for me!

src/internal/components/token-list/index.tsx
just-boris: @pan-kot still see the same outcome in the latest revision.

I do not think any parent role changes the behavior or aria label. It is an ultimate feature – when you apply the label, all inner content gets ignored.

Still curious to know reasons why you added the label in the first place

just-boris: Also, [quoting MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label)

> Note: aria-label is intended for use on interactive elements, or elements made to be interactive via other ARIA declarations

This does not sound suitable for this use-case

pan-kot: The ARIA label was added based on the feedback from a11y - it maintains the context so that if you navigate with Tab to the dismiss button the group name described with ARIA label will be announced. Thanks for sharing the concerns though, I'll double-check the approach.

gethinwebster (approved):

src/internal/components/token-list/index.tsx
just-boris: When I open in my inspector, I see the expanded content, but only when focusing one particular node. When going up or down, the tree shows "ignored"

https://user-images.githubusercontent.com/812240/228520387-3e45c95e-de8f-48b0-a0b2-b2425a6700cc.mov

just-boris: I also double checked what it means in Voice Over interpretation. Currently it announces the title twice (label causes the duplication, I am guessing).

> The ARIA label was added based on the feedback from a11y - it maintains the context so that if you navigate with Tab to the dismiss button the group name described with ARIA label will be announced

The `dismissLabel` is unique for each item. What extra context missing, I do not understand.

pan-kot: Here is the feedback I got:

"We want a label on the item (which is a group) so that when a user reaches a control inside of it they know which one they are affecting. The user may not have reached that control through linear or tab navigation, it could be the result of navigating by button, for example."

Of course, as it appears the addition of the aria-label causes more issues the decision needs to be re-evaluated. I'll check it

pan-kot: Turns out the "Ignored" property assigned in Chrome (but not e.g. Firefox) is not the same as hidden. It also applies e.g. when an element has role "presentation" or "none" set and means that the element's semantics not exposed to the a11y tree so the element e.g. can be excluded from the list this way. However, it does not prevent screen-readers to navigate inside. W/o "aria-label" set and "Ignored" assigned the option title is announced when entering the group. When the node is ignored its contents is not announced when entering the group and is being replaced by the ARIA-label contents. However, having an aria label set has benefits and now when the group element is navigated by clicking or using Tab navigation the group name is announced.

You can see there are plenty of "Ignored" nodes in e.g. our multselect test page.

Screenshot 2023-03-31 at 15 40 56

src/internal/components/token-list/styles.scss
```diff
@@ -10,20 +10,34 @@
...
-  display: flex;
-  flex-wrap: wrap;
-  gap: awsui.$space-xs;
-}
-.list-horizontal {
-  flex-direction: row;
-}
-.list-vertical {
-  flex-direction: column;
+  display: contents;
```
pan-kot, line 24: `display: contents` allows to omit the list from rendering so that the parent flex container styles apply. That makes list items and the following "show more" and "clear filters" buttons to be rendered in the same flex container which is required for the property-filter.

There is a risk that `display: contents` would also exclude the element from the a11y tree. My initial round of testing did not reveal any issues with it though.

src/token-group/token.tsx
```diff
@@ -21,10 +20,7 @@ interface TokenProps extends ItemAttributes {
 
 export function Token({ disabled, dismissLabel, onDismiss, children }: TokenProps) {
   return (
-    <div
-      className={clsx(styles.token, disabled && styles['token-disabled'])}
-      aria-disabled={disabled ? 'true' : undefined}
```
pan-kot, line 26: The Token is a visual but not semantic container. When rendered with TokenList the aria-disabled is set on list elements. When rendered as standalone the aria-disabled must be provided separately:

```
<div role="group" aria-label="Token" aria-disabled={true}>
  <Token>Token</Token>
<div>
```

src/property-filter/__tests__/property-filter.test.tsx
pan-kot: Verified here: https://github.com/cloudscape-design/components/pull/899

src/internal/components/option/index.tsx
```diff
@@ -59,15 +60,17 @@ const Option = ({
     />
   );
 
+  // When generic props are assigned assistive technologies treat option nodes as semantic (generic) groups.
+  // See https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/generic_role
+  const genericGroupProps = isGenericGroup
```
gethinwebster, line 65: I've read the comment, the linked page, and the code a few times now, and still don't understand what this is doing. Perhaps it's too early on a Monday, but perhaps this could be expanded?

gethinwebster: maybe also expanding on _why_ (I think it's because sometimes these attributes are set on the parent instead?)

gethinwebster (approved):

PR #888, opened 2023-03-24
feat: spinner variant for alert

32teeth:
### Description

We have a use case where the `<Alert />` component is on the stage while a backend validation call occurs. This might be a long call round trip and we would like to persist the `<Alert />` component with an `status-in-progress` icon.

> What was added

* `progress` typeToIcon
* `progress` to tests
* `progress` to styles

> Variants

Screen Shot 2023-03-24 at 16 03 37

> Animated SVG

spinner


johannes-weber:
Thank you for your PR! We want to make sure to fully understanding the specific use case you described. Could you share more details about the context, the user is in, when you'd like to display the in progress alert?

In General, for an action that takes anywhere from 1-10 seconds using a spinner (e.g. in the button the user triggers the backend validation) is the recommended pattern. Find more details in our [feedback mechanisms article](https://cloudscape.design/patterns/general/user-feedback/).

32teeth:
Any updates?

PR #892, opened 2023-03-27
fix: Ensure AppLayout only changes focus on user interactions

gethinwebster:
### Description

The previous focus management approach was prone to focusing when it shouldn't, e.g. on page load. This reworks the logic to only change focus when specifically the result of user interaction.

Related links, issue #, if available: AWSUI-20639

### How has this been tested?

Updated tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/app-layout/__tests__/common.test.tsx
```diff
@@ -187,16 +187,6 @@ describeEachAppLayout(() => {
         expect(findToggle(wrapper).getElement()).toBe(document.activeElement);
       });
 
-      test('updates active element in controlled mode', () => {
```
jperals, line 190: Is this test not relevant anymore?

jperals: OK I see. Would it make sense to have coverage for the case of focusing via user interactions then? (Maybe by changing this test instead of removing it)

src/app-layout/utils/use-focus-control.ts
```diff
@@ -18,15 +20,15 @@ export function useFocusControl(isOpen: boolean, restoreFocus = false): FocusCon
     close: useRef<ButtonProps.Ref>(null),
   };
   const previousFocusedElement = useRef<HTMLElement>();
+  const [shouldFocus, setShouldFocus] = useState(false);
```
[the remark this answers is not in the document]
gethinwebster: it's expected that the parent application changes state whenever this flag will change anyway, so does it actually make much difference here?

src/app-layout/__tests__/common.test.tsx
gethinwebster: this is covered in the integration tests

src/app-layout/utils/use-focus-control.ts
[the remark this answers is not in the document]
just-boris: The focus changes only when a state of a panel changes and this is definitely a react state change. Maybe we should keep `isOpen` in the effect dependencies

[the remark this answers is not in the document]
gethinwebster: makes sense, now that I look at it again I think the refs approach better represents what we're actually trying to achieve

src/app-layout/__integ__/app-layout-focus-delegation.test.ts
```diff
@@ -9,162 +9,197 @@ const wrapper = createWrapper().findAppLayout();
...
+                await expect(page.isFocused(wrapper.findToolsClose().toSelector())).resolves.toBe(true);
+              },
+              { pageName: 'with-fixed-header-footer', visualRefresh, mobile }
+            )
           );
-        },
-        { pageName: 'with-split-panel', visualRefresh, splitPanelPosition: 'side' }
-      )
+          // tests not relevant for mobile, as panel overlays content
+          if (!mobile) {
```
just-boris, line 161: If tests are not relevant for mobile, they can simply live outside the `forEach(mobile => {..})` block.

Personally I find it better to read than keeping in mind that `mobile` var is always `false` down in this condition

src/app-layout/utils/use-split-panel-focus-control.ts
```diff
@@ -0,0 +1,46 @@
...
+        refs.preferences.current?.focus();
+        break;
+    }
+    lastInteraction.current = null;
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, dependencies);
+
+  return {
+    refs,
+    setLastInteraction: (interaction: SplitPanelLastInteraction) => (lastInteraction.current = interaction),
```
just-boris, line 44: I think, this is a moment where it would be more readable to change the API of this hook to make it easier to understand that this call does not trigger a re-render on itself, without looking into the implementation

just-boris (approved):
Looks good, the comments are optional

jperals (approved):

PR #899, opened 2023-03-29
refactor: Token group list

pan-kot:
### Description

Reuse code between token-group and property-filter components by introducing internal token-list component.

This is also required to make an a11y improvement for token-group, multiselect and property-filter that all use token-list, see: https://github.com/cloudscape-design/components/pull/876.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/property-filter/__tests__/property-filter.test.tsx
```diff
@@ -718,11 +718,11 @@ describe('property filter parts', () => {
       expect(wrapper.findTokens()!).toHaveLength(1);
       expect(wrapper.findTokens()![0].findLabel().getElement()).toHaveTextContent('string : first');
       // show more
-      act(() => wrapper.findTokenToggle()!.find('button')!.click());
+      act(() => wrapper.findTokenToggle()!.click());
       expect(wrapper.findTokenToggle()!.getElement()).toHaveTextContent(i18nStrings.tokenLimitShowFewer);
       expect(wrapper.findTokens()!).toHaveLength(2);
       // show fewer
-      act(() => wrapper.findTokenToggle()!.find('button')!.click());
```
pan-kot, line 725: This can potentially affect customers. I will only merge the PR after a successful dry-run.

src/internal/components/token-list/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+import React, { useState } from 'react';
+import clsx from 'clsx';
+
+import { SelectToggle } from './select-toggle';
+import styles from './styles.css.js';
+import { I18nStrings } from './interfaces';
+import InternalSpaceBetween from '../../../space-between/internal';
+import { useUniqueId } from '../../hooks/use-unique-id';
+
+interface TokenListProps<Item> {
```
timogasda, line 13: minor: If you already have an `interfaces.ts`, you might as well put the Props interface in there

src/internal/components/token-list/select-toggle.tsx
```diff
@@ -1,29 +1,30 @@
...
   allHidden: boolean;
   expanded: boolean;
   numberOfHiddenOptions: number;
   onClick?: NonCancelableEventHandler<null>;
-  i18nStrings?: Pick<TokenGroupProps.I18nStrings, 'limitShowFewer' | 'limitShowMore'>;
+  i18nStrings?: I18nStrings;
 }
 
-const SelectToggle = ({
+export function SelectToggle({
```
timogasda, line 20: minor: Probably out of scope, but the name of this component is confusing to me. Something like `DisclosureToggle`, `ListDisclosureToggle`, or `ItemLimitToggle` maybe?

src/token-group/token.tsx
```diff
@@ -0,0 +1,34 @@
...
+import React from 'react';
+import clsx from 'clsx';
+
+import DismissButton from './dismiss-button';
+
+import styles from './styles.css.js';
+
+interface ItemAttributes {
+  children: React.ReactNode;
+  dismiss?: {
```
timogasda, line 13: Do we really need another object layer here? We do this very rarely, and for things like `i18nStrings` it makes sense because they usually have a lot of keys, and they can be reused across components.

Here I think it would be more straight forward to have `dismissLabel` and `onDismiss` as top-level props.

src/property-filter/index.tsx
```diff
@@ -256,11 +255,6 @@ const PropertyFilter = React.forwardRef(
 
       fireNonCancelableEvent(onLoadItems, { ...loadMoreDetail, firstPage: true, samePage: false });
     };
-    const [tokensExpanded, setTokensExpanded] = useState(false);
-    const toggleExpandedTokens = () => setTokensExpanded(!tokensExpanded);
-    const hasHiddenOptions = tokenLimit !== undefined && tokens.length > tokenLimit;
-    const slicedTokens = hasHiddenOptions && !tokensExpanded ? tokens.slice(0, tokenLimit) : tokens;
-    const controlId = useMemo(() => generateUniqueId(), []);
```
timogasda, line 263: nice!

timogasda (commented):
Looks great overall! Just minor comments.

src/internal/components/token-list/select-toggle.tsx
pan-kot: Thanks, I like the ItemLimitToggle 👍

src/token-group/token.tsx
timogasda: I think we have precedence in other parts of our code where we toggle features based on the event handler. So I don't think this is a big deal. Alternatively, we could add another property `dismissible` - we do have this pattern in some public components too.

But for me the condition with `onDismiss` is fine here.

src/token-group/internal.tsx
```diff
@@ -8,28 +8,51 @@ import checkControlled from '../internal/hooks/check-controlled';
...
-      })}
-    />
+    >
+      <TokenList
+        alignment={alignment}
+        items={items}
+        limit={limit}
+        renderItem={(item, itemIndex) => (
+          <Token
+            key={itemIndex}
```
pan-kot, line 46: Currently the `key` property is implicitly required because the items are passed directly to the space-between. That is going to change after https://github.com/cloudscape-design/components/pull/876 is merged. Then, each item will be wrapped with `li` in the token-list so the keys will no longer be needed.

gethinwebster (approved):

timogasda (approved):

PR #922, opened 2023-04-04
feat: Improve focus behavior of table inline editing

timogasda:
### Description

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/components/focus-lock/__integ__/focus-lock.test.ts
+++ b/src/internal/components/focus-lock/__integ__/focus-lock.test.ts
@@ -1,0 +1,41 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import useBrowser from '@cloudscape-design/browser-test-tools/use-browser';
+import { BasePageObject } from '@cloudscape-design/browser-test-tools/page-objects';
+import createWrapper, { ElementWrapper } from '../../../../../lib/components/test-utils/selectors';
+
+function setupTest(testFn: (page: BasePageObject, wrapper: ElementWrapper) => Promise<void>) {
+  return useBrowser(async browser => {
+    await browser.url('#/light/focus-lock');
+    const page = new BasePageObject(browser);
+    const wrapper = createWrapper();
+    await page.waitForVisible(wrapper.findButton().toSelector());
+    await testFn(page, wrapper);
+  });
+}
+
+describe('FocusLock', () => {
+  test(
+    'focuses first element with autofocus',
+    setupTest(async (page, wrapper) => {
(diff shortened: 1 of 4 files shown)


src/internal/components/focus-lock/__integ__/focus-lock.test.ts
```diff
@@ -0,0 +1,41 @@
...
+  return useBrowser(async browser => {
+    await browser.url('#/light/focus-lock');
+    const page = new BasePageObject(browser);
+    const wrapper = createWrapper();
+    await page.waitForVisible(wrapper.findButton().toSelector());
+    await testFn(page, wrapper);
+  });
+}
+
+describe('FocusLock', () => {
```
pan-kot, line 17: Can this be tested in a unit test instead?

pages/table/editable.page.tsx
```diff
@@ -207,7 +207,7 @@ const Demo = forwardRef(
       let value = newValue;
       await new Promise(r => setTimeout(r, 1000));
       errorsMeta.delete(currentItem);
-      if (value === 'inline') {
+      if (typeof value === 'string' && value.includes('inline')) {
         errorsMeta.set(currentItem, 'Server does not accept this value, try another');
         throw new Error('Inline error');
```
pan-kot, line 212: Should we reflect the server error in the demo? The focusing works but there is no error state/text visible.

src/table/body-cell/inline-editor.tsx
```diff
@@ -93,45 +97,44 @@ export function InlineEditor<ItemType>({
...
-                loading={currentEditLoading}
-              />
-            </SpaceBetween>
-          </span>
-        </div>
-      </FormField>
-    </form>
+    <FocusLock restoreFocus={true} ref={focusLockRef}>
+      <div role="dialog" aria-label={ariaLabels?.activateEditLabel?.(column)} onKeyDown={handleEscape}>
+        <form ref={clickAwayRef} onSubmit={onSubmitClick} className={styles['body-cell-editor-form']}>
```
pan-kot, line 102: Should we move the `clickAwayRef` to the dialog?

pan-kot (commented):
It would be great to update the demo with server error showing up somewhere to better understand how the feature is supposed to work and test it for a11y.

src/internal/components/focus-lock/__integ__/focus-lock.test.ts
timogasda: You could probably craft something that kinda works. But when it comes to things like focus management, I have much more confidence in an integration test. I try to use them sparingly, but I think in this case it's the better path.

pages/table/editable.page.tsx
timogasda: It already displays the server side error "Server does not accept this value, try another" (if there isn't another client-side error present already)

pan-kot (approved):

PR #233, opened 2022-09-01
feat: Use React's `useId` hook if available

connorlanigan:
### Description

This updates our `useUniqueId` hook to use [the native `useId` hook provided by React](https://reactjs.org/docs/hooks-reference.html#useid). If the native hook is not available, it falls back to our existing implementation.

The native `useId` hook has the advantage of generating IDs that are stable between server-rendering and client-side hydration, thus preventing hydration errors.

### How has this been tested?

[_How did you test to verify your changes?_]

Upgraded the dev dependency on React and ReactDOM to 18, and inspected the dev pages. They use the new IDs provided by React.

[_How can reviewers test these changes efficiently?_]
see above

[_Check for unexpected visual regressions, see [`CONTRIBUTING.md`](CONTRIBUTING.md#run-visual-regression-tests) for details._]

### Documentation changes

[*Do the changes include any API documentation changes?*]
- [ ] _Yes, this change contains documentation changes._
- [X] _No._

### Related Links

- AWSUI-9561

<details>
   <summary>Review checklist</summary>

_The following items are to be evaluated by the author(s) and the reviewer(s)._

#### Correctness

- [X] _Changes are backward-compatible if not indicated, see [`CONTRIBUTING.md`](CONTRIBUTING.md#public-apis)._
- [X] _Changes do not include unsupported browser features, see [`CONTRIBUTING.md`](CONTRIBUTING.md#browsers-support)._
- [ ] _Changes were manually tested for accessibility, see [accessibility guidelines](https://cloudscape.design/foundation/core-principles/accessibility/)._

#### Security

- [ ] _If the code handles URLs: all URLs are validated through [the `checkSafeUrl` function](https://github.com/cloudscape-design/components/blob/main/src/internal/utils/check-safe-url.ts)._

#### Testing

- [ ] _Changes are covered with new/existing unit tests?_
- [ ] _Changes are covered with new/existing integration tests?_
</details>

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #930, opened 2023-04-05
feat: Use native useId hook if available

just-boris:
### Description

Revived version of #233

Native hook produces deterministic ids in server-side rendering, which is useful feature and not possible to adopt without changes in our components

### How has this been tested?

* Our demos package uses React 18, so it will cover that condition
* Also deployed to my dev pipeline to have a closer look at all demos

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/attribute-editor/row.tsx
+++ b/src/attribute-editor/row.tsx
@@ -11,7 +11,7 @@ import { fireNonCancelableEvent, NonCancelableEventHandler } from '../internal/e
 import InternalGrid from '../grid/internal';
 import { InternalButton } from '../button/internal';
 import clsx from 'clsx';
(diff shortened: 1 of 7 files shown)


avinashbot (approved):
omg **yes**. i'll accept the demos justification, sounds fair to me. since it's a very simple core react hook, i would expect useId assumptions to only ever break in very obvious ways that demos tests would catch.

(there's one file `src/autosuggest/__tests__/autosuggest.test.tsx` that mocks `generateUniqueId` which does nothing now, but it also doesn't break anything, so that's optional if you want to fix it)

avinashbot (approved):

PR #931, opened 2023-04-06
fix: Use `header` slot of ExpandableSection for tutorial steps

gethinwebster:
### Description

The `headerText` slot is designed only for plain text, so applying styling within it causes unexpected side-effects, but that styling is necessary for 

### How has this been tested?

Tested locally, will follow-up with improved screenshot test coverage

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/tutorial-panel/components/tutorial-detail-view/task.tsx
+++ b/src/tutorial-panel/components/tutorial-detail-view/task.tsx
@@ -58,7 +58,7 @@ export function Task({ task, taskIndex, currentTaskIndex, expanded, onToggleExpa
 
         <div className={styles['expandable-section-wrapper']}>
           <InternalExpandableSection
-            headerText={
+            header={
               <span className={styles['expandable-section-header']}>
                 {i18nStrings.labelTotalSteps(task.steps.length)}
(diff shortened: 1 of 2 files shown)


src/tutorial-panel/components/tutorial-detail-view/task.tsx
```diff
@@ -58,7 +58,7 @@ export function Task({ task, taskIndex, currentTaskIndex, expanded, onToggleExpa
 
         <div className={styles['expandable-section-wrapper']}>
           <InternalExpandableSection
-            headerText={
+            header={
```
pan-kot, line 61: Do we plan to update this in some future?

pan-kot (approved):

src/tutorial-panel/components/tutorial-detail-view/task.tsx
gethinwebster: I think it would make sense, but the current reality is that having non-standard-sized headings in expandable section isn't really supported by the component, so might need some additions there to support this in a more elegant way.

PR #934, opened 2023-04-06
feat: Performance improvements for focus-visible implementation

just-boris:
### Description

Remove react state and avoid full page re-render when switching focus-visible state

### How has this been tested?

* Locally, visited a few pages and paid extra attention to the most heavily changed components
  * charts
  * date-range-picker
  * select
  * table and its inline editing
 * Screenshot tests in my pipeline
* Any extra checks from the reviewer is appreciated

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/date-range-picker/calendar/grids/index.tsx
+++ b/src/date-range-picker/calendar/grids/index.tsx
@@ -9,7 +9,6 @@ import InternalSpaceBetween from '../../../space-between/internal';
 import { Grid } from './grid';
 import styles from '../../styles.css.js';
 
-import useFocusVisible from '../../../internal/hooks/focus-visible/index';
 import { hasValue } from '../../../internal/utils/has-value';
 import { useDateCache } from '../../../internal/hooks/use-date-cache';
 import { moveNextDay, movePrevDay, moveNextWeek, movePrevWeek, getBaseDate } from '../../../calendar/utils/navigation';
@@ -162,10 +161,7 @@ export const Grids = ({
     }
   };
 
-  const focusVisible = useFocusVisible();
-
-  const isRangeVisible =
-    (selectedStartDate && selectedEndDate) || (gridHasFocus && !!focusVisible['data-awsui-focus-visible']);
+  const isRangeVisible = (selectedStartDate && selectedEndDate) || gridHasFocus;
 
   const rangeEnds: Date[] = [selectedStartDate ?? focusedDate, selectedEndDate ?? focusedDate].filter(hasValue);
 
--- a/src/internal/components/chart-plot/focus-outline.tsx
+++ b/src/internal/components/chart-plot/focus-outline.tsx
@@ -1,10 +1,10 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
-import React, { useEffect, useRef } from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 
 import styles from './styles.css.js';
-import useFocusVisible from '../../hooks/focus-visible/index';
 import { Offset } from '../interfaces';
+import { isModifierKey } from '../../hooks/focus-visible';
 
 export interface FocusOutlineProps {
   elementKey?: null | string | number | boolean;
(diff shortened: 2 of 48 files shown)


src/date-range-picker/calendar/grids/index.tsx
```diff
@@ -162,10 +161,7 @@ export const Grids = ({
     }
   };
 
-  const focusVisible = useFocusVisible();
-
-  const isRangeVisible =
-    (selectedStartDate && selectedEndDate) || (gridHasFocus && !!focusVisible['data-awsui-focus-visible']);
+  const isRangeVisible = (selectedStartDate && selectedEndDate) || gridHasFocus;
```
just-boris, line 164: This `focusVisible['data-awsui-focus-visible']` condition does not seem to change anything. Tested the component, the outline displayed as expected

src/internal/components/chart-plot/focus-outline.tsx
```diff
@@ -1,20 +1,46 @@
...
 import { Offset } from '../interfaces';
+import { isModifierKey } from '../../hooks/focus-visible';
 
 export interface FocusOutlineProps {
   elementKey?: null | string | number | boolean;
   elementRef?: React.RefObject<SVGSVGElement | SVGGElement>;
   offset?: Offset;
 }
 
+function useFocusVisibleState() {
```
just-boris, line 15: It is recommended to use css-only implementation for focus-visible (to allow migrate to native `:focus-visible` once our supported browsers allow that).

This is why I put this hook here to prevent reuses of this exceptional pattern

src/internal/hooks/use-base-component/index.ts
```diff
@@ -16,6 +17,7 @@ export interface InternalBaseComponentProps {
  */
 export default function useBaseComponent<T = any>(componentName: string) {
   useTelemetry(componentName);
+  useFocusVisible();
```
just-boris, line 20: I think we can safely assume that all our components use focus visible in some way, so let's register it once for all components instead of repeating

src/side-navigation/internal.tsx
```diff
@@ -243,10 +238,7 @@ function Link({ definition, expanded, activeHref, fireFollow }: LinkProps) {
 
   return (
     <>
-      {/* https://github.com/yannickcr/eslint-plugin-react/issues/2962 */}
-      {/* eslint-disable-next-line react/jsx-no-target-blank */}
       <a
-        {...focusVisible}
```
just-boris, line 249: The upstream issue is resolved, no need to ignore

src/internal/hooks/use-base-component/index.ts
pan-kot: The Box doesn't. But I agree with the solution nevertheless.

src/internal/hooks/focus-visible/__tests__/index.test.tsx
```diff
@@ -24,65 +20,64 @@ test('should disable focus by default', async () => {
...
+  render(<Fixture />);
   fireEvent.keyDown(document.body);
   fireEvent.mouseDown(document.body);
-  expect(await findByTestId('button')).not.toHaveAttribute('data-awsui-focus-visible');
+  expect(document.body).not.toHaveAttribute('data-awsui-focus-visible');
 });
 
-test('should add listeners only once', async () => {
-  const { findByTestId } = render(
+test('should work with multiple components', () => {
```
pan-kot, line 49: Do we need this test case at all?

src/internal/components/chart-plot/focus-outline.tsx
```diff
@@ -1,20 +1,46 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
-import React, { useEffect, useRef } from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 
 import styles from './styles.css.js';
-import useFocusVisible from '../../hooks/focus-visible/index';
 import { Offset } from '../interfaces';
+import { isModifierKey } from '../../hooks/focus-visible';
```
pan-kot, line 7: Would it make sense to move the util to `internal/keycode`?

src/internal/components/chart-plot/focus-outline.tsx
pan-kot: I think this part can be refactored the same way as other components. I can do it myself as a follow-up.

src/table/resizer/styles.scss
```diff
@@ -50,7 +50,8 @@ $active-separator-width: 2px;
       }
     }
   }
-  &.has-focus {
+  // stylelint-disable-next-line selector-combinator-disallowed-list
+  [data-awsui-focus-visible='true'] &.has-focus {
```
pan-kot, line 54: Should this check for `body[data-awsui-focus-visible='true']` ?

pan-kot (dismissed):

src/internal/hooks/focus-visible/__tests__/index.test.tsx
just-boris: This test is very similar to the one below, "should add listeners only once", but it has different assertions. Let's keep both, it does not cost us anything

src/internal/components/chart-plot/focus-outline.tsx
just-boris: That is a very strange looking file (some copy of test-utils code for non-test context). Let's not move it for now, can change later if it becomes used anywhere outside focus-visible context

src/table/resizer/styles.scss
just-boris: Correct. Fixed in the next revision

pan-kot (approved):

johannes-weber (approved):

PR #925, opened 2023-04-04
chore: AppLayout support for Beta testing drawers in Visual Refresh

DaemonCahill:
### Description

This PR adds the ability for users to create custom drawers in the `AppLayout` component. There are no changes to the interface in order to prevent widespread awareness and adoption during the feature testing window.

### How has this been tested?

<details>
   <summary>Review checklist</summary>

_The following items are to be evaluated by the author(s) and the reviewer(s)._

#### Correctness

- _Changes include appropriate documentation updates._
- _Changes are backward-compatible if not indicated, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#public-apis)._
- _Changes do not include unsupported browser features, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#browsers-support)._
- _Changes were manually tested for accessibility, see [accessibility guidelines](https://cloudscape.design/foundation/core-principles/accessibility/)._

#### Security

- _If the code handles URLs: all URLs are validated through [the `checkSafeUrl` function](https://github.com/cloudscape-design/components/blob/main/src/internal/utils/check-safe-url.ts)._

#### Testing

- _Changes are covered with new/existing unit tests?_
- _Changes are covered with new/existing integration tests?_
</details>

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #938, opened 2023-04-07
chore: Adds Drawers to AppLayout (Classic)

katiegeorge:
### Description

This adds a private beta implementation of `drawers` to the AppLayout. Drawers allows for additional triggers on the right-hand side panel below Tools, and allows those panels to be independently resizable. 

See the [API Proposal here](https://quip-amazon.com/6gc4AqIFbmd5#ZUd9AA5sZbg).

Also see the [Visual Refresh PR](https://github.com/cloudscape-design/components/pull/925) for this feature.

image

### How has this been tested?

Screenshot tests are passing. Our plan is to add unit tests as an immediate follow-on PR so we can add them to Visual Refresh and Classic simultaneously.

See [Bug Bash doc](https://quip-amazon.com/q42lAwE7edzU/AppLayout-drawers-Beta-Bug-Bash).

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/app-layout/drawer/index.tsx
```diff
@@ -59,12 +47,75 @@ export function Drawer({
...
+  refs,
 }: DesktopDrawerProps) {
+  const openButtonWrapperRef = useRef<HTMLElement | null>(null);
   const { TagName, iconName, getLabels } = togglesConfig[type];
   const { mainLabel, closeLabel, openLabel } = getLabels(ariaLabels);
   const drawerContentWidthOpen = isMobile ? undefined : width;
   const drawerContentWidth = isOpen ? drawerContentWidthOpen : undefined;
-  const openButtonWrapperRef = useRef<HTMLElement | null>(null);
+
+  const activeDrawer = drawers?.items.find(item => item.id === drawers.activeDrawerId);
```
just-boris, line 62: This component represents a single drawer, so finding the specific one should happen on a level above (app layout main component)

Currently this find logic runs from 1 to 4 times, depending on how many drawers are used (including the static navigation/tools ones)

src/app-layout/drawer/index.tsx
```diff
@@ -117,16 +168,56 @@ export function Drawer({
...
+      className={clsx(styles.drawer, styles['drawer-closed'], testutilStyles['drawer-closed'], {
+        [styles['drawer-mobile']]: isMobile,
+      })}
+    >
+      <div
+        style={{ top: topOffset, bottom: bottomOffset }}
+        className={clsx(styles['drawer-content'], styles['non-interactive'], contentClassName)}
+      >
+        {!isMobile && (
+          <aside aria-label="Drawers" className={styles['drawer-triggers']}>
```
just-boris, line 202: Hard-coded text. Need to expose a property for that (or have a reminder to not forget before final release)

src/app-layout/index.tsx
```diff
@@ -141,12 +147,49 @@ const OldAppLayout = React.forwardRef(
...
+    );
+
+    const toolsItem = {
+      id: 'tools',
+      content: tools,
+      ariaLabels: {
+        triggerButton: 'View tools',
+        closeButton: 'Close tools',
+        content: 'Tools',
+      },
```
just-boris, line 168: Does this text suppress the value coming from `ariaLabels` prop?

src/app-layout/index.tsx
```diff
@@ -218,10 +276,13 @@ const OldAppLayout = React.forwardRef(
...
     const effectiveToolsWidth =
-      toolsHide && (!splitPanelDisplayed || splitPanelPreferences?.position !== 'side')
+      toolsHide && (!splitPanelDisplayed || splitPanelPreferences?.position !== 'side') && !drawers
         ? 0
-        : toolsOpen
+        : selectedDrawer?.resizable
+        ? drawerSize
+        : toolsOpen || activeDrawerId
         ? toolsWidth
         : closedDrawerWidth;
```
just-boris, line 287: Having that many members in this condition exceeds my ability to understand the code. Let's extract this into a function, at least (and possibly simplify some branches)

src/app-layout/index.tsx
```diff
@@ -261,6 +326,14 @@ const OldAppLayout = React.forwardRef(
       [setSplitPanelSize, onSplitPanelResize]
     );
 
+    const onDrawerSizeSet = useCallback(
```
just-boris, line 329: Some legacy code in this file uses `useCallback`, but I would not add it to any new code we write

pages/app-layout/with-drawers.page.tsx
```diff
@@ -0,0 +1,151 @@
...
+                triggerButton: 'Security trigger button',
+                resizeHandle: 'Security resize handle',
+              },
+              content: <Security />,
+              id: 'security',
+              resizable: true,
+              size: drawersWidth,
+              onResize: (event: NonCancelableCustomEvent<{ size: number }>) => {
+                setDrawersWidth(event.detail.size);
+              },
```
just-boris, line 44: It probably should have been called during API review, but we need to move `onResize` on the top level, along with `onChange`.

This will allow to make `items` array fully static

src/app-layout/index.tsx
```diff
@@ -364,7 +463,8 @@ const OldAppLayout = React.forwardRef(
       toolsPadding:
         // tools padding is displayed in one of the three cases
         // 1. Nothing on the that screen edge (no tools panel and no split panel)
-        (toolsHide && (!splitPanelDisplayed || finalSplitPanePosition !== 'side')) ||
+        toolsHide ||
+        (drawers && !activeDrawerId && (!splitPanelDisplayed || finalSplitPanePosition !== 'side')) ||
```
just-boris, line 467: It looks like now there is a duality between the old `tools` and the new `drawers` API. 

I think, we should resolve this duality in one function, like `applyToolsStateToDrawers(drawers, {toolsHide, toolsOpen, tools})` and then use a single `rightDrawer` (or any better name) abstraction instead of repeat `tools || drawers` many times

src/app-layout/mobile-toolbar/index.tsx
```diff
@@ -106,6 +110,25 @@ export function MobileToolbar({
           onClick={onToolsOpen}
         />
       )}
+      {drawers && (
+        <aside aria-label="Drawers" className={clsx(styles['mobile-toggle'], styles['mobile-toggle-with-drawers'])}>
```
just-boris, line 114: 1. Another hard-coded label 
2. Can we avoid this wrapper all together and have a flat layout?

```html
<div className={clsx(styles['mobile-bar']}>
   <MobileToggle/> //navigation
   <div className={styles['mobile-bar-breadcrumbs']}></div>
   {drawers.map(toggleConfig = > <MobileToggle {...toggleConfig} />}
</div>
```

The `drawers` array need to be normalised to include tools in a single place, according to the comment above (TODO link)

src/app-layout/utils/use-drawer-focus-control.ts
```diff
@@ -0,0 +1,41 @@
...
+export interface DrawerFocusControlRefs {
+  toggle: RefObject<ButtonProps.Ref>;
+  slider: RefObject<HTMLDivElement>;
+}
+interface DrawerFocusControlState {
+  refs: DrawerFocusControlRefs;
+  setLastInteraction: (interaction: DrawerLastInteraction) => void;
+}
+
+export function useDrawerFocusControl(dependencies: DependencyList): DrawerFocusControlState {
```
just-boris, line 17: The long term plan was to consolidate focus control in a single controller. 

So instead of creating a new controller for the drawers, we need to refactor existing ones to support an array of possible focusables

pages/app-layout/with-drawers.page.tsx
```diff
@@ -0,0 +1,151 @@
...
+                  </Toggle>
+                </SpaceBetween>
+              </SpaceBetween>
+            }
+          >
+            <Containers />
+          </ContentLayout>
+        }
+        contentType="form"
+        maxContentWidth={999999999999999}
```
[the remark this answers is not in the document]
katiegeorge: Makes sense.

pages/app-layout/with-drawers.page.tsx
```diff
@@ -0,0 +1,151 @@
...
+    <svg viewBox="0 0 16 16" fill="none">
+      <path
+        d="M8 15H1V5h14v3m-6.5 4.25H16M12.25 8.5V16M5 1h6v4H5V1Z"
+        stroke="#fff"
+        strokeWidth="2"
+        strokeLinejoin="round"
+      />
+    </svg>
+  );
+}
```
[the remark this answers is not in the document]
katiegeorge: I don't have a super strong opinion here, but shouldn't our testing pages test as many versions of the component as possible? It could be possible there are small alignment or other issues. I can remove for now.

pages/app-layout/with-drawers.page.tsx
```diff
@@ -0,0 +1,151 @@
...
+  Toggle,
+} from '~components';
+import appLayoutLabels from './utils/labels';
+import { Breadcrumbs, Containers } from './utils/content-blocks';
+import ScreenshotArea from '../utils/screenshot-area';
+
+export default function WithDrawers() {
+  const [activeDrawerId, setActiveDrawerId] = useState<string | null>(null);
+  const [useDrawers, setUseDrawers] = useState(true);
+  const [useTools, setUseTools] = useState(true);
```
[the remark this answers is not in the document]
katiegeorge: Makes sense.

src/app-layout/drawer/index.tsx
```diff
@@ -59,12 +47,75 @@ export function Drawer({
...
+    const maxWidth = getMaxWidth();
+    const size = getLimitedValue(MIN_WIDTH, width, maxWidth);
+
+    if (isOpen && maxWidth >= MIN_WIDTH) {
+      onResize({ size });
+    }
+  };
+
+  const position = 'side';
+  const setBottomPanelHeight = () => {};
```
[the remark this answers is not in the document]
katiegeorge: Yeah I should have left a comment here. I am currently importing directly from the Split Panel to avoid copy/ pasting code, and didn't want to touch any of that code to avoid unexpected regressions. I think ideally, we extract the resize functions into a neutral folder, and have Split Panel and Drawer both use them. For now, I was going to leave like this. The other option is to copy the functions into the drawer code and refactor from there.

src/app-layout/drawer/index.tsx
katiegeorge: I can take a look, but may need some help moving all of this logic one level up. It's pretty entrenched right now.

src/app-layout/drawer/index.tsx
katiegeorge: Sure, I can copy what we do for `Tools`. I was under the impression we only did that because it could either be Tools or Navigation, and here it will always be Drawers.

src/app-layout/index.tsx
katiegeorge: Only for the Tools panel, since this won't be something that is user-controlled. When tools and drawers are both used, I create a tools item to append to the front of the drawers array.

src/app-layout/index.tsx
katiegeorge: Makes sense.

src/app-layout/index.tsx
katiegeorge: Is there an alternative? I was copying what Split Panel does just above this.

pages/app-layout/with-drawers.page.tsx
katiegeorge: The idea here is that each panel / drawer can separately be resizable, and have separate sizes if the user wants. Is that still possible if `onResize` moves up a level?

src/app-layout/index.tsx
katiegeorge: I'm a bit confused by this - I agree with the sentiment but let's discuss implementation more.

pages/app-layout/with-drawers.page.tsx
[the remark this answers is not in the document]
just-boris: In other testing pages we use more abstract shapes for custom icons, for example in token group: https://github.com/cloudscape-design/components/blob/7ed751a10e2ecf63cec865877bf8a6f75b6f1e54/pages/token-group/permutations.page.tsx#L152-L174

I think, it is important to not accidentally create an unofficial icon set and this is why generic shapes are recommended

src/app-layout/mobile-toolbar/index.tsx
katiegeorge: Makes sense.

src/app-layout/drawer/index.tsx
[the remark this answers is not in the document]
just-boris: ok, let's leave it as is for now (except extracting into `ResizableDrawer`)

src/app-layout/index.tsx
just-boris: If we support auto-adoption of the `tools` properties into `drawers` array it should adopt the values from `ariaLabels` too.

src/app-layout/index.tsx
just-boris: Just a plain function: 

```
const onDrawerSizeSet = ({detail}) =>  {}
```

pages/app-layout/with-drawers.page.tsx
just-boris: Yes, if event detail will also include drawer id. Then it will be closer to how resizable columns API is implemented in table

src/app-layout/index.tsx
katiegeorge: Note from live conversations: use the existing `ariaLabels` prop to pull these in, instead of hard-coding them.

src/app-layout/index.tsx
katiegeorge: Notes from live discussion: to-do in a follow-up.

src/app-layout/utils/use-drawer-focus-control.ts
katiegeorge: Note from live discussion: I had to add this (separate from the VR version) because of the focus state on the resize handle. We will consolidate after adding resize to VR drawers.

src/app-layout/mobile-toolbar/index.tsx
katiegeorge: Note: I updated the hard-coded label, but keeping the structure as before. To change in a follow-up.

src/app-layout/drawer/index.tsx
```diff
@@ -42,90 +24,143 @@ export interface DesktopDrawerProps {
 // * https://github.com/nvaccess/nvda/issues/5825
 // * https://github.com/nvaccess/nvda/issues/5247
 // * https://github.com/nvaccess/nvda/pull/8869 (reverted PR that was going to fix it)
+export const Drawer = React.forwardRef(
```
just-boris, line 27: Even internal components need to have more or less isolated API to easier and safer refactorings in the future.
Suggested API for this component

```js
<Drawer
  ...old props unchanged
  
   ariaLabels={{mainLabel, closeLabel, openLabel}} // this property changes shape
   resizeHandle={...} // new slot
/>
```

Usage examples


```js
// tools (non-resizable) drawer
<Drawer
   ariaLabels={{mainLabel: ariaLabels.tools, openLabel: ariaLabels.toolsToggle, closeLabel: ariaLabels.toolsClose }}
/>

// new (resizable) drawer
<Drawer
   ariaLabels={{mainLabel: ariaLabels.tools, openLabel: ariaLabels.toolsToggle, closeLabel: ariaLabels.toolsClose }}
/>
```

src/app-layout/drawer/index.tsx
```diff
@@ -42,90 +24,143 @@ export interface DesktopDrawerProps {
...
-  onLoseFocus,
-}: DesktopDrawerProps) {
-  const { TagName, iconName, getLabels } = togglesConfig[type];
-  const { mainLabel, closeLabel, openLabel } = getLabels(ariaLabels);
-  const drawerContentWidthOpen = isMobile ? undefined : width;
-  const drawerContentWidth = isOpen ? drawerContentWidthOpen : undefined;
-  const openButtonWrapperRef = useRef<HTMLElement | null>(null);
+    const getDrawersLabels = (labels: DrawerItemAriaLabels = {}) => ({
+      drawerMainLabel: labels?.content,
+      drawerOpenLabel: labels?.triggerButton,
```
just-boris, line 59: This value is created but never used (also see the comment above how to eliminate `getDrawersLabels` completely)

src/app-layout/drawer/index.tsx
```diff
@@ -42,90 +24,143 @@ export interface DesktopDrawerProps {
...
+            />
+            {children}
+          </TagName>
+        </div>
+      </div>
+    );
+  }
+);
+
+export function DrawerTriggersBar({ isMobile, topOffset, bottomOffset, drawers, contentClassName }: DrawerTriggersBar) {
```
just-boris, line 134: `DrawerTriggersBar` is used twice as a function name and as props type. I am very surprised it did not cause any problems because the function is exported but the type is not. 

To avoid all such problems we always call prop types with `Props` ending,  `DrawerTriggersBarProps`

src/app-layout/drawer/resizable-drawer.tsx
```diff
@@ -0,0 +1,111 @@
...
+  getMaxWidth,
+  refs,
+  activeDrawer,
+  ...props
+}: ResizableDrawerProps) => {
+  const { isOpen, children, isMobile } = props;
+
+  const MIN_WIDTH = activeDrawer?.size && activeDrawer.size < 280 ? activeDrawer?.size : 280;
+  const [relativeSize, setRelativeSize] = useState(0);
+  const focusVisible = useFocusVisible();
```
just-boris, line 44: I merged #934 which changes how `focusVisible` is used.

You need to remove this code all together (see similar examples in my PR). It now works out of the box

PR #945, opened 2023-04-11
fix: Add role=group to token-group and property-filter tokens

pan-kot:
### Description

Although technically it works, having aria-label, aria-description or aria-disabled on li elements is not recommended and has been reported by html-validator. Instead it is possible to use role=group for the tokens that are directly nested under li  elements.

### How has this been tested?

Tested manually with VO and NVDA.

<details>

#### Correctness

#### Security

#### Testing

</details>

---


timogasda (approved):
Code looks good to me. I would change the title of the PR to make it more suitable for the changelog. Keep in mind that the prefix (fix:) will be removed.

PR #963, opened 2023-04-14
fix: Allow to skip optional i18n strings on split panel

just-boris:
### Description

When `hidePreferences` button is used, there is no point in passing strings for the modal, because it will never show. Additionally, when `closeBehavior` is `hide`, open button is also not displayed.

### How has this been tested?

Updated the demo where we use this property to only include required strings

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/app-layout/utils/strings.ts
+++ b/pages/app-layout/utils/strings.ts
@@ -2,7 +2,14 @@
 // SPDX-License-Identifier: Apache-2.0
 import { SplitPanelProps } from '~components/split-panel';
 
+export const discreetSplitPanelI18nStrings: SplitPanelProps.I18nStrings = {
+  closeButtonAriaLabel: 'Close panel',
+  resizeHandleAriaLabel: 'Slider',
+};
+
 export const splitPaneli18nStrings: SplitPanelProps.I18nStrings = {
+  ...discreetSplitPanelI18nStrings,
+  openButtonAriaLabel: 'Open panel',
   preferencesTitle: 'Preferences',
   preferencesPositionLabel: 'Split panel position',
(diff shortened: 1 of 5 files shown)


src/split-panel/preferences-modal.tsx
```diff
@@ -16,13 +16,13 @@ import bottomPositionIconRefresh from './icons/bottom-icon-refresh';
...
-  positionBottom: string;
-  positionSide: string;
+  header?: string;
+  cancel?: string;
+  confirm?: string;
+  positionLabel?: string;
+  positionDescription?: string;
+  positionBottom?: string;
+  positionSide?: string;
 }
```
just-boris, line 26: I had two choices here

1. Add non-null assertion in the `split-panel/index.tsx` file
2. Mark them optional here

I picked the 2nd option because it is closer to runtime behavior (these properties may be missing)

pages/app-layout/utils/strings.ts
```diff
@@ -2,15 +2,19 @@
 // SPDX-License-Identifier: Apache-2.0
 import { SplitPanelProps } from '~components/split-panel';
 
+export const discreetSplitPanelI18nStrings: SplitPanelProps.I18nStrings = {
```
just-boris, line 5: This is how we call this "no preferences" mode in our design guidelines

src/split-panel/preferences-modal.tsx
rubencarvalho: I'm happy with the 2nd option.

rubencarvalho (approved):

PR #964, opened 2023-04-14
chore: Adds tests to Drawers

katiegeorge:
### Description

This PR adds unit and integration tests to the new beta feature `Drawers`.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/app-layout/__tests__/drawers.test.tsx
```diff
@@ -0,0 +1,68 @@
...
+      drawers: {
+        activeDrawerId: 'security',
+        onResize: (event: any) => onResize(event.detail),
+        items: drawersConfigurations.drawersResizableItems,
+      },
+    };
+    const { wrapper } = renderComponent(<AppLayout contentType="form" {...drawersOpen} />);
+    act(() => wrapper.findDrawersSlider().keydown(KeyCode.left));
+    expect(getComputedStyle(wrapper.findActiveDrawer().getElement()).width).toBe('290px');
+  });
```
connorlanigan, line 55: I'm not sure if I'm understanding this test correctly, but it looks like we're not actually controlling the size here since I don't see the `width` property (?)

I would recommend to also test that `onResize` has been called

src/app-layout/__tests__/main.test.tsx
```diff
@@ -115,6 +115,51 @@ describe.each([
   });
 });
 
+describe.each([
```
connorlanigan, line 118: We're passing a single array item to `describe.each(` here, so I think we can replace it with `describe(` and move the `find*` functions to standalone `const`s.

src/app-layout/__tests__/main.test.tsx
```diff
@@ -115,6 +115,51 @@ describe.each([
...
+        ],
+      },
+    };
+
+    const { wrapper, rerender } = renderComponent(<AppLayout contentType="form" {...drawers} />);
+    findToggle(wrapper).click();
+    expect(onChange).toHaveBeenCalledWith(expect.objectContaining({ detail: 'security' }));
+    expect(findElement(wrapper)).not.toBeNull();
+
+    rerender(<AppLayout contentType="form" {...drawers} />);
```
connorlanigan, line 156: This line rerenders with the same props as before, I think it can be removed

src/app-layout/__tests__/main.test.tsx
```diff
@@ -115,6 +115,51 @@ describe.each([
...
+            trigger: {
+              iconName: 'security',
+            },
+          },
+        ],
+      },
+    };
+
+    const { wrapper, rerender } = renderComponent(<AppLayout contentType="form" {...drawers} />);
+    findToggle(wrapper).click();
```
connorlanigan, line 152: I recommend to wrap every interaction in `act()`, to ensure side effects from `useEffect` hooks are executed as we expect them to

src/app-layout/__tests__/main.test.tsx
```diff
@@ -115,6 +115,51 @@ describe.each([
...
+  [
+    'drawers',
+    {
+      findElement: (wrapper: AppLayoutWrapper) => wrapper.findActiveDrawer(),
+      findToggle: (wrapper: AppLayoutWrapper) => wrapper.findDrawersTriggers()[0],
+      findClose: (wrapper: AppLayoutWrapper) => wrapper.findActiveDrawerCloseButton(),
+    },
+  ],
+] as const)('%s', (name, { findToggle, findElement, findClose }) => {
+  test('property is controlled', () => {
```
connorlanigan, line 128: Which property are we testing here? Should we set a value for `activeDrawerId`?

src/app-layout/__integ__/app-layout-drawers.test.ts
```diff
@@ -0,0 +1,102 @@
...
+  getDrawerSize() {
+    return this.getBoundingBox(wrapper.findActiveDrawer().toSelector());
+  }
+}
+
+function setupTest(testFn: (page: AppLayoutDrawersPage) => Promise<void>, url = '#/light/app-layout/with-drawers') {
+  return useBrowser(async browser => {
+    const page = new AppLayoutDrawersPage(browser);
+    await page.setWindowSize(viewports.desktop);
+    await browser.url(`${url}?visualRefresh=false`);
```
connorlanigan, line 47: Can we test VR as well?

connorlanigan (changes requested):

src/app-layout/__tests__/drawers.test.tsx
katiegeorge: So without the `size` property, it will default to `280`. I wasn't sure how to test with adding a specific size, since if we explicitly set the size to a constant, `onResize` won't work. Is there a way in a testing environment to set the `size` property to something controlled by `onResize`?

connorlanigan: In a testing environment we don't have a state outside of the component, instead we have to replicate the updated state by rerendering the component with the new properties that we want to be passed in.

For example:
```js
const { wrapper, rerender } = renderComponent(
     <AppLayout contentType="form" {...drawersOpen} size={280} />
);
act(() => wrapper.findDrawersSlider().keydown(KeyCode.left));
expect(onResize).toHaveBeenCalledTimes(1);

// To check that the component is not using an uncontrolled state
expect(getComputedStyle(wrapper.findActiveDrawer().getElement()).width).toBe(280);


// Simulate that we have updated the controlled state to 320px
rerender(
     <AppLayout contentType="form" {...drawersOpen} size={320} />
);
expect(getComputedStyle(wrapper.findActiveDrawer().getElement()).width).toBe(320);

```

src/app-layout/__tests__/main.test.tsx
katiegeorge: None of the other tests in this file are wrapped in `act()` - should I still do it just for this one?

src/app-layout/__tests__/main.test.tsx
katiegeorge: I updated the test to include the `activeDrawerId` and `rerender`. Let me know if it looks okay.

connorlanigan: Looks good!

src/app-layout/index.tsx
```diff
@@ -266,19 +265,33 @@ const OldAppLayout = React.forwardRef(
...
+    const getSizes = () => {
+      const sizes: { [id: string]: number } = {};
+      if (!drawers) {
+        return {};
+      }
 
-    const [drawerSize = defaultDrawerSize, setDrawerSize] = useControllable(
-      selectedDrawer?.size,
+      for (const item of drawers.items) {
+        if (item.resizable && item.size) {
```
connorlanigan, line 275: We'll also need to include drawers here that are not `resizable`, because they can still have a size that's different from the default. In the current state the non-resizable drawers always render at a width of `toolsWidth` pixels.

```suggestion
        if (item.size) {
```

src/app-layout/drawer/interfaces.ts
```diff
@@ -83,3 +85,13 @@ export interface SizeControlProps {
   setSidePanelWidth: (width: number) => void;
   setBottomPanelHeight: (height: number) => void;
 }
+
+export interface InternalDrawerProps {
+  drawers: {
```
connorlanigan, line 90: This field should be optional, to reflect that the prop will not always be set. Typescript will then show some additional type errors in `app-layout/index.ts`

```suggestion
  drawers?: {
```

src/app-layout/index.tsx
```diff
@@ -266,19 +265,33 @@ const OldAppLayout = React.forwardRef(
...
+          sizes[item.id] = item.size;
+        }
+      }
+      return sizes;
+    };
+
+    const [drawerSizes = {}, setDrawerSizes] = useControllable(
+      Object.keys(getSizes()).length > 0 ? getSizes() : undefined,
```
connorlanigan, line 283: non-blocking: Since we call the `getSizes` function twice (and always), we can do a tiny performance improvement by calling the function immediately when it's defined:

```js
const sizes = (() => {
      const sizes: { [id: string]: number } = {};
      if (!drawers) {
        return {};
      }

      // etc ...

      return sizes;
})();

File: src/app-layout/index.tsx [hunk 10]
```diff
@@ -266,19 +265,33 @@ const OldAppLayout = React.forwardRef(
...
       drawers?.onResize,
-      defaultDrawerSize,
+      {},
       {
         componentName: 'AppLayout',
-        controlledProp: 'drawerSize',
-        changeHandler: 'onDrawerSize',
+        controlledProp: 'size',
+        changeHandler: 'onResize',
```
connorlanigan, line 289: I'd recommend to extend this text to reflect that the properties are inside another property. The strings here are only used for a console.log call, i.e. for human consumption. They do not need to match any specific value from Javascript's perspective, they just need to be useful for customers to understand where to find the properties. They are only used in these two messages:

https://github.com/cloudscape-design/components/blob/6f85b644a2990f0b0fd0395ce384f9a8e872158b/src/internal/hooks/use-controllable/index.ts#L65-L68

https://github.com/cloudscape-design/components/blob/6f85b644a2990f0b0fd0395ce384f9a8e872158b/src/internal/hooks/use-controllable/index.ts#L78-L84

For example something like 
> You provided a `drawers.items[n].size` prop without an `drawers.onResize` handler. This will render a non-interactive component.

by setting
```js
controlledProp: 'drawers.items[n].size',
changeHandler: 'drawers.onResize',
```

connorlanigan (changes requested):

connorlanigan (approved):

PR #969, opened 2023-04-17
fix: Update keys calculation for virtual dropdown lists

just-boris:
### Description

Add this property according to [react-virtual docs](https://react-virtual-v2.tanstack.com/docs/api#options)

> keyExtractor: Function(index) => String | Integer
This function should be passed whenever dynamic measurement rendering is enabled and the size or order of items in the list changes.

We have variable size of the items, so it is our case.

Related links, issue #, if available: AWSUI-20777

### How has this been tested?

* Pushed to my dev pipeline for cross-browser tests
* Added unit tests on virtual scroll mode in general, because we missed unit test coverage on this feature completely

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/autosuggest/__tests__/virtual-scroll.test.tsx
+++ b/src/autosuggest/__tests__/virtual-scroll.test.tsx
@@ -1,0 +1,66 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import React from 'react';
+import { render } from '@testing-library/react';
+import Autosuggest, { AutosuggestProps } from '../../../lib/components/autosuggest';
+import createWrapper from '../../../lib/components/test-utils/dom';
+
+const defaultProps: AutosuggestProps = {
+  options: [
+    { value: '1', label: 'One' },
+    { value: '2', lang: 'Two' },
+    { value: '3', lang: 'Three' },
+  ],
+  virtualScroll: true,
+  enteredTextLabel: () => 'Use value',
+  value: '',
+  onChange: () => {},
+};
+
+function renderWithWrapper(ui: React.ReactElement) {
+  const { container } = render(ui);
+  return createWrapper(container).findAutosuggest()!;
+}
+
+describe('Virtual scroll support', () => {
+  test('should render plain virtual list', () => {
+    const wrapper = renderWithWrapper(<Autosuggest {...defaultProps} />);
+    wrapper.findNativeInput().focus();
+    expect(wrapper.findDropdown().findOptions()).toHaveLength(3);
+  });
+
+  test('should render a subset of items in virtual list', () => {
+    const wrapper = renderWithWrapper(
+      <Autosuggest
+        {...defaultProps}
+        options={Array.from({ length: 100 }, (_, index) => ({ value: `Option ${index}` }))}
+      />
+    );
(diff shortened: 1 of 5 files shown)


src/autosuggest/__tests__/virtual-scroll.test.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+  test('should render a subset of items in virtual list', () => {
+    const wrapper = renderWithWrapper(
+      <Autosuggest
+        {...defaultProps}
+        options={Array.from({ length: 100 }, (_, index) => ({ value: `Option ${index}` }))}
+      />
+    );
+    wrapper.findNativeInput().focus();
+    expect(wrapper.findDropdown().findOptions()).toHaveLength(6);
```
gethinwebster, line 40: how certain are we that this will remain `6`? isn't it somewhat dependent on a lot of things (3rd party implementation, item spacing, ...) that might change?

src/select/__tests__/virtual-scroll.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+    wrapper.openDropdown();
+    expect(wrapper.findDropdown().findOptions()).toHaveLength(3);
+  });
+
+  test('should render a subset of items in virtual list', () => {
+    const wrapper = renderWithWrapper(
+      <Select {...defaultProps} options={Array.from({ length: 100 }, (_, index) => ({ value: `Option ${index}` }))} />
+    );
+    wrapper.openDropdown();
+    expect(wrapper.findDropdown().findOptions()).toHaveLength(11);
```
gethinwebster, line 36: (same q here)

src/autosuggest/__tests__/virtual-scroll.test.tsx
just-boris: This is defined by `overscan` configuration: https://github.com/cloudscape-design/components/blob/c95120a2acca52461e18575eda3aee4b34e2b937/src/autosuggest/virtual-list.tsx#L37

One current item + 5 overscan.

src/select/__tests__/virtual-scroll.test.tsx
just-boris: For some reason, the overscan count doubles, I am guessing because it works two way due to some implementation details

src/autosuggest/__tests__/virtual-scroll.test.tsx
just-boris: Item size does not matter here, because it is a JSDOM test. It may only change if we upgrade `react-virtual` dependency, where changing this test would be totally fine.

src/select/__tests__/virtual-scroll.test.tsx
gethinwebster: seems pretty strange and likely to cause future flakiness, but I'm happy if you're convinced it should be stable

gethinwebster (approved):

just-boris:
Cancelling the change, because `keyExtractor` returns `undefined` for all keys on the initial render, which defeats the reason for the change.

This happens because using `useStableEventHandler`, which is required because react-virtual heavily relies in memoization and the code fails with "Maximum Update Depth Exceeded" error if passing an inline lambda instead

PR #977, opened 2023-04-20
fix: Pass through props for select in code editor preferences

gethinwebster:
### Description

Ensure that all relevant Select props can be provided for the code editor theme picker 

### How has this been tested?

New tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/code-editor/preferences-modal.tsx
+++ b/src/code-editor/preferences-modal.tsx
@@ -23,6 +23,9 @@ interface PreferencesModali18nStrings {
   theme: string;
   lightThemes: string;
   darkThemes: string;
+  themeFilteringPlaceholder?: string;
+  themeFilteringAriaLabel?: string;
+  themeFilteringClearAriaLabel?: string;
 }
 
 interface PreferencesModalProps {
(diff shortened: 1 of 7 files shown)


Al-Dani (approved):

src/code-editor/preferences-modal.tsx
```diff
@@ -97,6 +100,9 @@ export default (props: PreferencesModalProps) => {
               onChange={onThemeSelected}
               options={themeOptions}
               filteringType="auto"
```
just-boris, line 102: Did we consider removing filtering mode instead (to make the default setup automatically compliant)?

Since [our recommendation](https://cloudscape.design/components/code-editor/?tabId=api) is to limit allowed themes using `themes` prop, this list is unlikely to grow large

gethinwebster: it feels like this would probably be a breaking change. At the very least I'd want to give customers the option to re-enable filtering, in which case the properties would be needed anyway, but I could see an argument for adding that but turning it off by default, to improve the default accessibility via a slightly-breaking change (with re-opt-in option)

just-boris: Your call. But I feel we are missing a good simplification opportunity

PR #983, opened 2023-04-20
fix: Allow placing semantic block elements inside card definition header

avinashbot:
### Description

Users can't place a div inside the card header because the wrapper is a span. Normally, it's designed to hold text (like a table), but it's being used in a situation with some layout elements. Given the scope of the change, we decided we might as well do it to unblock them.

Related links, issue #, if available: AWSUI-20694

### How has this been tested?

I changed an element that already has `display: inline-block` styling. No tests. Sue me.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/cards/index.tsx
+++ b/src/cards/index.tsx
@@ -225,9 +225,9 @@ const CardsList = <T,>({
         >
           <div className={styles['card-inner']}>
             <div className={styles['card-header']}>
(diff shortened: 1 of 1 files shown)


timogasda (approved):
I guess I won't sue this time.

PR #1016, opened 2023-04-25
chore: Add app layout runtime API

just-boris:
### Description

Adding new API for configuring app layout outside React tree. Detailed proposal: https://tiny.amazon.com/tjbskv2t/quipK3xL

### How has this been tested?

* Added unit tests
* Tested on integration POC (links in the proposal)

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/__tests__/warnings.test.tsx
+++ b/src/app-layout/__tests__/warnings.test.tsx
@@ -20,10 +20,10 @@ const noop = () => undefined;
  */
 describe('AppLayout component', () => {
   test('warns when toolsOpen and toolsHide are both set', () => {
-    render(<AppLayout toolsHide={true} toolsOpen={false} onToolsChange={noop} />);
+    const { rerender } = render(<AppLayout toolsHide={true} toolsOpen={false} onToolsChange={noop} />);
     expect(console.warn).not.toHaveBeenCalled();
 
-    render(<AppLayout toolsHide={true} toolsOpen={true} onToolsChange={noop} />);
+    rerender(<AppLayout toolsHide={true} toolsOpen={true} onToolsChange={noop} />);
 
     expect(console.warn).toHaveBeenCalledTimes(1);
     expect(console.warn).toHaveBeenCalledWith(
--- a/src/app-layout/index.tsx
+++ b/src/app-layout/index.tsx
@@ -47,6 +47,8 @@ import RefreshedAppLayout from './visual-refresh';
 import { useInternalI18n } from '../internal/i18n/context';
 import { useSplitPanelFocusControl } from './utils/use-split-panel-focus-control';
 import { useDrawerFocusControl } from './utils/use-drawer-focus-control';
+import { awsuiPluginsInternal } from '../internal/plugins/api';
+import { DrawersLayout, convertRuntimeDrawers } from './runtime-api';
 
 export { AppLayoutProps };
 
@@ -56,6 +58,7 @@ const AppLayout = React.forwardRef(
     ref: React.Ref<AppLayoutProps.Ref>
   ) => {
     const { __internalRootRef } = useBaseComponent<HTMLDivElement>('AppLayout');
+    const [runtimeDrawers, setRuntimeDrawers] = useState<DrawersLayout>({ before: [], after: [] });
     const isRefresh = useVisualRefresh();
 
     const i18n = useInternalI18n('app-layout');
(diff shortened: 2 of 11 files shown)


src/test-utils/dom/app-layout/index.ts
```diff
@@ -63,7 +63,7 @@ export default class AppLayoutWrapper extends ComponentWrapper {
     return this.findByClassName(testutilStyles['drawers-mobile-triggers-container']);
   }
 
-  findDrawersTriggers(): ElementWrapper<HTMLButtonElement>[] | null {
+  findDrawersTriggers(): ElementWrapper<HTMLButtonElement>[] {
```
just-boris, line 66: Not directly related to the change, but I found it when writing tests. `findAll` util can never return `null`, so the type can be simplified

src/app-layout/index.tsx
```diff
@@ -73,10 +76,27 @@ const AppLayout = React.forwardRef(
...
+    const disableRuntimeDrawers = (props as any).__disableRuntimeDrawers;
+    const combinedDrawers = [...runtimeDrawers.before, ...(ownDrawers?.items ?? []), ...runtimeDrawers.after];
+    const finalDrawers = combinedDrawers.length > 0 ? { ...ownDrawers, items: combinedDrawers } : ownDrawers;
+
+    useEffect(() => {
+      if (disableRuntimeDrawers) {
+        return;
+      }
+      return awsuiPluginsInternal.appLayout.registerAppLayout(drawers =>
+        setRuntimeDrawers(convertRuntimeDrawers(drawers))
```
gethinwebster, line 89: Thinking about re-renders in this sort of scenario:

- A custom drawer is open
- Something else on the page registers a new custom drawer
- Result: I think at this point the visible custom drawer would re-render? I'm not _sure_ that I would expect this as a drawer developer

wdyt?

just-boris: This line in particular does not cause the drawers content to be remounted. If keys management in JSX is done correctly, it will work fine.

I will keep an eye on this use-case, but after this PR

src/app-layout/index.tsx
```diff
@@ -73,10 +76,27 @@ const AppLayout = React.forwardRef(
     const props = { contentType, headerSelector, footerSelector, ...rest, ariaLabels };
 
     const baseProps = getBaseProps(rest);
+    const ownDrawers = (props as any).drawers;
+    const disableRuntimeDrawers = (props as any).__disableRuntimeDrawers;
+    const combinedDrawers = [...runtimeDrawers.before, ...(ownDrawers?.items ?? []), ...runtimeDrawers.after];
+    const finalDrawers = combinedDrawers.length > 0 ? { ...ownDrawers, items: combinedDrawers } : ownDrawers;
+
+    useEffect(() => {
+      if (disableRuntimeDrawers) {
```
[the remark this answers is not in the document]
just-boris: Yes, there could be a use-case like `<AppLayout disableRuntimeDrawers={pageHasIframes} />`. 

I will update the code and add an extra test

src/internal/plugins/internal.ts
```diff
@@ -0,0 +1,106 @@
...
+  [storageKey]: AwsuiApi;
+}
+
+function findUpApi(currentWindow: WindowWithApi): AwsuiApi | undefined {
+  try {
+    if (currentWindow[storageKey]) {
+      return currentWindow[storageKey];
+    }
+
+    if (!currentWindow || currentWindow.parent === currentWindow) {
```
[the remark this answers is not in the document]
just-boris: My bad. I incorrectly ported another cross-frame finder implementation:

https://github.com/cloudscape-design/component-toolkit/blob/a1f138c17d8104e27e56f40410152337483ad73b/src/internal/base-component/metrics/log-clients.ts#L50

Changed the line above  to `currentWindow?.[storageKey]` and now all lines make sense

src/internal/plugins/internal.ts
```diff
@@ -0,0 +1,106 @@
...
+          scheduleUpdate();
+        },
+      },
+    },
+    awsuiPluginsInternal: {
+      appLayout: {
+        clearRegisteredDrawers() {
+          drawers = [];
+        },
+        registerAppLayout(listener) {
```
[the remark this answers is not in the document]
just-boris: Updated, it is `onDrawersRegistered` now

src/app-layout/__tests__/warnings.test.tsx
```diff
@@ -20,10 +20,10 @@ const noop = () => undefined;
  */
 describe('AppLayout component', () => {
   test('warns when toolsOpen and toolsHide are both set', () => {
-    render(<AppLayout toolsHide={true} toolsOpen={false} onToolsChange={noop} />);
+    const { rerender } = render(<AppLayout toolsHide={true} toolsOpen={false} onToolsChange={noop} />);
     expect(console.warn).not.toHaveBeenCalled();
 
-    render(<AppLayout toolsHide={true} toolsOpen={true} onToolsChange={noop} />);
+    rerender(<AppLayout toolsHide={true} toolsOpen={true} onToolsChange={noop} />);
```
just-boris, line 26: This test used to render two copies of app layout, which now emits a warning. 

The warning is legitimate, the test was not, so I fixed the test

src/app-layout/index.tsx
```diff
@@ -73,10 +76,31 @@ const AppLayout = React.forwardRef(
     const props = { contentType, headerSelector, footerSelector, ...rest, ariaLabels };
 
     const baseProps = getBaseProps(rest);
+    const ownDrawers = (props as any).drawers;
+    const disableRuntimeDrawers = (props as any).__disableRuntimeDrawers;
```
gethinwebster, line 80: [minor] interested in the difference between `drawers` and `__disableRuntimeDrawers` - is one "more" private than the other?

gethinwebster (approved):

src/app-layout/index.tsx
just-boris: Yes. `drawers` has potential to become an official public API. `__disableRuntimeDrawers` is not, it will be removed after runtime phase of this API ends

PR #1026, opened 2023-04-27
feat: Add use-sticky-columns util

rubencarvalho:
### Description

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/table/use-sticky-columns.ts
+++ b/src/table/use-sticky-columns.ts
@@ -1,0 +1,335 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import React, { useCallback, useEffect, useMemo, useRef } from 'react';
+import AsyncStore, { useReaction } from '../area-chart/model/async-store';
+import { useStableEventHandler } from '../internal/hooks/use-stable-event-handler';
+import clsx from 'clsx';
+import { useResizeObserver } from '../internal/hooks/container-queries';
+
+export const selectionColumnId = Symbol('selection-column-id');
+
+// We allow the table to have a minimum of 148px of available space besides the sum of the widths of the sticky columns
+// This value is an UX recommendation and is approximately 1/3 of our smallest breakpoint (465px)
+const MINIMUM_SCROLLABLE_SPACE = 148;
+
+type ColumnId = string | symbol;
+
+interface StickyStateProps {
+  visibleColumns: readonly ColumnId[];
+  stickyColumnsFirst: number;
+  stickyColumnsLast: number;
+}
+
+export interface StickyStateModel {
+  store: StickyColumnsStore;
+  style: {
+    wrapper?: React.CSSProperties;
+  };
+  refs: {
+    table: React.RefCallback<HTMLElement>;
+    wrapper: React.RefCallback<HTMLElement>;
+    cell: (columnId: ColumnId, node: null | HTMLElement) => void;
+  };
+}
+
+export interface StickyState {
+  cellStyles: Record<ColumnId, null | StickyStateCellStyles>;
+  scrollPadding: { left: number; right: number };
+}
+
+interface UpdateScrollProps {
+  wrapper: HTMLElement;
+  table: HTMLElement;
+}
+
+export interface StickyStateCellStyles {
+  padLeft: boolean;
+  lastLeft: boolean;
+  lastRight: boolean;
+  offset: { left?: number; right?: number };
+}
+
+interface UseStickyStylesProps {
+  stickyState: StickyStateModel;
+  columnId: ColumnId;
+  getClassName: (styles: null | StickyStateCellStyles) => Record<string, boolean>;
+}
+
+interface StickyStyles {
+  ref: React.RefCallback<HTMLElement>;
+  className?: string;
+  style?: React.CSSProperties;
+}
+
+export function useStickyStyles({ stickyState, columnId, getClassName }: UseStickyStylesProps): StickyStyles {
+  const ref = useRef<HTMLElement>(null) as React.MutableRefObject<HTMLElement>;
+  const setCell = stickyState.refs.cell;
+  const refCallback = useCallback(
+    node => {
+      ref.current = node;
+      setCell(columnId, node);
+    },
+    [columnId, setCell]
+  );
+
+  useReaction(
+    stickyState.store,
+    state => state.cellStyles[columnId],
+    props => {
+      const className = getClassName(props);
+
+      const cellElement = ref.current;
+      if (cellElement) {
+        Object.keys(className).forEach(key => {
+          if (className[key]) {
+            cellElement.classList.add(key);
+          } else {
+            cellElement.classList.remove(key);
+          }
+        });
+        cellElement.style.left = props?.offset?.left !== undefined ? `${props?.offset?.left}px` : '';
+        cellElement.style.right = props?.offset?.right !== undefined ? `${props?.offset?.right}px` : '';
+      }
+    }
+  );
+
+  const cellStyles = stickyState.store.get().cellStyles[columnId];
+  return {
+    ref: refCallback,
+    className: cellStyles ? clsx(getClassName(cellStyles)) : undefined,
+    style: cellStyles?.offset ?? undefined,
+  };
+}
+
+export function useStickyState({
+  visibleColumns,
+  stickyColumnsFirst,
+  stickyColumnsLast,
+}: StickyStateProps): StickyStateModel {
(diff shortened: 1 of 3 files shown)


pan-kot (dismissed):

src/table/use-sticky-columns.ts
```diff
@@ -0,0 +1,362 @@
...
+        scrollPadding: { left: this.stickyWidthLeft, right: this.stickyWidthRight },
+      }));
+    }
+  }
+
+  private updateScroll(props: UpdateCellStylesProps) {
+    const wrapperScrollLeft = props.wrapper.scrollLeft;
+    const wrapperScrollWidth = props.wrapper.scrollWidth;
+    const wrapperClientWidth = props.wrapper.clientWidth;
+    const tablePaddingLeft = parseInt(getComputedStyle(props.table).paddingLeft) || 0;
```
connorlanigan, line 265: I'd suggest to use `parseFloat` here instead of `parseInt`, to keep the precision of the value.

(In case you want to keep `parseInt`, I recommend adding the `radix` parameter to it to prevent any accidental conversion, see [MDN docs on parseInt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt))

src/table/use-sticky-columns.ts
```diff
@@ -0,0 +1,362 @@
...
+    const tablePaddingRight = parseInt(getComputedStyle(props.table).paddingRight) || 0;
+
+    this.isStuckToTheLeft = wrapperScrollLeft > tablePaddingLeft;
+    this.isStuckToTheRight = Math.ceil(wrapperScrollLeft) < wrapperScrollWidth - wrapperClientWidth - tablePaddingRight;
```
connorlanigan, line 269: Why do we use `Math.ceil` here? This could have a risk of producing an incorrect result e.g. when the page is zoomed in and then scrolled to a non-integer scroll position at the end of the table

src/table/use-sticky-columns.ts
```diff
@@ -0,0 +1,362 @@
...
+        acc[columnId] = null;
+        return acc;
+      }
+
+      // Determine the offset of the sticky column using the `cellOffsets` state object
+      const isFirstColumn = index === 0;
+      const stickyColumnOffsetLeft = this.cellOffsets.get(columnId)?.first ?? 0;
+      const stickyColumnOffsetRight = this.cellOffsets.get(columnId)?.last ?? 0;
+      const cellStyle = {
+        left: stickyColumnOffsetLeft !== undefined && stickySide === 'left' ? stickyColumnOffsetLeft : undefined,
```
connorlanigan, line 293: I think the first part (`stickyColumnOffsetLeft !== undefined`) of the conditional is redundant

src/table/use-sticky-columns.ts
```diff
@@ -0,0 +1,362 @@
...
+        right: stickyColumnOffsetRight !== undefined && stickySide === 'right' ? stickyColumnOffsetRight : undefined,
+      };
+      const lastLeftStickyColumnIndex = props.stickyColumnsFirst - 1;
+      const lastRightStickyColumnIndex = props.visibleColumns.length - props.stickyColumnsLast;
+      const tablePaddingLeft = parseInt(getComputedStyle(props.table).paddingLeft) || 0;
```
connorlanigan, line 298: I suggest `parseFloat` here as well, see comment above

connorlanigan, line 298: `getComputedStyle` can force the browser to relayout the page i. I'm not sure if the result of that is cached, but since this value should be the same for all cells I suggest to move this constant out of the loop, e.g. into line 274, to reduce the risk of a performance impact.

See also https://gist.github.com/paulirish/5d52fb081b3570c81e3a for more background

src/table/use-sticky-columns.ts
```diff
@@ -0,0 +1,362 @@
...
+
+  private updateCellOffsets = (props: UpdateCellStylesProps): void => {
+    const firstColumnWidths: number[] = [];
+    for (let i = 0; i < props.visibleColumns.length; i++) {
+      const element = props.cells[props.visibleColumns[i]];
+      const cellWidth = element.getBoundingClientRect().width ?? 0;
+      firstColumnWidths[i] = (firstColumnWidths[i - 1] ?? 0) + cellWidth;
+    }
+
+    const lastColumnsWidths: number[] = [];
```
connorlanigan, line 318: minor: one constant uses "Column*s*" plural while the other uses "Column" singular, that seems unintentional

src/table/use-sticky-columns.ts
```diff
@@ -0,0 +1,362 @@
...
+        stickyColumnsFirst,
+        stickyColumnsLast,
+      });
+    }
+  }, [store, stickyColumnsFirst, stickyColumnsLast, visibleColumns]);
+
+  // Update wrapper styles imperatively to avoid unnecessary re-renders.
+  useEffect(() => {
+    if (!hasStickyColumns) {
+      return;
```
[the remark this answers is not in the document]
pan-kot: 1. This condition: https://github.com/cloudscape-design/components/pull/1026/files#diff-9cfc241d7a1172294b6c893446be0a43d07609300fb1db65ad76c0855df90d56R251 ensures one more update is fired after the state changes which should happen before unsubscribe (can check it though);
2. Both hooks also return styles as React props so that the reset happens within the render when the state changes.

src/table/use-sticky-columns.ts
```diff
@@ -0,0 +1,362 @@
...
+        const cellElement = cellRef.current;
+        if (cellElement) {
+          Object.keys(className).forEach(key => {
+            if (className[key]) {
+              cellElement.classList.add(key);
+            } else {
+              cellElement.classList.remove(key);
+            }
+          });
+          cellElement.style.left = state?.offset?.left !== undefined ? `${state?.offset?.left}px` : '';
```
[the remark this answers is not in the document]
pan-kot: The state is nullable but the offset is not so can remove the ? after offset it looks like.

src/table/use-sticky-columns.ts
pan-kot: Good point 👍

src/table/use-sticky-columns.ts
rubencarvalho: When a table is used in its `full-page` variant in our AppLayout component, it sometimes produces the weird result where it's always "stuck to the right".
This is because - in certain browsers -  the `wrapperScrollLeft` value was a non-integer, lower than the element's `clientWidth` (by, for example, 0.5 px).
While I understand the risk you pointed out, if it does happen, I see it as being an edge-case with very low impact: the user would need to scroll 1 more pixel to be able to see the "stuck" state of the column.

src/table/use-sticky-columns.ts
```diff
@@ -0,0 +1,362 @@
...
+      setCell(columnId, node);
+    },
+    [columnId, setCell]
+  );
+
+  // Update cell styles imperatively to avoid unnecessary re-renders.
+  useEffect(
+    () => {
+      if (!stickyColumns.isEnabled) {
+        return;
```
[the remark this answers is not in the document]
rubencarvalho: It should be covered by the [condition](https://github.com/cloudscape-design/components/pull/1026/files#diff-9cfc241d7a1172294b6c893446be0a43d07609300fb1db65ad76c0855df90d56R251) posted above by Andrei, where `hadStickyColumns` allows for one extra styles update.

src/table/use-sticky-columns.ts
rubencarvalho: Good call, also great linked resource. Thanks!

connorlanigan (changes requested):

src/table/__tests__/use-sticky-columns.test.tsx
```diff
@@ -0,0 +1,211 @@
...
+  const { result, rerender } = renderHook(() =>
+    useStickyColumns({ visibleColumns: [], stickyColumnsFirst: 0, stickyColumnsLast: 0 })
+  );
+  result.current.refs.wrapper(tableWrapper);
+  result.current.refs.table(table);
+  const updateCellStylesSpy = jest.spyOn(result.current.store, 'updateCellStyles');
+
+  expect(updateCellStylesSpy).not.toHaveBeenCalled();
+
+  rerender({ visibleColumns: [], stickyColumnsFirst: 1, stickyColumnsLast: 0 });
```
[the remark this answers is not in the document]
pan-kot: Amazing catch! Fixed.

connorlanigan (approved):

PR #1040, opened 2023-05-01
feat: Make table header row sticky on mobile

rubencarvalho:
### Description

Description:
This PR updates the table component to make only the table header row sticky on mobile devices, instead of the entire header container. This improves the user experience on mobile devices by preserving the sticky behavior of the table header row, while allowing other header elements, such as title, filters or buttons, to scroll out of view.

Changes:
- Updated the sticky behavior in `src/container/internal.tsx` to take into account the new internal `__disableStickyMobile` prop.
- Added the prop to the sticky header in `src/container/use-sticky-header.ts` and made it disable the sticky feature on mobile devices by default (status-quo).
- Updated the sticky header test in `src/table/__integ__/sticky-header.test.ts` to reflect the new behavior on mobile devices.
- Added a `toolsHeaderWrapper` ref and calculated the `mobileStickyOffset` in `src/table/internal.tsx` to correctly offset the sticky table header row when on a mobile device and passed the new `__disableStickyMobile` as `false`.

### How has this been tested?

Existing and newly updated integration test.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/container/use-sticky-header.ts
+++ b/src/container/use-sticky-header.ts
@@ -54,7 +54,7 @@ export const useStickyHeader = (
    * to the default offset calculated in AppLayoutDomContext.
    */
   let computedOffset = `${effectiveStickyOffset}px`;
-  if (isRefresh && !hasInnerOverflowParents) {
+  if (isRefresh && !hasInnerOverflowParents && !isMobile) {
     computedOffset = `var(${customCssProps.offsetTopWithNotifications}, ${computedOffset})`;
   }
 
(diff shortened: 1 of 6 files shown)


src/container/use-sticky-header.ts
```diff
@@ -92,8 +95,3 @@ export const useStickyHeader = (
     stickyStyles,
   };
 };
-
-export function useSupportsStickyHeader() {
-  const isMobile = useMobile();
-  return supportsStickyPosition() && !isMobile;
```
YueyingLu, line 98: Why we do not need to take `supportsStickyPosition()` into consideration? What about the server-side?

rubencarvalho: Great call - I've actually overseen the SSR check. I've re-added it.

YueyingLu (approved):

PR #1048, opened 2023-05-03
fix: Ensure that container expandable section headers wrap as expected

gethinwebster:
### Description

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/expandable-section/__tests__/expandable-section.test.tsx
```diff
@@ -205,46 +205,18 @@ describe('Variant container with headerText', () => {
...
       headerAriaLabel: 'ARIA Label',
     });
-    const headerButton = wrapper.findHeader().find('[role="button"]')!.getElement();
+    const headerButton = wrapper.findHeader().getElement();
     const content = wrapper.findContent().getElement();
     expect(headerButton).toHaveAttribute('aria-label', 'ARIA Label');
     expect(content).toHaveAttribute('aria-label', 'ARIA Label');
-    expect(headerButton).not.toHaveAttribute('aria-labelledby');
-  });
-  test('set aria-labelledby when no headerAriaLabel is set', () => {
```
gethinwebster, line 223: standardized the different variants, so this was no longer required

src/expandable-section/expandable-section-header.tsx
```diff
@@ -141,30 +130,9 @@ const ExpandableContainerHeader = ({
...
-          <span>{children}</span>
-        </span>
-      </Wrapper>
-      {isContainer && (
-        <ScreenreaderOnly id={screenreaderContentId}>
-          {children} {headerCounter} {headerDescription}
-        </ScreenreaderOnly>
-      )}
-    </div>
+    <ExpandableDefaultHeader {...props} variant={variant}>
```
gethinwebster, line 133: I couldn't find a strong reason for the underlying DOM here to be so different, so have tried to standardize the different variants

YueyingLu (changes requested):

src/expandable-section/expandable-section-header.tsx
YueyingLu: The reason is to not have heading inside button, see AWSUI-18961

Al-Dani: this part was also significantly changed in a [recent PR](https://github.com/cloudscape-design/components/pull/1044)

Al-Dani (changes requested):

src/expandable-section/expandable-section-header.tsx
gethinwebster: ah got it, thanks! will take another look

gethinwebster: Updated to address this with styling only. That does mean a slight regression in terms of wrapping behavior of the header & icon (see screenshot tests), but it also used to be this way in the past, so I think this is acceptable

YueyingLu: What about we do not use counter in Header but add right after child? Like 
```
<span className={clsx(styles['icon-container'], styles[`icon-container-${variant}`])}>{icon}</span>
<span>{children} {headerCounter !== undefined && <span className={styles.counter}> {headerCounter}</span>}</span> 
```

gethinwebster: I'd rather keep as much as possible relying on the <Header> component rather than duplicating things in here

YueyingLu (approved):

PR #1050, opened 2023-05-03
fix: Don't unset form field context around popover trigger

avinashbot:
### Description

We added a wrapper unsetting form field context around dialog components like modal and popover in #850. That was a little over-eager with popover and also reset it around the trigger, not just the dialog. This moves the context provider to wrap just the content.

Related links, issue #, if available: AWSUI-21042

### How has this been tested?

One-time manual check. Not a use case I want to, uh, endorse.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/popover/__tests__/popover.test.tsx
+++ b/src/popover/__tests__/popover.test.tsx
@@ -98,7 +98,7 @@ describe('Dismiss button', () => {
       const wrapper = renderPopover({ children: 'Trigger', content: 'Popover', renderWithPortal });
       wrapper.findTrigger().click();
       act(() => {
-        document.dispatchEvent(new MouseEvent('mousedown'));
+        document.dispatchEvent(new MouseEvent('mousedown', { bubbles: true }));
       });
       expect(wrapper.findBody({ renderWithPortal })).toBeNull();
     });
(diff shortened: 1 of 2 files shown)


src/popover/__tests__/popover.test.tsx
```diff
@@ -98,7 +98,7 @@ describe('Dismiss button', () => {
       const wrapper = renderPopover({ children: 'Trigger', content: 'Popover', renderWithPortal });
       wrapper.findTrigger().click();
       act(() => {
-        document.dispatchEvent(new MouseEvent('mousedown'));
+        document.dispatchEvent(new MouseEvent('mousedown', { bubbles: true }));
```
jperals, line 101: Why was `{ bubbles: true }` not necessary before and now it is?

avinashbot: Codecov 😔

There's a click handler (line 160) that wasn't getting triggered before, and the PR's getting blocked over lack of coverage for that one line.

jperals: OK. This should have no effect on real users because native events bubble up by default, right? (So actually this is a better simulation of a real world interaction)

I would just add a comment in the tests to explain that.

jperals: > Well, I only changed the tests :P

Yeah well, my question was not very well formulated —what I wondered is whether this change being necessary in the test was a sign that something changed that could affect the user's experience too. But from what you say the test was also not firing the event correctly before this change, so in fact the bubbling would have been necessary already as well.

> Pushing back slightly cause I don't know what I'd put in a comment that explains more than what's obvious

Fair enough, no big concerns.

jperals (approved):

PR #1052, opened 2023-05-04
fix: Remove sticky styles from container on mobile

rubencarvalho:
### Description

This PR fixes a regression caused by https://github.com/cloudscape-design/components/pull/1040, where the container would get the sticky styles (a box shadow) on mobile even though only the table header was sticky.

### How has this been tested?

Existing screenshot tests.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/container/internal.tsx
+++ b/src/container/internal.tsx
@@ -9,6 +9,7 @@ import { InternalBaseComponentProps } from '../internal/hooks/use-base-component
 import { StickyHeaderContext, useStickyHeader } from './use-sticky-header';
 import { useDynamicOverlap } from '../internal/hooks/use-dynamic-overlap';
 import { useMergeRefs } from '../internal/hooks/use-merge-refs';
+import { useMobile } from '../internal/hooks/use-mobile';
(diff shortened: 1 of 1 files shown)


src/container/internal.tsx
```diff
@@ -97,7 +99,7 @@ export default function InternalContainer({
         styles.root,
         styles[`variant-${variant}`],
         fitHeight && styles['fit-height'],
-        isSticky && [styles['sticky-enabled']]
+        isSticky && !isMobile && [styles['sticky-enabled']]
```
jperals, line 102: I think it is not going to be obvious for future readers why we make this distinction. At minimum I would suggest to add a comment. Perhaps ideally there is room for some refactoring, by replacing the term "sticky" with "sticky header" or "sticky table header", depending on the case, for more clarity.

rubencarvalho: Good point, I added a comment + a variable to hopefully make it more clear.

jperals (approved):

PR #1056, opened 2023-05-04
fix: Table sticky scrollbar can be hidden globally

pan-kot:
### Description

The table sticky scrollbar has fixed position and need to adjust every time the table updates its size or position. We can detect size changes with the resize observer but position detection is fairly limited. For example, in board-components the items can be repositioned with D&D. Moreover, when dragging, the container's position can be set to absolute. That leads to issues when the sticky scrollbar position not being updated or being updated incorrectly.

Proposed a global flag to hide all sticky scrollbars during interactions such as D&D that can result in table container being moved around.

Before fix:

https://user-images.githubusercontent.com/20790937/236244608-69b4c4f5-1e5c-4638-b1ba-fe6d4d0673f0.mov

After fix (requires a small adjustment to board components to set/unset the body attribute):

https://user-images.githubusercontent.com/20790937/236245054-267975af-c1ac-4848-b35d-4419c584b708.mov

### How has this been tested?

Manual testing with new custom table test page and board components.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/table/use-sticky-scrollbar.ts
+++ b/src/table/use-sticky-scrollbar.ts
@@ -124,10 +124,9 @@ export function useStickyScrollbar(
   }, [wrapperEl]);
 
   useEffect(() => {
-    if (supportsStickyPosition() && tableRef.current) {
-      const observer = new ResizeObserver(entries => {
+    if (supportsStickyPosition() && wrapperRef.current && tableRef.current) {
+      const observer = new ResizeObserver(() => {
         if (scrollbarContentRef.current) {
-          scrollbarContentRef.current.style.width = `${entries[0].borderBoxSize[0].inlineSize}px`;
           updatePosition(
             tableRef.current,
             wrapperRef.current,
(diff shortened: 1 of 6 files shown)


src/table/use-sticky-scrollbar.ts
```diff
@@ -138,29 +137,13 @@ export function useStickyScrollbar(
           );
         }
       });
+      // Scrollbar width must be in sync with wrapper width.
+      observer.observe(wrapperRef.current);
```
pan-kot, line 141: This replaces the global resize observer from below and is better as it also detects container width changes not caused by window resize (see screen recordings).

src/table/use-sticky-scrollbar.ts
```diff
@@ -124,10 +124,9 @@ export function useStickyScrollbar(
   }, [wrapperEl]);
 
   useEffect(() => {
-    if (supportsStickyPosition() && tableRef.current) {
-      const observer = new ResizeObserver(entries => {
+    if (supportsStickyPosition() && wrapperRef.current && tableRef.current) {
+      const observer = new ResizeObserver(() => {
         if (scrollbarContentRef.current) {
-          scrollbarContentRef.current.style.width = `${entries[0].borderBoxSize[0].inlineSize}px`;
```
pan-kot, line 130: This line is not needed. We do the same here: https://github.com/cloudscape-design/components/pull/1056/files#diff-c2032c96c95559284e2956e45c92df9ff7e7f5960fab48f5cf3381ee7c528a56R70

avinashbot (approved):
Don't want to block this particular fix, but just a thought. Do we even need `supportsStickyPosition()` anymore? I guess it only really detects SSR and Jest, but most of the calculation is done in effects/resize observers anyway.

pan-kot:
> Don't want to block this particular fix, but just a thought. Do we even need `supportsStickyPosition()` anymore? I guess it only really detects SSR and Jest, but most of the calculation is done in effects/resize observers anyway.

Thanks, let's address it as a follow-up

pan-kot:
Decided to not implement the global visibility toggle. Instead, the scrollbars will be updated to become properly sticky.

PR #1064, opened 2023-05-05
feat: Integrate sticky columns in table component

rubencarvalho:
### Description

This PR integrates the existing sticky columns functionality into our Table component.

### How has this been tested?

Added unit tests.

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/table/body-cell/styles.scss
```diff
@@ -102,6 +102,48 @@ $border-placeholder: awsui.$border-item-width solid transparent;
     }
   }
 
+  &.sticky-cell {
+    position: sticky;
+    background: awsui.$color-background-container-content;
+    z-index: 801; // Our sticky elements should have z-index in the range of 800-850
+    &-pad-left:not(.has-selection) {
```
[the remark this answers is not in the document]
rubencarvalho: The selection cells should not get the padding.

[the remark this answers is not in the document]
pan-kot: why not?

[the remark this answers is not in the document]
pan-kot: Ah, the selection cell it is?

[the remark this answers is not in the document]
rubencarvalho: Design decision - adding padding is something we would prefer to avoid if not strictly necessary.

src/collection-preferences/utils.tsx
```diff
@@ -147,6 +151,68 @@ export const ContentDensityPreference = ({ label, description, value, onChange }
   </InternalCheckbox>
 );
 
+interface StickyColumnsPreferenceProps extends CollectionPreferencesProps.StickyColumnsPreference {
+  onChange: (value: any) => void;
```
timogasda, line 155: Why `any`?

src/collection-preferences/index.tsx
```diff
@@ -45,12 +46,14 @@ export default function CollectionPreferences({
...
+  stickyColumnsPreference,
+
   preferences,
   customPreference,
   ...rest
 }: CollectionPreferencesProps) {
-  const { __internalRootRef } = useBaseComponent(componentName);
-  checkControlled(componentName, 'preferences', preferences, 'onConfirm', onConfirm);
+  const { __internalRootRef } = useBaseComponent('CollectionPreferences');
+  checkControlled('CollectionPreferences', 'preferences', preferences, 'onConfirm', onConfirm);
```
timogasda, line 56: Why not keep using the existing `componentName` variable?

src/collection-preferences/utils.tsx
```diff
@@ -147,6 +151,68 @@ export const ContentDensityPreference = ({ label, description, value, onChange }
...
+  value?: { first?: number; last?: number };
+}
+interface StickyPreference extends CollectionPreferencesProps.StickyColumnsPreference {
+  onChange: (value: any) => void;
```
timogasda, line 159: `any` again

src/table/body-cell/td-element.tsx
```diff
@@ -3,6 +3,7 @@
 import clsx from 'clsx';
 import React from 'react';
 import styles from './styles.css.js';
+import { StickyColumnsModel, useStickyCellStyles } from '../use-sticky-columns.js';
```
timogasda, line 6: minor: remove the `.js`

src/table/internal.tsx
```diff
@@ -34,6 +34,8 @@ import useTableFocusNavigation from './use-table-focus-navigation';
 import { SomeRequired } from '../internal/types';
 import { TableTdElement } from './body-cell/td-element';
 import { useMobile } from '../internal/hooks/use-mobile';
+import { useStickyColumns } from './use-sticky-columns';
+import { selectionColumnId } from './use-sticky-columns';
```
timogasda, line 38: These imports can be combined

src/table/internal.tsx
```diff
@@ -156,6 +157,19 @@ const InternalTable = React.forwardRef(
     const hasSelection = !!selectionType;
     const hasFooter = !!footer;
 
+    const noStickyColumns = !stickyColumns?.first && !stickyColumns?.last;
```
timogasda, line 160: minor: negated boolean variables take extra time to "process" in our heads. can we make this `hasStickyColumns` like the other variables above?

src/table/thead.tsx
```diff
@@ -89,12 +92,30 @@ const Thead = React.forwardRef(
...
     const { columnWidths, totalWidth, updateColumn } = useColumnWidths();
 
+    const stickyStyles = useStickyCellStyles({
+      stickyColumns: stickyState,
+      columnId: selectionColumnId.toString(),
+      getClassName: props => ({
+        [styles['sticky-cell']]: !!props,
+        [styles['sticky-cell-pad-left']]: !!props?.padLeft,
+        [styles['sticky-cell-last-left']]: !!props?.lastLeft,
+        [styles['sticky-cell-last-right']]: !!props?.lastRight,
```
timogasda, line 102: This is feedback for the actual hook and not directly this PR, but: I'm seeing this object/code block repeated quite a few times. Can this logic not be integrated into the hook somehow?

src/table/thead.tsx
```diff
@@ -89,12 +92,30 @@ const Thead = React.forwardRef(
 
     const { columnWidths, totalWidth, updateColumn } = useColumnWidths();
 
+    const stickyStyles = useStickyCellStyles({
```
timogasda, line 95: side note/thought: It's unfortunate that we have to use this hook for every Thead, even though it's just used for the select cell. I feel like the select cell should be moved into its own component... 🤔

src/table/use-sticky-columns.ts
```diff
@@ -224,6 +231,23 @@ export function useStickyCellStyles({
...
+  if (s1 && s2) {
+    return (
+      s1.padLeft === s2.padLeft &&
+      s1.lastLeft === s2.lastLeft &&
+      s1.lastRight === s2.lastRight &&
+      s1.offset.left === s2.offset.left &&
+      s1.offset.right === s2.offset.right
+    );
+  }
+  return s1 === s2;
```
timogasda, line 244: Isn't this always false, though?

src/collection-preferences/__tests__/sticky-columns.test.tsx
```diff
@@ -0,0 +1,142 @@
...
+describe('Sticky columns preference selection', () => {
+  test('correctly displays titles', () => {
+    const wrapper = renderWithStickyColumnsPreferences({});
+    wrapper.findTriggerButton().click();
+    const preferencesModalWrapper = wrapper.findModal();
+
+    expect(preferencesModalWrapper!.findStickyColumnsPreference()!.findTitle('first').getElement()).toHaveTextContent(
+      'Stick first column(s)'
+    );
+    expect(preferencesModalWrapper!.findStickyColumnsPreference()!.findTitle('last').getElement()).toHaveTextContent(
```
timogasda, line 39: Shouldn't these be `preferencesModalWrapper!.findStickyColumnsPreference('first')!.findTitle()`instead?

src/collection-preferences/__tests__/sticky-columns.test.tsx
```diff
@@ -0,0 +1,142 @@
...
+    const wrapper = renderWithStickyColumnsPreferences({});
+    wrapper.findTriggerButton().click();
+    const preferencesModalWrapper = wrapper.findModal();
+
+    expect(
+      preferencesModalWrapper!.findStickyColumnsPreference()!.findDescription('first').getElement()
+    ).toHaveTextContent('Keep the first column(s) visible while horizontally scrolling table content.');
+    expect(
+      preferencesModalWrapper!.findStickyColumnsPreference()!.findDescription('last').getElement()
+    ).toHaveTextContent('Keep the last column visible while horizontally scrolling table content.');
```
timogasda, line 53: Same as above. The `findDescription` method doesnt have arguments, does it?

timogasda (changes requested):

src/table/thead.tsx
pan-kot: We need it for every sticky thead. Can be optimised as:

```
function Thead(props) {
  if (props.isSticky) {
    return <StickyThead {...props} />
  }
  return <NormalThead {...props} />
}
```

But is it worth it? The hook does not create a subscription for non-sticky cells. Only the refs are assigned.

timogasda: No, look at the whole file: https://github.com/cloudscape-design/components/blob/3de4577329a8eefca3693c2474287eb203d49edf/src/table/thead.tsx.

The `stickyStyles` value is only used for the selection header cell. All other Theads get their sticky state from the `stickyState` prop. 

So maybe we can move this `useStickyCellStyles` call up and pass down the `stickyState` for all Thead cells?

pan-kot: Ah, I see. We can't move the useStickyCellStyles up because it produces props for only once cell. Instead, we can create a separate element to the selection cell same as it was done for body cells. However, that's produce quite some diff.

src/table/thead.tsx
rubencarvalho: We explicitly wanted to leave this out to the consumer because it has the `styles` dependency, which vary from cell type to cell type.

src/collection-preferences/__tests__/sticky-columns.test.tsx
```diff
@@ -46,17 +46,26 @@ describe('Sticky columns preference selection', () => {
...
-      preferencesModalWrapper!.findStickyColumnsPreference()!.findDescription('last').getElement()
+      preferencesModalWrapper!.findStickyColumnsPreference('last')!.findDescription().getElement()
     ).toHaveTextContent('Keep the last column visible while horizontally scrolling table content.');
   });
   test('correctly displays labels', () => {
     const wrapper = renderWithStickyColumnsPreferences({});
     wrapper.findTriggerButton().click();
-    const firstColumnsOptions = wrapper.findModal()!.findStickyColumnsPreference()!.findOptions('first');
-    const lastColumnsOptions = wrapper.findModal()!.findStickyColumnsPreference()!.findOptions('last');
+    console.log(wrapper.findModal()!.findStickyColumnsPreference('first')!.findRadioGroup());
```
timogasda, line 58: `console.log` !

src/table/use-sticky-columns.ts
```diff
@@ -215,7 +215,11 @@ export function useStickyCellStyles({
       const unsubscribe = stickyColumns.store.subscribe(selector, (newState, prevState) =>
         updateCellStyles(selector(newState), selector(prevState))
       );
-      return unsubscribe;
+      return () => {
+        unsubscribe();
+        // Force the cleanup
+        updateCellStyles(null, { padLeft: false, lastLeft: false, lastRight: false, offset: {} });
```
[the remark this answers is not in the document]
pan-kot: I guess you can at least change the signature to allow `prev` to be null as well.

timogasda (approved):

src/table/use-sticky-columns.ts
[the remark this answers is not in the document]
rubencarvalho: What do you mean by "change the signature to allow `prev` to be null as well"?

Do you mean having something like
`  if (prev === null || !isCellStatesEqual(state, prev)) { ...`

timogasda (approved):

PR #1076, opened 2023-05-09
feat: Added full-width prop for buttons

pan-kot:
### Description

Added fullWidth property for normal, primary and link buttons.

AWSUI-19884

### How has this been tested?

New unit tests and permutations

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/__tests__/__snapshots__/documenter.test.ts.snap
```diff
@@ -2646,6 +2646,12 @@ This property only applies when an \`href\` is provided.",
       "optional": true,
       "type": "string",
     },
+    Object {
+      "description": "Makes the button occupy 100% of the parent container width. The button content is centered.",
```
jperals, line 2650: Has this been through content review?

just-boris: Form field has a `stretch` property. Should it be the same name here? 

Did you think of any other alternatives when preparing this PR?

gethinwebster:
Will there be updated "Usage" guidelines for this property, i.e. when it should be used from a design point of view?

src/__tests__/__snapshots__/documenter.test.ts.snap
pan-kot: I considered:
* More variants ("full-width", "primary-full-width", "link-full-width"). However, I thought a flag is unambiguous enough;
* A `stretch` property. I decided against that because the "full-width" sounds more like a variant while "stretch" is more like a property. The full width buttons not only have 100% width but also the content is centered and there might be more differences introduced in the future.

pan-kot:
> Will there be updated "Usage" guidelines for this property, i.e. when it should be used from a design point of view?

Yes, there will be a new guideline added. It is under preparation.

pan-kot:
> Will there be updated "Usage" guidelines for this property, i.e. when it should be used from a design point of view?

Yes, there will be a new guideline added. It is under preparation.

pan-kot:
> Will there be updated "Usage" guidelines for this property, i.e. when it should be used from a design point of view?

Yes, there will be a new guideline added. It is under preparation.

pan-kot:
> Will there be updated "Usage" guidelines for this property, i.e. when it should be used from a design point of view?

Yes, there will be a new guideline added. It is under preparation.

src/__tests__/__snapshots__/documenter.test.ts.snap
timogasda: I agree that it probably shouldn't be a variant because then we would need multiple new variants.

I don't have a strong opinion of `stretch` vs `fullWidth`, but I lean slightly more towards `stretch`.

jperals: Is the content review done now?

stirache: Suggestion: "Sets the button width to be 100% of the parent container width. Button content is centered."

stirache (changes requested):

stirache (approved):

timogasda (approved):

jperals (approved):

PR #1080, opened 2023-05-09
chore: default `xScaleType` to `categorical` for bar charts

jperals:
### Description

Some customers were confused by the fact that the `xScaleType` prop is not mandatory and defaults to `linear`, but needs to be set to `categorical` for bar charts to work correctly. This PR addresses this in two ways:
- Adds a clarification sentence in the API docs, stating that `categorical` needs to be used
- Defaults the prop to `categorical` for bar charts ~~and mixed line-bar charts~~

~~The usage guidelines will likely be updated as well.~~ Done

For backwards compatibility, this PR doesn't restrict the possible values to be only `categorical`. It also doesn't deprecate the prop because it is not discarded that these graphs support more types in the future.

Related issue: AWSUI-19806

### How has this been tested?

- Ran a dry-run build successfully

<details>

#### Correctness

#### Security

#### Testing

</details>

---


stirache (dismissed):

src/mixed-line-bar-chart/index.tsx
```diff
@@ -13,7 +13,7 @@ export { MixedLineBarChartProps };
 function MixedLineBarChart<T extends number | string | Date>({
   series = [],
   height = 500,
-  xScaleType = 'linear',
+  xScaleType = 'categorical',
```
timogasda, line 16: Would this not be a breaking change? I think we can get away with it in the Bar chart, because you can literally only use `categorical` there.

But you could theoretically build a Mixed line bar chart that only contains lines, right?

timogasda: No, I can use other scale types as long as I dont use bars. This is a Mixed Line Bar Chart:
Screenshot 2023-05-25 at 08 31 13

jperals: Right, I think I'll leave mixed charts unchanged.

jperals: Changed mixed charts back to default to `linear` (even though it makes little sense to use mixed charts to display only lines?)

timogasda (approved):

PR #1098, opened 2023-05-15
fix: drawer widths

nickhudkins:
### Description

Removed `controlled/uncontrolled` concept and allowed resize handling to be controlled, with a callback to provide the ability to persist the last value. Changing `size` to initialSize to indicate that we only use this value as... an initial size.

### How has this been tested?

Tested locally using `with-drawers.page.tsx`, and included writing `onResize` value to local storage, and restoring to initialState. I also enabled, disabled drawers, added drawers dynamically and more!

#### Correctness
It is correct.

#### Security
It is secure.

#### Testing
It has been tested

---

### DIFF
--- a/pages/app-layout/with-drawers.page.tsx
+++ b/pages/app-layout/with-drawers.page.tsx
@@ -14,6 +14,7 @@ import {
 import appLayoutLabels from './utils/labels';
 import { Breadcrumbs, Containers } from './utils/content-blocks';
 import ScreenshotArea from '../utils/screenshot-area';
+import type { DrawerItem } from '~components/app-layout/drawer/interfaces';
 
 export default function WithDrawers() {
   const [activeDrawerId, setActiveDrawerId] = useState<string | null>(null);
@@ -21,19 +22,12 @@ export default function WithDrawers() {
   const [hasTools, setHasTools] = useState(false);
   const [isToolsOpen, setIsToolsOpen] = useState(false);
 
-  const [widths, setWidths] = useState<{ [id: string]: number }>({
-    security: 500,
-  });
-
   const drawers = !hasDrawers
     ? null
     : {
         drawers: {
           ariaLabel: 'Drawers',
           activeDrawerId: activeDrawerId,
-          onResize: (event: NonCancelableCustomEvent<{ size: number; id: string }>) => {
-            setWidths({ ...widths, [event.detail.id]: event.detail.size });
-          },
           items: [
             {
               ariaLabels: {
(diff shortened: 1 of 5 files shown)


src/app-layout/drawer/interfaces.ts
```diff
@@ -75,7 +75,8 @@ export interface DrawerItem {
   };
   ariaLabels: DrawerItemAriaLabels;
   resizable?: boolean;
-  size?: number;
+  initialSize?: number;
```
just-boris, line 78: In other components we use `defaultSomething` to provide state only for the first render

pages/app-layout/with-drawers.page.tsx
```diff
@@ -45,7 +38,13 @@ export default function WithDrawers() {
               content: <Security />,
               id: 'security',
               resizable: true,
-              size: widths.security,
+              initialSize: 500,
+              onResize: ({ size }: { size: number }) => {
```
just-boris, line 42: I think, this won't work, because the actual value is `event.detail.size`.

I think, this approach (without `detail`) is preferred for this case

src/app-layout/index.tsx
```diff
@@ -727,6 +723,10 @@ const OldAppLayout = React.forwardRef(
                   size={!isResizeInvalid ? drawerSize : toolsWidth}
                   onResize={changeDetail => {
                     fireNonCancelableEvent(drawers.onResize, changeDetail);
+                    const drawerItem = drawerItems.find(({ id }) => id === changeDetail.id);
+                    if (drawerItem?.onResize) {
+                      fireNonCancelableEvent(drawerItem.onResize, changeDetail);
+                    }
```
just-boris, line 729:
```suggestion
                    drawerItem?.onResize?.({ size: changeDetail.size })
```

src/app-layout/index.tsx
```diff
@@ -269,31 +269,27 @@ const OldAppLayout = React.forwardRef(
...
-        componentName: 'AppLayout',
-        controlledProp: 'drawers.items[].size',
-        changeHandler: 'drawers.onResize',
-      }
-    );
+    const [drawerSizes, setDrawerSizes] = useState(() => getDrawerItemSizes());
+
+    useEffect(() => {
+      setDrawerSizes(prev => ({ ...getDrawerItemSizes(), ...prev }));
+    }, [getDrawerItemSizes]);
```
just-boris, line 292: Not sure if this is expected to work this way. When new drawers are being added (for example with the runtime API #1016) it will reset sizes for all previous drawers.

Fine for now, added this comment for future reference

nickhudkins: I think line 282 was intended to prevent this but with `sizes` being local to that function it will indeed overwrite. I need to re-test this because I did add and remove drawers while testing to confirm that other drawer sizes were left untouched. That said, I likely only added drawers with the default size.

nickhudkins: Ok, confirmed that this does work as I hoped. the size of drawers is stored in state (the counterpart of `setDrawerSizes`), we are using the callback style of setState so that we can set drawer sizes (to the calculated default sizes, then using any previous values to "override" those values. I could instead filter out any existing from the state update but "apply all and keep existing" felt fine to me. Let me know if you'd like me to change this.

pages/app-layout/with-drawers.page.tsx
nickhudkins: Yep! I think there might be another place with a similar issue :(. Hidden types eh?

just-boris (approved):

PR #1108, opened 2023-05-19
Add support for TypeScript 4.4's exactOptionalPropertyTypes compiler option to Cards.

CharlesStover:
### Description

TypeScript configurations with `exactOptionalPropertyTypes` turned on are unable to set props in some cases. See also aws/awsui-documentation#14.

This applies `undefined` support for `Cards` to test the waters with a small blast-radius PR.

```
// Before:
function MyCards({ loadingText }) {
  // Type '{ loadingText: string | undefined; ... }' is not assignable to type 'CardsProps<Item>' with
  // 'exactOptionalPropertyTypes: true'. Consider adding 'undefined' to the types of the target's properties.
  return <Cards cardDefinition={{}} items={[]} loadingText={loadingText} />
}
<MyCards /> // ❌

// After:
function MyCards({ loadingText }) {
  return <Cards cardDefinition={{}} items={[]} loadingText={loadingText} />
}
<MyCards /> // 😎👍
```

---

### DIFF
--- a/src/cards/interfaces.tsx
+++ b/src/cards/interfaces.tsx
@@ -54,12 +54,12 @@ export interface CardsProps<T = any> extends BaseComponentProps {
   /**
    * Renders the cards in a loading state. We recommend that you also set a `loadingText`.
    */
-  loading?: boolean;
+  loading?: boolean | undefined;
 
   /**
    * Specifies the text to display when in loading state.
    */
-  loadingText?: string;
+  loadingText?: string | undefined;
 
   /**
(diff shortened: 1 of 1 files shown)


johannes-weber:
Thank you for your pull request! We appreciate your contribution! However, we don't currently support the `exactOptionalPropertyTypes` flag.

As you can see the build failed at the documentation step. Fixing the Documenter is a big deal and our team can't afford it at the moment.

We will be keeping an eye on the demand for this API change on this topic. The need will be taken into account in future prioritization.

PR #1117, opened 2023-05-22
chore: Instrument single page create funnel metrics

michaeldowseza:
### Description

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/breadcrumb-group/item/item.tsx
+++ b/src/breadcrumb-group/item/item.tsx
@@ -146,27 +146,30 @@ export function BreadcrumbItem<T extends BreadcrumbGroupProps.Item>({
     onClick: isLast ? preventDefault : onClickHandler,
   };
 
+  const dataAttibutes: Record<string, string> = {};
+  if (isLast) {
+    dataAttibutes['data-analytics-context'] = 'breadcrumb';
+  }
+
   return (
-    <>
-      <div className={clsx(styles.breadcrumb, isLast && styles.last)}>
-        {isDisplayed && isCompressed ? (
-          <BreadcrumbItemWithPopover
-            item={item}
-            isLast={isLast}
-            anchorAttributes={anchorAttributes}
-            {...itemAttributes}
-          />
-        ) : (
-          <Item isLast={isLast} anchorAttributes={anchorAttributes} {...itemAttributes}>
-            <span className={styles.text}>{item.text}</span>
-          </Item>
-        )}
-        {!isLast ? (
-          <span className={styles.icon}>
-            <InternalIcon name="angle-right" />
-          </span>
-        ) : null}
-      </div>
-    </>
+    <div className={clsx(styles.breadcrumb, isLast && styles.last)} {...dataAttibutes}>
+      {isDisplayed && isCompressed ? (
+        <BreadcrumbItemWithPopover
+          item={item}
+          isLast={isLast}
+          anchorAttributes={anchorAttributes}
+          {...itemAttributes}
+        />
+      ) : (
+        <Item isLast={isLast} anchorAttributes={anchorAttributes} {...itemAttributes}>
+          <span className={styles.text}>{item.text}</span>
+        </Item>
+      )}
+      {!isLast ? (
+        <span className={styles.icon}>
+          <InternalIcon name="angle-right" />
+        </span>
+      ) : null}
+    </div>
   );
 }
--- a/src/button/internal.tsx
+++ b/src/button/internal.tsx
@@ -11,6 +11,7 @@ import { InternalBaseComponentProps } from '../internal/hooks/use-base-component
 import { checkSafeUrl } from '../internal/utils/check-safe-url';
 import { useMergeRefs } from '../internal/hooks/use-merge-refs';
 import LiveRegion from '../internal/components/live-region';
+import { useFunnel } from '../internal/analytics/hooks/use-funnel';
 
 type InternalButtonProps = Omit<ButtonProps, 'variant'> & {
   variant?: ButtonProps['variant'] | 'flashbar-icon' | 'breadcrumb-group' | 'menu-trigger' | 'modal-dismiss';
@@ -60,6 +61,8 @@ export const InternalButton = React.forwardRef(
     const buttonRef = useRef<HTMLElement>(null);
     useForwardFocus(ref, buttonRef);
 
+    const { funnelInteractionId, funnelSubmit } = useFunnel();
+
     const handleClick = useCallback(
       (event: React.MouseEvent) => {
         if (isNotInteractive) {
@@ -70,10 +73,14 @@ export const InternalButton = React.forwardRef(
           fireCancelableEvent(onFollow, { href, target }, event);
         }
 
+        if (funnelInteractionId && variant === 'primary') {
+          funnelSubmit();
+        }
+
         const { altKey, button, ctrlKey, metaKey, shiftKey } = event;
         fireCancelableEvent(onClick, { altKey, button, ctrlKey, metaKey, shiftKey }, event);
       },
-      [isAnchor, isNotInteractive, onClick, onFollow, href, target]
+      [isAnchor, isNotInteractive, onClick, onFollow, href, target, funnelInteractionId, variant, funnelSubmit]
     );
 
     const buttonClass = clsx(props.className, styles.button, styles[`variant-${variant}`], {
--- a/src/container/__tests__/analytics.test.tsx
+++ b/src/container/__tests__/analytics.test.tsx
@@ -1,0 +1,59 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import React from 'react';
+import { render, fireEvent } from '@testing-library/react';
+
+import Container from '../../../lib/components/container';
+
+import { FunnelMetrics } from '../../../lib/components/internal/analytics';
+
+import { mockFunnelMetrics } from '../../internal/analytics/__tests__/mocks';
+
+import {
+  AnalyticsFunnel,
+  AnalyticsFunnelStep,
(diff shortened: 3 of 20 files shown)


src/button/internal.tsx
```diff
@@ -70,10 +73,14 @@ export const InternalButton = React.forwardRef(
           fireCancelableEvent(onFollow, { href, target }, event);
         }
 
+        if (funnelInteractionId && variant === 'primary') {
```
DaemonCahill, line 76: Just a quick note here for offline discussion. This is a good example of an assumption about the relationship between the funnel context, the button variant, and the correct/incorrect implementation of a pattern that could be incorrect.

src/breadcrumb-group/item/item.tsx
```diff
@@ -146,27 +146,30 @@ export function BreadcrumbItem<T extends BreadcrumbGroupProps.Item>({
     onClick: isLast ? preventDefault : onClickHandler,
   };
 
+  const dataAttibutes: Record<string, string> = {};
+  if (isLast) {
+    dataAttibutes['data-analytics-context'] = 'breadcrumb';
+  }
+
   return (
-    <>
```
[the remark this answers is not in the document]
timogasda: nevermind.

src/header/internal.tsx
```diff
@@ -64,7 +64,10 @@ export default function InternalHeader({
           )}
         >
           <HeadingTag className={clsx(styles.heading, styles[`heading-variant-${variantOverride}`])}>
-            <span className={clsx(styles['heading-text'], styles[`heading-text-variant-${variantOverride}`])}>
+            <span
+              data-analytics="heading-text"
```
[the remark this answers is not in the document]
michaeldowseza: Added

src/button/internal.tsx
michaeldowseza: As discussed in DM, there are a few options here to improve this. For now we'll continue as is and iterate on received data.

src/input/internal.tsx
```diff
@@ -166,6 +166,10 @@ function InternalInput(
     attributes.type = 'text';
   }
 
+  if (ariaLabelledby) {
+    (attributes as any)['data-analytics-label-selector'] = `#${ariaLabelledby}`;
```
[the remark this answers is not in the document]
michaeldowseza: Added.

src/link/__tests__/index.test.tsx
```diff
@@ -158,4 +180,54 @@ describe('Link component', () => {
...
+    test('sends a externalLinkInteracted metric when an external link is clicked within a Funnel Context', () => {
+      const { container } = render(
+        <AnalyticsFunnel funnelType="single-page" optionalStepNumbers={[]} totalFunnelSteps={1}>
+          <Link href="#" external={true} />
+        </AnalyticsFunnel>
+      );
+      const wrapper = createWrapper(container).findLink()!;
+      wrapper.click();
+
+      expect(FunnelMetrics.externalLinkInteracted).toHaveBeenCalled();
```
timogasda, line 206: minor: not necessary because of the next line

src/link/__tests__/index.test.tsx
```diff
@@ -158,4 +180,54 @@ describe('Link component', () => {
...
+      expect(FunnelMetrics.externalLinkInteracted).toHaveBeenCalledWith(
+        expect.objectContaining({
+          funnelInteractionId: expect.any(String),
+          elementSelector: expect.any(String),
```
timogasda, line 210: I would suggest checking for the correct strings, too.

src/header/__tests__/header.test.tsx
```diff
@@ -69,3 +69,13 @@ test('supports title tag name override with non-default variant', () => {
   expect(wrapper.find('h3')).toBeNull();
   expect(wrapper.find('h2')!.getElement()).toHaveTextContent('title');
 });
+
+test('adds the correct data attribute', () => {
```
timogasda, line 73: minor: not a great name for a test (applies to some of the others in this PR, too). I would except this to atleast mention "analytics" somehow, and probably have it sit in its own analytics suite (even if its just one test case)

src/form/__tests__/analytics.test.tsx
```diff
@@ -0,0 +1,144 @@
...
+    funnelCancelled: jest.fn(),
+    funnelStepStart: jest.fn(),
+    funnelStepComplete: jest.fn(),
+    funnelStepNavigation: jest.fn(),
+    funnelSubStepStart: jest.fn(),
+    funnelSubStepComplete: jest.fn(),
+    funnelSubStepError: jest.fn(),
+    helpPanelInteracted: jest.fn(),
+    externalLinkInteracted: jest.fn(),
+  });
```
timogasda, line 28: You're using this mock in at least 3 places already. Might be worth moving it to a shared file.

src/form/__tests__/analytics.test.tsx
```diff
@@ -0,0 +1,144 @@
...
+        funnelInteractionId: expect.any(String),
+      })
+    );
+  });
+
+  test('sends a funnelCancelled metric when Form is unmounted without clicking submit', () => {
+    const { unmount } = render(<Form />);
+
+    // funnelCancel is called in the next tick after unmounting
+    unmount();
```
timogasda, line 84: I'm confused by this. The setup is exactly the same as in the previous test case (for `funnelStepComplete`). How are they different?

src/form-field/__tests__/form-field-analytics.test.tsx
```diff
@@ -0,0 +1,159 @@
...
+    expect(FunnelMetrics.funnelSubStepError).toHaveBeenCalledTimes(1);
+    expect(FunnelMetrics.funnelSubStepError).toHaveBeenCalledWith(
+      expect.objectContaining({
+        funnelInteractionId: 'mocked-funnel-id',
+        fieldErrorSelector: expect.any(String),
+        fieldLabelSelector: expect.any(String),
+        subStepSelector: expect.any(String),
+        stepNameSelector: expect.any(String),
+        subStepAllSelector: expect.any(String),
+        subStepNameSelector: expect.any(String),
```
timogasda, line 61: Would it make sense to check for any more specific strings?

src/container/__tests__/analytics.test.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+describe('Funnel Analytics', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+    mockFunnelMetrics();
+  });
+
+  test('renders normally when outside the context of a funnel', () => {
+    render(<Container />);
+    // No error occurred if the component rendered successfully
```
timogasda, line 42: This is not really a test case without any assertions. Is there something we can `expect` here?

src/container/__tests__/analytics.test.tsx
```diff
@@ -0,0 +1,76 @@
...
+            <input data-testid="input" />
+          </Container>
+        </AnalyticsFunnelStep>
+      </AnalyticsFunnel>
+    );
+
+    expect(FunnelMetrics.funnelSubStepStart).toHaveBeenCalledTimes(0);
+
+    fireEvent.focus(getByTestId('input'));
+    fireEvent.blur(getByTestId('input'));
```
timogasda, line 71: Can we use our test utils instead?

michaeldowseza: It is more decoupled to use a regular html input because using the Input component would create an unnecessary dependency.

src/form-field/__tests__/form-field-analytics.test.tsx
michaeldowseza: Most of the selectors use the useUniqueId hook so they have generated strings in them. Is it a good trade off to mock the module in this test?

src/form/__tests__/analytics.test.tsx
```diff
@@ -0,0 +1,144 @@
...
+
```
michaeldowseza, line 85: It is like this because a Form is both the funnel and a single step so I could move the assertions into a single test but its more that funnelStepComplete is always expected to fire when the form unmounts but funnelCancelled is only expected to fire if the Submit button is not clicked.

src/form/__tests__/analytics.test.tsx
michaeldowseza: Moved to a common folder in the analytics folder.

src/link/__tests__/index.test.tsx
michaeldowseza: Changed to haveBeenCalledTimes since we always want to make sure we're not duplicating calls.

src/form-field/__tests__/form-field-analytics.test.tsx
timogasda: If it's not too cumbersome, that would be great. Yeah

src/container/__tests__/analytics.test.tsx
timogasda: Ah i didn't see that you were using a native input. All good.

src/container/__tests__/analytics.test.tsx
michaeldowseza: Added an assertion. This also came with some brought some new changes to not add any of the attributes if outside of the context.

src/form-field/__tests__/form-field-analytics.test.tsx
michaeldowseza: I can maybe do as a follow up. I got push back in a different PR where I mocked a module.

src/link/__tests__/index.test.tsx
michaeldowseza: Changed funnelInteractionId but left elementSelector as its a generated id

timogasda (approved):

PR #1123, opened 2023-05-24
feat: Introduce new `minColumnWidth` property for Column Layout

timogasda:
### Description
This introduced a new way to create layouts using the Column Layout component. Instead of relying on the internal fixed breakpoint grid, you can specify the minimum width of your columns. Given that `minColumnWidth` value, together with the desired maximum number of columns (`columns` property), the component automatically creates a CSS Grid with equal columns.

Main use-case here is text-grid KPIs in resizable dashboards.

Related links, issue #, if available: `8bQdAft0KwOI`

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/column-layout/with-css.tsx
```diff
@@ -0,0 +1,86 @@
...
+  containerWidth: number | null
+): number {
+  if (!containerWidth) {
+    return columns;
+  }
+
+  // First, calculate how many columns we can have based on the current container width and minColumnWidth.
+  const targetColumnCount = Math.min(columns, Math.floor(containerWidth / minColumnWidth));
+
+  // When we start wrapping into fewer columns than desired, we want to keep the number of columns even.
```
YueyingLu, line 26: Why do we want to keep the number even? It may still waste some space.

src/column-layout/interfaces.ts
```diff
@@ -27,6 +27,14 @@ export interface ColumnLayoutProps extends BaseComponentProps {
    */
   disableGutters?: boolean;
 
+  /**
+   * Use together with `columns` to specify the desired minimum width for each column in pixels.
+   *
+   * The number of columns is determined based on the value of this property, the available space,
+   * and the maximum number of columns as defined by the `columns` property.
+   */
+  minColumnWidth?: number;
```
YueyingLu, line 36: When using columns with minColumnWidth it does not have limit for 4 columns. Shall we make it clearer in documentation?

src/column-layout/__tests__/column-layout.test.tsx
```diff
@@ -28,38 +33,48 @@ describe('ColumnLayout component', () => {
...
+              variant: 'text-grid',
+              borders: borders as ColumnLayoutProps['borders'],
+            });
+            expect(wrapper.getElement()).not.toHaveClass(styles['grid-vertical-borders']);
+            expect(wrapper.getElement()).not.toHaveClass(styles['grid-horizontal-borders']);
+          });
+        });
       });
     });
   });
```
YueyingLu, line 80: Can we add test for cssColumnlayout to check `columnLayoutColumnCount`?

timogasda: The main logic is tested with the `calculcateCssColumnCount` utility. But I'll add a test to make sure that the column value is properly transferred to the CSS Custom Property.

src/column-layout/interfaces.ts
timogasda: Good idea. Let me think about that.

src/column-layout/with-css.tsx
timogasda: The reason is to keep the behavior more aligned with the default Column layout.

src/column-layout/with-css.tsx
```diff
@@ -0,0 +1,86 @@
...
+      {flattenedChildren.map((child, i) => {
+        // If this react child is a primitive value, the key will be undefined
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        const key = (child as any).key;
+        const rowIndex = Math.floor(i / columnCount);
+
+        return (
+          <div
+            key={key}
+            className={clsx(styles['restore-pointer-events'], {
```
just-boris, line 74: Do we still need the restore-pointer-events trick in this implementation?

src/internal/generated/custom-css-properties/list.js
```diff
@@ -38,5 +38,6 @@ const customCssPropertiesList = [
   'flashbarStickyBottomMargin',
   'stackedNotificationsBottomMargin',
   'stackedNotificationsDefaultBottomMargin',
+  'columnLayoutColumnCount',
```
just-boris, line 48: How about just setting the property inline?

```jsx
<div className={styles['css-grid']} styles={{gridTemplateColumns: `repeat(var(${columnCount}, 1), 1fr)`}}>
```

Custom props are for cross-component sharing. Here the inline styles will work and the solution will be much simpler.

CSS-vars may bring us back old bugs where different column layouts cannot nest properly (not happening in the current PR, but opens a door I would rather not open)

src/column-layout/styles.scss
```diff
@@ -130,5 +131,49 @@ div.column-layout {
...
+
+    // Alternative implementation based on minimum column widths and CSS Grid
+    &.css-grid {
+      display: grid;
+      grid-template-columns: repeat(var(#{custom-props.$columnLayoutColumnCount}, 1), 1fr);
+
+      pointer-events: none;
+
+      /* stylelint-disable-next-line selector-max-universal */
+      &.grid-variant-text-grid > * {
```
just-boris, line 143: Can we assign a class name (`.column`) to the items, instead of fighting with stylelint?

src/column-layout/styles.scss
```diff
@@ -130,5 +131,49 @@ div.column-layout {
     &.grid-horizontal-borders {
       @include make-horizontal-borders;
     }
+
+    // Alternative implementation based on minimum column widths and CSS Grid
+    &.css-grid {
```
just-boris, line 136: How about extracting new styles into separate file, to have it properly co-located CSS and JS?

timogasda: I could do that, but I'm also reusing a good amount of styles/classes from the main implementation. Mainly things around borders. So I would still need to import those classes.

src/column-layout/styles.scss
timogasda: Sure!

src/internal/generated/custom-css-properties/list.js
timogasda: It's a good point. I've thought about doing it but I figured the CSS custom props approach would be a bit "nicer/cleaner". Not a strong opinion though.

src/column-layout/styles.scss
just-boris: Your call, but if there are not too many classes, I would copy them instead. The "overlay thinking" where I need to add up styles from two places together makes it harder to read and prone to bugs

src/column-layout/with-css.tsx
just-boris: Why not grid-gap? Unification with the other layout? Same idea as https://github.com/cloudscape-design/components/pull/1123#discussion_r1217886392 maybe it is worth separating two instances properly.

timogasda: Yep, I've thought about this a lot and decided to go with this more unified and conservative approach. Thank you for the additional data point though. I will think this over again. I would really like to remove this.

src/column-layout/with-css.tsx
YueyingLu: I wonder if it still makes sense if `minColumnWidth` is set. It feel it is implicitly taking even number of columns as higher priority than minColumnWidth. Although I think `minColumnWidth` should help to take as much space as possible.

YueyingLu: As discussed offline, it makes sense to have balanced view.

YueyingLu (dismissed):

src/column-layout/interfaces.ts
```diff
@@ -27,6 +31,14 @@ export interface ColumnLayoutProps extends BaseComponentProps {
    */
   disableGutters?: boolean;
 
+  /**
+   * Use together with `columns` to specify the desired minimum width for each column in pixels.
+   *
+   * The number of columns is determined based on the value of this property, the available space,
```
stirache, line 37: Suggestion: "The number of columns is determined by the value of  this property, the available space, ...."

stirache (changes requested):

src/column-layout/interfaces.ts
```diff
@@ -9,6 +9,8 @@ export interface ColumnLayoutProps extends BaseComponentProps {
   /**
    * Specifies the number of columns in each grid row.
    * Valid values are any integer between 1 and 4.
+   *
+   * Note that there is no upper limit when used in combination with `minColumnWidth`.
```
just-boris, line 13: This can be simplified

> Specifies the number of columns in each grid row. When `minColumnWidth` is not set, only up to 4 columns are supported.

src/column-layout/interfaces.ts
```diff
@@ -19,6 +21,8 @@ export interface ColumnLayoutProps extends BaseComponentProps {
 
   /**
    * Controls whether dividers are placed between rows and columns.
+   *
+   * Note: This is not supported when used with `minColumnWidth`.
```
just-boris, line 25: Not for this PR, but a general suggestion – how about marking this as `@deprecated` too?

src/column-layout/internal.tsx
```diff
@@ -36,31 +29,31 @@ export default React.forwardRef(function ColumnLayout(
...
+          {children}
+        </ColumnLayoutWithCSS>
+      ) : (
+        <ColumnLayoutWithGrid
+          columns={columns}
+          variant={variant}
+          borders={borders}
+          disableGutters={disableGutters}
+          __breakpoint={__breakpoint}
+          ref={ref}
```
just-boris, line 52: Why do we forward ref to one instance, but not another? If it is some deprecated usage we want to avoid in the new code, it needs a comment, to prevent future developers "fixing" this bug

src/column-layout/interfaces.ts
timogasda: done!

src/column-layout/with-css/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+    targetColumnCount < columns && isOdd(targetColumnCount) ? targetColumnCount - 1 : targetColumnCount
+  );
+}
+
+interface ColumnLayoutWithCSSProps
+  extends Pick<InternalColumnLayoutProps, 'minColumnWidth' | 'columns' | 'variant' | 'borders' | 'disableGutters'> {
+  children: React.ReactNode;
+}
+
+export default function ColumnLayoutWithCSS({
```
just-boris, line 36: After having another look, I find this name confusing. Because this instance uses container query hook, it is not pure CSS, so the name is confusing.

Possible alternatives for the naming: Legacy/Modern, Flex/Grid

just-boris (approved):
Approved with minor follow-up suggestions

src/column-layout/internal.tsx
timogasda: I was wondering this too and I couldn't figure out what this ref is for. So yeah, I'm happy to add a comment.

just-boris: This is needed only for this line: https://github.com/cloudscape-design/components/blob/d6b4e10cdd6684bbe9c86149d649902bca3b73af/src/column-layout/index.tsx#L21

Can be simplified by moving it down to "with grid" component

YueyingLu (approved):

PR #1124, opened 2023-05-24
feat: Add `detailPopoverFooter` property to all chart components

timogasda:
### Description
This adds a new property called `detailPopoverFooter` to all chart components. It allows customers to add custom content to the bottom of the chart detail popover. This can be used to e.g. drill down on a highlighted data segment.

Related links, issue #, if available: AWSUI-21023, `8b6OATazKD2G`

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/area-chart/test.page.tsx
+++ b/pages/area-chart/test.page.tsx
@@ -10,6 +10,7 @@ import ScreenshotArea from '../utils/screenshot-area';
 
 import Example from './example';
 import { createLinearTimeLatencyProps, createLogXYProps } from './series';
+import { Box, ButtonDropdown } from '~components';
 
 const logXProps = createLogXYProps({ xLog: true });
 const linearLatencyProps = createLinearTimeLatencyProps();
@@ -47,6 +48,19 @@ export default function () {
             highlightedSeries={highlightedSeries}
             onHighlightChange={onHighlightChange}
             xDomain={xDomain}
+            detailPopoverFooter={() => (
+              <Box margin={{ top: 'm' }}>
+                <ButtonDropdown
+                  items={[
+                    { id: '1', text: 'View' },
+                    { id: '2', text: 'Add to filter' },
+                  ]}
+                  expandToViewport={true}
+                >
+                  Actions
+                </ButtonDropdown>
+              </Box>
+            )}
           />
 
           <Example name="Log-x latency chart" {...logXProps} />
--- a/pages/area-chart/test.page.tsx
+++ b/pages/area-chart/test.page.tsx
@@ -5,12 +5,12 @@ import React, { useCallback, useState } from 'react';
 import { AreaChartProps } from '~components/area-chart';
 import { NonCancelableEventHandler } from '~components/internal/events';
 import SpaceBetween from '~components/space-between';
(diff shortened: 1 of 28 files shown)


pages/pie-chart/test.page.tsx
```diff
@@ -38,6 +39,11 @@ export default function () {
               ariaLabel="Food facts"
               size="medium"
               onHighlightChange={e => setActiveSegment(e.detail.highlightedSegment)}
+              detailPopoverFooter={segment => (
+                <Box margin={{ top: 'm' }}>
+                  <Button>Filter by {segment.title}</Button>
```
pan-kot, line 44: Why do we have different formats for pie- and other charts? Would it be safer to make the first argument always an object which can be extended with more properties in the future?

timogasda: Conceptually, we always pass the same data, it's just called differently in the components. But this is always the "x" value that is of the generic `T` type. But in pie chart we call them segments.

timogasda: I currently cannot imagine any other data that would be useful to pass down here. So I'd rather not have a one-key object paramter.

pages/area-chart/test.page.tsx
```diff
@@ -47,6 +48,19 @@ export default function () {
             highlightedSeries={highlightedSeries}
             onHighlightChange={onHighlightChange}
             xDomain={xDomain}
+            detailPopoverFooter={() => (
+              <Box margin={{ top: 'm' }}>
```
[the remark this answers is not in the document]
pan-kot: I think unless we want different spacing for different types of content we expect it would be great to have some default. The "total" can be supported separately and for something more custom I'd opt for providing a function to render the entire popover content.

pages/pie-chart/test.page.tsx
pan-kot: We can pass formatted value, series, total. Of course, we can always pass those as an extra argument if ever needed.
I don't mind keeping it as is.

pages/area-chart/test.page.tsx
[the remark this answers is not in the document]
pan-kot: Other than that, the code looks good 👍

pan-kot (dismissed):

src/mixed-line-bar-chart/chart-popover.tsx
```diff
@@ -60,6 +63,7 @@ function MixedChartPopover<T extends ChartDataTypes>(
               onMouseLeave={onMouseLeave}
             >
               <ChartSeriesDetails details={highlightDetails.details} />
+              {footer && <InternalBox margin={{ top: 's' }}>{footer}</InternalBox>}
```
pan-kot, line 66: Would it make sense to add the footer prop to the ChartPopover? This way the margin can be set once

timogasda: I've thought about it, but I think for now the content slot makes more sense. Because we're not actually introducing a formal visual "footer" concept (yet). Just like regular popovers dont really have a footer.

PR #1129, opened 2023-05-25
refactor: Use logging from toolkit

pan-kot:
### Description

Replace internal is-development and warn-once utils from those from the toolkit.

Depends on https://github.com/cloudscape-design/component-toolkit/pull/26

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/area-chart/internal.tsx
```diff
@@ -67,18 +66,15 @@ export default function InternalAreaChart<T extends AreaChartProps.DataTypes>({
   const containerRef = useRef<HTMLDivElement>(null);
   const popoverRef = useRef<HTMLDivElement>(null);
 
-  if (isDevelopment) {
```
just-boris, line 70: Why did you make this change?

`isDevelopment` is a built-time constant allowing minifiers to remove unused code when it is `false`. When it is built into `warnOnce`, extra code does not get removed.

just-boris: P.S. can you also check that the unused code gets removed properly, after it moved to a different package?

pan-kot: That's fair. Should we then ad `if (isDevelopment)` to everywhere the warnOnce is used? There is no consistency in that regard atm.

just-boris: Do your own research. The goal is to ensure that dev only code gets removed properly. 

When `warnOnce` becomes an empty function in prod, does it get removed completely? If yes, then we can keep plain `warnOnce` calls. If not, refactor is needed

pan-kot: Sounds good, I'll deep dive and share the findings.

pan-kot: The code under `isDevelopment` can be still found in the `awsui.js` bundle we generate for pages before and after the change.

just-boris: This particular snippet or anything else?

just-boris: You found a big issue. All of ESM is broken and we ship 20% extra size in the bundle.

When set up properly, it works, so we need to keep it. 

Let me know in DM if you are also interested fixing our webpack config for dev pages

pan-kot: I run some experiments. Interestingly, the code optimisation does sometimes remove code when internal isDevelopment is used and never does so when using the one from the toolkit. I say sometimes because more often than not it replaces the value of isDevelopment with `false` but keeps the code under IF statement still. It does optimise better if replacing all isDevelopment uses with `process.env.NODE_ENV !== 'production'`.

However, I wouldn't say the results are any impressive, here is what I've got (measured non gzipped size of awsui.js):

1. The current PR: `1.187.156 bytes`;
2. All warnOnce are wrapped with internal/isDevelopment: `1.187.629 bytes`;
3. All warnOnce are wrapper with process.env.NODE_ENV !== 'production': `1.179.707 bytes`.

The solution is internal/isDevelopment is worse because there is more code kept than removed.
The solution with explicit process.env is less than 1% space efficient.

just-boris: Let's keep this unchanged in the current PR and if you want to refactor this conditions, it is going to be a separate PR

pan-kot: Ok, will do so.

just-boris (approved):

PR #1130, opened 2023-05-26
chore: Prevent scroll when focusing SVG elements

timogasda:
### Description
There is currently a bug in pie chart when there are no visible labels, i.e. `hideTitles={true}` and `hideDescriptions={true}`. Clicking anywhere on the plot on Safari will cause the browser to scroll to the top. See https://github.com/cloudscape-design/components/discussions/1101.

The root cause seems to be the `focusSvgElement` element call that focused the SVG app node.

Related links, issue #, if available: AWSUI-21151

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


pan-kot (approved):

src/internal/utils/focus-svg-element.ts
```diff
@@ -2,7 +2,9 @@
 // SPDX-License-Identifier: Apache-2.0
 export default function focusSvgElement(element: SVGElement) {
   if (element.focus) {
-    element.focus();
+    // Using preventScroll here fixes a bug in Safari that causes the page to scroll to the top
+    // when hovering over a pie chart with no visible labels.
+    element.focus({ preventScroll: true });
```
just-boris, line 7: Since it is a common utility, it could be used in cases where the scroll is desired.

This is why it would be better to proxy this option via argument

just-boris: P.S. there is also an interesting else-condition which can be cleaned up

timogasda: Good point. Since we don't really need the IE11 condition anymore, we could just get rid of this utility entirely.

What do you think @just-boris @pan-kot ?

pan-kot (approved):

PR #1132, opened 2023-05-26
feat: Media container

rubencarvalho:
### Description

This is a modification of our [container](https://cloudscape.design/components/container/?tabId=playground) component, optimised for content-oriented websites and applications.

This change to the container container allows displaying a borderless media element, either on the left or top of the container.

Added a custom hook `useMedia` to update the responsive media properties (like position, width, height, and content). 

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/container/internal.tsx
```diff
@@ -100,6 +103,10 @@ export default function InternalContainer({
   // In this case we don't want the container to have sticky styles, as only the table header row will show as stuck on scroll.
   const shouldHaveStickyStyles = isSticky && !isMobile;
 
+  function getMediaStyles() {
+    return mediaPosition === 'top' ? { height: mediaHeight } : { width: mediaWidth };
```
timogasda, line 107: Why does this need to be a function? We can just say 

```
const mediaStyles = mediaPosition === 'top' ? { height: mediaHeight } : { width: mediaWidth };
```

src/container/internal.tsx
```diff
@@ -109,50 +116,65 @@ export default function InternalContainer({
         styles.root,
         styles[`variant-${variant}`],
         fitHeight && styles['fit-height'],
+        media?.content && (mediaPosition === 'top' ? styles['with-top-media'] : styles['with-side-media']),
```
timogasda, line 119: minor: I think I've counted `media?.content` five times here. Might be worth doing `const hasMedia = !!media?.content`.

That also would turn it into a proper boolean, and not `boolean | undefined`

src/container/internal.tsx
```diff
@@ -109,50 +116,65 @@ export default function InternalContainer({
...
+          styles['content-wrapper'],
+          media?.content && styles['content-wrapper-with-media'],
+          fitHeight && styles['content-wrapper-fit-height']
+        )}
       >
-        {children}
-      </div>
-      {footer && (
+        {header && (
+          <StickyHeaderContext.Provider value={{ isStuck }}>
```
timogasda, line 137: So what happens if a sticky header has media in it? 🤔

src/container/media.ts
```diff
@@ -0,0 +1,87 @@
...
+import { useEffect, useState } from 'react';
+import { Breakpoint, BREAKPOINTS_DESCENDING, matchBreakpointMapping } from '../internal/breakpoints';
+import { ContainerProps, MediaDefinition } from './interfaces';
+import { useContainerBreakpoints } from '../internal/hooks/container-queries';
+
+type MediaProps = keyof ContainerProps.Media;
+
+const MEDIA_PROPS: MediaProps[] = ['position', 'width', 'height', 'content'];
+
+const hasDefinedBreakpoint = (obj: any) => {
```
timogasda, line 13: I think we can do better than `any`. Maybe something like `obj: { breakpoint: Breakpoint }` ? Or a generic type

src/container/media.ts
```diff
@@ -0,0 +1,87 @@
...
+    position: media?.position || 'top',
+  };
+  const [mediaState, setMediaState] = useState(initialState);
+  const [breakpoint, breakpointRef] = useContainerBreakpoints(getBreakpointsForMedia(media));
+  useEffect(() => {
+    if (!media || !breakpoint) {
+      return;
+    }
+
+    setMediaState(prevMediaState => {
```
timogasda, line 31: Please consider if you really need a `useEffect` hook here: https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes.

src/container/media.ts
```diff
@@ -0,0 +1,87 @@
...
+      MEDIA_PROPS.forEach(attribute => {
+        if (!media[attribute]) {
+          return;
+        }
+
+        let value;
+
+        if (attribute === 'content') {
+          value = hasDefinedBreakpoint(media[attribute])
+            ? matchBreakpointMapping(media[attribute] as MediaDefinition.BreakpointMapping<any>, breakpoint)
```
timogasda, line 43: `as` (and `any`) are stripping away a lot of type safety here. Can we get rid of them?

src/container/media.ts
```diff
@@ -0,0 +1,87 @@
...
+      const newMediaState = { ...prevMediaState };
+
+      MEDIA_PROPS.forEach(attribute => {
+        if (!media[attribute]) {
+          return;
+        }
+
+        let value;
+
+        if (attribute === 'content') {
```
timogasda, line 41: This `if/else` is the meat of this hook. Can we please add some descriptive comments?

src/container/media.ts
```diff
@@ -0,0 +1,87 @@
...
+    return;
+  }
+
+  const breakpointKeys = new Set<Breakpoint>();
+
+  MEDIA_PROPS.forEach(key => {
+    if (media && typeof media[key] === 'object' && media[key] !== null) {
+      const breakpointMapping = media[key] as MediaDefinition.BreakpointMapping<any>;
+      Object.keys(breakpointMapping).forEach(key => {
+        const breakpoint = key as Breakpoint;
```
timogasda, line 79: Two more uses of `as` in this function. This might be an indicator that your props and/or arguments are not typed correctly.

src/container/styles.scss
```diff
@@ -61,11 +64,68 @@
...
+  &-with-media {
+    display: flex;
+    flex-direction: column;
+  }
+}
+
+.media {
+  overflow: hidden;
+
+  // stylelint-disable-next-line @cloudscape-design/no-implicit-descendant, selector-max-type
```
timogasda, line 91: Shouldn't we be using a direct descending selector (`>`) here?

Side note: Should we add some additional restrictions to the API docs or the Dev guidelines? Things like: "we only support `img`, `video`, `picture`", or "the media tag needs to be the first and only child of the `media` slot"

src/container/styles.scss
```diff
@@ -61,11 +64,68 @@
...
   border-top-left-radius: awsui.$border-radius-container;
   border-top-right-radius: awsui.$border-radius-container;
 
+  &.header-with-media {
+    background: none;
+    &:not(:empty) {
+      border-bottom: none;
+    }
+    &.content {
+      overflow: auto;
```
timogasda, line 125: Why is this needed? Why do we want containers with media to scroll their content?

src/expandable-section/__tests__/expandable-section.test.tsx
```diff
@@ -29,7 +33,7 @@ describe('Expandable Section', () => {
       expect(wrapper.findAll('button').length).toBe(0);
     });
   });
-  describe('slots', () => {
+  describe.only('slots', () => {
```
timogasda, line 36: `.only` ! 

(we really should add an eslint rule about this...)

src/expandable-section/__tests__/expandable-section.test.tsx
timogasda, line 1: Should we add some new tests?

timogasda (changes requested):

src/container/internal.tsx
rubencarvalho: This combination should not be possible for consumers of the Container.

`__stickyHeader` is a private property that we only use internally for having the sticky container throughout other components (like Cards or Table). I don't see us wanting both to coexist in these scenarios, so I think it is safe to ignore the corner case.

src/container/internal.tsx
rubencarvalho: Leftover from refactoring, will remove this!

src/container/styles.scss
rubencarvalho: I've thought about this. My thought was that there are some custom media components (like those offered by some frameworks, i.e.: `GatsbyImage` or `next/image`) that could potentially include more wrapping elements.
What do you think?

Regarding documentation, I will update! 👍

timogasda: mhm yeah, fair enough. this might lead to interesting questions down the line in terms of what we support. in case some people have weird mixed media components.

But for now I think this is fine.

src/container/internal.tsx
rubencarvalho: Nice, I updated it. And just like that... way cleaner!

Thanks!

src/container/media.ts
rubencarvalho: Indeed! Great call - no need for state even.

src/container/interfaces.ts
```diff
@@ -47,3 +50,48 @@ export interface ContainerProps extends BaseComponentProps {
    */
   variant?: 'default' | 'stacked';
 }
+
+export namespace ContainerProps {
+  /**
+   * The value for each property can either be a single value (which applies for all breakpoints) or an object where the key
+   * is one of the supported breakpoints (`xxs`, `xs`, `s`, `m`, `l`, `xl`) or `default`.
+   * The value of this key is applied for that breakpoint and those above it. You must provide a `default` value for `content`.
+   */
```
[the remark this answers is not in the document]
rubencarvalho: Good call. I'd rather show it. I've moved it down to the `content` slot.

src/container/interfaces.ts
```diff
@@ -47,3 +50,48 @@ export interface ContainerProps extends BaseComponentProps {
...
+
+    /**
+     * Defines the media slot's orientation within the container.
+     */
+    orientation: MediaDefinition.Orientation | MediaDefinition.BreakpointMapping<MediaDefinition.Orientation>;
+
+    /**
+     * Defines the width of the media slot when vertically oriented. Corresponds to the `width` CSS-property.
+     * When this value is set, the media element may be cropped, with `object-fit: cover` centering it.
+     * Note: This value is considered only when `orientation` is set to `vertical`.
```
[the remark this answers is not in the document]
rubencarvalho: Done!

src/container/internal.tsx
rubencarvalho: Done!

src/container/styles.scss
rubencarvalho: Removed!

src/container/media.ts
timogasda: Nevermind. `Object.keys` is just really useless in Typescript 😬

src/container/internal.tsx
```diff
@@ -74,7 +78,10 @@ export default function InternalContainer({
   const hasDynamicHeight = isRefresh && variant === 'full-page';
   const overlapElement = useDynamicOverlap({ disabled: !hasDynamicHeight || !__darkHeader });
 
-  const mergedRef = useMergeRefs(rootRef, subStepRef, __internalRootRef);
+  const [breakpoint, breakpointRef] = useContainerBreakpoints(BREAKPOINTS_DESCENDING.filter(bp => bp !== 'default'));
```
timogasda, line 81: Would it make sense to move the `useContainerBreakpoints` into `useMedia`?

timogasda: also, instead of listening to all breakpoints (except default), could you extract the list of breakpoints use from the `media` prop and use them instead?

timogasda (commented):
Looks great overall. Just one small suggestion

src/container/interfaces.ts
```diff
@@ -47,3 +74,47 @@ export interface ContainerProps extends BaseComponentProps {
...
+    content: React.ReactNode | MediaDefinition.ContentBreakpointMapping;
+
+    /**
+     * Defines the media slot's position within the container. Defaults to `top`.
+     */
+    position?: 'top' | 'side' | MediaDefinition.BreakpointMapping<'top' | 'side'>;
+
+    /**
+     * Defines the width of the media slot when positioned on the side. Corresponds to the `width` CSS-property.
+     * When this value is set, the media element may be cropped, with `object-fit: cover` centering it.
```
stirache, line 97: When this value is set, media elements larger than the defined width may be cropped, with 'object-fit: cover' centering it.

stirache (changes requested):

timogasda (approved):
Looks good!

PR #1135, opened 2023-05-30
chore: Add i18n support to chart components and popover

avinashbot:
### Description

First batch of post-beta components for internationalization! The actual support addition is rather small; a lot of this PR is just tests. As before, we're continuing to add the integration as close to its use as possible, so a lot of chart subcomponents were updated. There are some additional things to consider:

- Simpler unit tests using an internal-only "TestI18nProvider"
- The addition of a new `@i18n` tag provided to interfaces

### How has this been tested?

Unit testing. Generally each string is tested in unit tests. Also added a little test support provider so that we don't need to fake a locale and provide a full namespaced structure for each test case.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/area-chart/__tests__/area-chart-initial-state.test.tsx
+++ b/src/area-chart/__tests__/area-chart-initial-state.test.tsx
@@ -7,6 +7,7 @@ import AreaChart, { AreaChartProps } from '../../../lib/components/area-chart';
 import { KeyCode } from '@cloudscape-design/test-utils-core/dist/utils';
 import popoverStyles from '../../../lib/components/popover/styles.css.js';
 import { warnOnce } from '../../../lib/components/internal/logging';
+import TestI18nProvider from '../../../lib/components/internal/i18n/testing';
 import { cloneDeep } from 'lodash';
 import '../../__a11y__/to-validate-a11y';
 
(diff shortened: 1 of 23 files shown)


src/area-chart/__tests__/area-chart-initial-state.test.tsx
```diff
@@ -434,3 +435,31 @@ test('warns when data series have different x values', () => {
...
+        <AreaChart series={[areaSeries1]} statusType="finished" />
+      </TestI18nProvider>
+    );
+
+    // Show popover for the first data point.
+    wrapper.findApplication()!.focus();
+    // Pin popover.
+    wrapper.findApplication()!.keydown(KeyCode.enter);
+
+    console.log(wrapper.findDetailPopover()?.getElement().outerHTML);
```
gethinwebster, line 457: remove

src/internal/components/chart-status-container/index.tsx
```diff
@@ -54,26 +55,29 @@ export default function ChartStatusContainer({
...
       fireNonCancelableEvent(onRecoveryClick);
     };
     if (statusType === 'error') {
+      const renderedRecoveryText = i18n('recoveryText', recoveryText);
       return (
         <span>
-          <InternalStatusIndicator type="error">{errorText}</InternalStatusIndicator>{' '}
-          {recoveryText && (
+          <InternalStatusIndicator type="error">{i18n('errorText', errorText)}</InternalStatusIndicator>{' '}
+          {renderedRecoveryText && (
```
gethinwebster, line 70: Interesting: are we removing the possibility to hide the recovery text? Are there cases where there shouldn't be text here?

gethinwebster (changes requested):

src/internal/components/chart-status-container/index.tsx
avinashbot: Hm, that is true. Once you onboard to i18n, they'll be always displayed. It made sense before because we didn't want to add a blank link without any text, but thankfully we have another check: testing for `onRecoveryClick`. Adding that check as well.

gethinwebster (dismissed):

gethinwebster (approved):

PR #1142, opened 2023-05-31
feat: Add i18n support to attribute editor, tag editor

avinashbot:
### Description

It's a lot of files, but it reads quick! Adds i18n string support to a batch of related components. The strings themselves will come in a later PR, so this PR itself will not change anything for users.

Codecov is yelling at me so I'll add a couple of tests before merging, but you can still take a look in the meantime :)

### How has this been tested?

All i18n string additions are being tested with unit tests. As I'm sure you've noticed, there's a little internal test helper to make building test-specific i18n providers easy. Also, tests are most of the actual code in this PR.

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/autosuggest/interfaces.ts
```diff
@@ -75,7 +75,7 @@ export interface AutosuggestProps
   /**
    * Specifies a function that generates the custom value indicator (for example, `Use "${value}"`).
    */
-  enteredTextLabel: AutosuggestProps.EnteredTextLabel;
+  enteredTextLabel?: AutosuggestProps.EnteredTextLabel;
```
YueyingLu, line 78: The website API documentation will state `enteredTextLabel ` as `optional` in this case? But it is actually required value. As enteredTextLabel item is always shown, if the user does not set it, it will display an empty row. I am thinking maybe make it clear in the documentation, always provide it unless using i18n privider?

avinashbot: Once released, it will display on the website next to the property that it's supported by the i18n provider. So it's not entirely misleading on the website.

avinashbot: Following a similar thing on another PR, I think it's worth adding a dev warning for this though. One second...

YueyingLu (approved):

PR #1149, opened 2023-06-01
chore: Pages context settings

pan-kot:
### Description

Many pages use url-based settings and custom UI controls to alter the component/page properties for manual or automated testing (both integ- and screenshot). While the app-context already simplifies the setup, it still takes too much time to introduce custom controls especially when making quick experiments.

The context settings feature allows enhancing or replacing custom url parameters with a JSON object allowing for quicker experiments setup.

For integration tests the settings JSON property can be passed as url param to set page configuration.

The readonly mode can be used for screenshot testing for the configuration to be visible.

https://github.com/cloudscape-design/components/assets/20790937/07eba5ab-9032-4241-a19f-b7f473116b85

Compared to in-code properties adjustment the playground allows changing properties while preserving the page/component internal state:

https://github.com/cloudscape-design/components/assets/20790937/4cea0403-93aa-4dfc-9987-e0c86801a4b4

### How has this been tested?

Manual testing.

The VR tests fail due to the new header toggle: "Context settings".

<details>

#### Correctness

#### Security

#### Testing

</details>

---


timogasda (commented):
Code looks fine, and it's a fun idea :)

Should we have a chat with the rest of team before we're making this "a thing"?

pan-kot:
> Details

Sure, let's discuss!

just-boris (commented):
The current implementation only works for basic components without JSX. I think, you cannot use this playground outside button dropdown component, actually

Also, it is unclear what task you solve here. There are two existing approaches how to play with our components:

1. Run locally and make changes right in the pages code
2. Use codesandbox

Why do we need this one?

pan-kot:
> The current implementation only works for basic components without JSX. I think, you cannot use this playground outside button dropdown component, actually
> 
> Also, it is unclear what task you solve here. There are two existing approaches how to play with our components:
> 
> 1. Run locally and make changes right in the pages code
> 2. Use codesandbox
> 
> Why do we need this one?

The primary reason is to simplify creation of pages with settings that alter the component presentation so instead of including checkboxes, dropdowns etc. a JSON can be accepted instead. That is useful for experiments but can also be helpful for test pages that can use different settings coming from the url params.

The settings object does not have to be component props but any custom page configuration. Existing pages like table sticky-columns or autosuggest async test can benefit from having it as well.

pages/dropdown/focus-trap.page.tsx
```diff
@@ -39,44 +27,44 @@ export default function DropdownScenario() {
           <label>
```
pan-kot, line 27: The in-page controls can be removed provided the context settings are open during the screenshot tests for better dev experience.

src/internal/components/dropdown/__integ__/dropdown-focus-trap.test.ts
```diff
@@ -13,7 +12,8 @@ interface PageSettings {
 
 function setupTest(settings: PageSettings, testFn: (page: DropdownPageObject) => Promise<void>) {
   return useBrowser(async browser => {
-    await browser.url(`#/light/dropdown/focus-trap?${qs.stringify(settings)}`);
+    await browser.url(`#/light/dropdown/focus-trap?settings=${JSON.stringify({ ...settings, loopFocus: true })}`);
```
pan-kot, line 15: The settings object now needs to be complete, not partial. Unless the defaults handled differently in the page.

PR #1143, opened 2023-05-31
feat: Support interactive elements in Expandable section headers

jperals:
### Description

Related links, issue #, if available: AWSUI-20024

### How has this been tested?

- Added unit tests
- Added integration tests
- New permutations page for visual regression tests. Visual regression test will be added including focus states

BackstopJS tests are failing because a new page was added (`expandable-section/container-variant.permutations`) for which there is no reference.

<details>
   <summary>Review checklist</summary>

_The following items are to be evaluated by the author(s) and the reviewer(s)._

#### Correctness

- _Changes include appropriate documentation updates._
- _Changes are backward-compatible if not indicated, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#public-apis)._
- _Changes do not include unsupported browser features, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#browsers-support)._
- _Changes were manually tested for accessibility, see [accessibility guidelines](https://cloudscape.design/foundation/core-principles/accessibility/)._

#### Security

- _If the code handles URLs: all URLs are validated through [the `checkSafeUrl` function](https://github.com/cloudscape-design/components/blob/main/src/internal/utils/check-safe-url.ts)._

#### Testing

- _Changes are covered with new/existing unit tests?_
- _Changes are covered with new/existing integration tests?_
</details>

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1157, opened 2023-06-01
fix: Prevent headings from being cut off in Header component

jperals:
### Description

A wrapping area around the heading in the Header component has `overflow: hidden` which causes any bit of overflowing content to be cut off.

This is most obvious with italics text:
Screenshot 2023-06-01 at 22 19 48 copy

Screenshot 2023-06-01 at 22 19 48 copy

But even without italics, the very tip of certain characters is also sometimes cut off (screenshot diffs below):
Screenshot 2023-06-01 at 22 35 18
Screenshot 2023-06-01 at 22 38 31

This overflow rule comes from a mixin (`text-flex-wrapping`) which is also used by components such as popover or alert, where it is applied at the outermost level of the component tree, so that eventually cutting off content seems justified (the content is really at the border of the component), but this is not the case here.

This is also relevant for #1143, since without this fix the focus ring around the new clickable area is cut off:
Screenshot 2023-06-01 at 22 25 14

### How has this been tested?

Ran against my pipeline and verified the visual regression failures, which come from headings not being cut off anymore and are therefore desired changes.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/header/styles.scss
+++ b/src/header/styles.scss
@@ -45,7 +45,10 @@
 }
 
 .main {
-  @include styles.text-flex-wrapping;
+  // Like styles.text-flex-wrapping, but without overflow: hidden, to prevent headings from being partially cut off.
+  word-wrap: break-word;
+  max-width: 100%;
+
   :not(.root-no-actions) > & {
     margin-right: awsui.$space-xs;
   }


src/header/styles.scss
```diff
@@ -45,7 +45,10 @@
 }
 
 .main {
-  @include styles.text-flex-wrapping;
+  // Like styles.text-flex-wrapping, but without overflow: hidden, to prevent headings from being partially cut off.
+  word-wrap: break-word;
```
jperals, line 49: It's possible that these two rules are also not needed, especially as the child `.title` already has `word-break: break-word` which overrides this behavior. But I prefer to err on the safe side here and might consider further changes later in the scope of #1143.

abdhalees (approved):

PR #1175, opened 2023-06-05
chore: Adds resize to Drawers in VR

katiegeorge:
### Description

Resize in drawers has been available in Classic Cloudscape styles since April. This is creating feature parity in the Visual Refresh version of Drawers.

This also moves the resize helper functions (`use_keyboard_events` and `use_pointer_events`) out of Split Panel and into the AppLayout utils, since they are shared functions now between Split Panel and Drawers.

### How has this been tested?
The test page is the same: `with-drawers`

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/app-layout/__tests__/split-panel.test.tsx
```diff
@@ -283,7 +283,7 @@ test('should fire split panel resize event', () => {
     />
   );
   wrapper.findSplitPanel()!.findSlider()!.keydown(KeyCode.pageUp);
-  expect(onSplitPanelResize).toHaveBeenCalledWith({ size: 460 });
+  expect(onSplitPanelResize).toHaveBeenCalled();
```
katiegeorge, line 286: Because I changed `style.height` to `clientHeight`, this no longer grabs the size. There is an integration test that tests that the width is actually changing how we expect.

src/app-layout/visual-refresh/context.tsx
```diff
@@ -417,9 +422,25 @@ export const AppLayoutInternalsProvider = React.forwardRef(
       changeHandler: 'onChange',
     });
 
-    const activeDrawerWidth = 290;
+    const [drawersMaxWidth, setDrawersMaxWidth] = useState(toolsWidth);
 
-    const { refs: drawersRefs, setFocus: focusDrawersButtons } = useFocusControl(activeDrawerId);
+    const activeDrawer = drawers?.items.find((drawer: any) => drawer.id === activeDrawerId);
```
connorlanigan, line 427: Can we remove (or replace) the `any` here? We should be able to update [line 417 (`const drawers = ...`)](https://github.com/cloudscape-design/components/blob/5e16764b4613b934a2d00ec37f7b2bdba54c4415/src/app-layout/visual-refresh/context.tsx#L417) similar to how we have it in the classic AppLayout: https://github.com/cloudscape-design/components/blob/d5aa98e030232223a427f2f927e4ac4b91d6c162/src/app-layout/index.tsx#L132 Typescript should then infer the types in this file correctly

src/app-layout/utils/use-resize.tsx
```diff
@@ -0,0 +1,134 @@
...
+  drawers: DrawersProps;
+  drawersRefs: DrawerFocusControlRefs;
+  isToolsOpen: boolean;
+  drawersMaxWidth: number;
+}
+
+function useResize(drawerRefObject: React.RefObject<HTMLDivElement>, drawerResizeProps: DrawerResizeProps) {
+  const { activeDrawerId, drawers, drawersRefs, isToolsOpen, drawersMaxWidth } = drawerResizeProps;
+
+  const activeDrawer = drawers?.items.find((item: any) => item.id === activeDrawerId) ?? null;
```
connorlanigan, line 38: Can we remove `any` here, to increase the type safety? (see also other comment below)

src/app-layout/utils/use-resize.tsx
```diff
@@ -0,0 +1,134 @@
...
+  const getDrawerItemSizes = useCallback(() => {
+    const sizes: { [id: string]: number } = {};
+    if (!drawerItems) {
+      return {};
+    }
+
+    for (const item of drawerItems) {
+      if (item.defaultSize) {
+        if (item.defaultSize > drawersMaxWidth) {
+          sizes[item.id] = toolsWidth;
```
connorlanigan, line 52: What is the relation between `drawersMaxWidth` and `toolsWidth` - why does exceeding the former mean we use the latter? Could we potentially reduce this `if/else` block to something like:
```
sizes[item.id] = Math.min(item.defaultSize || drawersMaxWidth, drawersMaxWidth)
```

src/app-layout/utils/use-resize.tsx
```diff
@@ -0,0 +1,134 @@
...
+    // wait one frame to allow app-layout to complete its calculations
+    const handle = requestAnimationFrame(() => {
+      const maxSize = drawersMaxWidth;
+      setRelativeSize(((drawerSize - MIN_WIDTH) / (maxSize - MIN_WIDTH)) * 100);
+    });
+    return () => cancelAnimationFrame(handle);
+  }, [drawerSize, drawersMaxWidth, MIN_WIDTH]);
+
+  const drawerResize = (resizeDetail: { size: number; id: string }) => {
+    const drawerItem = drawers.items.find(({ id }: any) => id === resizeDetail.id);
```
connorlanigan, line 86: Same request about replacing/removing `any`, see other comments

src/app-layout/utils/use-resize.tsx
```diff
@@ -0,0 +1,134 @@
...
+    if (drawerItem?.onResize) {
```
connorlanigan, line 87: [`fireNonCancelableEvent` checks internally if the handler is undefined](https://github.com/cloudscape-design/components/blob/d5aa98e030232223a427f2f927e4ac4b91d6c162/src/internal/events/index.ts#L55-L57), so this condition here can be skipped

src/app-layout/utils/use-resize.tsx
```diff
@@ -0,0 +1,134 @@
...
+    }
+    setDrawerItemSizes({ ...drawerItemSizes, [resizeDetail.id]: resizeDetail.size });
+  };
+
+  const setSidePanelWidth = (width: number) => {
+    const maxWidth = drawersMaxWidth;
+    const size = getLimitedValue(MIN_WIDTH, width, maxWidth);
+    const id = activeDrawer?.id;
+
+    if (activeDrawer && id && maxWidth >= MIN_WIDTH) {
```
connorlanigan, line 98: minor: `activeDrawer` is always truthy if `id` is truthy, so it can be removed

src/app-layout/utils/use-pointer-events.ts
```diff
@@ -1,49 +1,62 @@
...
   const onDocumentPointerUp = useCallback(() => {
+    if (!panelRef || !panelRef.current) {
+      return;
+    }
     document.body.classList.remove(styles['resize-active']);
     document.body.classList.remove(styles[`resize-${position}`]);
     document.removeEventListener('pointerup', onDocumentPointerUp);
     document.removeEventListener('pointermove', onDocumentPointerMove);
-  }, [onDocumentPointerMove, position]);
+    panelRef.current.style.transitionProperty = 'border-color, opacity, width';
```
connorlanigan, line 58: Do these three properties correspond to something in a CSS file? I see a risk that the definition here might get out of sync with the CSS, but also at runtime it might get out of sync with React's virtual DOM, since imperative updating of DOM elements happens without React's knowledge. We could make this more reliable and maintainable by setting and removing a class on the DOM element instead of inline styles (e.g. something like `panelRef.current.classList.add(styles.disableTransitions)`)

connorlanigan (changes requested):

src/app-layout/utils/use-pointer-events.ts
katiegeorge: This is such a good idea! I hadn't thought of that.

src/app-layout/utils/use-resize.tsx
katiegeorge: I actually noticed a small bug in this code where the size wasn't being saved when I switched drawers. I updated it so now its much cleaner.

connorlanigan (changes requested):

connorlanigan (dismissed):

src/app-layout/__tests__/common.test.tsx
```diff
@@ -3,7 +3,14 @@
...
 import { act } from 'react-dom/test-utils';
 import { AppLayoutWrapper } from '../../../lib/components/test-utils/dom';
-import { describeEachAppLayout, isDrawerClosed, drawersConfigurations, renderComponent } from './utils';
+import {
+  describeEachAppLayout,
+  drawerWithoutLabels,
+  isDrawerClosed,
+  renderComponent,
+  singleDrawer,
+  singleDrawerOpen,
```
just-boris, line 12: Refactored all drawers configuration presets, made them typescript friendly and also removed duplicates

src/app-layout/__tests__/desktop.test.tsx
```diff
@@ -161,105 +148,94 @@ describeEachThemeAppLayout(false, () => {
...
-    const { wrapper } = renderComponent(<AppLayout contentType="form" {...resizableDrawer} />);
-
+    const { wrapper } = renderComponent(<AppLayout contentType="form" {...drawers} />);
     wrapper.findDrawersSlider()!.keydown(KeyCode.left);
-    expect(onDrawerResize).toHaveBeenCalled();
+
+    expect(onResize).toHaveBeenCalledWith({ size: expect.any(Number), id: 'security' });
+  });
+
+  test('should change size via mouse pointer on slider handle', () => {
```
just-boris, line 214: This test is what was missing in the coverage

src/app-layout/utils/use-resize.tsx
```diff
@@ -72,8 +72,9 @@ function useResize(drawerRefObject: React.RefObject<HTMLDivElement>, drawerResiz
   }, [drawerSize, drawersMaxWidth, MIN_WIDTH]);
 
   const drawerResize = (resizeDetail: { size: number; id: string }) => {
-    const drawerItem = drawers?.items.find((item: DrawerItemProps) => item.id === resizeDetail.id);
+    const drawerItem = drawers?.items.find(item => item.id === resizeDetail.id);
     fireNonCancelableEvent(drawerItem?.onResize, resizeDetail);
+    fireNonCancelableEvent(drawers?.onResize, resizeDetail);
```
just-boris, line 77: 🐛 Found the actual bug! There are two `onResize` callbacks, one on the drawer, the other one on the `drawers` object directly.

It was working in Classic, but not in Refresh. Fixed

src/app-layout/visual-refresh/context.tsx
```diff
@@ -14,7 +14,6 @@ import { applyDefaults } from '../defaults';
 import { AppLayoutContext } from '../../internal/context/app-layout-context';
 import { DynamicOverlapContext } from '../../internal/context/dynamic-overlap-context';
 import { AppLayoutProps } from '../interfaces';
-import { DrawerItemProps, DrawersProps, InternalDrawersProps } from './drawers';
```
just-boris: There was a duplicate type between refresh and classic. It also contributed to the issue with `drawers?.onResize` above.

Removed duplicate types, used types from classic in both. (It would be a good next step to move the types into `src/app-layout/interfaces.ts` file)

katiegeorge: Yeah, that is my plan after this. I'll consolidate as much of the code as possible when bringing it out of beta.

src/app-layout/styles.scss
```diff
@@ -7,6 +7,30 @@
...
+    // Prevents iframes from eating the mouseup event if the cursor is on top of it.
+    pointer-events: none;
+  }
+  &-side * {
+    cursor: ew-resize;
+  }
+  &-bottom * {
+    cursor: ns-resize;
+  }
+}
```
just-boris, line 25: There should be `/* stylelint-enable selector-combinator-disallowed-list, selector-max-type, selector-max-universal */` at the end of the section to finish the exceptional part

src/app-layout/__tests__/desktop.test.tsx
```diff
@@ -132,42 +148,147 @@ describeEachThemeAppLayout(false, () => {
...
+      drawers: {
+        onResize: ({ detail }) => onResize(detail),
+        activeDrawerId: 'security',
+        items: resizableDrawer.drawers.items,
+      },
+    };
+    const { wrapper } = renderComponent(<AppLayout contentType="form" {...drawersOpen} />);
+    wrapper.findDrawersSlider()!.fireEvent(new MouseEvent('pointerdown', { bubbles: true }));
+    const resizeEvent = new MouseEvent('pointermove', { bubbles: true });
+    // Object.defineProperty(resizeEvent, 'clientX', { value: -400 });
```
just-boris, line 226: Oops, I left some temp code here. I will push extra changes to fix it

just-boris (approved):

PR #1184, opened 2023-06-07
chore: Add findExpandButton to Expandable section test utils

jperals:
### Description

After #1143, the clickable area to expand and collapse expandable sections is restricted to a subset of the header elements when interactive elements are present in the header. Therefore, clicking on the entire header with `ExpandableSection.findHeader().click()` is not reliable anymore to expand an expandable section in tests, and a new function `findExpandButton` is provided in test utils for this purpose, which returns only the clickable area.

Related documents:
- API proposal: kSXNASqTs3Sk
- More information about changes in the clickable area: 7SZJAvbhbznX

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/expandable-section/__tests__/interactions.test.tsx
+++ b/src/expandable-section/__tests__/interactions.test.tsx
@@ -23,15 +23,15 @@ describe('Expandable Section - Interactions', () => {
       wrapper = renderExpandableSection({ onChange: onChangeSpy });
     });
     test('toggles from false to true when header is clicked, change event fires', () => {
-      wrapper.findHeader().click();
+      wrapper.findExpandButton().click();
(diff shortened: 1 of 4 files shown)


src/expandable-section/__tests__/interactions.test.tsx
```diff
@@ -23,15 +23,15 @@ describe('Expandable Section - Interactions', () => {
       wrapper = renderExpandableSection({ onChange: onChangeSpy });
     });
     test('toggles from false to true when header is clicked, change event fires', () => {
-      wrapper.findHeader().click();
```
jperals, line 26: This would still work today because when no interactive elements are present the clickable area is not restricted, but it might change in the future, therefore the code in the right is the new recommended way.

michaeldowseza (approved):

PR #1189, opened 2023-06-07
chore: Support much stricter typechecking for internal i18n provider usage

avinashbot:
### Description

Support for i18n strings in components usually looks like this:

```js
const i18n = useInternalI18n('component-name');
const myAriaLabel = i18n('myAriaLabel', ariaLabelFromProps);
const myFunction = i18n('myFunction', functionFromProps, format => (arg) => format({ argName: arg }));
```

The keys are determined by checking the provided messages object using string matching. This isn't perfect, of course, because this means typos can sneak in, or some format function arguments might be missed. To avoid unnecessarily overcomplicating the internationalization logic, I started off with simple unit tests as the validation mechanism with the hope that they would be good enough to catch these mistakes. But #1179 showed that this isn't enough, with no enforcement that all strings are checked in tests, and test strings not potentially matching the variables used by real non-test strings. So we need to go deeper 🕳️

The internal package that generates these strings now generates a TypeScript file with a fixed mapping of component → message keys → format string arguments. This generated file can now be used to enforce that message keys exist or that format strings arguments are provided. And it did manage to catch a few unnecessary/incorrect strings, so it does work. Now you get live IDE warnings like this:

```jsx
i18n('dismissAraiLabel', dismissAriaLabel)
// Argument of type '"dismissAraiLabel"' is not assignable to parameter of type '"dismissAriaLabel"'.

i18n('i18nStrings.tagLimit', i18nStrings?.tagLimit?.(availableTags, tagLimit), format =>
  format({ tagLimitAvailable: `${remainingTags === tagLimit}`, availableTags })
)
// Property '"tagLimit"' is missing in type '{ tagLimitAvailable: string; availableTags: number; }'
//   but required in type '{ tagLimitAvailable: string; availableTags: number; tagLimit: string | number; }'.
```

Now, it means we have an autogenerated TypeScript file we need to keep up to date along with the strings, but usually both can be updated together. And also, I now feel the benefits outweigh the negatives.

### How has this been tested?

The integration actually pointed out a couple of bugs, so this _is_ the test. The "logic" is mostly just TypeScript here.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/i18n/__tests__/test-component.tsx
+++ b/src/i18n/__tests__/test-component.tsx
@@ -28,7 +28,7 @@ export const MESSAGES: I18nProviderProps.Messages = {
 };
 
 declare module '../../../lib/components/i18n/messages-types' {
-  interface CloudscapeI18nFormatArgTypes {
+  interface I18nFormatArgTypes {
     'test-component': {
       topLevelString: never;
       topLevelFunction: { type: string };
--- a/src/i18n/context.ts
+++ b/src/i18n/context.ts
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import React, { useContext } from 'react';
-import { CloudscapeI18nFormatArgTypes } from './messages-types';
+import { I18nFormatArgTypes } from './messages-types';
 
 export type CustomHandler<ReturnValue, FormatFnArgs> = (formatFn: (args: FormatFnArgs) => string) => ReturnValue;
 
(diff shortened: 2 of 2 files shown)


src/wizard/__tests__/wizard.test.tsx
```diff
@@ -500,3 +501,44 @@ describe('Custom actions', () => {
...
+    );
+    const wrapper = createWrapper(container).findWizard()!;
+    expect(wrapper.find('li:nth-child(1)')!.getElement()).toHaveTextContent(
+      'Custom step 1 - Custom optional' + 'Step 1'
+    );
+    expect(wrapper.getElement()).toHaveTextContent('Custom step 1 of 3');
+    expect(wrapper.findCancelButton().getElement()).toHaveTextContent('Custom cancel');
+    expect(wrapper.findPrimaryButton().getElement()).toHaveTextContent('Custom next');
+    expect(wrapper.findSkipToButton()!.getElement()).toHaveTextContent('Custom skip to Step 3');
+    wrapper.findPrimaryButton().click();
```
connorlanigan, line 541: I'd recommend to wrap this in `act`

connorlanigan (changes requested):

src/wizard/__tests__/wizard.test.tsx
avinashbot: Wrapping interactions in `act` generally makes sense; but wizard updates seem to work fine without it, and we have test cases in the rest of the file that don't use it. So I don't want to pepper in an unexplained unnecessary `act` here.

just-boris: I would recommend not wrapping anything into act, because test utils already do this for you: https://github.com/cloudscape-design/test-utils/blob/main/packages/core/src/dom.ts#L65-L67

The only valid use-case for act on our end is fast-forwarding timeouts: `act(() => new Promise(resolve => setTimeout(resolve))`

avinashbot: Forgot we do that now! Resolving this comment then.

connorlanigan (approved):

PR #1190, opened 2023-06-07
chore: Measure bundle size in pull requests

connorlanigan:
### Description

This PR adds a new Github Actions workflow, and an essentially empty Vite project that the workflow uses. The workflow measures the Javascript bundle size of the Cloudscape package for the PR's branch and the `main` branch. It then attaches a status message in the Checks section of the PR that describes the new Javascript bundle size, along with information on how the Javascript bundle size changed in this PR.

screenshot of status message reporting a bundle size of 249 KB

As a high-level overview, the workflow works like this:
1. We build the essentially empty Vite project. This generates a Javascript bundle file. We record the size of that file as the __baseline__ for the later calculations.
2. We checkout the code of __the base branch__ (i.e. `main`, usually), and build the components package from there. We add these components to the Vite project and build it. We record the size of the Javascript bundle file again.
3. We checkout the code of __the pull request branch__, and build the components package from there. We add these components to the Vite project and build it. We record the size of the Javascript bundle file again.
4. We subtract the baseline size (from step 1) from each of the sizes from step 2 and 3, so that we only get the increase that is caused by Cloudscape.
5. We format the values and attach them to the latest commit of the pull request.

#### Caching

The workflow also caches the results of step 1 and step 2. These caches are scoped to the pull request (this is [a Github restriction](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache)), so they do _not_ persist across pull requests. If we wanted to cache across pull requests, we would have to perform step 1 and step 2 on the main branch regularly, but this would save us only about 2 minutes of time per PR.
The caching inside the pull request is mainly useful when you add further commits to an existing PR. In that case, only step 3 will be built again.

### How has this been tested?

Tested in a throwaway PR. The result can also be seen in this PR in the Checks section.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/.github/workflows/bundle-size.yml
+++ b/.github/workflows/bundle-size.yml
@@ -1,0 +1,170 @@
+name: Measure bundle size
+
+on:
+  pull_request:
+    branches:
+      - main
+
+permissions:
+  statuses: write
+
+jobs:
+  measure:
+    runs-on: ubuntu-latest
+    steps:
+      # ------- Setup -----------------------------------------------------
+      - uses: actions/github-script@v6
+        name: Create a pending status on the commit
+        with:
+          script: |
+            await github.rest.repos.createCommitStatus({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              sha: context.payload.pull_request.head.sha,
+              context: "Bundle size",
+              state: 'pending',
+              target_url: `https://github.com/${context.repo.owner}/${context.repo.repo}/actions/runs/${context.runId}`
+            })
+
+      - name: Checkout the PR branch
+        uses: actions/checkout@v3
+        with:
+          path: 'components'
+
+      - name: Setup the bundle measurement project
+        run: |
+          cp -r components/.github/workflows/bundle-size bundle-size
+          cd bundle-size
+          npm i --force
+
+      # ------- Baseline size measurement ---------------------------------
+
+      - name: Restore the cache for the baseline size measurement
+        uses: actions/cache/restore@v3
+        id: restore-cache-baseline
+        with:
+          path: bundle-size/output-baseline.txt
+          key: bundlesize-${{ hashFiles('bundle-size/build.js','bundle-size/index.html','bundle-size/main.jsx','bundle-size/package-lock.json','bundle-size/package.json') }}
+
+      - name: Measure the baseline size of the Vite project
+        if: steps.restore-cache-baseline.outputs.cache-hit != 'true'
+        run: |
+          cd bundle-size
+          npm run build
+          mv output.txt output-baseline.txt
+          cat output-baseline.txt
+
+      - name: Cache the baseline size
+        uses: actions/cache/save@v3
+        if: steps.restore-cache-baseline.outputs.cache-hit != 'true'
+        with:
+          path: bundle-size/output-baseline.txt
+          key: ${{ steps.restore-cache-baseline.outputs.cache-primary-key }}
+
+      # ------- Base branch size measurement ------------------------------
+
+      - name: Checkout the base branch
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.base.sha }}
+          path: 'components'
+
+      - name: Restore the cache for the base branch size measurement
+        uses: actions/cache/restore@v3
+        id: restore-cache-basebranch
+        with:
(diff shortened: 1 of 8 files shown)


YueyingLu (approved):
LGTM

.github/workflows/bundle-size/package-lock.json
```diff
@@ -0,0 +1,2789 @@
...
+      }
+    },
+    "@cloudscape-design/collection-hooks": {
+      "version": "1.0.21",
+      "resolved": "https://registry.npmjs.org/@cloudscape-design/collection-hooks/-/collection-hooks-1.0.21.tgz",
+      "integrity": "sha512-H2svaXsqHK/XX2CAgU0hxIqiiAh3rvlG490DpNHGbQ8ouXCURc8Hz0Hws+DbZ/etEczpyW44uBBO6xfIXE41iQ==",
+      "requires": {}
+    },
+    "@cloudscape-design/component-toolkit": {
+      "version": "1.0.0-beta.14",
```
just-boris, line 1946: This line looks very concerning. How much are we sure that it tests against the latest build and not this one? Same for all other our packages

.github/workflows/bundle-size.yml
```diff
@@ -0,0 +1,170 @@
...
+          cat output-baseline.txt
+
+      - name: Cache the baseline size
+        uses: actions/cache/save@v3
+        if: steps.restore-cache-baseline.outputs.cache-hit != 'true'
+        with:
+          path: bundle-size/output-baseline.txt
+          key: ${{ steps.restore-cache-baseline.outputs.cache-primary-key }}
+
+      # ------- Base branch size measurement ------------------------------
```
just-boris, line 64: Instead of building it from scratch, can we have a separate workflow storing the baseline size when building the mainline and then fetch this info with github API?


For example, board components fetch reference screenshots this way:

https://github.com/cloudscape-design/board-components/blob/2c590d64f2f688a057ef20bfb6fa51854777956d/.github/workflows/visual-regression.yml#L26-L33

.github/workflows/bundle-size.yml
```diff
@@ -0,0 +1,170 @@
...
+
+      - name: Measure the bundle size of the base branch
+        if: steps.restore-cache-basebranch.outputs.cache-hit != 'true'
+        run: |
+          cd components
+          npm i --force
+          npm run quick-build
+
+          cd ../bundle-size
+          rm -rf node_modules/@cloudscape-design/components
```
just-boris, line 87: What if there are nested dependencies? Very often, `tslib` and other our 3rd party deps appear to be nested there

.github/workflows/bundle-size.yml
```diff
@@ -0,0 +1,170 @@
...
+          cp main-with-cloudscape.jsx main.jsx
+          npm run build
+
+          mv output.txt output-pr.txt
+          cat output-pr.txt
+
+      - name: Update the commit status with calculated results
+        uses: actions/github-script@v6
+        with:
+          script: |
```
just-boris, line 127: That much of JS inline looks hard to read and maintain. Also there is a risk of injections (see our past issues on https://github.com/cloudscape-design/.github/pull/29).

I am on the fence about "Create a pending status on the commit" steps (short and might be okay). But the long script definitely deserves a separate file

.github/workflows/bundle-size/build.js
```diff
@@ -0,0 +1,41 @@
...
+  try {
+    unlinkSync('output.txt');
+  } catch (e) {}
+
+  const result = await build({
+    plugins: [react()],
+    logLevel: 'silent',
+  });
+
+  const jsFile = result.output.find(item => item.fileName.includes('index') && item.fileName.endsWith('.js'));
```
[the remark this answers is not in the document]
connorlanigan: The build here produces only three files, they are stored in `dist/`. The files are `index.html`, `assets/index-{a hash here}.css` and `assets/index-{a hash here}.js`. Vite seems to produce them in memory first, then [write them to disk](https://github.com/rollup/rollup/blob/3437c89d50ddb0e879fd46860094a37fd53078a8/src/rollup/rollup.ts#L273), and makes that memory available to us here as a return value. None of the files are larger than a megabyte, so I'm not worried about memory usage

PR #1199, opened 2023-06-09
chore: Remove dependency on react-focus-lock

avinashbot:
### Description

"Mom, can we use react-focus-lock?" "We have react-focus-lock at home"

Over the past year or so, we've mostly switched from react-focus-lock to using our own internal tab trap that's simpler to coordinate, and far less aggressive when it comes to focus management. This PR removes the final two components that still rely on the original focus lock and removes the dependency altogether. But I also couldn't resist a little refactor - moving things around so they read easier, mostly.

Related links, issue #, if available: AWSUI-18931

### How has this been tested?

Added a new focus lock test for the return backoff logic thing, but other than that, existing tests should pass.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/components/focus-lock/index.tsx
+++ b/src/internal/components/focus-lock/index.tsx
@@ -25,26 +25,8 @@ function FocusLock(
   { className, disabled, autoFocus, restoreFocus, children }: FocusLockProps,
   ref: React.Ref<FocusLockRef>
 ) {
-  useImperativeHandle(ref, () => {
-    return {
-      focusFirst,
-    };
-  });
-
   const returnFocusToRef = useRef<HTMLOrSVGElement | null>(null);
   const containerRef = useRef<HTMLDivElement | null>(null);
-
-  // Using a callback ref to detect component unmounts, which is safer than using useEffect.
-  const restoreFocusHandler = useCallback(
-    (elem: HTMLDivElement | null) => {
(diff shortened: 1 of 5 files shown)


src/internal/components/focus-lock/index.tsx
```diff
@@ -25,27 +25,9 @@ function FocusLock(
   { className, disabled, autoFocus, restoreFocus, children }: FocusLockProps,
   ref: React.Ref<FocusLockRef>
 ) {
-  useImperativeHandle(ref, () => {
```
avinashbot, line 28: These two "deleted" chunks were just moved to the bottom.

src/top-navigation/parts/overflow-menu/views/submenu.tsx
```diff
@@ -33,7 +33,7 @@ const SubmenuView = ({
   const headerId = useUniqueId('overflow-menu-header');
 
   return (
-    <FocusLock returnFocus={true}>
+    <FocusLock autoFocus={true}>
```
avinashbot, line 36: There is already a handler that returns focus to the trigger defined on the top navigation component. `returnFocus` wouldn't be great here because the component hotswaps two focus locks around.

timogasda (approved):
Best PR description. Ship it.

PR #1204, opened 2023-06-12
fix: Don't use dark mode for print media

gethinwebster:
### Description

"Print" typically prints on a white background (and removes dark backgrounds),
so in general it's better to use light mode for this media.

Related links, issue #, if available: https://github.com/cloudscape-design/theming-core/pull/47

### How has this been tested?

Tested in dev pipeline

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/hooks/use-visual-mode/index.ts
+++ b/src/internal/hooks/use-visual-mode/index.ts
@@ -8,6 +8,9 @@ import { useMutationObserver } from '../use-mutation-observer';
 import { isDevelopment } from '../../is-development';
 import { warnOnce } from '../../logging';
 
+// Note that this hook doesn't take into consideration @media print (unlike the dark mode CSS),
+// due to challenges with cross-browser implementations of media/print state change listeners.
+// This means that components using this hook will render in dark mode even when printing.
 export function useCurrentMode(elementRef: React.RefObject<HTMLElement>) {
   const [value, setValue] = useState<'light' | 'dark'>('light');
   useMutationObserver(elementRef, node => {
--- a/src/internal/styles/utils/theming.scss
+++ b/src/internal/styles/utils/theming.scss
@@ -3,9 +3,11 @@
  SPDX-License-Identifier: Apache-2.0
 */
 @mixin dark-mode-only($selector: '') {
-  /* stylelint-disable selector-combinator-disallowed-list, selector-class-pattern */
-  :global(#{$selector}.awsui-polaris-dark-mode) &,
-  :global(#{$selector}.awsui-dark-mode) & {
-    @content;
+  @media not print {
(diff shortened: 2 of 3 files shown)


src/internal/hooks/use-visual-mode/index.ts
```diff
@@ -11,10 +11,11 @@ import { warnOnce } from '../../logging';
...
   useMutationObserver(elementRef, node => {
-    const darkModeParent = findUpUntil(
-      node,
-      node => node.classList.contains('awsui-polaris-dark-mode') || node.classList.contains('awsui-dark-mode')
-    );
+    const darkModeParent =
+      findUpUntil(
+        node,
+        node => node.classList.contains('awsui-polaris-dark-mode') || node.classList.contains('awsui-dark-mode')
+      ) && !window?.matchMedia?.('print').matches;
```
just-boris, line 18: How about we will extract this as a separate function and reuse in `src/code-editor/util.ts`? It seems like it is now complex enough to not be duplicated

Also, can we write an extra test-case here? https://github.com/cloudscape-design/components/blob/main/src/internal/hooks/use-visual-mode/__tests__/use-visual-mode.test.tsx

src/internal/hooks/use-visual-mode/index.ts
```diff
@@ -8,6 +8,9 @@ import { useMutationObserver } from '../use-mutation-observer';
 import { isDevelopment } from '../../is-development';
 import { warnOnce } from '../../logging';
 
+// Note that this hook doesn't take into consideration @media print (unlike the dark mode CSS),
```
jperals, line 11: What was wrong with using `!window?.matchMedia?.('print').matches`?

gethinwebster: `window.matchMedia('print').matches` changes briefly to `true` and then immediately back to `false` when opening the print dialog, which doesn't give enough time for React to re-render.

`window.addEventListener('beforeprint')` (and `afterprint`) works better in some browsers but not consistently.

In general, the amount of & complexity of code that looked like it would be required for this seemed to outweigh the benefits of the usecase (it impacts CodeEditor and dropdown elements rendered in portals)

jperals: I see, but what is the scope of the fix now? The `dark-mode-only` mixin is only used for icons in the Code editor component, and I am no able to find any difference by manually testing in Chrome (nor in the Code editor component pages nor elsewhere).

gethinwebster: All CSS-based dark mode styling should be removed when printing. However, I think the PR deployment isn't showing this up because the supporting change (https://github.com/cloudscape-design/theming-core/commit/5febd2e) has not yet made it through our internal pipelines

jperals (approved):

PR #1213, opened 2023-06-14
fix: Use more standard aria-labels/descriptions in expandable section

gethinwebster:
### Description

Previously we had a workaround for VO+Safari that was causing problems
for other users (as the label & description were combined into one). This 
PR uses a more standard label/description approach.

Related links, issue #, if available: AWSUI-21215

### How has this been tested?

New unit tests, local testing in VO+Chrome, NVDA+FF

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/expandable-section/__tests__/expandable-section.test.tsx
+++ b/src/expandable-section/__tests__/expandable-section.test.tsx
@@ -137,11 +137,17 @@ describe('Expandable Section', () => {
 
   describe('a11y', () => {
     test('content region is labelled by header', () => {
-      const wrapper = renderExpandableSection();
-      const header = wrapper.findHeader().getElement();
+      const wrapper = renderExpandableSection({
+        variant: 'container',
+        headerText: 'Container header',
+        headerDescription: 'Container description',
+      });
+      const header = wrapper.findExpandButton().getElement();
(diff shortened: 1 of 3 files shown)


src/expandable-section/__tests__/expandable-section.test.tsx
```diff
@@ -323,16 +329,22 @@ describe('Variant container with headerText', () => {
...
+      headerText: 'Header component',
+    });
+    const headerButton = wrapper.findHeader().find('[role="button"]')!.getElement();
+    expect(headerButton).toHaveAccessibleName('Header component');
+  });
+  test('set aria description if description present', () => {
     const wrapper = renderExpandableSection({
       variant: 'container',
       headerText: 'Header component',
-      headerCounter: '(5)',
```
Al-Dani, line 330: do we now completely exclude counter from the announcement?

gethinwebster: From the button, yes, but given that the counter is no longer (since https://github.com/cloudscape-design/components/pull/1143) actually part of the clickable button area I think this makes more sense. It's still part of the heading's accessible name

Al-Dani (approved):

PR #1215, opened 2023-06-15
fix: Make inline editing focus triggers only once in case a re-render…

taheramr:
… happens

### Description

In React Strict Mode the inline-editing cell re-renders twice and the `useEffectOnUpdate` is triggered with the second re-render making the inline-editing cell get focused because of the double rendering.

This should not happen in production but it is still a bug.

AWSUI-21262

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/table/body-cell/index.tsx
+++ b/src/table/body-cell/index.tsx
@@ -2,8 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 import clsx from 'clsx';
 import styles from './styles.css.js';
-import React, { useRef, useState } from 'react';
-import { useEffectOnUpdate } from '../../internal/hooks/use-effect-on-update';
+import React, { useEffect, useRef, useState } from 'react';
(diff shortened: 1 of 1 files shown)


src/table/body-cell/index.tsx
```diff
@@ -44,13 +43,14 @@ function TableCellEditable<ItemType>({
   const tdNativeAttributes = {
     'data-inline-editing-active': isEditing.toString(),
   };
+  const isFocusMoveNeededRef = useRef(false);
```
jperals, line 46: Is this ref really necessary, or would replacing the `useEffectOnUpdate` with `useEffect` be enough?

jperals:
I think an integration test would make sense here.

jperals (approved):
I don't want to block this fix but I still think that an integration test for it would be good to have.

PR #1221, opened 2023-06-16
chore: Add check to ensure token value is string and add test

rubencarvalho:
### Description

In trying to construct a `PropertyFilterOperatorExtended<string[]>` in a PropertyFilter, a token value type of `string[]` doesn’t actually work, but not until you try to edit. At token edit time, there’s a util that tries to manipulate the value as a string that fails (trying to call `toLowerCase()` on it). This wasn't caught by TypeScript because we don't have an ESLint rule for checking unsafe calls of `any` typed values.

This PR makes the handling logic in the util more robust by checking the type first. 

Related links, issue #, if available: AWSUI-21107

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/property-filter/utils.ts
+++ b/src/property-filter/utils.ts
@@ -69,7 +69,10 @@ export function matchTokenValue(token: Token, filteringOptions: readonly Interna
       // exact match found: return it
       return { ...token, value: option.value };
     }
-    if (token.value.toLowerCase() === (option.label ?? option.value ?? '').toLowerCase()) {
+    if (
+      typeof token.value === 'string' &&
+      token.value.toLowerCase() === (option.label ?? option.value ?? '').toLowerCase()
+    ) {
       // non-exact match: save and keep running in case exact match found later
       bestMatch.value = option.value;
     }
--- a/src/property-filter/utils.ts
+++ b/src/property-filter/utils.ts
@@ -69,6 +69,9 @@ export function matchTokenValue(token: Token, filteringOptions: readonly Interna
       // exact match found: return it
       return { ...token, value: option.value };
     }
+
+    // By default, the token value is a string, but when a custom property is used,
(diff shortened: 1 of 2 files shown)


src/property-filter/utils.ts
```diff
@@ -69,7 +69,10 @@ export function matchTokenValue(token: Token, filteringOptions: readonly Interna
       // exact match found: return it
       return { ...token, value: option.value };
     }
-    if (token.value.toLowerCase() === (option.label ?? option.value ?? '').toLowerCase()) {
+    if (
+      typeof token.value === 'string' &&
```
timogasda, line 73: If we have to make this type check, doesn't that mean that our Typescript types are inaccurate? Can we fix that?

rubencarvalho: There is an ESLint rule that can detect this.

It currently shows:
25 reports of such “[unsafe calls](https://typescript-eslint.io/rules/no-unsafe-call/)”
61 reports of “[unsafe assignments](https://typescript-eslint.io/rules/no-unsafe-assignment/)” to variables
43 reports of “[unsafe arguments](https://typescript-eslint.io/rules/no-unsafe-argument/)” to functions
and 14 reports of “[unsafe returns](https://typescript-eslint.io/rules/no-unsafe-return)” in functions

We have this task as a internal BRF ticket.

rubencarvalho: So... looking at the type definition for the `token.value`, coming from the [collection hooks package](https://github.com/cloudscape-design/collection-hooks/blob/main/src/interfaces.ts#L144), it has to be `any`: by default, the token value is a `string` - however, when a custom property is used, the token value can be `any`.

I added a comment here to also make it more clear.

timogasda: Got it, thank you!

PR #1227, opened 2023-06-19
feat: adding variant colors to badge

32teeth:
### Description

This pull request extends the Badge component in the "cloudscape-design/components" repository by adding new color variants, including `critical`, `high`, `medium`, `low`, and `neutral`. These new variants provide users with more options to customize the appearance of the Badge component according to different severity levels or states.

### How has this been tested?

The changes have been tested using the following steps:

1. Set up a development environment for the "cloudscape-design/components" repository.
2. Create a test scenario for each new color variant (`critical`, `high`, `medium`, `low`, `neutral`).
3. Integrate the Badge component with different elements and use the new color variants.
4. Verify that the Badge component renders correctly with the expected colors for each variant.
5. Ensure that the changes do not impact the functionality or appearance of the existing Badge component.

### Screenshots
| Design Spec | Running demo assets |
|--|--|
| Screen Shot 2023-06-19 at 11 33 27| Screen Shot 2023-06-19 at 11 43 14|

#### Correctness

#### Security

#### Testing

</details>

---


jperals:
Thanks for your contribution. We plan to work on this in the mid-term, but this needs some further study especially because of accessibility issues. Since Badge is a relatively simple component, for the time being we would suggest that you implement your own custom badges using our existing design tokens (which you already used in this PR).

32teeth:
> study especially because of accessibility issues
@jperals wonder if you can help me understand better

> suggest that you implement your own custom badges using our existing design tokens
that is exactly what this pr does. wondering if you could recommend a solution that avoids using `className` since CloudScape disallows that

thanks, look forward to your response

jperals:
We have to make sure that changes are useful not just for a particular case but for as many use cases as possible, and think ahead of all the potential issues caused by the ways a new feature can be used, that's why we go through a process before going ahead with changes like this. For example, in this case:

- Since not all colors have enough contrast with each other, if multiple severity badges are going to be used together, we might have to question this severity palette
- Using badge color alone to communicate is not accessible, so we need to think how to prevent or at least discourage that.

(_Note_: the question is not so much whether your team in particular will use the feature in the ways pointed out above, but if anyone might do that)

What I meant by implementing your own custom badges, is creating a custom component based on the [existing Badge code](https://github.com/cloudscape-design/components/tree/main/src/badge); as you can see it is quite simple as it is essentially static content with [some styling](https://github.com/cloudscape-design/components/blob/main/src/badge/styles.scss).

This said, this PR can be a useful starting point once we get to tackle this issue.

PR #1238, opened 2023-06-21
feat: Support descriptions in Expandable section headers for footer variant

jperals:
### Description

After adding description for default variant (#1234), adding it to the footer variant too was a quite trivial change so we decided to follow up with it.

### How has this been tested?

- Adapted unit tests
- Added unit tests to make sure `findExpandButton` works in all variants and with or without descriptions
- Added permutations to existing permutations page
- Manually reviewed permutations page
- Verified that BackstopJS failures are due to the newly added permutations and look as expected

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/expandable-section/expandable-section-header.tsx
+++ b/src/expandable-section/expandable-section-header.tsx
@@ -244,11 +244,8 @@ export const ExpandableSectionHeader = ({
     );
   }
 
-  if (headerDescription && variant !== 'container' && variant !== 'default' && isDevelopment) {
-    warnOnce(
-      componentName,
-      'The `headerDescription` prop is only supported for the "default" and "container" variants.'
-    );
+  if (headerDescription && !variantSupportsDescription(variant) && isDevelopment) {
+    warnOnce(componentName, 'The `headerDescription` prop is not supported for this variant.');
   }
 
   const wrapperClassName = clsx(styles.wrapper, styles[`wrapper-${variant}`], expanded && styles['wrapper-expanded']);
--- a/src/expandable-section/expandable-section-header.tsx
+++ b/src/expandable-section/expandable-section-header.tsx
@@ -245,7 +245,7 @@ export const ExpandableSectionHeader = ({
   }
 
   if (headerDescription && !variantSupportsDescription(variant) && isDevelopment) {
-    warnOnce(componentName, 'The `headerDescription` prop is not supported for this variant.');
+    warnOnce(componentName, `The \`headerDescription\` prop is not supported for the ${variant} variant.`);
   }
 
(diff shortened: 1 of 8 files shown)


src/expandable-section/expandable-section-header.tsx
```diff
@@ -131,26 +131,26 @@ const ExpandableHeaderTextWrapper = ({
...
     </span>
   );
 
+  // If interactive elements are present, constrain the clickable area to only the icon and the header text
+  // to prevent nesting interactive elements.
+  const headerButtonListeners = hasInteractiveElements ? listeners : undefined;
+  // For the default and footer variants with description,
+  // include also the immediate wrapper around it to include the entire row for backwards compatibility,
+  // but exclude the description.
+  const headingTagListeners = !headerButtonListeners && !isContainer && description ? listeners : undefined;
```
jperals, line 150: Moving the listeners from the outer wrapper to the inner HeadingTag for the footer variant would break the Side navigation component which uses the footer variant Expandable section internally, because of its test-util selector. Taking a small conservative step back and move the listeners only in the case that a description is present.

Just in case, revamped tests (see above) to make sure that the now recommended `findExpandButton` test-util function works in all expected combinations.

src/expandable-section/expandable-section-header.tsx
```diff
@@ -242,11 +244,8 @@ export const ExpandableSectionHeader = ({
     );
   }
 
-  if (headerDescription && variant !== 'container' && variant !== 'default' && isDevelopment) {
-    warnOnce(
-      componentName,
-      'The `headerDescription` prop is only supported for the "default" and "container" variants.'
-    );
+  if (headerDescription && !variantSupportsDescription(variant) && isDevelopment) {
+    warnOnce(componentName, 'The `headerDescription` prop is not supported for this variant.');
```
YueyingLu, line 248: Can we output the variant name ? 
The headerDescription prop is not supported for ${variant} variant.

jperals: 👍 Done.

YueyingLu:
Just notice that the headerDescription API description needs to be updated. It is still writing `Use with the container variant.`

jperals:
Good point, and this is already incorrect now since descriptions are supported for default variant since #1234. I think I'll change it to:
> Use with the container, default or footer variants.

Now it's only the navigation variant that doesn't support descriptions, and I think it's unlikely that anyone will try to use descriptions with it, but just in case and for correctness, I prefer to keep this sentence and update it than just remove it.

YueyingLu (approved):

PR #998, opened 2023-04-24
chore: Move async-store to internal utils folder

rubencarvalho:
### Description

This PR moves the already abstracted async-store to the shared, internal utils folder. This makes it reusable in other components, in preparation for the sticky columns feature: https://github.com/cloudscape-design/components/pull/918

### How has this been tested?

Existing tests.

<details>
   <summary>Review checklist</summary>

_The following items are to be evaluated by the author(s) and the reviewer(s)._

#### Correctness

- _Changes include appropriate documentation updates._
- _Changes are backward-compatible if not indicated, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#public-apis)._
- _Changes do not include unsupported browser features, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#browsers-support)._
- _Changes were manually tested for accessibility, see [accessibility guidelines](https://cloudscape.design/foundation/core-principles/accessibility/)._

#### Security

- _If the code handles URLs: all URLs are validated through [the `checkSafeUrl` function](https://github.com/cloudscape-design/components/blob/main/src/internal/utils/check-safe-url.ts)._

#### Testing

- _Changes are covered with new/existing unit tests?_
- _Changes are covered with new/existing integration tests?_
</details>

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1240, opened 2023-06-22
refactor: Group reusable table elements and utilities

pan-kot:
### Description

The refactoring makes table building blocks better encapsulated including dedicated unit tests and even test pages built around a custom table to ensure the right level of abstraction. This will be eventually needed to support the upcoming features (see hQUSAdAwRtGN).

Currently, two features have been migrated: sticky-columns and sticky-scrollbar. Eventually, there is going to be much more, including sticky-header, table-wrapper, table-head, table-cell, selection, inline-editing, column-widths etc.

### How has this been tested?

Existing tests.

<details>

#### Correctness

#### Security

#### Testing

</details>

---


just-boris (commented):
Checked the linked quip doc. I do not think we are at this stage, when the component structure is already decided.

Moving the files ahead of time is a little pre mature (you may need to move them back)

pan-kot:
> Checked the linked quip doc. I do not think we are at this stage, when the component structure is already decided.
> 
> Moving the files ahead of time is a little pre mature (you may need to move them back)

The idea of the refactoring is to ensure the table features are well encapsulated. That is already true for sticky-columns and sticky-scrollbar but moving those elsewhere helps to secure it. The more features we move this way the better code we end up with regardless of the future plans.

My plan is for every feature moved to the internal/components/table-fragments to create a dedicated test page with a custom table and individual unit- and integration tests.

just-boris:
Reorganising code inside `src/table`, ok 👍 . Moving code into `src/internal` is premature, because it is still unclear if a new component is going to be created or not. (Also what if the new component will live in a separate repo, similar to board components)?

pan-kot:
> Reorganising code inside `src/table`, ok 👍 . Moving code into `src/internal` is premature, because it is still unclear if a new component is going to be created or not. (Also what if the new component will live in a separate repo, similar to board components)?

Alright, thanks for the clarification - I will move the code to under src/table.

src/area-chart/__tests__/area-chart-use-chart-model.test.tsx
```diff
@@ -8,7 +8,7 @@ import { ChartDataTypes } from '../../internal/components/cartesian-chart/interf
 import { act, render, fireEvent } from '@testing-library/react';
 import { AreaChartProps } from '../interfaces';
 import { KeyCode } from '../../internal/keycode';
-import { useReaction } from '../model/async-store';
+import { useReaction } from '../../internal/async-store';
```
just-boris, line 11: We discussed it in the past: https://github.com/cloudscape-design/components/pull/998

You have not done your part of formalizing this utility, so it also cannot be moved

just-boris (commented):
Part about moving table utils, approved ✅ 
Async store, no 🛑

src/area-chart/__tests__/area-chart-use-chart-model.test.tsx
pan-kot: The tool is already used in both area-chart and table. Why do you think keeping it under area-chart is better?

pan-kot: The sticky-columns util does not use the async-store utility to its full extent. Would it work for you if I update the util to use a simplified store and move that one to under internal?

pan-kot: I'll move the async-store back to under src/area-chart to unblock the PR.

Created a new PR to address the feedback on async store: https://github.com/cloudscape-design/components/pull/1246

src/area-chart/elements/highlighted-point.tsx
```diff
@@ -5,7 +5,7 @@ import React, { forwardRef, memo } from 'react';
 import HighlightedPoint from '../../internal/components/cartesian-chart/highlighted-point';
 
 import { ChartModel } from '../model';
-import { useSelector } from '../model/async-store';
+import { useSelector } from '../async-store';
```
just-boris, line 8: Weird to see moves inside chart folder in a commit named "Group reusable **table** elements", but okay

just-boris (approved):

PR #1244, opened 2023-06-23
fix: Avoid cutting off half the stroke when line is at the edge of chart

jperals:
### Description

Extend clipping rectangle vertically by half the stroke width in each direction vertically, to prevent the line stroke from being clipped in segments where it has the maximum or minimum value of the domain for several times in a row.

Issue: AWSUI-21329

### How has this been tested?

- Added permutation
- Manually tested with Chrome, Firefox and Safari

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/mixed-line-bar-chart/data-series.tsx
+++ b/src/mixed-line-bar-chart/data-series.tsx
@@ -1,6 +1,6 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
-import React, { useMemo } from 'react';
+import React, { useLayoutEffect, useMemo, useRef } from 'react';
 import clsx from 'clsx';
 
 import { useUniqueId } from '../internal/hooks/use-unique-id';
@@ -42,6 +42,8 @@ export default function DataSeries<T extends ChartDataTypes>({
   yScale,
 }: DataSeriesProps<T>) {
   const chartAreaClipPath = useUniqueId('awsui-mixed-line-bar-chart__chart-area-');
+  const seriesRef = useRef<SVGGElement | null>(null);
+  const strokeWidth = useRef(0);
 
   const stackedBarOffsetMaps: StackedOffsets[] = useMemo(() => {
     if (!stackedBars) {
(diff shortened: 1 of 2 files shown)


src/mixed-line-bar-chart/data-series.tsx
```diff
@@ -57,11 +59,17 @@ export default function DataSeries<T extends ChartDataTypes>({
     return calculateOffsetMaps(barData);
   }, [visibleSeries, stackedBars]);
 
+  useLayoutEffect(() => {
+    if (!strokeWidth.current && seriesRef.current) {
+      strokeWidth.current = parseInt(getComputedStyle(seriesRef.current).strokeWidth);
```
just-boris, line 64: This value will not apply on the current render. Do you see any risks of using stale value here?

Is this over-engineering really necessary? Can we just leave 3-4 pixels to account all possible widths?

jperals: > This value will not apply on the current render. Do you see any risks of using stale value here?

Right, because changes in refs values will not trigger a re-render. Is that what you mean?

> Is this over-engineering really necessary? Can we just leave 3-4 pixels to account all possible widths?

Main reason why I did it this way is because I assume the clipping has a purpose, i.e, to cut off any lines that stick out of the graph, and I wanted to minimize changes to that behavior, i.e, keep any eventual overflowing graphics to the minimum necessary.

just-boris: > Is that what you mean?

Yes 

> keep any eventual overflowing graphics to the minimum necessary

Still recommend to find a minimal but constant value and put it into the code. Sometimes a hard-coded value performs better than a dynamic detection. If you need an extra example to convince, have a look at these constants: https://github.com/cloudscape-design/components/blob/main/src/internal/components/cartesian-chart/constants.ts

jperals: Well, and I was mainly doing this because the stroke width is based on a design token (`border-line-chart-width`) but it's actually not exposed, so at the moment there is no way that this can be broken by theming.

just-boris (approved):

PR #1264, opened 2023-06-29
fix: Move inline-editor success icon to focus state

johannes-weber:
### Description

When a user edits a cell, the cell stays in focus after an successful edit.
Before this change, the success icon was displayed when the edited cell lost it's focus.

With this change, the success icon will be part of the cell's focus state and visible to the user. After the cell loses focus, the success icon get's hidden as well. See the new moved inline-editor success icon in action on [this PR deployment](https://d21d5uik3ws71m.cloudfront.net/components/43cb25ec0fae4cd65cfa470b3bbd99890a6b32e8/index.html#/light/table/editable):
- edit a cell and click save -> the success icon appears as part of the focused state; live region informing about the success gets rendered in the DOM as well
- move the focus away e.g. by tabbing to the next editable cell -> the success icon disappears.
- move back to the edited cell -> the success icon get's not visible.

Related links, issue #, if available: AWSUI-21298

### How has this been tested?

- manually tested in all supported browsers.
- added additional integration tests.
- run the performance tests in my pipeline, table tests are below the threshold.

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/table/__integ__/inline-editing.test.ts
```diff
@@ -62,6 +65,42 @@ test(
   })
 );
 
+test(
+  'success icon is displayed and aria live region is rendered in after a successful edit',
+  setupTest(async page => {
+    await page.click(cellRoot$);
+    await page.click(cellSaveButton.toSelector());
+    await expect(page.isFocused(cellEditButton$)).resolves.toBe(true);
+    await expect(page.isDisplayed(successIcon$)).resolves.toBe(true);
```
timogasda, line 74: It would make sense to also test that the icon was *not* there before

timogasda (changes requested):
Needs some optimization.

Also, please take another pass at the PR title. The `fix: ` suffix means that this will go into the release notes. I don't think "Handle inline-editor success icon" is a meaningful release note message

src/table/body-cell/index.tsx
```diff
@@ -85,29 +103,33 @@ function TableCellEditable<ItemType>({
       ) : (
         <>
           {column.cell(item)}
-          {successfulEdit && (
-            <>
-              <span
-                className={styles['body-cell-success']}
-                aria-label={ariaLabels?.successfulEditLabel?.(column)}
-                role="img"
+          <div className={styles['body-cell-editor-control']}>
```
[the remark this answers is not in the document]
johannes-weber: Thanks for the pointer! I'll address the feedback and reduce the nodes for the default state.

I was doing a local rendering performance test which didn't show a significant change. Doing that now with an increased data set and awaiting the performance test result in my pipeline.

src/table/__integ__/inline-editing.test.ts
johannes-weber: Added the additional checks to validate the elements are not existing before the cell edit process is starting.

src/table/body-cell/index.tsx
```diff
@@ -56,6 +56,28 @@ function TableCellEditable<ItemType>({
   const [hasFocus, setHasFocus] = useState(false);
   const showIcon = hasHover || hasFocus;
 
+  const prevSuccessfulEdit = useRef(successfulEdit);
+  const prevHasFocus = useRef(hasFocus);
```
timogasda, line 60: We have a hook for this: https://github.com/cloudscape-design/components/blob/main/src/internal/hooks/use-previous/index.ts

src/table/body-cell/index.tsx
```diff
@@ -56,6 +56,28 @@ function TableCellEditable<ItemType>({
...
+  const [showSuccessIcon, setShowSuccessIcon] = useState(false);
+
+  useEffect(() => {
+    // Early return if there hasn't been any successfulEdit state changes.
+    if (!successfulEdit && !prevSuccessfulEdit.current) {
+      return;
```
timogasda, line 66: If we exit early because of `successfulEdit`, doesn't that mean that we sometimes dont update `prevSuccessfulEdit`? I think this will be fixed automatically we switch to our `usePrevious` hook.

src/table/body-cell/index.tsx
```diff
@@ -56,6 +56,28 @@ function TableCellEditable<ItemType>({
...
+    }
+
+    // Show the success icon right after a successful edit (when successfulEdit switches to true)
```
timogasda, line 69: The comments are describing the opposite of what's happening inside the `if` blocks

timogasda (changes requested):

src/table/body-cell/index.tsx
johannes-weber: I wasn't aware of the `usePrevious`, thanks for the shout-out - updating the PR.

src/table/body-cell/index.tsx
```diff
@@ -9,6 +9,7 @@ import { TableTdElement, TableTdElementProps } from './td-element';
 import { InlineEditor } from './inline-editor';
```
timogasda, line 9: When I click on the success icon after a successful edit, the icon disappears. That seems like a strange interaction?

src/table/body-cell/index.tsx
```diff
@@ -9,6 +9,7 @@ import { TableTdElement, TableTdElementProps } from './td-element';
...
 import LiveRegion from '../../internal/components/live-region/index.js';
```
timogasda, line 10: Discarding an edit operation still displays the success icon, which is a little bit confusing.

To reproduce:
1. Start editing a cell
2. Submit the change - you see the success icon
3. Start editing *the same* cell
4. Cancel editing
5. Success icon is still there

src/table/body-cell/index.tsx
```diff
@@ -77,6 +93,7 @@ function TableCellEditable<ItemType>({
           column={column}
           item={item}
           onEditEnd={e => {
+            setShowSuccessIcon(false);
```
johannes-weber, line 96: Discarding an edit operation hides the success icon.

src/table/body-cell/index.tsx
johannes-weber: Good catch (both issues)! 
This actually happened because the button's onBlur event got fired which removes the success icon from the DOM, so the success icon did no receive the click anymore which would have bubbled up to the cell.

Fixed by preventing the button's blur event in the `onMouseDown` callback on the success-icon. When releasing the mouse, it triggers the cell's onClick which takes care of getting into the editable state.

src/table/body-cell/index.tsx
johannes-weber: Fixed.

src/table/body-cell/index.tsx
```diff
@@ -56,6 +57,21 @@ function TableCellEditable<ItemType>({
...
   const showIcon = hasHover || hasFocus;
 
+  const prevSuccessfulEdit = usePrevious(successfulEdit);
+  const prevHasFocus = usePrevious(hasFocus);
+  const [showSuccessIcon, setShowSuccessIcon] = useState(false);
+
+  useEffect(() => {
+    // Hide the success icon after a successful edit, when cell loses focus.
+    if (successfulEdit && prevSuccessfulEdit && !hasFocus && prevHasFocus) {
+      setShowSuccessIcon(false);
```
johannes-weber, line 67: This line is covered by this [newly added integ test case](https://github.com/cloudscape-design/components/pull/1264/files#diff-21cfe5e571a29f732cbfb7d3b2cf38c294086bbf84d2c047d1e61b14fefa2fbdR154-R177).

src/table/body-cell/index.tsx
```diff
@@ -85,12 +102,17 @@ function TableCellEditable<ItemType>({
...
+          {showSuccessIcon && showIcon && (
             <>
               <span
                 className={styles['body-cell-success']}
                 aria-label={ariaLabels?.successfulEditLabel?.(column)}
                 role="img"
+                onMouseDown={e => {
+                  // Prevent the editor's Button blur event to be fired when clicking the success icon.
+                  // This prevents unfocusing the button and triggers the `TableTdElement` onClick event which initiates the edit mode.
+                  e.preventDefault();
```
timogasda, line 114: Could this have been solved with `pointer-events` on the icon? That might be a bit more "elegant". If not, this is fine too.

src/table/body-cell/index.tsx
timogasda: Got it. I think we want to avoid disabling codecov checks in GitHub. So we have two options:

1) Add an inline comment here to ignore the line in codecov
2) Find a way to test this in a unit test (preferred)

src/table/body-cell/index.tsx
johannes-weber: I've tried that first but a `pointer-events: none` would results in a little flickering from focus state to hover state to edit state when clicking the success icon: `onBlur` gets triggered before `onClick` where Button's `onBlur` removes the focus state followed by the `onClick` which enables the edit state:

https://github.com/cloudscape-design/components/assets/569011/16ad87b0-e821-4f2a-a898-c2251a380abf

Intercepting the click on `onMouseDown` prevents `onBlur` from being called. This gets the cell directly into edit mode.

src/table/body-cell/index.tsx
johannes-weber: Makes sense, I added a unit test for this case. I also added more unit test scenarios that don't have to be integration tests and would run much faster. Codecov is happy now.

timogasda (approved):

PR #1266, opened 2023-06-29
feat: Pass form field context to text filter

avinashbot:
### Description

Recently encountered a scenario where a user needed to pass in "usage instructions" for a text filter in the table tools. After some discussion, we felt that the constraint text UX was indeed the way to go. We support proper form fields in tables anyway (see [select filter](https://cloudscape.design/examples/react/table-select-filter.html)), so this isn't too far off.

We also discussed and felt it wouldn't be right to support the form field context by itself without also exposing the form field properties (`ariaLabelledby`, `ariaDescribedby`, `controlId`), so this adds them too.

Needs a review with the team before merging.

Related links, issue #, if available: AWSUI-21356

### How has this been tested?

Unit tests!

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/text-filter/internal.tsx
+++ b/src/text-filter/internal.tsx
@@ -6,11 +6,13 @@ import InternalInput from '../input/internal';
 import { getBaseProps } from '../internal/base-component';
 import useForwardFocus from '../internal/hooks/forward-focus';
 import { fireNonCancelableEvent } from '../internal/events';
-import styles from './styles.css.js';
 import { InternalBaseComponentProps } from '../internal/hooks/use-base-component';
-import { TextFilterProps } from './interfaces';
 import { useUniqueId } from '../internal/hooks/use-unique-id';
+import { joinStrings } from '../internal/utils/strings';
+import { TextFilterProps } from './interfaces';
 import { SearchResults } from './search-results';
+
+import styles from './styles.css.js';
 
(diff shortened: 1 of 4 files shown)


src/text-filter/internal.tsx
```diff
@@ -6,12 +6,14 @@ import InternalInput from '../input/internal';
...
 import { fireNonCancelableEvent } from '../internal/events';
-import styles from './styles.css.js';
 import { InternalBaseComponentProps } from '../internal/hooks/use-base-component';
-import { TextFilterProps } from './interfaces';
 import { useUniqueId } from '../internal/hooks/use-unique-id';
+import { joinStrings } from '../internal/utils/strings';
+import { TextFilterProps } from './interfaces';
 import { SearchResults } from './search-results';
 
+import styles from './styles.css.js';
```
Al-Dani, line 15: looks like it's unnecessary

Al-Dani (changes requested):

src/text-filter/internal.tsx
avinashbot: Couldn't help myself; I didn't add it, just moved it down from deep inside the imports (see line 9)

Al-Dani (approved):

PR #1272, opened 2023-07-03
chore: Fix sticky columns cleanup

rubencarvalho:
### Description

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/table/sticky-columns/use-sticky-columns.ts
```diff
@@ -208,13 +208,18 @@ export function useStickyCellStyles({
...
+      };
+
+      update();
+
       return () => {
-        unsubscribe();
-        // Force the cleanup
-        updateCellStyles(null, { padLeft: false, lastLeft: false, lastRight: false, offset: {} });
+        // Unsubscribe / force cleanup
+        update();
```
pan-kot, line 222: I don't quite get this line. The `update()` call triggers subscribe, not unsubscribe - isn't it?

rubencarvalho: That makes sense. Not only does it look cleaner, by having the conditional checks it should also solve the `StrictMode` issue. I'll try that approach.

src/table/sticky-columns/use-sticky-columns.ts
```diff
@@ -171,56 +171,57 @@ export function useStickyCellStyles({
...
+      // If the node is not null (i.e., the table cell is being mounted or updated, not unmounted),
+      // set up a new subscription to the store's updates
+      if (node) {
+        unsubscribeRef.current = stickyColumns.store.subscribe(selector, (newState, prevState) => {
+          updateCellStyles(selector(newState), selector(prevState));
+        });
+      }
     },
+
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
pan-kot, line 199: I think it will be safer to declare the `updateCellStyles` inside the ref callback and add `store` to the dependency array. Otherwise if there would be a new dependency for updateCellStyles it might result in a bug which would be very difficult to track.

src/table/sticky-columns/use-sticky-columns.ts
```diff
@@ -208,18 +210,16 @@ export function useStickyCellStyles({
...
+        unsubscribeRef.current = stickyColumns.store.subscribe(selector, (newState, prevState) => {
+          updateCellStyles(selector(newState), selector(prevState));
+        });
+      }
     },
-    // getClassName is expected to be pure
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-    [stickyColumns.store, stickyColumns.isEnabled, columnId]
+
+    [columnId, setCell, stickyColumns.store, getClassName]
```
pan-kot, line 222: I think we should probably assume the getClassName is pure and exclude it from the dependency array. Otherwise it is going to have an impact on performance unless the getClassName is memoized by the consumer

pan-kot (approved):

PR #1275, opened 2023-07-04
fix: Include stepName and subStepName in events

connorlanigan:
### Description

With this change, we include the fields `stepName` and `subStepName` in the metrics event. Previously, we would only emit the selector to the elements containing those names. When a step is completed (i.e. at the point where `funnelStepComplete` is called), this selector does not point at the correct element anymore, because that element has been unmounted. To mitigate this, we save the name already at the start of the effect.

Related links, issue #, if available: nBvhAQL0oUIC , Line 33

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/analytics/components/analytics-funnel.tsx
+++ b/src/internal/analytics/components/analytics-funnel.tsx
@@ -21,6 +21,7 @@ import { FunnelProps, FunnelStepProps } from '../interfaces';
 import {
   DATA_ATTR_FUNNEL_STEP,
   getFunnelNameSelector,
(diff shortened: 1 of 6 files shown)


src/internal/analytics/components/analytics-funnel.tsx
connorlanigan, line 1: This file fixes the original bug (by extracting the name already _before_ the cleanup callback is called). The other files are mainly for API parity.

timogasda (approved):

PR #1280, opened 2023-07-04
fix: Re-emit error metrics if an error persists after attempting to submit

connorlanigan:
### Description

When a form is submitted (either in a Form component or in a page of a Wizard component), and an error is shown in one of the form fields, we emit the `funnelSubStepError` event. If a form-level error happens, we emit the `funnelError` event. When the user tries to submit the form again, and the same error happens, the error is not re-rendered. The result of this was that no new error event is emitted.

After this change, if there are errors after trying to submit a form (or a page, in the case of a Wizard), the error events are emitted again. This is achieved by applying a counter that is incremented on every attempt to submit the form, which causes the `useEffect` hooks to re-run.

Related links, issue #, if available: nBvhAQL0oUIC , Line 26

### How has this been tested?

I have added two new testing pages, which I have used for manual testing.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/form/with-validation.page.tsx
+++ b/pages/form/with-validation.page.tsx
@@ -2,7 +2,6 @@
 // SPDX-License-Identifier: Apache-2.0
 import React, { useState } from 'react';
 import {
-  AppLayout,
   BreadcrumbGroup,
   Button,
   Checkbox,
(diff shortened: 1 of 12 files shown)


johannes-weber (changes requested):
Please fix the failing build (type errors in `src/internal/analytics/__tests__/hooks/use-funnel.test.tsx`).

Missing tests
It'd be useful to have tests which covers
- the `funnelNextOrSubmitAttempt` counter inside `analytics-funnel.tsx` and
- the `useEffect` execution after the counter increased in the form and wizard example.

johannes-weber (approved):
Looks good to me!

johannes-weber (approved):

PR #1283, opened 2023-07-04
chore: Unsupport iconName in runtime drawers api

just-boris:
### Description

There is no guarantee that this iconName is going to be available in the receiving app layout, because it could be a different bundle.

Since we have no actual use-cases for `iconName` property it would be safer to unsupport the feature before it becomes problematic

### How has this been tested?

1. Updated dev page to exclude built-in icons
2. Added a unit test

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/plugins/drawers-controller.ts
+++ b/src/internal/plugins/drawers-controller.ts
@@ -1,13 +1,11 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 import { DrawerItem } from '../../app-layout/drawer/interfaces';
-import { IconProps } from '../../icon/interfaces';
 
-export type DrawerConfig = Omit<DrawerItem, 'content'> & {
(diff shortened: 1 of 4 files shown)


src/internal/plugins/drawers-controller.ts
```diff
@@ -1,13 +1,11 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 import { DrawerItem } from '../../app-layout/drawer/interfaces';
-import { IconProps } from '../../icon/interfaces';
 
-export type DrawerConfig = Omit<DrawerItem, 'content'> & {
+export type DrawerConfig = Omit<DrawerItem, 'content' | 'trigger'> & {
```
Al-Dani, line 5: is it still required to exist in `DrawerItem` interface?

just-boris: Yes, because the normal React props API still supports this feature

Al-Dani (approved):

PR #1290, opened 2023-07-06
fix: Prevent recovery button from wrapping to next line in dropdown footers

avinashbot:
### Description

The status indicator is a display-block component, for wrapping reasons. This also means we can't place any inline text next to it without it breaking to the next line if it takes up multiple lines. We can't switch the default to inline because it breaks wrapping in many circumstances (e.g. key-value pairs), so I'm adding this as an internal property for now.

Related links, issue #, if available: AWSUI-21388

### How has this been tested?

Permutation tests. With `wrapText={false}`, it will overflow past the page (which is, I guess, expected), so I added a wrapper div that hides overflow past the page.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/multiselect/screenshot.page.tsx
+++ b/pages/multiselect/screenshot.page.tsx
@@ -27,6 +27,7 @@ const options: MultiselectProps.Options = [
 export default function () {
   const [selected, setSelected] = useState<MultiselectProps.Options>(options.slice(0, 3));
   const [virtualScroll, setVirtualScroll] = useState<boolean>(false);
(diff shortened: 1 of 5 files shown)


johannes-weber (changes requested):
I've verified the changes locally, works as expected.

I recommend adding another example to the multiselect async page, which reproduces the issue described in the ticket and allows to do a visual regression test against this behaviour.

johannes-weber (approved):
LGTM!
_Feel free to assign review about the updated screenshot test directly to me._

PR #1305, opened 2023-07-10
chore: Make sticky scrollbar visible on top of sticky columns

rubencarvalho:
### Description

Tables with sticky columns will have the columns overlap the scrollbar:

Screenshot 2023-07-24 at 09 51 03

This is due to the `z-index` of the sticky columns being higher than that of the sticky scrollbar. This issue affects both Classic and VR.

As a fix, we increased the z-index of the scrollbar to be higher than that of the sticky columns. However, this introduced a bug: the scrollbar now intercepts clicks when users try to interact with elements (e.g.: checkboxes, radio buttons, links):

Screenshot 2023-07-24 at 10 20 34 copy

(It is important to note this only happens on Mac OS when the option to “Show scroll bars” is set to “When scrolling” - this is the OS default)

With this PR, we will allocate space for the scrollbar on the potential problematic tables (all Classic tables and the `full-page` variant of the VR table, when sticky columns are enabled and the table is scrollable):

Screenshot 2023-08-02 at 17 35 36 copy

Related links:
Problem statement: Quip `Ib8BAcSFMapr`
Solutions proposal: Quip `jb9CADaASMbV`

### How has this been tested?

Covered by existing screenshot tests. Run on my v3-rucarva pipeline.

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/table/sticky-scrollbar/styles.scss
```diff
@@ -23,4 +25,28 @@
     // For example it is set as "show when scrolling"
     margin-top: -15px;
   }
+
+  &-offset {
+    // Needed for when sticky columns are enabled.
+    // We offset the scrollbar lower for it not to overlap interactive elements
+    // such as checkboxes or links
+    z-index: 800; // Higher than sticky columns
```
jperals, line 33: What happens if it is not higher but the same than sticky columns?

rubencarvalho: Same problem.

- When no `z-index` was applied (default state without sticky columns enabled), even though the scrollbar is visually on top, it doesn't necessarily intercept click events because it isn't explicitly above the columns in the stacking context. It's in a separate stacking context due to its `position: sticky` property. The columns, being `position: static` (default value) and in a different stacking context, receive the clicks as normal.

- When a `z-index` is applied to both - even if the same value, they are in the same stacking context. The order of elements in the DOM now determines which one is on top. The scrollbar is later in the DOM, so it will be on top and start intercepting the clicks.

jperals: Thanks for the explanation! So if both have the same `z-index` but `position: sticky` was applied not to the main sticky scrollbar wrapper but to _a child_ of it (and perhaps only this child have non-zero height), so that the scrollbar is again in a different stacking context, would that help?

rubencarvalho: Unfortunately, that also doesn't work with our current sticky scrollbar implementation...

taheramr (approved):

PR #1309, opened 2023-07-11
fix: Fix enteredTextLabel message not being provided in property filter

avinashbot:
### Description

`enteredTextLabel` is provided by autosuggest strings, which when they receive `undefined`, should use the built-in ones. Unfortunately, when property filter receives `undefined`, it provides a fallback value to autosuggest, which breaks this assumption.

Also I noticed a bunch of the buttons submit forms, so I fixed that as well while I was here.

### How has this been tested?

The property filter, having been one of the beta components, lacked the GA component test suite. Added i18n tests to the component.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/property-filter/index.tsx
+++ b/src/property-filter/index.tsx
@@ -304,7 +304,7 @@ const PropertyFilter = React.forwardRef(
           <PropertyFilterAutosuggest
             ref={inputRef}
             virtualScroll={virtualScroll}
-            enteredTextLabel={i18nStrings.enteredTextLabel ?? (value => value)}
+            enteredTextLabel={i18nStrings.enteredTextLabel}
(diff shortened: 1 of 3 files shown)


src/property-filter/index.tsx
```diff
@@ -304,7 +304,7 @@ const PropertyFilter = React.forwardRef(
           <PropertyFilterAutosuggest
             ref={inputRef}
             virtualScroll={virtualScroll}
-            enteredTextLabel={i18nStrings.enteredTextLabel ?? (value => value)}
+            enteredTextLabel={i18nStrings.enteredTextLabel}
```
avinashbot, line 307: Here's the primary change, by the way

abdhalees (approved):

PR #1312, opened 2023-07-11
fix: Use appropriate aria-haspopup for select/multiselect with filtering

gethinwebster:
### Description

Uses `aria-haspopup="dialog"` for select & multiselect with filtering,
as the previous value of "listbox" is not appropriate for something that's
more complex than a simple list.

Also adjust roles and `aria-controls` to align.

Related links, issue #, if available: AWSUI-21334

### How has this been tested?

Unit tests, and tested with VO & Chrome

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/internal/components/dropdown/interfaces.ts
```diff
@@ -132,6 +132,15 @@ export interface DropdownProps extends ExpandToViewport {
...
   onBlur?: NonCancelableEventHandler<Pick<React.FocusEvent, 'target' | 'relatedTarget'>>;
+
+  /**
+   * ID for the dropdown content wrapper
+   */
+  dropdownContentId?: string;
+  /**
+   * HTML role for the dropdown content wrapper
+   */
+  dropdownContentRole?: string;
```
timogasda, line 143: Would it make sense to "enum" this like we do `ariaHasPopup`?

gethinwebster: Given that there are so many options I'm not sure it makes so much sense here. We could _perhaps_ limit it to the options we think might be relevant here, but especially as it's an internal interface I'm not sure it has so much value

timogasda: Right, but `ariaHasPopup` is an internal prop as well, right?

And yes, I would limit it to the values that we currently think are reasonable (which I think is just `dialog`?).

Not a blocking comment, though. I'll leave the final call up to you.

timogasda (approved):
One non-blocking comment

PR #1314, opened 2023-07-11
feat: Move i18nStrings.submitText in wizard to top level

avinashbot:
### Description

As signed off by the dev team, moving these situation-dependent labels out of `i18nStrings` and into the component proper. No new strings, we're all good on content review (unless `@deprecated` messages count?).

### How has this been tested?

Updated unit tests, but that's about it.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/wizard/__tests__/wizard.test.tsx
+++ b/src/wizard/__tests__/wizard.test.tsx
@@ -31,7 +31,7 @@ type DefaultWizardProps = Optional<WizardProps, 'i18nStrings' | 'steps'>;
 const renderDefaultWizard = (
   wizardProps?: DefaultWizardProps
 ): [WizardWrapper, (props: DefaultWizardProps) => void] => {
-  const defaultProps = { i18nStrings: DEFAULT_I18N_SETS[0], submitButtonText: 'Create record', steps: DEFAULT_STEPS };
+  const defaultProps = { i18nStrings: DEFAULT_I18N_SETS[0], steps: DEFAULT_STEPS };
   const [wrapper, rerender] = renderWizard({ ...defaultProps, ...wizardProps });
   return [wrapper, (newProps: DefaultWizardProps) => rerender({ ...defaultProps, ...newProps })];
 };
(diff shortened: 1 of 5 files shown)


src/wizard/interfaces.ts
```diff
@@ -48,13 +48,18 @@ export interface WizardProps extends BaseComponentProps {
    * - `cancelButton` (string) - The text of the button that enables the user to exit the flow.
    * - `previousButton` (string) - The text of the button that enables the user to return to the previous step.
    * - `nextButton` (string) - The text of the button that enables the user to move to the next step.
-   * - `submitButton` (string) - The text of the button that enables the user to submit the form.
+   * - `submitButton` (deprecated) (string) - The text of the button that enables the user to submit the form.
```
jperals, line 51: There are two other cases of deprecated nested props [here](https://github.com/cloudscape-design/components/blob/main/src/flashbar/interfaces.ts#L60) and [here](https://github.com/cloudscape-design/components/blob/main/src/flashbar/interfaces.ts#L62), and they use a different formatting. And I think that other formatting brings the opportunity to state what other property should be used instead (as you already did in the prop further down, but that string is not shown in our API docs).

src/wizard/__tests__/wizard.test.tsx
```diff
@@ -216,7 +218,7 @@ describe('Primary button', () => {
       if (index < DEFAULT_STEPS.length - 1) {
         expect(wrapper.findPrimaryButton().getElement()).toHaveTextContent(DEFAULT_I18N_SETS[0].nextButton!);
       } else {
-        expect(wrapper.findPrimaryButton().getElement()).toHaveTextContent(DEFAULT_I18N_SETS[0].submitButton);
+        expect(wrapper.findPrimaryButton().getElement()).toHaveTextContent(DEFAULT_I18N_SETS[0].submitButton!);
```
jperals, line 221: Since `renderDefaultWizard` uses the new `submitButtonText` prop, the case of using `i18nStrings.submitButton` is not covered now. This assertion passes because the value of both is the same. Since using the old prop should still be supported even if it's deprecated, I think we should keep covering this case in tests to make sure our fallback logic is not broken.

jperals:
Now both the new top-level property and the old one are optional, which introduces the possibility of not providing any value, which was not possible before. Thinking of ways to mitigate this, maybe with a dev warning?

just-boris:
Speaking of dev warnings, let's be very mindful about this. If every minor issue would print something to console, people will quickly lose interest in reading any of them.

How does it fail when both properties missing? If the error is self explanatory (unable to read property of undefined), it is self-explanatory without a warning

jperals:
Actually I thought for "gray area cases" like this we would fall back to the i18n provider if nothing is provided, since it's better to display a generic text like "Submit" than not displaying anything in such case (two-level fallback if you want to call it that way). Would the fact that the prop is now at the top level make it not suitable for the i18n provider?

jperals (approved):

PR #1321, opened 2023-07-13
chore: Update VR typography heading sizes and weight

jkuelz:
### Description

This is part of an ongoing initiative to increase the information density of VR in both comfortable mode and compact mode, better align with brand and marketing, and address VR customer feedback.

This change is the outcome of an extensive usability study where we A/B tested alternate heading type ramps with users. While the original heavy typography left a positive impression on users when it came to navigation and findability, when it comes to focusing on a task and diving deep, it often makes customers feel overwhelmed, especially in high dense interfaces. 

The new typography implemented in this commit (smaller sizes and removal of heavy weight) performed the best for users focused on completing tasks.

Screen Shot 2023-07-12 at 6 09 24 PM

### How has this been tested?

Ran through dev pipeline, reviewed with visual design squad, and tested with users.

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/table/__integ__/table.test.ts
```diff
@@ -38,10 +38,10 @@ test(
     const header = createWrapper().findContainer().findHeader().toSelector();
     await browser.url('#/light/table/full-page-variant?visualRefresh=true');
     const page = new BasePageObject(browser);
-    await expect(browser.execute(extractHeight, header)).resolves.toEqual({ height: '101px', marginBottom: '0px' });
+    await expect(browser.execute(extractHeight, header)).resolves.toEqual({ height: '91px', marginBottom: '0px' });
```
jperals, line 41: Nitpick: I think what this test tries to say is that the sum of the height and the bottom margin should stay the same on scrolling, so that the content below the header doesn't jump. After these changes, this is still true 👍, but I think this test would better assert just that equality (sum of height and bottom margin is the same before and after) rather than kind-of-magic numbers. Otherwise a similar update in the future could easily break this.

jperals (dismissed):

src/table/__integ__/table.test.ts
jkuelz: Updated! All that seems to matter is that the height is not smaller after scroll, otherwise it tries to recalculate indefinitely and jumps between the two sizes.

abdhalees (approved):

PR #1328, opened 2023-07-14
feat: Emit error events for Alerts

connorlanigan:
### Description

When an Alert is used to display an error, it now also emits `funnelError` or `funnelSubStepError` events, depending on whether it is placed inside a sub step (such as a container) or somewhere else in a funnel.

### How has this been tested?

Added unit tests, and tested it manually on the new dev page

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/alert/__tests__/alert-analytics.test.tsx
+++ b/src/alert/__tests__/alert-analytics.test.tsx
@@ -1,0 +1,183 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import React from 'react';
+import { act, render } from '@testing-library/react';
+
+import { FunnelMetrics } from '../../../lib/components/internal/analytics';
+import { useFunnel } from '../../../lib/components/internal/analytics/hooks/use-funnel';
+import Alert from '../../../lib/components/alert';
+
+import {
+  AnalyticsFunnel,
+  AnalyticsFunnelStep,
+  AnalyticsFunnelSubStep,
+} from '../../../lib/components/internal/analytics/components/analytics-funnel';
+
+import { mockFunnelMetrics } from '../../internal/analytics/__tests__/mocks';
+
+describe('Alert Analytics', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+    mockFunnelMetrics();
+  });
+
+  test('sends funnelSubStepError metric when the alert is placed inside a substep', () => {
+    render(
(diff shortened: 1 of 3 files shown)


src/alert/index.tsx
```diff
@@ -1,15 +1,56 @@
...
+          subStepAllSelector: getSubStepAllSelector(),
+        });
+      } else {
+        FunnelMetrics.funnelError({
+          funnelInteractionId,
+        });
+      }
+
+      return () => {
+        // eslint-disable-next-line react-hooks/exhaustive-deps
```
jperals, line 46: Does this comment line do anything here?

src/alert/__tests__/alert-analytics.test.tsx
```diff
@@ -0,0 +1,183 @@
...
+      </AnalyticsFunnel>
+    );
+
+    expect(FunnelMetrics.funnelError).not.toHaveBeenCalled();
+
+    expect(FunnelMetrics.funnelSubStepError).toHaveBeenCalledTimes(1);
+    expect(FunnelMetrics.funnelSubStepError).toHaveBeenCalledWith(
+      expect.objectContaining({
+        funnelInteractionId: 'mocked-funnel-id',
+        subStepSelector: expect.any(String),
```
jperals, line 41: Could we test against actual selectors instead of any string, or would it be complicated for some reason?

connorlanigan: That would be difficult, since the selectors contain a randomly generated ID for the substep, which is generated here: 
https://github.com/cloudscape-design/components/blob/eda050146b48c90e6cb47b7037832367e9047300/src/internal/analytics/components/analytics-funnel.tsx#L208

The object looks like this, for example:

```
{
  "funnelInteractionId": "mocked-funnel-id",
  "stepName": undefined,
  "stepNameSelector": ".step-name-selector",
  "stepNumber": 2,
  "subStepAllSelector": "[data-analytics-funnel-substep]",
  "subStepName": undefined,
  "subStepNameSelector": "[data-analytics-funnel-substep=\"substep0-1689587393670-611\"] [data-analytics-funnel-key=\"substep-name\"]",
  "subStepSelector": "[data-analytics-funnel-substep=\"substep0-1689587393670-611\"]"
}
```

src/alert/index.tsx
jperals: OK, thank you for the clarification. Maybe adding an explanatory comment would help here, especially since this rule `react-hooks/exhaustive-deps` doesn't sound related to this specific warning —but I wouldn't block the PR on this.

jperals (approved):

PR #1329, opened 2023-07-14
fix: Emit funnelStepComplete after completion, but not after cancellation

connorlanigan:
### Description

`funnelStepComplete` should be emitted when the funnel is complete. This makes processing the events afterwards easier, since every step has a start and and end event now. The one exception is that no events should be emitted when the flow is cancelled.

### How has this been tested?

Updated unit tests and manual testing in dev pages

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/analytics/components/analytics-funnel.tsx
+++ b/src/internal/analytics/components/analytics-funnel.tsx
@@ -85,7 +85,7 @@ export const AnalyticsFunnel = ({ children, ...props }: AnalyticsFunnelProps) =>
         FunnelMetrics.funnelSuccessful({ funnelInteractionId });
       } else {
         FunnelMetrics.funnelCancelled({ funnelInteractionId });
-        funnelState.current === 'cancelled';
+        funnelState.current = 'cancelled';
       }
     };
   }, []);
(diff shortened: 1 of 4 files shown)


src/internal/analytics/components/analytics-funnel.tsx
```diff
@@ -180,7 +180,7 @@ export const AnalyticsFunnelStep = ({ children, stepNumber, stepNameSelector }:
 
     return () => {
       //eslint-disable-next-line react-hooks/exhaustive-deps
-      if (funnelInteractionId && funnelState.current === 'default') {
+      if (funnelInteractionId && funnelState.current !== 'cancelled') {
```
Al-Dani, line 183: are there any other states besides `default` and `cancelled`?

connorlanigan: Yes, the possible values are defined in this type: https://github.com/cloudscape-design/components/blob/eda050146b48c90e6cb47b7037832367e9047300/src/internal/analytics/context/analytics-context.ts#L6

The `validating` state is similar to the `default` state, except that the funnel is in a loading state (detected thought the primary button having `loading={true}`). If the step unmounts during either of these two states (but the funnel hasn't been cancelled), then the step was successful, and should emit the event here.

The `complete` state is entered when the funnel has been successfully submitted, i.e. the resource has been created. We still want to emit `funnelStepComplete` when the step unmounts, so that the step has a start and end that can later be used in calculations.

Al-Dani (approved):

PR #1331, opened 2023-07-14
fix: Render Visual Refresh background height in sync with content

jperals:
### Description

Alternative approach to https://github.com/cloudscape-design/components/pull/1344, a bit less orthodox (adding imperative DOM updates) but more performant.

As a long-term fix, we want to refactor app layout so that resize observer calls are not necessary to keep the content and the background in sync.

Ticket: AWSUI-21410

### How has this been tested?

- Tested manually by locally switching the package to React 18 and using Chrome developer tools (CPU throttling, performance tools). Performance measurements yielded no performance regression
- Added page to manually test
- Added unit tests
- Deployed in my pipeline. Lighthouse test results report actually faster results than main:
  - Table view:
    - Score: +0.01
    - Bootup time: -118ms (-7.58%)
    - Total blocking time: -41ms (-6.38%)
  - Table inline editing view:
    - Score: +0.01
    - Bootup time: -22ms (-1.35%)
    - Total blocking time: -50ms (-7.27%)

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/app-layout/__tests__/dynamic-overlap.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+  test('sets dynamic overlap when content header is present', () => {
+    const { isDynamicOverlapSet, isDynamicOverlapDisabled } = renderApp({
+      content: <ComponentWithDynamicOverlap />,
+      contentHeader: 'Content header',
+    });
+    expect(isDynamicOverlapSet()).toBe('true');
+    expect(isDynamicOverlapDisabled()).toBe('false');
+  });
+
+  test('sets dynamic overlap when height is higher than 0', () => {
```
rubencarvalho, line 65: Could we have a test case where the height dynamically changes (for example, from `800` to `0`) and we check for the expected values of `isDynamicOverlapSet` and `isDynamicOverlapDisabled`?

Or is this not something that can happen?

jperals: It's not impossible that these state variables change during the lifecycle of an application but I think it should be rather rare. Some of these events should happen:

1. The height of a child component that uses the dynamic overlap hook changes from 0 to non-0, or the other way around from non-0 to 0
2. A child component that uses the dynamic overlap hook is added or removed
3. `contentHeader` is passed and then not passed anymore, or the other way around
4. The value of the `disableDynamicOverlap` property passed to the app layout changes

I just added a test that checks these state variables after re-rendering, based on the case 2 above. Case 1 would be tricky to test because we would need to simulate a call to the resize observer hook.

src/app-layout/visual-refresh/context.tsx
```diff
@@ -487,6 +490,38 @@ export const AppLayoutInternalsProvider = React.forwardRef(
...
+         * because they will rarely change within the lifecycle of an application.
+         */
+        const element = typeof layoutElement === 'function' ? layoutElement(null) : layoutElement?.current;
+        if (isOverlapDisabled || height <= 0) {
+          element?.style.removeProperty(customCssProps.overlapHeight);
+        } else {
+          element?.style.setProperty(customCssProps.overlapHeight, `${height}px`);
+        }
+      },
+      [layoutElement, props.contentHeader, props.disableContentHeaderOverlap]
```
avinashbot, line 522: `props.contentHeader` is a `ReactNode`, which is rarely stable between renders anyway. This means the function will update identity each render, making `useCallback` useless.

If it's just provided to check for the _existence_ of the property rather than the contents, maybe you could define a variable outside the function and use it internally? That would make it safer.

```
const hasContentHeader = !!props.contentHeader;
useCallback(() => { ... }, [layoutElement, hasContentHeader, props.disableContentHeaderOverlap]);
```

src/app-layout/visual-refresh/context.tsx
```diff
@@ -487,6 +490,38 @@ export const AppLayoutInternalsProvider = React.forwardRef(
...
+        /**
+         * React 18 will trigger a paint before the state is correctly updated
+         * (see https://github.com/facebook/react/issues/24331).
+         * To work around this, we bypass React state updates and imperatively update the custom property on the DOM.
+         * An alternative would be to use `queueMicrotask` and `flushSync` in the ResizeObserver callback,
+         * but that would have some performance impact as it would delay the render.
+         * Using React state for `isDynamicOverlapSet` and `isDynamicOverlapDisabled` is less problematic
+         * because they will rarely change within the lifecycle of an application.
+         */
+        const element = typeof layoutElement === 'function' ? layoutElement(null) : layoutElement?.current;
```
avinashbot, line 515: Not sure what `layoutElement(null)` does here. It just unsets the ref? I think it works here just because of internal implementation details, since `layoutElement` happens to be an ObjectRef usually. If we're using the ref for things other than resize observer, maybe another ref property (typed as `ObjectRef`) is needed.

avinashbot (changes requested):

src/app-layout/visual-refresh/context.tsx
jperals: This variable is returned by `useContainerQuery` which returns `React.Ref<any>`, which in turn is `RefCallback<any> | RefObject<any> | null`, but its value is set only by the Layout component [here](https://github.com/cloudscape-design/components/blob/main/src/app-layout/visual-refresh/layout.tsx#L87) so we know it's going to be an ObjectRef and not a RefCallback.

If it looks better, we could as well just do nothing if it is a function.

jperals: Updated.

avinashbot: Not ideal, but it does have a comment and it works for now. Not going to treat it as blocking.

src/internal/hooks/use-dynamic-overlap/index.ts
```diff
@@ -21,21 +23,30 @@ export interface UseDynamicOverlapProps {
 export function useDynamicOverlap(props?: UseDynamicOverlapProps) {
   const disabled = props?.disabled ?? false;
   const setDynamicOverlapHeight = useContext(DynamicOverlapContext);
-  const [overlapHeight, overlapElementRef] = useContainerQuery(rect => rect.contentBoxHeight);
+  const overlapElementRef = useRef(null);
 
-  useLayoutEffect(
-    function handleDynamicOverlapHeight() {
+  const getElement = useCallback(() => overlapElementRef.current, [overlapElementRef]);
```
avinashbot, line 28: 😭 I'm so not a fan of this ResizeObserver hook API

avinashbot (approved):

src/app-layout/__tests__/dynamic-overlap.test.tsx
rubencarvalho: Alright, that makes sense!

src/app-layout/visual-refresh/context.tsx
rubencarvalho: Should `layoutElement` somehow end up being used as a `RefCallback`, it will also cause this `useCallback` to recreate the `updateDynamicOverlapHeight` function on every render.

Anyway, it seems unlikely, so it is also non-blocking for me.

rubencarvalho (approved):

PR #1342, opened 2023-07-18
chore: Add optional chaining to focusedSegmentRef

rubencarvalho:
### Description

This is a corner case we didn't account for. To avoid `focusedSegmentRef.current is null` errors, we add optional chaining to this.

Related issue: AWSUI-21557

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


just-boris:
Are you sure this is the right fix for the issue? In your approach it calls `clearHighlightedSegment` when this ref is empty? Maybe we should do nothing instead (go into early return branch)?

What are the steps to reproduce the issue? 

And ideally, this should be covered by a test too.

rubencarvalho:
> Are you sure this is the right fix for the issue? In your approach it calls `clearHighlightedSegment` when this ref is empty, maybe we should do nothing instead (go into early return branch)?
> 
> What are the steps to reproduce the issue?
> 
> And ideally, this should be covered by a test too.

I can't reproduce this and I could not understand how this can happen (following the logic on our code). It must be a very corner case of a race condition where the segment is no longer highlighted but the popover is still present.

But that said, if `focusedSegmentRef` is the active segment, why would there be a popover if the segment wasn’t active? That's why I replaced the `!` with `?`, and if there isn’t a `focusedSegmentRef`, I think it is safe to `clearHighlightedSegment` (which also closes the popover).

Regarding the test, I'm not sure how to test the condition I changed, as I'm not sure there's a way of replicating this programatically. Happy to take a suggestion.

just-boris:
I am not sure if the expected behavior to go into `then` or `else` branch. It is a real 50/50

I think the best option is to go back to the original reporter and ask for details. Their current report is too vague.

just-boris:
Closing this one for now, until we get more details about the bug report

PR #1343, opened 2023-07-18
fix: Fix visual bugs in drawer toggle buttons

just-boris:
### Description

Fix focus and hover style issues with drawer triggers. 

Updated the non-drawer toggle buttons too, to have unified implementation

### How has this been tested?

* PR build
* Visual regression tests in my pipeline

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/app-layout/__integ__/app-layout-focus-delegation.test.ts
```diff
@@ -91,13 +91,14 @@ function setupTest(
           'drawers focus toggles between open and close buttons',
           setupTest(
             async page => {
-              await page.click(wrapper.findDrawersTriggers().get(2).toSelector());
+              const triggerSelector = wrapper.findDrawerTriggerById('pro-help').toSelector();
```
just-boris, line 94: `:nth-child` does not work anymore, because every button has a wrapper now. 

Since this feature is not in production yet, we can do such breaking changes

src/app-layout/drawer/styles.scss
```diff
@@ -48,13 +49,11 @@ $drawer-z-index-mobile: 1001;
   }
   .drawer-closed > & {
     width: constants.$sidebar-size-closed;
-    &:hover {
-      background: awsui.$color-background-layout-panel-hover;
-    }
-    &.non-interactive {
+    &.drawer-content-clickable {
+      cursor: pointer;
+      color: awsui.$color-text-interactive-default;
```
just-boris, line 54: Made some refactoring in this file to more clearly connect styles and their reasons.

Why is there a pointer cursor? Not because the drawer is closed, but because it is clickable

just-boris: There could be closed non-clickable drawers and the new CSS structure allows that

src/app-layout/drawer/styles.scss
```diff
@@ -63,34 +62,26 @@ $drawer-z-index-mobile: 1001;
   }
 }
 
-.trigger {
-  .drawer-content > .drawer-triggers > & {
```
just-boris, line 67: Before this PR, app layout used our button component and this line overrode button styles. 

Overriding styles is bad even if done for a good reason, it still may cause issues. 

In this PR I replaced it with custom styles on top of native `<button>` tag (see the `<ToggleButton>` component) which has the correct styles from the start, without overrides

src/app-layout/mobile-toolbar/styles.scss
```diff
@@ -12,7 +12,6 @@
 }
 
 .mobile-bar {
-  /* stylelint-disable-next-line plugin/no-unsupported-browser-features */
```
just-boris, line 15: Not needed anymore with our current browser support

style-dictionary/utils/token-names.ts
```diff
@@ -331,7 +331,6 @@ export type ColorsTokenName =
   | 'colorForegroundControlDefault'
   | 'colorForegroundControlDisabled'
   | 'colorShadowDefault'
-  | 'colorShadowLayoutToggle'
```
just-boris, line 334: Replaced this token with `colorBorderLayout` rendering the same look

jperals:
Before this change the focus rings in Classic didn't look right, they were cut off:

Screenshot 2023-07-20 at 10 24 02

After this change they are shown inside the corresponding toggle button, which gets rid of this particular problem, but there are still a couple things that I think could be improved:

1. They have some margin to the container horizontally but stick to the border vertically:

Screenshot 2023-07-20 at 10 22 40

2. Maybe more importantly, they are hard to distinguish over the currently active item:

Screenshot 2023-07-20 at 10 14 44

I see two possible ways to fix this:

1. Fine tune the dimensions to have some margin both horizontally and vertically, and use the bright variant of the focus ring color for the currently active item

2. Bring the focus ring back to be outside of the drawer toggle buttons, fixing the cutoff issue. As a quick experiment, by removing `overflow: auto` from the `drawer-content` they look like this:

Screenshot 2023-07-20 at 10 16 38

just-boris:
The plan is to reuse this treatment from calendar

image

jperals (approved):

PR #1273, opened 2023-07-03
feat: Add fit-height property to charts

pan-kot:
### Description

The new `fitHeight` property on all charts allows the content area to grow or shrink to fit the outer container. The `height` (mixed charts, area chart) and `size` (pie chart) properties control the minimal allowed height of the content in that case.

### How has this been tested?

New screenshot tests.

<details>
   <summary>Review checklist</summary>

_The following items are to be evaluated by the author(s) and the reviewer(s)._

#### Correctness

- _Changes include appropriate documentation updates._
- _Changes are backward-compatible if not indicated, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#public-apis)._
- _Changes do not include unsupported browser features, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#browsers-support)._
- _Changes were manually tested for accessibility, see [accessibility guidelines](https://cloudscape.design/foundation/core-principles/accessibility/)._

#### Security

- _If the code handles URLs: all URLs are validated through [the `checkSafeUrl` function](https://github.com/cloudscape-design/components/blob/main/src/internal/utils/check-safe-url.ts)._

#### Testing

- _Changes are covered with new/existing unit tests?_
- _Changes are covered with new/existing integration tests?_
</details>

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1355, opened 2023-07-20
feat: Charts fit-height

pan-kot:
### Description

Re-introduction of https://github.com/cloudscape-design/components/pull/1273 reverted here: https://github.com/cloudscape-design/components/pull/1341

Updated the internal chart-wrapper and chart-container to have less performance impact on the charts not using the new property.

### How has this been tested?

Tested in dev-pipeline.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/area-chart/model/use-chart-model.ts
+++ b/src/area-chart/model/use-chart-model.ts
@@ -1,7 +1,7 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 import { AreaChartProps } from '../interfaces';
-import React, { useEffect, useMemo, useRef, RefObject, MouseEvent } from 'react';
+import React, { useEffect, useMemo, useRef, RefObject, MouseEvent, useState } from 'react';
 import { findClosest, circleIndex } from './utils';
 
 import { nodeContains } from '../../internal/utils/dom';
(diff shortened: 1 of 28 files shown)


timogasda (approved):

src/pie-chart/pie-chart.tsx
```diff
@@ -281,60 +296,76 @@ export default <T extends PieChartProps.Datum>({
...
-          variant={variant}
-          focusedSegmentRef={focusedSegmentRef}
-          popoverTrackRef={popoverTrackRef}
-          highlightedSegment={highlightedSegment}
-          segmentAriaRoleDescription={i18nStrings?.segmentAriaRoleDescription}
-          onMouseDown={onMouseDown}
-          onMouseOver={onMouseOver}
+        <ChartPlot
+          ref={plotRef}
+          width="100%"
```
just-boris, line 311: Looks like this value is always this constant now, can be simplified

src/internal/components/chart-wrapper/index.tsx
```diff
@@ -35,25 +36,59 @@ export const ChartWrapper = forwardRef(
...
+        >
+          {defaultFilter}
+          {additionalFilters}
+        </InternalSpaceBetween>
+      </InternalBox>
+    );
+
+    const legendNode = legend && <InternalBox margin={{ top: 'm' }}>{legend}</InternalBox>;
+
+    if (fitHeight) {
```
just-boris, line 61: I see many `if(fitHeight) {...}` code duplication patterns now. Does it have to do anything with performance optimization?

src/area-chart/model/use-chart-model.ts
```diff
@@ -55,6 +58,14 @@ export default function useChartModel<T extends AreaChartProps.DataTypes>({
   const containerRef = useRef<HTMLDivElement>(null);
   const verticalMarkerRef = useRef<SVGLineElement>(null);
 
+  const plotMeasureRef = useRef<SVGLineElement>(null);
+  const [measuredHeight, setHeight] = useState(0);
+  useResizeObserver(
+    () => plotMeasureRef.current,
+    entry => fitHeight && setHeight(entry.borderBoxHeight)
```
just-boris, line 65: I am guessing the conditional `fitHeight && setHeight(...)` is the one that improves performance. When false, there are no extra state updates and no-op resize observer does not really cost anything.

Is this correct?

just-boris (commented):
non-blocking comments

src/area-chart/model/use-chart-model.ts
timogasda: Yes, I think this combined with the `fitHeight` if conditionals is giving the desired effect. But we can wait for @pan-kot to confirm.

pan-kot: Yes, that is correct. The hooks are still there and the resize observer is created yet the associated state is never updated when fitHeight=false. The performance improvements come from having less nodes being rendered when fitHeight=false.

PR #1364, opened 2023-07-25
feat: Autolabel table and cards with header

Al-Dani:
### Description

Scrollable tables and cards create an unlabelled tabIndex, which leads to screen readers reading out all the contents. This can be fixed manually by providing `ariaLabels.tableLabel(cardsLabel)` property, but it is not commonly used. To simplify this, we will automatically label table (cards) with a header. 
Because header is a slot, we will use context to provide id to the `<Header>` component, assuming that `<Header>` is the recommended way to entitle collections.
For the custom headers it's still required to provide `ariaLabels.tableLabel(cardsLabel)` property

Related links, issue #, if available: AWSUI-21380

### How has this been tested?

New unit tests are added, tested with VO+Safari

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/box/internal.tsx
+++ b/src/box/internal.tsx
@@ -1,12 +1,11 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
-import React, { useContext } from 'react';
+import React from 'react';
 import { getBaseProps } from '../internal/base-component';
 import clsx from 'clsx';
 import styles from './styles.css.js';
 import { BoxProps } from './interfaces';
 import { InternalBaseComponentProps } from '../internal/hooks/use-base-component';
-import { ScrollbarLabelContext } from '../internal/context/scrollbar-label-context';
 
 type InternalBoxProps = BoxProps & InternalBaseComponentProps;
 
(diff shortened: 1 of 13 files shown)


src/box/internal.tsx
```diff
@@ -24,7 +25,12 @@ export default function InternalBox({
   __internalRootRef = null,
   ...props
 }: InternalBoxProps) {
-  const baseProps = getBaseProps(props);
+  const wrapperHeadingId = useContext(ScrollbarLabelContext);
```
timogasda, line 28: Box is one of our most used components in the system. I'm a little scared of adding a React context to it. Have you checked this for potential performance degredation?

src/internal/context/scrollbar-label-context.ts
```diff
@@ -0,0 +1,5 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import { createContext } from 'react';
+
+export const ScrollbarLabelContext = createContext<string | undefined>(undefined);
```
timogasda, line 5: Why is this called `ScrollbarLabelContext`? It sounds like we're adding a label to a scrollbar.

Al-Dani: Naming is difficult! We are adding a label to table / cards kinda because of a scrollbar. But maybe, `CollectionLabelContext` sounds better

src/box/internal.tsx
Al-Dani: Box is definitely tricky, and there are more concerns (like pointing aria to a non-existing id). So I changed the implementation to provide an id from the header component.
But it might still worth checking for performance issues in a dev pipeline first

YueyingLu (dismissed):

timogasda (dismissed):

avinashbot (approved):
Rebase, implicitly approving

PR #1365, opened 2023-07-25
feat: Table roles helper and role="grid" for editable table

pan-kot:
### Description

Encapsulated table's semantic structure in a helper module and changed the table's role to "grid" for tables with editable cells.

### How has this been tested?

New unit tests to verify "grid" role for tables with editable cells.

Manual testing with VO+Chrome, VO+Safari, NVDA+Chrome, NVDA+Firefox.

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/table/table-role/table-role-helper.ts
```diff
@@ -0,0 +1,98 @@
...
+  descending: 'descending',
+} as const;
+const getAriaSort = (sortingState: SortingStatus) => stateToAriaSort[sortingState];
+
+/**
+ * Depending on its content the table can have different semantic representation which includes the
+ * ARIA role of the table component ("table", "grid", "treegrid") but also roles and other semantic attributes
+ * of the child elements. The TableRole helper encapsulates table's semantic structure.
+ */
+export class TableRole {
```
just-boris, line 18: We use React and this encourages us to use pure functions instead of mutable objects.

Especially, objects that you propagate across components. Does not happen in your code, because all `assign` functions do not modify `this`, but it paves the way for bugs.

Safer option would be to propagate the role directly, without helper

just-boris: On a higher level the it looks like this task is suitable to be solved with contexts

```js
<TableContext role={hasEditableCells ? 'grid' : 'table'} />
```

```js
const {getTableRowProps} = useTableContext();
```

---

We have several more props passing down the same way, so maybe table context may hold more than just this

pan-kot: I think contexts is an overkill here - instead I am passing a stable object's reference - it is simpler and no pub/sub is required. What do you think if I make the tableRole property readonly?

just-boris: You could just propagate a string role instead of the object

pan-kot: I think of that class as a way of binding the helper functions to the tableRole value - surely can have a module with the functions that would accept the tableRole as an additional argument, but I like the semantics of the helper more. I think it is a good idea to ensure the internal state never changes after initialisation.

just-boris: Classes are not suitable for this job. If you want to know more details, welcome to my DM

pan-kot: I updated the helper to be a function instead, please check.

just-boris:
I do not understand one thing

How is this better:

```js
const tableRole = createTableRoleHelper('table');
tableRole.assignTableRowProps({ firstIndex, rowIndex });
```


than this

```js
getRowProps({role: 'table', firstIndex, rowIndex })
```

I am trying really hard to understand the original motivation for grouping this, but I do not see it

pan-kot:
> I do not understand one thing
> 
> How is this better:
> 
> ```js
> const tableRole = createTableRoleHelper('table');
> tableRole.assignTableRowProps({ firstIndex, rowIndex });
> ```
> 
> than this
> 
> ```js
> getRowProps({role: 'table', firstIndex, rowIndex })
> ```
> 
> I am trying really hard to understand the original motivation for grouping this, but I do not see it

Same as you suggested above with the contexts. When there is a single helper it can be instantiated with more properties.
I will update it to the plain set of functions for now.

src/table/table-role/table-role-helper.ts
```diff
@@ -0,0 +1,80 @@
...
+): React.ThHTMLAttributes<HTMLTableCellElement> {
+  nativeProps.scope = 'row';
+
+  return nativeProps;
+}
+
+export function assignTableCellProps(
+  options: { tableRole: TableRole },
+  nativeProps: React.TdHTMLAttributes<HTMLTableCellElement> = {}
+): React.TdHTMLAttributes<HTMLTableCellElement> {
```
just-boris, line 74: On `Object.assign()`, the first argument receives new properties not the last. Let's avoid confusing people and stick to the convention.

Ideally, we should avoid using mutable objects at all, like this

```js
{...getTableProps(tableRole)} // the object gets inside that function, no chances of overwriting something that we should not
```

just-boris: Or for table cell

```
nativeAttributes.role = getTableCellRole(tableRole)
```

pan-kot: Sure, can change the attributes order. The assignment is made to avoid objects cloning which is then has to be done for all cells and is unnecessary. Sure, can create 10 utility methods to set all attributes individually but the point was to make a single place where the semantic attributes are assigned.

just-boris: You still can group all of this attribute management in a single file without mutating objects.

The ask is only about avoiding mutations, the grouping of the code is up to you

pan-kot: done

just-boris (approved):

PR #1367, opened 2023-07-25
chore: Emit `externalLinkInteracted` events for buttons

connorlanigan:
### Description

We're already emitting the `externalLinkInteracted` in the Link component, when the `external` prop is enabled: https://github.com/cloudscape-design/components/blob/ea832bbceffa379e959e4872c2dda7c97828a2c4/src/link/internal.tsx#L72-L109

There are also instances where the Button component is used to build an external link. The Button component does not have an `external` prop, but it allows to set the `iconName` and `target` to replicate the behaviour.  
When these properties are set, we will now send the same `externalLinkInteracted` event as in the Link component.

### How has this been tested?

Added unit test

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/button/__tests__/internal.test.tsx
```diff
@@ -23,3 +29,65 @@ test('supports __iconClass property', () => {
...
+  });
+
+  test('does not send any metrics when not in a funnel context', () => {
+    const { container } = render(<Button iconName="external" target="_blank" href="https://example.com" />);
+    createWrapper(container).findButton()!.click();
+
+    expect(FunnelMetrics.externalLinkInteracted).not.toHaveBeenCalled();
+  });
+
+  test('sends an externalLinkInteracted metric within a Funnel Context with iconName=external', () => {
```
rubencarvalho, line 46: Could we maybe add a test for when `iconName="external"` but no `href` is present?

rubencarvalho (changes requested):
LGTM, just left a comment about an additional test, to cover the case when `isAnchor` is `false` :)

rubencarvalho (approved):

PR #1374, opened 2023-07-26
fix: Merge mouseleave detection logic

just-boris:
### Description

Fixed two major issues

1. `event.relatedTarget` is not always a DOM node. Sometimes it is `window` and this breaks `node.contains(event.relatedTarget)`
2. There are two concurrent mouse leave handlers, for the SVG element and popover. When both fire same time, the second fails with an error, because the segment is already hidden. The new implementation is idempotent

Also did some additional refactoring, see the comments inline

Related links, issue #, if available: AWSUI-21557

### How has this been tested?

1. Added more tests
2. Checked other chart components, they already use `nodeContains`, so we are good

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/pie-chart/__tests__/pie-chart.test.tsx
+++ b/src/pie-chart/__tests__/pie-chart.test.tsx
@@ -528,6 +528,29 @@ describe('Details popover', () => {
     expect(detailPopover?.findHeader()?.getElement()).toHaveTextContent(defaultData[0].title);
   });
 
+  test('allow mouse to be move between segment and popover ', () => {
+    const { wrapper } = renderPieChart(<PieChart data={defaultData} />);
+    fireEvent.mouseOver(wrapper!.findSegments()[0].getElement());
+
+    expect(wrapper.findDetailPopover()).toBeTruthy();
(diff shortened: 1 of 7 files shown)


src/pie-chart/__tests__/pie-chart.test.tsx
```diff
@@ -528,6 +528,29 @@ describe('Details popover', () => {
     expect(detailPopover?.findHeader()?.getElement()).toHaveTextContent(defaultData[0].title);
   });
 
+  test('allow mouse to be move between segment and popover ', () => {
```
jperals, line 531: Nitpick: typo in "to be move" (could be either "to move" or "to be moved" instead)

jperals (approved):

src/pie-chart/__tests__/pie-chart.test.tsx
just-boris: grammar is hard. Since this issue is kinda critical, I am going to merge and fix it next time

PR #1375, opened 2023-07-26
chore: Use synthetic focus tracking in substeps

connorlanigan:
### Description

This PR fixes an issue with how the browser focus is reported for some form elements.  
When the user clicks on the label(!) of a form element (such as the label of a radio item, as opposed to the radio item itself), the focus is not directly moved to the form element. Rather, when the mouse button is pressed down, the focus is removed from the page (i.e. a `blur` element is fired, and `document.activeElement` points to e.g. the body element). Then, when the mouse button is released, the form element is focused.  
This means that we erroneously emitted `funnelSubStepComplete` and `funnelSubStepStart` events when interacting with e.g. radio items inside an already focused substep.

At its core, this PR introduces a separate state for tracking which substep is focused (`isFocusedSubStep`), that does not directly reflect the browser focus. It tracks the browser focus closely, but when the mouse button is pressed, the `isFocusedSubStep` is not updated yet. Instead, it is only updated when the mouse button is released again. At that point, we call `funnelSubStepComplete` if needed - this replicates what was previously the `onBlur` event.

Keyboard interaction behaves as before.

### How has this been tested?

Tested manually in the dev pages

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/internal/analytics/components/analytics-funnel.tsx
```diff
@@ -247,5 +257,49 @@ export const AnalyticsFunnelSubStep = ({ children }: AnalyticsFunnelSubStepProps
...
+        /*
+         Run this substep's own focus cleanup function if another substep
+         hasn't already done it for us.
+         */
+        focusCleanupFunction.current?.();
+      }
+    };
+    window.addEventListener('mousedown', onMouseDown);
+    window.addEventListener('mouseup', onMouseUp);
+    return () => {
```
pan-kot, line 290: Not sure how unlikely is that but if any of the effect dependencies change after mouse it pressed and before it is released, the mousePressed state can remain true which can lead to the wrong step being reported. Should the mousePresses state be reset in the effect cleanup?

src/internal/analytics/components/analytics-funnel.tsx
```diff
@@ -247,5 +257,49 @@ export const AnalyticsFunnelSubStep = ({ children }: AnalyticsFunnelSubStepProps
...
+      /*
+        Some mouse events result in an element being focused. However,
+        this happens only _after_ the onMouseUp event. We yield the
+        event loop here, so that `document.activeElement` has the
+        correct new value.      
+      */
+      await new Promise(r => setTimeout(r, 1));
+
+      if (!subStepRef.current || !subStepRef.current.contains(document.activeElement)) {
+        isFocusedSubStep.current = false;
```
pan-kot, line 279: Is this necessary to set this one to false here? I would imagine in the next similar flow it is going to be set to true anyways. If not, wouldn't it make sense to reset this outside of the condition block and also in the effect cleanup function?

src/internal/analytics/components/analytics-funnel.tsx
connorlanigan: The `mousePressed` state has the same scope/lifecycle as the effect, so if the effect gets cleaned up because the component unmounts then the `mousePressed` state will also disappear. If the effect is cleaned up only because the dependencies changed, then the new run of the effect will immediately attach new event listeners. So I don't think there is a risk that the `mousePressed` state will get stuck.

Additionally, if we would reset the mousePressed state in the cleanup, and the effect is re-run (e.g. because of strict mode in development, or in React's upcoming "Offscreen rendering"), then the state will be incorrect, because we cannot determine the correct value at the start of the effect.

src/internal/analytics/components/analytics-funnel.tsx
connorlanigan: `isFocusedSubStep.current` is checked in a separate file when an element in the substep is focused, so that the `onFocus` handling runs when the focus _enters_ the substep, but not when it moves around _inside_ the substep:
https://github.com/cloudscape-design/components/blob/3482439f8affd1a0d69be1009267db229b800dd0/src/internal/analytics/hooks/use-funnel.ts#L44-L54

If we do not set the variable to `false` here, then the next time the user enters this substep we will miss emitting the `funnelSubStepStart` event.

src/internal/analytics/components/analytics-funnel.tsx
pan-kot: alright, sounds good 👍

src/internal/analytics/components/analytics-funnel.tsx
pan-kot: Ah, so when there was a click inside the isFocusedSubStep shall remain true so that the focus is ignored?

pan-kot (approved):

src/internal/analytics/components/analytics-funnel.tsx
connorlanigan: Yes. The `onFocus` event listener is attached to the whole substep, and it fires every time that any element inside the container receives the focus, because the `onFocus` event bubbles (only in React). So we want to ignore all subsequent `onFocus` events if the focus is _already_ somewhere inside the substep.

PR #1378, opened 2023-07-27
fix: Ensure we generate valid ids

gethinwebster:
### Description

ids should start with a letter, so ensure that we do that by default,
and enable validity checking in our tests.

### How has this been tested?

Updated tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/hooks/use-unique-id/index.ts
+++ b/src/internal/hooks/use-unique-id/index.ts
@@ -14,5 +14,5 @@ const useIdFallback = () => {
 const useId: typeof useIdFallback = (React as any).useId ?? useIdFallback;
 
 export function useUniqueId(prefix?: string) {
-  return `${prefix ? prefix : ''}` + useId();
+  return `${prefix ? prefix : 'i'}` + useId();
 }
--- a/src/internal/hooks/use-unique-id/index.ts
+++ b/src/internal/hooks/use-unique-id/index.ts
@@ -14,5 +14,6 @@ const useIdFallback = () => {
 const useId: typeof useIdFallback = (React as any).useId ?? useIdFallback;
 
 export function useUniqueId(prefix?: string) {
+  // if no prefix provided, add a letter at the beginning to ensure valid HTML id
   return `${prefix ? prefix : 'i'}` + useId();
 }
--- a/src/__a11y__/to-validate-a11y.ts
+++ b/src/__a11y__/to-validate-a11y.ts
@@ -19,8 +19,6 @@ Axe.configure(spec);
 const htmlValidator = new HtmlValidate({
   extends: ['html-validate:recommended'],
   rules: {
-    // set relaxed to exclude error that id does not begin with letter
-    'valid-id': ['error', { relaxed: true }],
     'no-inline-style': 'off',
     'prefer-native-element': ['error', { exclude: ['listbox', 'button', 'region'] }],
     //TODO: revisit 'no-redundant-for' when fixing AWSUI-18968


src/internal/hooks/use-unique-id/index.ts
```diff
@@ -14,5 +14,5 @@ const useIdFallback = () => {
 const useId: typeof useIdFallback = (React as any).useId ?? useIdFallback;
 
 export function useUniqueId(prefix?: string) {
-  return `${prefix ? prefix : ''}` + useId();
+  return `${prefix ? prefix : 'i'}` + useId();
```
johannes-weber, line 17: I recommend adding a comment as to why the unique id must start with a letter. This might help future readers avoid confusion (Why is the unique id prefixed with an `i`?).

johannes-weber (changes requested):

just-boris:
Checked react docs, they suggest to drop `useId` return value directly into the ID attribute without prefixes: https://react.dev/reference/react/useId#generating-unique-ids-for-accessibility-attributes

In the spec, there is no requirement on starting with a letter: https://html.spec.whatwg.org/multipage/dom.html#global-attributes:the-id-attribute-2

> When specified on [HTML elements](https://html.spec.whatwg.org/multipage/infrastructure.html#html-elements), the [id](https://html.spec.whatwg.org/multipage/dom.html#the-id-attribute) attribute value must be unique amongst all the [IDs](https://dom.spec.whatwg.org/#concept-id) in the element's [tree](https://dom.spec.whatwg.org/#concept-tree) and must contain at least one character. The value must not contain any [ASCII whitespace](https://infra.spec.whatwg.org/#ascii-whitespace).

There needs to be at least one character anywhere, but not the first. I am guessing that the react team is aware about this requirement and their ids always include a character.


Are you sure there is a real issue?

gethinwebster:
```
Technically, the value for an id attribute may contain any character, except [whitespace](https://developer.mozilla.org/en-US/docs/Glossary/Whitespace) characters. However, to avoid inadvertent errors, only [ASCII](https://developer.mozilla.org/en-US/docs/Glossary/ASCII) letters, digits, '_', and '-' should be used, and the value for an id attribute should start with a letter.
```
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id

And certainly in Chrome at least, e.g. `document.querySelector('#1a')` throws an error "'#1a' is not a valid selector"

gethinwebster:
I guess not strictly necessary though. It was causing me some friction while trying to debug another issue, but I guess I can live with it as it is.

just-boris:
Interesting. This is not an issue with the id, but with a querySelector:

```
document.querySelector('#123'); // Failed to execute 'querySelector' on 'Document': '#123' is not a valid selector.
document.getElementById('123'); // works
document.querySelector('#' + CSS.escape('123')); // also works
```

We already escape selectors in some places: https://github.com/search?q=repo%3Acloudscape-design%2Fcomponents+CSS.escape&type=code

As well as in test-utils: https://github.com/search?q=repo%3Acloudscape-design/components%20escapeSelector&type=code


I think we should continue going the escape approach every time we need to select an id, because it is safer

just-boris:
I also found the origins of the MDN recommendation: https://github.com/mdn/content/blob/009fcd49671ef08b6ccd808a38f19e6049936907/files/en-us/web/html/global_attributes/id/index.md


> Note: Using characters except {{glossary("ASCII")}} letters, digits, '_', '-' and '.' may cause compatibility problems, as they weren't allowed in HTML 4. Though this restriction has been lifted in {{glossary("HTML5")}}, an ID should start with a letter for compatibility.

It was needed to support legacy modes and browsers, but not needed in environments we support

gethinwebster:
Ok, cancelling, seems like we're ok as we are

PR #1396, opened 2023-08-01
feat: Add AnchorNavigation component

rubencarvalho:
### Description

Adding the new Anchor Navigation component.

### How has this been tested?

Unit & integration tests. 

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/anchor-navigation/interfaces.ts
```diff
@@ -0,0 +1,67 @@
...
+  /**
+   * List of anchors.
+   */
+  anchors: AnchorNavigationProps.Anchor[];
+
+  /**
+   * Specifies the active anchor href.
+   * For using the component in a controlled manner, use together with 'disableTracking'.
+   */
+  activehref?: string;
```
jperals, line 26: Is the "h" intentionally lowercase?

The Side navigation component has `activeHref` with uppercase "H".

rubencarvalho: Not intentional! Fixed it.

jperals:
I think it would be good to add integration tests for features like scrolling on click and the scroll spy, as these are critical features which otherwise are not covered.

src/anchor-navigation/internal.tsx
```diff
@@ -0,0 +1,117 @@
...
+    setDisableTracking(disableTracking);
+
+    if (activeHref) {
+      setCurrentActiveHref(activeHref);
+    }
+  }, [setDisableTracking, setCurrentActiveHref, disableTracking, activeHref]);
+
+  useEffect(() => {
+    if (currentActiveHref) {
+      onActiveAnchorChangeHandler(memoizedAnchors.find(anchor => anchor.href === currentActiveHref));
```
jperals, line 52: It would be nice to have a test that asserts that `onActiveAnchorChange` is called once and only once on anchor change.

src/anchor-navigation/internal.tsx
```diff
@@ -0,0 +1,117 @@
...
+  onActiveAnchorChange,
+  disableTracking = false,
+  activeHref = '',
+  __internalRootRef = null,
+  ...props
+}: AnchorNavigationProps & InternalBaseComponentProps) {
+  const baseProps = getBaseProps(props);
+
+  const hrefs = useMemo(() => anchors.map(anchor => anchor.href), [anchors]);
+  const memoizedAnchors = useMemo(() => anchors, [anchors]);
```
jperals, line 27: What is this line for?

src/anchor-navigation/internal.tsx
```diff
@@ -0,0 +1,117 @@
...
+          );
+        })}
+      </ol>
+    </nav>
+  );
+}
+
+interface AnchorProps {
+  anchor: AnchorNavigationProps.Anchor;
+  fireFollow: (anchor: AnchorNavigationProps.Anchor, event: React.SyntheticEvent | Event) => void;
```
jperals, line 82: Why `fireFollow` and not `onFollow`?

src/anchor-navigation/scroll-spy.tsx
```diff
@@ -0,0 +1,95 @@
...
+  hrefs,
+}: {
+  hrefs: string[];
+}): [
+  string | undefined,
+  React.Dispatch<React.SetStateAction<string | undefined>>,
+  React.Dispatch<React.SetStateAction<boolean>>
+] {
+  const [currentHref, setCurrentHref] = useState<string>();
+  const [scrollY, setScrollY] = useState(isBrowser ? window.pageYOffset : 0);
```
jperals, line 17: Because `scrollY` is a state variable, every time the user scrolls the component will re-render, even if it does not need to. Is there a way we can optimize this?

src/anchor-navigation/scroll-spy.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const [disableTracking, setDisableTracking] = useState(false);
```
jperals, line 18: Could we pass `disableTracking` directly as a prop so we don't need this extra state variable? This seems unnecessary duplication.

src/anchor-navigation/scroll-spy.tsx
```diff
@@ -0,0 +1,95 @@
...
+    }
+  }, [updateScroll, scrollY]);
+
+  useEffect(() => {
+    if (disableTracking || !isBrowser) {
+      return;
+    }
+
+    // Main logic to get the new active href
+    let newCurrentHref;
```
jperals, line 81: Minor: no need to use a `let` and reassign, you can just use `setNewCurrentRef` twice.

src/anchor-navigation/scroll-spy.tsx
```diff
@@ -0,0 +1,95 @@
...
+    const lastAnchorId = hrefs[hrefs.length - 1]?.slice(1);
+    const lastAnchorElementExists = document.getElementById(lastAnchorId);
+    return !!lastAnchorElementExists && scrollY >= Math.floor(document.body.scrollHeight - window.innerHeight);
+  }, [scrollY, hrefs]);
+
+  // Find the href for which the element is within the viewport plus EXTRA_OFFSET
+  const findHrefInView = useCallback(() => {
+    return hrefs.find(href => {
+      const rect = getRectByHref(href);
+      return rect && rect.bottom <= window.innerHeight && rect.bottom - EXTRA_OFFSET >= 0;
```
jperals, line 47: I've noticed that when clicking on an anchor for which the content is shorter than the viewport, the next section will be activated instead (you can try this on the examples you provide in this PR using a big monitor: clicking e.g, on section 1.1 will highlight section 1.1.1).

Note that `rect.bottom` returns the coordinate _from the top of the viewport_ (not from the bottom) of the element's bottom edge, therefore `rect.bottom - EXTRA_OFFSET` will be negative for the element that we scrolled to the top. This might be leading to this function not returning any element and then falling back to `findLastHrefInView` as per the logic in line 86 below.

rubencarvalho: I've noticed testing. `EXTRA_OFFSET` is now the `scrollSpyOffset`, which should fix this.
I'll make sure to add integration tests covering these edge cases.

src/anchor-navigation/scroll-spy.tsx
```diff
@@ -0,0 +1,95 @@
...
+      window.addEventListener('scroll', updateScroll, {
+        capture: false,
+        passive: true,
+      });
+
+      return () => {
+        window.removeEventListener('scroll', updateScroll);
+      };
+    }
+  }, [updateScroll, scrollY]);
```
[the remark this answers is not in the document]
rubencarvalho: No, cleaned up!

src/anchor-navigation/internal.tsx
rubencarvalho: Was trying to follow the convention used in SideNavigation, but it was also confusing to me so I renamed it! 👍

src/anchor-navigation/internal.tsx
rubencarvalho: TODO

pages/anchor-navigation/expandable.page.tsx
```diff
@@ -0,0 +1,65 @@
...
+      <ScreenshotArea>
+        <SpaceBetween size="l">
+          <div className={styles['content-grid']}>
+            <TextContent />
+            <div>
+              <div className={styles['anchor-navigation']}>
+                <ExpandableSection variant="navigation" headerText="On this page">
+                  <AnchorNavigation
+                    ariaLabelledby="anchor-nav-heading"
+                    anchors={[
```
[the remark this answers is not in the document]
rubencarvalho: Added

src/anchor-navigation/scroll-spy.tsx
```diff
@@ -0,0 +1,77 @@
...
+      setCurrentHref(hrefs[hrefs.length - 1]);
+    } else {
+      setCurrentHref(findHrefInView() || (scrollY > 0 ? findLastHrefInView() : undefined));
+    }
+  }, [disableScrollSpy, isPageBottom, findHrefInView, findLastHrefInView, hrefs]);
+
+  useEffect(() => {
+    if (isBrowser) {
+      handleScroll();
+      window.addEventListener('scroll', handleScroll, { passive: true });
```
jperals, line 69: Did you consider using the Intersection Observer API instead? https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API

rubencarvalho: I did consider it initially. In fact, I implemented it (https://github.com/cloudscape-design/components/pull/1396/commits/00a711b3bcaf3421f1e6fdcc1657e3c503ca1422) but ran into some issues that made the code more complex than I would've liked...

rubencarvalho: There were some edge cases (scrolled all the way to the bottom, in which the element didn't intersect the IO), lack of granularity, IO sometimes not being triggered (when being linked to a specific section of the page or quickly scrolling on the trackpad)

pages/anchor-navigation/expandable.page.tsx
```diff
@@ -0,0 +1,75 @@
...
+        <TextSample />
+        <h5 id="section-1-2-1-1">Section 1.2.1.1</h5>
+        <TextSample />
+      </div>
+    </SpaceBetween>
+  );
+};
+
+export default function ExpandableAnchorNavigation() {
+  const navigateToItem = (id: string) => {
```
jperals, line 35: Nitpick: this function could be in `utils.tsx` so it can be de-duplicated

src/anchor-navigation/internal.tsx
```diff
@@ -0,0 +1,112 @@
...
+  const hrefs = useMemo(() => anchors.map(anchor => anchor.href), [anchors]);
+
+  const onFollowHandler = useCallback(
+    (anchor: AnchorNavigationProps.Anchor, sourceEvent: React.SyntheticEvent | Event) => {
+      fireCancelableEvent(onFollow, anchor, sourceEvent);
+    },
+    [onFollow]
+  );
+
+  const onActiveHrefChangeHandler = useStableCallback((newActiveAnchor: AnchorNavigationProps.Anchor | undefined) => {
```
jperals, line 35: Is this function necessary? Could we just call `fireNonCancellableEvent` directly on line 47 below instead?

src/anchor-navigation/scroll-spy.tsx
```diff
@@ -0,0 +1,79 @@
...
+  // Check if we're scrolled to the bottom of the page
+  const isPageBottom = useCallback(() => {
+    return lastAnchorElementExists && window.scrollY >= Math.floor(document.body.scrollHeight - window.innerHeight);
+  }, [lastAnchorElementExists]);
+
+  // Find the href for which the element is within the viewport
+  const findHrefInView = useCallback(() => {
+    return hrefs.find(href => {
+      const rect = getRectByHref(href);
+      return rect && rect.bottom <= window.innerHeight && rect.bottom - scrollSpyOffset >= 0;
```
jperals, line 38: Why is it
```rect.bottom - scrollSpyOffset >= 0``` and not
```rect.top - scrollSpyOffset >= 0```?
(or `rect.top >= scrollSpyOffset` for simplicity)

Comment (rubencarvalho) [reply src/anchor-navigation/scroll-spy.tsx]: Simplified it!

File: src/anchor-navigation/__integ__/anchor-navigation.test.ts [hunk 15]
```diff
@@ -0,0 +1,71 @@
...
+
+  test(
+    'the correct aria-labelledby attribute is applied',
+    setupTest(async page => {
+      return expect(await page.getElementAttribute(wrapper.toSelector(), 'aria-labelledby')).toBe('anchor-nav-heading');
+    })
+  );
+
+  test(
+    'clicking in a anchor link makes the respective anchor item active',
```
jperals, line 41: Minor typo: "**an** anchor link" :)

src/anchor-navigation/scroll-spy.tsx
```diff
@@ -0,0 +1,79 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
jperals, line 1: _Almost_ every file in the repo that contains only a hook has the `use-` prefix in the file name. I'd suggest to follow this pattern here as well.

src/anchor-navigation/scroll-spy.tsx
```diff
@@ -0,0 +1,79 @@
...
+  const getRectByHref = useCallback(href => {
+    return document.getElementById(href.slice(1))?.getBoundingClientRect();
+  }, []);
+
+  // Check if we're scrolled to the bottom of the page
+  const isPageBottom = useCallback(() => {
+    return lastAnchorExists && Math.ceil(window.scrollY) >= Math.floor(document.body.scrollHeight - window.innerHeight);
+  }, [lastAnchorExists]);
+
+  // Find the href for which the element is within the viewport
```
jperals, line 37: The _first_ href, right?

src/anchor-navigation/scroll-spy.tsx
```diff
@@ -0,0 +1,79 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import { useCallback, useEffect, useState } from 'react';
+
+const isBrowser = typeof window !== 'undefined';
+
+export default function useScrollSpy({
```
jperals, line 7: Looks like this hook assumes that the hrefs, and by extension the anchors passed in in the props, are sorted by order of appearance in the page. Is this a safe assumption? Or should be either a) make it explicit in the API description or b) let the component sort them

rubencarvalho: Good point. I'd be for `a)`.
For `b)`, I can imagine this could add some computational overhead, especially if the list of `hrefs` is long or if the DOM is complex. Also, I feel it becomes _too much_ magic.

jperals (approved):

PR #1398, opened 2023-08-01
chore: Expose more component configuration in DOM metadata property

connorlanigan:
### Description

This change extends the `__awsuiMetadata__` property of 28 components to include information about their configured properties.

It builds on https://github.com/cloudscape-design/components/pull/1391, build failures are expected until the other PR is merged.

### How has this been tested?

Manual testing in the dev pages

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/breadcrumb-group/index.tsx
+++ b/src/breadcrumb-group/index.tsx
@@ -12,7 +12,7 @@ export default function BreadcrumbGroup<T extends BreadcrumbGroupProps.Item = Br
   items = [],
   ...props
 }: BreadcrumbGroupProps<T>) {
-  const baseComponentProps = useBaseComponent('BreadcrumbGroup');
+  const baseComponentProps = useBaseComponent('BreadcrumbGroup', { items });
   return <InternalBreadcrumbGroup items={items} {...props} {...baseComponentProps} />;
 }
 
--- a/src/cards/index.tsx
+++ b/src/cards/index.tsx
@@ -57,7 +57,7 @@ const Cards = React.forwardRef(function <T = any>(
   }: CardsProps<T>,
   ref: React.Ref<CardsProps.Ref>
 ) {
-  const { __internalRootRef } = useBaseComponent('Cards');
+  const { __internalRootRef } = useBaseComponent('Cards', { loading, selectedItems, selectionType, items, trackBy });
   const baseProps = getBaseProps(rest);
   const isRefresh = useVisualRefresh();
   const isMobile = useMobile();
--- a/src/checkbox/index.tsx
+++ b/src/checkbox/index.tsx
@@ -9,7 +9,7 @@ import useBaseComponent from '../internal/hooks/use-base-component';
 export { CheckboxProps };
 
 const Checkbox = React.forwardRef(({ ...props }: CheckboxProps, ref: React.Ref<CheckboxProps.Ref>) => {
-  const baseComponentProps = useBaseComponent('Checkbox');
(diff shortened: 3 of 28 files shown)


src/breadcrumb-group/index.tsx
```diff
@@ -12,7 +12,7 @@ export default function BreadcrumbGroup<T extends BreadcrumbGroupProps.Item = Br
   items = [],
   ...props
 }: BreadcrumbGroupProps<T>) {
-  const baseComponentProps = useBaseComponent('BreadcrumbGroup');
+  const baseComponentProps = useBaseComponent('BreadcrumbGroup', { items });
```
just-boris, line 15: Not sure about logging nested objects. What is the use-case on consumer end for it? Are we sure they won't get `[object Object]` logged in their metrics?

src/cards/index.tsx
```diff
@@ -57,7 +57,7 @@ const Cards = React.forwardRef(function <T = any>(
   }: CardsProps<T>,
   ref: React.Ref<CardsProps.Ref>
 ) {
-  const { __internalRootRef } = useBaseComponent('Cards');
+  const { __internalRootRef } = useBaseComponent('Cards', { loading, selectedItems, selectionType, items, trackBy });
```
just-boris, line 60: Same question about serialization friendliness here

`selectedItems` and `items` are nested and possibly even recursive objects
`trackBy` can be a function

src/checkbox/index.tsx
```diff
@@ -9,7 +9,7 @@ import useBaseComponent from '../internal/hooks/use-base-component';
 export { CheckboxProps };
 
 const Checkbox = React.forwardRef(({ ...props }: CheckboxProps, ref: React.Ref<CheckboxProps.Ref>) => {
-  const baseComponentProps = useBaseComponent('Checkbox');
+  const baseComponentProps = useBaseComponent('Checkbox', { disabled: props.disabled, checked: props.checked });
```
just-boris, line 12: Why is there no `name` prop? Are they going to distinguish individual controls using contexts?

src/side-navigation/index.tsx
```diff
@@ -22,7 +22,7 @@ export default function SideNavigation({
   onChange,
   ...props
 }: SideNavigationProps) {
-  const { __internalRootRef } = useBaseComponent('SideNavigation');
+  const { __internalRootRef } = useBaseComponent('SideNavigation', { items, activeHref });
```
just-boris, line 25: `items` may contain `info` prop which is JSX

just-boris (commented):
We need to figure out treatment for complex objects and functions

1. What is an object contains JSX (flashbar, radio group and others)?
2. What do we do with functions (`trackBy`)?
3. What do we do with arrays in general? How is the consumer side going to use them?

just-boris:
The list of comments is not excessive, I only called repeating patterns once

just-boris:
I see, I can approve this change, because it is internal, but there are some caveats making it harder to revisit if any issues occur.

1. Changes propagate to production within weeks, so there will be moment when this property may contain different data depending on their dependency freshness
2. These values may throw when passed into `JSON.stringify`, this may cause intermittent errors which will be hard to capture
3. Traversing some structures (side navigation items for example) may be tricky, because they are recursive. I believe we can simplify work for consumer end, preparing more digestable info (for example, annotating individual items)

just-boris (approved):

PR #1400, opened 2023-08-02
Refactor sticky columns

pan-kot:
### Description

A first step in refactoring table sticky columns. As part of refactoring made a fix to apply table wrapper styles synchronously same as it is done for columns. The goal of refactoring is to make the util lifecycle better explicit with uniform refs and callbacks initialisation and a single useEffect to synchronise properties.

### How has this been tested?

Existing tests + dev pipeline

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/table/sticky-columns/use-sticky-columns.ts
+++ b/src/table/sticky-columns/use-sticky-columns.ts
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import React, { useCallback, useEffect, useMemo, useRef } from 'react';
-import AsyncStore from '../../area-chart/async-store';
+import AsyncStore, { useSelector } from '../../area-chart/async-store';
 import { useStableEventHandler } from '../../internal/hooks/use-stable-event-handler';
 import { useResizeObserver } from '../../internal/hooks/container-queries';
 import clsx from 'clsx';
@@ -21,9 +21,6 @@ const MINIMUM_SCROLLABLE_SPACE = 148;
 
 export interface StickyColumnsModel {
   store: StickyColumnsStore;
-  style: {
-    wrapper?: React.CSSProperties;
-  };
   refs: {
     table: React.RefCallback<HTMLElement>;
     wrapper: React.RefCallback<HTMLElement>;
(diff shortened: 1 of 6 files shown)


src/table/sticky-columns/use-sticky-columns.ts
```diff
@@ -2,53 +2,32 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import React, { useCallback, useEffect, useMemo, useRef } from 'react';
-import AsyncStore from '../../area-chart/async-store';
+import AsyncStore, { useSelector } from '../../area-chart/async-store';
```
just-boris, line 5: This utility has not passed the formal team review: https://github.com/cloudscape-design/components/pull/998

This code is not allowed to change:

pan-kot: ok, reverted the change related to the use of async store

src/table/sticky-columns/utils.ts
```diff
@@ -0,0 +1,51 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { CellOffsets, StickyColumnsCellState, StickyColumnsProps, StickyColumnsWrapperState } from './interfaces';
+
+export function isCellStatesEqual(s1: null | StickyColumnsCellState, s2: null | StickyColumnsCellState): boolean {
```
just-boris, line 6: This could be a basic `shallowEqual` check, but because you chose an inconvenient data structure, there is now confusing and bug prone code

src/table/sticky-columns/utils.ts
```diff
@@ -0,0 +1,51 @@
...
+  }
+  lastColumnsWidths.reverse();
+
+  const stickyWidthLeft = firstColumnsWidths[props.stickyColumnsFirst - 1] ?? 0;
+  const stickyWidthRight = lastColumnsWidths[props.stickyColumnsLast - 1] ?? 0;
+  const offsets = props.visibleColumns.reduce(
+    (map, columnId, columnIndex) =>
+      map.set(columnId, {
+        first: firstColumnsWidths[columnIndex - 1] ?? 0,
+        last: lastColumnsWidths[props.visibleColumns.length - 1 - columnIndex - 1] ?? 0,
```
just-boris, line 45: So for non sticky columns there will be `0` value. I guess, it currently works fine, because this value is never used, but it is definitely a confusion when debugging the code and seeing these phantom values

src/table/sticky-columns/use-sticky-columns.ts
```diff
@@ -260,9 +221,11 @@ interface UpdateCellStylesProps {
 }
 
 export default class StickyColumnsStore extends AsyncStore<StickyColumnsState> {
```
just-boris, line 223: Because I do not understand how this store works, I cannot review this file properly.

Count this approval as "I read it" instead of "I understand it"

just-boris (approved):

src/table/sticky-columns/utils.ts
pan-kot: Agree, it would be nice to choose a better data structure going forward. The idea is that we don't need to iterate over all columns when only sticky columns are needed.

rubencarvalho (approved):

PR #1403, opened 2023-08-02
feat: Provide accessible name to menu in button-dropdown

Al-Dani:
### Description

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/button-dropdown/internal.tsx
```diff
@@ -259,7 +263,8 @@ const InternalButtonDropdown = React.forwardRef(
             position="static"
             role="menu"
             decreaseTopMargin={true}
-            ariaLabelledby={hasHeader ? headerId : undefined}
+            ariaLabel={ariaLabel}
+            ariaLabelledby={hasHeader ? headerId : shouldLabelWithTrigger ? triggerId : undefined}
```
pan-kot, line 267: What is header and why trigger ARIA label is preferred over the header?

pan-kot (commented):
I would add a few unit tests to ensure the labels are assigned correctly for icon and non-icon variants

src/button-dropdown/internal.tsx
Al-Dani: `header` is a combination of `title` and `description`, these properties are available only internally and can be used with `variant: 'navigation'` inside TopNavigation. 

Priorities are the following: 

- if there is a header inside dropdown, than it is used to label menu list
- else if we have aria-label - use it
- else if the variant is _not_ `icon` or `inline-icon` - label by trigger text

In case of a split button variant I assume there will be an aria-label, because trigger itself is an icon and we can't auto-label dropdown with it

pan-kot (approved):

PR #1409, opened 2023-08-03
refactor: Code editor setup

pan-kot:
### Description

Make code editor setup better granular for it to be reusable to support the readonly version of the component (see https://github.com/cloudscape-design/components/pull/1408).

### How has this been tested?

Existing tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/code-editor/__tests__/code-editor.test.tsx
+++ b/src/code-editor/__tests__/code-editor.test.tsx
@@ -219,8 +219,7 @@ describe('Code editor component', () => {
     const textarea = editorMock.renderer.textarea;
     editorMock.renderer.textarea = null as any;
 
-    renderCodeEditor({ ariaLabel: 'test aria label' });
-    // expect no failure
+    expect(renderCodeEditor({ ariaLabel: 'test aria label' })).not.toThrowError();
 
     editorMock.renderer.textarea = textarea;
 
--- a/src/code-editor/__tests__/code-editor.test.tsx
+++ b/src/code-editor/__tests__/code-editor.test.tsx
@@ -219,7 +219,7 @@ describe('Code editor component', () => {
     const textarea = editorMock.renderer.textarea;
     editorMock.renderer.textarea = null as any;
 
-    expect(renderCodeEditor({ ariaLabel: 'test aria label' })).not.toThrowError();
+    expect(() => renderCodeEditor({ ariaLabel: 'test aria label' })).not.toThrowError();
 
     editorMock.renderer.textarea = textarea;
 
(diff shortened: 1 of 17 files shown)


src/code-editor/setup-editor.ts
```diff
@@ -102,6 +66,53 @@ export function setupEditor(
     }
   });
 
+  // HACK: Annotations aren't cleared when editor is empty.
```
johannes-weber, line 69: Why aren't the cleared? Are there any references which helps future readers understand why the annotations aren't cleared when editor is empty?

src/code-editor/use-editor.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+export function useSyncEditorValue(editor: null | Ace.Editor, value: string) {
+  useEffect(() => {
+    if (!editor) {
+      return;
+    }
+    if (value === editor.getValue()) {
+      return;
+    }
+    // TODO maintain cursor position?
```
johannes-weber, line 79: Not related to the scope of this PR: I see this TODO is there for a while - do you have any insights about what needs to be done and what's the impact of not having the cursor position maintained?

johannes-weber (commented):
Overall, the code looks quite good and simpler! I've two questions, not directly related to the scope of the PR - if you have insights on them, it be great to add that additional context.

src/code-editor/setup-editor.ts
pan-kot: I don't know actually, the code was copied as is

src/code-editor/use-editor.tsx
johannes-weber: Got it! I recommend to remove the TODO because:
- it's a question and not clear what exactly it means
- does not give any hints about what's needed to be done
- it creates more confusion.

johannes-weber (dismissed):

src/code-editor/__tests__/code-editor.test.tsx
```diff
@@ -211,6 +216,14 @@ describe('Code editor component', () => {
   });
 
   it('provides ariaLabel to the underlying textarea', () => {
+    const textarea = editorMock.renderer.textarea;
+    editorMock.renderer.textarea = null as any;
+
+    renderCodeEditor({ ariaLabel: 'test aria label' });
+    // expect no failure
```
johannes-weber, line 223: If the failure you mention in the comment resulting in an exception? If so, you could make it explicit by asserting against it:
```suggestion
    expect(renderCodeEditor({ ariaLabel: 'test aria label' })).not.toThrowError();
```

johannes-weber (approved):

PR #1416, opened 2023-08-04
chore: Drawer badge, mobile and overflow

katiegeorge:
### Description

Adds the following:
- Red notification badge property
- Overflow behavior, in which the overflowed items fall into an ellipsis button dropdown
- Active overflow item swaps places with the last main item, until it is no longer active
- Mobile overflow - first 2 items exist inline, with all the rest falling into an ellipsis dropdown button

Classic:
image

VR:
image

Mobile: 
image

image

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/__tests__/desktop.test.tsx
+++ b/src/app-layout/__tests__/desktop.test.tsx
@@ -12,7 +12,6 @@ import {
   singleDrawer,
   singleDrawerOpen,
   manyDrawers,
-  defineClientHeight,
 } from './utils';
 import AppLayout, { AppLayoutProps } from '../../../lib/components/app-layout';
 import styles from '../../../lib/components/app-layout/styles.css.js';
@@ -28,8 +27,6 @@ jest.mock('@cloudscape-design/component-toolkit', () => ({
   ...jest.requireActual('@cloudscape-design/component-toolkit'),
   useContainerQuery: () => [1300, () => {}],
 }));
-
-defineClientHeight(100);
 
 describeEachThemeAppLayout(false, () => {
   test('renders breadcrumbs and notifications inside of the main landmark', () => {
@@ -241,7 +238,7 @@ describeEachThemeAppLayout(false, () => {
   test('should render overflow item when expected', () => {
     const { wrapper } = renderComponent(<AppLayout contentType="form" {...manyDrawers} />);
 
-    expect(wrapper.findDrawersTriggers()!).toHaveLength(1);
+    expect(wrapper.findDrawersTriggers()!.length).toBeLessThan(100);
   });
 });
 
@@ -286,7 +283,8 @@ describe('Classic only features', () => {
 
   test('should render badge when defined', () => {
     const { wrapper } = renderComponent(<AppLayout contentType="form" {...manyDrawers} />);
-    expect(wrapper.findDrawersTriggers()[0]!.getElement().children[0]).toHaveClass(iconStyles.badge);
+
+    expect(wrapper.findByClassName(iconStyles.badge)!.getElement()).toBeInTheDocument();
   });
 });
 
@@ -307,6 +305,6 @@ describe('VR only features', () => {
   test('should render badge when defined', () => {
     const { wrapper } = renderComponent(<AppLayout contentType="form" {...manyDrawers} />);
 
-    expect(wrapper.findDrawersTriggers()[0]!.getElement()).toHaveClass(visualRefreshStyles.badge);
+    expect(wrapper.findByClassName(visualRefreshStyles.badge)!.getElement()).toBeInTheDocument();
   });
 });
--- a/src/app-layout/__tests__/desktop.test.tsx
+++ b/src/app-layout/__tests__/desktop.test.tsx
@@ -17,6 +17,7 @@ import AppLayout, { AppLayoutProps } from '../../../lib/components/app-layout';
 import styles from '../../../lib/components/app-layout/styles.css.js';
 import notificationStyles from '../../../lib/components/app-layout/notifications/styles.css.js';
 import visualRefreshStyles from '../../../lib/components/app-layout/visual-refresh/styles.css.js';
+import iconStyles from '../../../lib/components/icon/styles.css.js';
 import customCssProps from '../../../lib/components/internal/generated/custom-css-properties';
 import { KeyCode } from '../../internal/keycode';
 import { useVisualRefresh } from '../../../lib/components/internal/hooks/use-visual-mode';
@@ -289,6 +290,11 @@ describe('Classic only features', () => {
     act(() => wrapper.findDrawersTriggers()![0].click());
     expect(wrapper.findActiveDrawer()!.getElement().style.width).toBe('500px');
   });
+
+  test('should render badge when defined', () => {
+    const { wrapper } = renderComponent(<AppLayout contentType="form" {...manyDrawers} />);
+    expect(wrapper.findDrawersTriggers()[0]!.getElement().children[0]).toHaveClass(iconStyles.badge);
+  });
 });
 
 describe('VR only features', () => {
@@ -303,5 +309,11 @@ describe('VR only features', () => {
     const { wrapper } = renderComponent(<AppLayout contentType="form" {...resizableDrawer} />);
     act(() => wrapper.findDrawersTriggers()![0].click());
     expect(wrapper.findActiveDrawer()!.getElement()).toHaveClass(styles['with-motion']);
+  });
+
+  test('should render badge when defined', () => {
+    const { wrapper } = renderComponent(<AppLayout contentType="form" {...manyDrawers} />);
+
+    expect(wrapper.findDrawersTriggers()[0]!.getElement()).toHaveClass(visualRefreshStyles.badge);
   });
 });
--- a/src/app-layout/__tests__/desktop.test.tsx
+++ b/src/app-layout/__tests__/desktop.test.tsx
@@ -11,6 +11,7 @@ import {
   resizableDrawer,
   singleDrawer,
   singleDrawerOpen,
+  manyDrawers,
 } from './utils';
 import AppLayout, { AppLayoutProps } from '../../../lib/components/app-layout';
 import styles from '../../../lib/components/app-layout/styles.css.js';
@@ -25,6 +26,16 @@ jest.mock('@cloudscape-design/component-toolkit', () => ({
   ...jest.requireActual('@cloudscape-design/component-toolkit'),
   useContainerQuery: () => [1300, () => {}],
 }));
+
+const originalClientHeight = Object.getOwnPropertyDescriptor(HTMLElement.prototype, 'clientHeight') || {};
+
+beforeAll(() => {
+  Object.defineProperty(HTMLElement.prototype, 'clientHeight', { configurable: true, value: 100 });
+});
+
+afterAll(() => {
+  Object.defineProperty(HTMLElement.prototype, 'clientHeight', originalClientHeight);
+});
 
 describeEachThemeAppLayout(false, () => {
   test('renders breadcrumbs and notifications inside of the main landmark', () => {
@@ -231,6 +242,12 @@ describeEachThemeAppLayout(false, () => {
 
     act(() => wrapper.findDrawersTriggers()![0].click());
     expect(wrapper.findDrawersSlider()!.getElement()).toHaveAttribute('aria-valuenow', '0');
+  });
+
+  test('should render overflow item when expected', () => {
+    const { wrapper } = renderComponent(<AppLayout contentType="form" {...manyDrawers} />);
+
+    expect(wrapper.findDrawersTriggers()!).toHaveLength(1);
   });
 });
 
--- a/src/app-layout/__tests__/desktop.test.tsx
+++ b/src/app-layout/__tests__/desktop.test.tsx
@@ -12,6 +12,7 @@ import {
   singleDrawer,
   singleDrawerOpen,
   manyDrawers,
+  defineClientHeight,
 } from './utils';
 import AppLayout, { AppLayoutProps } from '../../../lib/components/app-layout';
 import styles from '../../../lib/components/app-layout/styles.css.js';
@@ -28,15 +29,7 @@ jest.mock('@cloudscape-design/component-toolkit', () => ({
   useContainerQuery: () => [1300, () => {}],
 }));
 
-const originalClientHeight = Object.getOwnPropertyDescriptor(HTMLElement.prototype, 'clientHeight') || {};
-
-beforeAll(() => {
-  Object.defineProperty(HTMLElement.prototype, 'clientHeight', { configurable: true, value: 100 });
-});
-
-afterAll(() => {
-  Object.defineProperty(HTMLElement.prototype, 'clientHeight', originalClientHeight);
-});
+defineClientHeight(100);
 
 describeEachThemeAppLayout(false, () => {
   test('renders breadcrumbs and notifications inside of the main landmark', () => {
--- a/src/app-layout/__tests__/mobile.test.tsx
+++ b/src/app-layout/__tests__/mobile.test.tsx
@@ -9,6 +9,7 @@ import {
   renderComponent,
   singleDrawer,
   singleDrawerOpen,
+  manyDrawers,
   splitPanelI18nStrings,
 } from './utils';
 import AppLayout, { AppLayoutProps } from '../../../lib/components/app-layout';
@@ -16,7 +17,9 @@ import SplitPanel from '../../../lib/components/split-panel';
 import { AppLayoutWrapper } from '../../../lib/components/test-utils/dom';
 import styles from '../../../lib/components/app-layout/styles.css.js';
 import toolbarStyles from '../../../lib/components/app-layout/mobile-toolbar/styles.css.js';
+import iconStyles from '../../../lib/components/icon/styles.css.js';
 import testUtilsStyles from '../../../lib/components/app-layout/test-classes/styles.css.js';
+
 import visualRefreshRefactoredStyles from '../../../lib/components/app-layout/visual-refresh/styles.css.js';
 import { findUpUntil } from '../../../lib/components/internal/utils/dom';
 
@@ -353,5 +356,10 @@ describeEachThemeAppLayout(true, theme => {
     const { wrapper } = renderComponent(<AppLayout contentType="form" {...singleDrawer} />);
     expect(wrapper.findDrawersTriggers()![0].getElement()).toHaveAttribute('aria-label', 'Security trigger button');
     expect(wrapper.findDrawersMobileTriggersContainer()!.getElement()).toHaveAttribute('aria-label', 'Drawers');
+  });
+
+  test('should render badge when defined', () => {
+    const { wrapper } = renderComponent(<AppLayout contentType="form" {...manyDrawers} />);
+    expect(wrapper.findDrawersTriggers()[0]!.getElement().children[0]).toHaveClass(iconStyles.badge);
   });
 });
--- a/src/app-layout/drawer/index.tsx
+++ b/src/app-layout/drawer/index.tsx
@@ -3,11 +3,14 @@
 import clsx from 'clsx';
 import React, { useRef } from 'react';
 import { ToggleButton, CloseButton, togglesConfig } from '../toggles';
-import InternalButtonDropdown from '../../button-dropdown/internal';
 
 import testutilStyles from '../test-classes/styles.css.js';
 import styles from './styles.css.js';
 import { DesktopDrawerProps, DrawerTriggersBarProps, DrawerItem } from './interfaces';
+import OverflowMenu from './overflow-menu';
+import { useContainerQuery } from '@cloudscape-design/component-toolkit';
+import { useDensityMode } from '@cloudscape-design/component-toolkit/internal';
+import { splitItems } from './drawers-helpers';
 
 // We are using two landmarks per drawer, i.e. two NAVs and two ASIDEs, because of several
 // known bugs in NVDA that cause focus changes within a container to sometimes not be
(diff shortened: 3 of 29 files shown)


src/app-layout/visual-refresh/trigger-button.scss
```diff
@@ -51,4 +52,15 @@ handleSplitPanelMaxWidth function in the context.
...
+
+  &.badge::after {
+    content: '';
+    position: absolute;
+    width: 12px;
+    height: 12px;
+    border-radius: 8px;
+    top: 0;
+    right: 0;
+    background-color: awsui.$color-text-status-error;
```
just-boris, line 64: Don't you find this risky to use semantically different token here? Should we introduce a separate token here to avoid accidental overrides or theming issues?

src/button-dropdown/interfaces.ts
```diff
@@ -136,6 +136,7 @@ export namespace ButtonDropdownProps {
     iconName?: IconProps.Name;
     iconUrl?: string;
     iconSvg?: React.ReactNode;
+    badge?: boolean;
```
just-boris, line 139: With this addition, it becomes a public API of the button-dropdown component too.

Fine with me, can be a nice feature, but it needs team consensus. Separate API proposal is not needed, but a brief discussion would be useful

src/button-dropdown/styles.scss
```diff
@@ -88,4 +88,9 @@ $dropdown-trigger-icon-offset: 2px;
 }
 .dropdown-trigger {
   display: contents;
+
+  & > .trigger-active,
+  & > .trigger-active:hover {
+    color: awsui.$color-text-layout-toggle-active;
+  }
```
just-boris, line 95: This also looks like some for of CSS overrides. Need to find an alternative solution

src/button/icon-helper.tsx
```diff
@@ -18,6 +18,8 @@ export interface ButtonIconProps {
   iconSize?: IconProps.Size;
   variant?: string;
   iconClass?: string;
+  badge?: boolean;
+  badgeColor?: 'blue' | 'red';
```
just-boris, line 22: Interesting things happen here. 

The blue color used in the top navigation for the same reason, highlight active notifications. Do we really need separate colors in different places? Also, even if different colors, how about using visual contexts to modify a single design token for both?

src/app-layout/drawer/index.tsx
```diff
@@ -131,47 +135,171 @@ export const Drawer = React.forwardRef(
...
+                  className={clsx(styles['drawer-trigger'], overflowItemIsActive() && styles['drawer-trigger-active'])}
+                >
+                  <InternalButtonDropdown
+                    expandToViewport={true}
+                    className={clsx(styles['trigger-overflow'])}
+                    items={getDrawerItems()
+                      .slice(getIndexOfOverflowItem(), getDrawerItems().length)
+                      .map(item => ({
+                        id: item.id,
+                        text: item.ariaLabels?.content || 'Content',
```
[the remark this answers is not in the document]
katiegeorge: This is going to be the new structure of drawer aria labels (from the API doc): 
```
    export interface DrawerAriaLabels {
        drawerName: string;
        closeButton?: string;
        triggerButton?: string;
        resizeHandle?: string;
    }
```

`drawerName` will be required. I plan on making all of the API changes in a future PR so as to not break anything in the beta

src/app-layout/mobile-toolbar/index.tsx
```diff
@@ -64,6 +67,35 @@ interface MobileToolbarProps {
...
+    ref: React.Ref<ButtonProps.Ref>,
+    isDisabled: boolean,
+    isExpanded: boolean,
+    ariaLabel?: string
+  ) => {
+    return (
+      <InternalButton
+        disabled={isDisabled}
+        onClick={event => {
+          event.preventDefault();
```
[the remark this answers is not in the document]
katiegeorge: I'm not sure, I don't remember adding this. I can remove if it doesn't do anything.

src/button/icon-helper.tsx
katiegeorge: We had a design team discussion about the blue vs. red color and decided to leave both for now. We may want to keep both variants in the future, but how would I use visual contexts in this case?

src/app-layout/drawer/index.tsx
```diff
@@ -131,47 +135,171 @@ export const Drawer = React.forwardRef(
...
   topOffset,
   bottomOffset,
   drawers,
   contentClassName,
   toggleClassName,
-}: DrawerTriggersBarProps) {
+}: DrawerTriggersBarProps) => {
+  const triggersContainerRef = useRef<HTMLDivElement>(null);
+
+  const containerHeight = triggersContainerRef.current?.clientHeight;
```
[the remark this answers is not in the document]
katiegeorge: I've tested this thoroughly with screen resizes and it does work. Interestingly, in the VR version, I had to add `useContainerHeight` ([link](https://github.com/cloudscape-design/components/pull/1416/files#diff-fa3c75e391e28b0bb11c7d5d5dbd152b856edd723ecb6c23996673830b0c5282R163)) for it to work on resize, but not here. I'll take a look at useContainerQuery.

src/app-layout/__tests__/desktop.test.tsx
```diff
@@ -279,6 +290,11 @@ describe('Classic only features', () => {
     act(() => wrapper.findDrawersTriggers()![0].click());
     expect(wrapper.findActiveDrawer()!.getElement().style.width).toBe('500px');
   });
+
+  test('should render badge when defined', () => {
+    const { wrapper } = renderComponent(<AppLayout contentType="form" {...manyDrawers} />);
+    expect(wrapper.findDrawersTriggers()[0]!.getElement().children[0]).toHaveClass(iconStyles.badge);
```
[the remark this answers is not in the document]
katiegeorge: This didn't quite work (Property 'findByClassName' does not exist on type 'ElementWrapper<HTMLButtonElement>[]') but I tried this instead, which is less brittle (found in breadcrumb-group.test.tsx):
`expect(wrapper.findByClassName(iconStyles.badge)!.getElement()).toBeInTheDocument();`

src/app-layout/drawer/index.tsx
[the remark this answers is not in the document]
katiegeorge: `useContainerQuery` worked like a charm. Killed `defineClientHeight` as well.

src/app-layout/drawer/index.tsx
```diff
@@ -131,47 +135,171 @@ export const Drawer = React.forwardRef(
...
+
+  const getIndexOfOverflowItem = () => {
+    if (containerHeight) {
+      const itemHeight = 38;
```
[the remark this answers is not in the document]
katiegeorge: Good call - it isn't. Is there a way to get `isCompactMode` here?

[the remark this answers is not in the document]
katiegeorge: I've tried using a container query to get the height here as well, but I'm getting lots of thrashing on resize.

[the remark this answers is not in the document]
katiegeorge: I added a simple `useCompactMode` hook to the internal folder.

src/app-layout/mobile-toolbar/index.tsx
```diff
@@ -122,19 +160,41 @@ export function MobileToolbar({
...
                 ariaLabel={item.ariaLabels?.triggerButton}
                 onClick={() => drawers.onChange({ activeDrawerId: item.id })}
                 ariaExpanded={drawers.activeDrawerId !== undefined}
                 testId={`awsui-app-layout-trigger-${item.id}`}
               />
             </span>
           ))}
+          {drawers?.items?.length && drawers?.items?.length > 2 && (
+            <span className={clsx(styles['mobile-toggle'], styles['mobile-toggle-type-drawer'])}>
+              <InternalButtonDropdown
```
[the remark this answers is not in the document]
katiegeorge: Pulled it into its own component.

src/app-layout/visual-refresh/drawers.tsx
```diff
@@ -127,6 +130,46 @@ function ActiveDrawer() {
...
+        badge={hasOverflowBadge}
+        badgeColor="red"
+        iconName="ellipsis"
+      />
+    );
+  };
+  return DropdownTrigger;
+};
+
+function useContainerHeight(ref: React.RefObject<HTMLDivElement>) {
```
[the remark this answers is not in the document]
katiegeorge: 🔪

src/button/icon-helper.tsx
just-boris: We can define a design token, `colorBackgroundNotificationBadge`, for example (also to address this comment https://github.com/cloudscape-design/components/pull/1416#discussion_r1287022139), default to red and then redefine it to blue in this context only: https://github.com/cloudscape-design/components/blob/main/style-dictionary/classic/contexts/top-navigation.ts

src/button-dropdown/interfaces.ts
just-boris: Just to clarify, after merging this PR, it becomes a defacto public API (not a beta like the drawers, but fully public).

This is why the discussion needs to happen before the merge

src/app-layout/__tests__/mobile.test.tsx
```diff
@@ -354,4 +357,9 @@ describeEachThemeAppLayout(true, theme => {
     expect(wrapper.findDrawersTriggers()![0].getElement()).toHaveAttribute('aria-label', 'Security trigger button');
     expect(wrapper.findDrawersMobileTriggersContainer()!.getElement()).toHaveAttribute('aria-label', 'Drawers');
   });
+
+  test('should render badge when defined', () => {
+    const { wrapper } = renderComponent(<AppLayout contentType="form" {...manyDrawers} />);
+    expect(wrapper.findDrawersTriggers()[0]!.getElement().children[0]).toHaveClass(iconStyles.badge);
```
just-boris, line 363:
```suggestion
 
 
 
 expect(wrapper.findByClassName(visualRefreshStyles.badge)!.getElement()).toBeInTheDocument();
```

src/app-layout/visual-refresh/drawers.tsx
```diff
@@ -169,6 +236,13 @@ function DesktopTriggers() {
     return null;
   }
 
+  const overflowItemHasBadge = () => {
+    const { overflowItems } = splitItems(drawers?.items);
+    return (overflowItems && overflowItems.filter(item => item.badge).length > 0) || false;
+  };
+
+  const { visibleItems, overflowItems } = splitItems(getDrawerItems());
+
```
just-boris, line 245: A bit of personal opinion, but I would simplify this code

```js
const { visibleItems, overflowItems } = splitItems(getDrawerItems());
const overflowItemHasBadge = overflowItems.find(item => item.badge);
```

No need for a function and also we can call `splitItems` only once

src/app-layout/visual-refresh/drawers.tsx
```diff
@@ -201,23 +276,42 @@ function DesktopTriggers() {
...
+              }}
+              ref={item.id === previousActiveDrawerId.current ? drawersRefs.toggle : undefined}
+              selected={item.id === activeDrawerId}
+              badge={item.badge}
+              testId={`awsui-app-layout-trigger-${item.id}`}
+            />
+          );
+        })}
+
+        {getDrawerItems() && getDrawerItems().length > getIndexOfOverflowItem() && (
```
just-boris, line 299:
```suggestion
{overflowItems.length > 0 && (
```

src/app-layout/visual-refresh/drawers.tsx
```diff
@@ -265,6 +359,16 @@ export function MobileTriggers() {
...
 
+  const overflowItemHasBadge = () => {
+    const overflowItems = drawers?.items.slice(1, drawers.items.length);
+    return overflowItems && overflowItems.filter(item => item.badge).length > 0;
+  };
+
+  const mobileItems = !toolsHide && tools ? drawers.items.slice(0, 1) : drawers.items.slice(0, 2);
+  const hasOverflowMenu = !toolsHide && tools ? drawers?.items?.length > 2 : drawers?.items?.length > 3;
+  const overflowItems =
+    !toolsHide && tools ? drawers.items.slice(1, drawers.items.length) : drawers.items.slice(2, drawers.items.length);
```
just-boris, line 370: I think this code could also reuse `splitItems` helper, if it would work like this

```
function splitItems(allItems: Array<DrawerItem>, splitIndex: number) {...}

// use in desktop mode
splitItems(drawers.items, getIndexOfOverflowItem());

// use in mobile mode
splitItems(drawers.items, 2);
```

src/app-layout/drawer/overflow-menu.tsx
```diff
@@ -0,0 +1,71 @@
...
+  hasOverflowBadge?: boolean;
+  hasActiveStyles?: boolean;
+}
+
+export default function OverflowMenu(props: OverflowMenuProps) {
+  const { drawersRefs, className, overflowItems, onItemClick, hasOverflowBadge, hasActiveStyles } = props;
+  return (
+    <InternalButtonDropdown
+      ref={drawersRefs ? drawersRefs.toggle : null}
+      className={className}
```
just-boris, line 56: There are still some CSS overrides which we need to avoid

src/app-layout/drawer/index.tsx
```diff
@@ -131,47 +134,136 @@ export const Drawer = React.forwardRef(
...
   isMobile,
   topOffset,
   bottomOffset,
   drawers,
   contentClassName,
   toggleClassName,
-}: DrawerTriggersBarProps) {
+}: DrawerTriggersBarProps) => {
+  const [containerHeight, triggersContainerRef] = useContainerQuery(rect => rect.contentBoxHeight);
+  const isCompactMode = useCompactMode();
```
just-boris, line 146: There is already an existing hook: https://github.com/cloudscape-design/component-toolkit/blob/main/src/internal/visual-mode/index.ts#L31

It just lives in another package to allow reuse for border components:

```js
import { useDensityMode } from '@cloudscape-design/component-toolkit/internal';
```

just-boris: I also have an alternative approach, which does not need complex maths.

We can render a 2nd hidden copy of the drawers bar which does not have overflow checks. Then we take the height of the hidden bar and compare with the available height and make adjustments.

The alternative approach has an upside that this can be a single code for all modes and visual refresh

src/app-layout/drawer/index.tsx
```diff
@@ -131,47 +134,136 @@ export const Drawer = React.forwardRef(
...
         {!isMobile && (
           <aside aria-label={drawers?.ariaLabel} className={clsx(styles['drawer-triggers-wrapper'], contentClassName)}>
-            {drawers?.items?.map((item: DrawerItem, index: number) => (
-              <span
-                key={index}
-                className={clsx(
-                  styles['drawer-trigger'],
-                  drawers.activeDrawerId === item.id && styles['drawer-trigger-active']
-                )}
-                onClick={() =>
```
just-boris, line 158: Why do we need to move the handler? Now there is an issue that the trigger outside blue area is not clickable
image

katiegeorge: The handler was being called twice, once on the span and once on the Toggle button. I needed to remove one of them. I can remove the one on the ToggleButton instead

src/app-layout/drawer/index.tsx
katiegeorge: This isn't complex - its one value for Comfortable mode, and one for Compact mode (In VR compact mode it doesn't actually change, so its 3 total values). I feel like rendering a hidden drawer is a bit overkill here

src/button-dropdown/interfaces.ts
katiegeorge: It wasn't my intention to make it public. I feel like more conversations need to happen before we release this feature in a button dropdown, how can I make this a private feature?

src/app-layout/visual-refresh/trigger-button.scss
katiegeorge: Good point, I will add an additional token.

src/app-layout/mobile-toolbar/index.tsx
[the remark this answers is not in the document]
katiegeorge: Looks like its part of how the `customTriggerBuilder` works in our internal button dropdown, so we do need to keep it.

src/button-dropdown/interfaces.ts
just-boris: There is no easy way to make it public. In the past we used to have a private `href` feature (needed internally for breadcrumb group, but we did not make it public). Eventually, it leaked anyway and we had to call it de-facto public API. See AWSUI-12000 more details

I think, making it public is easier than setting up a guards against leaking this. I will try to help you with the team discussions

src/button-dropdown/styles.scss
katiegeorge: Based on the styles of the button dropdown, I couldn't find another way to make the trigger colors match the rest of the buttons (be white). We need to access the button dropdown <button> itself, and this is the only way to do that. What alternative do you see?

src/button/icon-helper.tsx
katiegeorge: So I ended up realizing there are already design tokens used in the badge component itself - `awsui.$color-background-notification-red;` and `awsui.$color-background-notification-blue;` which I updated in the PR. What do you think of using those? They convey the same message without having to add anything additional.

src/button-dropdown/styles.scss
katiegeorge: The other solution I see is adding an internal "type" to the ButtonDropdown itself, which adds the proper styles.

src/app-layout/visual-refresh/drawers.tsx
```diff
@@ -161,6 +206,63 @@ function DesktopTriggers() {
...
+  const splitPanelHeight =
+    isSplitPanelOpen && splitPanelPosition === 'bottom' ? splitPanelReportedSize : splitPanelReportedHeaderHeight;
+  const containerHeight = useContainerHeight(triggersContainerRef) || triggersContainerRef.current?.clientHeight;
+
+  const getIndexOfOverflowItem = () => {
+    if (containerHeight) {
+      const ITEM_HEIGHT = 48;
+      const overflowSpot =
+        activeDrawerId && isSplitPanelOpen
+          ? (containerHeight - splitPanelReportedHeaderHeight) / 1.5
```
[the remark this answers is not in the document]
katiegeorge: image
I'm not sure we can - based on the grid lines here, there aren't any that correspond to the top of the split panel. Unless I'm missing something?

src/button/icon-helper.tsx
just-boris: I am not sure about exposing blue/red choice in the API. For me this choice is contextual, based on background (thus perfectly fits visual contexts use case).

Also, some may argue that badge should use "color-text" tokens and not "color-background", because there is no text inside the badge.

My personal opinion is still the same – "icon-badge" is a separate feature than the "badge" component and should use its own token. But we can discuss with our squad if there are different opinions

src/app-layout/visual-refresh/drawers.tsx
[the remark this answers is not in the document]
just-boris: Checked the implementation one more time and you are correct. Split panel sticks to the bottom using `align-self`, no easy wins here. I am okay with the current code with this discovery.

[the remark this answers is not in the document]
just-boris: However when inspecting this setup, I found a corner case

Here we want the overflow to consider split panel height:

Screenshot 2023-08-11 at 16 44 45

But in this one we probably do not (because drawers bar has the full height):

Screenshot 2023-08-11 at 16 44 58

What do you think? Is this important enough to address?

---

🤔  When trying to address this, the height calculation condition becomes even more convoluted. The idea from this comment may become more valuable https://github.com/cloudscape-design/components/pull/1416#discussion_r1288281996, because it separates tasks "make buttons overflow if they do not fit container" and "adjust container height to not overlap split panel". Two simpler tasks are usually more preferred than a single complex one.

src/button-dropdown/styles.scss
just-boris: For this thread I need to check out this package locally and inspect more closely.

I will try to find some time for it on Monday

src/app-layout/visual-refresh/drawers.tsx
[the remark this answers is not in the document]
katiegeorge: This calculation already takes this into account (see the `activeDrawerId &&` logic in the code).

src/button/icon-helper.tsx
katiegeorge: So I brought this back up with the design team today and we decided to just make the notification badge red everywhere. So this becomes a non-issue anyway!

src/app-layout/mobile-toolbar/index.tsx
```diff
@@ -88,6 +92,10 @@ export function MobileToolbar({
       document.body.classList.remove(styles['block-body-scroll']);
     }
   }, [anyPanelOpen]);
+
+  const splitNum = drawers?.items.length === 3 ? 3 : 2;
```
just-boris, line 96: Good catch. I think this logic should go inside the `splitItems` function, because it is the same for all cases

Also, it would be very convenient to test via `src/app-layout/drawer/__tests__/drawers-helpers.test.ts` setup

src/app-layout/visual-refresh/drawers.tsx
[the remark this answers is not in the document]
just-boris: It seems like there some fixes were made. I cannot reproduce the previous behavior on the latest version.

resolving the thread.

src/app-layout/mobile-toolbar/index.tsx
katiegeorge: It's only the case for mobile, not all. But I can add an `isMobile` argument to the function.

just-boris: Why is this mobile only? I think it makes sense for the desktop too. Every time there is a dropdown with just one item, it could be rendered without the overflow

just-boris:
Found that one thread got accidentally resolved:

https://github.com/cloudscape-design/components/pull/1416#discussion_r1297322638

In my opinion, dropdown menu with just one item is very annoying UX issue. I can approve the current revision, but I think we should fix it as a follow up

just-boris (dismissed):

src/app-layout/visual-refresh/trigger-button.scss
```diff
@@ -51,4 +52,24 @@ handleSplitPanelMaxWidth function in the context.
       background: awsui.$color-background-layout-toggle-selected-active;
     }
   }
+
+  &.badge {
+    clip-path: path(
+      'M38.0215 11.3161C37.1068 11.7545 36.0821 12 35 12C31.134 12 28 8.86599 28 5C28 3.91794 28.2455 2.89322 28.6839 1.97851C26.0575 0.710652 23.1118 0 20 0C8.9543 0 0 8.9543 0 20C0 31.0457 8.9543 40 20 40C31.0457 40 40 31.0457 40 20C40 16.8882 39.2893 13.9425 38.0215 11.3161Z'
```
just-boris, line 58: Do we still keep this clip-path? Does Safari like it?

katiegeorge: I checked this one on Safari - its fine because its on the button element and not directly on the svg. Still a mystery why Safari didn't like the svg clip-path specifically, want to do some more digging.

src/app-layout/mobile-toolbar/index.tsx
katiegeorge: This is already the case! I added it as a feature to the [split items helper](https://github.com/cloudscape-design/components/pull/1416/files#diff-a0890a9725d3f94a60ee2a7aeed898f9771db07811a439e6313e92aa1bbdbe93R14).

just-boris: I see a code simplification opportunity. I will follow up when all currently planned features added and the dust settles

PR #1423, opened 2023-08-08
feat: Allow dropdowns of autosuggest, select, and multiselect to expand beyond their trigger width

timogasda:
### Description
The main change in this PR is allowing the dropdowns of select, multiselect, and autosuggest to expand their width according to their content (up to a pre-defined maximum or 465px). Additionally, there are some other changes that support this new feature:

* Fixed a bug where dropdowns with `expandToViewport={true}` could expand beyond the trigger width (AWSUI-16369)
* Changed `getDropdownPosition` from using (9) positional arguments to using an object as argument

Related links, issue #, if available: AWSUI-19898, AWSUI-16369

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/autosuggest/virtual-resize.page.tsx
+++ b/pages/autosuggest/virtual-resize.page.tsx
@@ -39,7 +39,6 @@ export default function AutosuggestPage() {
           ariaLabel={'simple autosuggest'}
           selectedAriaLabel="Selected"
           virtualScroll={true}
-          expandToViewport={false}
         />
       </div>
     </div>
--- a/pages/autosuggest/virtual-resize.page.tsx
+++ b/pages/autosuggest/virtual-resize.page.tsx
@@ -10,7 +10,8 @@ declare const window: ExtendedWindow;
 
 const options = [
   {
-    value: 'A very long option label that wraps upon resizing',
+    value:
+      'A very very very very very very very very very very very very very long option label that wraps upon resizing',
     tags: ['tag1', 'tag2'],
     filteringTags: ['bla', 'opt'],
     description: 'description1',
@@ -38,6 +39,7 @@ export default function AutosuggestPage() {
           ariaLabel={'simple autosuggest'}
           selectedAriaLabel="Selected"
           virtualScroll={true}
+          expandToViewport={false}
         />
       </div>
     </div>
--- a/src/internal/components/dropdown/dropdown-fit-handler.ts
+++ b/src/internal/components/dropdown/dropdown-fit-handler.ts
@@ -125,15 +125,19 @@ export const getDropdownPosition = (
   preferCenter = false,
   stretchWidth = false,
   stretchHeight = false,
-  isMobile?: boolean
+  isMobile?: boolean,
+  stretchBeyondTriggerWidth?: boolean
 ): DropdownPosition => {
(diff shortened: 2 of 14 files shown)


pages/autosuggest/virtual-resize.page.tsx
```diff
@@ -10,7 +10,8 @@ declare const window: ExtendedWindow;
 
 const options = [
   {
-    value: 'A very long option label that wraps upon resizing',
+    value:
+      'A very very very very very very very very very very very very very long option label that wraps upon resizing',
```
timogasda, line 14: Extended this string so that it still wraps when needed by the integration test.

src/internal/components/dropdown/styles.scss
```diff
@@ -70,6 +70,11 @@
   &.occupy-entire-width {
     min-width: 100%;
   }
+  &.stretch-beyond-trigger-width {
+    width: max-content;
+    // This maps to the XXS container breakpoint as defined in src/internal/breakpoints.ts
+    max-width: 465px;
```
just-boris, line 76: Can we avoid this duplication by setting this value with inline styles from JS?

src/internal/components/dropdown/styles.scss
```diff
@@ -70,6 +70,11 @@
   &.occupy-entire-width {
```
just-boris, line 70: Are there any conflicts with this style

src/internal/components/dropdown/dropdown-fit-handler.ts
```diff
@@ -116,22 +121,46 @@ export const getInteriorAvailableSpace = (
...
+}: {
+  triggerElement: HTMLElement;
+  dropdownElement: HTMLElement;
+  overflowParents: ReadonlyArray<Dimensions>;
+  minWidth?: number;
+  preferCenter?: boolean;
+  stretchWidth?: boolean;
+  stretchHeight?: boolean;
+  isMobile?: boolean;
+  stretchBeyondTriggerWidth?: boolean;
```
just-boris, line 143: Reminder about this tech debt item: AWSUI-13040

This code is overloaded with flags which may contradict each other. There is an opportunity to split this code into separate "position strategies" for different use-cases.

Not a blocker, but I wanted to raise your attention to this

just-boris (commented):
What about multiselect and button dropdown? Do we need to include them here as well?

if not, I can approve

src/internal/components/dropdown/styles.scss
timogasda: No, in fact they complement each other. This class is still relevant and makes sure that the dropdown fills the trigger width (when `expandToViewport` is false).

src/internal/components/dropdown/styles.scss
timogasda: Good idea, let's do that.

src/internal/components/dropdown/dropdown-fit-handler.ts
timogasda: Yeah, totally valid point. Unfortunately, I didn't really have the bandwidth to take that on.

just-boris (approved):

PR #1317, opened 2023-07-12
chore: Improve info density of headers and containers in VR

jkuelz:
### Description

This is part of an ongoing initiative to increase the information density of VR in both comfortable mode and compact mode. This change includes the following updates to table, cards, header, and container (plus container adjacent components - e.g. modal, expandable section):

- Reduces container header vertical padding in VR (both modes - comfortable & compact)
- Refactor header component in VR to help make the future typography update more seamless. 
- Switch to using gaps and remove outer margin around the header, as well as on its inner elements, to provide more equal padding when wrapping with simpler implementation (both modes)
  - Use flex column gap instead of margin between header and actions so header takes up 100% horizontal space (both modes + classic) 
- Reduce the app layout gap above and below page headers (both modes)
- Removes extra padding above tables with and without headers in VR (both modes)
- Reduce card vertical padding in VR (compact mode only)

https://github.com/cloudscape-design/components/assets/15003460/258469da-0d3b-4b38-898d-e7b5564d9690

https://github.com/cloudscape-design/components/assets/15003460/e8761323-da59-4c62-b24d-1e5f430019ca

https://github.com/cloudscape-design/components/assets/15003460/cc963311-faf1-4b20-9dc2-f235810ed2a3

Relevant links: CR-96406829, demos [#89](https://github.com/cloudscape-design/demos/pull/89), board components [#257](https://github.com/cloudscape-design/board-components/pull/257)

### How has this been tested?

Ran through dev pipeline, and reviewed with visual design squad. Can view screenshot diffs in approval instance 56dec3dd-6922-4f98-9596-53240e1f5f63.

<details>
   <summary>Review checklist</summary>

_The following items are to be evaluated by the author(s) and the reviewer(s)._

#### Correctness

- _Changes include appropriate documentation updates._
- _Changes are backward-compatible if not indicated, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#public-apis)._
- _Changes do not include unsupported browser features, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#browsers-support)._
- _Changes were manually tested for accessibility, see [accessibility guidelines](https://cloudscape.design/foundation/core-principles/accessibility/)._

#### Security

- _If the code handles URLs: all URLs are validated through [the `checkSafeUrl` function](https://github.com/cloudscape-design/components/blob/main/src/internal/utils/check-safe-url.ts)._

#### Testing

[...]

PR #1425, opened 2023-08-09
fix: Asymmetric header vertical padding in closed expandable section

jkuelz:
### Description

We reduced header bottom padding in VR to improve info density in #1317, this introduced an inconsistency in the expandable section when it is closed, because heading vertical padding is no longer symmetrical. This PR increases the bottom padding of expandable section headers when it is closed so it looks more balanced.

https://github.com/cloudscape-design/components/assets/15003460/e86b4ce3-916a-4b15-ae7f-5e798cdedf28

Ideally, we would not add more space to the interface, however, we cannot reduce the top padding instead because we don't want a shift in the text between open and closed states. We also cannot have header text start at different heights because the board components use fixed alignment with the handles and settings buttons.

Related links, issue #, if available: AWSUI-21867

### How has this been tested?

Ran through dev pipeline, screenshot approval-instance: ce67a9fc-e53b-4d70-965f-0eb94b6ef0ac

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/header/internal.tsx
+++ b/src/header/internal.tsx
@@ -55,14 +55,7 @@ export default function InternalHeader({
       ref={__internalRootRef}
     >
       <div className={clsx(styles.main, styles[`main-variant-${variantOverride}`], isRefresh && styles.refresh)}>
-        <div
-          className={clsx(
-            styles.title,
-            styles[`title-variant-${variantOverride}`],
-            isRefresh && styles.refresh,
-            actions && [styles[`has-actions`]]
-          )}
-        >
+        <div className={clsx(styles.title, styles[`title-variant-${variantOverride}`], isRefresh && styles.refresh)}>
           <HeadingTag className={clsx(styles.heading, styles[`heading-variant-${variantOverride}`])}>
             <span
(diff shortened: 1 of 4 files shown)


src/header/internal.tsx
```diff
@@ -55,14 +55,7 @@ export default function InternalHeader({
       ref={__internalRootRef}
     >
       <div className={clsx(styles.main, styles[`main-variant-${variantOverride}`], isRefresh && styles.refresh)}>
-        <div
-          className={clsx(
-            styles.title,
-            styles[`title-variant-${variantOverride}`],
-            isRefresh && styles.refresh,
-            description && [styles[`root-has-description`]]
```
rubencarvalho, line 63: If I understand this correctly, when `description` was present, `.root-has-description` was applied, which removed the `padding-bottom`. Why don't we need this anymore?

jkuelz: The class was applied twice, once on the title (this instance) and the other on the root. The styles only referenced the class applied on the root, so this one was never actually used.

PR #1428, opened 2023-08-09
fix: Disallow dropping multiple files onto non-multiple FileUpload

gethinwebster:
### Description

Disallow dropping multiple files onto non-multiple FileUpload.

This ensures that drop behavior matches file selector behavior, 
and better matches with browser native inputs.

Related links, issue #, if available: AWSUI-21885

### How has this been tested?

New unit tests & local testing

<details>

#### Correctness

#### Security

#### Testing

</details>

---


connorlanigan (changes requested):
This new UX feels misleading to me - the component reacts to the user dragging files over the page, and explicitly offers/invites to drop the files there. Then when the user does as they've been invited to, the files are silently ignored.

https://github.com/cloudscape-design/components/assets/9086585/0fc44c43-5261-415b-b03e-6981ad73af43

This differs from e.g. the behaviour of the native file input in Chrome (tested on [the MDN page](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file)): when a single file is dragged, the button of the input slightly brightens, indicating possible interaction. When dragging _multiple_ files, the button does not change at all, i.e. it does not offer the user to drop the files anymore.

connorlanigan (approved):

PR #1434, opened 2023-08-10
fix: Refactor and fix AppLayout and side navigation panel state and behaviour

rubencarvalho:
### Description

When clicking a navigation link in mobile AppLayout side navigation, the panel should close automatically. Furthermore, on mobile viewports, the side navigation should also be closed by default.

This used to be handled in our code, but has regressed:

This behavior is supposed to be [handled by the AppLayout](https://github.com/cloudscape-design/components/blob/main/src/app-layout/index.tsx#L249-L257
).
However, our side navigation [was blocking this code](https://github.com/cloudscape-design/components/blob/main/src/side-navigation/internal.tsx#L231). I've split the event listener to be applied to both the header text and trigger button for the expandable section, so we don't need the `stopPropagation`.

There was also a unit test for it, but very artificial and this is why the regression was not caught: https://github.com/cloudscape-design/components/blob/715264b967c0811cde77f29ec7e98312d780cd45/src/app-layout/__tests__/mobile.test.tsx#L85

Related links: AWSUI-21892

### How has this been tested?

Existing tests, updated some "artificial ones" and added new ones to cover the expected behaviour.

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/app-layout/index.tsx
```diff
@@ -197,6 +197,14 @@ const OldAppLayout = React.forwardRef(
       }
     };
 
+    useEffect(() => {
+      // To guarantee an optimal UX on mobile, we forcely close the navigation on this viewport
```
timogasda, line 201: (minor) "guarantee" and "optimal" are some pretty big words, and don't really explain the motivation. I would say something like "Close navigation drawer on mobile so that the main content is visible"

rubencarvalho: Good point. Not really trying to sell a feature here. I'll update

src/app-layout/index.tsx
```diff
@@ -197,6 +197,14 @@ const OldAppLayout = React.forwardRef(
       }
     };
 
+    useEffect(() => {
+      // Close navigation drawer on mobile so that the main content is visible
+      if (isMobile) {
+        onNavigationToggle(false);
+      }
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [isMobile]);
```
just-boris, line 206: As opposed to adding more `eslint-disable-next-line` exceptions, I suggest converting `const onNavigationToggle = useCallback(() => {}, [])` to `const onNavigationToggle = useStableCallback(() => {})`. This value will be safe to put into effects

src/side-navigation/internal.tsx
```diff
@@ -412,6 +412,7 @@ function ExpandableLinkGroup({ definition, fireFollow, fireChange, activeHref, v
       variant="navigation"
       expanded={userExpanded ?? expanded}
       onChange={onExpandedChange}
+      __disableExpandChangeOnHeaderTextClick={disableExpandChangeOnHeaderTextClick}
```
timogasda, line 415: Can we do this without an internal prop? 

For example, what if we baked this behavior into the `navigation` variant of expandable section?

src/app-layout/__tests__/utils.tsx
```diff
@@ -87,7 +87,7 @@ export function describeEachAppLayout(callback: () => void) {
           (useMobile as jest.Mock).mockReset();
           (useVisualRefresh as jest.Mock).mockReset();
         });
-        callback();
+        callback(size as 'desktop' | 'mobile');
```
timogasda, line 90: minor: Instead of casting the value of the variable (technically not safe), it would be better to cast the array that you iterate over (`const size of ['desktop', 'mobile'] as const`)

src/side-navigation/internal.tsx
rubencarvalho: I've changed from using an internal prop (for detecting the desired behaviour of the trigger to expand the section) into the `"navigation"` variant of the expandable section.

> `navigation` - Use this variant in the navigation panel with anchors and custom styled content. It doesn't have any default styles.

This is a more impactful change, but between a successful dry-run (ID: 6564992503) and our current guidelines (above), I'm confident this is the best approach.

src/expandable-section/expandable-section-header.tsx
```diff
@@ -95,14 +95,15 @@ const ExpandableNavigationHeader = ({
   icon,
 }: ExpandableNavigationHeaderProps) => {
   return (
```
rubencarvalho, line 97: `<ExpandableNavigationHeader />` is rendered conditionally only for the `navigation` variant, so the change below is only applied to this variant.

https://github.com/cloudscape-design/components/blob/4c020a82878a4590218caacad2468144e033795a/src/expandable-section/expandable-section-header.tsx#L255-L265

src/side-navigation/internal.tsx
timogasda: Let's do it!

timogasda (approved):

PR #1444, opened 2023-08-14
fix: Refactor drawers overflow

just-boris:
### Description

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/drawer/index.tsx
+++ b/src/app-layout/drawer/index.tsx
@@ -135,14 +135,38 @@ export const Drawer = React.forwardRef(
   }
 );
 
-export const DrawerTriggersBar = ({
-  isMobile,
-  topOffset,
-  bottomOffset,
-  drawers,
-  contentClassName,
-  toggleClassName,
-}: DrawerTriggersBarProps) => {
+interface DrawerTriggerProps {
+  testUtilsClassName?: string;
+  ariaLabel: string | undefined;
+  ariaExpanded: boolean;
+  badge: boolean | undefined;
+  itemId?: string;
+  isActive: boolean;
+  trigger: DrawerItem['trigger'];
+  onClick: () => void;
+}
+
+const DrawerTrigger = React.forwardRef(
+  (
+    { testUtilsClassName, ariaLabel, ariaExpanded, badge, itemId, isActive, trigger, onClick }: DrawerTriggerProps,
+    ref: React.Ref<{ focus: () => void }>
+  ) => (
+    <div className={clsx(styles['drawer-trigger'], isActive && styles['drawer-trigger-active'])} onClick={onClick}>
+      <ToggleButton
+        ref={ref}
+        className={testUtilsClassName}
+        iconName={trigger.iconName}
+        iconSvg={trigger.iconSvg}
+        ariaLabel={ariaLabel}
+        ariaExpanded={ariaExpanded}
+        badge={badge}
+        testId={itemId && `awsui-app-layout-trigger-${itemId}`}
+      />
+    </div>
+  )
+);
+
+export const DrawerTriggersBar = ({ isMobile, topOffset, bottomOffset, drawers }: DrawerTriggersBarProps) => {
   const containerRef = React.useRef<HTMLDivElement>(null);
   const [containerHeight, triggersContainerRef] = useContainerQuery(rect => rect.contentBoxHeight);
   const isCompactMode = useDensityMode(containerRef) === 'compact';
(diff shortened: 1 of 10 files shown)


src/app-layout/drawer/index.tsx
```diff
@@ -159,37 +183,7 @@ export const DrawerTriggersBar = ({
...
-      }
-
-      return drawers.items;
-    }
-    return [];
-  };
-
-  const { visibleItems, overflowItems } = splitItems(getDrawerItems(), getIndexOfOverflowItem());
-  const overflowItemHasBadge = !!overflowItems?.find(item => item.badge);
+  const { visibleItems, overflowItems } = splitItems(drawers?.items, getIndexOfOverflowItem(), drawers?.activeDrawerId);
```
just-boris, line 186: Consolidating splitting by index and active item handing in one function allowed to simplify this code by a lot

src/app-layout/drawer/index.tsx
```diff
@@ -204,51 +198,51 @@ export const DrawerTriggersBar = ({
...
+          <aside
+            aria-label={drawers?.ariaLabel}
+            className={clsx(styles['drawer-triggers-wrapper'], testutilStyles['drawers-desktop-triggers-container'])}
+          >
             <>
-              {visibleItems?.map((item: DrawerItem, index: number) => {
+              {visibleItems.map((item, index) => {
                 return (
-                  <span
+                  <DrawerTrigger
```
just-boris, line 208: Extracted utility component to reuse between normal triggers and the overflow trigger

src/app-layout/drawer/interfaces.ts
```diff
@@ -45,8 +45,6 @@ export interface ResizableDrawerProps extends DesktopDrawerProps {
 }
 
 export interface DrawerTriggersBarProps {
-  contentClassName: string;
-  toggleClassName: string;
```
just-boris, line 49: These never change, so not much of a point to make them a props. Fewer props make it easier to understand components

src/app-layout/visual-refresh/drawers.tsx
```diff
@@ -286,21 +255,22 @@ function DesktopTriggers() {
...
 
-        {overflowItems && overflowItems.length > 0 && (
+        {overflowItems.length > 0 && (
           <OverflowMenu
-            drawersRefs={drawersRefs}
-            className={clsx(
-              styles['drawers-trigger'],
-              styles.trigger,
-              styles['drawers-trigger-overflow'],
-              overflowItemHasBadge && styles.badge
```
katiegeorge, line 296: It looks like we've lost the badge on the overflow menu when one of its items has a badge
image

src/app-layout/drawer/styles.scss
```diff
@@ -67,15 +67,15 @@ $drawer-z-index-mobile: 1001;
   flex-direction: column;
   text-align: center;
   align-items: stretch;
+  & > :not(:first-child) {
```
just-boris, line 70: Moving this here allowed to avoid all our need for the overrides

src/app-layout/mobile-toolbar/index.tsx
```diff
@@ -130,33 +129,41 @@ export function MobileToolbar({
...
                 iconName={item.trigger.iconName}
                 iconSvg={item.trigger.iconSvg}
                 badge={item.badge}
                 ariaLabel={item.ariaLabels?.triggerButton}
-                onClick={() => drawers.onChange({ activeDrawerId: item.id })}
                 ariaExpanded={drawers.activeDrawerId !== undefined}
                 testId={`awsui-app-layout-trigger-${item.id}`}
               />
-            </span>
+            </div>
```
just-boris, line 147: Changed from `<span>` to `<div>` in a few places, because button dropdown renders a `<div>` and it cannot be nested into `<span>`, so we need to upgrade all elements to div all way up

src/app-layout/visual-refresh/context.tsx
```diff
@@ -33,15 +33,15 @@ import { useContainerQuery } from '@cloudscape-design/component-toolkit';
 import useBackgroundOverlap from './use-background-overlap';
 
 interface AppLayoutInternals extends AppLayoutProps {
-  activeDrawerId?: string | null;
+  activeDrawerId?: string | undefined;
```
just-boris, line 36: There was a discrepancy between refresh and classic version. Classic used `undefined` to notify "no active drawer" state and refresh used "null".

To avoid surprises on consumers end when they implement dynamic switch to refresh, we should use a consistent approach

src/app-layout/visual-refresh/drawers.tsx
```diff
@@ -396,21 +358,31 @@ export function MobileTriggers() {
...
+              badgeColor="red"
+              onClick={clickHandler}
+              variant="icon"
+              __nativeAttributes={{ 'aria-haspopup': true }}
+            />
+          )}
           onItemClick={({ detail }) => {
             handleDrawersClick(detail.id);
           }}
-          hasOverflowBadge={overflowItemHasBadge}
```
just-boris, line 412: Badges on button dropdown triggers are not supported in this PR. 

My suggestion would be to make this feature built into button dropdown, so it will automatically render the badge when any dropdown item has it. 

It is useful for our use-case, so can also be useful for customers with standalone use-case in button dropdown

src/app-layout/visual-refresh/drawers.tsx
just-boris: Just finished annotating this PR with comments.

This change is sort of intended, pointed in another comment: https://github.com/cloudscape-design/components/pull/1444#discussion_r1293588290

src/app-layout/visual-refresh/drawers.tsx
katiegeorge: I see. So you're suggesting that we do that in a separate PR, and make it a built-in feature for the ButtonDropdown component. That's fine by me, but if there is design pushback we may have to re-implement it here this way.

src/app-layout/visual-refresh/context.tsx
katiegeorge: Agreed they should be consistent. In the API proposal, we actually agreed on changing both to use `null` instead of `undefined` since there is a valid use case for no active drawer open. So this is fine for now, but planning on changing both to use `null` in a future PR (there are also a few other API discrepancies between the two that I'll need to fix simultaneously)

just-boris: Fine by me either way as long as it is consistent

PR #1451, opened 2023-08-16
feat: Add focusSplitPanel() ref function to app layout

avinashbot:
### Description

Added a .focusSplitPanel() ref function to the app layout that focuses the first focusable element in the panel (the slider).

Related links, issue #, if available: AWSUI-21470?

### How has this been tested?

Unit tests.

<details>

#### Correctness

#### Security

#### Testing

</details>

---


taheramr (commented):
will this API be only available in VR?

src/split-panel/index.tsx
```diff
@@ -32,290 +32,301 @@ export { SplitPanelProps };
...
+    const [relativeSize, setRelativeSize] = useState(0);
+    const [maxSize, setMaxSize] = useState(size);
+    const minSize = position === 'bottom' ? MIN_HEIGHT : MIN_WIDTH;
+    const cappedSize = getLimitedValue(minSize, size, maxSize);
+    const appLayoutMaxWidth = isRefresh && position === 'bottom' ? contentWidthStyles : undefined;
+
+    useImperativeHandle(
+      ref,
+      () => ({
+        focus() {
```
just-boris, line 69: Should we call it `focusResizer` to allow adding `focusPanelClose` method without confusion in the future if needed

timogasda: On the other hand, a general `focus` method could be what people should always call when they need to "move focus to the split panel". And if the recommended element to focus changes in the future, we can easily swap out that logic without changing the recommendation.

avinashbot: Ah, Timo got there just before I did 😅

just-boris: If it would have been unconditional focus on the resizer, we would make it built in bound to the open state.

Since we expose an API instead, it seems like a context-dependent feature, so "just focus" may become limiting

just-boris (approved):

timogasda (approved):

PR #1461, opened 2023-08-17
chore: Allow runtime drawers to be open by default

just-boris:
### Description

Support `awsuiPlugins.appLayout.registerDrawer({ defaultActive: true })` to allow opening it on the initial rendering

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/__tests__/runtime-drawers.test.tsx
+++ b/src/app-layout/__tests__/runtime-drawers.test.tsx
@@ -7,6 +7,7 @@ import { InternalDrawerProps } from '../../../lib/components/app-layout/drawer/i
 import { awsuiPlugins, awsuiPluginsInternal } from '../../../lib/components/internal/plugins/api';
 import { DrawerConfig } from '../../../lib/components/internal/plugins/controllers/drawers';
 import createWrapper from '../../../lib/components/test-utils/dom';
+import { singleDrawer } from './utils';
 
 beforeEach(() => {
   awsuiPluginsInternal.appLayout.clearRegisteredDrawers();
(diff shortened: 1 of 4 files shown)


src/app-layout/utils/use-drawers.ts
```diff
@@ -50,12 +52,22 @@ function useRuntimeDrawers(
...
+  const prevDrawersRef = useRef<Array<RuntimeDrawerConfig>>([]);
+
   useEffect(() => {
     if (disableRuntimeDrawers) {
       return;
     }
     const unsubscribe = awsuiPluginsInternal.appLayout.onDrawersRegistered(drawers => {
       setRuntimeDrawers(convertRuntimeDrawers(drawers));
+      if (prevDrawersRef.current) {
+        const diff = drawers.filter(drawer => !prevDrawersRef.current.find(d => drawer.id === d.id));
```
YueyingLu, line 64: With it, the newly registered drawer will open even its priority is lower than the current open one. Is it expected?

src/app-layout/__tests__/runtime-drawers.test.tsx
```diff
@@ -73,6 +74,66 @@ describe('Runtime drawers', () => {
...
+      ...drawerDefaults,
+      id: 'second',
+      defaultActive: true,
+      mountContent: container => (container.textContent = 'second drawer content'),
+    });
+    const { wrapper } = await renderComponent(<AppLayout />);
+    expect(wrapper.findActiveDrawer()!.getElement()).toHaveTextContent('second drawer content');
+  });
+
+  test('keeps currently active drawer if a drawer with lower priority was registered', async () => {
```
YueyingLu, line 102: Could it be the second drawer is open because the second drawer is registered later, not the priority?

YueyingLu (changes requested):

src/app-layout/__tests__/runtime-drawers.test.tsx
just-boris: Good finding, I did two bugs which cancelled each other.

Updated the PR, with the new behavior. Now it is strictly in loading order. Loaded first gets open, all others do not (this is also better for UX to prevent them jumping too much)

src/app-layout/utils/use-drawers.ts
```diff
@@ -50,12 +52,21 @@ function useRuntimeDrawers(
...
 
+  const prevDrawersRef = useRef<Array<RuntimeDrawerConfig>>([]);
+
   useEffect(() => {
     if (disableRuntimeDrawers) {
       return;
     }
     const unsubscribe = awsuiPluginsInternal.appLayout.onDrawersRegistered(drawers => {
       setRuntimeDrawers(convertRuntimeDrawers(drawers));
+      if (!prevDrawersRef.current || !prevDrawersRef.current.find(drawer => drawer.defaultActive)) {
```
YueyingLu, line 63: If the old drawer is closed, shall the new defaultActive drawer open? Maybe check `activeDrawerId` so we wont open a new drawer when there is one open already?

just-boris: Good point, updated the implementation.

The new logic is also much simpler "defaultActive only works when there was no any open drawer before. If there was an open drawer (by a user or from another instance), new drawer won't open"

YueyingLu (approved):

PR #1462, opened 2023-08-18
refactor: Deduplicate drawers handling code between classic and refresh implementations

just-boris:
### Description

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/__tests__/desktop.test.tsx
+++ b/src/app-layout/__tests__/desktop.test.tsx
@@ -193,13 +193,6 @@ describeEachThemeAppLayout(false, () => {
     expect(wrapper.findDrawersSlider()!.getElement()).toHaveFocus();
   });
 
-  test(`Should fire resize when expected`, () => {
-    const { wrapper } = renderComponent(<AppLayout contentType="form" {...resizableDrawer} />);
-
-    act(() => wrapper.findDrawersTriggers()![0].click());
-    expect(wrapper.findDrawersSlider()!.getElement()).toHaveFocus();
-  });
-
   test('should change size via keyboard events on slider handle', () => {
     const onResize = jest.fn();
     const drawers: Required<InternalDrawerProps> = {
--- a/src/app-layout/index.tsx
+++ b/src/app-layout/index.tsx
@@ -1,7 +1,7 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 import clsx from 'clsx';
-import React, { useCallback, useEffect, useImperativeHandle, useMemo, useRef, useState } from 'react';
+import React, { useCallback, useEffect, useImperativeHandle, useRef, useState } from 'react';
 import { getBaseProps } from '../internal/base-component';
 import { useControllable } from '../internal/hooks/use-controllable';
 import { useMobile } from '../internal/hooks/use-mobile';
@@ -34,8 +34,6 @@ import { useVisualRefresh } from '../internal/hooks/use-visual-mode';
 import ContentWrapper, { ContentWrapperProps } from './content-wrapper';
 import { Drawer, DrawerTriggersBar } from './drawer';
 import { ResizableDrawer } from './drawer/resizable-drawer';
-import { DrawerItem, InternalDrawerProps } from './drawer/interfaces';
-import { togglesConfig } from './toggles';
 import { SideSplitPanelDrawer } from './split-panel-drawer';
 import useAppLayoutOffsets from './utils/use-content-width';
 import { isDevelopment } from '../internal/is-development';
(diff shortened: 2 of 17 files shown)


src/app-layout/__tests__/desktop.test.tsx
```diff
@@ -193,13 +193,6 @@ describeEachThemeAppLayout(false, () => {
     expect(wrapper.findDrawersSlider()!.getElement()).toHaveFocus();
   });
 
-  test(`Should fire resize when expected`, () => {
```
just-boris, line 196: Duplicate test, there is very same test above

src/app-layout/visual-refresh/context.tsx
```diff
@@ -394,68 +396,58 @@ export const AppLayoutInternalsProvider = React.forwardRef(
...
 
-    const activeDrawer = drawers?.items.find(drawer => drawer.id === activeDrawerId);
-
     const {
       refs: drawersRefs,
       setFocus: focusDrawersButtons,
       loseFocus: loseDrawersFocus,
       setLastInteraction: setDrawerLastInteraction,
-    } = useDrawerFocusControl([activeDrawerId, activeDrawer?.resizable], activeDrawerId !== undefined, true);
+    } = useDrawerFocusControl([activeDrawerId, activeDrawer?.resizable], true, true);
```
just-boris, line 426: This line contains a bug, `activeDrawerId` was never `undefined` before  (because it was `null`). This expression always resolved to `true` even when the drawer is closed.

To keep the changes in this PR minimal, I put `true` as a compatibility value and I am going to refactor focus management in the next PRs

src/app-layout/visual-refresh/context.tsx
```diff
@@ -394,68 +396,58 @@ export const AppLayoutInternalsProvider = React.forwardRef(
...
-      changeHandler: 'onChange',
-    });
+    const {
+      drawers,
+      activeDrawer,
+      activeDrawerId,
+      onActiveDrawerChange,
+      onActiveDrawerResize,
+      activeDrawerSize,
+      ...drawersProps
```
taheramr, line 406: minor:
shouldn't we just use the `ariaLabel` instead of the `...drawersProps` since its the only remaining return value from the hook?

src/app-layout/index.tsx
```diff
@@ -173,60 +146,29 @@ const OldAppLayout = React.forwardRef(
...
-        ? getAllDrawerItems()?.filter((drawerItem: DrawerItem) => drawerItem.id === activeDrawerId)[0]
-        : undefined;
+    const {
+      drawers,
+      activeDrawer,
+      activeDrawerSize,
+      activeDrawerId,
+      onActiveDrawerChange,
+      onActiveDrawerResize,
+      ...drawersProps
```
taheramr, line 156: same as the other comment for the `...drawersProps` vs just using the `ariaLabel` directly

src/app-layout/visual-refresh/context.tsx
```diff
@@ -394,68 +396,58 @@ export const AppLayoutInternalsProvider = React.forwardRef(
...
+    } = useDrawers(
+      props as InternalDrawerProps,
```
taheramr, line 408: I'm a bit confused with this line, so from my understanding when we render the AppLayout the runtime API attaches the required props, so we cast it here ( even if the props definition does not include `drawers` or `__disableRuntimeDrawers`.

should we then add some kind of guarding conditions in the `useDrawers` hook to check if the props have the required properties passed to it?

src/app-layout/visual-refresh/context.tsx
just-boris: There will be more changes in future PRs: https://github.com/cloudscape-design/components/pull/1416

Already prepared it here to not forget to update these lines every time a new feature is added

src/app-layout/visual-refresh/context.tsx
just-boris: Both of these properties are optional, what kind of guarding condition do see possible here?

src/app-layout/index.tsx
just-boris: Same answer: https://github.com/cloudscape-design/components/pull/1462#discussion_r1301621092

src/app-layout/visual-refresh/context.tsx
taheramr: I see, I looked again and its fine.

It just does not look correct to me when we are casting the whole props as `InternalDrawerProps`, so we are sending a lot of un-required fields to the hook. as a minor thingy, can we selectively send them?

like:
```
const useDrawersFields = {
drawers: (props as InternalDrawerProps)?.drawers,
__disableRuntimeDrawers: (props as InternalDrawerProps)?. __disableRuntimeDrawers
}
```

Will approve tho, just a suggestion .

taheramr (approved):

src/app-layout/visual-refresh/context.tsx
just-boris: I see. I will revisit it when `props.drawers` becomes public and will not require type casting

PR #1468, opened 2023-08-21
refactor: Reuse nodeBelongs in focus tracker utility

just-boris:
### Description

`nodeBelongs` utility already contains check for portals, no need for custom implementation via `viewportId`

This PR is a preparation for `containsOrEqual` removal https://github.com/cloudscape-design/components/pull/1465

### How has this been tested?

PR build ([this test suite](https://github.com/cloudscape-design/components/blob/main/src/date-range-picker/__integ__/events.test.ts) in particular covers the focus tracker functionality)

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/utils/node-belongs.ts
+++ b/src/internal/utils/node-belongs.ts
@@ -10,11 +10,11 @@ import { containsOrEqual, findUpUntil } from './dom';
  * @param container Container node
  * @param target Node that is checked to be a descendant of the container
  */
-export function nodeBelongs(container: Node | null, target: Node): boolean {
-  const portal = findUpUntil(
-    target as HTMLElement,
-    node => node instanceof HTMLElement && !!node.dataset.awsuiReferrerId
-  );
+export function nodeBelongs(container: Node | null, target: Node | EventTarget | null): boolean {
+  if (!(target instanceof Node)) {
+    return false;
+  }
+  const portal = findUpUntil(target as HTMLElement, node => !!node.dataset.awsuiReferrerId);
   const referrer = portal instanceof HTMLElement ? document.getElementById(portal.dataset.awsuiReferrerId ?? '') : null;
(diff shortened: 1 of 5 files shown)


src/internal/utils/node-belongs.ts
```diff
@@ -10,11 +10,11 @@ import { containsOrEqual, findUpUntil } from './dom';
...
-export function nodeBelongs(container: Node | null, target: Node): boolean {
-  const portal = findUpUntil(
-    target as HTMLElement,
-    node => node instanceof HTMLElement && !!node.dataset.awsuiReferrerId
-  );
+export function nodeBelongs(container: Node | null, target: Node | EventTarget | null): boolean {
+  if (!(target instanceof Node)) {
+    return false;
+  }
+  const portal = findUpUntil(target as HTMLElement, node => !!node.dataset.awsuiReferrerId);
```
connorlanigan, line 17: This type assertion (*`as HTMLElement`) is not safe either, right? We can't guarantee that `nodeBelongs` will never be called on e.g. an SVG element. Can we replace the type assertion with a check?

connorlanigan (changes requested):

src/internal/utils/node-belongs.ts
just-boris: Tried this while working on this change: https://github.com/cloudscape-design/components/compare/3d036443c0771b25e3143449f7cdbf65156289c1..550b17351b54911a505868360a7d05a5e4dc6bdb

It broke some logic in area chart, where we use this utility with SVG nodes. 


Fixing this properly would require refactoring `findUpUntil` which stretches the scope of this work a bit. 

Helping me fixing this part of our codebase would be also appreciated.

connorlanigan (approved):

PR #1470, opened 2023-08-22
fix: Prevent sticky scrollbar from overlapping table cells

avinashbot:
### Description

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/table/sticky-scrollbar/sticky-scrollbar.tsx
```diff
@@ -26,6 +27,12 @@ function StickyScrollbar(
   const scrollbarContentRef = React.useRef<HTMLDivElement>(null);
   const mergedRef = useMergeRefs(ref, scrollbarRef);
 
+  /**
+   * If the height of the scrollbar is 0, we're likely on a platform that uses
+   * overlay scrollbars (e.g. Mac).
+   */
+  const forceOffsetScrollbar = browserScrollbarSize().height === 0;
```
rubencarvalho, line 34: Maybe it's worth just briefly checking with UI folks to make sure we're OK with applying this treatment to all tables.

avinashbot: I'm reasonably confident this is the right move, especially because it's not the first time we've heard this from people. Still, good call - going to check with on-call design before I merge.

rubencarvalho (dismissed):
LGTM, approved pending confirmation from UI/UX folks 👍

src/table/sticky-scrollbar/sticky-scrollbar.tsx
```diff
@@ -39,7 +46,7 @@ function StickyScrollbar(
       ref={mergedRef}
       className={clsx(
         styles['sticky-scrollbar'],
-        offsetScrollbar && styles['sticky-scrollbar-offset'],
+        (offsetScrollbar || forceOffsetScrollbar) && styles['sticky-scrollbar-offset'],
```
rubencarvalho, line 49: If we do this we won't need the `offsetScrollbar` argument, because it will cover all cases.

avinashbot: Looks like sticky scrollbar _always_ enables the offset regardless of the scrollbar style. Was the overlap of the scrollbar over interactive content the only reason we did this? If so, then we might be able to.

```
const hasStickyColumns = !!((stickyColumns?.first ?? 0) + (stickyColumns?.last ?? 0) > 0);
```

rubencarvalho: > Looks like sticky scrollbar always enables the offset regardless of the scrollbar style.
I'm not sure I understood this.

> Was the overlap of the scrollbar over interactive content the only reason we did this? 
Yes!

avinashbot: I phrased it a little weird, but I get what you mean! I'll publish another revision with it enabled for all browsers with floating scrollbars. I have UX/visual design approval :)

rubencarvalho (approved):

PR #1476, opened 2023-08-23
chore: Mark the current funnel step name with a data attribute

connorlanigan:
### Description

This PR adds a data attribute to the Wizard. This attribute can be used to determine the name of the current step.

The Form component (the single-page counterpart to the Wizard) does not receive a similar attribute, because we do not consider it to have a proper step name. Instead, we're already using the funnel name as the step name. See the value for `stepNameSelector` in this line:
https://github.com/cloudscape-design/components/blob/99e8bbfd1a4012be17e3acbb617e4fcda7e65298/src/form/index.tsx#L39

I've added a test to make this explicit.

### How has this been tested?

Added unit tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/form/__tests__/analytics.test.tsx
+++ b/src/form/__tests__/analytics.test.tsx
@@ -7,11 +7,22 @@ import createWrapper from '../../../lib/components/test-utils/dom';
 import Button from '../../../lib/components/button';
 import Form from '../../../lib/components/form';
 import Modal from '../../../lib/components/modal';
+import BreadcrumbGroup from '../../../lib/components/breadcrumb-group';
 
 import { FunnelMetrics } from '../../../lib/components/internal/analytics';
 import { useFunnel } from '../../../lib/components/internal/analytics/hooks/use-funnel';
 
 import { mockFunnelMetrics } from '../../internal/analytics/__tests__/mocks';
+
+// JSDom does not support the `textContent` property. For this test, `textContent` is close enough.
+Object.defineProperty(HTMLElement.prototype, 'innerText', {
+  get() {
+    return this.textContent;
+  },
+  set(v) {
+    this.textContent = v;
+  },
+});
 
 describe('Form Analytics', () => {
   beforeEach(() => {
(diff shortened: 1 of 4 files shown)


src/form/__tests__/analytics.test.tsx
```diff
@@ -48,6 +59,27 @@ describe('Form Analytics', () => {
     );
   });
 
+  test('includes the current breadcrumb as the step name in the funnelStepStart event', () => {
```
avinashbot, line 62: Looks like an extra breadcrumb group test landed in here? Not complaining, always love extra tests

avinashbot (approved):

src/form/__tests__/analytics.test.tsx
```diff
@@ -7,12 +7,23 @@ import createWrapper from '../../../lib/components/test-utils/dom';
...
 import Form from '../../../lib/components/form';
 import Modal from '../../../lib/components/modal';
+import BreadcrumbGroup from '../../../lib/components/breadcrumb-group';
 
 import { FunnelMetrics } from '../../../lib/components/internal/analytics';
 import { useFunnel } from '../../../lib/components/internal/analytics/hooks/use-funnel';
 
 import { mockFunnelMetrics } from '../../internal/analytics/__tests__/mocks';
 
+// JSDom does not support the `textContent` property. For this test, `textContent` is close enough.
```
avinashbot, line 17: You mean "does not support the `innerText`"

taheramr (approved):

src/form/__tests__/analytics.test.tsx
connorlanigan: This is intentional: when the Form wants to determine the current step name (for the funnelStepStart event), it does not use any of its own content, but actually looks for a BreadcrumbGroup component on the page. It then reaches into that BreadcrumbGroup to extract the name.

avinashbot (approved):

PR #1479, opened 2023-08-23
fix: Fixes aria expanded and aria controls in AppLayout

katiegeorge:
### Description

We weren't properly using `aria-expanded` or `aria-controls` for the triggers in AppLayout. This adds those attributes to drawers, tools, navigation, and split panel when appropriate.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/drawer/index.tsx
+++ b/src/app-layout/drawer/index.tsx
@@ -66,7 +66,7 @@ export const Drawer = React.forwardRef(
           iconName={iconName}
           ariaLabel={openLabel}
           onClick={() => onToggle(true)}
-          ariaExpanded={false}
+          ariaExpanded={isOpen ? undefined : false}
         />
       </TagName>
     );
--- a/src/app-layout/drawer/index.tsx
+++ b/src/app-layout/drawer/index.tsx
@@ -133,7 +133,7 @@ interface DrawerTriggerProps {
   testUtilsClassName?: string;
   ariaLabel: string | undefined;
   ariaExpanded: boolean;
-  ariaControls: string | undefined;
+  ariaControls?: string;
   badge: boolean | undefined;
   itemId?: string;
   isActive: boolean;
@@ -203,6 +203,7 @@ export const DrawerTriggersBar = ({ isMobile, topOffset, bottomOffset, drawers }
         style={{ top: topOffset, bottom: bottomOffset }}
         className={clsx(styles['drawer-content'])}
         role="toolbar"
+        aria-orientation="vertical"
       >
         {!isMobile && (
           <aside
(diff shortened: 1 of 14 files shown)


src/app-layout/visual-refresh/drawers.tsx
```diff
@@ -220,11 +222,15 @@ function DesktopTriggers() {
...
         })}
+        role="toolbar"
+        aria-orientation="vertical"
       >
         {visibleItems.map(item => {
           return (
             <TriggerButton
               ariaLabel={item.ariaLabels?.triggerButton}
+              ariaExpanded={item.id === activeDrawerId}
+              ariaControls={item.id}
```
Al-Dani, line 233: So, when we select a drawer, we set it's `item.id` as an `activeDrawerId`, right? In that case all other trigger buttons will point to a non-existing element in the DOM, which is not good.

I think we need something like `ariaControls={item.id === activeDrawerId ? item.id : undefined}`, but we can double-check.

The same applies to other ariaControls :)

src/app-layout/mobile-toolbar/index.tsx
```diff
@@ -141,7 +141,7 @@ export function MobileToolbar({
                 iconSvg={item.trigger.iconSvg}
                 badge={item.badge}
                 ariaLabel={item.ariaLabels?.triggerButton}
-                ariaExpanded={drawers.activeDrawerId !== undefined}
+                ariaExpanded={drawers.activeDrawerId === item.id}
```
Al-Dani, line 144: no `aria-controls` here?

src/app-layout/drawer/index.tsx
```diff
@@ -189,6 +202,7 @@ export const DrawerTriggersBar = ({ isMobile, topOffset, bottomOffset, drawers }
         ref={triggersContainerRef}
         style={{ top: topOffset, bottom: bottomOffset }}
         className={clsx(styles['drawer-content'])}
+        role="toolbar"
```
Al-Dani, line 205: no `aria-orientation` here?

Al-Dani (changes requested):

src/app-layout/mobile-toolbar/index.tsx
katiegeorge: My thinking is that it isn't necessary on mobile, since you won't ever see the panel and the trigger button at the same time (its a full-screen overlay). So it would indicate that its controlling something that isn't in the DOM yet.

src/app-layout/visual-refresh/drawers.tsx
katiegeorge: I think this makes a lot of sense! Doesn't make sense to have a control for something that doesn't exist (see my other comment around mobile behavior)

src/app-layout/visual-refresh/trigger-button.tsx
```diff
@@ -12,20 +12,34 @@ export interface TriggerButtonProps {
   className?: string;
   iconName?: IconProps.Name;
   iconSvg?: React.ReactNode;
+  ariaExpanded: boolean | undefined;
+  ariaControls?: string;
```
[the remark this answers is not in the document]
katiegeorge: Good catch. I'll update to be consistent.

Al-Dani (approved):

PR #1492, opened 2023-08-29
feat: Table resize handler requires keyboard activation

pan-kot:
### Description

Made resize handles buttons that require activation with Space/Enter before accepting further keyboard input.
Fixed a bug with controlled mode when onWidthChange handler was never called - now it is called upon resize commit with Space/Enter. The operation can also be discarded with Escape.

### How has this been tested?

Updated existing and added new unit- and integration tests. Screenshot tests, dry-run.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/table/resizer/index.tsx
+++ b/src/table/resizer/index.tsx
@@ -9,6 +9,9 @@ import styles from './styles.css.js';
 import { KeyCode } from '../../internal/keycode';
 import { DEFAULT_COLUMN_WIDTH } from '../use-column-widths';
 import { useStableCallback } from '@cloudscape-design/component-toolkit/internal';
+import ScreenreaderOnly from '../../internal/components/screenreader-only';
+import { useUniqueId } from '../../internal/hooks/use-unique-id';
+import { joinStrings } from '../../internal/utils/strings';
 
 interface ResizerProps {
   onDragMove: (newWidth: number) => void;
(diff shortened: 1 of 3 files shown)


src/table/resizer/index.tsx
```diff
@@ -97,83 +108,138 @@ export function Resizer({
...
-        isDragging && styles['resizer-active'],
-        (resizerHasFocus || showFocusRing) && styles['has-focus']
-      )}
-      onMouseDown={event => {
-        if (event.button !== 0) {
-          return;
+  }, [headerCell, isDragging, isKeyboardDragging, onFinishStable, resizerHasFocus, handlers]);
+
+  const resizerWidthId = useUniqueId();
+  const resizerRole = isKeyboardDragging ? 'separator' : 'button';
```
YueyingLu, line 187: I have some concerns to change the role with state changing. Will it be confusing by users? 
In addition, the button is labelled by header name without any indication it is used to resize the cell.

src/table/resizer/index.tsx
```diff
@@ -97,83 +108,138 @@ export function Resizer({
...
+            setIsKeyboardDragging(false);
+            handlers?.resetColumnWidth();
+          }
+        }}
+        role={resizerRole}
+        {...resizerAriaProps}
+        tabIndex={tabIndex}
+        data-focus-id={focusId}
+      />
+      <ScreenreaderOnly id={resizerWidthId}>{headerCellWidthString}</ScreenreaderOnly>
```
YueyingLu, line 241: Is the 'aria-valuetext' announcing it already? Is it because of changing to button role?

pan-kot: Yes, the `aria-valuetext` is only added when the role is "separator".

src/table/resizer/index.tsx
pan-kot: I think that should not matter much whether it is the same element or, say, a focus transition from one element (button) to another (separator). The button is labelled the same way as the separator was. Probably a specific label or, say, ARIA role-description would make sense here, but am not sure if that is really necessary as so far we don't have it.

src/table/resizer/index.tsx
YueyingLu: I think it makes more sense to me to announce it after "Left"/"Right" key down, to give user more instant feedback when resizing. So we do not need to announce it after the resize is over.

YueyingLu: Looks like the aria-valuetext is not announced instantly when resizing using keyboard.

src/table/resizer/index.tsx
pan-kot: Yes, I am planning to go this way instead of adding the description preemptively.

src/table/resizer/index.tsx
pan-kot: I observe that more often than not with Chrome+VO but it is consistent with the current behaviour.

YueyingLu: You are right, looks good on Safari + VO.

PR #1493, opened 2023-08-29
chore: Table grid navigation replace widget with dialog

pan-kot:
### Description

Updated the grid navigation based on the team's feedback. The "widget" cells are no longer explicitly present, instead - the focus handling is now done uniformly for all cells and elements.

Whenever there is a need to support cell content requiring arrow-based navigation the consumer must ensure the content only becomes fully interactive upon activation, same as the built-in inline editing feature works. The table no longer interferes when the focused element or one of its parents has role="dialog" or data-awsui-table-suppress-navigation="true".

Updated the docs and the test page accordingly.

### How has this been tested?

Updated unit- and integration tests.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/table/table-role/grid-navigation.md
+++ b/src/table/table-role/grid-navigation.md
@@ -1,52 +1,48 @@
 # Table grid navigation
 
-Tables with interactive elements especially those featuring row selection or inline cell editing are assigned or can be assigned the role "grid" which implies extended keyboard navigation as per https://www.w3.org/WAI/ARIA/apg/patterns/grid.
+Tables with interactive elements require ARIA role "grid" and extended keyboard navigation as per https://www.w3.org/WAI/ARIA/apg/patterns/grid.
 
-## Navigation commands
+## Basic keyboard navigation
+
+The grid cells both text-only or even empty and those with one or multiple interactive elements are navigable. When a cell does not have interactive elements as its content the cell itself receives focus. Otherwise - one of the content elements is focused (can be the first, the last, or even the nth depending on the move direction).
 
 The list of supported navigation commands is:
 
-- `Right Arrow`: Moves focus one cell to the right. If focus is on the right-most cell in the row, focus does not move.
-- `Left Arrow`: Moves focus one cell to the left. If focus is on the left-most cell in the row, focus does not move.
-- `Down Arrow`: Moves focus one cell down. If focus is on the bottom cell in the column, focus does not move.
-- `Up Arrow`: Moves focus one cell up. If focus is on the top cell in the column, focus does not move.
-- `Page Down`: Moves focus down an customer-determined number of rows. If focus is in the last row of the grid, focus does not move.
-- `Page Up`: Moves focus up an customer-determined number of rows. If focus is in the first row of the grid, focus does not move.
-- `Home`: moves focus to the first cell in the row.
-- `End`: moves focus to the last cell in the row.
-- `Control + Home`: moves focus to the first cell in the first row.
-- `Control + End`: moves focus to the last cell in the last row.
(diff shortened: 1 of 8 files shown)


src/table/table-role/grid-navigation.md
```diff
@@ -1,52 +1,46 @@
 # Table grid navigation
 
-Tables with interactive elements especially those featuring row selection or inline cell editing are assigned or can be assigned the role "grid" which implies extended keyboard navigation as per https://www.w3.org/WAI/ARIA/apg/patterns/grid.
+Tables with interactive elements require ARIA role "grid" and extended keyboard navigation as per https://www.w3.org/WAI/ARIA/apg/patterns/grid.
```
timogasda, line 3: nit: I know the doc is called "Table grid navigation" but it's probably good to clarify: "Tables that use grid navigation and have interactive elements [...]". (or something like that)

src/table/table-role/grid-navigation.md
```diff
@@ -1,52 +1,46 @@
...
 
-## Navigation commands
+## Basic keyboard navigation
 
-The list of supported navigation commands is:
+The grid cells both text-only or even empty and those with one or multiple interactive elements are navigable. When a cell does not have interactive elements as its content the cell itself receives focus. Otherwise - one of the content elements is focused (can be the first, the last, or even the nth depending on the move direction).
```
timogasda, line 7: nit/writing: "All grid cells are navigable, no matter if [...]"

src/table/table-role/utils.ts
```diff
@@ -137,16 +124,32 @@ export function restoreTableFocusables(table: HTMLTableElement) {
...
+    if (tagName === 'td' || tagName === 'th') {
+      return false;
+    }
+    if (
+      current.getAttribute('role') === 'dialog' ||
+      current.getAttribute('data-awsui-table-suppress-navigation') === 'true'
+    ) {
+      return true;
+    }
+    current = current.parentElement;
```
timogasda, line 144: Could we use the existing `findUpUntil` util for this?

timogasda (commented):
Mostly minor comments.

src/table/table-role/utils.ts
```diff
@@ -4,41 +4,50 @@
 import { getFocusables as getActualFocusables } from '../../internal/components/focus-lock/utils';
 import { FocusedCell } from './interfaces';
 
+// For the grid to have a single Tab stop all interactive element indices are updated to be -999.
+// The elements having tab index -999 are eligible for keyboard navigation but not for Tab navigation.
+const PSEUDO_FOCUSABLE_TAB_INDEX = -999;
```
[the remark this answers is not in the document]
pan-kot: The trick here is that the table not only disables focus but also re-enables it. However, we don't want enabling focus for elements not disabled by the table which can happen prior or during the table interaction. The current logic assigns -999 and then changes it back to 0 assuming the content elements use -1 when it is needed to mute the focus.

This also works for the following complex case:
1. The user opens an in-cell dropdown element. The dropdown updates its focus-lock tabIndex to be 0;
2. The table overrides the focus-lock tabIndex to be -999;
3. The user closes the dropdown. The dropdown updates its focus-lock tabIndex to be -1.

The focus lock is no longer reachable with either Tab or focus navigation which is expected. When the focus leaves the table the focus-lock tabIndex remains -1.

src/table/table-role/utils.ts
pan-kot: Technically, yes -- but that would be awkward due to the `tagName === 'td' || tagName === 'th'` check. The idea is to exit as soon the cell element is reached as checking further is not expected. Because the findUpUntil does not have any exit mechanism it would take an external boolean flag to check if the condition has been met.

timogasda (approved):

PR #1495, opened 2023-08-30
chore: Creates internal drawer component

katiegeorge:
### Description

This is the first step to creating a Drawer component, which will release along with the public release of the drawers changes to the AppLayout. The goal of this PR is to get the code reviewed so that when the time comes to release the component publicly, it is mostly a "flip of a switch" to move the component out of the `internal` folder.

The purpose of the Drawer component is to provide a wrapper for the contents of a drawer, similar to the HelpPanel in the current tools slot or the SplitPanel in the splitPanel slot. It is built as a clone of the HelpPanel component, without the style overrides.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/i18n/messages-types.ts
+++ b/src/i18n/messages-types.ts
@@ -167,7 +167,7 @@ export interface I18nFormatArgTypes {
     "i18nStrings.warningIconAriaLabel": never;
   }
   "drawer": {
-    "loadingText": never;
+    "i18nStrings.loadingText": never;
   }
   "form": {
     "errorIconAriaLabel": never;
--- a/src/i18n/messages-types.ts
+++ b/src/i18n/messages-types.ts
@@ -156,6 +156,9 @@ export interface I18nFormatArgTypes {
       "unit": string | number;
     }
   }
+  "drawer": {
+    "loadingText": never;
+  }
   "flashbar": {
     "i18nStrings.ariaLabel": never;
     "i18nStrings.errorIconAriaLabel": never;
--- a/src/i18n/messages-types.ts
+++ b/src/i18n/messages-types.ts
@@ -166,6 +166,9 @@ export interface I18nFormatArgTypes {
     "i18nStrings.successIconAriaLabel": never;
     "i18nStrings.warningIconAriaLabel": never;
   }
+  "drawer": {
+    "loadingText": never;
+  }
   "form": {
     "errorIconAriaLabel": never;
   }
--- a/src/i18n/messages-types.ts
+++ b/src/i18n/messages-types.ts
@@ -156,9 +156,6 @@ export interface I18nFormatArgTypes {
       "unit": string | number;
     }
   }
-  "drawer": {
-    "loadingText": never;
-  }
   "flashbar": {
     "i18nStrings.ariaLabel": never;
     "i18nStrings.errorIconAriaLabel": never;
--- a/src/internal/components/drawer/index.tsx
+++ b/src/internal/components/drawer/index.tsx
@@ -1,0 +1,38 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import clsx from 'clsx';
+import React from 'react';
+import { getBaseProps } from '../../base-component';
+import InternalStatusIndicator from '../../../status-indicator/internal';
+import styles from './styles.css.js';
+import { applyDisplayName } from '../../utils/apply-display-name';
+import useBaseComponent from '../../hooks/use-base-component';
+import { DrawerProps } from './interfaces';
+import LiveRegion from '../../components/live-region';
+import { useInternalI18n } from '../../../i18n/context';
+
+export { DrawerProps };
+
+export default function Drawer({ header, children, loading, loadingText, ...restProps }: DrawerProps) {
+  const { __internalRootRef } = useBaseComponent('Drawer');
+  const baseProps = getBaseProps(restProps);
+  const i18n = useInternalI18n('drawer');
+  const containerProps = {
+    ...baseProps,
+    className: clsx(baseProps.className, styles.drawer),
+  };
+  return loading ? (
+    <div {...containerProps} ref={__internalRootRef}>
+      <InternalStatusIndicator type="loading">
+        <LiveRegion visible={true}>{i18n('loadingText', loadingText)}</LiveRegion>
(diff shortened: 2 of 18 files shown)


src/internal/components/drawer/interfaces.ts
```diff
@@ -0,0 +1,30 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import { BaseComponentProps } from '../../base-component';
+import React from 'react';
+
+export interface DrawerProps extends BaseComponentProps {
+  /**
+   * Header of the drawer.
```
jperals, line 8: Have these doc strings ben reviewed by content design?

src/internal/components/drawer/interfaces.ts
```diff
@@ -0,0 +1,30 @@
...
+  header?: React.ReactNode;
+
+  /**
+   * Main content of the drawer.
+   *
+   */
+  children?: React.ReactNode;
+
+  /**
+   * Renders the panel in a loading state. We recommend that you also set a `loadingText`.
```
jperals, line 21: "the panel" or "the drawer"?

src/internal/components/drawer/interfaces.ts
```diff
@@ -0,0 +1,30 @@
...
+   */
+  loading?: boolean;
+
+  /**
+   * Specifies the text that's displayed when the panel is in a loading state.
```
jperals, line 26: "the panel" or "the drawer"?

src/internal/components/drawer/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { DrawerProps } from './interfaces';
+import LiveRegion from '../../components/live-region';
+import { useInternalI18n } from '../../../i18n/context';
+
+export default function Drawer({ header, children, loading, loadingText, ...restProps }: DrawerProps) {
+  const baseProps = getBaseProps(restProps);
+  const i18n = useInternalI18n('drawer');
+  const containerProps = {
+    ...baseProps,
+    className: clsx(baseProps.className, styles.drawer),
```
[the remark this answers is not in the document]
just-boris: This is the current convention. All components inside the components repository need to support it for consistency. 

New packages (board-components and any other future one) do not have to do this

src/i18n/messages-types.ts
```diff
@@ -156,6 +156,9 @@ export interface I18nFormatArgTypes {
       "unit": string | number;
     }
   }
+  "drawer": {
+    "loadingText": never;
```
just-boris, line 160: What does `never` mean here? Is it temporary while the strings are not avaialble?

src/internal/components/drawer/interfaces.ts
katiegeorge: I'll updated to drawer.

src/internal/components/drawer/interfaces.ts
katiegeorge: They are copied word for word from the Help Panel documentation, other than `drawer` instead of `panel`. So they were reviewed when the help panel was released.

src/i18n/messages-types.ts
katiegeorge: If you look at this file, most of the types are `never`. Although I missed that this file is auto-generated, so I removed this update in this PR.

katiegeorge:
Realizing I need to update the `i18n` package with the updated drawer strings before merging this in. Moving to a draft PR until that is merged.

src/internal/components/drawer/styles.scss
```diff
@@ -0,0 +1,41 @@
...
+*/
+
+/* stylelint-disable @cloudscape-design/no-implicit-descendant, selector-max-type */
+@use '../../styles/tokens' as awsui;
+@use '../../styles' as styles;
+
+.drawer {
+  @include styles.styles-reset;
+  word-wrap: break-word;
+  padding: awsui.$space-scaled-l awsui.$space-panel-side-right awsui.$space-scaled-xxxl awsui.$space-panel-side-left;
```
jperals, line 13: Why does the help panel have zero bottom padding and the drawer `$space-scaled-xxxl`?

src/internal/components/drawer/styles.scss
```diff
@@ -0,0 +1,41 @@
...
+}
+
+.header {
+  @include styles.font-panel-header;
```
jperals, line 17: Would it make sense to de-duplicate this block?

jperals:
Do we want to add the new testing utilities as well? I am guessing we could add them to `src/test-utils/dom/internal/drawer`. And then you could use them in your tests instead of `findByClassName`.

src/internal/components/drawer/styles.scss
katiegeorge: Help panel has built-in padding for the last item (see [here](https://github.com/cloudscape-design/components/blob/main/src/help-panel/styles.scss#L76)), but since Drawers won't be content-based, the padding goes on the container.

src/internal/components/drawer/styles.scss
katiegeorge: Do you mean make it the same for help panel and drawer? The header scss is already different, since the drawer header doesn't have the same style overrides as help panel. And I don't want to overlap them too much, since they may diverge even more in the future (ie, there are talks of putting other action buttons in the drawer header).

katiegeorge:
> Do we want to add the new testing utilities as well? I am guessing we could add them to `src/test-utils/dom/internal/drawer`. And then you could use them in your tests instead of `findByClassName`.

My plan is to add the test utilities when we make this public. The API review document includes `findHeader` and `findContent` just like the Help Panel, so that's the plan going forward.

src/internal/components/drawer/styles.scss
jperals: The CSS rules in this `.header` selector look exactly the same to me as in help panel (see [here](https://github.com/cloudscape-design/components/blob/main/src/help-panel/styles.scss#L107)) with the only exception that the rule `border: 0` is missing here in the drawer —I assume you removed it because it's actually unnecessary (and it seems unnecessary in the help panel too).

Not sure how much if at all adding action buttons in the drawer header would need to affect this block, with this information I would personally still be inclined to deduplicate, but won't block the PR on this.

jperals (approved):

PR #1502, opened 2023-09-01
fix: JSDOM compatibility in mixed line bar chart

just-boris:
### Description

In JSDOM, when calling `element.blur()`, `event.relatedTarget` can be `Document` instance, which has no `dataset` property. 

In real browsers, it is `null` in this case, and works as expected

Related links, issue #, if available: AWSUI-22179

### How has this been tested?

Added an extra unit test

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/utils/node-belongs.ts
+++ b/src/internal/utils/node-belongs.ts
@@ -15,7 +15,10 @@ export function nodeBelongs(container: Node | null, target: Node | EventTarget |
   if (!(target instanceof Node)) {
     return false;
   }
-  const portal = findUpUntil(target as HTMLElement, node => !!node.dataset.awsuiReferrerId);
+  const portal = findUpUntil(
+    target as HTMLElement,
+    node => node instanceof HTMLElement && !!node.dataset.awsuiReferrerId
+  );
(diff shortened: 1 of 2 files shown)


src/internal/utils/node-belongs.ts
```diff
@@ -15,7 +15,10 @@ export function nodeBelongs(container: Node | null, target: Node | EventTarget |
   if (!(target instanceof Node)) {
     return false;
   }
-  const portal = findUpUntil(target as HTMLElement, node => !!node.dataset.awsuiReferrerId);
+  const portal = findUpUntil(
+    target as HTMLElement,
+    node => node instanceof HTMLElement && !!node.dataset.awsuiReferrerId
```
just-boris, line 20: The fix is to restore this line from the past revision: https://github.com/cloudscape-design/components/commit/b28e7f5e83d747a6a21c91150796eb57cdb1db20#diff-eb750a117dea1d7137bc6b95a880c6685b362f28af8d6202fe481e5cccefba27L16

I assumed this check useless, but apparently it is useful

taheramr (approved):

PR #1503, opened 2023-09-04
feat: Add jest shards to a11y tests

taheramr:
### Description
Currently, the a11y tests takes around 30~35 minutes to be executed and it increases the waiting time for a PR to have a successful workflow action by almost 15 minutes as the highest tests that takes time after the a11y are the integ tests ( 15 mins )

This PR introduces jest shards, by which we divide the a11y into shards and each shard executes a subset of the tests in a separate machine. To create the shards mechanism we'll also have to merge the changes in the dry-run github workflow definition in this [PR](https://github.com/cloudscape-design/.github/pull/57).

A successful workflow running with the yml and code updates: https://github.com/cloudscape-design/components/actions/runs/6108039649/job/16576588232

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


build-tools/tasks/a11y.js
```diff
@@ -4,8 +4,19 @@ const execa = require('execa');
 const glob = require('glob');
 const waitOn = require('wait-on');
 const { task } = require('../utils/gulp-utils.js');
+const { last } = require('lodash');
 
+function extractCurrentRunningShard() {
```
connorlanigan, line 9: This function assumes that the last parameter is always the shard, which will not always be the case. I'd recommend to extract the parameter by name instead, for example something like:
```js
const shardParameter = process.argv.find(arg => arg.startsWith("--shard="))
```

then we can also remove the `console.warn`.

connorlanigan (changes requested):

build-tools/tasks/a11y.js
just-boris: Can we pass the shard definition via env variables? 

Or if you want to have a CLI flag, here is the official recipe: https://github.com/gulpjs/gulp/blob/master/docs/recipes/pass-arguments-from-cli.md (but it is an extra dependency)

taheramr: the `parseArgs` function looks great but how do we feel if we used a polyfill until we update our workflows to use nodejs 18?

just-boris: It is linked on the node.js documentation page, so it is sort of an official recommendation

connorlanigan: The function seems to also be available in Node 16 (which we use in our dry-run):
https://nodejs.org/docs/latest-v16.x/api/util.html#utilparseargsconfig
So I think we won't need the polyfill.

just-boris: Nice!

connorlanigan (approved):

build-tools/tasks/a11y.js
taheramr: I used the `parseArgs` but I passed a `false` to the strict mode in its properties, reasons for that:

If the provided arguments in the `process.argv` does not have a matching key in the provided `options` it will throw an error, one parameter that gets passed is the `test:a11y`, I can also put in the options to overcome the error but using the `strict: false` is just more straightforward instead of placing an unused option.

just-boris: ok, seems appropriate for this use-case

just-boris (approved):

PR #1505, opened 2023-09-04
chore: Add vr class from window symbol value

YueyingLu:
### Description

When window[Symbol.for('awsui-visual-refresh-flag')] returns true, renders page in VR.
GlobalStyles: 100722878
Updating Dashboard...

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/app/index.tsx
+++ b/pages/app/index.tsx
@@ -17,8 +17,9 @@ import Header from './components/header';
 import StrictModeWrapper from './components/strict-mode-wrapper';
 import AppContext, { AppContextProvider, parseQuery } from './app-context';
 
+const awsuiVisualRefreshFlag = Symbol.for('awsui-visual-refresh-flag');
 interface ExtendedWindow extends Window {
-  [key: symbol]: (() => boolean) | undefined;
+  [awsuiVisualRefreshFlag]?: () => boolean;
 }
 declare const window: ExtendedWindow;
 
@@ -77,15 +78,9 @@ function App() {
 
 const history = createHashHistory();
 const { visualRefresh } = parseQuery(history.location.search);
-const { windowVR } = parseQuery(history.location.search);
 
 // The VR class needs to be set before any React rendering occurs.
-if (windowVR !== true && windowVR !== false) {
-  window[Symbol.for('isVisualRefresh')] = undefined;
-  document.body.classList.toggle('awsui-visual-refresh', visualRefresh);
-} else {
-  window[Symbol.for('isVisualRefresh')] = () => windowVR;
-}
+window[awsuiVisualRefreshFlag] = () => visualRefresh;
 
 render(
   <HashRouter>
--- a/pages/app/index.tsx
+++ b/pages/app/index.tsx
@@ -17,6 +17,11 @@ import Header from './components/header';
 import StrictModeWrapper from './components/strict-mode-wrapper';
 import AppContext, { AppContextProvider, parseQuery } from './app-context';
 
+interface ExtendedWindow extends Window {
+  [key: symbol]: (() => boolean) | undefined;
+}
+declare const window: ExtendedWindow;
+
 function isAppLayoutPage(pageId?: string) {
   const appLayoutPages = ['app-layout', 'content-layout', 'grid-navigation-custom'];
   return pageId !== undefined && appLayoutPages.some(match => pageId.includes(match));
@@ -72,9 +77,15 @@ function App() {
 
 const history = createHashHistory();
 const { visualRefresh } = parseQuery(history.location.search);
+const { windowVR } = parseQuery(history.location.search);
 
 // The VR class needs to be set before any React rendering occurs.
-document.body.classList.toggle('awsui-visual-refresh', visualRefresh);
+if (windowVR !== true && windowVR !== false) {
+  window[Symbol.for('isVisualRefresh')] = undefined;
+  document.body.classList.toggle('awsui-visual-refresh', visualRefresh);
+} else {
+  window[Symbol.for('isVisualRefresh')] = () => windowVR;
+}
 
 render(
   <HashRouter>
--- a/src/app-layout/__integ__/awsui-applayout.test.ts
+++ b/src/app-layout/__integ__/awsui-applayout.test.ts
@@ -5,6 +5,7 @@ import { BasePageObject } from '@cloudscape-design/browser-test-tools/page-objec
 import createWrapper from '../../../lib/components/test-utils/selectors';
 import mobileStyles from '../../../lib/components/app-layout/mobile-toolbar/styles.selectors.js';
 import { viewports } from './constants';
+import styles from '../../../lib/components/app-layout/visual-refresh/styles.selectors.js';
 
 const wrapper = createWrapper().findAppLayout();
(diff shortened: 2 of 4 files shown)


src/internal/hooks/use-visual-mode/index.ts
```diff
@@ -21,6 +26,9 @@ function detectVisualRefresh() {
 
 export function useVisualRefreshDynamic() {
   if (visualRefreshState === undefined) {
+    if (typeof window !== 'undefined' && window[Symbol.for('isVisualRefresh')]?.()) {
```
just-boris, line 29: Symbol names are global for all objects on the page, so we need to have a scoped name. Following the convention in [our runtime API](https://github.com/cloudscape-design/components/blob/df04533aa56157985792e1ceb8aaf965a77a193d/src/internal/plugins/api.ts#L6C20-L6C48), I would recommend `Symbol.for('awsui-visual-refresh-flag')`.

This is a moment where the longer name, the better

src/internal/hooks/use-visual-mode/index.ts
```diff
@@ -5,6 +5,11 @@ import { ALWAYS_VISUAL_REFRESH } from '../../environment';
 import { isDevelopment } from '../../is-development';
 import { warnOnce } from '@cloudscape-design/component-toolkit/internal';
 
+interface ExtendedWindow extends Window {
+  [key: symbol]: (() => boolean) | undefined;
```
just-boris, line 9: I think you can get more type safety by declaring the symbol via

```js
const awsuiVisualRefreshFlag = Symbol.for('awsui-visual-refresh-flag');
```

Check [this code demo](https://www.typescriptlang.org/play?ssl=3&ssc=1&pln=3&pc=72#code/KYDwDg9gTgLgBAbwL4G4BQaDGEB2BneAQwHc8BXASwDULzCAbAJWADMpg8ALAMXsIHM4AXjgBlAJ4BbAEYR6AOhbQAFAHIS5CgFoAbrTIMt7Nh05aWffqoCU6NBRwxgUFoUzA4AURBOcAE2A-AHUHPwhiOFBfPzw4EP9wxDQ4OABtDUoaOiZWdi5eAQBdAH4ALjhla2EAPjhZOWBCHHQkDDQAzD52OGx8eGJQ8PLvaMD4sOI7NAB6abhiaABrPCxcPAb5egh+ZQGE4nTSTP0GZhN8yxL5StsMWbgIGE5nODwpWXpYsLgcR-ml1b4DZbHZ7CapCQyOSKFSqdaSYBPBz8LTAT7AGyFaxAA) for details

src/internal/hooks/use-visual-mode/index.ts
```diff
@@ -21,6 +26,9 @@ function detectVisualRefresh() {
 
 export function useVisualRefreshDynamic() {
   if (visualRefreshState === undefined) {
+    if (typeof window !== 'undefined' && window[Symbol.for('isVisualRefresh')]?.()) {
+      document.body.classList.add('awsui-visual-refresh');
+    }
     visualRefreshState = detectVisualRefresh();
```
just-boris, line 32: Can we change the order of these checks and only read the symbol value if class name is not found?

This is needed for backward compatibility and also to prevent multiple access to the symbol when multiple instances load on the page. Only the first one will read the symbol, all following get it from the DOM, as before

src/internal/hooks/use-visual-mode/__tests__/use-visual-mode.test.tsx
```diff
@@ -38,4 +44,25 @@ describe('useVisualRefresh', () => {
     expect(console.warn).toHaveBeenCalledWith(expect.stringMatching(/Dynamic visual refresh change detected/));
     expect(screen.getByTestId('current-mode')).toHaveTextContent('false');
   });
+
+  describe('Window Symbol isVisualRefresh ', () => {
+    afterEach(() => {
+      window[Symbol.for('isVisualRefresh')] = undefined;
+    });
+
+    test('should return true when Window Symbol isVisualRefresh is true', () => {
```
just-boris, line 53: Should we also add a test when the symbol returns `false` value?

src/internal/hooks/use-visual-mode/__tests__/use-visual-mode.test.tsx
```diff
@@ -4,6 +4,12 @@ import React from 'react';
 import { useVisualRefresh, clearVisualRefreshState } from '../index';
 import { render, screen } from '@testing-library/react';
 
+declare global {
```
just-boris, line 7: I am guessing it is safe, because it is test-only code, but can we use the same `ExtendedWindow` approach here as well, for consistency?

src/app-layout/__integ__/awsui-applayout.test.ts
```diff
@@ -32,13 +33,21 @@ class AppLayoutPage extends BasePageObject {
...
+    windowVR = false,
+  },
   testFn: (page: AppLayoutPage) => Promise<void>
 ) {
   return useBrowser(async browser => {
     const page = new AppLayoutPage(browser);
     await page.setWindowSize(viewport);
-    await browser.url(`#/light/app-layout/${pageName}?visualRefresh=${visualRefresh}`);
+    await browser.url(
+      `#/light/app-layout/${pageName}?visualRefresh=${visualRefresh}${windowVR ? '&windowVR=true' : ''}`
```
just-boris, line 49: When we have too many query parameters I suggest using `URLSearchParams` to build the query string.  Example in our code: https://github.com/cloudscape-design/components/blob/df04533aa56157985792e1ceb8aaf965a77a193d/src/internal/components/dropdown/__integ__/dropdown-focus-trap.test.ts#L15-L19

pages/app/index.tsx
```diff
@@ -72,9 +77,15 @@ function App() {
 
 const history = createHashHistory();
 const { visualRefresh } = parseQuery(history.location.search);
+const { windowVR } = parseQuery(history.location.search);
```
just-boris, line 80: Why do we need a new flag? Can we change the logic of the existing query parameter instead?

src/internal/hooks/use-visual-mode/__tests__/use-visual-mode.test.tsx
just-boris: Also, what about the test when `awsui-visual-refresh` class already exists, but symbol is not defined or returns false?

pages/app/index.tsx
```diff
@@ -72,9 +77,15 @@ function App() {
...
 
 // The VR class needs to be set before any React rendering occurs.
-document.body.classList.toggle('awsui-visual-refresh', visualRefresh);
+if (windowVR !== true && windowVR !== false) {
+  window[Symbol.for('isVisualRefresh')] = undefined;
+  document.body.classList.toggle('awsui-visual-refresh', visualRefresh);
+} else {
+  window[Symbol.for('isVisualRefresh')] = () => windowVR;
+}
```
just-boris, line 88:
```suggestion
window[Symbol.for('isVisualRefresh')] = () => visualRefresh
```

just-boris (approved):
👍

PR #1508, opened 2023-09-05
fix: Cleanup sticky columns listeners when table cells unmount

pan-kot:
### Description

The missing subscription cleanup function caused a memory leak.

ref: AWSUI-22191

### How has this been tested?

New unit test added.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/table/body-cell/td-element.tsx
+++ b/src/table/body-cell/td-element.tsx
@@ -6,6 +6,7 @@ import styles from './styles.css.js';
 import { getStickyClassNames } from '../utils';
 import { StickyColumnsModel, useStickyCellStyles } from '../sticky-columns';
 import { TableRole, getTableCellRoleProps } from '../table-role';
+import { useMergeRefs } from '../../internal/hooks/use-merge-refs/index.js';
 
 export interface TableTdElementProps {
   className?: string;
@@ -74,6 +75,9 @@ export const TableTdElement = React.forwardRef<HTMLTableCellElement, TableTdElem
       columnId,
       getClassName: props => getStickyClassNames(styles, props),
     });
+
+    const mergedRef = useMergeRefs(stickyStyles.ref, ref);
+
     return (
       <Element
         style={{ ...style, ...stickyStyles.style }}
@@ -96,12 +100,7 @@ export const TableTdElement = React.forwardRef<HTMLTableCellElement, TableTdElem
         onClick={onClick}
         onMouseEnter={onMouseEnter}
         onMouseLeave={onMouseLeave}
-        ref={node => {
-          stickyStyles.ref(node);
-          if (node && ref) {
-            (ref as React.MutableRefObject<HTMLTableCellElement>).current = node;
-          }
-        }}
+        ref={mergedRef}
         {...nativeAttributes}
       >
         {children}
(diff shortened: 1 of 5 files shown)


src/table/sticky-columns/__tests__/use-sticky-columns.test.tsx
```diff
@@ -241,6 +241,38 @@ test('performs styles cleanup', () => {
   expect(elements.cells[0]).not.toHaveClass('sticky-cell');
 });
 
+test('cell subscriptions are cleaned up on ref change and un-mount', () => {
+  const unsubscribe = jest.fn();
+  const subscribe = jest.fn().mockImplementation(() => unsubscribe);
```
connorlanigan, line 246: just fyi, you can pass the mock implementation directly as a parameter:
```js
jest.fn(() => unsubscribe)
```

src/table/sticky-columns/use-sticky-columns.ts
```diff
@@ -201,6 +201,11 @@ export function useStickyCellStyles({
     [columnId, setCell, stickyColumns.store]
   );
 
+  // Subscriptions created in the refCallback must be cancelled when the component un-mounts.
+  useEffect(() => {
+    return () => unsubscribeRef.current?.();
```
connorlanigan, line 206: Looks good, but I don't quite understand yet why we need to do this. Shouldn't react call the refCallback with `null` when the component unmounts, and thus already cause the subscription to be cancelled in line 162?

pan-kot: Great question! That should indeed be the case.

pan-kot: Updated!

src/table/body-cell/td-element.tsx
```diff
@@ -97,11 +97,9 @@ export const TableTdElement = React.forwardRef<HTMLTableCellElement, TableTdElem
         onMouseEnter={onMouseEnter}
         onMouseLeave={onMouseLeave}
         ref={node => {
-          if (node) {
-            stickyStyles.ref(node);
-            if (ref) {
-              (ref as React.MutableRefObject<HTMLTableCellElement>).current = node;
-            }
+          stickyStyles.ref(node);
+          if (node && ref) {
```
connorlanigan, line 101: Can we remove the `ref` condition for this other ref as well? This way we will match the usual React expectations of refs being set to `null` when a component is unmounted
i.e.:
```
ref={node => {
  stickyStyles.ref(node);
  if (ref) {
    (ref as React.MutableRefObject<HTMLTableCellElement>).current = node;
  }
}

File: src/table/sticky-columns/__tests__/use-sticky-columns.test.tsx [hunk 4]
```diff
@@ -267,7 +265,7 @@ test('cell subscriptions are cleaned up on ref change and un-mount', () => {
   expect(subscribe).toHaveBeenCalledTimes(2);
   expect(unsubscribe).toHaveBeenCalledTimes(1);
 
-  unmount();
+  result.current.ref(null);
```
connorlanigan, line 268: Does this test still work if we use `unmount`? Internally the `unmount` call would then still call `result.current.ref(null)`, but it expresses a bit clearer what we want to simulate in this line than just calling our ref function directly

connorlanigan (changes requested):

connorlanigan (approved):

PR #1509, opened 2023-09-05
chore: Emit `funnelStepChange` events

connorlanigan:
### Description

This PR emits the `funnelStepChange` event whenever a funnel step is modified, i.e. when the substeps of the step change.

The events have been added in https://github.com/cloudscape-design/components/pull/1507.

### How has this been tested?

Added unit tests and tested manually in new dev pages.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/analytics/components/analytics-funnel.tsx
+++ b/src/internal/analytics/components/analytics-funnel.tsx
@@ -17,8 +17,8 @@ import { useVisualRefresh } from '../../hooks/use-visual-mode';
 
 import { PACKAGE_VERSION } from '../../environment';
 
-import { FunnelMetrics } from '../';
-import { FunnelProps, FunnelStepProps } from '../interfaces';
+import { FunnelMetrics } from '../index';
+import { FunnelProps, FunnelStepProps, SubStepConfiguration } from '../interfaces';
 
 import {
   DATA_ATTR_FUNNEL_STEP,
@@ -28,6 +28,7 @@ import {
   getSubStepNameSelector,
   getSubStepSelector,
 } from '../selectors';
+import { useDebounceCallback } from '../../hooks/use-debounce-callback';
 
 export const FUNNEL_VERSION = '1.0';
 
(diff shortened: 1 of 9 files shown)


src/internal/analytics/components/analytics-funnel.tsx
```diff
@@ -321,11 +382,15 @@ export const AnalyticsFunnelSubStep = ({ children }: AnalyticsFunnelSubStepProps
   useEffect(() => {
     if (!isNested) {
       subStepCount.current++;
+      onStepChange();
 
-      // eslint-disable-next-line react-hooks/exhaustive-deps
-      return () => void subStepCount.current--;
+      return () => {
+        // eslint-disable-next-line react-hooks/exhaustive-deps
```
avinashbot, line 388: Why? Is the ref not stable?

src/internal/analytics/components/analytics-funnel.tsx
```diff
@@ -200,6 +201,48 @@ export const AnalyticsFunnelStep = (props: AnalyticsFunnelStepProps) => {
...
+
+  /* We debounce this handler, so that multiple containers can change at once without causing 
+  too many events. */
+  const stepChangeCallback = useDebounceCallback(() => {
+    // We don't want to emit the event after the component has been unmounted.
+    if (!listenForSubStepChanges.current) {
+      return;
+    }
+
+    const subSteps = Array.from(document.querySelectorAll(getSubStepAllSelector())) as HTMLElement[];
```
avinashbot, line 230: Minor: Big `as` hater here, you can do `document.querySelectorAll<HTMLElement>(getSubStepAllSelector())`

avinashbot (approved):
Just minor stuff, if you feel like fixing in this PR or the next.

src/internal/analytics/components/analytics-funnel.tsx
connorlanigan: The ref is stable, but its `current` value is not. Although this is the whole point of using a ref, ESLint produces the following warning if you access the `current` property of a ref in the effect cleanup function:
> The ref value 'subStepCount.current' will likely have changed by the time this effect cleanup function runs. If this ref points to a node rendered by React, copy 'subStepCount.current' to a variable inside the effect, and use that variable in the cleanup function.

I find it a bit misleading that this is produced by the rule `react-hooks/exhaustive-deps`, but that is why we have to disable that rule for that line.

src/internal/analytics/components/analytics-funnel.tsx
connorlanigan: Ah that is very good, I was not aware that I can specify that as a generic parameter. Thank you! I'll cover that in the next PR.

PR #1511, opened 2023-09-05
chore: Emit `funnelChange` events

connorlanigan:
### Description

This PR emits the funnelChange event whenever a funnel is modified, i.e. when the steps of the funnel change.

The events have been added in https://github.com/cloudscape-design/components/pull/1507.

See also https://github.com/cloudscape-design/components/pull/1509 for the related funnel**Step**Change event, which handles changes in substeps rather than in steps.

### How has this been tested?
Manually tested in updated dev page, and added unit tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/analytics/components/analytics-funnel.tsx
+++ b/src/internal/analytics/components/analytics-funnel.tsx
@@ -30,7 +30,7 @@ import {
 } from '../selectors';
 import { useDebounceCallback } from '../../hooks/use-debounce-callback';
 
-export const FUNNEL_VERSION = '1.0';
+export const FUNNEL_VERSION = '1.1';
 
 type AnalyticsFunnelProps = { children?: React.ReactNode } & Pick<
   FunnelProps,
--- a/src/internal/analytics/components/analytics-funnel.tsx
+++ b/src/internal/analytics/components/analytics-funnel.tsx
@@ -227,7 +227,7 @@ function useStepChangeListener(handler: (stepConfiguration: SubStepConfiguration
       return;
     }
 
-    const subSteps = Array.from(document.querySelectorAll(getSubStepAllSelector())) as HTMLElement[];
+    const subSteps = Array.from(document.querySelectorAll<HTMLElement>(getSubStepAllSelector()));
 
     const subStepConfiguration = subSteps.map((substep, index) => {
       const name = substep.querySelector(getSubStepNameSelector())?.textContent ?? '';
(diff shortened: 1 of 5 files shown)


src/wizard/internal.tsx
```diff
@@ -183,3 +185,43 @@ export default function InternalWizard({
     </div>
   );
 }
+
+/**
+ * This hook observes the step configuration and events the `funnelChange` event when the steps change.
```
timogasda, line 190: *emits the `funnelChange` event?

src/wizard/internal.tsx
```diff
@@ -183,3 +185,43 @@ export default function InternalWizard({
...
+ */
+function useFunnelChangeEvent(funnelInteractionId: string | undefined, steps: WizardProps['steps']) {
+  const listenForStepChanges = useRef(false);
+
+  useEffect(() => {
+    // We prevent emitting the event before the funnel has stabilised.
+    const handle = setTimeout(() => (listenForStepChanges.current = true), 50);
```
timogasda, line 197: What does "stabilised" mean? And how certain are we that 50 is the right magic number here?

connorlanigan: We want to give some time for updates that happen directly after rendering, for example if some part of the step configuration is loaded asynchronously by the customer or for our effects to detect nested funnels after the first render — changes that are technically not part of the initial appearance, but from the user perspective happen invisibly because they're so fast. The number 50 is somewhat arbitrary, it is chosen so that it is shorter than any reasonable human interaction that we would legitimately count as a change to the configuration, and long enough that React can render at least once and e.g. run any asynchronous effects.

src/wizard/internal.tsx
connorlanigan: Good catch!

src/internal/analytics/components/analytics-funnel.tsx
```diff
@@ -30,7 +30,7 @@ import {
 } from '../selectors';
 import { useDebounceCallback } from '../../hooks/use-debounce-callback';
 
-export const FUNNEL_VERSION = '1.0';
+export const FUNNEL_VERSION = '1.1';
```
connorlanigan, line 33: The new funnel version indicates that consumers of these events can rely on the existence of `funnelChange` and `funnelStepChange` events to determine whether a funnel is dynamic (i.e. steps change) or static.

PR #1519, opened 2023-09-06
chore: Emit event when funnelComplete or funnelCancelled

YueyingLu:
### Description

Add 'cloudscape-creation-flow-done' event when funnelComplete or funnelCancelled. 

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/analytics/__tests__/components/analytics-funnel.test.tsx
+++ b/src/internal/analytics/__tests__/components/analytics-funnel.test.tsx
@@ -9,6 +9,7 @@ import {
   AnalyticsFunnel,
   AnalyticsFunnelStep,
   AnalyticsFunnelSubStep,
+  CREATION_EDIT_FLOW_DONE_EVENT_NAME,
 } from '../../../../../lib/components/internal/analytics/components/analytics-funnel';
 import { useFunnel, useFunnelSubStep } from '../../../../../lib/components/internal/analytics/hooks/use-funnel';
 import Button from '../../../../../lib/components/button';
@@ -663,6 +664,47 @@ describe('AnalyticsFunnelSubStep', () => {
       subStepSelector: expect.any(String),
       subStepNameSelector: expect.any(String),
     });
+  });
+});
+
+describe('emit cloudscape-creation-flow-done event', () => {
+  const doneSpy = jest.fn();
+
+  beforeEach(() => {
+    document.addEventListener(CREATION_EDIT_FLOW_DONE_EVENT_NAME, doneSpy);
+    jest.clearAllMocks();
+    jest.useFakeTimers();
+    mockFunnelMetrics();
+  });
+
+  afterEach(() => {
+    document.removeEventListener(CREATION_EDIT_FLOW_DONE_EVENT_NAME, doneSpy);
+  });
+
+  test('emit cloudscape-creation-flow-done event when funnelComplete', async () => {
+    const ChildComponent = () => {
+      const { funnelSubmit } = useFunnel();
+      return <button onClick={funnelSubmit}>Submit</button>;
+    };
+    const { getByText } = render(
+      <AnalyticsFunnel funnelType="single-page" optionalStepNumbers={[]} totalFunnelSteps={3}>
+        <ChildComponent />
+      </AnalyticsFunnel>
+    );
+    act(() => void jest.runAllTimers());
+    fireEvent.click(getByText('Submit'));
+    await waitFor(() => {
+      expect(doneSpy).toHaveBeenCalled();
+    });
+  });
+
+  test('emit cloudscape-creation-flow-done event when funnelCancelled', () => {
+    const { unmount } = render(
+      <AnalyticsFunnel funnelType="single-page" optionalStepNumbers={[]} totalFunnelSteps={1} />
+    );
+    act(() => void jest.runAllTimers());
+    unmount();
+    expect(doneSpy).toHaveBeenCalled();
   });
 });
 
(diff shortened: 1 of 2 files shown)


src/internal/analytics/__tests__/components/analytics-funnel.test.tsx
```diff
@@ -641,6 +641,48 @@ describe('AnalyticsFunnelSubStep', () => {
...
+describe('emit cloudscape-creation-flow-done event', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+    jest.useFakeTimers();
+    mockFunnelMetrics();
+  });
+
+  test('emit cloudscape-creation-flow-done event when funnelComplete', async () => {
+    let eventIsCalled = false;
+    document.addEventListener('cloudscape-creation-flow-done', () => {
```
timogasda, line 653: We should probably clean this up so it doesn't leak into other tests. Maybe move this up into the `describe` block (since the other test case uses it too), and then call `removeEventListener` when cleaning up

src/internal/analytics/__tests__/components/analytics-funnel.test.tsx
```diff
@@ -641,6 +641,48 @@ describe('AnalyticsFunnelSubStep', () => {
...
+    };
+    const { getByText } = render(
+      <AnalyticsFunnel funnelType="single-page" optionalStepNumbers={[]} totalFunnelSteps={3}>
+        <ChildComponent />
+      </AnalyticsFunnel>
+    );
+    act(() => void jest.runAllTimers());
+    fireEvent.click(getByText('Submit'));
+    await waitFor(() => {
+      expect(eventIsCalled).toBe(true);
```
timogasda, line 668: Adding to my previous comment, you can probably turn the event listener into a jest function and then do `expect(...).toHaveBeenCalled` here. Seems a bit more like standard jest testing

src/internal/analytics/components/analytics-funnel.tsx
```diff
@@ -52,6 +52,16 @@ export const AnalyticsFunnel = (props: AnalyticsFunnelProps) => {
...
 };
 
+const funnelCancelled = ({ funnelInteractionId }: { funnelInteractionId: string }) => {
+  FunnelMetrics.funnelCancelled({ funnelInteractionId });
+  document.dispatchEvent(new Event('cloudscape-creation-flow-done'));
+};
+
+const funnelComplete = ({ funnelInteractionId }: { funnelInteractionId: string }) => {
+  FunnelMetrics.funnelComplete({ funnelInteractionId });
+  document.dispatchEvent(new Event('cloudscape-creation-flow-done'));
```
johannes-weber, line 62: I recommend setting the event name in a const instead. This way, we'll only have to change one place if the constant value needs to be changed in the future.

src/internal/analytics/components/analytics-funnel.tsx
```diff
@@ -52,6 +52,16 @@ export const AnalyticsFunnel = (props: AnalyticsFunnelProps) => {
   return <InnerAnalyticsFunnel {...props} />;
 };
 
+const funnelCancelled = ({ funnelInteractionId }: { funnelInteractionId: string }) => {
```
johannes-weber, line 55: To me `funnelCancelled` and `funnelComplete` sound generic and requires reading thought the code to see what it does. What about a more specific name, e.g. `dispatchFunnelComplete` or `dispatchFunnelCompleteEvent`?

YueyingLu: I guess the `funnelCancelled` `funnelComplete` sounds generic and misleading to 'to cancel/complete the funnel'. Rename to `onFunnelComplete` and `onFunnelCancelled`, because the function is not only dispatching the event but also run FunnelMetrics function.

johannes-weber (approved):

src/internal/analytics/__tests__/components/analytics-funnel.test.tsx
```diff
@@ -666,6 +667,53 @@ describe('AnalyticsFunnelSubStep', () => {
...
+    const { getByText } = render(
+      <AnalyticsFunnel funnelType="single-page" optionalStepNumbers={[]} totalFunnelSteps={3}>
+        <ChildComponent />
+      </AnalyticsFunnel>
+    );
+    act(() => void jest.runAllTimers());
+    fireEvent.click(getByText('Submit'));
+    await waitFor(() => {
+      expect(onCompleteSpy).toHaveBeenCalled();
+      removeCreateFlowDoneEvent(onCompleteSpy);
```
timogasda, line 700: This cleanup does not get executed when there's a problem with the test code itself or when the test fails. That's what `beforeEach` and `afterEach` are for. A more idiomatic approach would look something like this:

```
describe('emit cloudscape-creation-flow-done event', () => {
  const doneSpy = jest.fn();

  beforeEach(() => {
    document.addEventListener(CREATION_FLOW_DONE_EVENT_NAME, doneSpy);
    // ...
  });
  afterEach(() => {
    document.removeEventListener(CREATION_FLOW_DONE_EVENT_NAME, doneSpy);
  });
});
```

src/internal/analytics/components/analytics-funnel.tsx
```diff
@@ -51,6 +51,17 @@ export const AnalyticsFunnel = (props: AnalyticsFunnelProps) => {
 
   return <InnerAnalyticsFunnel {...props} />;
 };
+export const CREATION_FLOW_DONE_EVENT_NAME = 'cloudscape-creation-flow-done';
```
connorlanigan, line 54: A general note about the naming: the funnel could also be an edit flow rather than a creation flow. As long as we keep that in mind when listening for the events, I'm fine with this naming.

timogasda (approved):

PR #1523, opened 2023-09-07
chore: Cleanup ESLint warnings in CI environment

connorlanigan:
### Description

In a typical PR, the linting step produces a large amount of warnings with these two messages:
```
warning  Prop "className" is forbidden on Components  react/forbid-component-props
```
```
warning  Prop "id" is forbidden on Components  react/forbid-component-props
```

Example log: [cloudscape-design/components/.../job/16574616087](https://github.com/cloudscape-design/components/actions/runs/6107496245/job/16574616087?pr=1517#step:7:1)

These warnings currently make up 479 of 503 total warnings (95.2%), and thus reduce the visibility of other (potentially more important) warnings immensely. This has already lead to a "warning fatigue" in the team. Since there are still many use cases for these properties, they cannot just all be removed from the code.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,6 +1,8 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+const isCI = process.env.CI;
+
 module.exports = {
   parser: '@typescript-eslint/parser',
   extends: [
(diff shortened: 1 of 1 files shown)


just-boris:
Should we just turn it off, it is not useful? Or scope down to "pages/" directory only, for example?

just-boris:
Every instance of `process.env.CI` is a possible hours spent on debugging why the build passes locally and fails in PR. Should be used very carefully

just-boris:
How about annotating all current usage with eslint-ignore, if we want to keep the rule?

connorlanigan:
That is an option, but, assuming that we want to review them rather than just annotate all of them, it will be a longer process (since there are 479 instances). I think in the meantime these warnings serve no purpose, so I would not want to delay making the warning output useful again until all these instances have been reviewed and updated or annotated.

just-boris:
Still not convinced for CI-only conditions here. You can try discussing it with the other devs. If the majority is fine with that, I will accept

connorlanigan:
Superseded by https://github.com/cloudscape-design/components/pull/1529.

PR #1528, opened 2023-09-08
fix: Resizable columns text content to not include column width

pan-kot:
### Description

When asserting column headers text content is must not include anything but the header text.

Ref: AWSUI-22297

### How has this been tested?

Added a new unit test to secure the property.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/table/resizer/index.tsx
+++ b/src/table/resizer/index.tsx
@@ -9,7 +9,6 @@ import styles from './styles.css.js';
 import { KeyCode } from '../../internal/keycode';
 import { DEFAULT_COLUMN_WIDTH } from '../use-column-widths';
 import { useStableCallback } from '@cloudscape-design/component-toolkit/internal';
-import ScreenreaderOnly from '../../internal/components/screenreader-only';
 import { useUniqueId } from '../../internal/hooks/use-unique-id';
 import { joinStrings } from '../../internal/utils/strings';
 import Portal from '../../internal/components/portal';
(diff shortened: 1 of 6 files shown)


src/table/resizer/index.tsx
```diff
@@ -238,7 +240,9 @@ export function Resizer({
         tabIndex={tabIndex}
         data-focus-id={focusId}
       />
-      <ScreenreaderOnly id={resizerWidthId}>{headerCellWidthString}</ScreenreaderOnly>
+      <Portal container={getDescriptionRoot?.()}>
```
just-boris, line 243: Is Lighthouse happy about rendering an extra portal for every table column?

Why do you need that much complexity here? Can you think if a simpler solution?

pan-kot: Thanks, I will run the tests to make sure.

I haven't found a much simpler solution so far. The portal can be made simpler if removing the container but that makes the a11y tools unhappy because of the extra nodes added that are not wrapped in any landmarks.

Another idea was to only add the description when the resize handle is in focus. However, that has still the potential to affect the tests.

just-boris: I need to look into the labelling logic overall, because I am not sure you implemented it correctly in the first place: https://github.com/cloudscape-design/components/pull/1492

This may need revisiting

just-boris: But Lighthouse test run need to happen first.

pan-kot: The lighthouse tests are green.

just-boris (approved):
This implementation will clearly backfire us more times, but if Lighthouse tests are green, I have no formal reasons to push back on this change

PR #1537, opened 2023-09-11
fix: Use recoveryText from i18n provider in Select

Al-Dani:
### Description

We missed couple of properties when implementing i18n 

Related links, issue #, if available: AWSUI-22320

### How has this been tested?

New `i18n` unit tests added

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/autosuggest/recovery-test.page.tsx
+++ b/pages/autosuggest/recovery-test.page.tsx
@@ -29,6 +29,7 @@ export default function () {
             enteredTextLabel={enteredTextLabel}
             expandToViewport={true}
             recoveryText="Try again"
+            onLoadItems={() => {}}
           />
         </SpaceBetween>
       </ScreenshotArea>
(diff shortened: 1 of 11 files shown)


abdhalees (approved):

just-boris:
Should there be any tests for it?

Al-Dani:
> Should there be any tests for it?

I thought about it, but not sure how these tests would look like. What we can do is for example remove from `pages` all properties that are coming from i18n and than rely on our a11y and screenshot tests. But this is also kinda manual and therefore pruned to errors

just-boris:
I think, there should be a general test suite for each component that checks that expected built-in strings are rendered.

Please discuss with other team members. It is not good to leave this behavior untested.

Al-Dani:
Created separate `i18n` unit tests for the components in PR. Apparently, we missed more than it was initially reported. So we will gradually add similar tests for other components to make sure we covered all messages.

abdhalees (approved):

PR #1541, opened 2023-09-12
chore: Include stepConfiguration and subStepConfiguration properties in start events

connorlanigan:
### Description

Previously, we've included the `stepConfiguration` and `subStepConfiguration` properties in the `funnelChange` and `funnelStepChange` events (https://github.com/cloudscape-design/components/pull/1509, https://github.com/cloudscape-design/components/pull/1511).

With this PR, we also include these properties in the corresponding start event, so that they can be used as a reference when the change events are emitted (otherwise it is difficult to figure out what changed).

### How has this been tested?
Extended unit tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/analytics/__tests__/components/analytics-funnel.test.tsx
+++ b/src/internal/analytics/__tests__/components/analytics-funnel.test.tsx
@@ -332,10 +332,29 @@ describe('AnalyticsFunnelStep', () => {
     expect(FunnelMetrics.funnelStepStart).toHaveBeenCalledTimes(1);
     expect(FunnelMetrics.funnelStepStart).toHaveBeenCalledWith({
       funnelInteractionId: mockedFunnelInteractionId,
+      stepName: undefined,
       stepNumber,
       stepNameSelector,
       subStepAllSelector: expect.any(String),
       totalSubSteps: 4,
+      subStepConfiguration: [
+        {
+          name: '',
+          number: 1,
+        },
+        {
+          name: '',
+          number: 2,
+        },
+        {
+          name: '',
+          number: 3,
+        },
+        {
+          name: '',
+          number: 4,
+        },
+      ],
     });
   });
 
--- a/src/internal/analytics/__tests__/components/analytics-funnel.test.tsx
+++ b/src/internal/analytics/__tests__/components/analytics-funnel.test.tsx
@@ -18,7 +18,7 @@ import Cards from '../../../../../lib/components/cards';
 import Table from '../../../../../lib/components/table';
 import ExpandableSection from '../../../../../lib/components/expandable-section';
 
-import { mockedFunnelInteractionId, mockFunnelMetrics } from '../mocks';
+import { mockedFunnelInteractionId, mockFunnelMetrics, mockInnerText } from '../mocks';
 
 describe('AnalyticsFunnel', () => {
   beforeEach(() => {
@@ -300,6 +300,8 @@ describe('AnalyticsFunnelStep', () => {
     mockFunnelMetrics();
   });
 
+  mockInnerText();
+
   test('renders children components', () => {
     const { getByText } = render(
       <AnalyticsFunnelStep stepNumber={1} stepNameSelector=".step-name-selector">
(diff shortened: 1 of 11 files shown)


src/wizard/__tests__/wizard.test.tsx
```diff
@@ -407,7 +407,7 @@ test('raises a warning when setting activeStepIndex without onNavigate listener'
 test('does not perform navigation when used in controlled mode', () => {
   const [wrapper] = renderDefaultWizard({ activeStepIndex: 1, onNavigate: () => {} });
   const checkStep = () => {
-    expect(wrapper.findContent()!.getElement()).toHaveTextContent(DEFAULT_STEPS[1].content as string);
+    expect(wrapper.findContent()!.getElement()).toHaveTextContent('Step 2, substep oneStep 2, substep two');
```
pan-kot, line 410: What this change is for? The content of the updated DEFAULT_STEPS seems as still matching the output.

src/internal/analytics/components/analytics-funnel.tsx
```diff
@@ -201,6 +206,19 @@ export const AnalyticsFunnelStep = (props: AnalyticsFunnelStepProps) => {
   return <InnerAnalyticsFunnelStep {...props} key={props.stepNumber} />;
 };
 
+function getSubStepConfiguration() {
+  const subSteps = Array.from(document.querySelectorAll<HTMLElement>(getSubStepAllSelector()));
+
+  const subStepConfiguration = subSteps.map((substep, index) => {
+    const name = substep.querySelector(getSubStepNameSelector())?.textContent ?? '';
```
pan-kot, line 213: Should we include the sub-steps w/o a name at all?

pan-kot: Is there a reason why here we use textContent and in getNameFromSelector we use innerText?

src/internal/analytics/__tests__/components/analytics-funnel.test.tsx
```diff
@@ -332,10 +332,29 @@ describe('AnalyticsFunnelStep', () => {
     expect(FunnelMetrics.funnelStepStart).toHaveBeenCalledTimes(1);
     expect(FunnelMetrics.funnelStepStart).toHaveBeenCalledWith({
       funnelInteractionId: mockedFunnelInteractionId,
+      stepName: undefined,
```
pan-kot, line 335: Why stepName is undefined here? Is there a real difference between this value being an empty string and undefined?

src/wizard/internal.tsx
```diff
@@ -225,3 +219,11 @@ function useFunnelChangeEvent(funnelInteractionId: string | undefined, steps: Wi
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [funnelInteractionId, stepTitles]);
 }
+
+export function getStepConfiguration(steps: WizardProps['steps']) {
```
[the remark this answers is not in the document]
connorlanigan: Good idea

src/wizard/__tests__/wizard.test.tsx
connorlanigan: The DEFAULT_STEPS content is now JSX, which means that we can't directly compare it anymore with the text content that is rendered in the DOM.

src/internal/analytics/components/analytics-funnel.tsx
connorlanigan: Good catch, I don't think there was any particular reason. The advantage of `innerText` is that it doesn't include invisible text. I'll update this to use `innerText`.

Steps without a name should still be included. This only happens for containers that don't have a heading, which we still want to be recognisable in the event data.

src/internal/analytics/__tests__/components/analytics-funnel.test.tsx
connorlanigan: I've updated the code in the next revision to have a proper step name

src/wizard/__tests__/wizard.test.tsx
pan-kot: got it 👍

src/internal/analytics/components/analytics-funnel.tsx
pan-kot: Should we have a better default than an empty string in that case? Like `name.trim() || null`?

pan-kot (approved):

pan-kot (approved):

PR #1543, opened 2023-09-12
fix: Fix controlled tools mode when drawers API is used

just-boris:
### Description

When drawers API is used `toolsOpen` is ignored. This introduced an issue for controlled mode, which is commonly used to integrate with help panel.

### How has this been tested?

Added extra unit and integration tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/utils/use-drawers.ts
+++ b/src/app-layout/utils/use-drawers.ts
@@ -86,7 +86,7 @@ export function useDrawers(
   const [activeDrawerId, setActiveDrawerId] = useControllable(
     ownDrawers?.activeDrawerId,
     ownDrawers?.onChange,
-    !toolsProps.toolsHide && toolsProps.toolsOpen ? TOOLS_DRAWER_ID : undefined,
+    undefined,
     {
       componentName: 'AppLayout',
       controlledProp: 'activeDrawerId',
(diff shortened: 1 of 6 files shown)


src/app-layout/utils/use-drawers.ts
```diff
@@ -105,8 +105,12 @@ export function useDrawers(
   if (toolsDrawer && combinedDrawers.length > 0) {
     combinedDrawers.unshift(toolsDrawer);
   }
-  const activeDrawer = combinedDrawers.find(drawer => drawer.id === activeDrawerId);
-  const activeDrawerIdResolved = activeDrawer?.id; // only defined when corresponding drawer exists
+  // support toolsOpen in runtime-drawers-only mode
+  let activeDrawerIdResolved =
```
taheramr, line 109: non-blocking: 
Can we make the `activeDrawerIdResolved` as a `const` instead? we can reference the actual active drawer id using the `activeDrawer` variable already.

Down below we can just pass the id by returning:

```
return {
activeDrawerId: activeDrawer?,id,
{..rest
}
```

taheramr (approved):
left a non-blocking comment

src/app-layout/utils/use-drawers.ts
just-boris: There is a re-assignment on line 113. I could work around it by introducing intermediate variables

```js
// support toolsOpen
const activeDrawerIdWithTools = toolsProps.toolsOpen && (ownDrawers?.items ?? []).length === 0 ? TOOLS_DRAWER_ID : activeDrawerId
const activeDrawer = combinedDrawers.find(drawer => drawer.id === activeDrawerIdWithTools);
// ensure that the variable is only defined when the respective drawer exist
const activeDrawerIdResolved = activeDrawer?.id;
```

But then the naming gets confusing. I think having a pipeline like is acceptable code

```js
let foo = something();
foo = somethingElse(foo)
foo = anotherOne(foo)
```

If you have anything more to add, I am open for DM discussion

PR #1555, opened 2023-09-14
fix: Constrain dropdown width to viewport

jperals:
### Description

After #1423, dropdowns of the autosuggest, select and multiselect components expand beyond the trigger width, like this:

localhost_8080_ (29)

The problem is, that doesn't account for the available space, so in a narrow viewport or in a narrow container with hidden overflow the dropdown will be cut off:

localhost_8080_ (28)

This is fixed here by opting out of that behavior if it was detected that the the dropdown overflows. So if that would happen, it will _not_ stretch beyond the trigger:

localhost_8080_ (30)

Issue: AWSUI-22328

### How has this been tested?

Added new test page, which I used for both running manual testing and for writing new integration tests.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/components/dropdown/dropdown-fit-handler.ts
+++ b/src/internal/components/dropdown/dropdown-fit-handler.ts
@@ -45,7 +45,7 @@ const getClosestParentDimensions = (element: HTMLElement): any => {
 export const defaultMaxDropdownWidth = getBreakpointValue('xxs');
 
 export const getAvailableSpace = (
-  trigger: HTMLElement,
+  trigger: DOMRect,
   dropdown: HTMLElement,
   overflowParents: ReadonlyArray<Dimensions>,
   stretchWidth = false,
@@ -62,12 +62,12 @@ export const getAvailableSpace = (
     : isMobile
     ? AVAILABLE_SPACE_RESERVE_MOBILE_HORIZONTAL
     : AVAILABLE_SPACE_RESERVE_DEFAULT;
-  const { bottom: triggerBottom, left: triggerLeft, right: triggerRight } = trigger.getBoundingClientRect();
+  const { bottom: triggerBottom, left: triggerLeft, right: triggerRight } = trigger;
 
   return overflowParents.reduce(
     ({ above, below, left, right }, overflowParent) => {
       const offsetTop = triggerBottom - overflowParent.top;
-      const currentAbove = offsetTop - trigger.offsetHeight - availableSpaceReserveVertical;
+      const currentAbove = offsetTop - trigger.height - availableSpaceReserveVertical;
       const currentBelow = overflowParent.height - offsetTop - availableSpaceReserveVertical;
       const currentLeft = triggerRight - overflowParent.left - availableSpaceReserveHorizontal;
       const currentRight = overflowParent.left + overflowParent.width - triggerLeft - availableSpaceReserveHorizontal;
@@ -122,7 +122,7 @@ export const getInteriorAvailableSpace = (
 };
 
 export const getDropdownPosition = ({
-  triggerElement,
+  triggerBox,
   dropdownElement,
   overflowParents,
   minWidth: desiredMinWidth,
@@ -132,7 +132,7 @@ export const getDropdownPosition = ({
   isMobile = false,
   stretchBeyondTriggerWidth = false,
 }: {
-  triggerElement: HTMLElement;
+  triggerBox: DOMRect;
   dropdownElement: HTMLElement;
   overflowParents: ReadonlyArray<Dimensions>;
   minWidth?: number;
@@ -144,7 +144,7 @@ export const getDropdownPosition = ({
 }): DropdownPosition => {
   // Determine the space available around the dropdown that it can grow in
   const availableSpace = getAvailableSpace(
-    triggerElement,
+    triggerBox,
     dropdownElement,
     overflowParents,
     stretchWidth,
@@ -152,7 +152,7 @@ export const getDropdownPosition = ({
     isMobile
   );
   // Determine the width of the trigger
-  const triggerWidth = triggerElement.getBoundingClientRect().width;
+  const triggerWidth = triggerBox.width;
   // Minimum width is determined by either an explicit number (desiredMinWidth) or the trigger width
   const minWidth = desiredMinWidth ? Math.min(triggerWidth, desiredMinWidth) : triggerWidth;
   // If stretchBeyondTriggerWidth is true, the maximum width is the XS breakpoint (465px) or the trigger width (if bigger).
@@ -273,10 +273,11 @@ export const calculatePosition = (
   dropdownElement.classList.remove(styles['dropdown-drop-up']);
 
   const overflowParents = getOverflowParentDimensions(dropdownElement, interior, expandToViewport, stretchHeight);
+  const triggerBox = triggerElement.getBoundingClientRect();
   const position = interior
     ? getInteriorDropdownPosition(triggerElement, dropdownElement, overflowParents, isMobile)
     : getDropdownPosition({
-        triggerElement,
+        triggerBox,
         dropdownElement,
         overflowParents,
         minWidth,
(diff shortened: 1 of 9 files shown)


src/internal/components/dropdown/index.tsx
```diff
@@ -197,6 +197,14 @@ const Dropdown = ({
     } else {
       target.style.width = position.width;
     }
+
+    // Prevent the dropdown width from stretching beyond the trigger width
+    // if that is going to cause the dropdown to be cropped because of overflow
+    if (position.overflows) {
+      target.classList.remove(styles['stretch-beyond-trigger-width']);
+      target.style.removeProperty('maxWidth');
```
pan-kot, line 205: Why maxWidth is removed?

src/internal/components/dropdown/index.tsx
```diff
@@ -197,6 +197,14 @@ const Dropdown = ({
     } else {
       target.style.width = position.width;
     }
+
+    // Prevent the dropdown width from stretching beyond the trigger width
+    // if that is going to cause the dropdown to be cropped because of overflow
+    if (position.overflows) {
+      target.classList.remove(styles['stretch-beyond-trigger-width']);
```
pan-kot, line 204: What is worrying is that the class is added and removed in different places. There is a chance that with another render of the dropdown the class is going to be re-added but not removed:

```
function DropdownConsumer () {
  const [counter, setCounter] = useState(0);
  useEffect(() => setInterval(() => setCounter(prev => prev + 1), 1000), []);
  return <Dropdown ... ><LargeContent /></Dropdown>
}
```

Also, what if the content changes? E.g. initially the dropdown renders a loading state and then the content appears. Does overflow calculation re-apply?

src/internal/components/dropdown/index.tsx
jperals: Because when `stretchBeyondTriggerWidth` is `true`, then both the `stretch-beyong-trigger-width` class and the `maxWidth` are applied (see line 106 [here](https://github.com/cloudscape-design/components/blob/main/src/internal/components/dropdown/index.tsx#L106)) and line 116 [here](https://github.com/cloudscape-design/components/blob/main/src/internal/components/dropdown/index.tsx#L118) respectively). Here we need to revert both changes.

Why is `maxWidth` applied separately from the class in the first place? I assume basically because its value comes from JavaScript, [here](https://github.com/cloudscape-design/components/blob/main/src/internal/breakpoints.ts#L11). I guess this could be refactored, e.g, by setting a CSS custom property that could be used in a ruleset under the class selector. This way adding and removing the class would be enough, and this part in particular might be a bit more readable but at the same time this would add a bit more complexity overall, for just removing one line, so not sure it's worth it.

src/internal/components/dropdown/index.tsx
jperals: You're right, there is an issue here because this `useLayoutEffect` hook has some dependencies so that it is called essentially when the dropdown changes from closed to open, but not if its content is changes while it's open. I think we will have to call `onDropdownOpen` (or at least part of it) essentially on every render. Here are two things to consider:

- Performance: the function call takes 2ms in average (on my machine), so I think this should not be a concern
- Behavior: should the dropdown placement change if its content changes? Since this hook only runs when opening, the current behavior is not to reposition the dropdown in any case while opened. I think this current behavior is good when select options are added or removed (which can happen with Autoselect) to prevent moving content that is already present in the page, but when changing the status from loading to content loaded, I think it would actually make sense to recalculate the position.

pan-kot: I agree that the dropdown placement should stay the same but the concern is it might not to. The class that fixes the issue is only removed in the effect so the next render can bring it back and as result break the positioning. This can probably be fixed by applying the class from within the effect as well, wdyt?

jperals: The code that (potentially) removes the class (and the maxWidth) is in a useLayoutEffect hook, so it will run after the component renders but before the browser paints the screen. I just moved this code to its own useLayoutEffect hook, because as said, we will have to check this on every render.

src/internal/components/dropdown/index.tsx
pan-kot: I think eventually we need to refactor that entire component but that's not in scope. What I would recommend is to add a code comment to explain why the max-width is removed.

src/internal/components/dropdown/index.tsx
pan-kot: Is there an advantage in using the layout effect with an empty dependencies array over computing the props in the render cycle, like:

```
const canOverflow = ...;
const maxWidth = canOverflow ? defaultMaxWidth : undefined;
const overflowClasses = canOverflow ? [styles['stretch-beyond-trigger-width']] : [];
```

pan-kot: Ah, no - you do need to have the layout for that kind of computation of course.

jperals: Exactly, we need to know what the rendered dimensions are going to be, so we need to render the component first...

pan-kot: I am thinking on a slightly different solution then to not introduce the potential performance degradation. What if when the overflow is banned it is reflected in a new state so that the next renders will not override it?

```
useLayoutEffect(() => {
  if (!canOverflow) {
    setBanOveflow(true);
  }
}, [...]);

return <div style={{ maxWidth: !banOverflow ? defaultMaxWidth : undefined }} className={clsx(..., !banOverflow && styles['stretch-beyond-trigger-width'])} />
```

Alternatively, the state can be replaced with e.g. a new class name (like styles['ban-overflow']) to remove the effects in CSS. This class name can be added in the layout effect but then we need to ensure it persists over the next renders which can be achieved by also storing this info in a ref.

jperals: Wouldn't the browser still do one paint with the layout that we don't want? Since this would "fix" the next render, but not the current one.

pan-kot: Not sure about this one. When the state is changed from the layout effect it probably won't render the previous one. If yes, the alternative approach can be used or the styles and classes can be also removed imperatively.

pan-kot: What you can also try is:

```
useLayoutEffect(() => {
  // ...
  if (!canOverflow) {
    banOverflowRef.current = true;
    dropdownRef.classList.add(styles['ban-overflow']);
  }
  // ...
}, [...]);

return <div ... className={clsx(..., banOverflowRef.current && styles['ban-overflow'])} />;
```

The above code ensures the `styles['ban-overflow']` is always present when we want to ban the overflow. In CSS you can use that class selector to negate the max-width and 'stretch-beyond-trigger-width' effects.

pan-kot (approved):

src/internal/components/dropdown/dropdown-fit-handler.ts
```diff
@@ -121,6 +129,69 @@ export const getInteriorAvailableSpace = (
...
+  dropdownElement: HTMLElement;
+  desiredMinWidth?: number;
+  stretchBeyondTriggerWidth?: boolean;
+}) => {
+  // Determine the width of the trigger
+  const triggerWidth = triggerElement.getBoundingClientRect().width;
+  // Minimum width is determined by either an explicit number (desiredMinWidth) or the trigger width
+  const minWidth = desiredMinWidth ? Math.min(triggerWidth, desiredMinWidth) : triggerWidth;
+  // If stretchBeyondTriggerWidth is true, the maximum width is the XS breakpoint (465px) or the trigger width (if bigger).
+  const maxWidth = stretchBeyondTriggerWidth ? Math.max(defaultMaxDropdownWidth, triggerWidth) : Number.MAX_VALUE;
```
pan-kot, line 148: why it defaults to MAX_VALUE and not undefined?

pan-kot (approved):

src/internal/components/dropdown/dropdown-fit-handler.ts
jperals: I didn't change this (only moved it around), but would it change anything? Since in line 152 we are doing a `Math.min` between this and another value which at that point should be defined. So I think the result should be the same.

pan-kot: got it 👍

PR #1560, opened 2023-09-15
feat: Accessible resize handles with keyboard activation

pan-kot:
### Description

Made table resizers activatable with keyboard so that an Enter/Space needs to be pressed before the arrow keys can be used for resizing.

Added the new API prop ariaLabels.resizerRoleDescription necessary to make the new behaviour better accessible for screen-reader users.

See K0jwA3Ib1kk0

### How has this been tested?

Added new tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/table/resizer/index.tsx
```diff
@@ -169,20 +181,51 @@ export function Resizer({
...
+        // aria-valuetext is needed because the VO announces "collapsed" when only aria-valuenow set without aria-valuemax
+        aria-valuetext={headerCellWidth.toFixed(0)}
+        aria-valuemin={minWidth}
+        data-focus-id={focusId}
+        onBlur={event => {
+          setResizerHasFocus(false);
+          if (isKeyboardDragging) {
+            setIsKeyboardDragging(false);
+          }
+          if (event.relatedTarget !== resizerToggleRef.current) {
```
timogasda, line 224: What does this check mean? Might be worth a comment.

src/table/resizer/index.tsx
```diff
@@ -169,20 +181,51 @@ export function Resizer({
...
-          setResizerHasFocus(false);
+        onBlur={event => {
+          if (event.relatedTarget !== resizerSeparatorRef.current) {
+            setResizerHasFocus(false);
+          }
         }}
-        {...resizerAriaProps}
+        // Using a custom role-description to make the element's purpose better clear.
+        // The role-description must include the word "button" to imply the interaction model.
+        // See https://www.w3.org/TR/wai-aria-1.1/#aria-roledescription
```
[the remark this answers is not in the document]
timogasda: Also, should this be part of the i18n provider?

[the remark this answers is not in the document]
pan-kot: Yes, the "ariaLabels.resizerRoleDescription" i18n key is already supported and translated.

src/table/resizer/index.tsx
pan-kot: Here, we have two actual elements that juggle focus -- the activation button and the separator. When there is a blur event we want to commit an update but not when the focus juggles between these two elements because that is merely an implementation detail.

pan-kot: I can add a code comment if you think that makes sense.

pan-kot: Yes, correct. I will add a code comment

pan-kot: Actually, I will add a code comment to another similar line is the file and remove the IF statement from here - it is not necessary to prevent the blur when focus moves back to the resizer button!

timogasda (approved):

PR #1567, opened 2023-09-20
fix: Don't display dropdown border when there is no content on autosuggest in VR

rubencarvalho:
### Description

When focusing the autosuggest input field on the ["No suggestions" example](https://cloudscape.design/components/autosuggest/?tabId=playground&example=no-suggestions) without an `empty` text, you can see the undesired collapsed border. This is because even when there is no suggestions and no empty text, we still render the dropdown content. This is a regression from Classic.

Customers need this to use Autosuggest to build a search experience. We want to support this use case, extending the component's guidelines to reflect it.

This PR adds an `shouldRenderDropdownContent` condition to the autosuggest component. This was needed because there were empty wrapper nodes passed on to the dropdown, even though there was no actual content inside of them.
This allows us to apply the correct styling to hide the borders in this case.

Related issue: AWSUI-22344

### How has this been tested?

Existing tests.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/autosuggest/internal.tsx
+++ b/src/autosuggest/internal.tsx
@@ -181,6 +181,8 @@ const InternalAutosuggest = React.forwardRef((props: InternalAutosuggestProps, r
     filteringResultsText: filteredText,
   });
 
+  const shouldRenderDropdownContent = !isEmpty || dropdownStatus.content;
+
   return (
     <AutosuggestInput
       {...restProps}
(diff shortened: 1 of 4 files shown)


src/internal/components/dropdown/styles.scss
```diff
@@ -108,13 +109,6 @@
   width: 100%;
 }
 
-// Child flex wrapper used because when combining an absolute position
-// and flexbox, IE11 will incorrectly size the container according to its parent
-// and not its content.
-.ie11-wrapper {
```
rubencarvalho, line 114: Cleaned up IE11 code, because why not!

src/autosuggest/internal.tsx
```diff
@@ -202,6 +201,7 @@ const InternalAutosuggest = React.forwardRef((props: InternalAutosuggestProps, r
       ariaLabel={ariaLabel}
       ariaRequired={ariaRequired}
       clearAriaLabel={clearAriaLabel}
+      isEmpty={isEmpty && !dropdownStatus.content}
```
pan-kot, line 204: An alternative way to not show the dropdown is to pass null as dropdownContent. This approach is used in the property-filter already.

See https://github.com/cloudscape-design/components/blob/main/src/internal/components/autosuggest-input/index.tsx#L282

pan-kot: Also take a look at potentially related PR: https://github.com/cloudscape-design/components/pull/1574. The autosuggest-input considers the presence of content but not footer. However, the empty state comes from the footer apparently.

src/internal/components/dropdown/index.tsx
```diff
@@ -120,15 +119,19 @@ const TransitionContent = ({
...
-            <DropdownContextProvider position={position}>
-              {header}
-              {children}
-              {footer}
-            </DropdownContextProvider>
-          </div>
+      <div
+        className={clsx(
+          styles['dropdown-content-wrapper'],
+          !header && !children && styles['is-empty'],
```
rubencarvalho, line 125: This is the line that I moved, the `is-empty` styles were being applied to the wrong node

src/internal/components/dropdown/index.tsx
```diff
@@ -102,7 +102,6 @@ const TransitionContent = ({
         [styles['with-limited-width']]: !stretchWidth,
         [styles['hide-upper-border']]: stretchWidth,
         [styles.interior]: interior,
-        [styles['is-empty']]: !header && !children,
```
rubencarvalho, line 105: `is-empty` styles don't belong here

src/internal/components/dropdown/styles.scss
```diff
@@ -50,6 +50,11 @@
     }
 
     border-bottom: none;
+
+    // Don't show dropdown border when there's no dropdown content
+    &.is-empty::after {
```
rubencarvalho, line 55: The `::after` pseudo element is being rendered on this node, so this gets the empty styles.

src/autosuggest/internal.tsx
```diff
@@ -181,6 +181,8 @@ const InternalAutosuggest = React.forwardRef((props: InternalAutosuggestProps, r
     filteringResultsText: filteredText,
   });
 
+  const shouldRenderDropdownContent = !isEmpty || dropdownStatus.content;
```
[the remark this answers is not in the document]
pan-kot: It is possible to add a unit test to cover the change?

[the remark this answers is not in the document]
rubencarvalho: > This might need to include dropdownStatus.footer. (In the property-filter the empty state is added to the sticky footer).

There is no `dropdownStatus.footer` here. If it `isSticky`, we will pass the `dropdownStatus.content` to the `dropdownFooter` property.

But if I got it right, I made the [change](https://github.com/cloudscape-design/components/pull/1567/files#diff-22d221ed5a882afacd0f403f72e9841f105c338be4766fa47a92ee2ef74907b5R282) to account for this concern: `open={open && (!!dropdownContent || !!dropdownFooter)}`.

> It is possible to add a unit test to cover the change?

Added!

pan-kot (approved):

PR #1577, opened 2023-09-25
fix: Disable opening animation for side Split panel

abdhalees:
### Description

A month ago we discovered that split panel opening animation is not working due to a typo, the fix was done in this PR https://github.com/cloudscape-design/components/pull/1471 .

After the fix was out we got reports that resizing the split panel is laggy and leaves a right gap while resizing, the issue is reproducible only in React 18 when `createRoot` is used and the page has many components like table.

the issue is due to some excessive rerendering issues, fixing that is not trivial as it needs refactoring different parts of AppLayout and SplitPanel  

After discussing with the team we decided to disable animations back as mitigation for now. 

Related links: AWSUI-22396

### How has this been tested?

- Upgrade to react 18
-  Use createRoot in dev pages
-  visit a dev page with split pane and a table
- before this commit you should see the same bug in the reported ticket

<details>

#### Correctness

#### Security

#### Testing

</details>

---


jperals:
Does the issue happen only when resizing, or also when showing and hiding? If only when resizing, can we remove the animation only when resizing? Because I don't think the animation should be applied while resizing anyway.

abdhalees:
> Does the issue happen only when resizing, or also when showing and hiding? If only when resizing, can we remove the animation only when resizing? Because I don't think the animation should be applied while resizing anyway.

@jperals animation should only happen when showing, however for it to work properly we need to pass the [split panel size](https://github.com/cloudscape-design/components/blob/main/src/app-layout/visual-refresh/split-panel.tsx#L144) which gets updated on resizing causing the excessive re-rendering issue, passing the size only when the split panel opens  is not possible as it will cause [wrapping issues](https://github.com/cloudscape-design/components/blob/main/src/app-layout/visual-refresh/split-panel.scss#L142) while opening the split panel, we also need to pass the last reported size after resizing finishes, I tried doing that by adding another property to both the split panel and app layout context that gets updated onkeyup after resizing or when split panel is closed but the problem was getting the first width when the page renders, multiple rerenders happen and it's tricky to get which is the last render before opening, solving that is theoretically possible but it needs lots of refactoring and optimising both in App layout and Split panel and it has a risk of breaking other more important behaviour.

jperals (approved):

PR #1578, opened 2023-09-25
fix: Do not provide artificial values into table column width change detail

just-boris:
…detail

### Description

Some of the `columnDefinitions` items might be hidden according to `visibleColumns`  and `columnDisplay` values. These should not receive any values in the event detail

Related links, issue #, if available: AWSUI-22345

### How has this been tested?

Updated related tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/table/internal.tsx
+++ b/src/table/internal.tsx
@@ -16,7 +16,7 @@ import { useRowEvents } from './use-row-events';
 import { SelectionControl, focusMarkers, useSelectionFocusMove, useSelection } from './selection';
 import { fireNonCancelableEvent } from '../internal/events';
 import { isDevelopment } from '../internal/is-development';
-import { ColumnWidthDefinition, ColumnWidthsProvider, DEFAULT_COLUMN_WIDTH } from './use-column-widths';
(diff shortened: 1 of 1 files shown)


src/table/internal.tsx
```diff
@@ -222,7 +222,7 @@ const InternalTable = React.forwardRef(
       onFocusMove: moveFocus,
       onResizeFinish(newWidth) {
         const widthsDetail = columnDefinitions.map(
-          (column, index) => newWidth[getColumnKey(column, index)] || (column.width as number) || DEFAULT_COLUMN_WIDTH
+          (column, index) => newWidth[getColumnKey(column, index)] || (column.width as number)
```
pan-kot, line 225: The type of widths array in the event detail is ReadonlyArray<number>. The current line produces ReadonlyArray<number | undefined> it looks like.

just-boris: And you are correct. When changing it to the correct type, I am getting a lot of TS errors and this is why I am cancelling this PR as it is a breaking change

PR #1581, opened 2023-09-25
chore: Disable implicit tools integration into drawers list

just-boris:
### Description

Implement this proposal: ZuAXAsmOsora

### How has this been tested?

* Updated related tests and added new dev pages for discovered corner-case
* Run screenshot tests in my dev pipeline

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/__tests__/drawers.test.tsx
+++ b/src/app-layout/__tests__/drawers.test.tsx
@@ -1,12 +1,11 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 import React from 'react';
-import { describeEachAppLayout, renderComponent, singleDrawer, manyDrawers, singleDrawerOpen } from './utils';
+import { describeEachAppLayout, renderComponent, singleDrawer, manyDrawers } from './utils';
 import createWrapper from '../../../lib/components/test-utils/dom';
 
 import { render } from '@testing-library/react';
 import AppLayout from '../../../lib/components/app-layout';
-import { TOOLS_DRAWER_ID } from '../../../lib/components/app-layout/utils/use-drawers';
 
 jest.mock('../../../lib/components/internal/hooks/use-mobile', () => ({
   useMobile: jest.fn().mockReturnValue(true),
@@ -17,7 +16,7 @@ jest.mock('@cloudscape-design/component-toolkit', () => ({
   useContainerQuery: () => [100, () => {}],
 }));
 
-describeEachAppLayout(size => {
+describeEachAppLayout(() => {
   test(`should not render drawer when it is not defined`, () => {
     const { wrapper, rerender } = renderComponent(<AppLayout contentType="form" toolsHide={true} {...singleDrawer} />);
     expect(wrapper.findDrawersTriggers()).toHaveLength(1);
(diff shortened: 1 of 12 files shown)


src/app-layout/__tests__/drawers.test.tsx
```diff
@@ -35,82 +34,14 @@ describeEachAppLayout(size => {
     const { wrapper } = renderComponent(<AppLayout contentType="form" {...emptyDrawerItems} />);
 
     expect(wrapper.findDrawersTriggers()).toHaveLength(0);
-    expect(wrapper.findToolsToggle()).toBeTruthy();
+    expect(wrapper.findToolsToggle()).toBeFalsy();
   });
 
-  test('should apply drawers treatment to the tools if at least one other drawer is provided', () => {
```
just-boris, line 41: This and other test from this file moved into `runtime-drawers.test.tsx`, because the tools+drawers merging feature is now only available for runtime drawers

src/app-layout/utils/use-drawers.ts
```diff
@@ -95,28 +110,32 @@ export function useDrawers(
...
+
+  function onActiveDrawerChange(newDrawerId: string | undefined) {
     setActiveDrawerId(newDrawerId);
-    fireNonCancelableEvent(ownDrawers?.onChange, newDrawerId);
-  });
+    if (hasOwnDrawers) {
+      fireNonCancelableEvent(ownDrawers?.onChange, newDrawerId);
+    }
+    if (!toolsProps.toolsHide) {
+      fireNonCancelableEvent(toolsProps.onToolsChange, { open: newDrawerId === TOOLS_DRAWER_ID });
```
just-boris, line 124: Removed repeating `onToolsChange` handlers from other places, this is the only one where we keep it to support compatibility between tools and drawers

src/app-layout/visual-refresh/context.tsx
```diff
@@ -36,7 +36,7 @@ import { useUniqueId } from '../../internal/hooks/use-unique-id';
 
 interface AppLayoutInternals extends AppLayoutProps {
   activeDrawerId: string | undefined;
-  drawers: Array<DrawerItem>;
+  drawers: Array<DrawerItem> | null;
```
just-boris, line 39: Empty array `[]` means that `<AppLayout drawers={[]}>` is used and `null` means that there are no drawers registered in any form. This distinction is needed to decide what we render on the right hand side

src/app-layout/visual-refresh/layout.tsx
```diff
@@ -50,7 +50,7 @@ export default function Layout({ children }: LayoutProps) {
 
   // Content gaps on the left and right are used with the minmax function in the CSS grid column definition
   const hasContentGapLeft = isNavigationOpen || navigationHide;
-  const hasContentGapRight = drawersTriggerCount <= 0 || hasOpenDrawer;
+  const hasContentGapRight = drawersTriggerCount === 0 || hasOpenDrawer;
```
just-boris, line 53: I do not think this count can ever be negative number, so decided to simplify for readability

src/app-layout/utils/use-drawers.ts
YueyingLu: Do we need to check if tools are ignored?

just-boris: We still have "multi drawer with tools mode" for runtime API, this is why the condition is here. See `applyToolsDrawer` function how exactly this works

YueyingLu (approved):

PR #1587, opened 2023-09-26
fix: Aligns drawers slider to split panel and matches side nav animation

katiegeorge:
### Description

The slider of drawers didn't line up vertically with the slider of split panel.

Before:
image

After:
image

Also, the drawer was animated both in and out, which didn't match the navigation on the other side. This removes the animation out so that the two match. There is a separate issue with the animation that creates lag in certain scenarios, but since that is a more complex fix, this is a holdover for the launch of drawers until that bug can be fixed.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/visual-refresh/drawers.scss
+++ b/src/app-layout/visual-refresh/drawers.scss
@@ -21,10 +21,10 @@
   &.has-open-drawer {
     background-color: awsui.$color-background-container-content;
     box-shadow: awsui.$shadow-panel;
-  }
 
-  @include styles.with-motion {
-    transition: background-color awsui.$motion-duration-refresh-only-fast;
+    @include styles.with-motion {
+      transition: background-color awsui.$motion-duration-refresh-only-fast;
+    }
   }
 
   @include styles.media-breakpoint-up(styles.$breakpoint-x-small) {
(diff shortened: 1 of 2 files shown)


src/app-layout/visual-refresh/drawers.scss
```diff
@@ -156,7 +156,7 @@
 
   > .drawer-slider {
     grid-column: 1;
-    grid-row: 3;
+    grid-row: 1 / span 3;
```
rubencarvalho, line 159: nit: I'm not sure if the number of rows might change, but we could be a bit more defensive here.
```suggestion
    grid-row: 1 / -1;
```

rubencarvalho:
Do we have a visual regression test that could cover this? :)

rubencarvalho (approved):
I left a couple of non-blocking comments 👍

src/app-layout/visual-refresh/drawers.scss
katiegeorge: This is a great solution! There are a couple other places in this file that use the `span` method, so they should all be consistent. I think this works for now, since the rows won't change.

katiegeorge:
> Do we have a visual regression test that could cover this? :)

We currently have screenshot tests for an open drawer, but not one with an open drawer and open split panel. I'll make a note to add one as a follow up.

PR #1594, opened 2023-09-28
chore: Change default Funnel Name selector for Forms and Modals

michaeldowseza:
### Description

We saw two issues with the funnel names reported by some customers that were either 1. reporting incorrectly due to asynchronously render their breadcrumbs for various reasons or 2. using a modal with an embedded form would use the breadcrumb as its funnel name which might be unrelated to the modal contents (e.g. Feedback dialogs). 

This PR tries to improve this for Forms by allowing it to rather prefer using the Form Header slot instead of the breadcrumb and for Modals it introduces a new context where it can provide the preferred funnel name selector to the Funnel.

New behaviour:
- Forms will now default to the Form Header slot instead of Breadcrumb
- Forms inside a Modal will default to the Modal Header instead of Breadcrumb

### How has this been tested?

Updated unit tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/form/__tests__/analytics.test.tsx
+++ b/src/form/__tests__/analytics.test.tsx
@@ -5,16 +5,18 @@ import { act, render } from '@testing-library/react';
 
 import createWrapper from '../../../lib/components/test-utils/dom';
 import Button from '../../../lib/components/button';
+import Container from '../../../lib/components/container';
 import Form from '../../../lib/components/form';
+import Header from '../../../lib/components/header';
 import Modal from '../../../lib/components/modal';
-import BreadcrumbGroup from '../../../lib/components/breadcrumb-group';
 
 import { FunnelMetrics } from '../../../lib/components/internal/analytics';
 import { useFunnel } from '../../../lib/components/internal/analytics/hooks/use-funnel';
 
 import { mockFunnelMetrics, mockInnerText } from '../../internal/analytics/__tests__/mocks';
-import Container from '../../../lib/components/container';
-import Header from '../../../lib/components/header';
+
+import formStyles from '../../../lib/components/form/styles.selectors.js';
+import modalStyles from '../../../lib/components/modal/styles.selectors.js';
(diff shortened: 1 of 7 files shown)


src/form/__tests__/analytics.test.tsx
```diff
@@ -350,4 +339,25 @@ describe('Form Analytics', () => {
...
+    render(
+      <Modal header="My Modal Funnel" visible={true}>
+        <Form>
+          <Container header={<Header>Substep one</Header>}></Container>
+          <Container header={<Header>Substep two</Header>}></Container>
+        </Form>
+      </Modal>
+    );
+
+    act(() => void jest.runAllTimers());
```
just-boris, line 353: Not blocking for this PR, but I have received customer complains about analytics caused "unwrapped act" warnings. We need to more carefully consider every case where we have to wrap some updates in `act` in order to work

src/internal/analytics/components/analytics-funnel.tsx
```diff
@@ -68,6 +68,8 @@ const InnerAnalyticsFunnel = ({ children, stepConfiguration, ...props }: Analyti
   const [funnelInteractionId, setFunnelInteractionId] = useState<string>('');
   const [submissionAttempt, setSubmissionAttempt] = useState(0);
   const isVisualRefresh = useVisualRefresh();
+  const inheritedFunnelNameSelector = useFunnelNameSelector();
+  const funnelNameSelector = props.funnelNameSelector || inheritedFunnelNameSelector || getFunnelNameSelector();
```
just-boris, line 72: Because `getFunnelNameSelector` is a function returning a constant value, I would recommend refactoring it into a plain constant.

Current state makes it prone to having side effects inside React's render function, possibly returning incorrect selectors

just-boris (approved):
Approved with follow up suggestions

PR #1603, opened 2023-10-02
fix: Remove unnecessary preventDefault in mouse event handlers

just-boris:
### Description

Remove this code because it triggers an error message: "Unable to preventDefault inside passive event listener due to target being treated as passive"

Related links, issue #, if available: related code was added in CR-56013320

### How has this been tested?

Locally, no differences

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/button-dropdown/category-elements/expandable-category-element.tsx
+++ b/src/button-dropdown/category-elements/expandable-category-element.tsx
@@ -45,8 +45,7 @@ const ExpandableCategoryElement = ({
     }
   };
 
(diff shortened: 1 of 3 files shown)


Al-Dani (approved):

just-boris:
@pan-kot since you added this code, can you have a look too?

Can you explain why you added `event.preventDefault()` in the first place, because for me it seems like it does nothing useful

pan-kot (approved):

PR #1611, opened 2023-10-04
refactor: Property filter internal settings lookup

pan-kot:
### Description

Refactored property filter as a preparation for the API extension to decouple properties and definitions. Added a new test page to imitate one of the current problems - when the properties are not provided the tokens lack labels, formatters, and custom forms that come from the properties. This problem is meant to be fixed by https://github.com/cloudscape-design/components/pull/689.

Related links: AWSUI-22200

### How has this been tested?

Existing tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/property-filter/controller.ts
```diff
@@ -127,21 +120,13 @@ export const parseText = (
...
-
 interface OptionGroup<T> {
   label: string;
   options: T[];
 }
 
 export const getAllValueSuggestions = (
-  filteringOptions: readonly InternalFilteringOption[],
-  filteringProperties: readonly InternalFilteringProperty[],
+  filteringSettings: FilteringSettings,
```
just-boris, line 129: This "knows it all" `filteringSettings` objects violates system design principles. Each component should receive only the minimum information it needs to operate. 

This "god object" breaks encapsulation between property filter sub components

just-boris (commented):
I see this PR as a step backwards in code quality. The main objective of decoupling properties and definitions could be solved without this

I can approve it you insist, but still recommend to re-consider your system design decisions

src/property-filter/controller.ts
pan-kot: In the component the filtering properties and filtering options are used together more often than not. If you think it makes sense I can ensure the functions only receive the data they actually consume e.g. if only options are used - then only the options will be passed, not the entire settings object.

On the brighter side this settings structure helps in the following ways:
1. Encapsulates the relations between properties and options (and is going to be extended to also include definitions);
2. Performs all necessary mappings in one place so that in the downstream functions there is no longer a need to find properties by key or options by property key.

src/property-filter/index.tsx
```diff
@@ -158,46 +159,58 @@ const PropertyFilter = React.forwardRef(
...
-    const internalFilteringOptions: readonly InternalFilteringOption[] = filteringOptions.map(option => {
-      const formatter = propertyByKey.get(option.propertyKey)?.getValueFormatter();
       return {
-        propertyKey: option.propertyKey,
-        value: option.value,
-        label: formatter ? formatter(option.value) : option.label ?? option.value ?? '',
+        properties: [...propertyByKey.values()],
+        options,
+        getProperty,
+        getPropertyOptions,
```
just-boris, line 205: 2nd system design issue. Components propagate data top down. Propagating getters violates this principle because communication becomes bi directional.


Ideally props propagation should be like a funnel, narrowing down more specific things on each new component level

src/property-filter/controller.ts
just-boris: Your "bright side" points is the book definition of the "god object" anti-pattern. Try to simplify individual components, and not create mega-controller (aka `filteringSettings`) where it is not needed.

src/property-filter/index.tsx
just-boris: Also see another comment below. Found another red flag in this implementation: https://github.com/cloudscape-design/components/pull/1611#discussion_r1348626190

src/property-filter/controller.ts
just-boris: Also see another comment below. Found another red flag in this implementation: https://github.com/cloudscape-design/components/pull/1611#discussion_r1348626190

src/property-filter/index.tsx
pan-kot: I don't think it violates the React principles as those functions are just getters - an interface on top of the data structure. An alternative could be to pass down structure itself like a bunch of dictionaries and collections, but the getters contribute to better encapsulation.

Consider:
```
function createMapping(...) {
  const propertiesByKey = new Map();
  // ... create mapping from properties
  return propertiesByKey;
}

// and

function createMapping(...) {
  const propertiesByKey = new Map();
  // ... create mapping from properties
  const getProperty = (key: string) => {
    const property = propertiesByKey.get(key);
    if (!key) {
      throw new Error('Invariant violation: missing property');
    }
    return property;
  }
  return { getProperty };
}
```

Both functions perform the same work but the second one results in a better API by validating the output so that consumers should no longer care about the missing properties. It also more restrictive as it does not offer the list of available properties and it can't be mutated.

src/property-filter/__tests__/utils.test.ts
```diff
@@ -108,47 +108,55 @@ describe('matchOperatorPrefix', () => {
 });
 
 describe('matchTokenValue', () => {
+  const property = filteringProperties.find(p => p.propertyKey === 'key') ?? null;
```
just-boris, line 111: `filteringProperties` is a constant object declared above. Do you need `?? null` if you can confidently say it is there?

or if it is not there, can this code be simplified?

src/property-filter/__tests__/common.ts
```diff
@@ -59,9 +59,12 @@ export function toInternalProperties(properties: FilteringProperty[]): InternalF
   }));
 }
 
-export function toInternalOptions(options: FilteringOption[]): InternalFilteringOption[] {
+export function toInternalOptions(
```
just-boris, line 62: This is a logic duplication. What if the logic of this function deviate of the real one here: https://github.com/cloudscape-design/components/blob/70bafc50860ae991853c407766ca4d4a11199965/src/property-filter/index.tsx#L185-L190

Can you trust these tests after it?

src/property-filter/index.tsx
```diff
@@ -149,65 +150,75 @@ const PropertyFilter = React.forwardRef(
...
-        operators: (property.operators ?? []).map(op => (typeof op === 'string' ? op : op.operator)),
-        defaultOperator: property.defaultOperator ?? '=',
-        getValueFormatter: operator => (operator ? extendedOperators.get(operator)?.format ?? null : null),
-        getValueFormRenderer: operator => (operator ? extendedOperators.get(operator)?.form ?? null : null),
-        externalProperty: property,
-      };
-    });
-
-    const propertyByKey = new Map(internalFilteringProperties.map(p => [p.propertyKey, p]));
+    const { internalProperties, internalOptions, internalQuery } = useMemo(() => {
```
just-boris, line 161: This memo is useless, because collection hooks produces new array on every render anyway: https://github.com/cloudscape-design/collection-hooks/blob/main/src/utils.ts#L116

Please do not add any `useMemo` until you are really sure it works.

pages/property-filter/single-use-properties.page.tsx
```diff
@@ -0,0 +1,36 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
just-boris, line 1: Do I understand correctly that this demo currently does not work?

How about not adding this demo into the mainline for now?

pan-kot: Alright.

just-boris (approved):

PR #1615, opened 2023-10-04
feat: Make Input aria-label available when using FormField

Al-Dani:
### Description

Currently when using Input inside FormField, `aria-labelledby` is always being set from FormField context. Because `aria-labelledby` has precedence over `aria-label` in accessible name calculation, it's not possible to use `aria-label` on Input to add additional context.

Related links, issue #, if available: AWSUI-22502

### How has this been tested?

New tests added

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/input/internal.tsx
+++ b/src/input/internal.tsx
@@ -101,7 +101,7 @@ function InternalInput(
 
   const attributes: React.InputHTMLAttributes<HTMLInputElement> = {
     'aria-label': ariaLabel,
-    'aria-labelledby': ariaLabelledby,
+    'aria-labelledby': ariaLabel ? undefined : ariaLabelledby,
     'aria-describedby': ariaDescribedby,
     name,
     placeholder,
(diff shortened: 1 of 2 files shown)


src/input/internal.tsx
```diff
@@ -101,7 +101,7 @@ function InternalInput(
 
   const attributes: React.InputHTMLAttributes<HTMLInputElement> = {
     'aria-label': ariaLabel,
-    'aria-labelledby': ariaLabelledby,
+    'aria-labelledby': ariaLabel && !rest.ariaLabelledby ? undefined : ariaLabelledby,
```
pan-kot, line 104: I would add a code comment to explain the expectations. Alternatively, a util might make sense as we have integration with the form field in several components

Al-Dani: I'll add a comment for now. We have a different treatment of this property across components, so we need to agree on a universal solution. But the issue with input is blocking customers

pan-kot (approved):

PR #1624, opened 2023-10-06
fix: Adds min-width to drawer

katiegeorge:
### Description

From Drawers Bug Bash.

No min-width on drawers caused situations like this on small screens:
image

Added a min-width of 290 for feature parity with tools.

It should be noted that situations like this can still happen, which isn't ideal, but should be handled holistically:
image

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/__integ__/app-layout-drawers.test.ts
+++ b/src/app-layout/__integ__/app-layout-drawers.test.ts
@@ -128,9 +128,7 @@ for (const visualRefresh of [true, false]) {
 
     test(
       `should not shrink drawer beyond min width`,
-      setupTest({}, async page => {
-        // this width is small enough that it would hit the min-width of a drawer
-        await page.setWindowSize({ ...viewports.desktop, width: 700 });
+      setupTest({ screenSize: { ...viewports.desktop, width: 700 } }, async page => {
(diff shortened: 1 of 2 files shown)


src/app-layout/__integ__/app-layout-drawers.test.ts
```diff
@@ -126,6 +126,18 @@ for (const visualRefresh of [true, false]) {
       })
     );
 
+    test(
+      `should not shrink drawer beyond min width`,
+      setupTest({}, async page => {
+        // this width is small enough that it would hit the min-width of a drawer
+        await page.setWindowSize({ ...viewports.desktop, width: 700 });
```
just-boris, line 133: Screen size can go here in the config

```suggestion
      setupTest({ screenSize: { ...viewports.desktop, width: 700 } }, async page => {
```

The upside of this, it resizes the window before opening the page, so we do not need to worry if all resize listeners already worked or not

just-boris: I can approve after apply this change

just-boris (approved):

PR #1632, opened 2023-10-09
fix: Removes jumping drawer caret by removing motion class

katiegeorge:
### Description

From Drawers Bug Bash.

In VR, the close caret was jumping when resizing via keyboard due to a motion class. This removes that class on keypress.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/utils/use-keyboard-events.ts
+++ b/src/app-layout/utils/use-keyboard-events.ts
@@ -3,6 +3,7 @@
 import React from 'react';
 import { KeyCode } from '../../internal/keycode';
 import { SizeControlProps } from './interfaces';
+import styles from '../styles.css.js';
 
 const KEYBOARD_SINGLE_STEP_SIZE = 10;
 const KEYBOARD_MULTIPLE_STEPS_SIZE = 60;
@@ -27,6 +28,8 @@ export const useKeyboardEvents = ({ position, onResize, panelRef }: SizeControlP
     let maxSize;
 
     const { panelHeight, panelWidth } = getCurrentSize(panelRef);
+
+    panelRef?.current?.classList.remove(styles['with-motion']);
 
     if (position === 'side') {
       currentSize = panelWidth;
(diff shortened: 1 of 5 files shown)


gethinwebster (approved):

src/app-layout/utils/use-keyboard-events.ts
```diff
@@ -21,13 +22,17 @@ const getCurrentSize = (panelRef?: React.RefObject<HTMLDivElement>) => {
...
-export const useKeyboardEvents = ({ position, onResize, panelRef }: SizeControlProps) => {
+export const useKeyboardEvents = ({ position, onResize, panelRef, hasTransitions = false }: SizeControlProps) => {
   return (event: React.KeyboardEvent) => {
     let currentSize;
     let maxSize;
 
     const { panelHeight, panelWidth } = getCurrentSize(panelRef);
 
+    if (hasTransitions) {
+      panelRef?.current?.classList.remove(styles['with-motion']);
```
just-boris, line 33: We do not add this class programmatically, React does it in `<ActiveDrawer />`. Removing it imperatively may put it out of think with React, which may think the class is still there.

It should always be symmetric. Either we fully manage the class via `classList.add` and `classList.remove` or we fully delegate it to React via `useState` or something.

I am fine either way as long as they do not mix

src/app-layout/visual-refresh/context.tsx
```diff
@@ -399,7 +401,14 @@ export const AppLayoutInternalsProvider = React.forwardRef(
...
     });
 
+    const previousActiveDrawerId = useRef(activeDrawerId);
+
+    useEffect(() => {
+      previousActiveDrawerId.current = activeDrawerId;
+    }, [activeDrawerId]);
+
     const handleDrawersClick = (id: string | undefined, skipFocusControl?: boolean) => {
+      setPrevActiveDrawerId(previousActiveDrawerId.current);
```
just-boris, line 411: This is not guaranteed to be always the previous. For example, opening the drawer programmatically via "info" link does not visit this line, so `prevActiveDrawerId` won't be updated.

Following the comment from above I would recommend to avoid relying on previous value that much. Open for more discussion in DM

katiegeorge:
Closing in favor of https://github.com/cloudscape-design/components/pull/1653 since bugs keep popping up related to animations. Removing animations until we can solve for split panel and drawer holistically.

PR #1633, opened 2023-10-09
fix: Fixes scroll discrepencies between drawer and tools

katiegeorge:
### Description
From Drawers Bug Bash.

With both tools and drawers, having both tools and a side split panel open simultaneously was not allowing for adequate resize of split panel.

This merges the variables into a single variable when calculating the max width of the split panel, so that the two function the same.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/__integ__/runtime-drawers.test.ts
+++ b/src/app-layout/__integ__/runtime-drawers.test.ts
@@ -17,6 +17,7 @@ for (const visualRefresh of [true, false]) {
           `#/light/app-layout/runtime-drawers?${new URLSearchParams({
             hasDrawers: 'false',
             hasTools: 'true',
+            splitPanelPosition: 'side',
             visualRefresh: `${visualRefresh}`,
           }).toString()}`
         );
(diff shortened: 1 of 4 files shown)


src/app-layout/visual-refresh/context.tsx
```diff
@@ -541,17 +541,10 @@ export const AppLayoutInternalsProvider = React.forwardRef(
       function handleSplitPanelMaxWidth() {
         const contentGapRight = 80; // Approximately 40px when rendered but doubled for safety
         const toolsFormOffsetWidth = 160; // Approximately 80px when rendered but doubled for safety
-        const toolsOffsetWidth = isToolsOpen ? toolsWidth : 0;
-        const activeDrawerOffsetWidth = activeDrawerId ? drawerSize : 0;
+        const panelOffsetWidth = isToolsOpen ? toolsWidth : activeDrawerId ? drawerSize : 0;
```
just-boris, line 544: Tools width should be accounted only when drawers are not used

```suggestion
        const panelOffsetWidth = drawers ? activeDrawerId ? drawerSize : 0 ? isToolsOpen ? toolsWidth : 0;
```

(or refactor this to if-else, because it is getting too clunky)

src/app-layout/__integ__/runtime-drawers.test.ts
```diff
@@ -38,6 +39,26 @@ for (const visualRefresh of [true, false]) {
       })
     );
 
+    test.only(
```
just-boris, line 42: Missing debug bit

```suggestion
    test(
```

src/app-layout/__integ__/runtime-drawers.test.ts
```diff
@@ -38,6 +39,26 @@ for (const visualRefresh of [true, false]) {
...
+      'should resize equally with tools or drawers',
+      setupTest(async page => {
+        await page.setWindowSize({ ...viewports.desktop, width: 1800 });
+        await page.click(wrapper.findToolsToggle().toSelector());
+        await page.click(wrapper.findSplitPanel().findOpenButton().toSelector());
+        await page.click(wrapper.findSplitPanel().findPreferencesButton().toSelector());
+        const tile = createWrapper().findModal().findContent().findTiles().findItemByValue('side');
+        await page.click(tile.toSelector());
+        await page.click('button=Confirm');
```
just-boris, line 51: I would be more happy to add support for the query parameter on this page as well, rather than doing it manually

avinashbot (approved):

PR #1638, opened 2023-10-11
refactor: Column auto width is set in column widths utils

pan-kot:
### Description

Moving column width related logic loser together. It is a step towards resolving the issue visible with React 18 when containerWidth changes in an effect causing the auto width to be calculated one render too late. As result the last column can be resized below its min width for a short while.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/table/thead.tsx
+++ b/src/table/thead.tsx
@@ -130,7 +130,7 @@ const Thead = React.forwardRef(
 
           {columnDefinitions.map((column, colIndex) => {
             const columnId = getColumnKey(column, colIndex);
-            const columnWidth = resizableColumns && columnWidths ? columnWidths[columnId] : column.width;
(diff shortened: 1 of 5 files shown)


src/table/thead.tsx
```diff
@@ -46,7 +46,6 @@ export interface TheadProps {
 const Thead = React.forwardRef(
   (
     {
-      containerWidth,
```
Al-Dani, line 49: I guess no need to path it as a prop here
https://github.com/cloudscape-design/components/blob/8c93f3423013523ec63f20af7c6587f2bc013aa9/src/table/internal.tsx#L211

Al-Dani (approved):

PR #1642, opened 2023-10-11
fix: Render a hidden copy of tools drawer for compatibility when drawers are used

just-boris:
### Description

In a single-drawer (tools only) mode, app layout renders `tools` slot content even when it is closed. Some teams are relying on this. When rendering multiple drawers we tried to render only the active one, but this broke compatibility.

This PR restores compatibility, we now always keep the tools content in the DOM, only hide it when it is closed
 
### How has this been tested?

Added a dev page `pages/app-layout/runtime-drawers-imperative.page.tsx` to reproduce this case and added several unit tests to capture this behavior

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/__tests__/main.test.tsx
+++ b/src/app-layout/__tests__/main.test.tsx
@@ -169,29 +169,3 @@ test('a11y', async () => {
   );
   await expect(container).toValidateA11y();
 });
-
-test('drawers a11y', async () => {
-  const { container } = renderComponent(
-    <AppLayout
-      navigationOpen={true}
-      toolsOpen={true}
-      splitPanelOpen={true}
-      navigation={<div></div>}
-      content={<div></div>}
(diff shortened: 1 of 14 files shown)


src/app-layout/__tests__/main.test.tsx
```diff
@@ -169,29 +169,3 @@ test('a11y', async () => {
   );
   await expect(container).toValidateA11y();
 });
-
-test('drawers a11y', async () => {
```
just-boris, line 173: This test does not work anymore, because we now render drawer HTML, even when it is closed.

I do not think there is a way to make this test work while keeping the current implementation.

The real browser tests will cover this anyway: https://github.com/cloudscape-design/components/tree/main/src/__a11y__

PR #1644, opened 2023-10-12
feat: Support links and nested items in chart popovers

jperals:
### Description

This PR adds a new API `DetailPopoverSeriesContent` which enables two new features for chart popovers:

1. Arbitrary content for keys and values, in order to support drilldown links to navigate to a different page. This applies to all chart types line, bar, mixed line-bar, pie and donut charts (all chart types except area charts).
1. Nested key-value pairs, to see further breakdown information for a datum of a series. These nested lists of key-value pairs can be either plain lists or expandable. This applies to line, bar and mixed line-bar charts (not to area, pie and donut charts).

Also includes additions to the testing utils API.

API proposal document: `MlKVAmdIvCL8`

### How has this been tested?

- Added unit tests
- Added example dev pages for which visual regression tests will be provided

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/internal/components/chart-series-details/index.tsx
```diff
@@ -1,49 +1,117 @@
...
+              [styles['with-sub-items']]: subItems?.length,
+              [styles.expandable]: expandable,
             })}
           >
-            <div className={styles.key}>
-              {markerType && color && <ChartSeriesMarker type={markerType} color={color} />}
-              <span>{key}</span>
-            </div>
-            <InternalBox textAlign="right">{value}</InternalBox>
+            {subItems?.length && expandable ? (
```
pan-kot, line 72: not: can this be simplified a bit with introducing more components e.g. <ExpandableItemsGroup subItems={subItems} ... />, <StaticItemsGroup subItems={subItems} ... />?

```
{hasSubItems && expandable ? <ExpandableitemsGroup ... /> : <StaticItemsGroup ... />}
```

src/internal/components/chart-series-details/styles.scss
```diff
@@ -7,10 +7,54 @@
 @use '../../styles/tokens' as awsui;
 @use './motion';
 
+@use '../chart-series-marker/styles' as marker;
+
+@mixin bold {
```
pan-kot, line 12: Do we really need this mixin? It is just one line of code using twice. If it is important to secure the value is in sync a variable can do.

src/internal/components/chart-series-marker/styles.scss
```diff
@@ -6,12 +6,15 @@
 @use '../../styles' as styles;
 @use '../../styles/tokens' as awsui;
 
+$marker-width: 1.4 * styles.$base-size;
+$marker-margin-right: awsui.$space-xxs;
```
pan-kot, line 10: Why the variable for the right margin is needed? (And why there is no variable for the top margin?)

src/line-chart/interfaces.ts
```diff
@@ -24,6 +24,13 @@ export interface LineChartProps<T extends ChartDataTypes>
    * See the usage guidelines for more details.
    */
   emphasizeBaselineAxis?: boolean;
+
+  /**
+   * Specifies custom rendering of the series displayed in the chart popover.
+   * Use this for wrapping keys or values in links, or to display an additional
+   * level of nested items.
+   */
+  detailPopoverSeriesContent?: MixedLineBarChartProps.DetailPopoverSeriesContent<T>;
```
pan-kot, line 33: Can we reuse this via CartesianChartProps? As we don't provide the feature yet for the AreaChart we can exclude it from there with:

```
// Omitting detailPopoverSeriesContent because the feature is not yet supported ...
interface AreaChartProps<T> extends Omit<CartesianChartProps<T>, 'detailPopoverSeriesContent'>{ ... }
```

src/mixed-line-bar-chart/__integ__/announcements.test.ts
```diff
@@ -0,0 +1,72 @@
...
+        const getLabel = () => page.getElementAttribute(bar, 'aria-label');
+        await page.hoverElement(bar);
+        await page.waitForAssertion(async () => {
+          const label = await getLabel();
+          for (const text of alwaysVisibleTexts) {
+            expect(label).toContain(text);
+          }
+          // Collapsed sub-items should not be announced
+          expect(label).toContain(collapsedText);
+          expect(label).not.toContain('[object object]');
```
pan-kot, line 25: Would it be easier and safer to test against the actual value e.g. `expect(label).toBe('....')`

src/test-utils/dom/internal/chart-popover.ts
```diff
@@ -4,6 +4,25 @@ import { ComponentWrapper, ElementWrapper } from '@cloudscape-design/test-utils-
...
+    return this.findByClassName(chartSeriesDetails.key);
+  }
+
+  findValue(): ElementWrapper | null {
+    return this.findByClassName(chartSeriesDetails.value);
+  }
+}
+
+export class ChartPopoverSeriesWrapper extends ChartPopoverSeriesItemWrapper {
+  findSubItems(): Array<ChartPopoverSeriesItemWrapper> {
```
pan-kot, line 20: 1. Do we really need to differentiate between items and sub-items in test utils?
2. Would it make sense to add utils like getItemByIndex(index) or getItemByTitle(title)?

src/bar-chart/__tests__/chart-series-details.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+const series = [barSeries, thresholdSeries];
+
+function renderChart(props: Omit<BarChartProps<string>, 'series'>) {
+  const { container } = render(<BarChart {...props} series={series} />);
+  return {
+    wrapper: new BarChartWrapper(container),
+  };
+}
+
+testChartSeriesDetails({ renderChart });
```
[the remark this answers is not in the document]
jperals: From the customer perspective they are different components, and we have the behavior unified in the mixed line-bar chart but that is a matter of internal implementation, there could always be divergences so I feel better testing the feature in all of them.

src/test-utils/dom/internal/chart-popover.ts
jperals: The testing utils were part of the API review so I would prefer to not deviate unless having a strong reason. To the specific questions:

1. I see a clear hierarchy between top-level items, which represent actual series displayed in the graph, and their sub-items, which are arbitrarily defined by the customers, so I think an API that reflects that makes sense. And consider the case where sub-items are expandable, it's not clear how a flat-list approach should work (should all items always be returned? If not, should the results update and the indices change when expanding/collapsing?) and it would be more confusing IMO.
2. Regarding `getItemByIndex`, you can already access the items by index in unit tests or use `.get(0)` in integration tests, so would it bring much? Regarding `getItemByTitle`, maybe it wouldn't hurt, but I'm not sure iI would add it before having specific data about its need.

src/mixed-line-bar-chart/__integ__/announcements.test.ts
jperals: Refined the test a bit [here](https://github.com/cloudscape-design/components/pull/1644/commits/7c8c1496906031dec36a1076d338a94be7ba5ff3), added a few more texts and using `toBe` once in line 44. However, the reason to use `toContain` is basically the expand/collapse functionality: when expanding, sub-items are inserted just after their parent and not at the end, so I need to be able compose by picking and choosing elements.

src/expandable-section/expandable-section-header.tsx
```diff
@@ -190,12 +188,19 @@ const ExpandableHeaderTextWrapper = ({
...
         <>
-          <HeadingTag
-            className={clsx(styles['header-wrapper'], headingTagListeners && styles['click-target'])}
-            {...headingTagListeners}
-          >
-            {headerButton}
-          </HeadingTag>
+          {variant === 'compact' ? (
+            <div className={clsx(styles['header-wrapper'], styles['header-wrapper-compact'])}>
+              <HeadingTag className={styles['click-target']}>{headerButton}</HeadingTag>
```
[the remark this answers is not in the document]
jperals: Because we only want the key to be clickable and not the value, whereas the default and footer variants today make the entire header wrapper clickable, although this is something we want to change in the future (see doc: `7SZJAvbhbznX`). When that happens, this code will be a bit simpler. For now, I have refactored this is a bit further, so hopefully it's a bit more clear.

src/expandable-section/expandable-section-header.tsx
```diff
@@ -190,12 +188,19 @@ const ExpandableHeaderTextWrapper = ({
...
+              {headerActions}
```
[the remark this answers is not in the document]
jperals: I have refactored this and now the condition does not depend on whether the variant is `compact` but on whether actions are present and supported for the variant. This also brings us closer to support actions for default and footer variants (AWSUI-21326) and when that happens this can be unified.

Although the slot is called `headerActions`, I think using it for other elements is OK, and this is less confusing or forced than other options I can think of. Note that `headerText` is a string-only prop and the ReactNode `header` prop is deprecated.

src/internal/components/chart-series-details/index.tsx
```diff
@@ -1,49 +1,117 @@
...
   const baseProps = getBaseProps(restProps);
   const className = clsx(baseProps.className, styles.root);
+  const detailsRef = useRef<HTMLDivElement | null>(null);
+
+  // Once the component has rendered, pass its content in plain text
+  // so that it can be used by screen readers.
+  useEffect(() => {
+    if (setPopoverText) {
+      if (detailsRef.current) {
+        setPopoverText(getSeriesDetailsText(detailsRef.current));
```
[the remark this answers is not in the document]
jperals: Almost but it doesn't add commas and screen readers read everything as one sentence without stopping. JAWS, NVDA and VoiceOcer all briefly pause when a comma is found, see [here](https://www.deque.com/blog/dont-screen-readers-read-whats-screen-part-1-punctuation-typographic-symbols/#:~:text=The%20house%20is%20green%2C%20red%2C%20and%20orange.), and since the current approach adds commas between key-value pairs, this would be a regression, even if a small one.

src/internal/components/chart-series-details/index.tsx
```diff
@@ -1,49 +1,117 @@
...
   color?: string;
   isDimmed?: boolean;
+  subItems?: ReadonlyArray<ChartDetailPair>;
+  expandable?: boolean;
 }
+export type ExpandedStates = Record<number, boolean>;
 
 export interface ChartSeriesDetailsProps extends BaseComponentProps {
   details: ReadonlyArray<ChartSeriesDetailItem>;
+  expandedStates?: ExpandedStates;
```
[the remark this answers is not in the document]
jperals: This has to be defined outside because we want to keep the expanded state of each expandable series of each X coordinate independently, and otherwise the state would be lost when closing and therefore unmounting the popover. And it needs to be optional because this component is used by all charts but the feature is not supported by area or pie charts.

src/internal/components/chart-series-details/index.tsx
```diff
@@ -1,49 +1,117 @@
...
+}
 
 export interface ChartSeriesDetailItem extends ChartDetailPair {
   markerType?: ChartSeriesMarkerType;
   color?: string;
   isDimmed?: boolean;
+  subItems?: ReadonlyArray<ChartDetailPair>;
+  expandable?: boolean;
 }
+export type ExpandedStates = Record<number, boolean>;
```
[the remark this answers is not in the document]
jperals: Switched to a Set [here](https://github.com/cloudscape-design/components/pull/1644/commits/eed080b6d937174b80c34e58d76c62482fd5aca7), thanks for the suggestion.

src/internal/components/chart-series-marker/styles.scss
jperals: I'm using them in ChartSeriesDetails, [here](https://github.com/cloudscape-design/components/pull/1644/files/a9ad08dfb8e982f84e72ba623f1ac1b049857d78#diff-759e11b0626a727a2deb0cfbe10103be7c565cffc8194fb3cb06a189920601b7R48), to make sure non-expandable sub-items are left-aligned with their parents.

src/internal/components/chart-series-details/index.tsx
jperals: Refactored [here](https://github.com/cloudscape-design/components/pull/1644/commits/42cb7b1cb1f3834d85fbd5b0bad23f7cd9222143)

src/internal/components/chart-series-details/styles.scss
jperals: Replaced by a variable [here](https://github.com/cloudscape-design/components/pull/1644/commits/2bd0f79899478e2344440aceb9589caa92aaddd5)

src/__tests__/__snapshots__/documenter.test.ts.snap
```diff
@@ -2156,6 +2156,25 @@ Do not use \`ariaLabel\` and \`ariaLabelledby\` at the same time.",
...
     },
+    Object {
+      "description": "Specifies customization of the content displayed in the chart popover.
+Use this for wrapping keys or values in links, or to display an additional
+level of nested items.",
+      "inlineType": Object {
+        "name": "MixedLineBarChartProps.DetailPopoverSeriesContent",
+        "parameters": Array [
+          Object {
+            "name": "__namedParameters",
```
just-boris, line 2167: Is this really expected to print it on our API documentation like this?

src/expandable-section/expandable-section-header.tsx
```diff
@@ -183,19 +182,29 @@ const ExpandableHeaderTextWrapper = ({
...
         <>
-          <HeadingTag
-            className={clsx(styles['header-wrapper'], headingTagListeners && styles['click-target'])}
-            {...headingTagListeners}
-          >
-            {headerButton}
-          </HeadingTag>
+          {actions ? (
+            <div className={clsx(styles['header-wrapper'], styles['header-wrapper-with-actions'])}>
+              <HeadingTag className={styles['click-target']} {...headingTagListeners}>
```
just-boris, line 194: Why is the `click-target` class added here without conditions but in the same code below it has a condition: `headingTagListeners && styles['click-target']`

Is there a potential for a bug? Any chances to reduce duplicated code?

src/internal/components/chart-series-details/index.tsx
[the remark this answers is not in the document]
just-boris: Such details need to be incapsulated in shared utility components. There is nothing chart specific, so our utility component should be responsible for that.

Do not add a tech debt by doing a custom version of some typical operation.

src/mixed-line-bar-chart/__integ__/detail-popover-series-content.test.ts
```diff
@@ -0,0 +1,65 @@
...
+describe('Detail popover series content', () => {
+  describe('keeps expanded state independently for every datum', () => {
+    test.each(['hover', 'click'])('on %s', interaction =>
+      setupTest('#/light/mixed-line-bar-chart/drilldown?expandableSubItems=true', async page => {
+        const textUnderFirstExpandableSection = 'AWS Config';
+        const textUnderSecondExpandableSection = 'Amazon Elastic Container Service';
+        const wrapper = createWrapper().findMixedLineBarChart();
+        let currentIndex: number | undefined;
+
+        const goToDatum = (index: number, interaction: string) => {
```
just-boris, line 15: Why does this code not follow our standard practices of creating a sub class for the base page object with page specific interaction methods?

src/mixed-line-bar-chart/__integ__/detail-popover-series-content.test.ts
```diff
@@ -0,0 +1,65 @@
...
+          const barGroup = wrapper.findBarGroups().get(index).toSelector();
+          if (interaction === 'hover') {
+            return page.hoverElement(barGroup);
+          } else {
+            // If a popover was pinned, we need to unpin it before pinning another one.
+            if (currentIndex) {
```
just-boris, line 21: Looking at this triple-nested code I am wondering if this code needs to have a test on a test. 

Any chances to simplify this code?

src/mixed-line-bar-chart/__integ__/detail-popover-series-content.test.ts
```diff
@@ -0,0 +1,65 @@
...
+            // intercepts the click and this is seen by Webdriver as an error and thrown as such, although it works for us
+            // (the component manages the event accordingly as coming from the corresponding bar group).
+            return page.buttonDownOnElement(barGroup);
+          }
+        };
+        const getPopoverHeaderText = () => page.getText(wrapper.findDetailPopover().findHeader().toSelector());
+        const getKeyValuePairsText = async () =>
+          (await page.getElementsText(wrapper.findDetailPopover().findSeries().toSelector())).join('\n');
+
+        page.setWindowSize({ width: 900, height: 800 });
```
just-boris, line 35: Missing await?

src/mixed-line-bar-chart/chart-popover.tsx
```diff
@@ -65,7 +68,26 @@ function MixedChartPopover<T extends ChartDataTypes>(
...
               onBlur={onBlur}
             >
-              <ChartSeriesDetails details={highlightDetails.details} />
+              <ChartSeriesDetails
+                key={highlightDetails.position}
+                details={highlightDetails.details}
+                setPopoverText={setPopoverText}
+                expandedSeries={expandedSeries[highlightDetails.position]}
+                setExpandedState={(index, isExpanded) =>
+                  setExpandedSeries(oldState => {
```
just-boris, line 77: Why are there no tests?

src/expandable-section/interfaces.ts
```diff
@@ -12,6 +12,8 @@ export namespace ExpandableSectionProps {
   export type HeadingTag = 'h1' | 'h2' | 'h3' | 'h4' | 'h5';
 }
 
+export type InternalVariant = ExpandableSectionProps.Variant | 'compact';
```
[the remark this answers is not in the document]
jperals: Discussed with the designer and the project team, the use case we see is this one which is supported via this API, and we don't currently see the need to support this variant standalone. We see a possibility that this variant could make sense to use in popovers, but we don't have enough data today to support that and we prefer to go for the two-day door option which is to keep the variant internal.

src/__tests__/__snapshots__/documenter.test.ts.snap
jperals: Thanks for spotting this. Refactored the interface to fix this [here](https://github.com/cloudscape-design/components/pull/1644/commits/8216ba7eddbb632e53666f6d770a2387a39e3fb3).

Note: unfortunately we don't currently have a way to show all the named parameters in the API documentation, instead we show the single parent parameter. A similar example of this is the `renderAriaLive` function of the Table component, see https://cloudscape.design/components/table/?tabId=api#properties

src/expandable-section/expandable-section-header.tsx
jperals: Refactored [here](https://github.com/cloudscape-design/components/pull/1644/commits/db54fea422a9bc0f6ae136b85905c5b659dfb85b), with the caveat that now there is an unconditional wrapping div even if there is nothing in the actions slot, but I guess the benefit of code simplicity overrules this?

src/mixed-line-bar-chart/__integ__/detail-popover-series-content.test.ts
jperals: Refactored [here](https://github.com/cloudscape-design/components/pull/1644/commits/3c067a09fbaca7e4e237a5c75e419b75cd4b3138)

src/mixed-line-bar-chart/__integ__/detail-popover-series-content.test.ts
jperals: Hopefully it is better to understand after the refactoring I mentioned above? [3c067a0](https://github.com/cloudscape-design/components/pull/1644/commits/3c067a09fbaca7e4e237a5c75e419b75cd4b3138)

src/mixed-line-bar-chart/__integ__/detail-popover-series-content.test.ts
jperals: Thanks for spotting, handled also in the commit mentioned above. [3c067a0](https://github.com/cloudscape-design/components/pull/1644/commits/3c067a09fbaca7e4e237a5c75e419b75cd4b3138)

src/mixed-line-bar-chart/chart-popover.tsx
jperals: Thanks, same as above, handled in [3c067a0](https://github.com/cloudscape-design/components/pull/1644/commits/3c067a09fbaca7e4e237a5c75e419b75cd4b3138)

src/mixed-line-bar-chart/__integ__/detail-popover-series-content.test.ts
```diff
@@ -0,0 +1,65 @@
...
+        let currentIndex: number | undefined;
+
+        const goToDatum = (index: number, interaction: string) => {
+          const barGroup = wrapper.findBarGroups().get(index).toSelector();
+          if (interaction === 'hover') {
+            return page.hoverElement(barGroup);
+          } else {
+            // If a popover was pinned, we need to unpin it before pinning another one.
+            if (currentIndex) {
+              page.click(wrapper.findDetailPopover().findDismissButton().toSelector());
```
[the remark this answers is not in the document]
jperals: Thanks, same as above, handled in [3c067a0](https://github.com/cloudscape-design/components/pull/1644/commits/3c067a09fbaca7e4e237a5c75e419b75cd4b3138)

src/mixed-line-bar-chart/chart-popover.tsx
jperals: Added unit test coverage [here](https://github.com/cloudscape-design/components/pull/1644/files#diff-cf210ece79a3b18b604dcbb61742c9c3daaa12afd61ac6d8720af4b917efcb4aR105).

Also covered with integration tests including hover behavior [here](https://github.com/cloudscape-design/components/pull/1644/files#diff-8193a37686e2c08dc27e3187fa2e6d21748c49c8412f28e7506fd808e06fe69aR41).

pages/line-chart/drilldown.page.tsx
```diff
@@ -0,0 +1,211 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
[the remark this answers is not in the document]
pan-kot: I do think that different scenarios must be considered like lines only or bars only - I would just scope those to the component where the feature is actually implemented.

src/bar-chart/__tests__/chart-series-details.test.tsx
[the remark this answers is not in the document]
pan-kot: On the other hand, everything possible with line- and bar- chart is also possible with the mixed chart, so it must be covered accordingly. If the mixed chart includes all scenarios then duplicating those in the different charts sounds unnecessary. I would only keep there tests that are relevant for the logic that is exclusive, e.g. test if the properties are propagated or that a chart-specific dev warning is fired. Should we add more exclusive features we should define new tests for those.

src/mixed-line-bar-chart/__tests__/test-chart-series-details.tsx
```diff
@@ -0,0 +1,229 @@
...
+          detailPopoverSeriesContent: ({ series, y }) => ({
+            key: <Link>{series.title}</Link>,
+            value: <Link>{y}</Link>,
+          }),
+        });
+
+        wrapper.findApplication()!.focus();
+        expect(warnOnce).toHaveBeenCalledWith(
+          'MixedLineBarChart',
+          'Use a ReactNode for the key or the value of a series, but not for both'
```
pan-kot, line 224: This might require additional explanation. The intention is to disallow links on both sides, but the ReactNode does not necessarily mean a link.

src/mixed-line-bar-chart/chart-container.tsx
```diff
@@ -372,6 +376,13 @@ export default function ChartContainer<T extends ChartDataTypes>({
       if (isPopoverPinned) {
         dismissPopover();
       } else {
+        // If the focus is on the document body when pinning the popover,
+        // Safari will focus any interactive element that may exist inside the popover body
+        // right *after* we focus the dismiss button.
+        // We prevent this here by focusing the chart plot before pinning the popover.
+        // For the purpose of this workaround, we could focus anything except the document body
+        // --the focus will move to the dismiss button once the popover appears in the next frame, anyway.
+        plotRef.current?.focusApplication();
```
pan-kot, line 385: Is there any reference why Safari does that? Can we do this change as a separate PR (we do already allow interactive content in the popover so the fix makes sense on its own it seems)?

src/internal/components/chart-series-details/index.tsx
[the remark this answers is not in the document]
jperals: `getSeriesDetailsText` is chart-specific. On the other hand, I'm thinking that it might make sense to make a separate PR with the changes related to live region from this PR and #1751, and unify how cartesian charts and pie charts generate its content.

src/popover/container.tsx
```diff
@@ -7,7 +7,7 @@ import { useResizeObserver } from '@cloudscape-design/component-toolkit/internal
```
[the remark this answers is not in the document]
jperals: I'll give this a try.

src/mixed-line-bar-chart/chart-popover.tsx
pan-kot: If I get this correctly, the expandedSeries keeps expanded states per X coordinate and series, like:

```
{
  'x=1': Set(0),
  'x=2': Set(),
  'x=3': Set(1) 
}
```

In the above example, the x=1, x=2, and x=3 are visited X coordinates and 0 and 1 are expanded series. Series 0 is expanded for x=1 only and series 1 is expanded for x=3 only.

I have two questions therefore:
1. Why do we decide to keep expanded state per x coordinate at all? When I drill-down on different coordinates it might not be very intuitive as of why sometimes a series is expanded.
2. What happens if the data is refreshed after the visit? The X coordinates are less concerning because those are tracked by formatted values. However, the series are tracked by indices. If the series order or structure change the state might not longer stay valid. This can easily happen if the series are filtered out with the default filter.

src/popover/container.tsx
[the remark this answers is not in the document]
jperals: Published PR: #1756, and removed those changes from this PR.

src/mixed-line-bar-chart/chart-popover.tsx
jperals: Thanks for the detailed feedback. To your questions:
1. This was the outcome after discussing with design. Personally I don't have a strong opinion.
2. Changed to use series title as discussed offline.

src/mixed-line-bar-chart/__tests__/test-chart-series-details.tsx
jperals: Done in [3c330ad](https://github.com/cloudscape-design/components/pull/1644/commits/3c330adc6f9a0204174c570c4693dc758977d55a).

src/mixed-line-bar-chart/chart-container.tsx
jperals: Opened a separate PR: https://github.com/cloudscape-design/components/pull/1784 and removed these changes from this one.

src/internal/components/chart-series-details/index.tsx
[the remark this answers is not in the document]
pan-kot: It might be a general improvement to the live region, not specific to charts. Say, if the passed html includes a list we can separate the entries with commas automatically instead of taking textContent from the entire thing.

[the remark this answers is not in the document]
jperals: One problem with lists is that they can be nested, and the `innerText` of the parent already includes the `innerText` of each child.

In order to overcome this here, I am using this specific selector in `getSeriesDetailsText`:

```
`.${styles.announced},.${expandabeleSectionHeaderStyles.header}`
```
where the `styles.announced` class is applied to all key-value pairs except to the ones of top-level items which have expandable sub-items. For these ones, `expandabeleSectionHeaderStyles.header` is used. I can't just add a wrapper around the `ExpandableSection` component because this would also include the children, so I would still have the same problem.

I think one way that such abstraction could work is by either passing either a selector or a list of elements to the function. WDYT?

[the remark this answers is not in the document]
pan-kot: Let's decouple this problem. I created a test page to visualise/test how live region extracts content: https://github.com/cloudscape-design/components/pull/1790

Let's use it with your example and try to get the desired announcement. We can either insert the invisible text (e.g. commas) to the list or modify the `extractInnerText` helper to analyse the nodes and make format the extracted content the way we want.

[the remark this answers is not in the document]
jperals: Note that this extracted value is not only used for live region but also for aria labels: [here](https://github.com/cloudscape-design/components/blob/00f52848e1a195893f817f619d0c9995de95e9ae/src/mixed-line-bar-chart/chart-container.tsx#L602) and [here](https://github.com/cloudscape-design/components/blob/00f52848e1a195893f817f619d0c9995de95e9ae/src/mixed-line-bar-chart/chart-container.tsx#L610). So I don't think this can be encapsulated in the live region. What we can do is make this function more generic as mentioned above, and let all charts use it (right now pie and donut charts don't, and use a different approach).

src/mixed-line-bar-chart/format-highlighted.ts
```diff
@@ -56,11 +71,40 @@ function getSeriesDetail<T>(internalSeries: InternalChartSeries<T>, targetX: T):
   if (isDataSeries(series)) {
     for (const datum of series.data) {
       if (matchesX(targetX, datum.x)) {
+        const customContent = detailPopoverSeriesContent
+          ? detailPopoverSeriesContent({ series, x: targetX, y: datum.y })
+          : undefined;
+        const hasSubItems = !!customContent?.subItems?.length;
+        const isExpandable = customContent?.expandable && hasSubItems;
+        const isKeyString = typeof customContent?.key === 'string';
+        const key = customContent?.key && (!isExpandable || isKeyString) ? customContent.key : series.title;
```
pan-kot, line 80: Can we just use the series title? What is the point in trying for the content key first?

pan-kot: But what do we lose if we just use the series.title?

jperals: This is passed to render the popover details, so we would lose the ability to render links in the key of the key-value pairs. Similar with `customContent.value` which contains the ReactNode value if `detailPopoverSeriesContent` was provided.

pan-kot: ah, got it now - thanks!

pan-kot (approved):

src/bar-chart/interfaces.ts
```diff
@@ -30,12 +29,6 @@ export interface BarChartProps<T extends ChartDataTypes>
    */
   horizontalBars?: boolean;
 
-  /**
-   * When set to `true`, adds a visual emphasis on the zero baseline axis.
-   * See the usage guidelines for more details.
-   */
-  emphasizeBaselineAxis?: boolean;
```
jperals, line 37: Moved to the new `CommonMixedChartProps` interface which contains the props from the mixed chart that are also shared by both line and bar charts. Those props are currently `emphasizeBaselineAxis` and `detailPopoverSeriesContent`.

pan-kot (approved):

PR #1646, opened 2023-10-12
chore: Apply funnel name data attribute to mobile breadcrumbs

michaeldowseza:
### Description

Desktop viewport breadcrumbs has a data attribute for to identify a potential funnel name. In a mobile viewport a different node is rendered and this data attribute was missing.

Changes made:
- Small refactor to name these at attributes related to funnels. This is done as the attributes are applied to the inner most text span instead of the root breadcrumb item element and it might have been confusing to not find your data attributes there.
- These specific funnel data attributes are now applied to both desktop and mobile versions of the breadcrumbs

### How has this been tested?

- Updated unit test

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/breadcrumb-group/__integ__/breadcrumb-group.test.ts
+++ b/src/breadcrumb-group/__integ__/breadcrumb-group.test.ts
@@ -31,12 +31,12 @@ class BreadcrumbGroupPage extends BasePageObject {
     return this.click(dropdownWrapper.findItems().get(index).toSelector());
   }
 }
-const setupTest = (testFn: (page: BreadcrumbGroupPage) => Promise<void>) => {
+const setupTest = (testFn: (page: BreadcrumbGroupPage, browser: WebdriverIO.Browser) => Promise<void>) => {
   return useBrowser(async browser => {
     const page = new BreadcrumbGroupPage(browser);
     await browser.url(`#/light/breadcrumb-group/events`);
     await page.waitForVisible(breadcrumbGroupWrapper.toSelector());
-    await testFn(page);
+    await testFn(page, browser);
   });
 };
 describe('BreadcrumbGroup', () => {
@@ -105,6 +105,15 @@ describe('BreadcrumbGroup', () => {
       await expect(page.isExisting(createWrapper().find(`.${styles['item-popover']}`).toSelector())).resolves.toBe(
         false
       );
+    })
+  );
+
+  test(
+    'Attachs funnel name attribute to last breadcrumb item',
+    setupTest(async (page, browser) => {
+      await page.setMobileViewport();
+      const funnelName = await browser.$('[data-analytics-funnel-key="funnel-name"]').getText();
+      expect(funnelName).toBe('Sixth that is very very very very very very long long long text');
     })
   );
 });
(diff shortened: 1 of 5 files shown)


src/breadcrumb-group/item/item.tsx
```diff
@@ -112,17 +114,14 @@ const BreadcrumbItemWithPopover = <T extends BreadcrumbGroupProps.Item>({
...
-const Item = ({ anchorAttributes, dataAttributes, children, isLast, ...itemAttributes }: ItemProps) =>
+const Item = ({ anchorAttributes, children, isLast, ...itemAttributes }: ItemProps) =>
   isLast ? (
-    <span {...itemAttributes} {...dataAttributes}>
-      {children}
-    </span>
+    <span {...itemAttributes}>{children}</span>
   ) : (
-    <a {...itemAttributes} {...anchorAttributes} {...dataAttributes}>
+    <a {...itemAttributes} {...anchorAttributes}>
```
michaeldowseza, line 124: The removed dataAttributes here were initially added by me for funnels but they were never used since we only add the attribute to the last item. Not sure why I added it, so removed it.

src/breadcrumb-group/__tests__/breadcrumb-item.test.tsx
```diff
@@ -30,6 +32,16 @@ const items = [
...
 describe('BreadcrumbGroup Item', () => {
+  let originalMatchMedia: typeof window['matchMedia'];
+
+  beforeAll(() => {
+    originalMatchMedia = window.matchMedia;
+  });
+
+  afterAll(() => {
+    window.matchMedia = originalMatchMedia;
+  });
```
michaeldowseza, line 43: This is also a bit of a "meh" cleanup attempt for this approach since all the tests passed without this being here. 

Is it needed? - ¯\_(ツ)_/¯. 
Do I feel better knowing the original function is restored - yes.

src/breadcrumb-group/__integ__/breadcrumb-group.test.ts
```diff
@@ -31,12 +31,12 @@ class BreadcrumbGroupPage extends BasePageObject {
     return this.click(dropdownWrapper.findItems().get(index).toSelector());
   }
 }
-const setupTest = (testFn: (page: BreadcrumbGroupPage) => Promise<void>) => {
+const setupTest = (testFn: (page: BreadcrumbGroupPage, browser: WebdriverIO.Browser) => Promise<void>) => {
```
jperals, line 34: Is this second parameter necessary? The `browser` used in line 39 comes from line 35, right?

michaeldowseza: Yes - The browser object in Line 39 comes from line 35. Line 34 is the type definition of testFn (Line 39).

src/breadcrumb-group/__integ__/breadcrumb-group.test.ts
```diff
@@ -107,4 +107,13 @@ describe('BreadcrumbGroup', () => {
       );
     })
   );
+
+  test(
+    'Attachs funnel name attribute to last breadcrumb item',
+    setupTest(async (page, browser) => {
+      await page.setMobileViewport();
```
jperals, line 114: Why only test this on mobile viewport?

michaeldowseza: I added a desktop viewport test as a unit test as its possible there. Mobile viewport can't be tested there because of the necessary css breakpoint conditions.

jperals: Well, I guess the unit test below covers the general case. No strong opinion here.

PR #1647, opened 2023-10-12
feat: Add ariaLabel and ariaLabelledby to progress bar

avinashbot:
### Description

Does what it says on the tin. People could provide visual labels through the `label` property, but it turns out in some contexts, like inside a table cell, the visual label itself doesn't provide enough context. So additional programmatic context needs to be added. This allows for that.

Related links, issue #, if available: AWSUI-22557

### How has this been tested?

Unit tests!

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/progress-bar/internal.tsx
+++ b/src/progress-bar/internal.tsx
@@ -20,9 +20,10 @@ const clamp = (value: number, lowerLimit: number, upperLimit: number) => {
 interface ProgressProps {
   value: number;
   isInFlash: boolean;
-  labelId: string;
+  ariaLabel?: string;
+  ariaLabelledby?: string;
 }
-export const Progress = ({ value, isInFlash, labelId }: ProgressProps) => {
+export const Progress = ({ value, isInFlash, ariaLabel, ariaLabelledby }: ProgressProps) => {
   const roundedValue = Math.round(value);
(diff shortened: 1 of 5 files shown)


src/progress-bar/internal.tsx
```diff
@@ -36,7 +37,9 @@ export const Progress = ({ value, isInFlash, labelId }: ProgressProps) => {
         )}
         max={MAX_VALUE}
         value={progressValue}
-        aria-labelledby={labelId}
+        aria-label={ariaLabel}
+        // Ensures aria-label takes precedence over aria-labelledby
```
gethinwebster, line 41: can we just pass the attributes through and let browsers handle this?

avinashbot: We spotted recently that `aria-labelledby` overrides `aria-label` if present - and it's particularly notable in this case because users may provide a visual label (which is linked using `aria-labelledby`) and might want to provide a different aria label.

So even if they don't provide an `ariaLabelledby` explicitly, they might still be bitten by this strange behavior. So manually doing this is the most "unsurprising" and expected behavior.

gethinwebster: ah of course, got it!

PR #1649, opened 2023-10-13
refactor: Remove column widths sync from table sticky header

pan-kot:
### Description

Table's header and sticky header render the same Thead component responsible for assigning column widths. When resizable columns feature is used there is no need to additionally sync the widths because there is no difference between them.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/table/__tests__/columns-width.test.tsx
+++ b/src/table/__tests__/columns-width.test.tsx
@@ -194,28 +194,51 @@ describe('with stickyHeader=true', () => {
     window.CSS.supports = originalFn;
   });
 
-  test('mirrors column width on sticky header copy, but not minWidth/maxWidth', () => {
-    const extractSize = (thead: ElementWrapper) =>
-      thead.findAll('tr > *').map(column => {
-        const { minWidth, maxWidth, width } = column.getElement().style;
-        return { minWidth, maxWidth, width };
-      });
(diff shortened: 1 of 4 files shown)


just-boris:
I would not call it unnecessary, check this page: https://d21d5uik3ws71m.cloudfront.net/components/86c5f63d33dc2c2022cbd17d0f064f08c7f43126/index.html#/light/table/sticky-header

image


After the change, the header and column sizes do not match.

pan-kot:
> I would not call it unnecessary, check this page: https://d21d5uik3ws71m.cloudfront.net/components/86c5f63d33dc2c2022cbd17d0f064f08c7f43126/index.html#/light/table/sticky-header
> 
> image
> After the change, the header and column sizes do not match.

Great observation! :) I didn't realise the sync is necessary when the resizable columns are not used. I will submit another revision to make the change properly scoped and add an extra unit test.

pan-kot:
> There are still broken cases (which work in mainline)
> 
>  Screen.Recording.2023-10-13.at.14.55.32.mov 
> Why do you dislike this code so much? ;) why is it important for you to remove it?

Thanks, I will take a look. The reason I dislike that code is because it includes the "auto" width condition. I would like that one to be figured out by the resizable columns helper exclusively.

pan-kot:
Closing due to high complexity.

PR #1650, opened 2023-10-13
fix: Consider container paddings in table calculations

pan-kot:
### Description

In VR styles container can have paddings (those are actually applied to the table element). The paddings must be considered when compared container width against table width - otherwise the isScrollable state is applied 40px later. It also causes the last column resizer to not auto-scroll the container.

Before fix (see how the last column is getting smaller than its min-width before container scroll applies):

https://github.com/cloudscape-design/components/assets/20790937/d7808de6-1a73-478a-bf50-fbcf7380ebbf

After fix:

https://github.com/cloudscape-design/components/assets/20790937/3e08a81d-6412-46d4-b995-0ba10e684a53

### How has this been tested?

Added an integration test for regression.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/table/resizable-columns-rounding.page.tsx
+++ b/pages/table/resizable-columns-rounding.page.tsx
@@ -39,11 +39,11 @@ const items: Array<Item> = [
 ];
 
 export default function () {
-  const [width, setWidth] = useState(609);
+  const [width, setWidth] = useState(649);
   return (
     <>
       <h1>Table in container with special size </h1>
-      <button id="shrink-container" onClick={() => setWidth(595)}>
+      <button id="shrink-container" onClick={() => setWidth(635)}>
         Resize container
       </button>
       <div style={{ width }}>
--- a/src/table/internal.tsx
+++ b/src/table/internal.tsx
@@ -1,7 +1,7 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 import clsx from 'clsx';
-import React, { useCallback, useImperativeHandle, useMemo, useRef } from 'react';
+import React, { useCallback, useImperativeHandle, useRef } from 'react';
 import { TableForwardRefType, TableProps } from './interfaces';
 import { getVisualContextClassname } from '../internal/components/visual-context';
 import InternalContainer, { InternalContainerProps } from '../container/internal';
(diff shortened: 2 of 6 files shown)


src/table/internal.tsx
```diff
@@ -119,6 +119,14 @@ const InternalTable = React.forwardRef(
     const [tableWidth, tableMeasureRef] = useContainerQuery<number>(rect => rect.contentBoxWidth);
     const tableRefObject = useRef(null);
 
+    const containerContentWidth = useMemo(() => {
```
pan-kot, line 122: Ideally is to avoid these computations in the code. In theory, we can apply the paddings to the container itself in which case its measured content width is going to be correct. However, that requires complex adjustments applied to other styles and also resizable- and sticky columns features.

pages/table/resizable-columns-rounding.page.tsx
```diff
@@ -39,11 +39,11 @@ const items: Array<Item> = [
 ];
 
 export default function () {
-  const [width, setWidth] = useState(609);
```
pan-kot, line 42: Adjusted the values by 40px which is the sum of table paddings.

src/table/internal.tsx
```diff
@@ -119,6 +119,14 @@ const InternalTable = React.forwardRef(
...
+      if (!containerWidth || !tableRefObject.current) {
+        return null;
+      }
+      const tableStyle = getComputedStyle(tableRefObject.current);
```
just-boris, line 126: Here you render the computed value of the previous frame, not your actual state. There could be all sorts of issues, let's avoid this path.

You can only read DOM inside `useEffect` hook

just-boris: P.S. if you need to account paddings, can we just target a different element in our container query 🤔

pan-kot: Yes, that is true. The canonical way would be that have a useEffect and a piece of state which is I would expect going to impact performance. I can explain the possible issues in the comments (changing Classic -> VR, changing table's variant) or come up with a better solution. Would you recommend some?

just-boris: Already provided my suggestion: https://github.com/cloudscape-design/components/pull/1650#discussion_r1358239770

pan-kot: Ah, see it now. I don't think querying a different element would work. First of all - the paddings are applied to the table and not the container itself and changing that is very tricky. Second, the container width with paddings is also used for setting the empty row width.

just-boris: In any way, the approach in the PR is a no go. We can discuss alternatives, but in a different setting

pan-kot: I updated the approach

src/table/styles.scss
```diff
@@ -68,7 +68,8 @@
   scrollbar-width: none; /* Hide scrollbar in Firefox */
   &.variant-stacked,
   &.variant-container {
```
just-boris, line 70: What happens in the full page variant? If all works as expected, should this be reflected in the tests?

pan-kot: It works just fine because there are no extra paddings. Added a test for this variant as well.

just-boris (approved):

PR #1656, opened 2023-10-16
fix: Prevent Flashbar from delaying render when not using animations

jperals:
### Description

This fixes two issues:
- Ticket AWSUI-22525: The collapsible flashbar was making use of TransitionGroup even when motion is disabled, although without defining any animation in that case, so there would not be any perceivable issue for users, but since TransitionGroup delays the removal of items by one frame, this would break tests that remove items. This is fixed by ~~adopting the same approach as the [non-collapsible flashbar](https://github.com/cloudscape-design/components/blob/main/src/flashbar/non-collapsible-flashbar.tsx#L46), with separate list rendering for motion enabled and motion disabled, where TransitionGroup is only used when motion is enabled.~~ replacing the wrapping TransitionGroup with a React fragment when motion is disabled.
- An additional issue was found and fixed by working on this: the logic for scrolling the expand/collapse button when collapsing the flashbar was only called when motion is enabled.

### How has this been tested?

- Added unit tests to cover flash dismissal for collapsible and non-collapsible flashbar (with motion disabled)
- Reorganized unit tests so that as much as possible of the collapsible flashbar functionality is tested with motion enabled and disabled (this uncovered the second issue mentioned above)
- Manually tested on local dev pages, with motion enabled and disabled, and compared with mainline deployment

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/flashbar/__tests__/flashbar.test.tsx
```diff
@@ -6,29 +6,9 @@ import Flashbar, { FlashbarProps } from '../../../lib/components/flashbar';
 import Button from '../../../lib/components/button';
 import createWrapper from '../../../lib/components/test-utils/dom';
 import styles from '../../../lib/components/flashbar/styles.css.js';
-import { createFlashbarWrapper, findList } from './common';
+import { createFlashbarWrapper, findList, testFlashDismissal } from './common';
 import { DATA_ATTR_ANALYTICS_FLASHBAR } from '../../../lib/components/internal/analytics/selectors';
-
-let mockUseAnimations = false;
```
jperals, line 12: Simplified disabling motion in test by using our own official method instead of mocking

src/flashbar/collapsible-flashbar.tsx
```diff
@@ -93,6 +93,10 @@ export default function CollapsibleFlashbar({ items, ...restProps }: FlashbarPro
         focusFlashById(ref.current, mostRecentItem.id);
       }
     }
+    // When collapsing, scroll up if necessary to avoid losing track of the focused button
+    else if (wasFlashbarStackExpanded && !isFlashbarStackExpanded && notificationBarRef.current) {
+      scrollElementIntoView(notificationBarRef.current);
```
jperals, line 98: By moving this here, this will be called regardless of motion being enabled or not. Before, it was under a condition `if (initialAnimationState)`, and `initialAnimationState` is only initialized if motion is enabled.

src/flashbar/__tests__/collapsible.test.tsx
```diff
@@ -45,297 +46,317 @@ const defaultProps = {
...
-    });
-
-    it('shows toggle element with desired text', () => {
-      const customToggleButtonText = 'Custom text';
-      const flashbar = renderFlashbar({
-        i18nStrings: {
-          notificationBarText: customToggleButtonText,
-        },
+  for (const withAnimations of [false, true]) {
+    describe(withAnimations ? 'with animations' : 'without animations', () => {
```
just-boris, line 50: Could save one indentation level by using `describe.each([true, false])('withAnimations=%s')`

src/flashbar/__tests__/collapsible.test.tsx
```diff
@@ -45,297 +46,317 @@ const defaultProps = {
...
+        });
+      });
     });
+  }
+
+  test('dismisses items', () => {
+    // Test this feature only without animations because TransitionGroup delays item removals by one frame.
+    // Customers should disable animations in their tests too:
+    // https://cloudscape.design/foundation/visual-foundation/motion/#implementation
+    disableMotion(true);
```
just-boris, line 358: If you manipulate with motion inside a single test, should the `beforeAll` hook convert to `beforeEach` to ensure the clean state between tests?

src/flashbar/__tests__/flashbar.test.tsx
just-boris: Nice

src/flashbar/collapsible-flashbar.tsx
just-boris: 1. Let's not use dnd-kit code in unrelated use-cases. We have our own version of it by the way: https://github.com/cloudscape-design/components/blob/2a073a233c1fd0704d55cb043269fab4d2a90819/src/internal/hooks/use-previous/index.ts#L8
2. I think, this code does not need this value at all, because the hook already depends on `isFlashbarStackExpanded`, so if `isFlashbarStackExpanded === false` it automatically means it was `true` before. If you want to skip initial render, there is `useEffectOnUpdate` which is more idiomatic approach to solve this: https://github.com/cloudscape-design/components/blob/2a073a233c1fd0704d55cb043269fab4d2a90819/src/internal/hooks/use-effect-on-update.ts#L6

src/flashbar/collapsible-flashbar.tsx
```diff
@@ -188,106 +187,164 @@ export default function CollapsibleFlashbar({ items, ...restProps }: FlashbarPro
...
-      style={
-        !isFlashbarStackExpanded || transitioning
-          ? {
-              [customCssProps.flashbarStackDepth]: stackDepth,
-            }
-          : undefined
-      }
-    >
-      <TransitionGroup component={null}>
+  const renderListWithoutAnimations = () => {
```
just-boris, line 190: This is too much code copy paste isn't it? It can get our of sync very easily

Any options how reduce this duplication?

src/flashbar/collapsible-flashbar.tsx
jperals: Oh, imported from the wrong place, thanks for spotting this.

Ideally I'd like to keep using a layout effect instead of a "normal" effect to prevent flickering, i.e, a frame where the flashbar collapsed but the scrolling hasn't happened yet.

pages/flashbar/interactive.page.tsx
```diff
@@ -13,7 +13,8 @@ export default function InteractiveFlashbar() {
   };
 
   const add = (type: FlashbarProps.Type, hasHeader = false) => {
-    setItems(items => [generateItem({ type, id: nextId.current.toString(), dismiss, hasHeader }), ...items]);
```
jperals, line 16: This makes sure that the new item has the correct id and not the next one. Since React batches state updates, the old logic could lead to two items having the same id.

src/flashbar/collapsible-flashbar.tsx
jperals: Fixed the import

src/flashbar/collapsible-flashbar.tsx
jperals: Grouped the common props together

src/flashbar/__tests__/collapsible.test.tsx
jperals: Sorry I missed this comment. This test is outside the `describe` block that contains the `beforeAll` hook, so this call should not affect the tests in that block.

just-boris: This code contains a hidden danger, why not to resolve it by converting beforeAll to beforeEach

src/flashbar/collapsible-flashbar.tsx
jperals: Because `useEffectOnUpdate` uses `useEffect` and not `useLayoutEffect`, the scroll is delayed to the next render. See test failure: https://github.com/cloudscape-design/components/actions/runs/6576775598/job/17866925600

src/flashbar/__tests__/collapsible.test.tsx
jperals: Changed

src/flashbar/__tests__/collapsible.test.tsx
jperals: Changed

src/flashbar/collapsible-flashbar.tsx
jperals: I added a test which compares the HTML output. Note:
- For this I needed the test to remove some a11y-related attributes that are not relevant to the visual result, because they have dynamic values
- Only for the collapsed state. For the expanded state, it gets trickier since we do need to add classes for the motion case, for the animations to work

If we want to be even more precise here, my suggestion would be to consider actually making screenshot tests with motion enabled too for this component.

src/flashbar/collapsible-flashbar.tsx
just-boris: `findNotificationBar().click()` uses vanilla `HTMLElement.click` function instead of operating via our test utils.

Use our test utils and I expect it will work fine

src/flashbar/__tests__/collapsible.test.tsx
```diff
@@ -174,169 +185,185 @@ describe('Collapsible Flashbar', () => {
...
+    testFlashDismissal({ stackItems: true });
+  });
+
+  test('produces the same visual HMTL output with motion enabled and disabled', () => {
```
jperals, line 360: This test is less relevant after last commit, but it helped me catch that if using `<TransitionGroup>` without `component={null}`, it adds a wrapping div. This could actually break styles, eventually.

src/flashbar/collapsible-flashbar.tsx
```diff
@@ -369,3 +368,6 @@ const NotificationTypeCount = ({
     </span>
   );
 };
+
+const ListWrapper = ({ children, withMotion }: { children: ReactNode; withMotion: boolean }) =>
+  withMotion ? <TransitionGroup component={null}>{children}</TransitionGroup> : <>{children}</>;
```
jperals, line 373: Added this small component because TransitionGroup adds an extra div by default unless `component` is null, but using this same prop in React.Fragment triggers runtime (not build time) errors.

An alternative would be to have this inline inside the main component:
```
const ListWrapper = isReducedMotion ? TransitionGroup : React.Fragment;
(...)
<ListWrapper {...isReducedMotion && { component: null } }>
  (...)
</ListWrapper>
```
Equally valid, just a bit uglier IMO.

src/flashbar/__tests__/collapsible.test.tsx
just-boris: When using with `removeInvisibleAttributes` the test is not very trust worthy.

Feel free to keep it, but I do not find it very useful

src/flashbar/__tests__/flashbar.test.tsx
```diff
@@ -44,409 +36,411 @@ afterEach(() => {
...
+  useVisualRefresh = state;
+  disableMotion(!state);
+}
 
-  for (const withAnimations of [false, true]) {
-    describe(withAnimations ? 'with animations' : 'without animations', () => {
-      beforeEach(() => {
-        useAnimations = withAnimations;
-      });
+describe.only('Flashbar component', () => {
```
just-boris, line 44:
```suggestion
describe('Flashbar component', () => {
```

src/flashbar/collapsible-flashbar.tsx
just-boris: Fine by me

just-boris (commented):
`describe.only` is important to fix, I will approve after this

src/flashbar/__tests__/flashbar.test.tsx
jperals: Thanks for spotting, removed

jperals:
> When using with removeInvisibleAttributes the test is not very trust worthy.
> Feel free to keep it, but I do not find it very useful

Removed it

just-boris (approved):

PR #1669, opened 2023-10-18
fix: Cache dropdown's overflow parents for better jsdom performance

jperals:
### Description

`getOverflowParents` is expensive on testing environments because we are calling `getComputedStyle` on every element in the parent chain and Jest does not yet incorporate a version of jsdom that includes [this fix](https://github.com/jsdom/jsdom/pull/3482). This causes some packages to have test timeouts after #1555.

This PR handles this by caching the result of `getOverflowParents` in the dropdown component. Note that it does not cache their dimensions, which will still be recalculated as necessary.

Ticket: AWSUI-22613

### How has this been tested?

- Successful dry-run build to previously failing package
- Successfully ran tests in previously failing package, and observed that the run time is reduced significantly (~3x).

<details>
   <summary>Review checklist</summary>

_The following items are to be evaluated by the author(s) and the reviewer(s)._

#### Correctness

- _Changes include appropriate documentation updates._
- _Changes are backward-compatible if not indicated, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#public-apis)._
- _Changes do not include unsupported browser features, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#browsers-support)._
- _Changes were manually tested for accessibility, see [accessibility guidelines](https://cloudscape.design/foundation/core-principles/accessibility/)._

#### Security

- _If the code handles URLs: all URLs are validated through [the `checkSafeUrl` function](https://github.com/cloudscape-design/components/blob/main/src/internal/utils/check-safe-url.ts)._

#### Testing

- _Changes are covered with new/existing unit tests?_
- _Changes are covered with new/existing integration tests?_
</details>

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1679, opened 2023-10-20
fix: Adjust dropdown width only on resize

jperals:
### Description

Alternative fix to https://github.com/cloudscape-design/components/pull/1669, without caching anything.

Ticket: AWSUI-22613

Note: dropdowns do not expand beyond the trigger width when virtual scroll is enabled. This is a known issue: AWSUI-22357

### How has this been tested?

- Successful dry-run build to same failing package as in https://github.com/cloudscape-design/components/pull/1669. The build time is similar (11m 47s then, 12m 35s now)
- Manually tested against existing page `dropdown/width` (which also has [integration tests](https://github.com/cloudscape-design/components/blob/main/src/internal/components/dropdown/__integ__/width.test.ts))

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/components/dropdown/index.tsx
+++ b/src/internal/components/dropdown/index.tsx
@@ -23,6 +23,7 @@ import TabTrap from '../tab-trap/index.js';
 import { getFirstFocusable, getLastFocusable } from '../focus-lock/utils.js';
 import { useUniqueId } from '../../hooks/use-unique-id/index.js';
 import customCssProps from '../../generated/custom-css-properties';
+import { useResizeObserver } from '@cloudscape-design/component-toolkit/internal';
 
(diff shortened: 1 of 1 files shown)


avinashbot (approved):

just-boris (approved):
The fix is still very weird and suboptimal, but I guess I cannot get anything better than this

just-boris:
The correct fix would be to fix this condition in the first place: https://github.com/cloudscape-design/components/blob/0c5e691a1bed9e316c38a4dc446dc9c7f92710b3/src/internal/components/dropdown/index.tsx#L107

The current fix is a patch that may explode in any moment, because there are two connected pieces of logic live in very different files and lifecycles

PR #1680, opened 2023-10-20
fix: Close chart popover when focus moves away

jperals:
### Description

Since custom content is supported in chart popovers via `detailPopoverFooter`, it is possible to insert interactive elements, such as links and buttons, in a chart popover. In such a case, it is possible to navigate into a chart popover with the keyboard, even without necessarily pinning it. However, there were a couple issues when navigating inside the popover without pinning it:

- In bar, line, mixed line-bar and pie charts, the popover would stay open when navigating away. This can be especially problematic if the popover is tall, since it could be covering focused legend items below.
- In area charts, it was not possible to navigate inside the popover by tabbing. When doing so, the popover would close.

This PR fixes these two issues and establishes one single behavior for all graphs as required by a11y: allow to navigate in and close when navigating out. This behavior mimics the mouse behavior.

Doc: `GVf7Asm4sCUs/Initial-a11y-review-for-Chart-drill-down-features#temp:C:EVOe4154480d00441e78e4cdb80e`

### How has this been tested?

- Added integration tests
- Manually tested

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/area-chart/model/use-chart-model.ts
+++ b/src/area-chart/model/use-chart-model.ts
@@ -16,6 +16,7 @@ import { throttle } from '../../internal/utils/throttle';
 import { useReaction } from '../async-store';
 import { useHeightMeasure } from '../../internal/hooks/container-queries/use-height-measure';
 import { useStableCallback } from '@cloudscape-design/component-toolkit/internal';
+import { nodeBelongs } from '../../internal/utils/node-belongs';
 
 const MAX_HOVER_MARGIN = 6;
 const SVG_HOVER_THROTTLE = 25;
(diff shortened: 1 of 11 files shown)


src/area-chart/model/use-chart-model.ts
```diff
@@ -273,10 +274,10 @@ export default function useChartModel<T extends AreaChartProps.DataTypes>({
     };
 
     // A callback for svg blur to clear all highlights unless the popover is pinned.
-    const onSVGBlur = () => {
+    const onSVGBlur = (event: React.FocusEvent<Element>) => {
       // Pinned popover stays pinned even if the focus is lost.
       // If blur is not caused by the popover, forget the previously highlighted point.
-      if (!interactions.get().isPopoverPinned) {
+      if (!nodeBelongs(containerRef.current, event.relatedTarget) && !interactions.get().isPopoverPinned) {
```
jperals, line 280: This fixes not being able to navigate into the popover. In that situation we blur the SVG but not the popover, so e check if we blur the container which contains both the SVG and the popover. This makes it equivalent to the behavior of the mixed line-bar chart [here](https://github.com/cloudscape-design/components/blob/main/src/mixed-line-bar-chart/chart-container.tsx#L393) and the pie chart [here](https://github.com/cloudscape-design/components/blob/main/src/pie-chart/pie-chart.tsx#L263).

abdhalees (approved):

PR #1683, opened 2023-10-24
fix: Scroll first selected [multi-]select option into view when opening

gethinwebster:
### Description

Restore old behavior (also similar to native controls) where the current item will be scrolled into view when opening a select/multiselect dropdown

Related links, issue #, if available: AWSUI-22621

### How has this been tested?

New unit tests, local testing, screenshot testing for regressions, and new screenshot tests (to be added in separate review)

<details>

#### Correctness

#### Security

#### Testing

</details>

---


pages/select/select.test.page.tsx
```diff
@@ -70,87 +71,89 @@ export default function SelectPage() {
...
-          <Box variant="p">Dropdown should not be rendered beyond the container with overflow hidden</Box>
-          <div id="smallest_container" style={{ overflow: 'hidden', height: '500px', padding: 0 }}>
-            <div style={{ overflow: 'hidden', height: '900px' }}>
-              <Select
-                id="select_overflow"
-                placeholder="Choose option"
-                selectedOption={null}
-                options={optionsOverflow}
-              />
+      <ScreenshotArea>
```
just-boris, line 74: Why did you decide to convert this page into screenshot testing, when we already have another dedicated one: https://github.com/cloudscape-design/components/blob/main/pages/select/screenshot.page.tsx

src/internal/components/options-list/utils/use-open-state.ts
```diff
@@ -9,10 +9,12 @@ interface OpenStateProps {
 
 export const useOpenState = ({ onOpen, onClose }: OpenStateProps) => {
   const [isOpen, setIsOpen] = useState(false);
+  const [openedWithKeyboard, setOpenedWithKeyboard] = useState(false);
 
-  const openDropdown = () => {
+  const openDropdown = (isKeyboard?: boolean) => {
```
just-boris, line 14: Do we have any instances where we do not know the origin of the action? 

I think for the integrity of our code it would be better to have this always explicitly defined

src/internal/components/options-list/utils/use-highlight-option.ts
```diff
@@ -65,12 +67,13 @@ export function useHighlightedOption<OptionType>({
   return [
     { highlightType, highlightedIndex, highlightedOption },
```
just-boris, line 68: Since we already return an object here (some sort of `highlightedState`) can we include the new field on the top level rather than introducing a nested object

just-boris (approved):
Approved, but I still find these code style suggestions useful to address

src/internal/components/options-list/utils/use-open-state.ts
gethinwebster: good point, will update

src/internal/components/options-list/utils/use-highlight-option.ts
gethinwebster: the existing object properties get passed around to other components as individual properties not the full object, so it feels to me cleaner to pass it around in this way rather than creating an additional property on many subcomponents

pages/select/select.test.page.tsx
gethinwebster: to avoid the risk of adding a completely new scenario to an existing page (as none of the selects on that page have enough items to be suitable for this test) impacting the correctness & reliability of those existing tests

just-boris (approved):

PR #1688, opened 2023-10-25
fix: Fix last item of collapsible flashbar not being dismissible

jperals:
### Description

The class `flash-with-motion` makes a flashbar flash enter or leave with animations, based on `.enter` and `.leave` suffixes that React Transition adds when the element enters or leaves: see https://github.com/cloudscape-design/components/blob/main/src/flashbar/motion.scss.

In #1656, I merged the logic to apply this class in the collapsible and non-collapsible versions of the flashbar but it turns out, there is a subtle difference:  in the non-collapsible flashbar these animations only play in Visual Refresh, but in the collapsible flashbar they are supposed to play not only in Visual Refresh but also in Classic, and it should stay like that.

Disabling React-Transition-based enter and leave animations for Classic had the following unintended consequences in Classic:

1. The enter animation did not look as intended
2. Even worse problem: dismissed items occasionally staying in the DOM. This is because our internal Transitions component [relies](https://github.com/cloudscape-design/components/blob/main/src/internal/components/transition/index.tsx#L46) on `transitionend` events, which are note emitted if the `.flashbar-with-motion` class, which [defines](https://github.com/cloudscape-design/components/blob/main/src/flashbar/motion.scss) the transition delays, is not present.

### How has this been tested?

In the `flashbar/interactive` page, followed these steps:
- Remove all initial flashes
- Enable "stack items"
- Add one flash
- Dismiss it and verify that it disappears

Ticket: AWSUI-23010

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/flashbar/collapsible-flashbar.tsx
+++ b/src/flashbar/collapsible-flashbar.tsx
@@ -45,7 +45,7 @@ export default function CollapsibleFlashbar({ items, ...restProps }: FlashbarPro
     setInitialAnimationState(rects);
   }, [getElementsToAnimate]);
 
-  const { animateFlash, baseProps, breakpoint, isReducedMotion, isVisualRefresh, mergedRef, ref } = useFlashbar({
+  const { baseProps, breakpoint, isReducedMotion, isVisualRefresh, mergedRef, ref } = useFlashbar({
     items,
     ...restProps,
     onItemsAdded: newItems => {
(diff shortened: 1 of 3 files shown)


abdhalees (approved):

just-boris:
This change should include an automated test. Even if you want to merge this quickly to fix the issue, you need to follow up with the test later

jperals:
> This change should include an automated test. Even if you want to merge this quickly to fix the issue, you need to follow up with the test later

Acknowledged. I'll leave the ticket open until I have the tests.

PR #1694, opened 2023-10-26
fix: Cartesian charts support long left labels

pan-kot:
### Description

Chart labels should not be prioritised over the graphic. The fix introduces max width to the left labels of cartesian charts.

### How has this been tested?

Screenshot tests (the test page added, to be tested in the dev. pipeline)

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/mixed-line-bar-chart/with-long-left-labels.page.tsx
+++ b/pages/mixed-line-bar-chart/with-long-left-labels.page.tsx
@@ -37,7 +37,7 @@ export default function () {
                 case 'Potatoes':
                   return 'Potatoes (fried, ~400g)';
                 case 'Chocolate':
-                  return 'Chocolate (none was left)';
+                  return 'Chocolate (delicious milk chocolate slice)\n(comment: none was left)';
                 case 'Apples':
                   return 'Apples (1 apple)';
                 case 'Oranges':
--- a/pages/mixed-line-bar-chart/with-long-left-labels.page.tsx
+++ b/pages/mixed-line-bar-chart/with-long-left-labels.page.tsx
@@ -3,7 +3,6 @@
 import React from 'react';
 
 import Box from '~components/box';
-import Button from '~components/button';
 import MixedLineBarChart from '~components/mixed-line-bar-chart';
 import { colorChartsThresholdInfo } from '~design-tokens';
 import ScreenshotArea from '../utils/screenshot-area';
@@ -32,7 +31,6 @@ export default function () {
           xScaleType="categorical"
           ariaLabel="Dinner chart"
           ariaDescription={barChartInstructions}
-          detailPopoverFooter={xValue => <Button>Filter by {xValue}</Button>}
           xTickFormatter={value =>
             (() => {
               switch (value) {
--- a/pages/mixed-line-bar-chart/with-long-left-labels.page.tsx
+++ b/pages/mixed-line-bar-chart/with-long-left-labels.page.tsx
@@ -1,0 +1,56 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import React from 'react';
+
+import Box from '~components/box';
+import Button from '~components/button';
+import MixedLineBarChart from '~components/mixed-line-bar-chart';
+import { colorChartsThresholdInfo } from '~design-tokens';
+import ScreenshotArea from '../utils/screenshot-area';
+
+import { barChartInstructions, commonProps, data3, data4 } from './common';
+
+export default function () {
+  return (
+    <ScreenshotArea>
+      <Box variant="h1">Mixed chart with long left labels</Box>
+      <Box padding="l">
+        <MixedLineBarChart
+          {...commonProps}
+          horizontalBars={true}
+          hideFilter={true}
+          id="chart"
+          height={250}
+          series={[
+            { title: 'Happiness', type: 'bar', data: data4.filter(({ x }) => x !== 'Chocolate') },
+            { title: 'Threshold', type: 'threshold', y: 420, color: colorChartsThresholdInfo },
+          ]}
+          xDomain={data3.map(d => d.x)}
+          yDomain={[0, 650]}
+          xTitle="Dinner"
+          yTitle="Calories (kcal)"
+          xScaleType="categorical"
+          ariaLabel="Dinner chart"
+          ariaDescription={barChartInstructions}
+          detailPopoverFooter={xValue => <Button>Filter by {xValue}</Button>}
+          xTickFormatter={value =>
+            (() => {
+              switch (value) {
+                case 'Potatoes':
+                  return 'Potatoes (fried, ~400g)';
+                case 'Chocolate':
+                  return 'Chocolate (none was left)';
+                case 'Apples':
+                  return 'Apples (1 apple)';
+                case 'Oranges':
+                  return 'Oranges (fresh orange juice, 300ml)';
+                default:
+                  return value;
+              }
+            })()
+          }
+        />
+      </Box>
+    </ScreenshotArea>
+  );
+}
--- a/src/area-chart/chart-container.tsx
+++ b/src/area-chart/chart-container.tsx
@@ -78,7 +78,7 @@ function ChartContainer<T extends AreaChartProps.DataTypes>({
 }: ChartContainerProps<T>) {
   const [leftLabelsWidth, setLeftLabelsWidth] = useState(0);
   const [containerWidth, containerWidthRef] = useContainerWidth(DEFAULT_CHART_WIDTH);
-  const maxLeftLabelsWidth = Math.round((containerWidth ?? 0) / 2);
+  const maxLeftLabelsWidth = Math.round(containerWidth / 2);
 
   const bottomLabelsProps = useBottomLabels({
     ticks: model.computed.xTicks,
--- a/src/area-chart/chart-container.tsx
+++ b/src/area-chart/chart-container.tsx
@@ -125,7 +125,7 @@ function ChartContainer<T extends AreaChartProps.DataTypes>({
           ticks={model.computed.yTicks}
           tickFormatter={yTickFormatter as TickFormatter}
           autoWidth={setLeftLabelsWidth}
-          maxWidth={maxLeftLabelsWidth}
+          maxLabelsWidth={maxLeftLabelsWidth}
         />
       }
       bottomAxisLabel={<AxisLabel axis="x" position="bottom" title={xTitle} />}
@@ -162,14 +162,14 @@ function ChartContainer<T extends AreaChartProps.DataTypes>({
           />
 
           <LeftLabels
-            width={model.width}
-            height={model.height}
+            plotWidth={model.width}
+            plotHeight={model.height}
             scale={model.computed.yScale}
             ticks={model.computed.yTicks}
             tickFormatter={yTickFormatter}
             title={yTitle}
             ariaRoleDescription={yAxisAriaRoleDescription}
-            maxWidth={maxLeftLabelsWidth}
+            maxLabelsWidth={maxLeftLabelsWidth}
           />
 
           <AreaDataSeries model={model} />
--- a/src/area-chart/chart-container.tsx
+++ b/src/area-chart/chart-container.tsx
@@ -78,6 +78,7 @@ function ChartContainer<T extends AreaChartProps.DataTypes>({
 }: ChartContainerProps<T>) {
   const [leftLabelsWidth, setLeftLabelsWidth] = useState(0);
   const [containerWidth, containerWidthRef] = useContainerWidth(DEFAULT_CHART_WIDTH);
+  const maxLeftLabelsWidth = Math.round((containerWidth ?? 0) / 2);
 
   const bottomLabelsProps = useBottomLabels({
     ticks: model.computed.xTicks,
@@ -124,6 +125,7 @@ function ChartContainer<T extends AreaChartProps.DataTypes>({
           ticks={model.computed.yTicks}
           tickFormatter={yTickFormatter as TickFormatter}
           autoWidth={setLeftLabelsWidth}
+          maxWidth={maxLeftLabelsWidth}
         />
       }
       bottomAxisLabel={<AxisLabel axis="x" position="bottom" title={xTitle} />}
@@ -167,6 +169,7 @@ function ChartContainer<T extends AreaChartProps.DataTypes>({
             tickFormatter={yTickFormatter}
             title={yTitle}
             ariaRoleDescription={yAxisAriaRoleDescription}
+            maxWidth={maxLeftLabelsWidth}
           />
 
           <AreaDataSeries model={model} />
--- a/src/internal/components/cartesian-chart/__tests__/left-labels.test.tsx
+++ b/src/internal/components/cartesian-chart/__tests__/left-labels.test.tsx
@@ -1,0 +1,40 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import React from 'react';
+import { render } from '@testing-library/react';
+import LeftLabels from '../../../../../lib/components/internal/components/cartesian-chart/left-labels';
+import { getVisibleTicks } from '../../../../../lib/components/internal/components/cartesian-chart/label-utils';
+import { NumericChartScale } from '../../../../../lib/components/internal/components/cartesian-chart/scales';
+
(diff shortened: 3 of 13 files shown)


src/area-chart/chart-container.tsx
```diff
@@ -78,6 +78,7 @@ function ChartContainer<T extends AreaChartProps.DataTypes>({
 }: ChartContainerProps<T>) {
   const [leftLabelsWidth, setLeftLabelsWidth] = useState(0);
   const [containerWidth, containerWidthRef] = useContainerWidth(DEFAULT_CHART_WIDTH);
+  const maxLeftLabelsWidth = Math.round((containerWidth ?? 0) / 2);
```
jperals, line 81: `containerWidth` is a number, so I believe `?? 0` is not necessary

pan-kot: It is `null | number`

jperals: `useContainerWidth` returns `[width: number, ref: React.Ref<HTMLDivElement>]`. Anyway this is not critical

pan-kot: Ah, sure thing! I didn't even noticed that is not useContainerQuery we have in there which produce a nullable value. Thanks!

pages/mixed-line-bar-chart/with-long-left-labels.page.tsx
```diff
@@ -0,0 +1,56 @@
...
+          {...commonProps}
+          horizontalBars={true}
+          hideFilter={true}
+          id="chart"
+          height={250}
+          series={[
+            { title: 'Happiness', type: 'bar', data: data4.filter(({ x }) => x !== 'Chocolate') },
+            { title: 'Threshold', type: 'threshold', y: 420, color: colorChartsThresholdInfo },
+          ]}
+          xDomain={data3.map(d => d.x)}
```
jperals, line 28: It works because both have the same x values, but maybe better to use the same data, either `data3` or `data4` in the `series` and in the `xDomain`?

pages/mixed-line-bar-chart/with-long-left-labels.page.tsx
```diff
@@ -0,0 +1,56 @@
...
+          yDomain={[0, 650]}
+          xTitle="Dinner"
+          yTitle="Calories (kcal)"
+          xScaleType="categorical"
+          ariaLabel="Dinner chart"
+          ariaDescription={barChartInstructions}
+          detailPopoverFooter={xValue => <Button>Filter by {xValue}</Button>}
```
jperals, line 35: This prop is not necessary here, we can remove complexity

pages/mixed-line-bar-chart/with-long-left-labels.page.tsx
pan-kot: That is a copy from the test.page.tsx. If we change the structure of the test data then all pages need to be adjusted, so I think it is find for now.

src/area-chart/chart-container.tsx
```diff
@@ -167,6 +169,7 @@ function ChartContainer<T extends AreaChartProps.DataTypes>({
             tickFormatter={yTickFormatter}
             title={yTitle}
             ariaRoleDescription={yAxisAriaRoleDescription}
+            maxWidth={maxLeftLabelsWidth}
```
jperals, line 172: I think `width` is the width of the tick line (which spans the entire chart width when the labels correspond to the y axis —therefore the width of the chart container itself, the right side) whereas the new `maxWidth` corresponds to the width reserved for the labels (the left side). Can we make the naming more clear to avoid confusion here?

pan-kot: Yes, I though about it too. We have the same problem with other cartesian chart components potentially but I can change it for the left labels and measure at least.

src/internal/components/cartesian-chart/left-labels.tsx
```diff
@@ -42,15 +45,17 @@ function LeftLabels({
...
-      return cacheRef.current[label];
+    if (labelToHeightCache.current[label] !== undefined) {
+      return labelToHeightCache.current[label];
     }
     if (virtualTextRef.current && virtualTextRef.current.getBBox) {
       virtualTextRef.current.textContent = label;
-      cacheRef.current[label] = virtualTextRef.current.getBBox().height;
-      return cacheRef.current[label];
+      labelToHeightCache.current[label] = virtualTextRef.current.getBBox().height;
+      labelToWidthCache.current[label] = virtualTextRef.current.getBBox().width;
```
jperals, line 57: No need to call `virtualTextRef.current.getBBox()` twice. Could `labelToHeightCache` and `labelToWidthCache` be merged in the same object? Maybe we can just cache the raw boundingBox?

src/internal/components/cartesian-chart/responsive-text.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+  return (
+    <text ref={textRef} x={x} y={y} style={{ textAnchor: 'end' }} className={className}>
+      {children}
+    </text>
+  );
+}
+
+function renderTextContent(textNode: SVGTextElement, text: string, visibleWidth: number) {
+  let visibleLength = text.length;
```
jperals, line 31: I find this name a bit confusing, would be more like `maxLength` or `maxVisibleLength`

src/internal/components/cartesian-chart/left-labels.tsx
```diff
@@ -87,14 +90,25 @@ function LeftLabels({
...
+                  x2={plotWidth}
                   y2={0}
                   aria-hidden="true"
                 />
               )}
-              <text className={styles.ticks__text} x={-(TICK_LENGTH + TICK_MARGIN)} y={0}>
-                {lines.join(' ')}
-              </text>
+
+              {lines.map((line, lineIndex) => {
```
pan-kot, line 99: Turns out, the left cartesian chart labels never were multiline although the collision detection is supported for that case. Made it multiline to make the truncation target individual lines same as in the pie chart. It is now also possible to control the wrapping on the consumer end.

src/internal/components/cartesian-chart/responsive-text.tsx
pan-kot: It is not max but current visible length. We start from the full length and make it smaller incrementally until finding the max fitting length. However, it becomes the max fitting length only at the last iteration.

src/internal/components/cartesian-chart/left-labels.tsx
```diff
@@ -42,15 +45,15 @@ function LeftLabels({
 
   const yOffset = axis === 'x' && scale.isCategorical() ? Math.max(0, scale.d3Scale.bandwidth() - 1) / 2 : 0;
 
-  const cacheRef = useRef<{ [label: string]: number }>({});
+  const labelToBoxCache = useRef<{ [label: string]: DOMRect }>({});
```
jperals, line 48: Maybe this logic could be somehow contained inside `formatTicks`? We could pass it `virtualTextRef.current` instead of `getLabelSpace` and this component would be a bit cleaner

pan-kot: then we'd also need to pass down the cache and repeat the same for the bottom labels

src/internal/components/cartesian-chart/label-utils.ts
```diff
@@ -80,3 +80,12 @@ function hasIntersection(a: FormattedTick, b: FormattedTick) {
   const rightEdge = right.position - right.space / 2;
   return leftEdge > rightEdge;
 }
+
+/* istanbul ignore next */
```
jperals, line 84: I'd add a comment to justify why part of the code can be uncovered

src/internal/components/cartesian-chart/left-labels.tsx
jperals: The box information (or just the width if that's the only part we use) of each line could come with each item returned from `getVisibleTicks`?

pan-kot: It could, but feels like a bigger refactoring that is better to be done separately. I would probably apply the formatting first with line-splitting and then in a separate function measure and cache the widths and heights.

src/internal/components/cartesian-chart/left-labels.tsx
```diff
@@ -87,14 +90,25 @@ function LeftLabels({
...
                 />
               )}
-              <text className={styles.ticks__text} x={-(TICK_LENGTH + TICK_MARGIN)} y={0}>
-                {lines.join(' ')}
-              </text>
+
+              {lines.map((line, lineIndex) => {
+                const lineTextProps = {
+                  x: -(TICK_LENGTH + TICK_MARGIN),
+                  y: lineIndex * TICK_LINE_HEIGHT,
```
jperals, line 102: This will render every line below each other starting from the original first line, but shouldn't it be better to center them vertically? I also did this [here](https://github.com/cloudscape-design/components/pull/1691/files#diff-cac1f0bfda43aa85aac8be6bb1fffe700f614c938b433eff37ae7c6ab83a0aa9R100) and I centered them by adding `- (lines.length - 1) * TICK_LINE_HEIGHT * 0.5`.

```
y={lineIndex * TICK_LINE_HEIGHT - (lines.length - 1) * TICK_LINE_HEIGHT * 0.5}
```
or it could be simplified a bit
```
y={TICK_LINE_HEIGHT * (lineIndex - (lines.length - 1) * 0.5)}
```

src/internal/components/responsive-text/responsive-text-utils.ts
```diff
@@ -0,0 +1,7 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+/* istanbul ignore next */
```
jperals, line 4: Same here, I would add a comment

src/internal/components/cartesian-chart/left-labels.tsx
pan-kot: This requires a design decision, I like the right alignment better. We can also check how it works for bottom labels where the multi-line is supported.

src/internal/components/cartesian-chart/__tests__/left-labels.test.tsx
```diff
@@ -0,0 +1,76 @@
...
+  default: () => <text>...</text>,
+}));
+
+describe('cartesian chart left labels', () => {
+  beforeEach(() => {
+    (getVisibleTicks as any).mockClear();
+  });
+
+  test('all ticks are visible', () => {
+    (getVisibleTicks as any as jest.Mock).mockReturnValueOnce([
```
just-boris, line 29:
```suggestion
    jest.mocked(getVisibleTicks).mockReturnValueOnce([
```

This helper changes nothing in runtime but makes the code more readable

src/internal/components/cartesian-chart/__tests__/left-labels.test.tsx
```diff
@@ -0,0 +1,76 @@
...
+jest.mock('../../../../../lib/components/internal/components/cartesian-chart/label-utils', () => ({
+  ...jest.requireActual('../../../../../lib/components/internal/components/cartesian-chart/label-utils'),
+  getVisibleTicks: jest.fn().mockReturnValue([]),
+  getLabelBBox: jest.fn().mockReturnValue({ width: 0, height: 0 }),
+}));
+
+jest.mock('../../../../../lib/components/internal/components/responsive-text', () => ({
+  __esModule: true,
+  default: () => <text>...</text>,
+}));
```
just-boris, line 21: Is it really necessary to mock everything here?

The more things you mock, the less authentic the tests become

just-boris (commented):
You can count my comments as non-blocking, but I tried to raise the code quality a bit here

src/internal/components/cartesian-chart/left-labels.tsx
jperals: > I like the right alignment better

But I'm talking about _vertically_ centering, so items can still be anchored to the right. [Here](https://d21d5uik3ws71m.cloudfront.net/components/5b3f734fa9edd6ba4cdc3e4108999a1e8a6d038d/index.html#/light/bar-chart/left-label-wrap) is a deployed example of how that looks like.

src/internal/components/cartesian-chart/label-utils.ts
pan-kot: I prefer ignoring the lines that are not going to be covered due to the limitations like jsdom support or, say, invariant violation exceptions. I am happy to have less than 100% coverage but I think it is great when the missing lines shown in the report represent the missing opportunity.

src/internal/components/cartesian-chart/left-labels.tsx
pan-kot: Oh, I din't realise that. That's great 👍  Feel free to push a commit to the PR.

src/internal/components/responsive-text/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+  }, [maxWidth, children]);
+
+  return (
+    <text ref={textRef} x={x} y={y} style={{ textAnchor: 'end' }} className={className}>
+      {children}
+    </text>
+  );
+}
+
+export function renderTextContent(textNode: SVGTextElement, text: string, visibleWidth: number) {
```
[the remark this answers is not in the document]
pan-kot: That code was copied over from the pie chart, that's why. I will change it

src/internal/components/cartesian-chart/label-utils.ts
just-boris: Even in this case, you can still narrow it down to a specific condition instead of marking the whole function

src/internal/components/cartesian-chart/label-utils.ts
```diff
@@ -80,3 +80,12 @@ function hasIntersection(a: FormattedTick, b: FormattedTick) {
...
+export function getLabelBBox(element: null | SVGTextElement, label: string) {
+  if (element && element.getBBox) {
+    element.textContent = label;
```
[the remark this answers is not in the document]
pan-kot: Thanks, will apply the second recommendation

src/internal/components/cartesian-chart/__tests__/left-labels.test.tsx
pan-kot: True, but the ResponsiveText relies on the API which needs to be mocked as well. By mocking the component I am just making sure the ResponsiveText and not <text> is getting rendered. The implementation is of no significance.

src/internal/components/cartesian-chart/left-labels.tsx
jperals: Done :)

jperals:
In the latest deployment, the truncation is not happening anymore: https://d21d5uik3ws71m.cloudfront.net/components/e63c053ed1556d0209c8a4ce148d5bff94a52493/index.html#/light/mixed-line-bar-chart/with-long-left-labels

src/internal/components/cartesian-chart/label-utils.ts
just-boris: Any news on this thread?

If you wanted to ignore only JSDOM branch, it should be this way

```js
function getSVGTextSize(element: null | SVGTextElement): undefined | { width: number; height: number } {
  /* istanbul ignore if  */
  if (element && element.getBBox) {
    return element.getBBox();
  }
  return undefined;
}

```

pan-kot: Thanks, I am familiar with the syntax. The problem is that it will only ignore the line inside of the IF not ignoring the function itself and the two remaining lines. That is enough to make the diff coverage check unhappy. At the same time, the entire function is meant to be mocked. I can add a test for the function alone to make it covered but that does not seem that necesary.

just-boris: So, you created a problem for yourself by choosing too aggressive mocking strategy: https://github.com/cloudscape-design/components/pull/1694#discussion_r1373437193

Ignoring this is not fine, it needs to be solved properly. I recommend revisit the mocks, they are hiding too much

pan-kot: This function returns the BBox of an element, what do we win if get it covered?

just-boris: What do we win from marking it as ignored?

jperals: The last statement `return undefined` is not necessary. Would removing that line help?

Or alternatively, write it in a single line:

```
return element && element.getBBox ? element.getBBox() : undefined;
```

jperals (approved):

PR #1696, opened 2023-10-27
fix: Consider container paddings in table calculations (2)

pan-kot:
### Description

This is the re-introduction of https://github.com/cloudscape-design/components/pull/1650 that was reverted due to a regression in empty state.

The sticky empty state cell requires its width to be explicitly set as container width plus table paddings. Previously the paddings were always included and not those are have to be taken care of explicitly.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/table/internal.tsx
+++ b/src/table/internal.tsx
@@ -39,7 +39,7 @@ import { useCellEditing } from './use-cell-editing';
 import { LinkDefaultVariantContext } from '../internal/context/link-default-variant-context';
 import { CollectionLabelContext } from '../internal/context/collection-label-context';
 import { useFunnelSubStep } from '../internal/analytics/hooks/use-funnel';
-import { EmptyStateCell } from './empty-state-cell';
+import { NoDataCell } from './node-data-cell';
 
 const SELECTION_COLUMN_WIDTH = 54;
 const selectionColumnId = Symbol('selection-column-id');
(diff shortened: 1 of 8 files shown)


gethinwebster:
Is the empty state part covered by tests in any way?

src/table/internal.tsx
```diff
@@ -363,26 +364,16 @@ const InternalTable = React.forwardRef(
...
-                          {loading ? (
-                            <InternalStatusIndicator type="loading" className={styles.loading} wrapText={true}>
-                              <LiveRegion visible={true}>{loadingText}</LiveRegion>
-                            </InternalStatusIndicator>
-                          ) : (
-                            <div className={styles.empty}>{empty}</div>
-                          )}
-                        </div>
-                      </td>
+                      <EmptyStateCell
```
gethinwebster, line 367: [minor naming] I don't think it's ideal that `EmptyStateCell` also renders the loading state, perhaps something like `FullTableCell` or `NoDataCell` instead (not sure either is ideal though, open to other suggestions!)

pan-kot:
> Is the empty state part covered by tests in any way?

Yes, it has a dedicated screenshot test.

src/table/internal.tsx
pan-kot: I like the `NodeDataCell` a lot - will update 👍

just-boris (approved):

PR #1698, opened 2023-10-27
fix: Fixes focus controls in drawers

katiegeorge:
### Description

Part of Drawers Bug Bash.

This simplifies and fixes the drawer focus controls to use the same hook as tools, and also adds a test to make sure focus works as expected with overflow and on-page triggers.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/app-layout/drawer/index.tsx
```diff
@@ -182,12 +182,17 @@ export const DrawerTriggersBar = ({
...
   drawers,
+  drawerRefs,
   onDrawerChange,
 }: DrawerTriggersBarProps) => {
   const containerRef = React.useRef<HTMLDivElement>(null);
+  const previousActiveDrawerId = useRef(activeDrawerId);
   const [containerHeight, triggersContainerRef] = useContainerQuery(rect => rect.contentBoxHeight);
   const isCompactMode = useDensityMode(containerRef) === 'compact';
 
+  if (activeDrawerId) {
```
avinashbot, line 193: I'm not sure what this code is doing, could you explain it? Why assign a ref only if the ref was previously active?

src/app-layout/interfaces.ts
```diff
@@ -220,6 +220,11 @@ export namespace AppLayoutProps {
      */
     focusToolsClose(): void;
 
+    /**
+     * Focuses the active drawer. Use this to focus the active drawer after opening it programmatically.
+     */
+    focusActiveDrawerClose(): void;
```
avinashbot, line 226: Non-blocking, but I guess the name's not really accurate anymore, is it? For resizable drawers, we focus on the slider, so maybe it should be closer to split panel's `focusActiveDrawer`? There was a fair bit of discussion about naming when it came to split panel too so... I'll leave it up to you.

katiegeorge: Good point, I updated it to just `focusActiveDrawer`.

src/app-layout/drawer/index.tsx
katiegeorge: We do the same thing [here](https://github.com/cloudscape-design/components/blob/main/src/app-layout/visual-refresh/drawers.tsx#L184), with the comment: 
> Changes to the activeDrawerId need to be tracked by the previousActiveDrawerId property in order to appropriately apply the ref required to manage focus control.

katiegeorge: I think its basically to make sure the focus control works property on drawer change, not just open / close. When I just checked, I realized I did have to add a new dependency to the focus hook for when the active id changes

src/app-layout/utils/use-focus-control.ts
```diff
@@ -41,9 +51,8 @@ export function useFocusControl(isOpen: boolean, restoreFocus = false): FocusCon
     shouldFocus.current = false;
   };
 
-  // We explictly want this effect to run when only `isOpen` changes
   // eslint-disable-next-line react-hooks/exhaustive-deps
-  useEffect(doFocus, [isOpen]);
+  useEffect(doFocus, [isOpen, dependencies]);
```
avinashbot, line 55: You probably want this, or else it'll fire on every render because the array is a new object each time:

```suggestion
  useEffect(doFocus, [isOpen, ...dependencies]);
```

But passing dependencies through hooks isn't ideal, and React recommends against it. Maybe it's a sign that we shouldn't be doing useEffect in here and instead have the calling component call `setFocus` when its own dependencies change.

just-boris: This code is a generic support for "when switching from non-resizable drawer to resizable the resize handle should be focused"

Do you have any suggestions about this high level behavior? Where should the focus go when switching between already open drawers?

just-boris: +1 to that comment. I think we can avoid changing this effect all together. Let's do programmatic focus inside drawer toggle click handler

src/app-layout/drawer/index.tsx
just-boris: Looks like this code is intended to do the same job as this already existing code: https://github.com/cloudscape-design/components/blob/e9fbf2367717b27300bc11c194ca66a60b3f0af2/src/app-layout/utils/use-focus-control.ts#L34-L37

Can we leverage the existing `loseFocus` feature somehow?

src/app-layout/utils/use-focus-control.ts
katiegeorge: What do you mean by this?

just-boris: We can keep `useFocusControl` unchanged and do `focusToolsButtons(true)` when we switch drawers

src/app-layout/drawer/index.tsx
katiegeorge: It's different, I believe. This section of the code only gets run when programmatically opening a drawer, not when changing drawers via triggers. Its working as expected now, let's keep it this way and we can revisit later if we need.

src/app-layout/utils/use-focus-control.ts
katiegeorge: From DMs: explicitly passing down the `activeDrawerId` instead of dependencies here.

just-boris (approved):

avinashbot (approved):

PR #1707, opened 2023-10-31
chore: Add a direction setting to the url parameters for internal pages.

DaemonCahill:
This PR adds a direction setting to the url parameters for internal pages. This URL parameter must be set manually, and requires a page refresh in order to see the document changes take effect.

Screenshot 2024-01-03 at 5 03 45 PM


jperals:
We have (at least) two integration tests that rely on clicking directly on the checkboxes (in code editor and table). Those tests will fail if this change is merged as is, unless they are changed accordingly as well. Also:
- If the main reason for grouping the settings is the BackstopJS visual tests, is it not possible to set up BackstopJS to exclude the top bar?
- If we decide to change to the settings UX, would it be better to use a less invasive component such as a button dropdown, so that it is possible to better see the effect on the page while changing the settings?

pages/app/app-context.tsx
```diff
@@ -67,6 +69,8 @@ export function AppContextProvider({ children }: { children: React.ReactNode })
   const { mode, pageId } = match ? match.params : { mode: undefined, pageId: undefined };
   const urlParams = parseQuery(location.search) as AppUrlParams;
 
+  document.querySelector('html')?.setAttribute('dir', urlParams.direction);
```
just-boris, line 72: This is React. 

In React you do not do such thing bluntly inside the render function. Even if it seems working, it will cause you a hard time dealing with issues later

just-boris:
@jperals the previous header implementation has a problem that header height is not static. The more items we add to the header, the more it wraps on mobile.

This becomes a problem, when you suddenly need to fix irrelevant tests in app layout, dropdown and other full-screen cases.

This is why we wrap the settings into a modal, to prevent this from happening.

pages/app/app-context.tsx
DaemonCahill: Per offline discussion we are moving forward with a page refresh requirement, and following the implementation for visual refresh: `document.body.classList.toggle('awsui-visual-refresh', visualRefresh);`

DaemonCahill:
> If we decide to change to the settings UX, would it be better to use a less invasive component such as a button dropdown or popover, so that it is possible to better see the effect on the page while changing the settings?

@jperals Good recommendation, I will update the UI accordingly.

DaemonCahill:
@jperals I updated the demo video to show the Popover use. @just-boris The Direction implementation has been updated to match the Visual Refresh implementation. Should the integration test updates be handled within the scope of this PR, or in a follow up by someone else?

pages/app/components/theme-switcher.tsx
```diff
@@ -26,44 +26,72 @@ export default function ThemeSwitcher() {
...
-          checked={urlParams.motionDisabled}
-          onChange={event => setUrlParams({ motionDisabled: event.target.checked })}
-        />
-        Disable motion
-      </label>
-    </SpaceBetween>
+    <Popover
+      position="bottom"
+      size="large"
+      content={
```
just-boris, line 32: From our usage gudelines

> Always use a header when the popover can be dismissed.

We need to add a header

pages/app/components/theme-switcher.tsx
```diff
@@ -26,44 +26,72 @@ export default function ThemeSwitcher() {
...
+        <SpaceBetween direction="vertical" size="xs">
+          <label>
+            Theme
+            <select defaultValue={THEME}>
+              <option value={THEME}>{THEME}</option>
+            </select>
+          </label>
```
just-boris, line 39: This select does not work anymore, because there is only one option. Time to remove it

pages/app/components/theme-switcher.tsx
```diff
@@ -26,44 +26,72 @@ export default function ThemeSwitcher() {
...
+
+          <label>
+            Direction
+            <select
+              defaultValue={urlParams.direction}
+              onChange={event => {
+                setUrlParams({ direction: event.target.value as 'rtl' | 'ltr' });
+                window.location.reload();
```
jperals, line 47: Why do we need to reload?

DaemonCahill: There will be a subset of components (Dropdown, Popover) that will need to detect directionality at runtime and apply styles through the CSSOM. If one of these components is in an open state and the document direction is changed they will not respond dynamically without a mutation observer. This is something of an edge case that we can revisit, but after discussion with @just-boris we decided to proceed with the same approach as Visual Refresh to start.

pages/app/components/theme-switcher.tsx
DaemonCahill: Added a header!

pages/app/components/theme-switcher.tsx
DaemonCahill: Removed!

jperals:
> We have (at least) two integration tests that rely on clicking directly on the checkboxes (in code editor and table)

Here I was referring to internal pipeline integration tests but I see that 10 integration tests from this package are failing as well ([link](https://github.com/cloudscape-design/components/actions/runs/6724978026/job/18278634961?pr=1707)). Not sure about the former ones but I'd expect the latter ones to be fixed as part of this PR. I'd say this is a good opportunity to refactor how these integration tests change the settings and make this a bit more robust.

DaemonCahill:
@just-boris @jperals I discussed this PR internally and to unblock myself I reduce the scope to simply adding support for a URL parameter. We can discuss further refactoring of the dev pages with the rest of the team, but that is outside the scope of what is required for Phase 1 of RTL support.

just-boris (approved):

PR #1711, opened 2023-11-01
chore: Increase offset between link text and underline

avinashbot:
### Description

Check the internal research doc, the people love a heavy offset. This won't fit into the existing variant map, so I had to create a font-size map. And to prevent `text-decoration` from overriding all the little sub-properties, I switched to `text-decoration-line` which is supported by every browser in our support matrix.

Related links, issue #, if available: AWSUI-22009

### How has this been tested?

It's a styling change, and I expect the one big permutation page to collect all the differences.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/link/constants.scss
+++ b/src/link/constants.scss
@@ -9,7 +9,7 @@
 
 $link-font-sizes: (
   'body-s': (
-    'text-underline-offset': 0.25em,
+    'text-underline-offset': 0.3em,
     'text-decoration-thickness': 1px,
   ),
   'body-m': (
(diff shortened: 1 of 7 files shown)


src/link/constants.scss
```diff
@@ -7,13 +7,52 @@
 @use '../internal/styles/typography/constants.scss' as styles;
 @use '../internal/styles/tokens' as awsui;
 
+$link-font-sizes: (
+  'body-s': (
+    'text-underline-offset': 0.3em,
```
timogasda, line 12: Should we convert this to px?

avinashbot: We decided to stay with `em` on purpose here, not only because it's a similar standard value for all font sizes, but also because it scales nicer if customers decide to theme font sizes.

just-boris (approved):

PR #1715, opened 2023-11-03
fix: DateRangerPicker with hh timeInputFormat

michaeldowseza:
Normalizes returned value when timeInput is of the format hh

### Description

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/date-range-picker/__tests__/date-range-picker-absolute.test.tsx
+++ b/src/date-range-picker/__tests__/date-range-picker-absolute.test.tsx
@@ -239,6 +239,108 @@ describe('Date range picker', () => {
         expect(wrapper.findDropdown()!.findEndDateInput()!.findNativeInput().getElement()).toHaveValue('2021/03/17');
         expect(wrapper.findDropdown()!.findEndTimeInput()!.findNativeInput().getElement()).toHaveValue('23:59:59');
       });
+
+      (['hh:mm:ss', 'hh:mm', 'hh'] as DateRangePickerProps['timeInputFormat'][]).forEach(timeInputFormat => {
+        test(`sets start and end time to the beginning and end of day when only selecting a date with format ${timeInputFormat}`, () => {
+          const onChangeSpy = jest.fn();
+          const { wrapper } = renderDateRangePicker({
+            ...defaultProps,
+            timeInputFormat,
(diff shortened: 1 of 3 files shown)


src/date-range-picker/__tests__/date-range-picker-absolute.test.tsx
```diff
@@ -241,6 +241,34 @@ describe('Date range picker', () => {
       });
     });
 
+    describe('time formats', () => {
+      test('correctly handles a time format of hh', () => {
```
just-boris, line 245: How about adding a test for `hh:mm` too? To demonstrate consistency between modes

src/date-range-picker/__tests__/date-range-picker-absolute.test.tsx
```diff
@@ -241,6 +241,34 @@ describe('Date range picker', () => {
...
+        const onChangeSpy = jest.fn();
+        const { wrapper } = renderDateRangePicker({
+          ...defaultProps,
+          timeInputFormat: 'hh',
+          onChange: event => onChangeSpy(event.detail),
+        });
+
+        act(() => wrapper.findTrigger().click());
```
just-boris, line 253: Test utils already call `act` when needed. No need to do it one more time

src/date-range-picker/__tests__/date-range-picker-absolute.test.tsx
michaeldowseza: Added for all time formats 👍🏻

src/date-range-picker/__tests__/date-range-picker-absolute.test.tsx
michaeldowseza: Removed from my tests. The entire file wraps everything in act though. I can do a follow up PR to remove it from the rest as well.

just-boris: Yes we have a lot of old code, pre-dating this change: https://github.com/cloudscape-design/test-utils/pull/25

Any cleanups would be nice, easy wins there

just-boris (approved):
Is approving a draft PR even possible?

PR #1718, opened 2023-11-07
fix: Last table resizable column stable width in React 18

pan-kot:
### Description

In React 18 when the table container changes width the last column can change its width style from "auto" to "{width}px" and back. However, the change happens with a slight delay. As result, there is a frame when the width is set incorrectly: when the column is set to "auto" but there is not enough space, the actual column width can be too small which makes the content wrap. The content wrapping can increase the table's height by a lot for that little period of time.

ref: AWSUI-22655

### How has this been tested?

* New unit test for resize observer part.
* New integration tests to secure real and sticky headers to have the same sizes.
* Screenshot tests (before merging).
* Manual testing with React 16 and React 18.

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/anchor-navigation/use-scroll-spy.tsx
```diff
@@ -37,7 +37,7 @@ export default function useScrollSpy({
   }, [hrefs]);
 
   // Get the bounding rectangle of an element by href
-  const getRectByHref = useCallback(href => {
+  const getRectByHref = useCallback((href: string) => {
```
just-boris, line 40: Ideally we should remove useCallback from our code, not fixing it. But okay

src/table/use-column-widths.tsx
```diff
@@ -80,6 +82,36 @@ export function ColumnWidthsProvider({ visibleColumns, resizableColumns, childre
     }
   };
 
+  // Update getContainer when deps change to trigger new observation.
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const getContainer = useCallback(() => containerRef.current, [resizableColumns, visibleColumns, columnWidths]);
```
just-boris, line 87: `visibleColumns` is not a stable value, so all your memoization efforts are futile.

Acceptable for this PR, because some other code already uses it too, but you just made it harder to fix when it becomes a real problem

src/table/use-column-widths.tsx
```diff
@@ -80,6 +82,36 @@ export function ColumnWidthsProvider({ visibleColumns, resizableColumns, childre
...
+  useResizeObserver(getContainer, entry => {
+    if (!resizableColumns) {
+      return;
+    }
+
+    setContainerWidth(entry.contentBoxWidth);
+
+    const lastColumnId = visibleColumns[visibleColumns.length - 1]?.id;
+    const lastCellElement = lastColumnId ? cellsRef.current[lastColumnId] : null;
+    if (!lastCellElement) {
```
just-boris, line 101: How is this possible? How can you have a table without the last column?

src/table/use-column-widths.tsx
```diff
@@ -80,6 +82,36 @@ export function ColumnWidthsProvider({ visibleColumns, resizableColumns, childre
...
+      return;
+    }
+
+    const totalWidth = visibleColumns.reduce((total, { id }) => total + (columnWidths[id] || DEFAULT_COLUMN_WIDTH), 0);
+    const stretchLastColumn = Boolean(entry.contentBoxWidth > totalWidth);
```
just-boris, line 106: What does this `Boolean()` do?

src/table/use-column-widths.tsx
```diff
@@ -80,6 +82,36 @@ export function ColumnWidthsProvider({ visibleColumns, resizableColumns, childre
...
+
+    if (stretchLastColumn && lastCellElement.style.width !== 'auto') {
+      lastCellElement.style.width = 'auto';
```
just-boris, line 109: If we moved this code here, it probably should be removed in the place where it was before. So we will not have race conditions to fragments of code doing the same.

src/table/use-column-widths.tsx
pan-kot: It have to be in the dependency array nevertheless. If the getContainer was stable, the onObserve function would capture the variables just once and never update those again.

src/table/use-column-widths.tsx
pan-kot: It is technically possible to have a table with no columns. This can be the case for example when the columns configuration is resolved asynchronously.

src/table/use-column-widths.tsx
pan-kot: In this case - nothing, will remove it. It was copied from the similar line below where the conversion is required to get the correct type.

src/table/use-column-widths.tsx
just-boris: 👀 it is first time I hear about this use-case. Do we have any unit tests confirming that our table works in this setup?

src/table/use-column-widths.tsx
just-boris: This is related to the comments below. I think resize observer callback should not use `visibleColumns` value, because it has access to the DOM and knows actually displayed columns

src/table/use-column-widths.tsx
pan-kot: Ideally, yes. Practically - that is very tricky for two reasons:

1. When the value e.g. a style or a class name is set imperatively, it can be then overridden by React for the same node the next render. That's why, for example, in sticky-columns implementation we return wrapper and cell styles as properties even though those are set by the util.
2. If not setting the value with React at all, we'd have to follow the component's lifecycle. In that case the style should be set not only in the observer's callback, but also upon component's mounting and any relevant properties change.

src/table/use-column-widths.tsx
pan-kot: Great call out. I haven't realised that in this version I am not using column.width. I will update the code to get the last column element from the existing reference.

src/table/use-column-widths.tsx
just-boris: You have two choices

1. Add `setStretchLastColumn()` state here and let React do its job
2. Extract `applyColumnWidths()` as a separate function and call it in both resize observer and in React lifecycle

Both options sound good for me, but it needs to be only one

src/table/use-column-widths.tsx
pan-kot: Surprisingly, I haven't found tests or permutations for that case, but the component does work when passing an empty column definitions array.

just-boris: All right, something to think about later. For this story I am wondering if we can make even earlier return, together with `!resizableColumns`, so even the container width is not tracked

Technically, if there are no columns, there are no resizable columns too 🤷

src/table/use-column-widths.tsx
pan-kot: I like the second approach and will explore it 👍 

How the setStretchLastColumn approach is supposed to work? I can't see how it can fix the issue.

src/table/use-column-widths.tsx
pan-kot: Makes sense!

src/table/use-column-widths.tsx
just-boris: > How the setStretchLastColumn approach is supposed to work?

So the `stretchLastColumn`  becomes the state.

Then all code uses this state. React takes care of re-rendering

pan-kot: It does re-render everything already but too late. I thought about the second suggestion and the question is how to apply the function within the React lifecycle exactly. The easiest solution I found was to execute it when the ref is assigned as it gives access to the node and fires after the thead render. This actually allows keeping all code within the util and is less complex than I originally anticipated. I will share my second revision soon.

just-boris (dismissed):

src/table/__integ__/resizable-columns.test.ts
```diff
@@ -178,11 +192,25 @@ describe.each([true, false])('StickyHeader=%s', sticky => {
   );
 
   test(
-    'should render "width: auto" for the last on big screens and explicit value on small',
```
pan-kot, line 181: Having width="auto" on the sticky header last column technically works because it only makes the column take the full width and the remaining columns have explicitly synced widths anyways. However, it is simpler to just synchronise all column widths by offset value instead.

src/table/thead.tsx
```diff
@@ -133,27 +131,10 @@ const Thead = React.forwardRef(
...
-              if (colIndex === columnDefinitions.length - 1 && containerWidth && containerWidth > totalWidth) {
-                // let the last column grow and fill the container width
-                widthOverride = 'auto';
-              }
-            }
             return (
               <TableHeaderCell
                 key={columnId}
                 className={headerCellClass}
-                style={{
```
pan-kot, line 152: This part is taken care of by the column widths util. The exact moment when the width is set in the element lifecycle is now slightly different so having all widths set the same way reduces the risk of the properties being out of sync at any given point in time.

src/table/use-column-widths.tsx
```diff
@@ -63,26 +64,77 @@ const WidthsContext = createContext<WidthsContext>({
...
+      setElementWidth(cellElement, 'minWidth', column?.minWidth);
+    }
+    if (!resizableColumns) {
+      setElementWidth(cellElement, 'width', column?.width);
+      setElementWidth(cellElement, 'minWidth', column?.minWidth);
+      setElementWidth(cellElement, 'maxWidth', column?.maxWidth);
+    }
+
+    // Synchronize cell and sticky cells widths.
+    setElementWidth(stickyCellElement, 'width', cellElement?.offsetWidth);
```
pan-kot, line 111: The sticky column widths were synced in the use-sticky-header before. It worked because first the width was set during thead rendering and then useLayoutEffect of the use-sticky-header fired. Now, the widths are set in the useEffect which happens slightly after than it was before. Besides, the widths can be only synced properly after the browser can figure out the widths of the columns that have no explicit value set, that's why the timeout is added.

src/table/thead.tsx
pan-kot: Update: this part is still needed as otherwise the initial rendering performance is impacted: every update to the styles of rendered nodes issues a reflow.

src/table/__integ__/resizable-columns.test.ts
```diff
@@ -178,11 +192,25 @@ describe.each([true, false])('StickyHeader=%s', sticky => {
...
+    'sticky and real column headers must have identical widths on big screen',
+    setupStickyTest(async page => {
+      const stickyHeaders = await page.getFirstTableHeaders();
+      const realHeaders = await page.getLastTableHeaders();
+
+      expect(stickyHeaders.length).toBe(realHeaders.length);
+      expect(stickyHeaders.map(el => el.offsetWidth)).toEqual(realHeaders.map(el => el.offsetWidth));
```
just-boris, line 201: This test does not work as you expect. It compares `undefined` to `undefined`.

Why this happens – I will leave a learning opportunity for you

src/table/__tests__/resizable-columns.test.tsx
```diff
@@ -431,3 +437,11 @@ describe('column header content', () => {
     expect(getResizeHandle(1)).toHaveAttribute('aria-roledescription', 'resize button');
   });
 });
+
+test('should set last column width to "auto" when container width exceeds total column width', () => {
+  jest.mocked(useResizeObserver).mockImplementation((_target, cb) => cb({ contentBoxWidth: 451 } as any));
```
just-boris, line 442: If there is a non round value, it needs a comment explaining why

src/table/use-column-widths.tsx
```diff
@@ -63,26 +66,71 @@ const WidthsContext = createContext<WidthsContext>({
...
+        setElementWidth(cellElement, 'width', columnWidths[column.id]);
+      }
+      setElementWidth(cellElement, 'minWidth', column?.minWidth);
+    } else {
+      setElementWidth(cellElement, 'width', column?.width);
+      setElementWidth(cellElement, 'minWidth', column?.minWidth);
+    }
+    if (!resizableColumns) {
+      setElementWidth(cellElement, 'maxWidth', column?.maxWidth);
+    }
```
just-boris, line 109: Multiple `setElementWidth` may cause extra reflows can't they?

Also, there is a little bit of asymmetry, not all widths are reset in all conditions. Are you sure there will be no stale widths after state update?

How about refactoring this code into separate parts

```js
const {minWidth, width, maxWidth} = getColumnStyles(...);
setElementStyle(cellElement, {minWidth, width, maxWidth})
```

This solves both problems and makes the code more readable

src/table/__tests__/resizable-columns.test.tsx
pan-kot: Makes sense , will add

src/table/use-column-widths.tsx
pan-kot: Thanks, the refactoring makes sense. In that particular case it won't make the performance better because the widths are all set in the thead render anyways. Those are updated imperatively only when there is a mismatch.

src/table/__integ__/resizable-columns.test.ts
pan-kot: Makes perfect sense - great catch!

just-boris (approved):

PR #1724, opened 2023-11-08
fix: Property filter token editor match by key

pan-kot:
### Description

In a recent change to the property filter component the matching logic was updated to compare filtering properties by reference. That does not work though for the token editor with async options because in the editor the property is cached in React state. Then, if the properties are updated as result of the onLoadItems call - the options can no longer be matched.

ref: AWSUI-24057

### How has this been tested?

The issue is reproducible in the existing `async-loading.integ.page.tsx`.  Without the fix there are no suggestions in the token editor. With the fix all suggestions are available.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/property-filter/__tests__/property-filter.test.tsx
+++ b/src/property-filter/__tests__/property-filter.test.tsx
@@ -1213,6 +1213,41 @@ describe('property filter parts', () => {
         })
       );
     });
+
+    test('token editor has matched property options', () => {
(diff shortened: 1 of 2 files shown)


just-boris (commented):
I read your "Testing" PR description part, but I still think it could be a dedicated test. The issue is very easy to reproduce in unit tests. We have a bunch of similar tests for selection tracking already

just-boris (approved):

PR #1727, opened 2023-11-09
chore: Remove unnecessary use of act in date picker tests

michaeldowseza:
### Description

Test utils was updated to do this by default. It is no longer required in our tests.

### How has this been tested?

Existing unit tests still pass.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/date-picker/__tests__/date-picker.test.tsx
+++ b/src/date-picker/__tests__/date-picker.test.tsx
@@ -1,7 +1,7 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 import * as React from 'react';
-import { act, render } from '@testing-library/react';
+import { render } from '@testing-library/react';
 import createWrapper from '../../../lib/components/test-utils/dom';
 import DatePicker, { DatePickerProps } from '../../../lib/components/date-picker';
 import DatePickerWrapper from '../../../lib/components/test-utils/dom/date-picker';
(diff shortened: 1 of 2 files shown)


src/date-picker/__tests__/date-picker.test.tsx
```diff
@@ -193,23 +193,23 @@ describe('Date picker - direct date input', () => {
...
+      getByTestId(outsideId).focus();
       expect(onBlurSpy).toHaveBeenCalledTimes(1);
     });
 
     test('should fire blur handler on open calendar when focusing another element', () => {
       wrapper.findOpenCalendarButton().click();
       expect(wrapper.findCalendarDropdown()).not.toBeNull();
 
-      act(() => getByTestId(outsideId).focus());
+      getByTestId(outsideId).focus();
```
just-boris, line 212: It would be very interesting if it passes. `getByTestId` returns you the native element instance, so `focus()` will not have any act wrappers

michaeldowseza: It passed 👀

michaeldowseza: Do we trust that it passes or would this be introducing something flaky?

just-boris: This test passing means, we do not have any `useEffect` depending on this focus event. If there will be any effects, they will not fire and you cannot write the proper assertion.

`act()` forces pending effects to fire, that's the whole trick

just-boris (approved):

PR #1729, opened 2023-11-13
chore: Migrate Button component to logical properties

DaemonCahill:
This PR is a subset of the changes drafted in #1713 to ensure that the migration to logical properties across the codebase will not create unintended failures upon merging.


.stylelintrc
```diff
@@ -49,7 +49,61 @@
       }
     ]
   },
+  "overrides": [
+    {
+      "files": ["./src/button/*.scss"],
+      "rules": {
+        "property-disallowed-list": [
```
just-boris, line 56: Why do we need this long list, if `csstools/use-logical` seem to enforce the same rules anyway?

src/button/styles.scss
```diff
@@ -110,40 +115,38 @@
...
+    margin-inline-end: calc(-1 * #{awsui.$space-xxs});
   }
 
   &.variant-inline-icon {
-    padding-top: 0;
-    padding-bottom: 0;
-    border: 0;
+    padding-block: 0;
+    border-block: 0;
+    border-inline: 0;
```
just-boris, line 130: Are these really equivalent? Shouldn't be be like this?

```css
padding-block-start: 0;
padding-block-end: 0;
border: 0; // why changing this, if it is zero all way round?
```

.stylelintrc
DaemonCahill: You are correct that there are redundancies between these rules and the plugin. The first reason is that the plugin is more permissive with shorthands (i.e. border, margin, padding) than we want to be instead of opting for logical only shorthands (i.e. border-inline, padding-block). The second is that, over the long term, it is likely that we change and/or remove this plugin and the rules ensure that these properties never make an appearance in our codebase. A minimal alternative to this can be seen in the parent PR: https://github.com/cloudscape-design/components/pull/1713/files#diff-a3f38f2cae79a3819f93ff1a9d4cd281cbe8f95696e14a29864f08796d3dc568

I'm inclined to use the comprehensive list but am open to an alternative opinion.

src/button/styles.scss
DaemonCahill: Logical properties have new single axis two property shorthands (i.e. `padding-block`, `padding-inline`). Four property shorthands (i.e. `margin`, `padding`, `border`) are a risk because they require that the 1st/3rd and 2nd/4th values are exactly the same, thus introducing the potential for bugs. Pull quote from the technical proposal:

> As seen above, the tradeoff with using logical properties is that some common physical shorthands (example: margin: 1px 2px 3px 4px) [do not have logical counterparts](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_logical_properties_and_values/Margins_borders_padding#margin_shorthands). However, logical properties do offer new shorthands such as margin-inline and margin-block that were previously missing from the web platform. Additionally, there is an [ongoing proposal to add a flow relative syntax for shorthands](https://github.com/w3c/csswg-drafts/issues/1282). Therefore, the recommendation is to add physical shorthands to the disallowed list given the viable alternatives.

just-boris: Makes sense, but...

This type of migration converts it from a dummy no-brainer "find and replace type" of code review to additional scrutiny on every PR, because the CSS is not equivalent.

So here is the choice

1. Dumb down the changes to fully equivalent `padding-top` -> `padding-block-start`
2. Break it down into even more PRs, one component at most for each one

.stylelintrc
just-boris: What's stopping us from applying your change from another PR right away? The shorter list looks good for me

just-boris (approved):

PR #1731, opened 2023-11-14
refactor: Update resizer styles to avoid ::before selector

pan-kot:
### Description

Removing `::before` selectors from table header cell styles because the same is used by the focus-highlight mixin. Assigning different styles for `::before` may produce unexpected results. The change is needed to support the upcoming keyboard navigation feature that makes table body- and header- cells focusable. It also fixes an existing bug: when the resizer is focused with keyboard and then hovered over the styles collide:

https://github.com/cloudscape-design/components/assets/20790937/d5e3438f-0362-45a2-9185-3a834cbca513

### How has this been tested?

* Existing unit tests now cover the presence of the divider;
* Manual testing for sticky cell styles hiding resize separator conditionally;
* Screenshot tests.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/table/resizer/index.tsx
+++ b/src/table/resizer/index.tsx
@@ -24,6 +24,10 @@ const AUTO_GROW_START_TIME = 10;
 const AUTO_GROW_INTERVAL = 10;
 const AUTO_GROW_INCREMENT = 5;
 
+export function Divider({ className }: { className?: string }) {
+  return <span className={clsx(styles.divider, styles['divider-disabled'], className)} />;
+}
+
 export function Resizer({
   onWidthUpdate,
   onWidthUpdateCommit,
@@ -207,6 +211,7 @@ export function Resizer({
         data-focus-id={focusId}
       />
       <span
+        className={clsx(styles.divider, isDragging && styles['divider-active'])}
         ref={resizerSeparatorRef}
         id={separatorId}
         role="separator"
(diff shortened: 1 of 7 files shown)


src/table/resizer/styles.scss
```diff
@@ -15,6 +15,31 @@
 $handle-width: awsui.$space-l;
 $active-separator-width: 2px;
 
+.divider {
+  th:not(:last-child) > & {
```
jperals, line 19: Could we remove this level of nesting? Like

```
th:not(:last-child) > .divider {
```

src/table/resizer/index.tsx
```diff
@@ -24,6 +24,10 @@ const AUTO_GROW_START_TIME = 10;
 const AUTO_GROW_INTERVAL = 10;
 const AUTO_GROW_INCREMENT = 5;
 
+export function Divider({ className }: { className?: string }) {
+  return <span className={clsx(styles.divider, styles['divider-disabled'], className)} />;
```
jperals, line 28: Looks like the `divider-disabled` class is applied unconditionally, is this intended?

pan-kot: Yes, the table header renders either Resizer or Divider (when resizableColumns=false).

jperals: But what is the point of the `divider-disabled` class if it is going to be always present?

So this selector
```
.divider.divider-disabled
```
is equivalent to
```
.divider
```

pan-kot: The .divider class is also assigned for the separator element of the Resizer component.

jperals (approved):

PR #1734, opened 2023-11-15
fix: Flashbar renders invisible interactive content

connorlanigan:
### Description

This fixes a bug where interactive content of the flashbar would be rendered twice, which leads to duplicated tab stops if the content contains interactive elements like links.

Instead of duplicating the content into the LiveRegion, the flashbar now provides references to the existing content.

Related links, issue #, if available: AWSUI-24287

### How has this been tested?
Tested in a dev page, with VoiceOver

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/components/live-region/index.tsx
+++ b/src/internal/components/live-region/index.tsx
@@ -13,8 +13,15 @@ export interface LiveRegionProps extends ScreenreaderOnlyProps {
   delay?: number;
   visible?: boolean;
   tagName?: 'span' | 'div';
-  children: React.ReactNode;
   id?: string;
+  /**
+   * Use a list of strings and/or existing DOM elements for building the
+   * announcement text. This avoids rendering separate content just for this
+   * LiveRegion.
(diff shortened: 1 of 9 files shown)


src/internal/components/live-region/index.tsx
```diff
@@ -76,9 +95,32 @@ function LiveRegion({
     does not impact the performance. If it does, prefer using a string as children prop.
   */
   useEffect(() => {
+    function getSourceContent() {
+      if (source) {
+        return source
+          .map(source => {
```
gethinwebster, line 101: having `source` as the variable name for both the parent and child is a bit hard to keep track of here

gethinwebster (changes requested):
looks good overall, but can we add some sort of tests for this? it feels like it could quite easily regress otherwise

gethinwebster (approved):

gethinwebster (approved):

PR #1735, opened 2023-11-15
chore: Migrate "A" components to logical properties

DaemonCahill:
This PR is a subset of the changes drafted in #1713 that affect only the subset of components that being with the letter "A".

### DIFF
--- a/src/app-layout/visual-refresh/trigger-button.scss
+++ b/src/app-layout/visual-refresh/trigger-button.scss
@@ -99,18 +99,12 @@ handleSplitPanelMaxWidth function in the context.
 
 .dot {
   position: absolute;
-<<<<<<< HEAD
-  width: 9px;
-  height: 9px;
-  border-radius: 8px;
-=======
-  inline-size: 10px;
-  block-size: 10px;
+  inline-size: 9px;
+  block-size: 9px;
   border-start-start-radius: 8px;
   border-start-end-radius: 8px;
   border-end-start-radius: 8px;
   border-end-end-radius: 8px;
->>>>>>> 87e5ce54 (Migrate ‘A’ components to logical properties.)
   background-color: awsui.$color-background-badge-icon;
   inset-block-start: 0;
   inset-inline-end: 0;


.stylelintrc
```diff
@@ -51,7 +51,16 @@
...
     {
-      "files": ["./src/button/*.scss"],
+      "files": [
+        "./src/alert/*.scss",
+        "./src/anchor-navigation/*.scss",
+        "./src/annotation-context/*.scss",
+        "./src/app-layout/*.scss",
+        "./src/area-chart/*.scss",
+        "./src/attribute-editor/*.scss",
+        "./src/autosuggest/*.scss",
```
timogasda, line 61: Since you're doing it by the letter, you could probably just do `"./src/a*/*.scss"` 😄

src/alert/styles.scss
```diff
@@ -25,9 +25,14 @@
   position: relative;
   display: grid;
   grid-template-columns: 1fr;
-  border-radius: awsui.$border-radius-alert;
-  border: awsui.$border-field-width solid;
-  padding: awsui.$space-alert-vertical awsui.$space-alert-horizontal;
+  border-start-start-radius: awsui.$border-radius-alert;
+  border-start-end-radius: awsui.$border-radius-alert;
+  border-end-start-radius: awsui.$border-radius-alert;
+  border-end-end-radius: awsui.$border-radius-alert;
```
timogasda, line 31: I'm not sure if the "no shorthands" rule makes a ton of sense for `border-radius`. It's almost always the same value for all.

just-boris: We discussed it on the previous PR: https://github.com/cloudscape-design/components/pull/1729#discussion_r1391666829

`border-radius` is a positional property: `[top] [right] [bottom] [left]`. For safety and ease of coding we decided to ban all cases rather than analysing possible exceptions.

Open for discussion (we can write a custom stylelint rule), but I would not block the ongoing work because of that

timogasda: I would imagine it would be as simple as removing "border" from this line: `"property-disallowed-list": ["border", "border-radius", "border-style", "margin", "padding"],`

just-boris: True, for `border` the CSS grammar does not allow setting more than one value. So it is all-way-round anyway.

But the `border-radius` will have to stay expanded. For 4 `border:` occurrences in this PR we can revert them back

cc @DaemonCahill

just-boris (commented):
All good from my end, but we need to address this thread: https://github.com/cloudscape-design/components/pull/1735/files?w=1#r1395369006

.stylelintrc
```diff
@@ -51,7 +51,16 @@
...
+        "./src/button/*.scss"
+      ],
       "rules": {
         "property-disallowed-list": ["border", "border-radius", "border-style", "margin", "padding"],
```
just-boris, line 65: Based on the parallel thread we should remove `border` from that list and add `border-width`

.stylelintrc
DaemonCahill: You have a keen eye!

src/alert/styles.scss
DaemonCahill: Good callout! I will update the linting rules to keep the `border` shorthand but still block `border-radius`.

timogasda (approved):

just-boris (approved):

connorlanigan (dismissed):

connorlanigan (approved):

connorlanigan (dismissed):

katiegeorge (approved):

PR #1739, opened 2023-11-16
fix: Table empty cell no jitter upon resizing with React 18

pan-kot:
### Description

Update empty cell to observe table container width changes and apply new width immediately. That prevents state when table width temporarily becomes bigger than the wrapper width making the scrollbar appear. The problem was only reproducible with React 18.

See a similar fix to the last column width: https://github.com/cloudscape-design/components/pull/1718

Before:

https://github.com/cloudscape-design/components/assets/20790937/9fb6353b-455b-4f73-8077-431f20977f06

After:

https://github.com/cloudscape-design/components/assets/20790937/195e4837-ec2f-4c5f-81b6-dcc71fdd931b

ref: AWSUI-24405

### How has this been tested?

Manual testing

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/table/node-data-cell.tsx
```diff
@@ -1,54 +1,46 @@
...
   empty,
   tableRef,
+  containerRef,
 }: NoDataCellProps) {
-  const [tablePaddings, setTablePaddings] = useState(containerWidth);
+  const cellContentRef = useRef<HTMLDivElement>(null);
 
-  useEffect(() => {
-    if (tableRef.current) {
+  useResizeObserver(containerRef, ({ contentBoxWidth }) => {
```
just-boris, line 32: For a better performance we probably should attach only one observer for the ref (because each observation callback is a separate react rendering loop).

Fine by now, because there is no react state inside, but this is something to keep an eye for the future

just-boris (dismissed):

src/table/__tests__/empty-state.test.tsx
```diff
@@ -61,19 +62,29 @@ function findMergedCell(wrapper: TableWrapper) {
   return wrapper.findByClassName(styles['cell-merged']);
 }
 
+function mockResizeObserver(contentBoxWidth: number) {
+  // Use "act" to imitate firing observer's callback after component is rendered.
+  // In actual code the callback is fired twice - before and after rendering but the second part depends on ResizeObserver polyfill.
+  jest
+    .mocked(useResizeObserver)
+    .mockImplementation((_target, cb) => act(() => cb({ contentBoxWidth } as unknown as ContainerQueryEntry)));
```
just-boris, line 70: This is incorrect place for `act`. As per [the docs](https://legacy.reactjs.org/docs/testing-recipes.html#act), it is a part of "Arrange-Act-Assert", it is an action _outside_ the system under test. Putting it inside the hook mock is _inside_ the system and this is incorrect.


If you needed to handle the initial state rendering, you could make the `renderTable` async

```js
async function renderTable() {
    // render table synchronously
   return act(() => Promise.resolve());  // or another delay if you need it
}
```

pan-kot: That is a hack to make the mock behave in a particular way. When in DOM the polyfill ResizeObserver fires upon creating in the use-effect which happens after the component is mounted. In the test environment that does not happen.

just-boris: Your comment does not contradict my suggestion. Why do you prefer your option?

just-boris: Did an extra dive into the same story on parallel PR: https://github.com/cloudscape-design/components/pull/1745#discussion_r1399388599

I think, both mock cases should have the same treatment

src/table/__tests__/empty-state.test.tsx
```diff
@@ -61,19 +62,49 @@ function findMergedCell(wrapper: TableWrapper) {
...
+  const cbRef: { current: (entry: ContainerQueryEntry) => void } = { current: () => {} };
+  jest.mocked(useResizeObserver).mockImplementation((_target, cb) => {
+    // The table uses more than one resize observer.
+    // The callback must be triggered for all to ensure the expected one is targeted as well.
+    const prev = cbRef.current;
+    cbRef.current = entry => {
+      prev(entry);
+      cb(entry);
+      return 5;
+    };
```
just-boris, line 75: 1. Same comment about using a list of callbacks instead of recursive chains: https://github.com/cloudscape-design/components/pull/1745#discussion_r1422909903 Think about your own debug experience and more readable stack traces
2. What does this number 5 do? In the types, you are saying it returns `void`

src/table/__tests__/empty-state.test.tsx
```diff
@@ -61,19 +62,49 @@ function findMergedCell(wrapper: TableWrapper) {
...
+
+    useLayoutEffect(() => {
+      cb({ contentBoxWidth } as unknown as ContainerQueryEntry);
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, []);
+
+    useEffect(() => {
+      cb({ contentBoxWidth } as unknown as ContainerQueryEntry);
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, []);
```
just-boris, line 85: Are you sure you need both and not just one?

just-boris (commented):
Almost there, but let's resolve comments and merge the other PR #1745 first

src/table/__tests__/empty-state.test.tsx
pan-kot: Thanks. The number 5 return was for debugging, I forgot to get rid of it

src/table/__tests__/empty-state.test.tsx
```diff
@@ -61,19 +62,42 @@ function findMergedCell(wrapper: TableWrapper) {
...
+
+  jest.mocked(useResizeObserver).mockImplementation((_target, cb) => {
+    // The table uses more than one resize observer.
+    // The callback must be triggered for all to ensure the expected one is targeted as well.
+    callbacks.push(cb);
+
+    useEffect(() => {
+      cb({ contentBoxWidth } as unknown as ContainerQueryEntry);
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+    });
```
just-boris, line 77: In the other PR you had and empty array, here you do not. Which one is correct?

just-boris (approved):

src/table/__tests__/empty-state.test.tsx
pan-kot: Thanks, I removed it accidentally. Having an empty dependency array is correct

just-boris (approved):

PR #1387, opened 2023-07-28
fix: Render Visual Refresh background height in sync with content

jperals:
This reverts commit 4773c009abd5393c5e8cd7765f9031b238dc29a9 and adds a fix for the commit that it had reverted.

### Description

#1331 had a regression for the case where no component uses the `useDynamicOverlap` hook but content header is passed. In this case, `isDynamicOverlapDisabled` was initialized to `true` and never set back to `false` because `updateDynamicOverlap` was never run.

This is fixed by initializing `isDynamicOverlapDisabled` to `false` unless `props.disableDynamicOverlap` is passed as `true`.

### How has this been tested?

- Added new unit test

<details>
   <summary>Review checklist</summary>

_The following items are to be evaluated by the author(s) and the reviewer(s)._

#### Correctness

- _Changes include appropriate documentation updates._
- _Changes are backward-compatible if not indicated, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#public-apis)._
- _Changes do not include unsupported browser features, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#browsers-support)._
- _Changes were manually tested for accessibility, see [accessibility guidelines](https://cloudscape.design/foundation/core-principles/accessibility/)._

#### Security

- _If the code handles URLs: all URLs are validated through [the `checkSafeUrl` function](https://github.com/cloudscape-design/components/blob/main/src/internal/utils/check-safe-url.ts)._

#### Testing

- _Changes are covered with new/existing unit tests?_
- _Changes are covered with new/existing integration tests?_
</details>

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1745, opened 2023-11-20
fix: Last table resizable columns stable with React 18 (#2)

pan-kot:
### Description

Re-introduction of https://github.com/cloudscape-design/components/pull/1718 but with a fix to default column IDs.

### How has this been tested?

New integration tests to secure regression.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/table/resizable-columns.page.tsx
+++ b/pages/table/resizable-columns.page.tsx
@@ -104,12 +104,14 @@ type PageContext = React.Context<
 
 export default function App() {
   const { urlParams, setUrlParams } = useContext(AppContext as PageContext);
-  const wrapLines = urlParams.wrapLines ?? false;
-  const stickyHeader = urlParams.stickyHeader ?? false;
-  const resizableColumns = urlParams.resizableColumns ?? true;
-  const fullPage = urlParams.fullPage ?? false;
-  const withColumnIds = urlParams.withColumnIds ?? true;
-  const withSelection = urlParams.withSelection ?? false;
+  const {
+    wrapLines = false,
+    stickyHeader = false,
+    resizableColumns = true,
+    fullPage = false,
+    withColumnIds = true,
+    withSelection = false,
+  } = urlParams;
 
   const [renderKey, setRenderKey] = useState(0);
   const [columns, setColumns] = useState(columnsConfig);
--- a/pages/table/resizable-columns.page.tsx
+++ b/pages/table/resizable-columns.page.tsx
@@ -97,6 +97,8 @@ type PageContext = React.Context<
     stickyHeader: boolean;
     resizableColumns: boolean;
     fullPage: boolean;
+    withColumnIds?: boolean;
+    withSelection?: boolean;
   }>
 >;
 
@@ -106,6 +108,8 @@ export default function App() {
   const stickyHeader = urlParams.stickyHeader ?? false;
   const resizableColumns = urlParams.resizableColumns ?? true;
   const fullPage = urlParams.fullPage ?? false;
+  const withColumnIds = urlParams.withColumnIds ?? true;
+  const withSelection = urlParams.withSelection ?? false;
 
   const [renderKey, setRenderKey] = useState(0);
   const [columns, setColumns] = useState(columnsConfig);
@@ -161,6 +165,12 @@ export default function App() {
             <Checkbox checked={fullPage} onChange={event => setUrlParams({ fullPage: event.detail.checked })}>
               Full page table
             </Checkbox>
+            <Checkbox checked={withColumnIds} onChange={event => setUrlParams({ withColumnIds: event.detail.checked })}>
+              Columns have IDs
+            </Checkbox>
+            <Checkbox checked={withSelection} onChange={event => setUrlParams({ withSelection: event.detail.checked })}>
+              With row selection
+            </Checkbox>
           </div>
           <div>
             {columnsConfig.map(column => (
@@ -190,9 +200,10 @@ export default function App() {
           key={renderKey}
           header={<Header>Simple table</Header>}
           stickyHeader={stickyHeader}
-          columnDefinitions={columns}
+          columnDefinitions={columns.map(col => (withColumnIds ? col : { ...col, id: undefined }))}
           resizableColumns={resizableColumns}
-          columnDisplay={columnDisplay}
+          columnDisplay={withColumnIds ? columnDisplay : undefined}
+          selectionType={withSelection ? 'single' : undefined}
           items={items}
           wrapLines={wrapLines}
           sortingColumn={sorting?.sortingColumn}
--- a/src/table/__integ__/resizable-columns.test.ts
+++ b/src/table/__integ__/resizable-columns.test.ts
@@ -60,6 +60,20 @@ class TablePage extends BasePageObject {
     return element.getAttribute('style');
   }
 
+  getFirstTableHeaderWidths() {
+    return this.browser.execute(() => {
+      const tables = document.querySelectorAll('table');
+      return Array.from(tables[0].querySelectorAll('th')).map(el => el.offsetWidth);
+    });
+  }
+
+  getLastTableHeaderWidths() {
+    return this.browser.execute(() => {
+      const tables = document.querySelectorAll('table');
+      return Array.from(tables[tables.length - 1].querySelectorAll('th')).map(el => el.offsetWidth);
+    });
+  }
+
   async getColumnMinWidth(columnIndex: number) {
     const columnSelector = tableWrapper
       // use internal CSS-selector to always receive the real table header and not a sticky copy
@@ -177,13 +191,13 @@ describe.each([true, false])('StickyHeader=%s', sticky => {
     })
   );
 
-  test(
-    'should render "width: auto" for the last on big screens and explicit value on small',
+  test.each([1680, 620])('sticky and real column headers must have identical widths for screen width %s', width =>
     setupStickyTest(async page => {
-      await expect(page.getColumnStyle(4)).resolves.toContain('width: auto;');
-      await page.setWindowSize({ ...defaultScreen, width: 620 });
-      await expect(page.getColumnStyle(4)).resolves.toContain('width: 120px;');
-    })
+      await page.setWindowSize({ ...defaultScreen, width });
+      const stickyHeaderWidths = await page.getFirstTableHeaderWidths();
+      const realHeaderWidths = await page.getLastTableHeaderWidths();
+      expect(stickyHeaderWidths).toEqual(realHeaderWidths);
+    })()
   );
 
   // The page width of 620px is an empirical value defined for the respective test page in VR
--- a/src/table/__integ__/resizable-columns.test.ts
+++ b/src/table/__integ__/resizable-columns.test.ts
@@ -123,17 +123,29 @@ class TablePage extends BasePageObject {
   }
 }
 
-const setupTest = (testFn: (page: TablePage) => Promise<void>) => {
+interface PageConfig {
+  stickyHeader?: boolean;
+  withColumnIds?: boolean;
+  withSelection?: boolean;
+}
+
+const setupTest = (testFn: (page: TablePage) => Promise<void>, config?: PageConfig) => {
   return useBrowser({ ...defaultScreen }, async browser => {
     const page = new TablePage(browser);
-    await browser.url('#/light/table/resizable-columns?visualRefresh=false');
+    const params = new URLSearchParams({
+      visualRefresh: 'false',
+      stickyHeader: config?.stickyHeader !== undefined ? String(config.stickyHeader) : 'false',
+      withColumnIds: config?.withColumnIds !== undefined ? String(config.withColumnIds) : 'true',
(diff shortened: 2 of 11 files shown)


src/table/utils.ts
```diff
@@ -27,7 +27,7 @@ export const getTrackableValue = <T>(trackBy: TableProps.TrackBy<T> | undefined,
 };
 
 export const getColumnKey = <T>(column: TableProps.ColumnDefinition<T>, index: number) => {
-  return column.id || index;
+  return (column.id || index).toString();
```
pan-kot, line 30: The fix to columns being out of sync between header and sticky header.

src/table/use-column-widths.tsx
```diff
@@ -63,26 +66,76 @@ const WidthsContext = createContext<WidthsContext>({
...
-      cellsRef.current[columnId] = node;
+      ref.current[columnId] = node;
     } else {
-      delete cellsRef.current[columnId];
+      delete ref.current[columnId];
     }
   };
 
+  const getColumnStyles = (sticky: boolean, columnId: PropertyKey): React.CSSProperties => {
+    const column = visibleColumns.find(column => column.id === columnId);
```
pan-kot, line 91: This comparison didn't work because column.id was of type "number" while columnId was of type "string" (converted by Object.keys()).

src/table/use-column-widths.tsx
```diff
@@ -63,26 +66,78 @@ const WidthsContext = createContext<WidthsContext>({
...
+  // Imperatively sets width style for a cell avoiding React state.
+  // This allows setting the style as soon container's size change is observed.
+  const updateColumnWidths = useStableCallback(() => {
+    for (const column of visibleColumns) {
+      setElementWidths(cellsRef.current[column.id], getColumnStyles(false, column.id));
+    }
+    // Sticky column widths must be synchronized once all real column widths are assigned.
+    if (Object.keys(stickyCellsRef.current).length > 0) {
+      for (const { id } of visibleColumns) {
+        setElementWidths(stickyCellsRef.current[id], getColumnStyles(true, id));
```
just-boris, line 125: Don't you want to add a check here that this column id is sticky, instead of relying on some code 3 levels down below to capture this with an undefined check?

pages/table/resizable-columns.page.tsx
```diff
@@ -106,6 +108,8 @@ export default function App() {
   const stickyHeader = urlParams.stickyHeader ?? false;
   const resizableColumns = urlParams.resizableColumns ?? true;
   const fullPage = urlParams.fullPage ?? false;
+  const withColumnIds = urlParams.withColumnIds ?? true;
+  const withSelection = urlParams.withSelection ?? false;
```
just-boris, line 112: Isn't this mix of true and false defaults is hard to follow? Any chances of making it more ergonomic?

src/table/__tests__/resizable-columns.test.tsx
```diff
@@ -19,6 +20,11 @@ jest.mock('../../../lib/components/internal/utils/scrollable-containers', () =>
   }),
 }));
 
+jest.mock('@cloudscape-design/component-toolkit/internal', () => ({
+  ...jest.requireActual('@cloudscape-design/component-toolkit/internal'),
+  useResizeObserver: jest.fn().mockImplementation((_target, cb) => cb({ contentBoxWidth: 0 })),
```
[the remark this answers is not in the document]
pan-kot: Thanks, I will try it out

src/table/use-column-widths.tsx
pan-kot: I don't get it. Do you suggest checking the `stickyCellsRef.current[id]` to be defined?

pages/table/resizable-columns.page.tsx
pan-kot: Will try

src/table/thead.tsx
```diff
@@ -133,27 +131,11 @@ const Thead = React.forwardRef(
...
-              }
-              if (colIndex === columnDefinitions.length - 1 && containerWidth && containerWidth > totalWidth) {
-                // let the last column grow and fill the container width
-                widthOverride = 'auto';
-              }
-            }
             return (
               <TableHeaderCell
                 key={columnId}
+                style={getColumnStyles(sticky, columnId)}
```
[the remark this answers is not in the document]
just-boris: See my example above. It is not about "manually duplicate id" but about table internal logic using indexes as a fallback. This is not documented anywhere (and it should not). But we need to inform what is supported and what is not

src/table/use-column-widths.tsx
just-boris: At the very least, yes. But this is also not solid, because sticky is a dynamic preference, can change in runtime and I am not sure how much I can trust `stickyCellsRef` object.

Reading column definition sounds better

src/table/thead.tsx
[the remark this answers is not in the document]
just-boris: In any way, you can treat this as a documentation improvement follow-up. Do you track these somewhere?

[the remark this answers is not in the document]
pan-kot: Yeah I though about that but the type of the `id` API property is "string" and the type of the automatically generated one is "number". Unless we convert the second type or use a type-agnostic comparison there are no restriction to the provided ID values as soon there are no explicit duplicates.

src/table/use-column-widths.tsx
pan-kot: So we do read visibleColumns to get the list of IDs to sync. The expectation is that the sticky header was already rendered and the references are available, however, the setElementWidths accepts an element or undefined as its first argument so no explicit check is needed. If we add an explicit check here there is nothing extra we can do if the ref is not available.

just-boris: So, "relying on some code 3 levels down" is your final word on this.

Clearly see being a source of a similar scale regressions as the past revert, but feel free to go ahead.

src/table/thead.tsx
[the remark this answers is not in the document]
just-boris: If you think there is nothing to improve or follow up, fine by me

src/table/__tests__/resizable-columns.test.tsx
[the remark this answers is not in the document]
just-boris: So you updated the custom implementation for one test, but not the default one. Why?

just-boris:
How many tests do we have for sticky + resizable columns? How much are you sure this fix will work fine there?

src/table/__tests__/resizable-columns.test.tsx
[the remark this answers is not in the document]
pan-kot: The other tests do not depend on it. I will check if I can make the mock better scoped.

src/table/__tests__/resizable-columns.test.tsx
```diff
@@ -431,3 +438,37 @@ describe('column header content', () => {
...
   });
 });
+
+test('should set last column width to "auto" when container width exceeds total column width', () => {
+  const totalColumnsWidth = 150 + 300;
+
+  let outsideCb: (entry: ContainerQueryEntry) => void = () => {};
+  jest.mocked(useResizeObserver).mockImplementation((_target, cb) => {
+    // The table uses more than one resize observer.
+    // The callback must be triggered for all to ensure the expected one is targeted as well.
```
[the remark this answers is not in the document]
pan-kot: The would have been nice, but the target is the same as currently we attach multiple observers to the same wrapper measure DOM node. This can be fixed after 1743 and 1745 are merged by ensuring a single observer with some API for registering listeners.

src/table/__tests__/resizable-columns.test.tsx
```diff
@@ -431,3 +438,37 @@ describe('column header content', () => {
...
+
+    useLayoutEffect(() => {
+      cb({ contentBoxWidth: totalColumnsWidth + 1 } as unknown as ContainerQueryEntry);
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, []);
+
+    useEffect(() => {
+      cb({ contentBoxWidth: totalColumnsWidth + 1 } as unknown as ContainerQueryEntry);
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, []);
```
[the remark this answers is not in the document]
pan-kot: will remove it

src/table/__tests__/resizable-columns.test.tsx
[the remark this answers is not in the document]
just-boris: Sounds like a use case for a map:

```js
const observedNodes = new Map<HTMLElement, Array<Callback>>();

// ....

useLayoutEffect(() => {
  cb({ contentBoxWidth: initialValue } as unknown as ContainerQueryEntry);

  observedNodes.get(target).push(cb);
  return () => {
     observedNodes.set(target, observedNodes.get(target)).filter(value => cb !== value);      
  }
}, []);
```

usage in tests

```
function notifyResize(target, newSize) {
    act(() => {
       observedNodes.get(target).forEach(cb => cb(newSize))
    })
}
```

Why did I put `act`? See another related thread: https://github.com/cloudscape-design/components/pull/1727#discussion_r1387867723

just-boris:
> We now have integration tests securing the widths

This should be a hyperlink

src/table/__tests__/resizable-columns.test.tsx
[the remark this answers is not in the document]
pan-kot: Makes sense, I will add the `act`. See no big reason to have that observed nodes map though. W/o a map all table resize observer callbacks are triggered but I don't expect those to have any effect on the feature anyways.

pan-kot:
> > We now have integration tests securing the widths
> 
> This should be a hyperlink

These tests are added in the PR: https://github.com/cloudscape-design/components/pull/1745/files#diff-8df512bec543b70dc3e5044984b2cd2ccc6d31e9b7f45c1c78fed4e890875e03R272

src/table/__integ__/resizable-columns.test.ts
```diff
@@ -60,6 +60,20 @@ class TablePage extends BasePageObject {
     return element.getAttribute('style');
   }
 
+  getFirstTableHeaderWidths() {
+    return this.browser.execute(() => {
+      const tables = document.querySelectorAll('table');
```
jperals, line 65: Instead of using `document.querySelectorAll` and picking the first item, we could simplify a little bit here by just using `document.querySelector`

src/table/use-column-widths.tsx
```diff
@@ -63,26 +66,80 @@ const WidthsContext = createContext<WidthsContext>({
...
+
+  // Observes container size and requests an update to the last cell width as it depends on the container's width.
+  useResizeObserver(containerRef, ({ contentBoxWidth: containerWidth }) => {
+    containerWidthRef.current = containerWidth;
+    updateColumnWidths();
+  });
+
   // The widths of the dynamically added columns (after the first render) if not set explicitly
   // will default to the DEFAULT_COLUMN_WIDTH.
   useEffect(() => {
```
jperals, line 140: Before resorting to imperative DOM updates, have you looked for possible effects and unnecessary render cycles that can be removed? It would be good if we can fix more of these React 18 flickering issues as in #1684 rather than as in #1387.

I have tried removing this effect [here](https://github.com/cloudscape-design/components/commit/062c48b927c0843c4131ad8c89332fe733c47bea), I saw no noticeable difference and all tests passed —however unfortunately this didn't seem to remove the React 18 flickering either. Now trying another POC change on top [here](https://github.com/cloudscape-design/components/commit/6042ad771eb387e041b894c435ecef299ad681ca) to let the provider pass down the updated column widths right away instead of relying on the state (which will be reflected only on the next render cycle).

pan-kot: The column widths and their total width do not update when the container width change. Once container width changes, the last column width style is determined from comparison between total width and container width so I don't see how removing a state can help but it would be lovely to find a simpler solution of course.

src/table/__tests__/resizable-columns.test.tsx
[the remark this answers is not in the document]
just-boris: Still not fixed. If you do not want a `Map`, you could at least have an array of callbacks, to avoid hard-coded `prev` and `cb` recursive calls

src/table/use-column-widths.tsx
just-boris: Reminder about this thread. Still want to hear your plan how to deal with possible bugs in this code

src/table/use-column-widths.tsx
just-boris: This code reads DOM, so it should live inside effect to ensure we read the actual state and not some previous frames.

If you separate state-based calculations, it can go outside the effect. But current code cannot.

src/table/use-column-widths.tsx
pan-kot: The code is already updated. The setElementWidths no longer accepts undefined as target.

just-boris (approved):

PR #1751, opened 2023-11-23
feat: Support links in pie chart popover content

jperals:
### Description

Splitting this part out of #1644.

### How has this been tested?
- Added example dev page for integration and visual regression tests
- Added integration tests that cover keyboard navigation and announcements

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/internal/components/chart-plot/index.tsx
```diff
@@ -218,7 +218,11 @@ function ChartPlot(
         </g>
       </svg>
 
-      <LiveRegion>{ariaLiveRegion}</LiveRegion>
+      {typeof ariaLiveRegion === 'string' ? (
+        <LiveRegion>{ariaLiveRegion}</LiveRegion>
+      ) : (
+        <LiveRegion source={[ariaLiveRegion]} />
```
jperals, line 224: Use the new `source` attribute introduced in #1734 for content with potentially navigable elements

pages/pie-chart/with-links.page.tsx
```diff
@@ -0,0 +1,106 @@
...
+>;
+
+import { commonProps, data1, FoodData, segmentDescription1 } from './common';
+
+export default function () {
+  const { urlParams, setUrlParams } = useContext(AppContext as DemoContext);
+
+  return (
+    <>
+      <h1>Pie chart with links</h1>
```
pan-kot, line 25: Can we make it simpler by rendering two separate charts, one with value links and one with key links? This would allow to remove extra elements and will make it easier to understand the setup.

src/internal/components/chart-plot/index.tsx
```diff
@@ -39,7 +39,7 @@ export interface ChartPlotProps {
   activeElementKey?: null | string | number | boolean;
   activeElementRef?: React.RefObject<SVGGElement>;
   activeElementFocusOffset?: Offset;
-  ariaLiveRegion?: React.ReactNode;
+  ariaLiveRegion?: RefObject<HTMLElement> | string;
```
pan-kot, line 42: Can this be simplified to just `RefObject<HTMLElement>`?

src/pie-chart/__integ__/with-links.test.ts
```diff
@@ -0,0 +1,42 @@
...
+    it(
+      'allows to focus internal popover content',
+      setupTest(useLinks, async page => {
+        const expectedFocusedText =
+          useLinks === 'keys'
+            ? // The external link icon adds an extra space at the end
+              'Popularity '
+            : '50% ';
+        await page.click('#focus-target');
+        await page.keys(['Tab', 'Enter']);
```
pan-kot, line 26: I would add code comments to explain what a sequence of keys is for, e.g.
```
// Pin popover on the first chart segment
await page.keys(['Tab', 'Enter']);
// Move focus to the second interactive element of the popover
await page.keys('Tab');
```

src/internal/components/chart-plot/index.tsx
pan-kot: Or, maybe, instead of providing the `ariaLiveRegion` from the pie-chart render `<LiveRegion source={[ariaLiveRegion]} />` from the pie-chart.

pages/pie-chart/with-links.page.tsx
jperals: Done by splitting it in two pages. This will make it more convenient for eventual visual regression tests.

src/internal/components/chart-plot/index.tsx
jperals: The problem is that this is used by other chart types as well, which use string. Before, this was not needed only because `ReactNode` includes string, but in fact pie/donut charts pass down the element whereas cartesian charts build the comma-separated string on their side. This even leads to subtle differences because pie/donut chart live region content does not have commas, and IMO TBH these two approaches should be unified, but here I'm trying to minimize the amount of changes.

src/pie-chart/__integ__/with-links.test.ts
jperals: Done

src/internal/components/chart-plot/index.tsx
pan-kot: Yes, that's why my second comment. I just think the string | ref API is a bit difficult to follow. What if we remove the ariaLiveRegion property from the chart-plot? It is only used to render the LiveRegion component which can well be done from outside. To not increase the PR size this can be scoped to pie chart only

src/internal/components/chart-series-details/index.tsx
```diff
@@ -19,14 +19,14 @@ export interface ChartSeriesDetailsProps extends BaseComponentProps {
   details: ReadonlyArray<ChartSeriesDetailItem>;
 }
 
-export default memo(ChartSeriesDetails);
+export default memo(forwardRef(ChartSeriesDetails));
 
-function ChartSeriesDetails({ details, ...restProps }: ChartSeriesDetailsProps) {
+function ChartSeriesDetails({ details, ...restProps }: ChartSeriesDetailsProps, ref?: React.Ref<HTMLDivElement>) {
```
just-boris, line 24: What is happening here? Why do you make this change, but there is no any use of this ref?

Also, how can `ref` argument be undefined? Isn't it always supplied by react?

pan-kot (dismissed):

src/internal/components/chart-series-details/index.tsx
jperals: > What is happening here? Why do you make this change, but there is no any use of this ref?

This ref is used [here](https://github.com/cloudscape-design/components/pull/1751/files#diff-3828e76418234b3e69ed5fee63c4b82beac8da7d413fc71047088ac70ce00120R127). Before, we could pass the element as child of the live region but now that there might be interactive elements, we need to do it with a ref and pass it to the live region as `source`.

> how can ref argument be undefined? Isn't it always supplied by react?

Right, changed [here](https://github.com/cloudscape-design/components/pull/1751/commits/058739dc4d86cef8a781aac06343b40a3562280d).

pan-kot (approved):

PR #1756, opened 2023-11-29
fix: Keep chart popover position on resize

jperals:
### Description

Splitting #1644 in smaller parts.

The popover container recalculates their position when any of this happens:
- on scroll
- on page resize
- on popover resize
- on click (anywhere on the page)

However, when a click comes from inside the popover itself, this can lead to situations where the popover moves away from the cursor. This can be especially critical with chart popovers, because they are shown on hover, so that clicking on them can lead to the actual popover disappearing because the hover was lost:

https://github.com/cloudscape-design/components/assets/1257272/7175af27-30fd-47c3-9f9d-e0480c42a0e4

After these changes, the popover position is kept:

https://github.com/cloudscape-design/components/assets/1257272/b43b86ea-8602-49a3-b21b-c853ffb7daf8

In order to reduce any possible unexpected impact on other popovers, these changes are applied here only on chart popovers, via a new prop `keepPosition` in the internal popover container component.

### How has this been tested?

- Added integration tests
- Added unit tests
- Manually tested on the new dev page
- Manually tested on a website dry-run playground to make sure it also works on React 18

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/mixed-line-bar-chart/__integ__/mixed-line-bar-chart.test.ts
```diff
@@ -441,6 +442,25 @@ describe('Details popover', () => {
     })
   );
 
+  describe('keeps the popover position when it resizes due to interacting with the popover itself', () => {
+    test.each(['hover', 'click', 'keyboard'])('Interaction type: %s', interactionType =>
+      setupPopoverPositionTest(async page => {
+        await page.setWindowSize({ width: 900, height: 500 });
+        expect(page.findDetailPopover()).toBeTruthy();
+        await page.focusCoordinate(0, interactionType);
```
pan-kot, line 450: nit: The name of the utility is a little confusing. How about something like: `page.openPopoverOnBarGroup(0, interactionType)` ?

pan-kot (commented):
I wonder if we can find a simpler solution here for chart popovers specifically. We already know that having on-hover popovers with as much content and especially interactive content can be problematic when it comes to accessibility. What if we pin the popover upon any click- or focus event inside the popover?

src/mixed-line-bar-chart/__integ__/mixed-line-bar-chart.test.ts
jperals: Changed in the last commit. Also removed the previous line which didn't make sense, and was passing only because test util functions like `wrapper.findDetailPopover` returns a wrapper in integration tests even if the element isn't found.

src/mixed-line-bar-chart/__tests__/mixed-chart.test.tsx
```diff
@@ -1006,31 +1031,60 @@ describe('Details popover', () => {
     wrapper.findApplication()!.focus();
     expect(wrapper.findDetailPopover()?.findContent()?.getElement()).toHaveTextContent('Details about Group 1');
   });
-});
 
-test('highlighted series are controllable', () => {
-  const { wrapper, rerender } = renderMixedChart(
-    <MixedLineBarChart series={[lineSeries, thresholdSeries]} highlightedSeries={lineSeries} />
-  );
+  test('highlights relevant x-thresholds when navigating line series', () => {
```
jperals, line 1035: Moved this test inside the "Details popover" describe block

src/popover/container.tsx
```diff
@@ -47,145 +48,193 @@ export default function PopoverContainer({
...
-    const containingBlockRect = containingBlock ? containingBlock.getBoundingClientRect() : viewportRect;
-
-    const bodyBorderWidth = getBorderWidth(body);
-    const contentRect = contentRef.current.getBoundingClientRect();
-    const contentBoundingBox = {
-      width: contentRect.width + 2 * bodyBorderWidth,
-      height: contentRect.height + 2 * bodyBorderWidth,
-    };
+  const updatePositionHandler = useCallback(
+    (onResize = false) => {
```
jperals, line 70: Adding this `onResize` parameter made this entire function block appear in the diff because of the indentation. The actual relevant changes in it are lines 120 and 132.

jperals:
> > I wonder if we can find a simpler solution here for chart popovers specifically. We already know that having on-hover popovers with as much content and especially interactive content can be problematic when it comes to accessibility. What if we pin the popover upon any click- or focus event inside the popover?
> 
> Notice that if we do that, we would solve the problem that the popover completely disappears accidentally (because it's pinned), but it would still move away.
> 
> Consulted with design, the general idea of pinning when interacting sounds good in principle, but not so happy about the popover still repositioning, so I see this more as a potential separate improvement.
> 
>  pinned.mov

I would rather consider making this behavior generic for all popovers in the future, after #1644 is released. This way at least `keepPositionWhenInteracting` can be removed.

src/popover/container.tsx
```diff
@@ -47,145 +48,206 @@ export default function PopoverContainer({
...
 
   // Store the handler in a ref so that it can still be replaced from outside of the listener closure.
   const positionHandlerRef = useRef<() => void>(() => {});
 
+  // Remember when the popover was clicked in order to prevent eventually undesired position recalculations coming from that interaction.
+  const clickedInside = useRef(false);
+
+  // Due to applying max height in some cases, it can take more than one render to have the popover in its final desired size.
+  // This ref allows us to track such series of render cycles triggered by a click inside a popover.
+  const resizingAfterInternalClick = useRef(false);
```
pan-kot, line 70: What if the resize happens asynchronously? It can be 1ms after a click but then the same issue would occur - the popover will reposition. Not relevant for the upcoming feature but does not sound very reliable.

Do we have a use case where the reposition of the chart popover is actually needed? If not, how about adding a flag `keepPosition` to the internal popover and set it to `true` in all chart popovers. When set, the popover would calculate its placement only once upon the first render, or, if we keep the popover open when navigating between series we can possibly also reposition when trackKey changes.

src/mixed-line-bar-chart/__tests__/mixed-chart.test.tsx
```diff
@@ -11,6 +11,18 @@ import chartWrapperStyles from '../../../lib/components/internal/components/char
 import { lineSeries3 } from './common';
 import createComputedTextLengthMock from './computed-text-length-mock';
 import { KeyCode } from '@cloudscape-design/test-utils-core/dist/utils';
+import { CalculatePositionProps } from '../../../lib/components/popover/utils/positions';
+
+let mockCalculatePosition: jest.Mock | undefined;
+jest.mock('../../../lib/components/popover/utils/positions', () => {
```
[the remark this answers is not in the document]
jperals: Coverage. This allows the tests to run past [this condition](https://github.com/cloudscape-design/components/pull/1756/files#diff-10a52bf3488068cab46d14def970625a0a8a12f9ed574481790843a9205c94a6L78).

src/popover/container.tsx
jperals: > What if the resize happens asynchronously? It can be 1ms after a click but then the same issue would occur - the popover will reposition. Not relevant for the upcoming feature but does not sound very reliable.

If a resize happens for another reason (e.g, an asynchronous content) unrelated to user interaction, even if it happens to happen almost at the same time, then it should reposition. If it happens due to that interaction, then it shouldn't. But it's true that the relationship between the click and the resize on which we base preventing the reposition is a bit fragile (I'm using `requestAnimationFrame` as a proxy for "when the cycle of renders triggered by this interaction has ended", but I can't find a better way to do exactly this with the React official toolset).

> Do we have a use case where the reposition of the chart popover is actually needed?

The use case that I was trying to keep covered was internal asynchronous content updates that could change the popover position without user interaction. Like in [this page](https://d21d5uik3ws71m.cloudfront.net/components/c8cb419f57999afa860ddd22d9a1ca92c919af18/index.html#/light/popover/scrollable). But I guess that is quite far fetched especially for chart popovers. So I'm switching back to  basically ignore popover resize events altogether in chart popovers, as you describe, which was also pretty much the initial approach of this PR.

> if we keep the popover open when navigating between series we can possibly also reposition when trackKey changes

Yes, this is already taken care of [here](https://github.com/cloudscape-design/components/pull/1756/files#diff-10a52bf3488068cab46d14def970625a0a8a12f9ed574481790843a9205c94a6R175).

src/mixed-line-bar-chart/__tests__/mixed-chart.test.tsx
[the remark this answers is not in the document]
pan-kot: Would it be possible to create a shared mock somewhere under `src/popover/__tests__` and use it instead?

src/mixed-line-bar-chart/__tests__/mixed-chart.test.tsx
```diff
@@ -966,7 +991,7 @@ describe('Details popover', () => {
     wrapper.findChart()!.fireEvent(new MouseEvent('mousedown', { bubbles: true }));
 
     wrapper.findDetailPopover()!.findDismissButton()!.keydown(KeyCode.escape);
-    jest.runAllTimers();
+    act(() => void jest.runAllTimers());
```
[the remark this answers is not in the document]
pan-kot: It looks strange that act is used to wrap the runAllTimers helper. Can we improve it by replacing the timer utils with the waitFor?

src/popover/container.tsx
```diff
@@ -47,145 +48,186 @@ export default function PopoverContainer({
...
+      ) {
+        return;
+      }
+
+      requestAnimationFrame(() => {
+        updatePositionHandler();
+      });
+    };
+
+    const onKeyUp = (event: KeyboardEvent) => {
```
pan-kot, line 206: If the click event is no longer recorded, does it make sense to introduce the new onKeyUp handler in scope of this PR?

src/mixed-line-bar-chart/__tests__/mixed-chart.test.tsx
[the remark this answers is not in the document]
jperals: Done [here](https://github.com/cloudscape-design/components/pull/1756/commits/c62dd313e1adef98cbc4f464341645dfe939bf6d).

src/mixed-line-bar-chart/__tests__/mixed-chart.test.tsx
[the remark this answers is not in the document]
jperals: Done [here](https://github.com/cloudscape-design/components/pull/1756/commits/c7a11f6e9a54b52138c41d8a4316142c32eb1d43). Interstingly this also makes these tests around 15% faster (80ms vs 90ms and 67ms vs 83ms). I'm guessing not all timers need to run to  get to the state where the assertions pass.

src/popover/container.tsx
```diff
@@ -47,145 +48,186 @@ export default function PopoverContainer({
   size,
   fixedWidth,
   variant,
+  keepPositionOnResize,
```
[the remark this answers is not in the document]
jperals: I think repositioning when the window size changes still makes sense.

Summary of the current behavior:
- On window resize, go through the full loop of recalculating position and eventually making scrollable (unchanged)
- On scroll, reposition but with the precalculated position handler (unchanged)
- On trackKey change, go through the full loop (unchanged)
- On click, if it is inside the popover do nothing (changed), if it is unchanged go through the full loop as it used to do before
- On keyup, if it is space or enter do the same as on click (new)
- On popover content resize, do not reposition in any case (changed)

src/mixed-line-bar-chart/__tests__/mixed-chart.test.tsx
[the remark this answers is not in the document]
jperals: Reverted because unfortunately the mock stopped working as soon as it was in another file.

src/popover/container.tsx
jperals: Just removed. In fact, not every space or enter event is equivalent to a click (the user could be typing in an input, for example).

src/popover/container.tsx
```diff
@@ -47,145 +48,181 @@ export default function PopoverContainer({
...
-      TODO: extend this to Enter and Spacebar?
+    TODO: extend this to Enter and Spacebar?
     */
-    const updatePosition = () => requestAnimationFrame(() => updatePositionHandler());
+
+    const onClick = (event: UIEvent | KeyboardEvent) => {
+      if (
+        // Do not update position if the click was on the popover itself
+        // and keepPositionOnResize is true.
+        (keepPositionOnResize && nodeContains(popoverRef.current, event.target)) ||
```
pan-kot, line 195: The click has now nothing to do with resize. Should we introduce a separate flag like keepPositionOnClick?

src/popover/container.tsx
```diff
@@ -47,145 +48,181 @@ export default function PopoverContainer({
   size,
   fixedWidth,
   variant,
+  keepPositionOnResize,
 }: PopoverContainerProps) {
   const bodyRef = useRef<HTMLDivElement | null>(null);
   const contentRef = useRef<HTMLDivElement | null>(null);
   const popoverRef = useRef<HTMLDivElement | null>(null);
   const arrowRef = useRef<HTMLDivElement | null>(null);
+  const internalPositionRef = useRef<InternalPosition | null>(null);
```
pan-kot, line 57: This one is shaky. It works because every time the internal position is set a render as triggered by also setting the popoverStyle.

I would rather keep the state as is and introduce a new variable previousInternalPositionRef.

```
const shouldKeepPosition = keepPositionOnResize && onResize;
const fixedInternalPosition = shouldKeepPosition && previousInternalPositionRef.current;

// ...
        
if (!shouldKeepPosition) {
  setInternalPosition(newInternalPosition);
}
setPopoverStyle(...);
```

wdyt?

src/popover/container.tsx
```diff
@@ -47,145 +48,181 @@ export default function PopoverContainer({
...
-    const containingBlockRect = containingBlock ? containingBlock.getBoundingClientRect() : viewportRect;
-
-    const bodyBorderWidth = getBorderWidth(body);
-    const contentRect = contentRef.current.getBoundingClientRect();
-    const contentBoundingBox = {
-      width: contentRect.width + 2 * bodyBorderWidth,
-      height: contentRect.height + 2 * bodyBorderWidth,
-    };
+  const updatePositionHandler = useCallback(
+    (onResize = false) => {
```
pan-kot, line 67: The name of the variable (onResize) is confusing: the handler is called in two resize events: window resize and content resize. I would make it more specific like `contentResizeObserved`

src/popover/container.tsx
jperals: Hm, not very keen to introduce yet another prop, but I get your point. Maybe I'd rather change its name, like `keepPosition`.

src/popover/container.tsx
jperals: What about `onContentResize`?

src/popover/container.tsx
```diff
@@ -191,8 +191,8 @@ export default function PopoverContainer({
 
     const onClick = (event: UIEvent | KeyboardEvent) => {
       if (
-        // Do not update position if the click was on the popover itself and keepPosition is true.
-        (keepPosition && nodeContains(popoverRef.current, event.target)) ||
+        // Do not update position if keepPosition is true.
+        keepPosition ||
```
jperals, line 199: Simplified this condition to not check whether the event was in the popover, because when dismissing, there is a frame before where the dismiss button disappears while the popover is still visible. This would cause `nodeContains(popoverRef.current, event.target)` to be false and update the popover position undesirably for that one frame before disappearing.

For the use case that uses the `keepPosition` prop, we don't need to consider clicks outside the popover for resizing because such clicks would close it anyway.

src/popover/container.tsx
jperals: Got it, done it in [e61c627](https://github.com/cloudscape-design/components/pull/1756/commits/e61c627220c339f96c89589bcf30c0ce8e3d9169).

I am calling `setInternaPosition` unconditionally though because `calculatePosition` will always return the value of the passed-in `fixedInternalPosition` as `internalPosition` if it is defined, so the permanency of this value is already guaranteed this way.

src/popover/container.tsx
jperals: Changed to `onContentResize`

src/popover/container.tsx
```diff
@@ -47,145 +48,181 @@ export default function PopoverContainer({
...
 
-    window.addEventListener('click', updatePosition);
-    window.addEventListener('resize', updatePosition);
+    window.addEventListener('click', onClick);
+
+    window.addEventListener('resize', updatePositionOnResize);
     window.addEventListener('scroll', refreshPosition, true);
+
+    // Update the position on the first render, right before painting
+    updatePositionHandler();
```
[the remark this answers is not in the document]
jperals: Removed, thanks. I must have introduced this change in the very beginning and forgot about it.

src/popover/container.tsx
pan-kot: Do we still need to check for target? I think that is merely a performance optimisation which looks unnecessary

pan-kot (approved):

src/popover/container.tsx
jperals: True that it's not that related to the PR, but it prevents unnecessary recalculations and re-renders.

PR #1767, opened 2023-11-30
chore: Refine VR dark mode styles and replace shadows with strokes

at-susie:
Make updates on border color, text color, and background color for selectable item and tiles component.

### Description

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/box/internal.tsx
```diff
@@ -43,7 +44,8 @@ export default function InternalBox({
     styles[`color-${color || 'default'}`],
     styles[`font-size-${fontSize || 'default'}`],
     styles[`font-weight-${fontWeight || 'default'}`],
-    styles[`t-${textAlign}`]
+    styles[`t-${textAlign}`],
+    isRefresh && styles.refresh
```
jperals, line 48: Is this one necessary? I don't see any `.refresh` selector in the stylesheets of this component.

at-susie: We do no longer need these changes in box component. I originally added to differentiate the color from h1 to h5 variant in box component, but I figured out that it's not necessary. 

I'll remove them and push the change.

style-dictionary/visual-refresh/metadata/colors.ts
```diff
@@ -273,6 +273,11 @@ const metadata: StyleDictionary.MetadataIndex = {
     public: true,
     themeable: true,
   },
+  colorBorderDividerDimmed: {
+    description: 'The border color for table row dividers.',
```
jperals, line 277: It looks like this token is not used only for tables. What about just "The border color for table row dividers."?

style-dictionary/visual-refresh/metadata/colors.ts
```diff
@@ -299,6 +304,12 @@ const metadata: StyleDictionary.MetadataIndex = {
     public: true,
     themeable: true,
   },
+  colorBorderSelectableItemFocused: {
+    description:
+      'The color of focus states for dropdown items, for example, the focus ring around selectable elements in the dropdown of button dropdown, select, and multi-select.',
```
jperals, line 309: I would use a full stop instead of comma before "for example", like in the previous token descriptions: "The color of focus states for dropdown items. For example, ..."

style-dictionary/visual-refresh/metadata/colors.ts
at-susie: It's also supposed to be used for [the service homepage](https://refresh.cloudscape.aws.dev/examples/react/landing-page.html?awsui-visual-refresh=true).

style-dictionary/visual-refresh/metadata/colors.ts
at-susie: Agreed! Will change it

style-dictionary/visual-refresh/metadata/colors.ts
```diff
@@ -273,6 +273,11 @@ const metadata: StyleDictionary.MetadataIndex = {
     public: true,
     themeable: true,
   },
+  colorBorderDividerDimmed: {
```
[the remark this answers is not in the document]
jperals: Those existing tokens are private (not public tokens), so it's a bit different because:
1. They can't be themed, so using a appearance-specific term like "dimmed" or a color name is OK
2. It doesn't matter that much in general, because we can change this name at any time without affecting anyone.

Once a token is public, then their name becomes part of a public API. So it's a good time to rethink its name :)

style-dictionary/visual-refresh/metadata/colors.ts
at-susie: Now I got what you mean. I'll then change it to `colorBorderRowDivider`. Make sense?

src/header/styles.scss
```diff
@@ -137,6 +137,7 @@
     // whether there are buttons present or not; otherwise configurable dashboard handles are misaligned.
     &.refresh {
       padding-top: space-heading-button-diff(awsui.$line-height-heading-xl);
+      color: awsui.$color-text-heading-primary;
```
jperals, line 140: Should we apply this to Classic as well? Classic inherits the design tokens from Visual Refresh, which means that Classic users will also see this `color-text-heading-primary` token listed, with the description "The default color for heading 1 (...)", so they will expect to be able to theme primary headings with it but it will have no effect for them.

src/internal/components/selectable-item/styles.scss
```diff
@@ -82,6 +82,20 @@
     box-shadow: inset 0 0 0 awsui.$border-control-focus-ring-shadow-spread awsui.$color-border-item-focused;
   }
 
+  &.refresh {
```
jperals, line 85: Same comment here: should we make Classic themeable too?

style-dictionary/visual-refresh/metadata/colors.ts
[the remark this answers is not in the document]
at-susie: Had offline discussion, and we decided to go with `colorBorderDividerSecondary`. This way the color token can be more scalable for the future. Now this token is used for the row dividers for table and collection preferences as horizontal dividers, but it could be used for vertical dividers in the future.

src/internal/styles/forms/mixins.scss
```diff
@@ -147,7 +147,6 @@
     @include utils.base-pseudo-element;
```
jperals, line 147: Since this is not a shadow anymore, I would give a try at making this an actual outline, and it looks like actually this enter `&::after` block can be replaced with
```
outline: 1px solid awsui.$color-grey-350;
```
and we can probably remove the specific shadow styling for the stacked container in `src/container/styles.scss` as well as the box shadow definitions in `style-dictionary/visual-refresh/shadows.ts`.

I would also rename the mixin from `container-shadow` to `container-border`.

style-dictionary/visual-refresh/metadata/colors.ts
```diff
@@ -299,6 +304,12 @@ const metadata: StyleDictionary.MetadataIndex = {
     public: true,
     themeable: true,
   },
+  colorBorderSelectableItemFocused: {
```
jperals, line 307: Would `colorBorderDropdownItemFocused` be a better name? We already have 6 public design tokens using the term `DropdownItem`, like `colorBorderDropdownItemDefault` or `colorBorderDropdownItemHover`. If this refers to the same kind of item, I think it should be named as `DropdownItem` too.

src/cards/styles.scss
```diff
@@ -98,7 +102,7 @@
     > .card-inner {
       background-color: awsui.$color-background-item-selected;
       &::before {
-        border: awsui.$border-item-width solid awsui.$color-border-item-selected;
+        outline: awsui.$border-item-width solid awsui.$color-border-item-selected;
```
abdhalees, line 105: why was this changed to outline in both themes?

src/container/styles.scss
```diff
@@ -44,13 +48,9 @@
...
     }
     // Replace container border with a divider
-    &-stacked + &-stacked::before {
+    &-stacked + &-stacked:not(.refresh)::before {
       @include shared.divider;
     }
-    // Reset container shadow to prevent unwanted overflow
-    &-stacked + &-stacked::after {
-      box-shadow: awsui.$shadow-container-stacked;
-    }
```
abdhalees, line 53: why was this removed?

src/internal/components/selectable-item/styles.scss
```diff
@@ -64,6 +64,7 @@
   }
 
   &.selected {
+    z-index: 2;
```
abdhalees, line 67: I don't think we should be playing with the z-index in dropdowns, why is this needed?

src/internal/styles/forms/mixins.scss
```diff
@@ -147,7 +147,32 @@
...
+    z-index: 1;
+  }
+
+  &::after {
+    @include utils.base-pseudo-element;
+    outline: 1px solid awsui.$color-border-divider-default;
+    border-radius: awsui.$border-radius-container;
   }
 
   // HACK: because of the issues with IE's box shadow rendering we draw
```
abdhalees, line 178: let's remove this from the new mixin, we don't support IE11 anymore

src/header/styles.scss
```diff
@@ -127,11 +127,13 @@
 
 .title {
   @include styles.text-wrapping;
-  color: awsui.$color-text-heading-default;
-
+  &:not(&-variant-h1) {
```
jkuelz, line 130: I thought we decided to make all headings the same color?

src/internal/styles/forms/mixins.scss
```diff
@@ -147,13 +147,26 @@
...
+    z-index: 1;
   }
 
-  // HACK: because of the issues with IE's box shadow rendering we draw
-  // a border and don't mind the pixel fitting
-  @media screen and (-ms-high-contrast: active) {
-    border: constants.$control-border-width solid awsui.$color-border-container-top;
+  &::after {
+    @include utils.base-pseudo-element;
+    outline: 1px solid awsui.$color-border-divider-default;
```
jkuelz, line 168: The `@container-shadow` mixin is intended to provide the style definition for containers, whether that's via shadow, border, or both. 
Since we removed the box shadow definitions in `style-dictionary/visual-refresh/shadows.ts`, and this seems to be the only line that is different between these two mixins, can we create/use a token for this border instead of creating a whole new mixin for just the border? That way it can be used for both classic and VR without needing to add the `isVisualRefresh` hook in downstream components like container, cards, modal, etc. We can rename the mixin to something more generic like @container-style or @container-outline if that helps.

style-dictionary/visual-refresh/shadows.ts
```diff
@@ -5,30 +5,21 @@ import { StyleDictionary } from '../utils/interfaces';
...
-    dark: '0px 1px 8px 2px rgba(0, 7, 22, 0.6)',
-  },
-  shadowContainerStacked: {
-    light:
-      '-1px 1px 1px 0px #e9ebed, 1px 1px 1px 0px #e9ebed, 0px 9px 8px -7px rgb(0 7 22 / 12%), 8px 0px 8px -7px rgb(0 7 22 / 12%), -8px 0px 8px -7px rgb(0 7 22 / 12%)',
-    dark: '0px 9px 8px -7px rgb(0 7 22 / 60%), 8px 0px 8px -7px rgb(0 7 22 / 60%), -8px 0px 8px -7px rgb(0 7 22 / 60%)',
-  },
   shadowContainerActive: {
     light: '0px 1px 1px 1px #e9ebed, 0px 6px 36px #0007161a',
     dark: '0px 1px 1px 1px #192534, 0px 6px 36px #000716',
```
jkuelz, line 10: We want to update the shadow for board component containers in their dragged state, right? At least for dark mode?

jkuelz, line 10: We want to update the shadow for board component containers in their dragged state, right? At least for dark mode?

src/container/shared.scss
```diff
@@ -19,6 +19,10 @@ $header-focus-visible-padding: calc(#{awsui.$space-scaled-s} - #{awsui.$border-d
   @include styles.container-shadow;
 }
 
+@mixin borders {
+  @include styles.container-border;
```
jkuelz, line 23: It's not quite clear why this mixin wouldn't just be part of the @borders-and-shadows mixin above? Or why either of these mixins are needed, as they are just wrappers for individual mixins.

style-dictionary/classic/colors.ts
```diff
@@ -104,6 +105,7 @@ const tokens: StyleDictionary.ColorsDictionary = {
   colorTextExpandableSectionDefault: '{colorTextInteractiveDefault}',
   colorTextExpandableSectionHover: '{colorTextInteractiveHover}',
   colorTextExpandableSectionNavigationIconDefault: '{colorTextIconCaret}',
+  colorTextHeadingPrimary: { light: '{colorGrey900}', dark: '{colorGrey200}' },
```
jkuelz, line 108: We also need to add `colorTextHeadingDefault:  { light: '{colorGrey900}', dark: '{colorGrey200}' },` if we don't want the changes to apply to classic as well (which I don't think we do).

style-dictionary/visual-refresh/colors.ts
```diff
@@ -143,8 +145,8 @@ const tokens: StyleDictionary.ColorsDictionary = {
   colorStrokeCodeEditorGutterActiveLineDefault: { light: '{colorGrey300}', dark: '{colorGrey750}' },
   colorStrokeCodeEditorGutterActiveLineHover: { light: '{colorGrey100}', dark: '{colorGrey900}' },
   colorTextAccent: { light: '{colorBlue600}', dark: '{colorBlue500}' },
-  colorTextBodyDefault: { light: '{colorGrey900}', dark: '{colorGrey300}' },
-  colorTextBodySecondary: { light: '{colorGrey600}', dark: '{colorGrey300}' },
```
jkuelz, line 147: We need to add these definitions to `style-dictionary/classic/colors.ts` definition, otherwise the new values will update in classic as well. Previously they were the same between classic and VR, so they were just being inherited.

src/internal/components/selectable-item/styles.scss
```diff
@@ -72,13 +73,20 @@
     }
   }
 
+  &.refresh {
+    &.selected {
+      border-color: awsui.$color-border-dropdown-item-selected;
+    }
+  }
+
```
jkuelz, line 81: Why is this refresh class needed? Is there a reason this isn't just a design token?

abdhalees (approved):

PR #1777, opened 2023-12-05
fix: Chart detail popover should open when focus moves out and in again

YueyingLu:
### Description

When focus moves out, the popover is closed, see https://github.com/cloudscape-design/components/blob/c17dd790be5b158190eb295036b34da22992098e/src/mixed-line-bar-chart/chart-container.tsx#L392-L410

The highlightedX is not updated at this moment. So when the focus moves back, the popover not show up https://github.com/cloudscape-design/components/blob/c17dd790be5b158190eb295036b34da22992098e/src/mixed-line-bar-chart/chart-container.tsx#L345-L349

The issue is only when using keyboard. The change opens the Popover when focus moves back.

Related links, issue #AWSUI-24744

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/components/chart-plot/index.tsx
+++ b/src/internal/components/chart-plot/index.tsx
@@ -48,6 +48,7 @@ export interface ChartPlotProps {
   onFocus?: (event: React.FocusEvent<SVGGElement>, trigger: 'mouse' | 'keyboard') => void;
   onBlur?: (event: React.FocusEvent<SVGGElement>) => void;
   onKeyDown?: (event: React.KeyboardEvent<SVGGElement>) => void;
+  showPopoverOfHighlightedItem?: () => void;
   children: React.ReactNode;
 }
 
@@ -87,6 +88,7 @@ function ChartPlot(
     onKeyDown,
     onFocus,
     onBlur,
+    showPopoverOfHighlightedItem,
     focusOffset = DEFAULT_PLOT_FOCUS_OFFSET,
     activeElementFocusOffset = DEFAULT_ELEMENT_FOCUS_OFFSET,
     ...restProps
@@ -137,6 +139,11 @@ function ChartPlot(
       const codes = [KeyCode.space, KeyCode.enter, KeyCode.up, KeyCode.left, KeyCode.right, KeyCode.down];
       if (codes.indexOf(event.keyCode) !== -1) {
         applicationRef.current!.focus();
+        // when focus on the plot, check if there is highlighted item and show popover
+        // because if focus moves to already highlighted item, highlightedX stays the same, popover may not show up
+        if (showPopoverOfHighlightedItem) {
+          showPopoverOfHighlightedItem();
+        }
       }
     }
   };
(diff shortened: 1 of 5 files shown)


abdhalees (changes requested):
there is a failing integ test

src/internal/components/chart-plot/index.tsx
```diff
@@ -137,6 +139,11 @@ function ChartPlot(
       const codes = [KeyCode.space, KeyCode.enter, KeyCode.up, KeyCode.left, KeyCode.right, KeyCode.down];
       if (codes.indexOf(event.keyCode) !== -1) {
         applicationRef.current!.focus();
+        // when focus on the plot, check if there is highlighted item and show popover
+        // because if focus moves to already highlighted item, highlightedX stays the same, popover may not show up
+        if (showPopoverOfHighlightedItem) {
+          showPopoverOfHighlightedItem();
```
YueyingLu, line 145: The change was covered by integration test.

src/mixed-line-bar-chart/chart-container.tsx
```diff
@@ -504,6 +504,12 @@ export default function ChartContainer<T extends ChartDataTypes>({
 
   const isRefresh = useVisualRefresh();
 
+  const showPopoverOfHighlightedItem = () => {
+    if (!isPopoverOpen && (highlightedX !== null || highlightedPoint !== null)) {
+      showPopover();
```
YueyingLu, line 509: The change is covered by integration test.

abdhalees (approved):

jperals:
This component has quite a few state variables that have to be kept in sync, which is prone to bugs like this, so it might be possible to do this by removing logic instead of adding. An alternative fix is to keep `isPopoverOpen` to `true` when focusing out, but only render it if there are highlighted elements, see this POC: https://github.com/cloudscape-design/components/pull/1821/files

YueyingLu:
> This component has quite a few state variables that have to be kept in sync, which is prone to bugs like this, so it might be possible to do this by removing logic instead of adding. An alternative fix is to keep `isPopoverOpen` to `true` when focusing out, but only render it if there are highlighted elements, see this POC: https://github.com/cloudscape-design/components/pull/1821/files

Thank you for the POC. When tabbing out the chart, the highlightedItem still exists, but the popover should close. This approach will keep the Popover open, even has tabbed out the chart by keyboard.

pan-kot:
The issue only exists in bar charts but not in area-, line- or pie charts. Why? In all charts there is state that keeps the highlighted element yet no workaround is required. My guess is that we clear the state on blur and a consistent approach is to ensure the same for the bar chart.

pan-kot (approved):

src/mixed-line-bar-chart/chart-container.tsx
```diff
@@ -401,11 +398,8 @@ export default function ChartContainer<T extends ChartDataTypes>({
       !(blurTarget instanceof Element) ||
       !nodeBelongs(containerRefObject.current, blurTarget)
     ) {
-      setHighlightedPoint(null);
+      clearHighlightedSeries();
```
YueyingLu, line 401: Bar chart will not remember the highlighted item when tab out. However this is not consistent behavior between Bar, Line and Area charts.

PR #1782, opened 2023-12-06
feat: Copy to clipboard component

pan-kot:
### Description

Introducing copy to clipboard as a coded component.

### How has this been tested?

New unit tests and a page for screenshot tests.

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/copy-to-clipboard/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+        wrapText={false}
+        formAction="none"
+      />
+    </InternalPopover>
+  );
+
+  return (
+    <div ref={__internalRootRef} {...baseProps} className={clsx(baseProps.className, styles.root, testStyles.root)}>
+      {variant === 'inline' ? (
+        <div className={styles['inline-container']}>
```
just-boris, line 85: If this is inline, why are these divs? What about appropriate HTML nesting? Such as using copy-to-clipboard inside `<p>`?

src/copy-to-clipboard/styles.scss
```diff
@@ -0,0 +1,23 @@
...
+.inline-container {
+  display: inline-block;
+  word-break: break-all;
+
+  &-trigger {
+    display: inline-block;
+    margin-inline-end: awsui.$space-scaled-xxs;
+  }
+  &-text {
+    display: contents;
```
just-boris, line 21: How many of these styles would be needed after addressing the comment above about `<div>`?

src/popover/internal.tsx
```diff
@@ -141,7 +143,7 @@ function InternalPopover(
       className={clsx(popoverClasses, styles['popover-content'])}
       data-awsui-referrer-id={referrerId}
     >
-      {visible && (
+      {visible && !__hidden && (
```
just-boris, line 146: On the API discussion, there was a question if this private API in the scope of this feature or not. Please discuss with the team before pushing it.

src/copy-to-clipboard/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+  const triggerVariant = (
+    {
+      button: 'normal',
+      icon: 'icon',
+      inline: 'inline-icon',
+    } as const
+  )[variant];
+
+  const trigger = (
+    <InternalPopover
```
just-boris, line 63: Should we set `renderWithPortal=true` from the start to support correct positions inside table, for example?

src/test-utils/dom/copy-to-clipboard/index.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  findCopyButton(): ButtonWrapper {
+    return this.findComponent(`.${ButtonWrapper.rootSelector}`, ButtonWrapper)!;
+  }
+
+  findStatusText(): null | ComponentWrapper {
+    return this.findComponent(`.${PopoverWrapper.rootSelector}`, PopoverWrapper)!.findContent();
+  }
+
+  findTextToCopy(): null | ComponentWrapper {
```
just-boris, line 19: Why do these return a `ComponentWrapper`?

src/copy-to-clipboard/index.tsx
pan-kot: Will update

src/test-utils/dom/copy-to-clipboard/index.ts
pan-kot: Should be ElementWrapper, will update 👍

src/popover/internal.tsx
pan-kot: Removed this for now. We can add it later if really needed.

src/copy-to-clipboard/styles.scss
just-boris: @pan-kot this question remains open. Did you try cleaning up these styles?

just-boris (commented):
[One comment](https://github.com/cloudscape-design/components/pull/1782#discussion_r1427863827) from the previous review left unaddressed

src/copy-to-clipboard/styles.scss
pan-kot: Thanks, updated now

just-boris (approved):

just-boris (approved):

PR #1788, opened 2023-12-06
chore: Migrate "B/C/D/E/F" components to logical properties

DaemonCahill:
This PR is a subset of the changes drafted in #1713 that affect only the subset of components that being with the letters "B/C/D/E/F".

### DIFF
--- a/src/container/shared.scss
+++ b/src/container/shared.scss
@@ -7,8 +7,6 @@
 @use '../internal/styles/tokens' as awsui;
 
 $header-padding-horizontal: awsui.$space-container-horizontal;
-$header-padding: awsui.$space-container-header-top awsui.$space-container-horizontal
-  awsui.$space-container-header-bottom;
 $footer-padding: awsui.$space-scaled-s awsui.$space-container-horizontal;
 
 // HACK: Remediate focus border for expandable section


src/container/styles.scss
```diff
@@ -156,23 +162,27 @@
...
   }
 
   &:not(:empty) {
-    border-bottom: awsui.$border-container-sticky-width solid awsui.$color-border-container-divider;
+    border-block-end: awsui.$border-container-sticky-width solid awsui.$color-border-container-divider;
   }
 
   &.with-paddings {
-    padding: shared.$header-padding;
+    padding-block-start: awsui.$space-container-header-top;
```
pan-kot, line 173: Should we now remove the shared padding variables?

src/container/styles.scss
```diff
@@ -156,23 +162,27 @@
...
+    margin-block-end: calc(#{awsui.$line-height-heading-xl} - #{awsui.$line-height-heading-l});
   }
 
   &:not(:empty) {
-    border-bottom: awsui.$border-container-sticky-width solid awsui.$color-border-container-divider;
+    border-block-end: awsui.$border-container-sticky-width solid awsui.$color-border-container-divider;
   }
 
   &.with-paddings {
-    padding: shared.$header-padding;
```
pan-kot, line 167: Should we now remove the shared padding variables?

pan-kot (dismissed):

src/container/styles.scss
DaemonCahill: Removed!

pan-kot (approved):

PR #1792, opened 2023-12-07
chore: Migrate "G/H/I/J/K/L/M" components to logical properties

DaemonCahill:
This PR is a subset of the changes drafted in https://github.com/cloudscape-design/components/pull/1713 that affect only the subset of components that being with the letters "G/H/I/J/K/L/M".

### DIFF
--- a/src/internal/components/chart-series-details/styles.scss
+++ b/src/internal/components/chart-series-details/styles.scss
@@ -75,7 +75,7 @@ $font-weight-bold: awsui.$font-weight-heading-s;
 
 .list:not(.compact) > .list-item:not(:first-child),
 .inner-list-item {
-  margin-top: awsui.$space-scaled-xxs;
+  margin-block-start: awsui.$space-scaled-xxs;
 }
 
 .list-item.with-sub-items:not(.expandable) > .key-value-pair {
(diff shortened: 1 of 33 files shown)


src/internal/components/chart-series-details/styles.scss
```diff
@@ -55,8 +55,10 @@ $font-weight-bold: awsui.$font-weight-heading-s;
...
 .list {
   display: flex;
-  margin: 0;
-  padding: 0;
+  margin-block: 0;
+  margin-inline: 0;
+  padding-block: 0;
+  padding-inline: 0;
   list-style: none;
   flex-direction: column;
```
Al-Dani, line 64: this file also has margin-top
https://github.com/cloudscape-design/components/blob/a011f57bf4c17a6ae236449b9562093df20a19ab/src/internal/components/chart-series-details/styles.scss#L78

and width 
https://github.com/cloudscape-design/components/blob/a011f57bf4c17a6ae236449b9562093df20a19ab/src/internal/components/chart-series-details/styles.scss#L89

Al-Dani (changes requested):

src/internal/components/chart-series-details/styles.scss
DaemonCahill: Good find! I wonder why those weren't flagged by the linter. Investigating...

Al-Dani (approved):

Al-Dani (approved):

PR #1793, opened 2023-12-08
chore: Migrate "N/O/P/Q/R/S" components to logical properties

DaemonCahill:
This PR is a subset of the changes drafted in https://github.com/cloudscape-design/components/pull/1713 that affect only the subset of components that being with the letters "N/O/P/Q/R/S".

### DIFF
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -51,7 +51,15 @@
   },
   "overrides": [
     {
-      "files": ["./src/button/*.scss"],
+      "files": [
+        "./src/button/*.scss",
+        "./src/n*/*.scss",
+        "./src/o*/*.scss",
+        "./src/p*/*.scss",
+        "./src/q*/*.scss",
+        "./src/r*/*.scss",
+        "./src/s*/*.scss"
+      ],
       "rules": {
         "property-disallowed-list": ["border", "border-radius", "border-style", "margin", "padding"],
         "csstools/use-logical": "always"
--- a/src/split-panel/styles.scss
+++ b/src/split-panel/styles.scss
@@ -188,7 +188,7 @@ $app-layout-drawer-width: calc(#{awsui.$space-layout-toggle-diameter} + 2 * #{aw
   flex-direction: column;
   z-index: 1; // should be above .content-bottom
   padding-block: 0;
-  padding-inline: awsui.$space-scaled-2x-xxxl;
+  padding-inline: awsui.$space-layout-content-horizontal;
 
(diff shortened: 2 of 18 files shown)


src/split-panel/styles.scss
```diff
@@ -100,16 +97,19 @@ $app-layout-drawer-width: calc(#{awsui.$space-layout-toggle-diameter} + 2 * #{aw
...
 }
 
 .slider {
-  padding: 0;
+  padding-block: 0;
+  padding-inline: 0;
   cursor: ns-resize;
-  margin: 2px 0 0 0;
-  height: $slider-width;
+  margin-block-start: 2px;
```
taheramr, line 109: Can we use the same shorthand syntax like `margin-block: 2px 0` and `margin-inline: 0 0`?

src/split-panel/styles.scss
```diff
@@ -125,26 +125,28 @@ $app-layout-drawer-width: calc(#{awsui.$space-layout-toggle-diameter} + 2 * #{aw
 
 .slider-side {
   cursor: ew-resize;
-  margin: 0 0 0 2px;
+  margin-block: 0;
```
taheramr, line 128: Same to my previous comment, looking at the codebase there is a lot of instances in which we transform `margin: x y z f` into separate `margin-block-[start|end]` and `margin-inline-[start|end]`, would be nice if we kept it as consistent as it was, wdyt?

taheramr:
The changes are generally looking good (left a couple of small comments), I tried to test a couple of dev pages to verify that the proper logical properties were applied but this is too big of a change to be verified only as a PR.

I'm suggesting you can run some-sort of a quick bug-bash for each subset of components so people can make sure incorrect spacings or properties being used, I myself get confused a lot between `[propertyName]-block` and `[propertyName]-inline`, maybe can you also share how you did the change? was there some kind of script that read all the proerties and changed them or was it all manual?

.stylelintrc
```diff
@@ -51,7 +51,15 @@
   },
   "overrides": [
     {
-      "files": ["./src/button/*.scss"],
+      "files": [
+        "./src/button/*.scss",
+        "./src/n*/*.scss",
```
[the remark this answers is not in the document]
DaemonCahill: Yes!

src/split-panel/styles.scss
DaemonCahill: Yes!

taheramr (approved):

PR #1794, opened 2023-12-08
chore: Migrate "T/U/V/W/X/Y/Z" components to logical properties

DaemonCahill:
This PR is a subset of the changes drafted in https://github.com/cloudscape-design/components/pull/1713 that affect only the subset of components that being with the letters "T/U/V/W/X/Y/Z".

### DIFF
--- a/src/table/body-cell/styles.scss
+++ b/src/table/body-cell/styles.scss
@@ -27,10 +27,12 @@ $success-icon-padding-right: calc(#{$edit-button-padding-right} + #{$icon-width-
 
 .body-cell {
   box-sizing: border-box;
-  padding: $cell-vertical-padding $cell-horizontal-padding $cell-vertical-padding-w-border;
-  border-top: awsui.$border-divider-list-width solid transparent;
+  padding-block-start: $cell-vertical-padding;
+  padding-block-end: $cell-vertical-padding-w-border;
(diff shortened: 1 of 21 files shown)


pan-kot (dismissed):

src/table/body-cell/styles.scss
```diff
@@ -315,35 +329,31 @@ $success-icon-padding-right: calc(#{$edit-button-padding-right} + #{$icon-width-
 
       &:hover {
         background-color: awsui.$color-background-dropdown-item-hover;
-        border: awsui.$border-divider-list-width solid awsui.$color-border-editable-cell-hover;
-        left: calc(-1 * #{awsui.$border-divider-list-width});
-        right: calc(-1 * #{awsui.$border-divider-list-width});
-
+        border-block: awsui.$border-divider-list-width solid awsui.$color-border-editable-cell-hover;
+        border-inline: awsui.$border-divider-list-width solid awsui.$color-border-editable-cell-hover;
```
michaeldowseza, line 333: I guess this is fine so I assume it was done by eslint. Maybe a bit unnecessary in this case since the border is uniform anyway so doesn't really benefit from logical properties.

michaeldowseza (approved):

PR #1800, opened 2023-12-13
feat: Add foundational generative AI sparkles icons

jkuelz:
### Description

Adds two icons for gen AI foundation: one filled, and one un-filled. The filled icon should only be used in the 'small' size, otherwise the unfilled version is the default.

### How has this been tested?

Building in pipeline AwsUi-v3-katiegeo

<details>

#### Correctness

#### Security

#### Testing

</details>

---


pages/icon/size-inherit.page.tsx
```diff
@@ -15,6 +16,7 @@ export default function IconScenario() {
   const [height, setHeight] = useState(16);
   const [visibility, setVisibility] = useState(false);
   const unitHeight = height + UNITS;
+  const [iconName, setIconName] = useState('settings' as IconProps.Name);
```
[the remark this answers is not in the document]
just-boris: Also, I would recommend using url params instead of state, as it shown here: https://github.com/cloudscape-design/components/blob/d48075d6af501105cc5ad6b7e9e4a16b400ab142/pages/content-layout/without-header.page.tsx#L20

This will come in handy when you write screenshot tests, because you can open the desired icon by the direct url

pages/icon/text-align.page.tsx
```diff
@@ -14,29 +14,29 @@ export default function IconScenario() {
       </ul>
       <ScreenshotArea>
         <Box fontSize="display-l" fontWeight="light">
-          <Icon name="settings" size="large" /> Display large <Icon name="external" size="inherit" />
+          <Icon name="gen-ai" size="large" /> Display large <Icon name="external" size="inherit" />
```
just-boris, line 17: Why does the other page make the icon configurable and here we change the default? This will break screenshot tests, is it expected?

src/icon/internal.tsx
```diff
@@ -103,12 +103,20 @@ const InternalIcon = ({
 
   const validIcon = name && Object.prototype.hasOwnProperty.call(icons, name);
 
+  function getIconName(name: IconProps.Name) {
+    if (name === 'gen-ai') {
+      return iconSize === 'small' ? icons['gen-ai-filled'] : icons['gen-ai'];
+    } else {
+      return icons[name];
+    }
```
just-boris, line 111: This logic should have a unit test.

Also, this setup allows pulling the filled icon by its direct name `<Icon name="gen-ai-filled">`. Look at this screenshot testing page, it is there: https://d21d5uik3ws71m.cloudfront.net/components/f4463ee9af1facac5acb07a052aa4305e435fe2d/index.html#/light/icon/variant-success

just-boris: So I am thinking if we should abandon the automatic path or keep pushing it and solve these issues

jkuelz: I thought it was showing there because [that page](https://github.com/cloudscape-design/components/blob/main/pages/icon/icons-list.tsx#L19 ) uses unsafe `as` casting, which as you pointed out above, overrides type checks. Same reason it shows up in the icons dropdown list.

I think we should push forward and solve. I see two paths, we can either 
1) Rely on safe/correct typings and generate both the filled and unfilled version and conditionally apply them based on the size, or 
2) Figure out a way to just generate just one icon and add the fill class for the small size.

just-boris: There is a third way. We can pull custom SVGs directly in our code. Example: https://github.com/cloudscape-design/components/blob/main/src/split-panel/icons/side-position.tsx

We can keep `gen-ai` variant in the main icons list to make listing work and pull the special `gen-ai-filled` from the code

pages/icon/text-align.page.tsx
jkuelz: I was going to use this page for the screenshot tests, knowing that it would break screenshot tests on the first run. The other page was just to make it easier to see any icon in our dev environment.

themes/icons/gen-ai.svg
```diff
@@ -0,0 +1,4 @@
+<svg viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg">
+<path d="M7.98139 1.04717C7.98805 1.03029 8.01195 1.03029 8.01861 1.04717L9.9799 6.0201L14.9528 7.98139C14.9697 7.98805 14.9697 8.01195 14.9528 8.01861L9.9799 9.9799L8.01861 14.9528C8.01195 14.9697 7.98805 14.9697 7.98139 14.9528L6.0201 9.9799L1.04717 8.01861C1.03029 8.01195 1.03029 7.98805 1.04717 7.98139L6.0201 6.0201L7.98139 1.04717Z" />
+<path d="M2.5 2.65L2.54455 2.55455L2.64 2.51L2.54455 2.46545L2.5 2.37L2.45545 2.46545L2.36 2.51L2.45545 2.55455L2.5 2.65Z" class="filled"/>
```
just-boris, line 3: Just a small sanity check, is this little dot expected?

image

just-boris (approved):

themes/icons/gen-ai.svg
jkuelz: renders fine for me in the dev-page preview.
Screenshot 2023-12-21 at 4 03 37 PM

src/icon/internal.tsx
```diff
@@ -103,12 +103,32 @@ const InternalIcon = ({
...
 
+  function iconMap(name: IconProps.Name) {
+    const genAiFilledIcon = `<svg
+        width="12"
+        height="12"
+        viewBox="0 0 16 16"
+        xmlns="http://www.w3.org/2000/svg"
+        focusable="false"
+        aria-hidden="true"
+        data-testid="gen-ai-filled"
```
timogasda, line 114: What is this data-testid for? We support data attributes on our components. People could just add that if necessary.

src/icon/internal.tsx
```diff
@@ -103,12 +103,32 @@ const InternalIcon = ({
 
   const validIcon = name && Object.prototype.hasOwnProperty.call(icons, name);
 
+  function iconMap(name: IconProps.Name) {
+    const genAiFilledIcon = `<svg
```
timogasda, line 107: Why can this not be in a normal file like the other icons? Would be good to keep the process consistent.

It also means that this icon does not get passed through our `generate-icons` build task, which (among other things) applies some optimizations on the SVGs.

src/icon/internal.tsx
```diff
@@ -103,12 +103,32 @@ const InternalIcon = ({
...
+        xmlns="http://www.w3.org/2000/svg"
+        focusable="false"
+        aria-hidden="true"
+        data-testid="gen-ai-filled"
+      >
+        <path d="m8 4.4 1.018 2.582L11.6 8 9.018 9.018 8 11.6 6.982 9.018 4.4 8l2.582-1.018L8 4.4ZM2.405 2.41l.002-.003.003-.002-.003-.002-.002-.003-.002.003-.003.002.003.002.002.003Z" class="filled"></path>
+      </svg>`;
+
+    if (name === 'gen-ai' && iconSize === 'small') {
+      return genAiFilledIcon;
```
timogasda, line 120: Why create this magic switch? We usually have separate filled and unfilled versions of icons. Shouldn't we stick to that approach?

src/icon/internal.tsx
timogasda: Looks like @just-boris suggested this in an earlier review. Are you sure this was a good idea?

src/icon/internal.tsx
jkuelz: It is used for the accompanying unit tests.

src/icon/internal.tsx
jkuelz: It started out as a normal file so that it would be part of the build task, that was my preferred path as well. But Boris pointed out the challenges with that approach along with that fact that we already exclude some svgs from this build task, so it seemed low risk here.

src/icon/internal.tsx
timogasda: It seems a little strange to me that we use this on our own components, because then this also is rendered in our customers' applications. But doing a little code search shows me that we've done this in a few places in the system already, which I didn't realize. I guess it's fine then

src/icon/internal.tsx
timogasda: But we've never used it for SVGs that are served via the `Icon` component, which is a little different I think.

If it's a one-time thing I'm fine with this approach. If this is going to happen more in the future, I think we need a proper mechanism for it.

PR #285, opened 2022-09-18
Make popover body scrollable when needed

pan-kot:
### Description

When popover content is tall sometimes it can be cut by the viewport. The fix allows the content to become scrollable when necessary.

Screenshot 2022-09-18 at 11 25 57

### How has this been tested?

* Manual tests for overflow using new test page;
* Screenshot tests;
* Existing placement integration tests;
* New unit tests for updated positioning logic.

### Documentation changes

[*Do the changes include any API documentation changes?*]
- [ ] _Yes, this change contains documentation changes._
- [x] _No._

### Related Links

[*Attach any related links/pull request for this change*]

<details>
   <summary>Review checklist</summary>

_The following items are to be evaluated by the author(s) and the reviewer(s)._

#### Correctness

- [ ] _Changes are backward-compatible if not indicated, see [`CONTRIBUTING.md`](CONTRIBUTING.md#public-apis)._
- [ ] _Changes do not include unsupported browser features, see [`CONTRIBUTING.md`](CONTRIBUTING.md#browsers-support)._
- [ ] _Changes were manually tested for accessibility, see [accessibility guidelines](https://cloudscape.design/foundation/core-principles/accessibility/)._

#### Security

- [ ] _If the code handles URLs: all URLs are validated through [the `checkSafeUrl` function](https://github.com/cloudscape-design/components/blob/main/src/internal/utils/check-safe-url.ts)._

#### Testing

- [ ] _Changes are covered with new/existing unit tests?_
- [ ] _Changes are covered with new/existing integration tests?_
</details>

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1804, opened 2023-12-14
fix: Prevent chart popovers from overflowing horizontally

jperals:
### Description

When a popover can't find a position where it could fit in its container or viewport without being cut off, it tries to find the one where it would be cut off by the least amount. Bar charts in mobile screens bring this logic quite to the limit, because the popover triggers are the bars which can occupy most of the viewport height, leaving practically no vertical space, but the sides can also be insufficient and lead to cut-off popovers:

localhost_8080_ (46)

With the new behavior, we give more preference to vertical than horizontal overflow and in mobile screens we allow the popover to vertically expand beyond the viewport and scroll to it if necessary:

localhost_8080_ (47)

Issue: AWSUI-24348

### How has this been tested?

- Added unit tests and integration tests
- Manually tested on browser and mobile device

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/popover/interfaces.ts
```diff
@@ -93,12 +93,12 @@ export interface Offset {
   top: number;
 }
 
-export interface BoundingBox {
+export interface Dimensions {
```
jperals, line 96: Renamed these two interfaces hopefully for more clarity.

`BoundingOffset` was supposed to mean "bounding box with offset". However I would expect a bounding offset to already contain the offset (left and top coordinates) besides the width and the height.

jperals:
Will keep iterating, setting back to draft.

src/popover/__tests__/mobile.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+});
+
+function renderPopover() {
+  const { container } = render(<Popover>Trigger</Popover>);
+  return createWrapper(container).findPopover();
+}
+
+describe('Scrolls beyond the viewport if necessary on mobile screen', () => {
+  test('scrolls up when the container has more space at the top', () => {
+    (calculatePosition as jest.Mock).mockReturnValue({
```
just-boris, line 48: Personal opinion, but I find this version more readable

```suggestion
    jest.mocked(calculatePosition).mockReturnValue({
```

src/popover/container.tsx
```diff
@@ -119,11 +122,15 @@ export default function PopoverContainer({
       const shouldKeepPosition = keepPosition && onContentResize && !!previousInternalPositionRef.current;
       const fixedInternalPosition = (shouldKeepPosition && previousInternalPositionRef.current) ?? undefined;
 
+      // On mobile screens, allow the popover to open outside of the viewoprt and scroll to it,
+      // if there is no position where it can fit without being cropped.
+      const allowVerticalOverflow = isMobile && !shouldKeepPosition;
```
just-boris, line 127: `isMobile` depends on the window _width_. Vertical overflow happens when the window _height_ is too small. 

Should we rely on a different condition to make overflow happening?

src/popover/utils/positions.ts
```diff
@@ -228,55 +227,74 @@ export function calculatePosition({
...
+      return { internalPosition, boundingBox };
     }
+
+    const newOption = { boundingBox, internalPosition, visibleArea };
+    bestOption = getBestOption(newOption, bestOption);
   }
 
   // Use best possible placement.
-  const internalPosition = bestPositionOutsideViewport?.internalPosition || 'right-top';
+  const internalPosition = bestOption?.internalPosition || 'right-top';
```
just-boris, line 268: Is this a part of the issue, that by default we fallback to `right-top` position, which is hard to align on small screens?

Should we change it to something more reasonable?

src/popover/container.tsx
```diff
@@ -148,11 +156,15 @@ export default function PopoverContainer({
...
       if (scrollable) {
-        body.style.maxHeight = boundingOffset.height + 'px';
+        body.style.maxHeight = boundingBox.height + 'px';
         body.style.overflowX = 'hidden';
         body.style.overflowY = 'auto';
       }
 
+      if (allowVerticalOverflow) {
+        scrollRectangleIntoView(boundingBox);
+      }
```
just-boris, line 166: I do not understand the logic around this scroll behavior. Let's try together

1. First, we try all possible positions. If we can fit popover fully, exit with the chosen position
2. If we could not fit, we try to find best possible option, giving preference to positions with more available height
3. Limit the popover height and add overflow
4. Update page scroll (this particular line). At this point, all offsets calculated on step 2 are not valid anymore, because elements position changed.

Should the step 4 happen somewhere else in the workflow, or not happen at all?

src/popover/container.tsx
```diff
@@ -57,9 +59,10 @@ export default function PopoverContainer({
   const arrowRef = useRef<HTMLDivElement | null>(null);
   const previousInternalPositionRef = useRef<InternalPosition | null>(null);
 
-  const [popoverStyle, setPopoverStyle] = useState<CSSProperties>(INITIAL_STYLES);
+  const [popoverStyle, setPopoverStyle] = useState<Offset>(INITIAL_STYLES);
   const [internalPosition, setInternalPosition] = useState<InternalPosition | null>(null);
   const isRefresh = useVisualRefresh();
+  const isMobile = useMobile();
```
just-boris, line 65: `isMobile` is just a constant value of some width. Why should 700px and 600px wide screens work differently?

src/popover/container.tsx
jperals: Your understanding is quite correct, and it's true that the top position value will be off sync for one frame. The scrolling itself then triggers a call to the [positionHandler](https://github.com/cloudscape-design/components/blob/04597474026edf3fb2fd665797dffc96d8ca57d9/src/popover/container.tsx#L175C7-L175C7), which corrects this because it recalculates the position relative to the new position of the trigger. I can't noticeable reproduce the initial misplacement (even with CPU throttling) but I will try to change this if it doesn't get too complicated.

src/popover/container.tsx
jperals: The reason to distinguish between mobile and desktop is to avoid the scrolling while hovering, so I'll try something else somehow based on hover vs tap rather than on the screen size.

pages/popover/placement-test.page.tsx
```diff
@@ -64,14 +65,20 @@ export default function () {
...
                   content={<Content />}
                   dismissAriaLabel="Close"
                   triggerType="custom"
                 >
-                  <button id="popover-trigger">Open popover</button>
+                  <button
+                    id="popover-trigger"
+                    style={triggerHeight ? { display: 'block', height: `${triggerHeight}px` } : undefined}
+                  >
+                    {triggerHeight === undefined && 'Open popover'}
```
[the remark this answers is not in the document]
jperals: To make sure it is narrow enough that it has more space to the sides than to the top or bottom, and to test that it will expand to the top or bottom despite that. The text is OK when simulating a normal inline button, but for these cases it makes the testing conditions less deterministic.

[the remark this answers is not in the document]
just-boris: Sounds like one more reason to introduce separate „trigger variant“ param to configure this page

src/popover/container.tsx
just-boris: What does it have to do with device type? If the bug says „popover does not fit vertically“ - the fix probably should count on vertical size. If anybody has a 1000x200 window for some reason, the fix should work for them too

jperals: The device type distinction comes from two reasons:
- On desktop chart popovers are opened on hover, and triggering auto-scroll on hover can lead to a jarring experience
- A scrollable popover on mobile is more concerning than on desktop because there are no scrollbars that indicate its scrollability, so it's preferable to auto-scroll the window instead

Also considering a use case like the one depicted [here](https://github.com/cloudscape-design/components/pull/285), where chaning the behavior seems unnecessary. Although for such a case, setting a minimum height to the popover before going for auto-scroll vs internal popover scroll would also work.

src/popover/container.tsx
just-boris: What if we move the step 4 after 1? Is there a chance that normal position will fit after scrolling?

src/popover/container.tsx
just-boris: I think I got what you wanted to do here. You want to apply fix only when popover opens by tapping on it, skipping the hover interactions. 

If this is what you want, we can dive into solutions for this

just-boris: > A scrollable popover on mobile is more concerning than on desktop because there are no scrollbars

There are no scrollbars on macs, big and small. If this is a factor you find important for the fix, it should be cross-platform.

----

As for the conditions to activate this behavior, this sounds like it should be activated when `isPinned` flag on chart-popover is active. What do you think?

jperals: > There are no scrollbars on Apple devices, big and small. If this is a factor you find important for the fix, it should be cross-platform.

Actually to be precise there are no scrollbars on Apple devices (by default) _when using touch interfaces_ (mobile or trackpad), so I am more inclined to be specific to tap vs click as you suggested earlier. Using `isPinned` could lead to undesired resizing flows when pinning for popovers that were already open on hover (i.e, chart popovers).

I guess instead of
> A scrollable popover on mobile is more concerning than on desktop because there are no scrollbars that indicate its scrollability

to be more precise I should have said
> A scrollable popover on Apple devices with touch interface is more concerning than on desktop because there are no scrollbars that indicate its scrollability (by default)

PS: To check that "Automatically based on mouse or trackpad" is the default option, you can run `defaults delete -g AppleShowScrollBars` and reopen the Appearance settings.

src/popover/utils/positions.ts
jperals: This should not be relevant because the only case where this value would be used would be if _all positioning options_ would be completely off screen (not a single pixel inside the viewport). We could change this, but it is not the cause of the issue.

src/popover/container.tsx
jperals: Handled this [here](bff2287b713bd775cfde3144183552067e52b381) by giving the popover the after-scroll position already right before (eventually) scrolling.

src/popover/container.tsx
just-boris: Parked for offline discussion. I will get back to reviewing this after the offline discussion happens

src/mixed-line-bar-chart/__integ__/mixed-line-bar-chart.test.ts
```diff
@@ -496,6 +496,33 @@ describe('Details popover', () => {
...
+  );
+
+  test(
+    'does not scroll on hover',
+    setupTest('#/light/bar-chart/drilldown', async page => {
+      await page.setWindowSize({ width: 360, height: 650 });
+      await page.windowScrollTo({ top: 150 });
+      const barChart = createWrapper().findBarChart();
+      const barGroup = barChart.findBarGroups().get(1).toSelector();
+      await page.hoverElement(barGroup);
```
just-boris, line 520: For a more solid test, you need to add the same assertion before and after hover interaction

```js
await expect(page.getWindowScroll()).resolves.toEqual({ top: 150, left: 0 });
```

src/popover/container.tsx
```diff
@@ -34,10 +33,12 @@ export interface PopoverContainerProps {
   variant?: 'annotation';
   // When keepPosition is true, the popover will not recalculate its position when it resizes nor when it receives clicks.
   keepPosition?: boolean;
+  // When allowVerticalScroll is true, we will scroll to the the popover if it overflows the viewport even when choosing the best possible position for it.
+  // Do not use this if the popover is open on hover, in order to avoid unexpected movement.
+  allowVerticalScroll?: boolean;
```
just-boris, line 38: A better name would be `allowScrollToFit`, something that suggests there will be scroll outside, not inside the content

src/internal/utils/scrollable-containers.ts
```diff
@@ -84,3 +84,28 @@ export function scrollElementIntoView(
...
+
+/**
+ * For elements with fixed position, the browser's native scrollIntoView API doesn't work,
+ * so we need to manually scroll to the element's position.
+ * Supports only vertical scrolling.
+ */
+export function scrollRectangleIntoView(box: BoundingBox) {
+  const scrollAmount = calculateScroll(box);
+  if (scrollAmount) {
+    window.scrollBy(0, scrollAmount);
```
just-boris, line 109: What if the chart displayed inside scrollable container? For example a split panel? Do we need to support this?

just-boris (approved):

src/mixed-line-bar-chart/__integ__/mixed-line-bar-chart.test.ts
jperals: What could happen between `await page.windowScrollTo({ top: 150 });` and `await page.hoverElement(barGroup);` that could alter the scroll position? Or is there a reason why the first one would not set it correctly in the first place?

just-boris: `windowScrollTo` may hit the page limit and not scroll fully as expected.

More assertions is better than less assertions

jperals: OK, added.

src/popover/container.tsx
jperals: Changed

just-boris (approved):

src/internal/utils/scrollable-containers.ts
jperals: Is this what you mean? It works as usual with the caveat that the potentially invisible part of the chart leads to an apparent white space between it and the popover (the second screenshot below makes this more apparent, because of the chart's scrolling position). This could be mitigated by scrolling the scrolling container, but I'm not sure we want the extra complexity.

Screenshot 1: chart more or less centered in its scrollable container

localhost_8080_ (48)

Screenshot 2: chart at the bottom, leaving most of its upper part cropped

localhost_8080_ (49)

just-boris (approved):

PR #1815, opened 2023-12-15
fix: Fix funnel handling when component is already unmounted

just-boris:
### Description

This code should not fire when component is already unmounted. 

The current behavior causes issues in tests for our consumers

Related links, issue #, if available: AWSUI-26406

### How has this been tested?

Is there a way to tear down the document in Jest? I do not think so.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/analytics/components/analytics-funnel.tsx
+++ b/src/internal/analytics/components/analytics-funnel.tsx
@@ -279,16 +279,25 @@ function useStepChangeListener(stepNumber: number, handler: (stepConfiguration: 
     };
   }, []);
 
+  useEffect(() => {
+    const handle = setTimeout(
+      () => subStepConfiguration.current.set(stepNumber, getSubStepConfiguration()),
+      SUBSTEP_CHANGE_DEBOUNCE
+    );
+    return () => {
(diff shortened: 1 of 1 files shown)


Al-Dani (dismissed):

src/internal/analytics/components/analytics-funnel.tsx
```diff
@@ -279,16 +279,25 @@ function useStepChangeListener(stepNumber: number, handler: (stepConfiguration:
     };
   }, []);
 
+  useEffect(() => {
+    const handle = setTimeout(
+      () => subStepConfiguration.current.set(stepNumber, getSubStepConfiguration()),
```
Al-Dani, line 284: aren't we making this update inside `stepChangeCallback` already?

just-boris: The value needs to be initialized on the first render too.

Previously it worked because the line was before `listenForSubStepChanges` condition.

Now it is separate code like it should have been in the first place

Al-Dani (approved):

PR #1820, opened 2023-12-19
fix: Focus previously highlighted element when navigating back from p…

jperals:
…opover

### Description

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/components/chart-plot/index.tsx
+++ b/src/internal/components/chart-plot/index.tsx
@@ -49,6 +49,7 @@ export interface ChartPlotProps {
   onBlur?: (event: React.FocusEvent<SVGGElement>) => void;
   onKeyDown?: (event: React.KeyboardEvent<SVGGElement>) => void;
   children: React.ReactNode;
+  hasHighlight?: boolean;
 }
 
 /**
@@ -89,6 +90,7 @@ function ChartPlot(
     onBlur,
     focusOffset = DEFAULT_PLOT_FOCUS_OFFSET,
     activeElementFocusOffset = DEFAULT_ELEMENT_FOCUS_OFFSET,
+    hasHighlight,
     ...restProps
   }: ChartPlotProps,
   ref: React.Ref<ChartPlotRef>
@@ -114,12 +116,11 @@ function ChartPlot(
     plotClickedRef.current = true;
   };
   const onPlotFocus = (event: React.FocusEvent<SVGSVGElement>) => {
-    if (event.target === svgRef.current && !plotClickedRef.current) {
-      setPlotFocused(true);
-    }
     // The click should focus the underling application bypassing the svg.
-    else if (plotClickedRef.current) {
+    if (plotClickedRef.current || hasHighlight) {
       applicationRef.current!.focus();
+    } else if (event.target === svgRef.current) {
+      setPlotFocused(true);
     }
   };
   const onPlotClick = (event: React.MouseEvent<SVGSVGElement, MouseEvent>) => {
--- a/src/mixed-line-bar-chart/chart-container.tsx
+++ b/src/mixed-line-bar-chart/chart-container.tsx
@@ -336,6 +336,8 @@ export default function ChartContainer<T extends ChartDataTypes>({
     return highlightedPoint?.datum?.x ?? null;
   }, [highlightedPoint, verticalMarkerX, highlightedGroupIndex, barGroups]);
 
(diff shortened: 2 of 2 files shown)


src/internal/components/chart-plot/index.tsx
```diff
@@ -89,6 +90,7 @@ function ChartPlot(
     onBlur,
     focusOffset = DEFAULT_PLOT_FOCUS_OFFSET,
     activeElementFocusOffset = DEFAULT_ELEMENT_FOCUS_OFFSET,
+    hasHighlight,
```
pan-kot, line 93: The `activeElementKey` property is used to focus the target element of the chart. Is it possible to define hasHighlight as `!!activeElementKey`?

src/internal/components/chart-plot/index.tsx
```diff
@@ -167,7 +167,8 @@ function ChartPlot(
   return (
     <>
       <svg
-        {...restProps}
```
jperals, line 170: Remove destructured props, only use explicit ones for better clarity. `onFocus` and `onBlur` were being assigned twice but it was not easy to tell because of this.

src/internal/components/chart-plot/index.tsx
```diff
@@ -85,11 +85,12 @@ function ChartPlot(
...
     onClick,
     onKeyDown,
-    onFocus,
-    onBlur,
     focusOffset = DEFAULT_PLOT_FOCUS_OFFSET,
     activeElementFocusOffset = DEFAULT_ELEMENT_FOCUS_OFFSET,
-    ...restProps
+    onMouseMove,
+    onMouseOut,
+    onApplicationBlur,
```
jperals, line 92: Renamed `onFocus` and `onBlur` to `onApplicationFocus` and `onApplicationBlur` because these functions are called when focusing and blurring the application (inner highlighting), not the wrapping SVG plot.

src/mixed-line-bar-chart/chart-container.tsx
```diff
@@ -209,7 +209,6 @@ export default function ChartContainer<T extends ChartDataTypes>({
    */
   const highlightedPointRef = useRef<SVGGElement>(null);
   const highlightedGroupRef = useRef<SVGRectElement>(null);
-  const [isPlotFocused, setPlotFocused] = useState(false);
```
jperals, line 212: `setPlotFocused` was being called when focusing the application (not the plot), so in reality this should have been called `[isApplicationFocused, setApplicationFocused]`. But the chart plot component already has its internal `isApplicationFocused` state, so that this duplication was not really necessary.

src/internal/components/chart-plot/index.tsx
```diff
@@ -114,12 +115,11 @@ function ChartPlot(
...
   };
   const onPlotFocus = (event: React.FocusEvent<SVGSVGElement>) => {
-    if (event.target === svgRef.current && !plotClickedRef.current) {
-      setPlotFocused(true);
-    }
-    // The click should focus the underling application bypassing the svg.
-    else if (plotClickedRef.current) {
+    // The click should focus the underlying application bypassing the svg.
+    if (plotClickedRef.current || !!activeElementKey) {
       applicationRef.current!.focus();
```
timogasda, line 120: Can we add a comment here explaining why we focus the application here?

And likewise a similar comment for the `else` case.

src/internal/components/chart-plot/index.tsx
jperals: Did this in the [last commit](9b48f677cf9f4c40db68e5901771bbbc52d40f5f), but note that this required some further small changes because `hasHighlight` keeps its value when blurring the application whereas `activeElementKey` doesn't, so when tabbing back the entire SVG plot would be focused instead of the application. This is because in mixed charts, `activeElementKey` was only passed as true when `isPlotFocused` is true in chart container. This `isPlotFocused` tracks whether the application (not the plot, confusingly) is focused, so this part of the state is essentially a duplication of `isApplicationFocused` in chart plot. So I opted for removing `isPlotFocused` in chart container and handling this part of the state only with `isApplicationFocused` in the chart plot.

After these changes I manually tested VoiceOver in Chrome and Firefox for pie, area and bar chart.

src/internal/components/chart-plot/index.tsx
jperals: Done

timogasda (approved):

pan-kot (approved):
Great work!!

PR #1827, opened 2023-12-22
fix: Fixed rounded corner misalignment in media container

at-susie:
### Description
We discovered that there is a small gap in rounded corner sizes in Container with media variant.
rounded-corner-gap

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/container/styles.scss
+++ b/src/container/styles.scss
@@ -11,6 +11,7 @@
   @include styles.styles-reset;
   word-wrap: break-word;
   position: relative;
+  overflow: hidden;
 
   &.fit-height {
     display: flex;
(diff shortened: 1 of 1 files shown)


src/container/styles.scss
```diff
@@ -11,6 +11,7 @@
   @include styles.styles-reset;
   word-wrap: break-word;
   position: relative;
+  overflow: hidden;
```
just-boris, line 14: We cannot do this, because it will make absolute elements (dropdowns and popover) cropped by the container borders.

at-susie: That's so true. I'll update.
Another option that I was thinking but with a hesitation is to reduce the border-radius size from the media wrapper `.media` like `border-top-left-radius: calc(awsui.$border-radius-container - 1px);`.

just-boris (approved):
The change looks good, but let's wait until after holidays to merge this

timogasda (approved):

PR #1828, opened 2023-12-22
feat: Handle drawers focus when focusTools is fired 

Al-Dani:
### Description

This is a fallback for a situation when app layout has both tools and drawers set. Tools panel won't be rendered, if drawers are present. The fallback logic of rendering tools content as a drawer already exists (somewhere), but existing focus management logic doesn't work at the moment.  

Related links, issue #, if available: AWSUI-26132

### How has this been tested?

new unit test added

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/__tests__/utils.tsx
+++ b/src/app-layout/__tests__/utils.tsx
@@ -284,32 +284,3 @@ export const singleDrawerPublic: Array<AppLayoutProps.Drawer> = [
     },
   },
 ];
-
-export const multipleDrawers: Array<AppLayoutProps.Drawer> = [
-  {
-    ariaLabels: {
-      closeButton: 'Help close button',
-      drawerName: 'Help drawer content',
-      triggerButton: 'Help trigger button',
-      resizeHandle: 'Help resize handle',
-    },
-    content: <span>Help</span>,
-    id: 'help',
-    trigger: {
-      iconName: 'status-info',
-    },
-  },
-  {
-    ariaLabels: {
-      closeButton: 'Security close button',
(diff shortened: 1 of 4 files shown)


timogasda (dismissed):

src/app-layout/index.tsx
```diff
@@ -491,7 +491,13 @@ const OldAppLayout = React.forwardRef(
           onNavigationToggle(false);
         }
       },
-      focusToolsClose: () => focusToolsButtons(true),
+      focusToolsClose: () => {
+        if (drawers && drawers.length) {
```
katiegeorge, line 495: We've been using the variable `hasDrawers` in this file instead of just `drawers` (defined [here](https://github.com/cloudscape-design/components/blob/main/src/app-layout/index.tsx#L185)), better to keep it consistent I would think.

src/app-layout/__tests__/utils.tsx
```diff
@@ -284,3 +284,32 @@ export const singleDrawerPublic: Array<AppLayoutProps.Drawer> = [
     },
   },
 ];
+
+export const multipleDrawers: Array<AppLayoutProps.Drawer> = [
```
katiegeorge, line 288: Is there a reason you need to add this instead of using `singleDrawerPublic`?

Al-Dani: I added it to simulate a real use-case with 2 drawers, when one of them is a help panel. But for the test it doesn't really matter, you are right :)

timogasda (approved):

PR #1842, opened 2023-12-28
chore: Update react virtual to v3

pan-kot:
### Description

Upgraded react-virtual to the latest version (v3, see: https://tanstack.com/virtual/v3/docs/adapters/react-virtual) to fix npm install issue caused by React 18 not being listed in peer dependencies of v2.

Additionally, removed the workaround introduced for v2 to avoid "Maximum update depth exceeded" error. A similar problem is possible in v3 but it does not cause the error. Instead, the virtual list bounces between two or more sizes with a small delay which is not expected to cause major UX issues. The problem is rare and can be mitigated by updating item styles so that list total size to no longer affect individual item heights.

Rel: AWSUI-26338, https://github.com/cloudscape-design/components/pull/1202

### How has this been tested?

Existing unit and integration tests, manual tests, screenshot tests.

<details>

#### Correctness

#### Security

#### Testing

</details>

---


__mocks__/@tanstack/react-virtual.ts
```diff
@@ -0,0 +1,40 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
just-boris, line 1: If we use a global mock, how do we ensure that unit tests of our consumers will not fail? Can we not use such ultimate tool?

src/autosuggest/virtual-list.tsx
```diff
@@ -26,22 +26,25 @@ const VirtualList = ({
...
     estimateSize: useCallback(() => 31, [width, highlightText]),
+    overscan: 5,
   });
 
+  const totalItems = autosuggestItemsState.items.length;
   useEffect(() => {
-    if (autosuggestItemsState.highlightType.moveFocus) {
-      rowVirtualizer.scrollToIndex(autosuggestItemsState.highlightedIndex);
+    const index = autosuggestItemsState.highlightedIndex;
+    if (autosuggestItemsState.highlightType.moveFocus && index >= 0 && index < totalItems) {
```
just-boris, line 44: What changed here? Why do we need to validate `index` now?

just-boris (commented):
Also questions about testing

1. PR build fails
2. Did you measure performance differences?
3. For the manual testing, how much are you confident in your own testing? Did you ask other team members to help?

package.json
```diff
@@ -29,6 +29,7 @@
     "@dnd-kit/sortable": "^7.0.2",
     "@dnd-kit/utilities": "^3.2.1",
     "@juggle/resize-observer": "^3.3.1",
+    "@tanstack/react-virtual": "^3.0.1",
```
just-boris, line 32: Should we pin this dependency to avoid unexpected versions installed by our consumers?

pan-kot:
The measure bundle size job fails because we replace its version of components with a local copy but do not account for newly added dependencies:

```
run: |
  cd components
  npm i --force
  npm run quick-build

  cd ../bundle-size
  rm -rf node_modules/@cloudscape-design/components
  cp -r $GITHUB_WORKSPACE/components/lib/components node_modules/@cloudscape-design/components
  cp main-with-cloudscape.jsx main.jsx
  npm run build
```

package.json
pan-kot: We don't pin other dependencies, why should we pin this one?

src/autosuggest/virtual-list.tsx
pan-kot: In v3 there is an error thrown when executing scroll to an index for which there is no item defined, see https://github.com/TanStack/virtual/blob/main/packages/virtual-core/src/index.ts#L822

package.json
just-boris: This used to be pinned before because of bugs. Check the pre-github commit history

just-boris:
> do not account for newly added dependencies

Can you make it work?

__mocks__/@tanstack/react-virtual.ts
pan-kot: Updated

pan-kot:
> Also questions about testing
> 
> 1. PR build fails
> 2. Did you measure performance differences?
> 3. For the manual testing, how much are you confident in your own testing? Did you ask other team members to help?

1. Fixed
2. I did not measure it but tested with throttling on a list of 1000 items. The v2 feels slightly faster when scrolling but nothing significant. I did not observe any difference to the initial rendering performance
3. I tested using the available test pages plus specifically tested for https://github.com/cloudscape-design/components/pull/1202 scenario. I think the manual testing done by both of us plus CI and dry-run should be enough, but I can ask for more testing if you think it is necessary

__mocks__/@tanstack/react-virtual.ts
just-boris: What is updated here? This mock still applies to all of our tests. We never test against the actual virtual instance in JSDOM.

Do you think it is okay? What if something fails in our live consumers?

package.json
just-boris: What is your final decision here? How about asking the team opinion?

src/autosuggest/virtual-list.tsx
just-boris: ok for now, but doesn't it sounds like a problem with our code?

pan-kot: I agree, there is an opportunity for improvement. In this particular case the issue comes from the fact we use -1 as the initial value of the highlighted index.

package.json
pan-kot: Agreed on having this dependency pinned.

__mocks__/@tanstack/react-virtual.ts
just-boris: How do we release this to live, if it has required upgrade steps to our consumers?

pan-kot:
Closed in favour of https://github.com/cloudscape-design/components/pull/1897

PR #1845, opened 2023-12-29
refactor: Page options loader helper

pan-kot:
### Description

Introduce a common utility to demonstrate and test async loading feature in autosuggest, select, multiselect.

### How has this been tested?

Manual testing, existing integration tests.

<details>

#### Correctness

#### Security

#### Testing

</details>

---


pages/property-filter/async-loading.integ.page.tsx
```diff
@@ -25,7 +26,7 @@ const filteringProperties = [
     groupValuesLabel: `Label values`,
   },
 ] as const;
-const serverFilteringOptions = range(1000).map(value => ({
+const filteringOptions = range(1000).map(value => ({
```
pan-kot, line 29: In the original page implementation all 1000 options are resolved as a single page. I updated this to be 10 pages with 100 items each instead to showcase progressive loading. That revealed what I would consider a bug: once a property is selected the fetch is called twice (for the first and the second page already). I updated the integration test accordingly and added a comment.

pages/property-filter/async-loading.integ.page.tsx
```diff
@@ -51,31 +47,10 @@ export default function () {
...
-      request.current.filteringOperator !== filteringOperator
-    ) {
-      // there is another request in progress, discard the result of this one
-      return;
-    }
-    setStatus('finished');
-    setFilteringOptions(items);
-  };
+  const [fetchTarget, setFetchTarget] = useState<'properties' | 'options'>('properties');
+  const propertiesLoader = useOptionsLoader({ timeout: 100 });
```
Al-Dani, line 51: it looks like we're passing filteringProperties as a constant anyway, does it play a role anywhere?

pages/common/options-loader.ts
```diff
@@ -0,0 +1,143 @@
...
+declare const window: ExtendedWindow;
+const pendingRequests: Array<PendingRequest> = (window.__pendingRequests = []);
+
+export interface OptionsLoaderProps {
+  pageSize?: number;
+  timeout?: number;
+  randomErrors?: boolean;
+}
+
+export interface OptionsLoaderState<Item> {
```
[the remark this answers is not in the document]
pan-kot: great finding!

pages/common/options-loader.ts
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  function issuePendingRequest(pageNumber: number, filteringText: string) {
+    const params = { filteringText, pageNumber };
+    return new FakeRequest<APIResponse>(pageNumber, (resolve, reject) =>
+      pendingRequests.push({ resolve, reject, params })
+    );
+  }
+
+  function fetchItems({ sourceItems, firstPage, filteringText: nextFilteringText }: FetchItemsProps<Item>) {
```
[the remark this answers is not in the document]
pan-kot: Yes, I agree. However, passing that as a hook argument won't work because this way it will be delayed by one render cycle for consecutive loading case

pages/property-filter/async-loading.integ.page.tsx
pan-kot: Yes, that is to match the current implementation more or less. Extending the test page to also feature async loading for properties could be a good next step. It is used here:
```
const status = { properties: propertiesLoader, options: optionsLoader }[fetchTarget].status;
```

As you can see, the status is taken from the properties or options loader depending on the current fetch target.

PR #1750, opened 2023-11-23
chore: Grid navigation alt 2

pan-kot:
### Description

Related links, issue #, if available: n/a

### How has this been tested?

<details>
   <summary>Review checklist</summary>

_The following items are to be evaluated by the author(s) and the reviewer(s)._

#### Correctness

- _Changes include appropriate documentation updates._
- _Changes are backward-compatible if not indicated, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#public-apis)._
- _Changes do not include unsupported browser features, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#browsers-support)._
- _Changes were manually tested for accessibility, see [accessibility guidelines](https://cloudscape.design/foundation/core-principles/accessibility/)._

#### Security

- _If the code handles URLs: all URLs are validated through [the `checkSafeUrl` function](https://github.com/cloudscape-design/components/blob/main/src/internal/utils/check-safe-url.ts)._

#### Testing

- _Changes are covered with new/existing unit tests?_
- _Changes are covered with new/existing integration tests?_
</details>

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1846, opened 2023-12-30
refactor: Simpler grid navigation implementation

pan-kot:
### Description

Use simpler grid navigation implementation (no mutation observer and a single variable to keep focused cell) based on https://github.com/cloudscape-design/components/pull/1750 to prepare for the register-based implementation based on https://github.com/cloudscape-design/components/pull/1760.

### How has this been tested?

New integration tests, reworked unit tests.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/table/table-role/__integ__/grid-navigation.test.ts
+++ b/src/table/table-role/__integ__/grid-navigation.test.ts
@@ -1,22 +1,30 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-import { BasePageObject } from '@cloudscape-design/browser-test-tools/page-objects';
 import useBrowser from '@cloudscape-design/browser-test-tools/use-browser';
 import createWrapper from '../../../../lib/components/test-utils/selectors';
+import { GridNavigationPageObject } from './page-object';
 
-interface ExtendedWindow extends Window {
-  refreshItems: () => void;
+interface Options {
+  actionsMode?: 'dropdown' | 'inline';
 }
-declare const window: ExtendedWindow;
+
+const setupTest = (
+  { actionsMode = 'dropdown' }: Options,
+  testFn: (page: GridNavigationPageObject) => Promise<void>
+) => {
+  return useBrowser(async browser => {
+    const page = new GridNavigationPageObject(browser);
+    const query = new URLSearchParams({ actionsMode });
+    await browser.url(`#/light/table-fragments/grid-navigation-custom/?${query.toString()}`);
+    await page.waitForVisible('table');
+    await testFn(page);
+  });
+};
 
 test(
   'cell action remains focused when row re-renders',
-  useBrowser({}, async browser => {
-    const page = new BasePageObject(browser);
-    await browser.url('#/light/table-fragments/grid-navigation-custom/?actionsMode=inline');
-    await page.waitForVisible('table');
-
+  setupTest({ actionsMode: 'inline' }, async page => {
     await page.click('button[aria-label="Update item"]');
     await expect(page.isFocused('button[aria-label="Update item"]')).resolves.toBe(true);
   })
@@ -24,11 +32,7 @@ test(
 
 test(
   'cell focus stays in the same position when row gets removed',
-  useBrowser({}, async browser => {
-    const page = new BasePageObject(browser);
-    await browser.url('#/light/table-fragments/grid-navigation-custom/?actionsMode=inline');
-    await page.waitForVisible('table');
-
+  setupTest({ actionsMode: 'inline' }, async page => {
     await page.click('button[aria-label="Delete item"]');
     await expect(page.isFocused('button[aria-label="Delete item"]')).resolves.toBe(true);
   })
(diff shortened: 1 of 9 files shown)


src/table/table-role/grid-navigation.tsx
```diff
@@ -100,13 +91,28 @@ class GridNavigationHelper {
     this._pageSize = pageSize;
   }
 
+  public refresh() {
+    if (this._table) {
+      const cellSuppressed = this.focusedCell ? this.isSuppressed(this.focusedCell.element) : false;
```
michaeldowseza, line 96: Should this run after the focusedCell needs to be updated?

src/table/table-role/grid-navigation.tsx
```diff
@@ -100,13 +91,28 @@ class GridNavigationHelper {
     this._pageSize = pageSize;
   }
 
+  public refresh() {
+    if (this._table) {
```
michaeldowseza, line 95: Can this be run before the processor has initialised?

src/table/table-role/grid-navigation.tsx
```diff
@@ -135,7 +144,13 @@ class GridNavigationHelper {
   };
 
   private onFocusout = () => {
-    this.focusedCell = null;
+    // When focus leaves the cell and the cell becomes no longer belong to the table it indicates the focused element has been unmounted.
```
michaeldowseza, line 147: Minor: "When focus leaves the cell and the cell ~becomes~ no longer belongs to the table, it indicates the focused element has been unmounted."

src/table/table-role/grid-navigation.tsx
```diff
@@ -168,74 +183,50 @@ class GridNavigationHelper {
...
-              ensureSingleFocusable(this.table, cell);
-              moveFocusBy(this.table, cell, { y: 0, x: 0 });
-            }
-          }
-        }
-      }
-    }
-  };
+  private moveFocusBy(cell: FocusedCell, delta: { x: number; y: number }) {
+    getNextFocusable(this.table, cell, delta)?.focus();
```
michaeldowseza, line 230: Missing return statement?

src/table/table-role/table-role-helper.ts
```diff
@@ -103,6 +103,10 @@ export function getTableColHeaderRoleProps(options: {
     nativeProps['aria-sort'] = getAriaSort(options.sortingStatus);
   }
 
+  if (options.tableRole === 'grid') {
```
michaeldowseza, line 106: The same if check appears on line 98. Can combine these two.

src/table/table-role/table-role-helper.ts
```diff
@@ -117,5 +121,9 @@ export function getTableCellRoleProps(options: { tableRole: TableRole; colIndex:
     nativeProps.scope = 'row';
   }
 
+  if (options.tableRole === 'grid') {
```
michaeldowseza, line 124: Similar to previous comment. The same if check appears on line 116.

src/table/table-role/utils.ts
```diff
@@ -33,62 +28,61 @@ export function findFocusinCell(event: FocusEvent): null | FocusedCell {
   }
 
   const cellFocusables = getFocusables(cellElement);
-  const elementIndex = cellFocusables.indexOf(element);
+  const elementIndex = cellFocusables.indexOf(focusedElement);
 
-  return { rowIndex, colIndex, rowElement, cellElement, element, elementIndex };
+  return { rowIndex, colIndex, rowElement, cellElement, element: focusedElement, elementIndex };
```
michaeldowseza, line 33: Why do we return the element here if its the same as the focusedElement provided?

src/table/table-role/utils.ts
```diff
@@ -33,62 +28,61 @@ export function findFocusinCell(event: FocusEvent): null | FocusedCell {
...
 }
 
 /**
  * Moves table focus in the provided direction. The focus can transition between cells or interactive elements inside cells.
```
michaeldowseza, line 37: Comment no longer matches implementation

michaeldowseza: Same with the inline comments here that mention that it moves focus

src/table/table-role/__integ__/grid-navigation.test.ts
```diff
@@ -28,9 +34,56 @@ test(
   })
 );
 
+test(
+  'cell focus stays in the same position when row gets removed with auto-refresh',
+  useBrowser({}, async browser => {
+    const page = new BasePageObject(browser);
+    await browser.url('#/light/table-fragments/grid-navigation-custom');
+    await page.waitForVisible('table');
```
michaeldowseza, line 42: Minor: Don't we normally abstract this into something like a `setupTest` function to remove some of this duplication? Although since it just a test file I don't have any problem with the duplication 🤷🏻‍♂️

src/table/table-role/grid-navigation.tsx
pan-kot: That makes sense, will update

src/table/table-role/grid-navigation.tsx
pan-kot: It should not, but that depends on the order of effects execution. If the effect that fires the update runs before the one that initialises the table there going to be an error.

src/table/table-role/grid-navigation.tsx
pan-kot: No, this one just focuses on the next applicable element. The case statements return to prevent multiple matches. The same can be achieved by using `break` after each case statement.

src/table/table-role/__integ__/grid-navigation.test.ts
pan-kot: Makes sense, will update!

PR #1851, opened 2024-01-02
chore: Migrate "A/B/C/D/E/F" components to logical properties

DaemonCahill:
This PR is a subset of the changes drafted in https://github.com/cloudscape-design/components/pull/1713 that affect only the subset of components that being with the letters "A/B/C/D/E/F".

### DIFF
--- a/src/calendar/styles.scss
+++ b/src/calendar/styles.scss
@@ -18,11 +18,12 @@
 .calendar {
   display: block;
   // IE11 does not calculate the height correctly when in nested flex containers (@see https://github.com/philipwalton/flexbugs#flexbug-3)
-  width: awsui.$size-calendar-grid-width;
+  inline-size: awsui.$size-calendar-grid-width;
   overflow: auto;
 
   &-inner {
-    margin: awsui.$space-xs;
+    margin-block: awsui.$space-xs;
+    margin-inline: awsui.$space-xs;
   }
 
   &-header {
@@ -35,7 +36,8 @@
     @include styles.font-body-m;
     font-weight: typographyConstants.$font-weight-bold;
     color: calendar.$header-color;
-    margin: 0;
+    margin-block: 0;
+    margin-inline: 0;
   }
 
   &-next-month-btn {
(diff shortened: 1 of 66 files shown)


rubencarvalho (dismissed):

src/code-editor/ace-editor.scss
```diff
@@ -12,10 +12,10 @@
 
 .code-editor-refresh :global .ace_editor {
   .ace_gutter {
-    border-top-left-radius: calc(awsui.$border-radius-code-editor - awsui.$border-item-width);
+    border-start-start-radius: calc(awsui.$border-radius-code-editor - awsui.$border-item-width);
```
just-boris, line 15: Since these are overriding built-in ace styles, how does it work? Do logical properties override non-logical?

src/container/styles.scss
```diff
@@ -50,12 +51,12 @@
   }
 
   // To ensure the top border/divider is visible on sticky elements which have a higher z-index
-  &.sticky-enabled:not(.refresh) {
+  &.sticky-enabled {
```
just-boris, line 54: Is this change expected?

src/code-editor/ace-editor.scss
DaemonCahill: Logical and physical properties are effectively setting the same value and it just comes down to selector specificity.

DaemonCahill: It would be no different than having two selectors set different values for the same property. The highest specificity selector value wins. There is no inherent hierarchy to physical vs logical properties to my knowledge.

src/container/styles.scss
DaemonCahill: Nope, good catch! Must have been a merge resolution error.

just-boris (dismissed):

src/calendar/styles.scss
```diff
@@ -18,11 +18,12 @@
 .calendar {
   display: block;
   // IE11 does not calculate the height correctly when in nested flex containers (@see https://github.com/philipwalton/flexbugs#flexbug-3)
-  width: awsui.$size-calendar-grid-width;
+  inline-size: awsui.$size-calendar-grid-width;
```
just-boris, line 21: IE11 and `inline-size` make no sense in one context. There is probably an opportunity to remove some unneeded styles or update the comment if this style needed for other browsers too

src/code-editor/ace-editor.scss
just-boris: 👍

just-boris (approved):

just-boris (approved):

PR #1852, opened 2024-01-03
fix: Ensure toggle is visible in high-contrast/forced-color mode

gethinwebster:
### Description

Ensure that toggle graphics are visible in forced-color mode. Previously they weren't, because the component relies purely on background color to "draw" the control, and forced-color overrides all background colors.

Related links, issue #, if available: AWSUI-24890

### How has this been tested?

Tested locally by emulating forced-color mode in Chrome

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/toggle/styles.scss
+++ b/src/toggle/styles.scss
@@ -23,6 +23,9 @@ $shadow-color: rgba(0, 0, 0, 0.25);
 }
 
 .toggle-control {
+  // this is a "drawing" element, and therefore we want it (and children) to retain
+  // background colors in high-contrast/forced-color mode
+  forced-color-adjust: none;
   @include styles.make-control-size($toggle-height, $toggle-width);
   background: awsui.$color-background-toggle-default;
   border-radius: 0.8 * styles.$base-size;


src/toggle/styles.scss
```diff
@@ -23,6 +23,9 @@ $shadow-color: rgba(0, 0, 0, 0.25);
 }
 
 .toggle-control {
+  // this is a "drawing" element, and therefore we want it (and children) to retain
+  // background colors in high-contrast/forced-color mode
+  forced-color-adjust: none;
```
michaeldowseza, line 28: Safari doesn't have support for this property yet.

gethinwebster: as far as I can tell that means that forced-color/high-contrast theme isn't a thing in Safari at all, so it doesn't matter that it's not supported as there's no way of putting the browser into a mode where the property would need to be supported

michaeldowseza: Safari supports the forced-color media query but I couldn't quite figure out how to emulate this correctly with Safari either. Closest I got was increased contrast mode via MacOS but that didn't really do the same thing.

michaeldowseza (approved):

PR #1856, opened 2024-01-04
chore: Table grid navigation register

pan-kot:
### Description

Reimplemented table grid navigation using single tab stop navigation context. The feature is tested on a custom table as of yet - integration with the table component is the next step.

### How has this been tested?

Existing unit- and integration tests. Updated unit test for focus context to include suppression mechanism.

<details>

#### Correctness

#### Security

#### Testing

</details>

---


pages/table-fragments/grid-navigation-custom.page.tsx
```diff
@@ -284,21 +289,22 @@ export default function Page() {
...
+                        </Cell>
                       ))}
                     </tr>
                   </thead>
                   <tbody>
                     {sortedItems.map((item, rowIndex) => (
                       <tr key={item.id} {...getTableRowRoleProps({ tableRole, rowIndex, firstIndex: 0 })}>
                         {visibleColumnDefinitions.map((column, colIndex) => (
-                          <td
+                          <Cell
```
jperals, line 300: [Before these changes](https://d21d5uik3ws71m.cloudfront.net/components/935524f97ebab6c43351cb269c7945f1c5ce4ffb/dev-pages/index.html#/light/table-fragments/grid-navigation-custom) the select and input fields in the "Select" and "Input" columns could be accessed directly with the grid navigation, and [now](https://d21d5uik3ws71m.cloudfront.net/components/87eabd0a9d0d6855cea278e4c4aacf423ac762cc/dev-pages/index.html#/light/table-fragments/grid-navigation-custom) the entire cell is focused instead and one further tab is necessary to access these inner interactive elements. Is this intended?

src/table/table-role/__tests__/stubs.tsx
```diff
@@ -21,25 +25,25 @@ export const idColumn = { header: 'ID', cell: (item: Item) => item.id };
 export const nameColumn = {
   header: (
     <span>
-      Name <button aria-label="Sort by name" tabIndex={0} />
+      Name <Button aria-label="Sort by name" />
```
jperals, line 28: Is this supposed to use the `Button` component? It doesn't seem to be imported and we are not using its props (e.g, `ariaLabel` instead of `aria-label`.

pages/table-fragments/grid-navigation-custom.page.tsx
pan-kot: The select and input components are not yet registered with single tab focus navigation context, that's why the focusing works differently.

src/table/table-role/__tests__/stubs.tsx
pan-kot: No, that is a custom button component defined down below.

src/internal/context/__tests__/single-tab-stop-navigation-context.test.tsx
```diff
@@ -16,10 +16,25 @@ function Button(props: React.HTMLAttributes<HTMLButtonElement>) {
 }
 
 test('does not override tab index when keyboard navigation is not active', () => {
-  renderWithSingleTabStopNavigation(<Button id="button" />);
+  renderWithSingleTabStopNavigation(<Button id="button" />, { navigationActive: false });
   expect(document.querySelector('#button')).not.toHaveAttribute('tabIndex');
 });
 
+test('does not override tab index when keyboard navigation is suppressed', () => {
```
jperals, line 23: Keyboard navigation being "active" is the same as it being "suppressed"? I think I understand this but the terms are a bit confusing here. If I understand correctly what's happening is that there are two kinds of "navigation": the default, browser-native keyboard navigation and the single-stop grid navigation that is being implemented. Can we make this more clear even if it's just in the test descriptions?

pan-kot: Thanks! When the navigation is "active" the table instructs children elements to "mute" user focus and it also listens to keyboard commands. However, some children elements can be exempted. When the focus lands on those elements the navigation is suppressed. That is to support inline-editing scenario when during the editing the table navigation should not intervene and the focus should not be muted for the elements inside of the dialog. I updated the test for it to be hopefully better clear.

src/table/table-role/grid-navigation.tsx
```diff
@@ -92,17 +100,30 @@ class GridNavigationProcessor {
   }
 
   public refresh() {
-    if (this._table) {
-      // Update focused cell indices in case table rows, columns, or firstIndex change.
-      if (this.focusedCell) {
-        this.focusedCell = this.findFocusedCell(this.focusedCell.element);
+    // Timeout ensures the newly rendered content elements are registered.
+    setTimeout(() => {
```
jperals, line 104: Would [`flushSync`](https://react.dev/reference/react-dom/flushSync) be useful here to avoid timeouts?
> Call `flushSync` to force React to **flush any pending work** and update the DOM synchronously.

pan-kot: It would be neat to have flushAsync of some sort to run the update when ready but not forcing anything. The refresh is called with every render and the flushSync can cause significant performance impact.

jperals (approved):

PR #1860, opened 2024-01-04
chore: Table grid navigation register context

pan-kot:
### Description

Implementing new technical design for table grid navigation, see KtAVARsTM7ko (documentation), https://github.com/cloudscape-design/components/pull/1856 (next PR), https://github.com/cloudscape-design/components/pull/1760 (POC).

In the PR I am adding grid navigation context. The context is to be used to control focusability of the interactive elements inside the table. The interactive elements must adjust the tab index when used inside a table with keyboard navigation active.

### How has this been tested?

New unit tests.

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/button/internal.tsx
```diff
@@ -134,9 +135,16 @@ export const InternalButton = React.forwardRef(
       [styles['full-width']]: shouldHaveContent && fullWidth,
     });
 
+    const explicitTabIndex =
+      __nativeAttributes && 'tabIndex' in __nativeAttributes ? __nativeAttributes.tabIndex : undefined;
```
timogasda, line 139: Could this be simplified to `__nativeAttributes?.tabIndex` ?

src/checkbox/__tests__/checkbox.test.tsx
```diff
@@ -220,3 +222,34 @@ test('Should set aria-describedby and aria-labelledby from ariaLabelledby and ar
   expect(checkboxInputAriaDescribedby).toBe('description-id' + ' ' + toggleDescriptionId);
   expect(checkboxInputAriaLabelledby).toBe(toggleLabelId + ' ' + 'label-id');
 });
+
+describe('table grid navigation support', () => {
+  test('does not override tab index when keyboard navigation is not active', () => {
+    renderWithGridNavigation({ target: null }, <Checkbox id="checkbox" checked={false} />);
+    expect(document.querySelector('#checkbox input')).not.toHaveAttribute('tabIndex');
```
timogasda, line 229: Let's use our official selector: `checkboxWrapper.findNativeInput().getElement()`

src/link/__tests__/index.test.tsx
```diff
@@ -274,3 +275,27 @@ describe('Link component', () => {
     });
   });
 });
+
+describe('table grid navigation support', () => {
+  test('does not override tab index for button link when keyboard navigation is not active', () => {
+    renderWithGridNavigation({ target: null }, <Link id="link" />);
+    expect(document.querySelector('#link')).toHaveAttribute('tabIndex', '0');
```
timogasda, line 282: I suggest we stick to our test utils

src/radio-group/__tests__/radio-group.test.tsx
```diff
@@ -308,3 +309,25 @@ describe('value', () => {
...
   });
 });
+
+describe('table grid navigation support', () => {
+  test('does not override tab index when keyboard navigation is not active', () => {
+    renderWithGridNavigation(
+      { target: null },
+      <RadioGroup id="radio" value={null} items={[{ value: '1', label: 'One' }]} />
+    );
+    expect(document.querySelector('#radio input')).not.toHaveAttribute('tabIndex');
```
timogasda, line 319: Test utils please :)

src/table/table-role/__tests__/grid-navigation-context.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+import { useGridNavigationFocusable } from '../../../../lib/components/table/table-role';
+import { renderWithGridNavigation } from './utils';
+
+function Button(props: React.HTMLAttributes<HTMLButtonElement>) {
+  const buttonRef = useRef<HTMLButtonElement>(null);
+  const { tabIndex } = useGridNavigationFocusable(buttonRef, { tabIndex: props.tabIndex });
+  return <button {...props} ref={buttonRef} tabIndex={tabIndex} />;
+}
+
+function ButtonAlt(props: React.HTMLAttributes<HTMLButtonElement>) {
```
timogasda, line 16: For better understanding, should we call these buttons something like `ButtonWithRef` and `ButtonWithCallback`?

src/table/table-role/__tests__/grid-navigation-context.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+  const getElement = useCallback(() => document.querySelector(`#${props.id}`), [props.id]);
+  const { tabIndex } = useGridNavigationFocusable(getElement, { tabIndex: props.tabIndex });
+  return <button {...props} tabIndex={tabIndex} />;
+}
+
+test.each([Button, ButtonAlt])('does not override tab index when keyboard navigation is not active', Button => {
```
timogasda, line 22: Will this produce two test cases with the exact same name? Or is jest smart enough to somehow add the name of the Button component into the test case?

I'm asking because imagine a scenario where this test fails but only for `ButtonAlt`. Would you be able to tell from the Jest output which button variant failed?

timogasda (changes requested):

src/table/table-role/__tests__/utils.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+function FakeGridNavigationProvider({ target, children }: { target: null | string; children: React.ReactNode }) {
+  const [focusTarget, setFocusTarget] = useState<null | Element>(null);
+  useEffect(() => {
+    if (!target) {
+      setFocusTarget(null);
+    } else {
+      setFocusTarget(document.querySelector(target));
+    }
+  }, [target]);
```
just-boris, line 16: This looks like too much complexity for a mock. Does it really need that much to represent the real code?

src/table/table-role/__tests__/grid-navigation-context.test.tsx
pan-kot: Makes sense, I will think on how to modify the name depending on the selected case

src/table/table-role/grid-navigation-context.tsx
```diff
@@ -0,0 +1,43 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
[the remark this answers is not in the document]
pan-kot: The grid navigation is a feature of the table and those interactive elements that can be rendered in the content have a dependency on the table as result. I would move it to a shared folder in case there is a indeed an opportunity to reuse the logic but it is hard to make this call just yet as it is not clear which parts can be shared. Say, the navigation commands in a different component with a similar focus overtake can be different.

src/table/table-role/__tests__/utils.tsx
```diff
@@ -0,0 +1,26 @@
...
+  return (
+    <GridNavigationContext.Provider value={{ focusTarget, keyboardNavigation: !!focusTarget }}>
+      {children}
+    </GridNavigationContext.Provider>
+  );
+}
+
+export function renderWithGridNavigation({ target }: { target: null | string }, ui: React.ReactNode) {
```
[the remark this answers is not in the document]
pan-kot: I will take a look and try to make it closer

src/table/table-role/__tests__/utils.tsx
pan-kot: The intention is to provide the expected state to the context which is a boolean flag and an html element. The challenge with the element is that it needs to be obtained. The mock can be made simpler by accepting the element as a prop but then using the mock is going to be more challenging.

src/table/table-role/grid-navigation-context.tsx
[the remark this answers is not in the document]
just-boris: Why not now?

src/table/table-role/interfaces.ts
```diff
@@ -18,3 +18,5 @@ export interface FocusedCell {
   cellElement: HTMLTableCellElement;
   element: HTMLElement;
 }
+
+export type FocusableDefinition = (() => HTMLElement) | React.RefObject<HTMLElement>;
```
just-boris, line 22: If this is our internal only API, can we agree on one format and use only the callback or only the refs?

src/table/table-role/__tests__/utils.tsx
just-boris: An alternative option

```js
let setCurrentTarget;
render(<FakeGridNavigationProvider ref={el => setCurrentTarget = el}><Button /></FakeGridNavigationProvider>)

setCurrentTarget(wrapper.findButton().getElement());
```

what pros and cons do you see here and there?

src/button/internal.tsx
pan-kot: I can't because of the __nativeAttributes type:

```
(React.HTMLAttributes<HTMLAnchorElement> & React.HTMLAttributes<HTMLButtonElement>) | Record<`data-${string}`, string>
```

src/table/table-role/grid-navigation-context.tsx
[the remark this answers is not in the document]
pan-kot: Actually, I was thinking more about the grid navigation feature which is coupled to the table as of yet. The context is less coupled though. What it actually controls is focusability of the elements. When present, the elements underneath can be instructed to "mute" user focus so it can be some generic UserFocusController context or alike. However, we might want to make alterations to some elements like select/multiselect when used within keyboard navigation and this can in theory be different depending on what kind of navigation is that.

What do you think if I move the context to under /internal and rename as such:
```
interface SingleTabStopNavigationContext {
  focusTarget: null | Element;
  navigationActive: boolean;
}
```

This can be extended in the future with more properties to carry more information the components can use to alter their behaviour if needed.

timogasda (dismissed):

src/table/table-role/grid-navigation-context.tsx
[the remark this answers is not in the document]
just-boris: Everything that does not make every component importing from `../table`, works

just-boris (changes requested):
the imports `../table/table-role` should move to more central location under `internal`.

Having it inside `table` creates problems in the future to unwind the dependencies, so it is a must have before the merge

just-boris (approved):

PR #1863, opened 2024-01-05
test: Introduce snapshot testing for test-utils selectors

pan-kot:
### Description

A new test for test-utils selectors extracts all referenced selectors, groups them by component, and captures a snapshot. It makes it visible when selectors used in test-utils are updated which might cause a mismatch between components and test infrastructure in case different versions of the package are used.

The selectors copy the trim-hash approach from https://github.com/cloudscape-design/test-utils/blob/main/packages/core/src/utils.ts#L51. We can consider adding the utility to extract selectors to the test-utils package to be consumed in the test.

See: https://sim.amazon.com/issues/AWSUI-22332

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


package-lock.json
```diff
@@ -14207,9 +14207,9 @@
       }
```
abdhalees, line 14207: are these changes necessary?

pan-kot: No, those are just resulted from running npm install - can remove those from the pr

src/test-utils/dom/collection-preferences/sticky-columns-preference.ts
```diff
@@ -10,12 +10,12 @@ export default class StickyColumnsPreferenceWrapper extends ComponentWrapper {
   static lastRootSelector = styles['sticky-columns-last'];
 
   findTitle(): ElementWrapper {
-    return this.findComponent(`.${styles[`sticky-columns-form-field`]}`, FormFieldWrapper)!.findLabel()!;
+    return this.findComponent(`.${styles['sticky-columns-form-field']}`, FormFieldWrapper)!.findLabel()!;
```
just-boris, line 13: Why is this needed? What happens with the extractor when a template string used?

src/__tests__/test-utils-selectors.test.tsx
```diff
@@ -0,0 +1,141 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import fs from 'fs';
+import Path from 'path';
+import glob from 'glob';
+import { transformSync, types, PluginObj, NodePath } from '@babel/core';
```
just-boris, line 7: This module is not declared in our dependencies. You should `npm install --save-dev @babel/core` to not break our builds.

src/__tests__/test-utils-selectors.test.tsx
```diff
@@ -0,0 +1,141 @@
...
+import { flatten, zip } from 'lodash';
+
+// The test extracts generated selectors from the compiled output and matches those against the snapshot.
+test('test-utils selectors', () => {
+  // Find referenced selector files and properties.
+  const selectorsFilePathToUsedProperties = new Map<string, Set<string>>();
+  for (const file of glob.sync('lib/components/test-utils/selectors/**/*.js')) {
```
just-boris, line 14: Are you sure it is a good idea to exctract this info from the compiled code? Would it be easier to extract it from typescript sources? (`import` syntax is much easier to statically analyse, I can see how the code below simplifies)

src/__tests__/test-utils-selectors.test.tsx
```diff
@@ -0,0 +1,141 @@
...
+    const selectorVars = new Map<string, string>();
+    return {
+      visitor: {
+        // Find require statements that import selectors.
+        CallExpression(path: NodePath<types.CallExpression>) {
+          if (path.node.callee.type === 'Identifier' && path.node.callee.name === 'require') {
+            const argument = path.node.arguments[0];
+            if (argument.type === 'StringLiteral' && argument.value.endsWith('selectors.js')) {
+              if (path.parent.type === 'VariableDeclarator' && path.parent.id.type === 'Identifier') {
+                selectorVars.set(path.parent.id.name, Path.resolve(file, '..', argument.value));
```
just-boris, line 44: Why is there an extra `..`? At least it needs an explaining comment

src/__tests__/test-utils-selectors.test.tsx
```diff
@@ -0,0 +1,141 @@
...
+              }
+            }
+          }
+        },
+        // Find selector references and extract used property names.
+        MemberExpression(path: NodePath<types.MemberExpression>) {
+          function parseNested(expression: types.MemberExpression, propertyName: string) {
+            if (expression.object.type === 'Identifier' && selectorVars.has(expression.object.name)) {
+              const filePath = selectorVars.get(expression.object.name)!;
```
just-boris, line 53: This fails if it ever hits this code


```js
import styles from '...../selectors.js'

findSomething() {
   const styles = getSomethingElse();
   return this.find(styles[....]); // this will be incorrectly classified as coming from the import
}
```

Babel has functionality to track variables and where they are coming from: https://github.com/jamiebuilds/babel-handbook/blob/master/translations/en/plugin-handbook.md#toc-scopes

Use scopes to track this properly

src/__tests__/test-utils-selectors.test.tsx
```diff
@@ -0,0 +1,141 @@
...
+    if (!element) {
+      continue;
+    } else if (element.type === 'TemplateElement') {
+      literal += element.value.raw;
+    } else if (element.type === 'Identifier') {
+      literal += '*';
+    } else if (element.type === 'NumericLiteral') {
+      literal += element.value;
+    } else {
+      throw new Error('Unhandled template literal structure.');
```
just-boris, line 124: Can we also add file name and line number to the error message to make it easier for other devs to investigate the issues?

Same applies to all errors here. Also, another reason to switch to the source parsing, to make it easier to track down issues.

src/__tests__/test-utils-selectors.test.tsx
```diff
@@ -0,0 +1,141 @@
...
+    }
+  }
+  return literal;
+}
+
+// Match property against the used properties supporting the wildcards ("*").
```
just-boris, line 130: In which condition is there a wildcard?

src/test-utils/dom/collection-preferences/sticky-columns-preference.ts
pan-kot: This change is not required. The templates strings like these ones are handled just fine. If a template string includes an identifier it is then replaced with a wildcard, e.g. `flash-type-*` to match all properties starting from `flash-type-`.

src/__tests__/test-utils-selectors.test.tsx
pan-kot: Thanks, I will try this out.

src/__tests__/test-utils-selectors.test.tsx
pan-kot: Thanks, I will add a comment

src/__tests__/test-utils-selectors.test.tsx
pan-kot: Explained here: https://github.com/cloudscape-design/components/pull/1863#discussion_r1444768436

src/__tests__/test-utils-selectors.test.tsx
pan-kot: This should be visible in the stack trace. I also plan to move the extractor code to the test-utils package eventually so it can be reused across packages.

just-boris: > This should be visible in the stack trace

Can you show an example?

src/test-utils/dom/collection-preferences/sticky-columns-preference.ts
just-boris: Let's just throw an error saying that dynamic property access is not allowed and simplify everything

pan-kot: I think it works well so far. An alternative could be to update the source code of the test-utils, for example, instead of `styles[`flash-type-${type}`]` we can do:

```
function getFlashStyleByType(type) {
  switch (type) {
    case 'error':
      return styles.error;
    case 'in-progress':
      return styles['in-progress'];
    // ...
  }
}
```

But I would rather keep the current solution with the extractor.

src/__tests__/test-utils-selectors.test.tsx
pan-kot: Ah, no. Sorry, I misunderstood the suggestion. I will add file path and line number to every thrown exception.

src/test-utils/dom/collection-preferences/sticky-columns-preference.ts
just-boris: Do we have any need for this case today? If not, why do you need to think about it?

pan-kot: Yes, there are 22 selectors matched this way:
```
"awsui_content-display-description_tc96w",
"awsui_content-display-option-content_tc96w",
"awsui_content-display-option-list_tc96w",
"awsui_content-display-option_tc96w",
"awsui_content-display-title_tc96w",
"awsui_visible-content-group-label_tc96w",
"awsui_visible-content-group_tc96w",
"awsui_visible-content-groups_tc96w",
"awsui_visible-content-option-label_tc96w",
"awsui_visible-content-option_tc96w",
"awsui_visible-content-title_tc96w",
"awsui_visible-content-toggle_tc96w",
"awsui_flash-type-error_1q84n",
"awsui_flash-type-in-progress_1q84n",
"awsui_flash-type-info_1q84n",
"awsui_flash-type-success_1q84n",
"awsui_flash-type-warning_1q84n",
"awsui_result-container-error_11huc",
"awsui_result-container-success_11huc",
"awsui_navigation-link-active_1xupv",
"awsui_navigation-link-disabled_1xupv",
"awsui_navigation-link-label_1xupv",
```

src/__tests__/test-utils-selectors.test.tsx
```diff
@@ -0,0 +1,141 @@
...
+function matchProperties(usedProperties: string[], property: string) {
+  for (const testProperty of usedProperties) {
+    if (testProperty === property) {
+      return true;
+    }
+    if (testProperty.includes('*') && new RegExp(testProperty.replace(/\*/g, '.*')).test(property)) {
```
just-boris, line 136: Dynamically constructing regexp is unsafe, because there could be other special characters which can be treated as control statements. 

Learn how to write safe regexps and update

src/__tests__/test-utils-selectors.test.tsx
pan-kot: This actually fails in the described scenario when running the script against uncompiled sources (in the compiled ones the var names are generated). However, in the worst case scenario the output would contain a few extra selectors which is fine. The case is extremely unlikely though: not only the variable name has to be duplicated, the accessed property must exist. I propose to not introduce scopes tracking unless there is a simple way of adding it.

src/__tests__/test-utils-selectors.test.tsx
pan-kot: How much do we need it for this case? Yes, there can be in theory some weird template string so that the resulting expression will match more or less properties in the worst scenario.

just-boris: If you want to have a weaknesses in your solution, I cannot do anything with that

just-boris: But at least you could add `^` and `$` to ensure it matches the full strings only

src/__tests__/test-utils-selectors.test.tsx
```diff
@@ -0,0 +1,144 @@
...
+  const splitSelector = selector.replace('.', '').split('_');
+  if (splitSelector.length >= 5) {
+    splitSelector.splice(splitSelector.length - 2, splitSelector.length);
+    return splitSelector.join('_');
+  }
+  return selector;
+}
+
+function resolveSelectorsPath(importPath: string) {
+  return 'lib/components/' + importPath.replace(/\.\.\//g, '');
```
just-boris, line 125: Still unclear what is happening here. Why did you not use `path.resolve(from, to)` to follow relative paths?

src/__tests__/test-utils-selectors.test.tsx
just-boris: This is what you needed to do


```js
export default function (babel) {
  const { types: t } = babel;

  return {
    name: "ast-transform", // not required
    visitor: {
      MemberExpression(path) {
        const binding = path.scope.getBinding(path.node.object.name);
        if (binding.kind !== 'module' || !binding.path.parent.source.value.endsWith('selectors.js')) {
          return;
        }
        // this is an expected import
      }
    }
  };
}
```

Live demo: https://astexplorer.net/#/gist/9fcb7096d184ca40423225db5f41baeb/1f4d1737b0dfd6fb7fa025ec9da416f3dd1a1f93

Do you still think it is difficult?

pan-kot: Many thanks for the reference, I was not familiar with getBinding API. I will add the code

src/__tests__/test-utils-selectors.test.tsx
pan-kot: The import path is taken from `src/test-utils/selectors/{component}`. The selectors path should resolve to `lib/components/{component}/styles.selectors.js`.

If using `Path.resolve` this can be achieved with e.g.
```
Path.resolve(testUtilFilePath, importPath).replace('/src/', '/lib/')
```

I like the current approach better.

Though, this needs to be parametrised somewhat when moving the extractor part to the test-utils package so that it can be reused.

pan-kot: The import path is taken from `src/test-utils/selectors/{component}`. The selectors path should resolve to `lib/components/{component}/styles.selectors.js`.

If using `Path.resolve` this can be achieved with e.g.
```
Path.resolve(testUtilFilePath, importPath).replace('/src/', '/lib/')
```

I like the current approach better.

Though, this needs to be parametrised somewhat when moving the extractor part to the test-utils package so that it can be reused.

pan-kot: The solution does support nested paths like `../../../app-layout/test-classes/styles.selectors.js` but I will give your solution a try

src/__tests__/test-utils-selectors.test.tsx
```diff
@@ -0,0 +1,162 @@
...
+          }
+        },
+      },
+    } as PluginObj;
+  }
+  const source = fs.readFileSync(file, 'utf-8');
+  transformSync(source, { babelrc: false, configFile: false, plugins: [...defaultPlugins, extractor] })?.code;
+}
+
+function extractComponentSelectors(file: string, usedProperties: string[], onExtract: (selector: string) => void) {
```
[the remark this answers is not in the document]
pan-kot: Updated

src/__tests__/test-utils-selectors.test.tsx
```diff
@@ -0,0 +1,162 @@
...
+        ImportDeclaration(path: NodePath<types.ImportDeclaration>) {
+          if (path.node.source.value.endsWith('selectors.js') && path.node.specifiers.length === 1) {
+            const specifier = path.node.specifiers[0];
+            if (specifier.type === 'ImportDefaultSpecifier') {
+              selectorToFilePath.set(specifier.local.name, resolveSelectorsPath(path.node.source.value));
+            } else {
+              throw new Error(`Unhandled styles import type at ${file}:${path.node.loc?.start.line}.`);
+            }
+          }
+        },
```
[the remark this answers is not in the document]
pan-kot: Updated

src/__tests__/test-utils-selectors.test.tsx
pan-kot: Updated

PR #1876, opened 2024-01-10
chore: Replace use of all: initial with manual list of properties and values

DaemonCahill:
This PR removes the use of `all: initial` which is replaced at build time by a list of values contained in the postCSSInitial plugin. That package is no longer maintained, and our use was only for inherited properties. Given that there is no native way to target only inherited properties, `all: initial` will be replaced by a manual list of properties and values we wish to reset to fully encapsulate the styles of our components.


just-boris (approved):

src/internal/styles/utils/styles-reset.scss
```diff
@@ -6,8 +6,30 @@
 @use '../typography' as typography;
 
 @mixin styles-reset {
-  /* stylelint-disable-next-line plugin/no-unsupported-browser-features */
-  all: initial;
```
timogasda, line 10: > Given that there is no native way to target only inherited properties

Isn't that exactly what `all: initial` does? We only used the postcss plugin for backwards compatibility with older browsers like IE11. Why can't we just use `all: initial`?

DaemonCahill: Interestingly, no. My understanding is that our production code never actually had `all: initial` in it. Instead, the `postCSSInitial` plugin would convert that property declaration into a list of property declarations. Additionally, we configured the plugin to only impact _inherited_ properties. When I tried to use the native behavior of `all: initial` it impacted _all_ properties and broke many things unintentionally. Based on scouring MDN it doesn't appear that there is an equivalent property for _all (inherited) properties_ which is how we were actually using the plugin/property combination. So I resorted to just setting the list manually.

DaemonCahill: As a side note, if you are incredulous and want to do a little research on your own, I would gladly welcome the second set of eyes. I am still in partial disbelief that there isn't an easier way to accomplish setting only inherited properties to their initial value.

just-boris: In other words, our `postcss-initial` polyfill was too loose and did not represent the native behavior fully.

We briefly tried to fix all regressions caused by switching to the genuine property, but it was too much for this project. So, this PR implements a safe option – just do the job of the polyfill but as a SASS mixin.

@timogasda if you think we should pursue the native property adoption, we can sync in DM, we will show you what exact work it means.

PR #1883, opened 2024-01-15
feat: Introduce flag to remove high-contrast header

fralongo:
### Description

- Addition of global flags
- Addition of first global flag to remove high-contrast header

### How has this been tested?

 Locally

<details>

#### Correctness

- _Changes are backward-compatible if not indicated, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#public-apis)._ Yes
- _Changes do not include unsupported browser features, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#browsers-support)._ Yes
- _Changes were manually tested for accessibility, see [accessibility guidelines](https://cloudscape.design/foundation/core-principles/accessibility/)._ Yes

#### Security

- _If the code handles URLs: all URLs are validated through [the `checkSafeUrl` function](https://github.com/cloudscape-design/components/blob/main/src/internal/utils/check-safe-url.ts)._ N/A

#### Testing

- _Changes are covered with new/existing unit tests?_ Yes
- _Changes are covered with new/existing integration tests?_ N/A
</details>

---

### DIFF
--- a/pages/app/index.html
+++ b/pages/app/index.html
@@ -27,7 +27,7 @@
       if (location.hash.includes('file-upload')) {
         csp['img-src'] = 'blob:';
       }
-
+      
       const cspString = Object.entries(csp)
         .map(([key, value]) => `${key} ${value}`)
         .join('; ');
--- a/pages/app/index.tsx
+++ b/pages/app/index.tsx
@@ -17,9 +17,15 @@ import Header from './components/header';
 import StrictModeWrapper from './components/strict-mode-wrapper';
 import AppContext, { AppContextProvider, parseQuery } from './app-context';
 
+interface GlobalFlags {
+  removeHighContrastHeader?: boolean;
+}
 const awsuiVisualRefreshFlag = Symbol.for('awsui-visual-refresh-flag');
+const awsuiGlobalFlagsSymbol = Symbol.for('awsui-global-flags');
+
 interface ExtendedWindow extends Window {
   [awsuiVisualRefreshFlag]?: () => boolean;
+  [awsuiGlobalFlagsSymbol]?: GlobalFlags;
 }
 declare const window: ExtendedWindow;
 
@@ -77,10 +83,15 @@ function App() {
 }
 
 const history = createHashHistory();
-const { direction, visualRefresh } = parseQuery(history.location.search);
+const { direction, visualRefresh, removeHighContrastHeader } = parseQuery(history.location.search);
 
 // The VR class needs to be set before any React rendering occurs.
 window[awsuiVisualRefreshFlag] = () => visualRefresh;
+if (removeHighContrastHeader) {
+  window[awsuiGlobalFlagsSymbol] = {
+    removeHighContrastHeader: true,
+  };
+}
 
 // Apply the direction value to the HTML element dir attribute
 document.documentElement.setAttribute('dir', direction);
--- a/pages/app/index.tsx
+++ b/pages/app/index.tsx
@@ -87,10 +87,11 @@ const { direction, visualRefresh, removeHighContrastHeader } = parseQuery(histor
 
 // The VR class needs to be set before any React rendering occurs.
 window[awsuiVisualRefreshFlag] = () => visualRefresh;
+if (!window[awsuiGlobalFlagsSymbol]) {
+  window[awsuiGlobalFlagsSymbol] = {};
+}
 if (removeHighContrastHeader) {
-  window[awsuiGlobalFlagsSymbol] = {
-    removeHighContrastHeader: true,
-  };
+  window[awsuiGlobalFlagsSymbol].removeHighContrastHeader = true;
 }
 
 // Apply the direction value to the HTML element dir attribute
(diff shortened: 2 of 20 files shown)


pages/app/index.html
```diff
@@ -27,12 +27,19 @@
...
       }
-
+      
       const cspString = Object.entries(csp)
         .map(([key, value]) => `${key} ${value}`)
         .join('; ');
 
       document.write('<meta http-equiv="Content-Security-Policy" content="' + cspString + '" />');
+      
+      if (location.hash.includes('removeHighContrastHeader')) {
```
just-boris, line 37: This script is only for building the CSP configuration, because it is not possible to set it otherwise.

Can we keep this script clean and set it in a more appropriate place, same as we do for visual refresh: https://github.com/cloudscape-design/components/blob/9f65ea408177be13561dc2c2e2f56e90f3b1f351/pages/app/index.tsx#L82-L86

fralongo: Great point, addressed

src/internal/utils/global-flags.ts
```diff
@@ -0,0 +1,29 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+export const awsuiGlobalFlagsSymbol = Symbol.for('awsui-global-flags');
+
+interface GlobalFlags {
+  removeHighContrastHeader?: boolean;
+}
+
+interface ExtendedWindow extends Window {
+  [awsuiGlobalFlagsSymbol]?: GlobalFlags;
```
connorlanigan, line 10: We could potentially add something like `top: ExtendedWindow` to this interface, so that we don't have to do the type assertion for `window.top` below

connorlanigan (changes requested):

src/internal/utils/global-flags.ts
fralongo: I tried that, but it created a recursive structure that caused complications in testing

src/internal/utils/global-flags.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export const getGlobalFlag = (flagName: keyof GlobalFlags): GlobalFlags[keyof GlobalFlags] | undefined => {
+  let globalFlags: GlobalFlags = {};
+  const topWindow = getTopWindow();
+  if (typeof window !== 'undefined') {
+    if (window[awsuiGlobalFlagsSymbol]) {
+      globalFlags = window[awsuiGlobalFlagsSymbol];
+    } else if (topWindow && topWindow[awsuiGlobalFlagsSymbol]) {
+      globalFlags = topWindow[awsuiGlobalFlagsSymbol];
+    }
```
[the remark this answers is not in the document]
fralongo: We currently only support one flag, I wouldn't even be able to test this dedupe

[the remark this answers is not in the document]
fralongo: Done

src/internal/utils/global-flags.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export const getTopWindow = (): ExtendedWindow | undefined => {
+  return window.top as ExtendedWindow;
+};
+
+export const getGlobalFlag = (flagName: keyof GlobalFlags): GlobalFlags[keyof GlobalFlags] | undefined => {
+  try {
+    if (typeof window !== 'undefined') {
+      const topWindow = getTopWindow();
+      return window[awsuiGlobalFlagsSymbol]?.[flagName] ?? topWindow?.[awsuiGlobalFlagsSymbol]?.[flagName];
```
just-boris, line 22: I think we should keep `window.top` and `window` separately. Because if something goes wrong in `getTopWindow` , we do not want to bail out immediately.

Something like this:

```js
function readFlag(window) {
  return window[awsuiGlobalFlagsSymbol]?.[flagName]
}

try {
   const ownFlag = readFlag(window);
   if(ownFlag !== undefined) {
       return ownFlag;
   }
   return readFlag(getTopWindow());
}
```

fralongo: Addressed

just-boris (dismissed):

pages/app/index.tsx
```diff
@@ -77,10 +83,15 @@ function App() {
 }
 
 const history = createHashHistory();
-const { direction, visualRefresh } = parseQuery(history.location.search);
+const { direction, visualRefresh, removeHighContrastHeader } = parseQuery(history.location.search);
 
 // The VR class needs to be set before any React rendering occurs.
 window[awsuiVisualRefreshFlag] = () => visualRefresh;
+if (removeHighContrastHeader) {
+  window[awsuiGlobalFlagsSymbol] = {
```
connorlanigan, line 91: I'd recommend that we only set the individual flag of the object, currently with this code we'd override any existing flag object

connorlanigan (changes requested):

pages/app/index.tsx
fralongo: Addressed

connorlanigan (dismissed):

just-boris (approved):

just-boris (approved):

PR #1885, opened 2024-01-15
fix: Prevent filtering highlighting from crashing when match is too long

avinashbot:
### Description

Quick fix for a rare case. Imagine pasting a 100,000 character string into the autosuggest component 😨 

Tested manually on Chrome and Firefox as well.

Related links, issue #, if available: P112982985

### How has this been tested?

Added a test case.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/components/option/highlight-match.tsx
+++ b/src/internal/components/option/highlight-match.tsx
@@ -5,6 +5,12 @@ import styles from './styles.css.js';
 import clsx from 'clsx';
 
 const splitOnFiltering = (str: string, highlightText: string) => {
+  // We match by creating a regex using user-provided strings, so we skip
+  // highlighting if the generated regex would be too memory intensive.
+  if (highlightText.length > 100000) {
+    return { noMatches: [str], matches: null };
+  }
+
   // Filtering needs to be case insensitive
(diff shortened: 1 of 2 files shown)


src/internal/components/option/highlight-match.tsx
```diff
@@ -23,7 +23,9 @@ const Highlight = ({ str }: HighlightMatchProps) =>
   str ? <span className={clsx(styles['filtering-match-highlight'])}>{str}</span> : null;
 
 export default function HighlightMatch({ str, highlightText }: HighlightMatchProps) {
-  if (!str || !highlightText) {
+  // splitOnFiltering creates a regex using user-provided strings, so we skip
+  // highlighting if the generated regex would be too memory intensive.
+  if (!str || !highlightText || highlightText.length > 100000) {
```
pan-kot, line 28: Would it make more sense to introduce the check as part of `splitOnFiltering` before the regex is created? When too long the function can return `[str]` as noMatches and `null` as matches.

avinashbot: Both seem equally fine to me. If you find that approach nicer, I could make that change now.

pan-kot (approved):

PR #1893, opened 2024-01-17
feat: Table grid navigation

pan-kot:
### Description

The PR integrates grid navigation (enhanced keyboard navigation) feature into the table component and makes it available through a public API.

Related links:
* API: 5uPHAylzAuFG
* Tech design: KtAVARsTM7ko
* Draft guidelines: U6xBAYcsDaYu
* Bug bash: 8QVkA8fIBunR
* Previous PR: https://github.com/cloudscape-design/components/pull/1856

### How has this been tested?

More integration tests to existing pages (inline editing, resizable columns, shift selection). Screenshot tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---


pages/table/resizable-columns.page.tsx
```diff
@@ -173,6 +175,15 @@ export default function App() {
             <Checkbox checked={withSelection} onChange={event => setUrlParams({ withSelection: event.detail.checked })}>
               With row selection
             </Checkbox>
+            <Checkbox
+              checked={enableKeyboardNavigation}
+              onChange={event => {
+                setUrlParams({ enableKeyboardNavigation: event.detail.checked });
```
jperals, line 181: Enabling keyboard navigation makes all row cells focusable with arrow keys but not header cells (unless they have interactive elements in them). Is this intentional?

jperals: I see that if resizable columns are enabled, then the resizers are focusable and the header cells themselves are not (again, unless they have interactive elements). Just wanted to confirm if this a conscious decision?

pages/table/sticky-columns.page.tsx
```diff
@@ -190,6 +191,16 @@ export default () => {
             >
               Sorting disabled
             </Checkbox>
+
+            <Checkbox
+              checked={urlParams.enableKeyboardNavigation}
+              onChange={event => {
+                setUrlParams({ enableKeyboardNavigation: event.detail.checked });
```
jperals, line 198: Shouldn't the `enableKeyboardNavigation` URL param be applied to the Table(s) below?

src/internal/context/single-tab-stop-navigation-context.tsx
```diff
@@ -37,8 +38,13 @@ export function useSingleTabStopNavigation(focusable: null | FocusableDefinition
 
   useEffect(() => {
     if (focusable) {
-      const changeHandler = (element: null | Element, suppressed: boolean) =>
-        setFocusTargetActive(focusable.current === element || suppressed);
+      const changeHandler = (element: null | Element, suppressed: boolean) => {
```
jperals, line 41: Is this just to avoid calls to `setFocusTargetActive`? Is this extra complexity worth it?

src/table/body-cell/styles.scss
```diff
@@ -32,6 +33,21 @@ $success-icon-padding-right: calc(#{$edit-button-padding-right} + #{$icon-width-
...
   font-weight: inherit;
+
+  @include focus-visible.when-visible {
+    @include styles.focus-highlight(
+      (
+        'vertical': calc(-1 * #{awsui.$space-scaled-xxs}),
+        'horizontal': calc(-1 * #{awsui.$space-scaled-xxs}),
+      )
+    );
+    // @mixin focus-highlight sets cell's position to "relative".
```
jperals, line 44: WDYT about adding an optional parameter to the mixin instead to override the `position`, to avoid these CSS overrides?

src/table/body-cell/td-element.tsx
```diff
@@ -98,10 +103,17 @@ export const TableTdElement = React.forwardRef<HTMLTableCellElement, TableTdElem
           stickyStyles.className
         )}
         onClick={onClick}
+        // Scroll cell into view when it is focused with keyboard.
+        onFocus={event => {
+          if (!event.currentTarget.contains(getMouseDownTarget())) {
+            event.target.scrollIntoView?.({ block: 'center' });
```
jperals, line 109: Why not use our existing utility `scrollElementIntoView`?

src/table/header-cell/index.tsx
```diff
@@ -84,26 +85,31 @@ export function TableHeaderCell<ItemType>({
...
     >
       <div
+        ref={clickableHeaderRef}
         data-focus-id={`sorting-control-${String(columnId)}`}
         className={clsx(styles['header-cell-content'], {
           [styles['header-cell-fake-focus']]: focusedComponent === `sorting-control-${String(columnId)}`,
         })}
         aria-label={
-          column.ariaLabel
+          column.ariaLabel && !sortingDisabled
```
jperals, line 112: Why the change in this condition?

src/table/header-cell/th-element.tsx
```diff
@@ -42,13 +46,24 @@ export function TableThElement({
...
+  const { tabIndex: cellTabIndex } = useSingleTabStopNavigation(cellRefObject);
+  const getMouseDownTarget = useMouseDownTarget();
 
   return (
     <th
+      data-focus-id={`header-${String(columnId)}`}
+      // Scroll cell into view when it is focused with keyboard.
+      onFocus={event => {
+        if (!event.currentTarget.contains(getMouseDownTarget())) {
+          event.target.scrollIntoView?.({ block: 'center' });
```
jperals, line 60: Same here, why not use `scrollElementIntoView`?

pages/table/resizable-columns.page.tsx
pan-kot: Yes, it is. When the column is not sortable but resizable the resize handle is the only focusable element of that column and that's why it is chosen as the navigation target. Likewise, if a body cell would include a text content and a button - only the button but not the text content nor the entire cell will be user-focusable.

pages/table/sticky-columns.page.tsx
pan-kot: It is, in this test page the urlSettings are passed to each table as `{...urlParams}`

src/table/body-cell/td-element.tsx
pan-kot: Thanks, will check it

src/internal/context/single-tab-stop-navigation-context.tsx
pan-kot: Yes, the Lighthouse tests show a noticeable improvement when having this optimisation in place. However, I am changing it further as there is still a small performance impact, see https://github.com/cloudscape-design/components/pull/1901

src/table/header-cell/index.tsx
pan-kot: Oh that was actually not intended - will revert 👍

src/table/body-cell/styles.scss
pan-kot: That is technically possible but as there is no named arguments I'd have to pass the defaults for border-radius and box-shadow plus include it twice (for normal and for sticky cell). I think that is increasing the complexity.

src/table/header-cell/styles.scss
```diff
@@ -17,6 +17,25 @@
...
+
+  @include focus-visible.when-visible {
+    @include styles.focus-highlight(
+      (
+        'vertical': calc(-1 * #{awsui.$space-scaled-xxs}),
+        'horizontal': calc(-1 * #{awsui.$space-scaled-xxs}),
+      )
+    );
+  }
+  // stylelint-disable-next-line selector-combinator-disallowed-list
```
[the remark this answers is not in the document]
pan-kot: The `body[data-awsui-focus-visible='true']` is required to only show focus outline when the keyboard input was detected. We don't document it elsewhere in the code but it could make sense to take a holistic look and maybe introduce a mixin utility.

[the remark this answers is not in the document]
jperals: This block is written twice:
```
    @include styles.focus-highlight(
      (
        'vertical': calc(-1 * #{awsui.$space-scaled-xxs}),
        'horizontal': calc(-1 * #{awsui.$space-scaled-xxs}),
      )
    );
```

src/table/table-role/grid-navigation.tsx
```diff
@@ -74,8 +74,9 @@ class GridNavigationProcessor {
 
   // State
   private focusedCell: null | FocusedCell = null;
-  private focusables = new Map<Element, boolean>();
+  private focusables = new Set<Element>();
   private focusHandlers = new Map<Element, FocusableChangeHandler>();
+  private focusablesState = new WeakMap<Element, boolean>();
```
jperals, line 80: What is the advantage of having `focusables` and `focusablesState` separated?

src/table/header-cell/styles.scss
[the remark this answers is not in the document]
pan-kot: Yes, but these are nested differently, I am not sure it is possible to de-duplicate.

src/table/table-role/grid-navigation.tsx
pan-kot: The entries from `focusablesState` are not removed upon unregistering. This way, if the same DOM node is unregistered and registered again (happens on every element render), certain extra work can be avoided because the focusable state remains cached.

src/table/header-cell/styles.scss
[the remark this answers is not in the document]
jperals: I think we could make a local mixin, but not a blocker.

src/table/header-cell/th-element.tsx
```diff
@@ -42,13 +45,17 @@ export function TableThElement({
...
   });
 
-  const mergedRef = useMergeRefs(stickyStyles.ref, cellRef);
+  const cellRefObject = useRef<HTMLTableCellElement>(null);
+  const mergedRef = useMergeRefs(stickyStyles.ref, cellRef, cellRefObject);
+  const { tabIndex: cellTabIndex } = useSingleTabStopNavigation(cellRefObject);
 
   return (
     <th
+      data-focus-id={`header-${String(columnId)}`}
```
jperals, line 54: Why is this necessary?

pan-kot: The data-focus-id is used to keep focus outlines in sync between header and sticky header. Previously it was only possible to focus on sortable headers, select-all checkbox and resize handles. Now it is also possible to focus on header cells themselves when there are no other focus targets inside.

src/table/use-table-focus-navigation.ts
```diff
@@ -26,6 +27,7 @@ function iterateTableCells<T extends HTMLElement>(
  * @param numRows - The number of rows in the table.
  */
 function useTableFocusNavigation<T extends { editConfig?: TableProps.EditConfig<any> }>(
+  tableRole: TableRole,
```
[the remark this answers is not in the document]
just-boris: This code will exist for the entire v3.0 lifetime because backward compatibility.

Do not treat anything as a temporary solution here.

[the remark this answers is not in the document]
pan-kot: That's fair. I updated the signature.

src/table/use-table-focus-navigation.ts
```diff
@@ -128,15 +138,15 @@ function useTableFocusNavigation<T extends { editConfig?: TableProps.EditConfig<
   );
 
   useEffect(() => {
-    if (!tableRoot.current) {
+    if (!tableRoot.current || tableRole === 'grid') {
```
jperals, line 141: Why not listen for events when `tableRole === 'grid'`?

pan-kot: The "grid" table role is only set when the keyboard navigation is active. It overtakes the table focus navigation developed for tables with inline editing. W/o this, both navigations will be active at the same time causing the cursor to jump two cells at a time in some cases.

jperals (approved):

PR #1901, opened 2024-01-22
refactor: Better efficient grid navigation

pan-kot:
### Description

Refactored grid navigation and single tab stop navigation context for better performance. The main change is that the focusables are registered as elements rather than element references. That makes the overall process simpler and removes the need to iterate over focusables to transform a set of refs into a set of elements.

### How has this been tested?

Existing tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/table/table-role/grid-navigation.tsx
+++ b/src/table/table-role/grid-navigation.tsx
@@ -3,7 +3,12 @@
 
 import React from 'react';
 import { useEffect, useMemo } from 'react';
-import { defaultIsSuppressed, findTableRowByAriaRowIndex, findTableRowCellByAriaColIndex } from './utils';
+import {
+  defaultIsSuppressed,
+  findTableRowByAriaRowIndex,
+  findTableRowCellByAriaColIndex,
+  getClosestCell,
+} from './utils';
 import { FocusedCell, GridNavigationProps } from './interfaces';
 import { KeyCode } from '../../internal/keycode';
 import { useStableCallback } from '@cloudscape-design/component-toolkit/internal';
@@ -11,7 +16,6 @@ import { nodeBelongs } from '../../internal/utils/node-belongs';
 import { getAllFocusables } from '../../internal/components/focus-lock/utils';
 import {
   SingleTabStopNavigationContext,
-  FocusableDefinition,
   FocusableChangeHandler,
 } from '../../internal/context/single-tab-stop-navigation-context';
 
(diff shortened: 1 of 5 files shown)


src/table/table-role/grid-navigation.tsx
```diff
@@ -244,33 +254,25 @@ class GridNavigationProcessor {
...
-    this.focusables.forEach(focusable => {
-      const element = focusable.current;
-      const handler = this.focusHandlers.get(focusable)!;
-      handler(this.focusTarget, this.isSuppressed(element));
-    });
+    for (const [focusable, prevIsFocusable] of this.focusables) {
+      const isFocusable = this.focusTarget === focusable || this.isSuppressed(focusable);
+      if (isFocusable !== prevIsFocusable) {
+        this.focusables.set(focusable, isFocusable);
+        this.focusHandlers.get(focusable)!(isFocusable);
```
pan-kot, line 261: Execute the handler only if the state changes to avoid unneeded re-renders.

src/table/table-role/grid-navigation.tsx
```diff
@@ -70,8 +74,8 @@ class GridNavigationProcessor {
 
   // State
   private focusedCell: null | FocusedCell = null;
-  private focusables = new Set<FocusableDefinition>();
-  private focusHandlers = new Map<FocusableDefinition, FocusableChangeHandler>();
+  private focusables = new Map<Element, boolean>();
```
jperals, line 77: What is the advantage of using a Map of boolean here vs a Set?

pan-kot: The `this.focusables` property keeps all elements that were registered and their state (focusable / not focusable). Can't use a set because it is necessary to iterate over focusables to issue updates or do a cleanup.

src/table/table-role/grid-navigation.tsx
```diff
@@ -244,49 +249,43 @@ class GridNavigationProcessor {
 
   private updateFocusTarget() {
     this.focusTarget = this.getSingleFocusable();
-    this.focusables.forEach(focusable => {
-      const element = focusable.current;
-      const handler = this.focusHandlers.get(focusable)!;
-      handler(this.focusTarget, this.isSuppressed(element));
-    });
+    for (const [focusable, prevIsFocusable] of this.focusables) {
```
jperals, line 252: When I first read this I thought `focusable` and `prevIsFocusable` were both boolean and the latter was keeping the value of the previous render (similar to using our `usePrevious` hook). What about some minor renaming, like

```
for (const [focusableElement, isFocusable] of this.focusables) {
   const newIsFocusable = this.focusTarget === focusable || this.isSuppressed(focusable);
   if (newIsFocusable !== isFocusable) {
    ...
```

jperals (approved):

PR #1902, opened 2024-01-22
feat: Expandable groups support in top navigation menu utilities

pan-kot:
### Description

Updated top navigation API to officially support expandable groups in menu utilities.

See AWSUI-29742

### How has this been tested?

Updated test page

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/top-navigation/interfaces.ts
```diff
@@ -98,6 +99,7 @@ export namespace TopNavigationProps {
     type: 'menu-dropdown';
     description?: string;
     items: ButtonDropdownProps.Items;
+    expandableGroups?: boolean;
```
jperals, line 102: Wouldn't it be better to be able to set this property for each group individually?

pan-kot: That is not supported in the button-dropdown.

pan-kot:
> Looks like there are a couple visual bugs in Visual Refresh:
> 
> * Extra top border displayed at the bottom of the nested group
> * Top border of the parent item increases on hover
> 
> Screenshot 2024-01-22 at 14 54 34

Apparently, the issues are also relevant to the button dropdown and are reproducible in there as well. Fixed in scope of this PR.

src/button-dropdown/category-elements/styles.scss
```diff
@@ -59,20 +59,6 @@
         box-shadow: inset 0 0 0 awsui.$border-control-focus-ring-shadow-spread awsui.$color-border-item-focused;
       }
     }
-
-    &.variant-navigation {
```
pan-kot, line 63: So this one is supposed to be added when expandable-header is set. However, we didn't support expandable menus in top navigation up until now and the styles were causing incorrect menu item display when highlighted.

src/button-dropdown/item-element/styles.scss
```diff
@@ -23,7 +23,7 @@
   &:first-child {
     margin-top: 0;
   }
-  &.last {
+  &:not(.has-expandable-groups).last {
```
pan-kot, line 26: This one is supposed to add dividers between groups when not expandable. Because it was always added the dividers were also added to expandable case causing extra border at the bottom of the list.

pages/button-dropdown/item-element.permutations.page.tsx
```diff
@@ -17,6 +17,7 @@ const permutations = createPermutations<ItemProps>([
     highlighted: [false, true],
     last: [false, true],
     hasCategoryHeader: [false],
+    hasExpandableGroups: [false],
```
jperals, line 20: Are these additions necessary? Won't the component default to false anyway?

src/button-dropdown/item-element/styles.scss
jperals: This applies to the elements _inside_ a group, right? It's not like the element itself has a group. Should we use a naming like `.is-expandable-group` or `.in-expandable-group` instead of `.has-expandable-groups`? Same for the `hasExpandableGroups` prop when passed down to the inner level.

pages/button-dropdown/item-element.permutations.page.tsx
pan-kot: I made the property required to not forget to pass it on accident

src/button-dropdown/item-element/styles.scss
jperals: Approving as I don't consider this critical but would be nice to resolve this to make sure this doesn't confuse us in the future.

jperals (approved):

PR #1906, opened 2024-01-23
fix: Remove html form from table cell inline editor

pan-kot:
### Description

The form we use inside table cell inline editor cause issues if the table is nested as part of another form:
* Causes invalid HTML (form nesting is not allowed);
* In React 17+ causes page refresh.

If we remove the form the issue is gone but then the submit cannot be completing by pressing Enter while focused on an input. Instead, we can:
* Move the form to a portal and refer to it using the `form` attribute. That prevents form nesting and also resolves the issue with page refresh.
* Stop form submit event propagation for it to not reach the outer form. 

See AWSUI-30406

### How has this been tested?

Existing unit- and integration tests, manual testing with React 16 and React 18.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/table/body-cell/inline-editor.tsx
+++ b/src/table/body-cell/inline-editor.tsx
@@ -11,6 +11,8 @@ import { Optional } from '../../internal/types';
 import FocusLock, { FocusLockRef } from '../../internal/components/focus-lock';
 import LiveRegion from '../../internal/components/live-region';
 import { useInternalI18n } from '../../i18n/context';
+import Portal from '../../internal/components/portal';
+import { useUniqueId } from '../../internal/hooks/use-unique-id';
 
 // A function that does nothing
 const noop = () => undefined;
@@ -42,6 +44,7 @@ export function InlineEditor<ItemType>({
   const i18n = useInternalI18n('table');
 
   const focusLockRef = useRef<FocusLockRef>(null);
+  const formId = useUniqueId();
 
   const cellContext = {
     currentValue: currentEditValue,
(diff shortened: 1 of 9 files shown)


src/table/body-cell/inline-editor.tsx
```diff
@@ -112,7 +111,7 @@ export function InlineEditor<ItemType>({
         aria-label={ariaLabels?.activateEditLabel?.(column, item)}
         onKeyDown={handleEscape}
       >
-        <form onSubmit={onSubmitClick} className={styles['body-cell-editor-form']}>
+        <div className={styles['body-cell-editor-form']}>
```
johannes-weber, line 114: I get the motivation behind this change. Out of curiosity: would the change from a `from` to a `div` element impact a11y?  Semantically a form would be correct because the tag contains interactive controls for submitting information ([MDN reference](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form)).

pan-kot: I don't think the form is needed there for accessibility but that needs checking. The editor is wrapped with a dialog role which makes it similar to some other components of ours, for example collection preferences or split panel settings. We do not use the `form` in either.

pan-kot: Oh, and the date range picker as well

just-boris: This breaks auto submitting edits on "Enter" feature. 

I would recommend finding alternative fixes for the issue. Form tag should stay, because it is semantically expected here

pan-kot: Yes, that is true. I updated my implementation.

just-boris: This version does not support custom controls, as it introduces secret props unavailable for consumers.

There must be a simple solution and there is. Talked to the original customer, recommended to abandon inline editing feature all together. 

They do not depend on this PR anymore, so we can close it and claim this use-case unsupported

just-boris:
Closed as the fix is not needed

PR #1911, opened 2024-01-25
feat: Add human-readable format option to Date range picker display

jperals:
### Description

Ticket: `AWSUI-19518`
API proposal: `M4haAbUQB589`

### How has this been tested?

- Unit tests
- Added permutations page to test the component width, which can change depending on the props
- Localizations have been reviewed by the localization team

BackstopJS visual regression tests are failing because of the newly added permutations page, which can't be compared to `main` because it doesn't exist there.

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/date-range-picker/styles.scss
```diff
@@ -14,7 +14,12 @@ $calendar-header-color: awsui.$color-text-body-default;
 
 .root {
   @include styles.styles-reset;
-  max-width: 32em;
+  &:not(.wide) {
+    max-width: 32em;
```
Al-Dani, line 18: Scott' PR (https://github.com/cloudscape-design/components/pull/1851) for A-F components hasn't been merged yet, but I think we will need to update it to `max-inline-size`

src/internal/utils/date-time/format-date-localized.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  if (hideTimeOffset) {
+    return formattedDateTime;
+  }
+
+  const formattedTimeOffset = formatTimeOffset(isoDate, timeOffset);
+  return formattedDateTime + ' ' + formattedTimeOffset;
+}
+
+function formatTimeOffset(isoDate: string, offsetInMinutes?: number) {
```
Al-Dani, line 46: we have similar logic here: https://github.com/cloudscape-design/components/blob/main/src/internal/utils/date-time/format-timezone-offset.ts

Can we try avoiding duplications?

Al-Dani (changes requested):

src/date-range-picker/styles.scss
jperals: Thanks, updated.

src/internal/utils/date-time/format-date-localized.ts
jperals: Note that there are two functions because one localizes the time offset for the ISO format (e.g, `+01:00`) and the other one for the human-readable format (e.g, `(UTC+1)`).

But I now refactored this to make things hopefully a bit more clear. The old `formatTimezoneOffset` function is now `formatTimeOffsetISO`, and the new one is `formatTimeOffsetLocalized`. They are in the same file and the common bits are extracted into small functions.

Also renamed the file `format-timezone-offset.ts` to `format-time-offset.ts`. Time zone (as opposed to time offset) does not play a role in these functions, so I think it's better to avoid mixing both concepts unnecessarily.

Al-Dani (approved):

Al-Dani (dismissed):

PR #1896, opened 2024-01-18
chore: optimization high contrast header removal

at-susie:
### Description
Implement design optimizations for high contrast header removal from VR.

wizard-optimization

Related links, issue #, if available: n/a

### How has this been tested?

<details>
   <summary>Review checklist</summary>

_The following items are to be evaluated by the author(s) and the reviewer(s)._

#### Correctness

- _Changes include appropriate documentation updates._
- _Changes are backward-compatible if not indicated, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#public-apis)._
- _Changes do not include unsupported browser features, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#browsers-support)._
- _Changes were manually tested for accessibility, see [accessibility guidelines](https://cloudscape.design/foundation/core-principles/accessibility/)._

#### Security

- _If the code handles URLs: all URLs are validated through [the `checkSafeUrl` function](https://github.com/cloudscape-design/components/blob/main/src/internal/utils/check-safe-url.ts)._

#### Testing

- _Changes are covered with new/existing unit tests?_
- _Changes are covered with new/existing integration tests?_
</details>

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1917, opened 2024-01-26
fix: Fix content header pointer events

pan-kot:
### Description

Quick fix for a regression introduced in https://github.com/cloudscape-design/components/pull/1896

The new element is preventing header clicks when there is scroll.

https://github.com/cloudscape-design/components/assets/20790937/76c72afe-752e-46c8-b879-cda88cd9ac8c

Ref: AWSUI-31054

### How has this been tested?

Manual testing using `/full-page-table-with-variable-header-height` test page

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/container/styles.scss
+++ b/src/container/styles.scss
@@ -218,6 +218,7 @@
       &::before {
         content: '';
         border-bottom: none;
+        pointer-events: none;
       }
     }
   }
--- a/src/container/styles.scss
+++ b/src/container/styles.scss
@@ -193,6 +193,7 @@
       content: '';
 
       position: absolute;
+      pointer-events: none;
       right: 0;
       left: 0;
       bottom: 0;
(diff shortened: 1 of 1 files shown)


src/container/styles.scss
```diff
@@ -218,6 +218,7 @@
       &::before {
         content: '';
         border-bottom: none;
+        pointer-events: none;
```
michaeldowseza, line 221: This would only fix it in the case when there is a high contrast header. Would this not also need to be applied above to the pseudo element block on line 192?

pan-kot: Yep, updating!

michaeldowseza (approved):

jperals (approved):

jperals:
It would be good to have tests even if it's as follow-up

PR #1918, opened 2024-01-27
chore: Add RTL support for priority components

DaemonCahill:
This PR adds full RTL support for a subset of priority components (`Badge`, `BreadcrumbGroup`, `Button`, `Container`, `Header`, `Icon`, `Link`) in addition to adding a new mixin for detecting document direction for a given subtree.


src/icon/styles.scss
```diff
@@ -42,3 +42,58 @@
...
 }
+
+/*
+Mirror all icons that are assymetrical or indicate a passage of 
+time when the icon direction is RTL. For browsers that do not 
+support the :dir() pseudoclass provide a fallback that depends 
+on the global document dir attribute.
+*/
+/* stylelint-disable plugin/no-unsupported-browser-features */
+@supports selector(:dir(rtl)) {
```
just-boris, line 53: I think it is already time to introduce a reusable mixin, similar to this one: https://github.com/cloudscape-design/components/blob/b958f52a3f68fa733386315e9d283f90c8d8e54e/src/internal/styles/motion/mixins.scss#L10-L22

DaemonCahill: Agreed!

src/internal/styles/direction/index.scss
```diff
@@ -0,0 +1,6 @@
+/*
+ Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ SPDX-License-Identifier: Apache-2.0
+*/
+
+@forward './mixins';
```
just-boris, line 6: Do we really need multiple layers of abstract files which will never come useful?

src/icon/styles.scss
```diff
@@ -42,3 +42,9 @@
   inset-block-start: 0px;
   inset-inline-end: -3px;
 }
+
+@include styles.with-direction(
+  '.icon:is(.name-angle-left-double, .name-angle-left, .name-angle-right-double, .name-angle-right, .name-arrow-left, .name-caret-left-filled, .name-caret-right-filled, .name-audio-full, .name-audio-half, .name-audio-off, .name-external, .name-redo, .name-shrink, .name-undo, .name-view-vertical)'
+) {
```
just-boris, line 48: 1. Need to have one more pass on `$selector` property. In our previous discussions I assumed it will be a short selector like `.icon`. But seeing this real use-case, I am not sure we want to go this way. To be discussed with other team members
2. We should make the 2nd `$direction` argument required too. Implicit `rtl` default is confusing

src/internal/components/dropdown/styles.scss
```diff
@@ -115,3 +115,8 @@
 .stretch-trigger-height {
   block-size: 100%;
 }
+
+@include styles.with-direction('.dropdown:is(.dropdown-drop-left, .dropdown-drop-right)') {
```
just-boris, line 119: For instance, this selector looks okay for embedded string, but the icon one is too long

src/internal/styles/direction/mixins.scss
```diff
@@ -0,0 +1,20 @@
...
+
+/* stylelint-disable @cloudscape-design/no-implicit-descendant, plugin/no-unsupported-browser-features, selector-combinator-disallowed-list */
+@mixin with-direction($selector, $direction: rtl) {
+  @supports selector(:dir(#{$direction})) {
+    #{$selector}:dir(#{$direction}) {
+      @content;
+    }
+  }
+
+  @supports not selector(:dir(rtl)) {
```
just-boris, line 14: Why is this dynamic `:dir(#{$direction})` in the expression before and hard coded here?

src/spinner/styles.scss
```diff
@@ -6,36 +6,44 @@
 @use '../internal/styles' as styles;
 @use './mixins' as mixins;
 @use '../internal/styles/tokens' as awsui;
+@use '../internal/generated/custom-css-properties/index' as custom-props;
```
just-boris, line 9: Do we really need to go this way? Why is this not `transform: scaleX(-1);` trick, same as icons?

src/internal/styles/direction/index.scss
DaemonCahill: I was just following the existing pattern. No opinion here. I can flatten it if you prefer.

just-boris: Yes please. The forward in `src/internal/styles/index.scss` makes sense, but this one does not. It is only useful when you have multiple files to forward

src/icon/styles.scss
DaemonCahill: 1. `Icon` will be the ugliest by far. I would consider this an edge case. We could also simplify the `Icon` code having a single selector passed into the CSS (`icon.should-invert`) and do the computation in JavaScript instead of the long form expression of all of the icon names in the mixin.

2. Also fine by me.

src/internal/components/dropdown/styles.scss
DaemonCahill: See above comment!

src/spinner/styles.scss
DaemonCahill: I was hung up on this for a long time. I started with the same solution as icons but for reasons I don't understand I couldn't get a transform to create an identical inversion. Ultimately I had to change the rotation targets to get it to work. I'll DM you a video.

just-boris (approved):

just-boris (approved):

PR #1927, opened 2024-02-02
fix: Prevent button dropdown divider from being shown for the last item

avinashbot:
### Description

It's the deeply nested item that has the bottom border, which makes it annoying. I tried moving the responsibility of who renders the section divider to a different component, but that had some side effects that made me unsure if I could do this without breaking something somewhere. So we just keep track of when the item or category or expandable section is the last one in the menu, and skip rendering a divider if it is. It's messy, but also the cleanest way of solving this for existing code.

Related links, issue #, if available: AWSUI-21049

### How has this been tested?

Changed some permutation pages. Going to test it in my dev pipeline to see if it's worth adding another page, but I'm okay with this.

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/button-dropdown/items-list.tsx
```diff
@@ -25,9 +27,14 @@ export default function ItemsList({
...
 }: ItemListProps) {
   const isMobile = useMobile();
+  const isVisualRefresh = useVisualRefresh();
 
   const elements = items.map((item, index) => {
     if (!isItemGroup(item)) {
+      const last = index === items.length - 1 || isItemGroup(items[index + 1]);
+      const showDivider =
+        !isLast && isItemGroup(items[index + 1]) && (!hasExpandableGroups || isMobile || !isVisualRefresh);
+
```
abdhalees, line 37: naming here is a bit confusing, what's `last` and what's `isLast`?

avinashbot: - `last` is the prop for the list item - whether the item is last in its category/section
- `isLast` is the prop for the category/section - whether the category/section is the last one in the dropdown

I was considering renaming these props, but decided against it. On second thought, maybe it's best I do.

abdhalees (approved):

PR #1941, opened 2024-02-07
feat: Access autosuggest options from the bottom via up arrow key

jperals:
### Description

The autosuggest component supports accessing the first option from the top by pressing the down arrow key when the input field is focused. This change adds the ability to access the first one from the bottom by pressing the up arrow key.

Notes:
- This change applies specifically to autosuggest and not other components that use the internal dropdown such as button dropdown or select, because it only makes sense for elements that focus the list elements via arrow keys. Button dropdown and select have a different behavior, where the first item from the top is auto-selected on open.

- The behavior does not depend on whether the autosuggest dropdown opens upwards or downwards —that is, users can choose to focus the first or last item of the list unconditionally. This is more in line with native implementations of similar behaviors (for example using `input` and `datalist`), and simplifies the solution a lot.

Ticket: `AWSUI-21659`

### How has this been tested?

Added one integration test

<details>

#### Correctness

#### Security

#### Testing

</details>

---


cansuaa (dismissed):

src/autosuggest/__integ__/autosuggest.test.ts
```diff
@@ -103,6 +103,18 @@ describe(`Simple Autosuggest`, () => {
...
+    'should move selection to the suggestion list from keyboard from bottom to top',
+    setupTest(async page => {
+      await page.focusInput();
+      await page.keys(['option']);
+      await page.keys(['ArrowUp']);
+      await page.assertHighlightedOptionContains('Option');
+      await page.keys(['ArrowUp']);
+      await page.assertHighlightedOptionContains('Option 2');
+    })
+  );
```
johannes-weber, line 116: The `useHighlightedOption` hook is used in the select component as well (https://github.com/cloudscape-design/components/blob/5916c50679362422d86fef72d7c60ce4a8e1ce31/src/select/utils/use-select.ts#L85). 
It kicks in, when when using a select with `filteringType = auto` where no option is selected ([link to select dev page](https://d21d5uik3ws71m.cloudfront.net/components/d2d6f9da4632d334d602dffb8111da4200d7dd28/dev-pages/index.html#/light/select/select.test)). 

Should we add another test for the select component where filteringType is set to auto?

jperals: Actually it uses it regardless of the value of `filteringType`. The condition for the behavior to appear in the Select component is that no option is already selected when opening the dropdown. So this can be seen for example in all the component instances [here](https://d21d5uik3ws71m.cloudfront.net/components/69b0f9acda604584feb5e786ee0edc6b950abe74/dev-pages/index.html#/light/select/select.test) except the first one, which has a pre-selected value. I manually checked all the component's dev pages that the behavior is consistent and added integration tests for Select. No other component besides Autosuggest and Select uses this hook. Thanks for pointing this out!

johannes-weber (approved):

PR #1625, opened 2023-10-06
fix: Make split panel max width calculation deterministic

just-boris:
### Description

Move the code from useEffects into normal code. This prevents layout jitter when resizing split panel or a drawer too quickly, so the effects do not fire on time.

Related links, issue #, if available: AWSUI-22606

### How has this been tested?

* PR build
* Screenshot tests in my pipeline
* Manually to ensure there is no jitter

<details>
   <summary>Review checklist</summary>

_The following items are to be evaluated by the author(s) and the reviewer(s)._

#### Correctness

- _Changes include appropriate documentation updates._
- _Changes are backward-compatible if not indicated, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#public-apis)._
- _Changes do not include unsupported browser features, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#browsers-support)._
- _Changes were manually tested for accessibility, see [accessibility guidelines](https://cloudscape.design/foundation/core-principles/accessibility/)._

#### Security

- _If the code handles URLs: all URLs are validated through [the `checkSafeUrl` function](https://github.com/cloudscape-design/components/blob/main/src/internal/utils/check-safe-url.ts)._

#### Testing

- _Changes are covered with new/existing unit tests?_
- _Changes are covered with new/existing integration tests?_
</details>

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1952, opened 2024-02-12
fix: Make split panel max width calculation deterministic

just-boris:
### Description

Re-apply #1625 but this time using a different approach for reading document width.

See this particular commit for differences to the past attempt: https://github.com/cloudscape-design/components/commit/09d76f9da679cd89e8a683f9372eb3b56c770e0c

### How has this been tested?

* PR build
* Screenshot tests in the dev pipeline
* Verified with the affected customer that this version works for them

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/app-layout/utils/use-content-width.ts
```diff
@@ -3,17 +3,24 @@
...
 
-export default function useAppLayoutOffsets(element: HTMLDivElement | null): { left: number; right: number } {
-  const [offsets, setOffsets] = useState({ left: 0, right: 0 });
+interface AppLayoutOffsets {
+  left: number;
+  right: number;
+  width: number;
+}
+
+export default function useAppLayoutOffsets(element: HTMLDivElement | null): AppLayoutOffsets {
```
avinashbot, line 12: Very minor: with the addition of `width`, this is no longer just providing "widths", so the name of the function seems a little confusing now.

src/app-layout/split-panel/provider.tsx
```diff
@@ -3,58 +3,59 @@
...
+  const [maxHeight, setMaxHeight] = useState(size);
+  const minSize = position === 'bottom' ? SPLIT_PANEL_MIN_HEIGHT : SPLIT_PANEL_MIN_WIDTH;
+  const maxSize = position === 'bottom' ? maxHeight : maxWidth;
   const cappedSize = getLimitedValue(minSize, size, maxSize);
+  const relativeSize = ((size - minSize) / (maxSize - minSize)) * 100;
 
   const onResizeWithValidation = (newSize: number) => {
-    const maxSize = position === 'side' ? getMaxWidth() : getMaxHeight();
-    const isResizeValid = position === 'side' ? maxSize >= MIN_WIDTH : maxSize >= MIN_HEIGHT;
+    const maxSize = position === 'side' ? maxWidth : getMaxHeight();
```
avinashbot, line 30: After changing `getMaxWidth` to `maxWidth`, this value is no longer recalculated after resizing, isn't it?

just-boris: It is now bound to react re-renders. Resize updates the state, the new state produces new width

src/app-layout/utils/use-content-width.ts
just-boris: Tried to come up with a new name, but I could not. `useAppLayoutRect` is the best I could think of. 

What do you think?

avinashbot: I don't have any good suggestions either... `useAppLayoutRect` is probably the best one.

just-boris: Applied

avinashbot (approved):

PR #1958, opened 2024-02-13
revert: Make Calendar and Date Picker code more granularity-agnostic

just-boris:
Reverts cloudscape-design/components#1948

Design token changes caused unforeseen issues. Revering for now


just-boris:
Not needed found a solution on the other end

PR #1961, opened 2024-02-14
fix: Prevent button dropdown divider from being shown for the last item

avinashbot:
### Description

So, this is essentially a minor refactor of the dropdown component. Gently moving the responsibility of rendering bottom borders away from the last item of the dropdown (😱) to the dropdown itself. But because it's a mess of negative margins, I couldn't find a way to make this visually identical. But unlike the time we reverted it last, none of it feels like it worsens any user's experience.

I would invite you to check out my dev pipeline yourself as part of this change: AwsUi-v3-dwaraa

Related links, issue #, if available: AWSUI-21049, AWSUI-32517

### How has this been tested?

Visually. Check my dev pipeline for all the expected changes. There are a bunch, concentrated on the button dropdown tests, but these are all either imperceptible (1px changes) or sometimes even desirable (e.g. alignment of inner dropdowns).

<details>

#### Correctness

#### Security

#### Testing

</details>

---


jperals:
Screenshot 2024-02-27 at 18 06 07

Looks like a border between two groups ("Security" and "Support") is missing in `top-navigation/scenario-full-page`.

src/button-dropdown/item-element/index.tsx
```diff
@@ -49,8 +50,9 @@ const ItemElement = ({
         [styles.highlighted]: highlighted,
         [styles.disabled]: disabled,
         [styles.first]: first,
-        [styles.last]: last,
+        [styles.last]: lastInSection,
```
taheramr, line 53: would it make since to also rename the styles to be `styles.lastInSection`?

jperals (commented):
Looks like [this change](https://github.com/cloudscape-design/components/commit/9600b9f9c4fa4f7ac8e8cb958c7780d0083a76db) fixes the 1px diff, WDYT? I'm deploying this to my pipeline.

src/button-dropdown/item-element/index.tsx
avinashbot: I renamed the component property to lastInSection to disambiguate it from the lastInDropdown property, but this kinda applies to both, so `last` works nicely as a CSS name here. The names don't *have to* match up, so I didn't change it in the CSS. What do you think?

jperals (dismissed):

PR #1962, opened 2024-02-14
fix: Allow reuploading the same file in FileUpload

abdhalees:
### Description

When a user uploads a file then change it's content and uploads it again (with the same name and path) `onChange` event isn't triggered.
this is the native html input behaviour, however if the user uploads the file with drag and drop it `onChange` is triggered, to make the behaviour consistent between our 2 upload methods and to allow customers to validate the input and ask end users to fix the file content and upload it again this PR resets the input value after a file is uploaded.

This change does not have a unit test since unit and e2e testing libraries does not mock the on change event in a way that mimics the real event so it doesn't call on change when the same file is uploaded.
see comments for more details on the unit testing decision.

Related links, issue #, if available: AWSUI-32402

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


package.json
```diff
@@ -60,6 +60,7 @@
     "@rollup/plugin-node-resolve": "^15.0.1",
     "@testing-library/jest-dom": "^5.16.4",
     "@testing-library/react": "^12.1.5",
+    "@testing-library/user-event": "^14.5.2",
```
johannes-weber, line 63: Have you considered any alternatives instead of adding the new dependency?

just-boris: Good catch. This library conflicts with our test utils, so it is better to avoid it.

Related past thread: https://github.com/cloudscape-design/board-components/pull/235#discussion_r1195047786

src/file-upload/__tests__/file-upload.test.tsx
```diff
@@ -63,7 +64,17 @@ function renderStateful(props: Partial<FileUploadProps> = {}) {
...
 function StatefulFileUpload({ value: initialValue = [], ...rest }: Partial<FileUploadProps>) {
   const [value, setValue] = useState(initialValue);
-  return <FileUpload {...defaultProps} {...rest} value={value} onChange={event => setValue(event.detail.value)} />;
+  return (
+    <FileUpload
+      {...defaultProps}
+      {...rest}
+      value={value}
+      onChange={event => {
+        onChange(event);
```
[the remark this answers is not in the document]
just-boris: And also inject it via props. Passing over a global mutable variable is a fast track to make your tests dependent on each other

package.json
abdhalees: I've tried a few different options before using that library: 

1. fire a `change` event on the native input, but that would defeat the purpose of the test because `change` event isn't fired when the same file is re-uploaded, and here I would be forcing to fire which isn't true in a real browser.

2. Add an integ test with text on the page the counts how many times `onChange` was called, but that had some limitations: 
   - I need to add a file in the page so I can upload it and that can be tricky in CI environment.
   -  webdriver upload function is only supported when running Chrome or using a Selenium Grid with the Selenium standalone server: https://webdriver.io/blog/2019/06/25/file-upload/
   - the technique in the above example uses `setValue` and that actually triggers on change even if the same file is re-uploaded so the test becomes trivial.

3. copy this [block of code in user-event library](https://github.com/testing-library/user-event/blob/main/src/utility/upload.ts#L36-L55) that mimics the a real file input behaviour into our test code. 


So I was left with only 2 options: use `@testing-library/user-event` or copy it's `upload` function implementation, I went with the first option.

@just-boris I don't understand from that thread how this library conflicts with our test-utils, the point about to exposing this functionality to customers shouldn't conflict with customers using a popular library to do the same functionality.

I went with adding the library because I saw that it had other useful utilities that could improve other tests, e.g. [Clipboard](@testing-library/user-event) for [CopyToClipboard](https://github.com/cloudscape-design/components/blob/main/src/copy-to-clipboard/__tests__/copy-to-clipboard.test.tsx#L20)


I'm happy to disagree and commit, also let me know if there is a better way to test this other than the ways I tried

just-boris: I think writing an e2e test is the only option here. 

User-event works for testing the change events, but that library has other downsides (in their click events).

So I would prefer not seeing this library at all, than remembering a long list of allowed/not allowed methods

abdhalees: @just-boris e2e test wouldn't work as I mentioned in `2.3` above because it actually calls `onChange` even if the same file is re-uploaded, that has to do with the fact that `<input type='file'>` does not allow setting a value programmatically to anything other than an empty string https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file#notes

so when a e2e test framework sets a value to a file input it's also not the real user experience

just-boris: Discussed in DM. Let's try these options

1. `Simulate.change` from `react-dom/test-utils`
2. Override methods on the input element our own way
3. If both options are unfeasible, let's skip test coverage for this case, unfortunately

abdhalees: tried both options: 
1. Simulate.change has the same behaviour as fireEvent 
2. I tried overriding the methods but seems like it's more complicated than the lines I've included above from user-event, other overrides are needed elsewhere, see [here](https://github.com/testing-library/user-event/blob/main/src/utils/dataTransfer/FileList.ts), [here](https://github.com/testing-library/user-event/blob/main/src/utils/edit/setFiles.ts#L28) and [here](https://github.com/testing-library/user-event/blob/main/src/event/dispatchEvent.ts#L8)
I've tried to only override what's needed for my test but it's not enough, based on the amount of overrides needed I think we're much diverting from a real input element so I just removed the test.

just-boris: Then we call it as "untestable", because testing against user-event does not represent real browser behavior

:(

johannes-weber (approved):
I recomment updating the PR description with the latest state. This would make it easier for future readers why no tests have been added and the thoughts behind without reading through the comments.

PR #1942, opened 2024-02-07
feat: Add support for month granularity in Calendar and Date Picker

jperals:
### Description

Adds month granularity option to Calendar and Date Picker components.

Visual Refresh:
Screenshot 2024-02-16 at 11 15 10

Classic:

Screenshot 2024-02-16 at 11 16 06

Ticket: `AWSUI-8139`

API doc: `GaXpAdXYZDe2`

### How has this been tested?

- Added unit tests
- Visual regression tests will be provided

<details>
   <summary>Review checklist</summary>

_The following items are to be evaluated by the author(s) and the reviewer(s)._

#### Correctness

- _Changes include appropriate documentation updates._
- _Changes are backward-compatible if not indicated, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#public-apis)._
- _Changes do not include unsupported browser features, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#browsers-support)._
- _Changes were manually tested for accessibility, see [accessibility guidelines](https://cloudscape.design/foundation/core-principles/accessibility/)._

#### Security

- _If the code handles URLs: all URLs are validated through [the `checkSafeUrl` function](https://github.com/cloudscape-design/components/blob/main/src/internal/utils/check-safe-url.ts)._

#### Testing

- _Changes are covered with new/existing unit tests?_
- _Changes are covered with new/existing integration tests?_
</details>

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1963, opened 2024-02-14
refactor: Prepare calendar code to support month granularity

jperals:
### Description

Refactor Calendar code to prepare it for month granularity (#1942).

Essentially:
- Refactor the internal Calendar Grid and Calendar Header components to be purely presentational.
- Move some of the logic into separate hooks (for labels, keyboard navigation, and content respectively). These hooks will manage most of the distinct behavior that depends on granularity.
- Minor renaming, e.g, `moveDate` and `getBaseDate` to the more specific `moveDay` and `getBaseDay` respectively so that they can be distinguished from the new counterparts `moveMonth` and `getBaseMonth` which will be introduced.

### How has this been tested?

Existing tests plus 3 more to cover interactions with space key (see code comments).

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/date-picker/__tests__/date-picker-calendar.test.tsx
```diff
@@ -273,26 +275,29 @@ describe('Date picker calendar', () => {
         expect(findCalendarHeaderText(wrapper)).toBe('April 2018');
       });
 
-      test('should allow initially selected date to be re-selected with enter', () => {
-        wrapper.findCalendar()!.findSelectedDate().keydown(KeyCode.enter);
+      test.each(selectionKeys)('should allow initially selected date to be re-selected with %s', key => {
```
jperals, line 278: As the keyboard navigation logic was moved to a new file, Codecov revealed that using space key was not covered, only enter.

src/calendar/styles.scss
```diff
@@ -56,11 +56,11 @@
 
   &-grid {
     inline-size: 100%;
+    table-layout: fixed;
```
jperals, line 59: This rules gives all cells in a table the same width. This way, the same rule can be used for any number of row cells instead of specifying a specific fraction of the total (e.g, 1/7 or 1/3) at the cell level.

https://developer.mozilla.org/en-US/docs/Web/CSS/table-layout

https://caniuse.com/?search=table-layout

src/calendar/use-calendar-grid-keyboard-navigation.ts
```diff
@@ -0,0 +1,68 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
jperals, line 1: Extracted from the Calendar's internal Grid component.

src/calendar/use-calendar-grid-keyboard-navigation.ts
```diff
@@ -0,0 +1,68 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import React from 'react';
+import { CalendarProps } from './interfaces';
+import { moveNextDay, movePrevDay, moveNextWeek, movePrevWeek } from './utils/navigation';
+import { KeyCode } from '../internal/keycode';
+import { isSameMonth } from 'date-fns';
+
+export default function useCalendarGridKeyboardNavigation({
```
timogasda, line 10: Maybe a philosophical question but is it really a hook if it uses no React hooks? (https://react.dev/reference/react/hooks)

This just feels like a utility function and therefore the nomenclature (`use*`) is a bit misleading.

src/calendar/use-calendar-grid-content.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+import React, { useMemo } from 'react';
+import styles from './styles.css.js';
+import clsx from 'clsx';
+import { normalizeStartOfWeek } from '../internal/utils/locale';
+import { getCalendarMonth } from 'mnth';
+import ScreenreaderOnly from '../internal/components/screenreader-only/index.js';
+import { renderDayName } from './utils/intl';
+
+export default function useCalendarGridContent({
```
timogasda, line 12: You're using `useMemo` here but it also doesn't really feel like a hook because you're mixing two things here.

Would it make sense to split this up into 1) a hook/utility to get rows, and 2) a pure functional component for the header ?

src/calendar/styles.scss
timogasda: Sounds good. Let's double-check that there's no regressions with a full screenshot test

timogasda (commented):
Looks great overall. Left some minor comments.

src/calendar/styles.scss
jperals: Ran the tests. There are some sub-pixel diffs in Firefox and Safari but they are stable across runs, not flaky. So they will require manual approval but only once.

Screenshot 2024-02-14 at 17 11 49

src/calendar/use-calendar-grid-content.tsx
jperals: I think this will make a bit more sense when granularity enters the scene. For month granularity there is no header, and it would be strange to have a function that returns `null` if `granularity` equals `month`. Also, note that `header` uses the first row in `rows` to generate its cell content. So there is some connection between the two, where different content for the grid is returned depending on the granularity.

In any case, we can always revisit this in #1942 soon.

src/calendar/use-calendar-grid-keyboard-navigation.ts
jperals: Went the full React way and added a `useCallback`...

just-boris: It is not required to `useCallback` to satisfy philosophy of the hooks. Here is an example, similar to yours: https://github.com/cloudscape-design/components/blob/4f7fc2c307de932b06c207931b3a76d4cf994364/src/app-layout/utils/use-keyboard-events.ts#L24

Worked well so far.

---

If you strongly dislike use-less hooks, you can make it a plain function:

```js
import {handleGridNavigation} from './handle-grid-navigation'

onGridKeyDownHandler={event => handleGridNavigation(...)}
```

just-boris: Another note, since we have `switch (event.keyCode)` in way too many places, is this the time to introduce a utility already?

```js
handleKeyNavigation(event, {
   onActivate: () => /*enter or space*/
   onUp: () => {},
   onDown: () => {},
   onLeft: () => {},
   onRight: () => {},
})
```

src/calendar/use-calendar-grid-keyboard-navigation.ts
```diff
@@ -0,0 +1,68 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import React from 'react';
+import { CalendarProps } from './interfaces';
+import { moveNextDay, movePrevDay, moveNextWeek, movePrevWeek } from './utils/navigation';
+import { KeyCode } from '../internal/keycode';
+import { isSameMonth } from 'date-fns';
```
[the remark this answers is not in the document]
jperals: Done

src/calendar/use-calendar-grid-content.tsx
pan-kot: Why not splitting the hook into `useCalendarGridRows` hook and `CalendarGridHeader` component that takes rows as a prop?

jperals: OK, split it as you suggested [here](https://github.com/cloudscape-design/components/pull/1963/commits/6a9a0dabf22fd7d5497720664143c0f00a84ffe1).

src/calendar/use-calendar-grid-keyboard-navigation.ts
jperals: Thanks both for the feedback. So I think we have two main questions here:

- Hook or no hook: since this seems to be an essentially philosophical question, and as Boris pointed out, we have other examples of this, I am inclined to keep it as a hook. And related to this question, the sub-question about `useCallback`: I could not measure any measurable performance impact of using it, and couldn't detect re-renders by not using it, so I removed it.
- A utility for keyboard navigation sounds like a good idea. I can explore it but I'd say in a different PR; then it can be integrated here.

pan-kot (approved):

PR #1993, opened 2024-02-21
feat: Emit latency metrics

connorlanigan:
### Description

This PR adds support for emitting metrics about the loading latency of the page. These metrics are emitted only on the Button, Table and Spinner components

There are four types of metrics emitted:
- when a component mounts (`mounted`)
- when a component enters a loading state (`loading-started`)
- when a component exits a loading state (`loading-finished`)
- when a component unmounts while it is still in a loading state (`loading-cancelled`)

- Document: `bfdPAfbGRNH6`

### How has this been tested?
Added unit tests. Also manually tested in dev pages.

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/internal/hooks/use-latency-metrics/is-in-viewport.ts
```diff
@@ -0,0 +1,84 @@
...
+      observe: () => {},
+      unobserve: () => {},
+    };
+  }
+  return new IntersectionObserver(callback);
+}
+
+const observer = createIntersectionObserver(function isInViewportObserver(entries) {
+  // This avoids interference when the IntersectionObserver is called again during the delay.
+  const mapSnapshot = map;
```
avinashbot, line 61: You can probably just keep the same map and reuse it by calling `.clear()`

src/internal/hooks/use-latency-metrics/index.ts
```diff
@@ -0,0 +1,125 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { useEffect, useRef } from 'react';
+import { useIdFallback } from '../use-unique-id';
+import { useEffectOnUpdate } from '../use-effect-on-update';
+import { isInViewport } from './is-in-viewport';
+import { metrics } from '../../analytics/metrics';
+
+export function useLatencyMetrics(
```
avinashbot, line 10: Seems like five parameters is pushing it - maybe it's time to create a params object?

src/internal/hooks/use-latency-metrics/index.ts
```diff
@@ -0,0 +1,125 @@
...
+            inViewport: undefined,
+            metadata: { instanceId },
+            loading: isLoadingOrSpinner,
+            loadingDuration,
+          },
+          Date.now()
+        );
+      }
+    };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
avinashbot, line 68: I definitely need an explainer comment for this, or ideally not have this at all... we really shouldn't be reaching for this hack as often as we do. If we don't want unrelated properties from causing the effect to be re-run, let's enforce that - by putting it in a ref or a state that never updates.

Effects will send double-events in dev (not that it matters, maybe), but the discrepancy will also cause headaches. Ideally, if we want to react to mount/unmount events, a callback ref is the ideal way to do it. But I know that's easier said than done, so I push for it more than just strongly suggesting it :P

src/internal/hooks/use-latency-metrics/is-in-viewport.ts
```diff
@@ -0,0 +1,84 @@
...
+  // Cleanup
+  return () => {
+    clearTimeout(timeoutHandle);
+    mapSnapshot.delete(element);
+    observer.unobserve(element);
+  };
+}
+
+function createIntersectionObserver(callback: IntersectionObserverCallback) {
+  if (typeof IntersectionObserver === 'undefined') {
```
avinashbot, line 50: Where is this the case? JSDOM?

src/internal/hooks/use-latency-metrics/is-in-viewport.ts
```diff
@@ -0,0 +1,84 @@
...
+    }
+  }, MAX_DELAY_MS);
+});
+
+function defer(callback: () => void, maxDelayMs: number) {
+  if (typeof requestIdleCallback === 'function') {
+    requestIdleCallback(callback, { timeout: maxDelayMs });
+  } else {
+    // requestIdleCallback is not supported in Safari
+    setTimeout(callback, maxDelayMs);
```
avinashbot, line 82: This might as well be `setTimeout(callback)` - the browser will just run it at the end of the event loop. It's not like waiting 1 second is going to help.

connorlanigan: Hm, I'm not sure I understand. Wouldn't the 1 second delay still allow the browser to do more important tasks first, e.g. fetching/processing data? Especially when the page first loads, the metrics code would otherwise be in competition with all other tasks and React effects that run after the page has been downloaded.

src/internal/hooks/use-latency-metrics/index.ts
connorlanigan: I agree, I'll try to improve it with ref objects.

I'm hesitant to use a _callback_ ref, since it seems semantically different from when an Effect runs. For example, if/when React uses [its proposed Offscreen rendering](https://react.dev/blog/2023/03/22/react-labs-what-we-have-been-working-on-march-2023#offscreen-rendering), the lifecycle of Effects could be separate from the one of callback refs.

avinashbot (approved):
One optional comment about setTimeout. But I also don't really think it makes sense to worry about performance for something this simple, so it's not a major thing.

PR #2000, opened 2024-02-22
feat: Table expandable rows

pan-kot:
### Description

New API to support expandable table rows.

Related changes in collection-hooks: https://github.com/cloudscape-design/collection-hooks/pull/64

Rel: caPkAgYq1LHW, UdbqAd69rbSm

Screenshot 2024-03-14 at 09 34 10

### How has this been tested?

New unit- tests, a sample integration test, screenshot tests, dry-run.

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/table/__tests__/expandable-rows.test.tsx
katiegeorge, line 1: General comment - once we decide on the behavior we want for inline edit + expandable rows, we should add a test here for that case.

src/table/body-cell/styles.scss
```diff
@@ -26,11 +26,33 @@ $edit-button-padding-right: calc(
...
+  inset-block: 0;
+  display: flex;
+  align-items: center;
+  margin-inline-start: calc(-1 * (#{awsui.$space-m} + #{awsui.$space-xs}));
+}
+
+@mixin cell-offset($padding) {
+  padding-inline-start: $padding;
+
+  @for $i from 1 through 9 {
```
katiegeorge, line 40: Do we have a screenshot test for this "extreme nesting" case of 10+?

src/table/__tests__/expandable-rows.test.tsx
pan-kot: Thanks, will ensure that! Currently there are permutations to ensure the alternative design for the first column. If disabling the editing instead, I will secure that with a unit test.

src/table/body-cell/styles.scss
pan-kot: Yes, the extreme nesting case is the first permutation.

src/table/expandable-rows/styles.scss
```diff
@@ -0,0 +1,35 @@
+/*
+ Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ SPDX-License-Identifier: Apache-2.0
+*/
+
+@use '../../internal/styles/index' as styles;
+@use '../../internal/styles/tokens' as awsui;
+
+.expand-toggle-hidden.expand-toggle-hidden {
```
jperals, line 9: Why two times the same class name for the same element?

src/table/interfaces.tsx
```diff
@@ -324,8 +326,19 @@ export interface TableProps<T = any> extends BaseComponentProps {
   /**
    * Use this property to activate advanced keyboard navigation and focusing behaviors.
    * When set to `true`, table cells become navigable with arrow keys, and the entire table has a single tab stop.
+   *
+   * By default, the keyboard navigation is active for tables with expandable rows.
    */
   enableKeyboardNavigation?: boolean;
+
+  /**
+   * Use this property to provide define expandable table rows. The expandableRows configuration object consists of
```
jperals, line 335: Double verb: "provide define".

Shouldn't we add a colon at the end of the sentence?

src/table/expandable-rows/styles.scss
pan-kot: That is needed for specificity. The same class is used below to assign other styles and it takes precedence. An alternative could be to assign both `expand-toggle` and `expand-toggle-hidden`. However, the `expand-toggle` is also used in test-utils and it must not match the hidden toggles.

src/table/interfaces.tsx
pan-kot: Thanks, updated 👍

src/table/body-cell/styles.scss
jperals: Looks like there is no indication at all of further nesting past the 9th level. Would it be possible to hard-disallow more than 9 levels instead?

pan-kot: It is possible, but we decided not to. Sometimes, it is the end user who decides the shape of the data and we don't want that data to become inaccessible. The recommendation though will be to avoid having deep nesting and add mechanisms to prevent deep nesting on the consumer end.

src/table/expandable-rows/expand-toggle-button.tsx
```diff
@@ -0,0 +1,46 @@
...
+  return (
+    <button
+      ref={buttonRef}
+      tabIndex={tabIndex}
+      aria-label={isExpanded ? collapseButtonLabel : expandButtonLabel}
+      aria-expanded={isExpanded}
+      className={styles['expand-toggle']}
+      onClick={onExpandableItemToggle}
+    >
+      {isExpanded ? (
```
jperals, line 31: In expandable section we turn the icon 90 degrees with a small animation, instead of replacing it: https://github.com/cloudscape-design/components/blob/main/src/expandable-section/motion.scss#L18

Do we want to do the same here?

pan-kot: @katiegeorge, what do you think?

src/table/body-cell/styles.scss
```diff
@@ -26,11 +26,33 @@ $edit-button-padding-right: calc(
...
+    &.body-cell-expandable-level-#{$i} {
+      padding-inline-start: calc($padding + $i * (#{awsui.$space-m} + #{awsui.$space-xs}));
+    }
+  }
+  &.body-cell-expandable-level-next {
+    padding-inline-start: calc($padding + 9 * (#{awsui.$space-m} + #{awsui.$space-xs}));
+  }
```
cansuaa, line 47: Users can nest more than 9th level but the padding for those will be the same as 9th? We are not stopping them from nesting this deep but also not supporting the styling, isn't it contradicting?

src/table/__tests__/expandable-rows.test.tsx
```diff
@@ -0,0 +1,272 @@
...
+    });
+
+    table.findExpandToggle(1)!.click();
+    expect(onExpandableItemToggle).toBeCalledTimes(1);
+    expect(onExpandableItemToggle).toBeCalledWith(
+      expect.objectContaining({ detail: { item: simpleItems[0], expanded: false } })
+    );
+
+    table.findExpandToggle(3)!.click();
+    expect(onExpandableItemToggle).toBeCalledTimes(2);
```
cansuaa, line 217: Why is it called twice while expanding? Or is it because the function was called before this as well?

src/table/body-cell/styles.scss
pan-kot: The deep nesting will be restricted with the guidelines. However, throwing an error or not showing the data after certain level is risky because sometimes the nesting can be hard to control/restrict on the consumer end. We still want consumers to do so, but if they do not succeed we prefer the data to remain accessible.

src/table/__tests__/expandable-rows.test.tsx
pan-kot: The event counter is cumulative. For each `.click()` issued the asserted count needs to be increased.

src/table/expandable-rows/expand-toggle-button.tsx
pan-kot: I added animation for the expand toggle. No animation for the content though.

src/__a11y__/a11y-page-object.ts
```diff
@@ -92,11 +92,22 @@ export default class A11yPageObject extends BasePageObject {
...
     await expect(this.getUndefinedTexts()).resolves.toHaveLength(0);
   }
 }
+
+// The message says: "aria-level values greater than 6 are not supported in all screenreader and browser combinations".
+// However, that is relevant for heading roles but not for treegrid. In treegrid there can be more levels of nesting.
+function ariaLevelViolationsFilter(violation: Axe.Result) {
+  return !(
+    violation.id === 'aria-valid-attr-value' &&
+    violation.nodes.every(node => node.all.every(entry => entry.id === 'aria-level'))
```
jperals, line 111: Do we lose coverage with this? Can we distinguish between heading levels and treegrid levels?

src/table/expandable-rows/styles.scss
jperals: Can't the test utils exclude the hidden class instead? Something like

```
this.findNativeTable().find(`tbody tr:nth-child(${rowIndex}) .${expandableRowsStyles['expand-toggle']}:not(${expandableRowsStyles['expand-toggle-hidden']})`)
```

pan-kot: Updated!

src/table/expandable-rows/expand-toggle-button.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+import React, { useRef } from 'react';
+import { useSingleTabStopNavigation } from '../../internal/context/single-tab-stop-navigation-context';
+import styles from './styles.css.js';
+import InternalIcon from '../../icon/internal';
+import { ExpandableItemProps } from './expandable-rows-utils';
+import clsx from 'clsx';
+
+export function ExpandToggle({ isExpandable, ...props }: ExpandableItemProps) {
+  return isExpandable ? <ExpandToggleButton {...props} /> : <ExpandTogglePlaceholder />;
```
jperals, line 12: Actually why do we render the ExpandTogglePlaceholder in case the row is not expandable, instead of just nothing?

pan-kot: Great question. It was previously needed to keep the consistent offsets, but now that is actually handled with CSS of the body-cell. Will update.

jperals (approved):

PR #2001, opened 2024-02-22
fix: Fix click interactions of dropdown components inside portals

youdz:
### Description

Dropdown and Autosuggest had a bug were they only looked at `element.contains()` on the containers to identify clicks and focus/blur events outside of the dropdown. This broke when the some of the contents of a dropdown were rendered through a portal themselves, which specifically happened for the [PropertyFilter with custom forms](https://cloudscape.design/components/property-filter?tabId=playground&example=with-custom-form) when the custom form contained an informational popover or a Select.

Nested dropdowns are clearly not recommended by Cloudscape, but this seems like a general enough issue to want to address as it seems to be recurring (recently in Charts).

In order to fix this, I simply leveraged the existing `nodeBelongs()` utility which was used to fix the mentioned charts issue, using it as a 1:1 replacement for `element.contains()` in Dropdown and Autosuggest.

Related links, issue #, if available: None

### How has this been tested?

1. Tested locally by adding a custom form with Select to the local dev PropertyFilter page. Tested with screenreader, keyboard-only and mouse interactions.
2. Added 5 unit tests and confirmed the entire `unit` suite passed.
3. Confirmed the entire `integ` suite passed with no modifications to tests.

I updated one of the PropertyFilter dev pages to include one custom form with a Select. You can try it yourself to ensure click and focus behaviors are correct.
cloudscape-2001

Once again, I know this pattern isn't recommended, but it's the most straightforward way to test portals in Dropdown.

<details>

#### Correctness

- _Changes are backward-compatible if not indicated, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#public-apis)._ ✅
- _Changes do not include unsupported browser features, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#browsers-support)._ ✅
- _Changes were manually tested for accessibility, see [accessibility guidelines](https://cloudscape.design/foundation/core-principles/accessibility/)._ ✅

#### Security

- _If the code handles URLs: all URLs are validated through [the `checkSafeUrl` function](https://github.com/cloudscape-design/components/blob/main/src/internal/utils/check-safe-url.ts)._ N/A

#### Testing

- _Changes are covered with new/existing unit tests?_ ✅
- _Changes are covered with new/existing integration tests?_ N/A
</details>

---

### DIFF
--- a/src/internal/components/dropdown/__tests__/dropdown.test.tsx
+++ b/src/internal/components/dropdown/__tests__/dropdown.test.tsx
@@ -62,6 +62,21 @@ describe('Dropdown Component', () => {
       expect(handleCloseDropdown).toBeCalled();
     });
 
+    test('does not fire close event when a portaled element inside dropdown is clicked', async () => {
+      const handleCloseDropdown = jest.fn();
+      renderDropdown(
+        <Dropdown trigger={<button />} onDropdownClose={handleCloseDropdown} open={true}>
+          <Dropdown trigger={<button />} open={true} expandToViewport={true}>
+            <button data-testid="inside">inside</button>
+          </Dropdown>
+        </Dropdown>
+      );
+      await runPendingEvents();
+
+      act(() => screen.getByTestId('inside').click());
+      expect(handleCloseDropdown).not.toBeCalled();
+    });
+
     test('does not fire close event when a self-destructible element inside dropdown was clicked', async () => {
       function SelfDestructible() {
         const [visible, setVisible] = useState(true);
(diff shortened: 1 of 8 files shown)


src/internal/components/dropdown/__tests__/dropdown.test.tsx
```diff
@@ -88,6 +103,44 @@ describe('Dropdown Component', () => {
       expect(screen.getByTestId('after-dismiss')).toBeTruthy();
     });
   });
+
+  describe('dropdown focus events', () => {
+    test('fires focus and blur events when focus transitions from and to an outside element', async () => {
```
youdz, line 108: This unit test was completely missing, so I added it in order to test both inside and outside cases.

src/internal/utils/node-belongs.ts
```diff
@@ -17,8 +17,12 @@ export function nodeBelongs(container: Node | null, target: Node | EventTarget |
   }
   const portal = findUpUntil(
     target as HTMLElement,
-    node => node instanceof HTMLElement && !!node.dataset.awsuiReferrerId
+    node => node === container || (node instanceof HTMLElement && !!node.dataset.awsuiReferrerId)
   );
+  if (portal === container) {
```
timogasda, line 22: Feels unlikely, but we should probably check here that we actually found a portal (`portal && portal === container`). Otherwise there could be a weird edge case where `nodeBelongs` returns true for `container==null` when the target is not inside a portal.

package.json
```diff
@@ -10,8 +10,8 @@
   "scripts": {
     "quick-build": "gulp quick-build",
     "build": "NODE_ENV=production gulp build",
-    "test": "gulp test",
-    "test:unit": "gulp test:unit",
+    "test": "TZ=UTC gulp test",
+    "test:unit": "TZ=UTC gulp test:unit",
```
timogasda, line 14: This looks unrelated. I remember the timezone being an issue for some tests a long time ago. I wonder if this could be better resolved by configuring the jest environment.

Either way, let's leave it out of this PR because it's unrelated.

timogasda (changes requested):

package.json
youdz: Ah right, this was something I asked early on in public channels and was told to just add it along with the change.

But I can confirm several date-related tests break if you do no run them on a machine in UTC. Unfortunately jest config doesn't have a good way of doing this (I remember facing the same problem multiple times), so the two typical options are to either set it in the script like this, or make sure to use the jest's `useFakeTimers` and `setSystemTime` in each test that checks dates.

Removed in the new revision. 👍

timogasda (approved):

timogasda:
Further internal testing showed no regression. Merging this now.

PR #2027, opened 2024-02-29
feat: Add test API to retrieve option groups in Select and Multiselect

jperals:
### Description

Ticket: `AWSUI-34239`

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/select/__tests__/test-utils.test.tsx
timogasda, line 1: I'm curious why you decided to create a test suite to test the new test utils instead of just using them in the existing unit tests (or adding a couple of new ones)?

src/internal/components/options-list/__tests__/test-indexes.test.ts
```diff
@@ -33,25 +33,25 @@ describe('Test indexes', () => {
       { groupIndex: 2, inGroupIndex: 1, throughIndex: 4 },
     ]);
   });
-  test('skips "use-entered" option and group headers', () => {
+  test('skips "use-entered" option but not group headers', () => {
```
timogasda, line 36: Do we think that this could be somehow a breaking change in the test util behavior?

src/select/__tests__/test-utils.test.tsx
jperals: My thinking is that testing utils is a public API, so they should have their own specific tests that make it hard for us to break these utils. Otherwise the tests could be refactored or some of them even removed as the code evolves, and nothing would strongly prevent these utils from breaking.

I see that this is not something we usually do, but maybe we should...?

src/internal/components/options-list/__tests__/test-indexes.test.ts
jperals: I don't think so. Note that this test covers an internal util and not something directly customer facing, and it is a pretty much intentional change in order to achieve this new test util functionality. What `test-indexes.ts` does is generate the attributes necessary for each list item so that they can be found by the test utils. Selectable options get a `data-test-index` attribute which indicates their 1-based position in the list (parent groups excluded), and if they are children of a group, also `data-group-index` and `data-child-index` to indicate the index of the parent group and their own index in the group. What this PR changes is that also the parent groups get an attribute, `data-group-index` (with the same value as their children), so that they can be found by the test utils as well.

The existing test util methods to find options all use `data-test-index` or `data-child-index`, which only non-parents have, so they will not find any parent groups.

The existing test utils are used in tests (e.g, `findOption` 10 times and `findOptions` 58 times), but we can always add specific tests for them if we want (related to the comment above).

timogasda: Alright, sounds safe to me. I would still consider doing a dry-run just in case.

timogasda (approved):

src/internal/components/options-list/__tests__/test-indexes.test.ts
jperals: Just ran the build successfully (`6767344096`). Will merge :)

PR #2037, opened 2024-03-05
add token-group focus handling prototype

johannes-weber:
### Description

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/internal/components/token-list/token-focus-controller.tsx
```diff
@@ -8,34 +8,27 @@ import styles from './styles.css.js';
...
 
     const tokenElements = tokenListRef.current.querySelectorAll(tokenSelector);
     const toggleButton = tokenListRef.current.querySelector(toggleButtonSelector);
 
-    const activeItemIndices: number[] = [];
-    for (let i = 0; i < tokenElements.length; i++) {
-      if (!tokenElements[i].querySelector('[aria-disabled="true"]')) {
-        activeItemIndices.push(i);
-      }
-    }
```
johannes-weber, line 27: With this change, disabled token's should not be skipped when finding the next focusable token.

avinashbot (commented):
Tested it with VoiceOver and didn't find any regressions. Did find a very specific case where focus got lost though, but it's not new (just keep removing the last item over and over again). Not enough of a concern to block this improvement though.

Using useEffect for focus management is rough, and ideally we'd do it in the handlers (focus should only move on direct interaction anyway).

avinashbot (approved):

PR #2031, opened 2024-03-02
chore: Upgrade lockfiles to version 3

just-boris:
### Description

1. Run `npm install --lockfile-version 3` to convert lockfiles to the latest version
2. Updated unlocker script and removed a duplicate inside bundle size workflow

Related links, issue #, if available: n/a

### How has this been tested?

PR build

<details>
   <summary>Review checklist</summary>

_The following items are to be evaluated by the author(s) and the reviewer(s)._

#### Correctness

- _Changes include appropriate documentation updates._
- _Changes are backward-compatible if not indicated, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#public-apis)._
- _Changes do not include unsupported browser features, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#browsers-support)._
- _Changes were manually tested for accessibility, see [accessibility guidelines](https://cloudscape.design/foundation/core-principles/accessibility/)._

#### Security

- _If the code handles URLs: all URLs are validated through [the `checkSafeUrl` function](https://github.com/cloudscape-design/components/blob/main/src/internal/utils/check-safe-url.ts)._

#### Testing

- _Changes are covered with new/existing unit tests?_
- _Changes are covered with new/existing integration tests?_
</details>

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2042, opened 2024-03-06
chore: Package versioning in bundle size workflow

connorlanigan:
### Description

This CR makes the bundle size workflow more independent from Cloudscape versions. It removes Cloudscape from the package.json file as a dependency, so that the current Cloudscape package in NPM cannot affect the version that this workflow installs.

### How has this been tested?
Ran the commands from the workflow locally.

<details>

#### Correctness

#### Security

#### Testing

</details>

---


.github/workflows/bundle-size/package.json
```diff
@@ -7,11 +7,10 @@
...
-    "@vitejs/plugin-react": "^4.0.0",
-    "react": "^18.2.0",
-    "react-dom": "^18.2.0",
-    "typescript": "^5.1.3",
-    "vite": "^4.5.2"
+    "@vitejs/plugin-react": "^4",
+    "react": "^18",
+    "react-dom": "^18",
+    "typescript": "^5",
+    "vite": "^5"
```
connorlanigan, line 14: I also upgraded from Vite 4 to Vite 5

.github/workflows/bundle-size/package-lock.json
```diff
@@ -1,1536 +0,0 @@
-{
```
just-boris, line 1: Don't we need this lockfile to pin build tools versions (Vite)?

Also, since https://github.com/cloudscape-design/components/pull/2031, Cloudscape packages are gone from this package-lock file

.github/workflows/bundle-size/package.json
just-boris: Good doing an upgrade, but how about keeping the default npm behavior here? It adds full version number here.

I know it does not mean much because the real version is in the lock file, but having to manually format the file after installing a new module is annoying

just-boris (commented):
Looks good overall, how about opening a dummy PR on top of this to test it in real environment? Or do you want to do it after the merge?

.github/workflows/bundle-size/package.json
connorlanigan: Hm, I find it more readable to only see the major version, since it more clearly communicates what we inted to install, e.g. Vite 5, React 18, etc. If you add another dependency with e.g. `npm install node-fetch` it does not modify the other entries, so there is no manual maintenance required

just-boris: You still will have to chase down inconsistencies on every code review. We already got similar story with `*` versions in some other cases, it did not scale super well.

Not a blocking comment anyway, just sharing some experience.

connorlanigan:
> Looks good overall, how about opening a dummy PR on top of this to test it in real environment? Or do you want to do it after the merge?

Good idea. I've now created a draft PR on top of this, the workflow seems to work correctly: https://github.com/cloudscape-design/components/pull/2044. It reports an increase of 10 KB (compressed), the new file is 32 KB.

.github/workflows/bundle-size/package-lock.json
connorlanigan: With the lockfile, the build unfortunately fails: https://github.com/cloudscape-design/components/actions/runs/8171937882/job/22341231084#step:6:16
According to the Rollup log output, this appears to be a bug in NPM, where the lockfile apparently needs to be created on the same platform and architecture as the consumer of the lockfile

just-boris:
Thanks. I would also check if the decrease can be detected, just to be extra sure

.github/workflows/bundle-size/package-lock.json
just-boris: Coincidentally, I am currently fixing the same build error in codeview: https://github.com/cloudscape-design/code-view/pull/22

It was easily fixable by doing a clean install.

Having lockfile is a must here. You do not want to block the entire team PR process, when an expected change in Vite or Typesctipt happens and bundle size step breaks and blocks PRs from merging

connorlanigan: Ah, indeed - I had to regenerate the lockfile locally two times but now it works 😄

PR #2049, opened 2024-03-08
chore: Replace usage of shell commands with node apis

michaeldowseza:
### Description

- Replace usage of exec cp with fs
- Replace usage of exec rsync wish fs-extra
- Add cross-env to handle env variables in npm scripts
- Add npm-run-all to replace && in npm scripts

### Why
Using shell commands makes our packages OS specific. 

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


build-tools/tasks/themeable-source.js
```diff
@@ -2,14 +2,15 @@
 // SPDX-License-Identifier: Apache-2.0
 const { parallel } = require('gulp');
 const path = require('path');
-const { promises: fsp } = require('fs');
+const { promises: fsp, cpSync } = require('fs');
+const fse = require('fs-extra');
```
just-boris, line 6: I do not see this dependency in our `package.json`

just-boris (commented):
ok for replacing plain shell commands with node API, but not sure we can market it as "OS support", because if we do not have a CI on that platform it does not count

michaeldowseza:
> ok for replacing plain shell commands with node API, but not sure we can market it as "OS support", because if we do not have a CI on that platform it does not count

I will rename. I guess it only indirectly improves the support.

build-tools/tasks/themeable-source.js
michaeldowseza: Hmm, the dependency was available to me. It must be coming in as a transitive. Will add to package json to make it explicit.

just-boris (approved):

PR #2056, opened 2024-03-12
feat: Slider component

katiegeorge:
### Description

Slider component. 
image

At its simplest, the component looks something like this:
```
<Slider min={0} max={100} value={40} />
```

It also accepts optional tick marks and reference labels. It can be either continuous or stepped.

See `slider/basic` and `slider/permutations` for test pages.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


connorlanigan:
There seems to be a slight difference in alignment between the start and end position of the slider: when the slider is at its lowest value, the blue circle's _left edge_ is aligned with the edge of the track - but when the slider is at its highest value, the blue circle's _center_ is aligned with the edge of the track, which overall makes the slider extend a bit further to the right than to the left (highlighted with red lines in the screenshot below). Is this expected? Can we make the alignment on both ends the same (whether edge-aligned or center-aligned)
alignment

connorlanigan:
When the slider is used on a touch screen, the default touch interaction (scrolling) still happens at the same time, which can make it harder to see and control what you're configuring. It also diverges from the native behaviour of sliders on touch devices (at least Android and as far as I know also iOS). Can we disable scrolling while the slider is being dragged?



https://github.com/cloudscape-design/components/assets/9086585/545798fb-faf4-4f67-89a0-214bbe7832eb

src/slider/utils.ts
```diff
@@ -0,0 +1,24 @@
...
+};
+
+export const findLowerAndHigherVals = (array: Array<number>, value: number) => {
+  array.sort((a, b) => a - b);
+
+  const index = array.indexOf(value);
+  const lower = array[index - 1];
+  const higher = array[index + 1];
+
+  return { lower, higher };
```
pan-kot, line 21: When using index access like `array[index - 1]` TypeScript (with our settings) assumes the value is always present, yet it might be undefined. You can add the return type to the function `{ lower: undefined | number, higher: undefined | number }` to express the real type. Otherwise, the consumer code might rely on the values being always defined.

src/slider/__tests__/utils.test.tsx
```diff
@@ -0,0 +1,33 @@
...
+  test('works with indivisible step', () => {
+    expect(getStepArray(6, 0, 20)).toEqual([0, 6, 12, 18]);
+  });
+
+  test('works with negative numbers', () => {
+    expect(getStepArray(5, -10, 10)).toEqual([-10, -5, 0, 5, 10]);
+  });
+});
+
+describe('findLowerAndHigherVals', () => {
```
pan-kot, line 25: I would add two more cases: when there is no lower value and when there is no higher value.

src/slider/index.tsx
```diff
@@ -0,0 +1,177 @@
...
+  const tooltipPositionAdjust =
+    hasValue && (value <= min ? THUMB_SIZE - THUMB_SIZE / 2 : value >= max ? THUMB_SIZE + THUMB_SIZE / 2 : THUMB_SIZE);
+
+  useEffect(() => {
+    if (rangeRef.current) {
+      rangeRef.current.style.width = `${percent}%`;
+    }
+  }, [percent]);
+
+  const popoverContent = hasValue && (
```
pan-kot, line 59: nit: This is not really a popover content but the popover itself. I would rename the variable to e.g. popoverElement

src/slider/index.tsx
```diff
@@ -0,0 +1,177 @@
...
+    if (value !== undefined && valueFormatter && valueFormatter(value)) {
+      return valueFormatter(value);
+    }
+
+    if (value && referenceValues && valueFormatter && !valueFormatter(value)) {
+      const valueArray = [min, ...referenceValues, value, max];
+      const prev = valueFormatter(findLowerAndHigherVals(valueArray, value)?.lower);
+      const next = valueFormatter(findLowerAndHigherVals(valueArray, value)?.higher);
+
+      return `${value}, between ${prev} and ${next}`;
```
pan-kot, line 73: This has to use an i18n property e.g. some `i18nStrings.valueTextRange(before, value, after)`. Also, how does this work when there is no lower or no higher value?

src/slider/index.tsx
```diff
@@ -0,0 +1,177 @@
...
+        <div
+          className={clsx(styles['slider-track'], {
+            [styles.disabled]: disabled,
+          })}
+        />
+        {step && tickMarks && (
+          <div
+            className={clsx(styles.ticks)}
+            style={{
+              [customCssProps.sliderLabelCount]: (max - min) / step,
```
pan-kot, line 99: CSS can probably handle that but I would round the result anyways. The count value must be an integer logically.

src/slider/interfaces.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  /**
+   * Indicates the minimum value.
+   */
+  min: number;
+
+  /**
+   * Indicates the maximum value.
+   */
+  max: number;
```
pan-kot, line 21: What happens to the component when min is greater than max?

src/slider/interfaces.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  /**
+   * Formats the values. This will format both the labels and the tooltip.
+   */
+  valueFormatter?: (value: number) => string;
+
+  /**
+   * onChange handler.
+   */
+  onChange?: NonCancelableEventHandler<SliderProps.ChangeDetail>;
```
pan-kot, line 31: Usually we annotate handlers with something like:

```
/**
  * Called when the user selects a value.
  * The event `detail` contains the current `value`.
  */
```

src/slider/interfaces.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  /**
+   * How big the step size is.
+   */
+  step?: number;
```
pan-kot, line 36: What happens to the component if step is greater than (max - min)?

src/slider/interfaces.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  /**
+   * Whether or not the slider is disabled.
+   */
+  disabled?: boolean;
+
+  /**
+   * Labels shown between the minimum and maximum values.
+   */
+  referenceValues?: Array<number>;
```
pan-kot, line 56: It is better to accept `ReadonlyArray<number>` to guarantee that we are not going to update it.

Do we expect the reference values array to be sorted? Can it include values outside of the min/max range?

src/slider/slider-labels.tsx
```diff
@@ -0,0 +1,107 @@
...
+        values.push(referenceValues[i]);
+        lastValue = referenceValues[i];
+      }
+    }
+    return values;
+  };
+
+  function getLabelPosition(index: number) {
+    const minDistance = (max - min) / LABEL_THRESHOLD;
+    const colSpan = Math.floor(minDistance / 2);
```
pan-kot, line 42: What is colSpan?

src/slider/slider-labels.tsx
```diff
@@ -0,0 +1,107 @@
...
+          [customCssProps.sliderMinEnd]: max - min > LABEL_THRESHOLD ? getLabelPosition(0).min : 1,
+        }}
+      >
+        {valueFormatter ? valueFormatter(min) : min}
+      </span>
+      {getVisibleReferenceValues().map((step, index) => {
+        return (
+          <span
+            role="listitem"
+            key={`step-${index}`}
```
pan-kot, line 79: nit: the key can be just index, there is no need to add a suffix or a prefix to it unless you need to render multiple elements with the same index

style-dictionary/visual-refresh/borders.ts
```diff
@@ -34,6 +34,7 @@ export const tokens: StyleDictionary.BordersDictionary = {
   borderRadiusItem: '8px',
   borderRadiusInput: '8px',
   borderRadiusPopover: '{borderRadiusInput}',
+  borderRadiusSliderThumb: '50%',
```
pan-kot, line 37: Do we expect this radius to ever change?

src/slider/index.tsx
timogasda, line 1: We should stick to the standard structure of having a basic `index.tsx` file for the public component interface, and then an `internal.tsx` file that contains the actual implementation. See for example: https://github.com/cloudscape-design/components/blob/main/src/form-field/index.tsx

timogasda: It's also missing the new external props hook (`getExternalProps`). I suggest you have a look at the "Creating a component" in this document: `i0b0A07hAfPB`.

src/slider/index.tsx
```diff
@@ -0,0 +1,177 @@
...
+  hideFillLine,
+  valueFormatter,
+  ...rest
+}: SliderProps) {
+  const { __internalRootRef } = useBaseComponent('Slider');
+  const baseProps = getBaseProps(rest);
+
+  const rangeRef = useRef<HTMLDivElement>(null);
+  const handleRef = useRef<HTMLDivElement>(null);
+  const [showPopover, setShowPopover] = React.useState(false);
```
timogasda, line 41: nit: We usually go for simply `useState` (without `React.`)

src/slider/index.tsx
```diff
@@ -0,0 +1,177 @@
...
+
+  const percent = hasValue && getPercent(Math.max(Math.min(value, max), min), min, max);
+
+  // at the min and max values, the thumb is not centered.
+  const tooltipPositionAdjust =
+    hasValue && (value <= min ? THUMB_SIZE - THUMB_SIZE / 2 : value >= max ? THUMB_SIZE + THUMB_SIZE / 2 : THUMB_SIZE);
+
+  useEffect(() => {
+    if (rangeRef.current) {
+      rangeRef.current.style.width = `${percent}%`;
```
timogasda, line 55: This probably doesn't need to be an effect. Can we apply this directly to the range element?

src/slider/index.tsx
```diff
@@ -0,0 +1,177 @@
...
+  const [isActive, setIsActive] = React.useState(false);
+  const labelsId = useUniqueId('labels');
+  const hasValue = value !== undefined;
+  const formFieldContext = useFormFieldContext(rest);
+
+  const percent = hasValue && getPercent(Math.max(Math.min(value, max), min), min, max);
+
+  // at the min and max values, the thumb is not centered.
+  const tooltipPositionAdjust =
+    hasValue && (value <= min ? THUMB_SIZE - THUMB_SIZE / 2 : value >= max ? THUMB_SIZE + THUMB_SIZE / 2 : THUMB_SIZE);
```
timogasda, line 51: Can you please add a bit more explanation on what's happening here?

src/slider/styles.scss
```diff
@@ -0,0 +1,260 @@
...
+
+  &:hover::-webkit-slider-thumb {
+    @include mixins.base-thumb-hover-styles();
+  }
+
+  &:hover::-moz-range-thumb {
+    @include mixins.base-thumb-hover-styles();
+  }
+
+  &:focus::-webkit-slider-thumb,
```
timogasda, line 210: Looks like we're not using the usual `focus-visible` helpers. Is this intentional?

timogasda (commented):
I just did a relatively shallow review. Great work! Please do not wait for my ship it

src/test-utils/dom/slider/index.ts
```diff
@@ -0,0 +1,21 @@
...
+
+  findNativeInput(): ElementWrapper<HTMLInputElement> {
+    return this.findByClassName(styles.thumb)!;
+  }
+
+  findRange(): ElementWrapper<HTMLDivElement> {
+    return this.findByClassName(styles['slider-range'])!;
+  }
+
+  findReferenceLabels(): ElementWrapper<HTMLSpanElement>[] {
```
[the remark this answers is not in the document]
katiegeorge: Fair point. I'll move this to the internal testing folder.

src/test-utils/dom/slider/index.ts
```diff
@@ -0,0 +1,21 @@
...
+import BaseInputWrapper from '../input/base-input.js';
+
+export default class SliderWrapper extends BaseInputWrapper {
+  static rootSelector = styles.root;
+
+  findNativeInput(): ElementWrapper<HTMLInputElement> {
+    return this.findByClassName(styles.thumb)!;
+  }
+
+  findRange(): ElementWrapper<HTMLDivElement> {
```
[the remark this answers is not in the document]
katiegeorge: I use it for our internal tests, so I figured customers might want to use it as well. For example:

```
  test('Range clamps at 100% if value is greater than max', () => {
    const wrapper = renderSlider({ min: 0, max: 100, value: 150 });

    expect(wrapper.findRange().getElement()).toHaveStyle('width: 100%');
  });
```

We can move this to the internal testing folder if we don't think customers would need it.

pages/slider/basic.page.tsx
```diff
@@ -0,0 +1,259 @@
...
+  const satisfactionScaleOptions = [
+    { value: 0, label: 'Extremely unsatisfied' },
+    { value: 1, label: 'Somewhat satisfied' },
+    { value: 2, label: `Not satisfied or unsatisfied` },
+    { value: 3, label: 'Somewhat satisfied' },
+    { value: 4, label: 'Extremely satisfied' },
+  ];
+
+  React.useEffect(() => {
+    setError2(value2 < 0 || value2 > 100 || value2 % 1 !== 0);
```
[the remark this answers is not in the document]
katiegeorge: When I take it out of the `useEffect`, the error only updates one state change after I want it. For example, the invalid state only kicks in when I update the value from 101 - 102, not from 100 - 101. Am I missing something?

pages/slider/basic.page.tsx
```diff
@@ -0,0 +1,259 @@
...
+            <Slider
+              value={value2}
+              onChange={({ detail }) => {
+                setValue2(detail.value);
+              }}
+              min={0}
+              max={100}
+              tickMarks={true}
+              step={25}
+              invalid={error2}
```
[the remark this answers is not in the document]
katiegeorge: Good catch!

pages/slider/basic.page.tsx
```diff
@@ -0,0 +1,259 @@
...
+          tickMarks={true}
+        />
+      </FormField>
+      <FormField label="Stepped slider">
+        <Slider
+          value={value3}
+          onChange={({ detail }) => {
+            setValue3(detail.value);
+          }}
+          valueFormatter={value => `${value} is the very long and precise value at this point.`}
```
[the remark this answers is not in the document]
katiegeorge: We use the tooltip when the user hovers or focuses the value, which will have the same text as the reference value.

src/slider/interfaces.ts
katiegeorge: It should not be valid. Shall we throw a warning?

src/slider/interfaces.ts
katiegeorge: It should not be valid. Shall we throw a warning?

src/slider/index.tsx
katiegeorge: true. its actually called tooltip now so i'll update it to be that.

src/slider/__tests__/utils.test.tsx
katiegeorge: Good catch.

src/slider/interfaces.ts
katiegeorge: It doesn't have to be sorted, since the calculation just takes the value itself and places it in that column along the slider. If its outside the min / max, it just isn't shown.

katiegeorge: Added a test for that, by the way.

style-dictionary/visual-refresh/borders.ts
katiegeorge: Not that I know of, but it seemed like it was good practice to make border radii tokens. Do you see a reason why not?

src/slider/slider-labels.tsx
katiegeorge: Hmm, I don't like using indices as keys. From the [React docs](https://react.dev/learn/rendering-lists#why-does-react-need-keys): 

> You might be tempted to use an item’s index in the array as its key. In fact, that’s what React will use if you don’t specify a key at all. But the order in which you render items will change over time if an item is inserted, deleted, or if the array gets reordered. Index as a key often leads to subtle and confusing bugs.

Since this is a read only array, I don't think this is possible, but still seems like good practice to not use them. I updated to just use `step`.

pages/slider/basic.page.tsx
```diff
@@ -0,0 +1,259 @@
...
+      </FormField>
+      <FormField label="How hot do you like your salsa?">
+        <Slider
+          value={value7}
+          onChange={({ detail }) => {
+            setValue7(detail.value);
+          }}
+          min={0}
+          max={100}
+          referenceValues={[7, 8, 9, 10, 25, 37, 39, 52, 53, 59, 67, 68, 93]}
```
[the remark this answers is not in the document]
katiegeorge: Happy to explain this one to you offline - went back and forth on the calculations here, but TL;DR - each label gets a set amount of space even if it can technically take up more.

pages/slider/basic.page.tsx
```diff
@@ -0,0 +1,259 @@
...
+export default function InputsPage() {
+  const [value, setValue] = React.useState(4);
+  const [temperature, setTemperature] = React.useState(72.4);
+  const [ticks, setTicks] = React.useState(false);
+  return (
+    <div style={{ padding: 30 }}>
+      <h1>Sliders demo</h1>
+      <SpaceBetween size="xxl">
+        <Sliders />
+        <Container header={<Header>This is a header</Header>}>
```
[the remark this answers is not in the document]
katiegeorge: Visually, to make sure it blends well in a form.

pages/slider/basic.page.tsx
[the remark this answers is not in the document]
pan-kot: The useEffect cannot be just removed. Instead, you need to update the validation together with value change:

```
<Slider onChange={({ detail }) => {
  setValue(detail.value);
  setError(validateValue(detail.value));
}} />
```

pages/slider/basic.page.tsx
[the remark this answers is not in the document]
pan-kot: Not sure how much is that needed. I think it is important to verify it works well with the form field and that's it. More complex integration can probably be featured in the demos.

src/test-utils/dom/slider/index.ts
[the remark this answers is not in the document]
pan-kot: Yes, I have seen how it is used in internal tests, but we should not expect customers to assert against our internal class-names or styles.

src/slider/interfaces.ts
pan-kot: I think we should use warnOnce but also provide some reasonable fallback

src/slider/interfaces.ts
pan-kot: I think a warning and some fallback would do. Not sure what can be the fallback exactly. Maybe we can renderd the component but in a disabled state or something

style-dictionary/visual-refresh/borders.ts
timogasda: more tokens = bigger CSS bundle

katiegeorge: Okay, I'll remove the token then!

src/slider/interfaces.ts
katiegeorge: I think we just do what the HTML reference does, which is a slider that cannot be moved. That plus a warnOnce makes sense to me.

src/slider/index.tsx
katiegeorge: The `getExternalProps` hook is only currently used in 7 components, I didn't see it as a requirement. The slider doesn't have an internal-only properties (indicated with the __ prefix), which is what that hook filters out. Do we still want it here?

src/slider/styles.scss
katiegeorge: The webkit-slider-thumb and moz-thumb elements are extremely finicky when it comes to pseudo elements like focus and hover. You have to define them separately, for example, or they won't work. I tried using the `focus-visible` helpers here but it broke the focus state. Is there any risk to not using it?

src/slider/slider-labels.tsx
katiegeorge: the grid column span that the label will take up. this is to give the label the most space it can have.

src/slider/index.tsx
katiegeorge: Update: Added it, and updated to use internal vs. index. Still unclear if it should be a requirement for components without internal props.

src/slider/internal.tsx
```diff
@@ -0,0 +1,188 @@
...
+import { SliderProps } from './interfaces.js';
+import useBaseComponent from '../internal/hooks/use-base-component/index.js';
+import { fireNonCancelableEvent } from '../internal/events/index.js';
+import { useFormFieldContext } from '../internal/context/form-field-context.js';
+import { useUniqueId } from '../internal/hooks/use-unique-id/index.js';
+import Tooltip from '../internal/components/tooltip/index.js';
+import SliderLabels from './slider-labels.js';
+import { getPercent, getStepArray, findLowerAndHigherValues, valuesAreValid, THUMB_SIZE } from './utils.js';
+import customCssProps from '../internal/generated/custom-css-properties/index.js';
+import { warnOnce } from '@cloudscape-design/component-toolkit/internal';
```
timogasda, line 17: nit: move this up to the other external dependencies. (I wish we had an eslint plugin for import order...)

src/slider/internal.tsx
```diff
@@ -0,0 +1,188 @@
...
+  step,
+  disabled,
+  ariaLabel,
+  referenceValues,
+  tickMarks,
+  hideFillLine,
+  valueFormatter,
+  ...rest
+}: SliderProps) {
+  const { __internalRootRef } = useBaseComponent('Slider');
```
timogasda, line 33: Don't use `useBaseComponent` again after using it in `internal.tsx`. You can grab `__internalRootRef` from your incoming props.

src/slider/styles.scss
timogasda: The main purpose of the utility is being a kind of polyfill for `:focus-visible`, so it makes sure that we consistently show focus outlines when we detect a keyboard user. Right now you're kinda taking care of that using mouse and keyboard events.

The risk here would be that the behavior ends up being inconsistent in some edge cases.

timogasda: Slightly off-topic because it's design feedback but: I noticed that we're using different color shades to indicate focused state. Is there a reason why we're not using the standard focus outline color around the knob?

pages/slider/basic.page.tsx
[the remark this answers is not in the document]
pan-kot: Actually, there is one corner case. The tooltip is only shown for the selected value. In a readonly Slider with long labels those labels cannot be read in full.

src/slider/utils.ts
```diff
@@ -0,0 +1,40 @@
...
+  sortedArray = [...sortedArray.sort((a, b) => a - b)];
+
+  const index = sortedArray.indexOf(value) || 0;
+  const lower = sortedArray[index - 1] || undefined;
+  const higher = sortedArray[index + 1] || undefined;
+
+  return { lower, higher };
+};
+
+export const valuesAreValid = (referenceValues: ReadonlyArray<number> | undefined) => {
```
pan-kot, line 30: nit: the `valuesAreValid` is always called with defined `referenceValues`. You can update the type and remove the unnecessary IF condition.

src/slider/internal.tsx
```diff
@@ -0,0 +1,188 @@
...
+      const valueArray = [min, ...referenceValues, value, max];
+      const prevAndNext = findLowerAndHigherValues(valueArray, value);
+      const prev = prevAndNext.lower && valueFormatter(prevAndNext.lower);
+      const next = prevAndNext.higher && valueFormatter(prevAndNext.higher);
+
+      if (prev === undefined) {
+        return `${value}, between ${valueFormatter(min)} and ${next}`;
+      }
+      if (next === undefined) {
+        return `${value}, between ${prev} and ${valueFormatter(max)}`;
```
pan-kot, line 83: The hardcoded labels need to be replaced with i18n and tested with I18nProvider.

pan-kot (changes requested):

src/slider/interfaces.ts
```diff
@@ -0,0 +1,71 @@
...
+   */
+  step?: number;
+
+  /**
+   * Hides the colored fill line, so only the handle is visible.
+   */
+  hideFillLine?: boolean;
+
+  /**
+   * Show the tick marks along the axis. Use with stepped sliders, except in extreme cases.
```
pan-kot, line 45: not: along which axis? I'd say the "slider line" or something along the lines

src/slider/interfaces.ts
```diff
@@ -0,0 +1,71 @@
...
+   */
+  tickMarks?: boolean;
+
+  /**
+   * Whether or not the slider is disabled.
+   */
+  disabled?: boolean;
```
pan-kot, line 52: Should we support readOnly? See sixtAeSZSxjR

src/slider/internal.tsx
```diff
@@ -0,0 +1,188 @@
...
+  const { __internalRootRef } = useBaseComponent('Slider');
+  const baseProps = getBaseProps(rest);
+
+  const handleRef = useRef<HTMLDivElement>(null);
+  const [showTooltip, setShowTooltip] = useState(false);
+  const [isActive, setIsActive] = useState(false);
+  const labelsId = useUniqueId('labels');
+  const hasValue = value !== undefined;
+  const formFieldContext = useFormFieldContext(rest);
+  const { ariaLabelledby, ariaDescribedby, controlId, invalid } = formFieldContext ?? rest;
```
pan-kot, line 42: Can formFieldContext be undefined?

src/slider/internal.tsx
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    if (value && referenceValues && valueFormatter && !valueFormatter(value)) {
+      const valueArray = [min, ...referenceValues, value, max];
+      const prevAndNext = findLowerAndHigherValues(valueArray, value);
+      const prev = prevAndNext.lower && valueFormatter(prevAndNext.lower);
+      const next = prevAndNext.higher && valueFormatter(prevAndNext.higher);
+
+      if (prev === undefined) {
+        return `${value}, between ${valueFormatter(min)} and ${next}`;
```
pan-kot, line 80: What if the value equals to min or max? Do we want to adjust the text in that case?

src/slider/internal.tsx
```diff
@@ -0,0 +1,188 @@
...
+            })}
+            style={{ [customCssProps.sliderRangeInlineSize]: `${percent}%` }}
+          />
+        )}
+      </div>
+
+      <input
+        aria-label={ariaLabel}
+        aria-labelledby={ariaLabel && !rest.ariaLabelledby ? undefined : ariaLabelledby}
+        aria-describedby={labelsId && valueFormatter ? labelsId : ariaDescribedby}
```
pan-kot, line 144: labelsId is always defined, no need to check for it

pan-kot: when both the formatter and the ariaDescribedBy are defined, should we join them?

src/slider/internal.tsx
```diff
@@ -0,0 +1,188 @@
...
+              [styles.disabled]: disabled,
+            })}
+            style={{ [customCssProps.sliderRangeInlineSize]: `${percent}%` }}
+          />
+        )}
+      </div>
+
+      <input
+        aria-label={ariaLabel}
+        aria-labelledby={ariaLabel && !rest.ariaLabelledby ? undefined : ariaLabelledby}
```
pan-kot, line 143: why do we check for `!rest.ariaLabelledby` here?

src/slider/slider-labels.tsx
```diff
@@ -0,0 +1,108 @@
...
+export interface SliderLabelProps {
+  min: number;
+  max: number;
+  referenceValues?: ReadonlyArray<number>;
+  valueFormatter?: (value: number) => string;
+  labelsId: string;
+}
+export default function SliderLabels({ min, max, referenceValues, valueFormatter, labelsId }: SliderLabelProps) {
+  const isMobile = useMobile();
+  const LABEL_THRESHOLD = isMobile ? 5 : 10;
```
pan-kot, line 19: Does LABEL_THRESHOLD represent the max number of labels to show? If so, would it make sense to rename the variable to e.g. `VISIBLE_LABELS_COUNT`?

src/slider/slider-labels.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+  const getVisibleReferenceValues = () => {
+    if (!referenceValues || referenceValues.length === 0) {
+      return [];
+    }
+
+    const values: Array<number> = [];
+    let lastValue = 0;
+    for (let i = 0; i <= referenceValues.length; i++) {
```
pan-kot, line 28: Do we expect the array to be sorted?

src/slider/utils.ts
```diff
@@ -0,0 +1,40 @@
...
+  }
+  return steps;
+};
+
+export const findLowerAndHigherValues = <T extends number>(
+  array: ReadonlyArray<T>,
+  value: T
+): { lower: undefined | T; higher: undefined | T } => {
+  let sortedArray = [...array];
+  sortedArray = [...sortedArray.sort((a, b) => a - b)];
```
pan-kot, line 21: nit: the array is copied on the prev line, no need to destructure it again

```
let sortedArray = [...array];
sortedArray.sort((a, b) => a - b);
```
or
```
const sortedArray = [...array].sort((a, b) => a - b);
```

src/slider/internal.tsx
katiegeorge: Yes, I have this on my to-do list. Thanks for the reminder.

src/slider/interfaces.ts
katiegeorge: Good question. If we should, I want to say let's keep that separate and do that alongside the rest of the readOnly updates.

src/slider/slider-labels.tsx
katiegeorge: No, it shouldn't matter.

src/slider/internal.tsx
katiegeorge: I assumed it was undefined when a Slider was used outside of a form field. So, use the form field properties when inside a form field, and use from props when not.

src/slider/internal.tsx
katiegeorge: It is impossible for the min & max not to be labeled, with how the API currently works. So that won't happen.

src/slider/styles.scss
katiegeorge: This came directly from my sessions with visual design. We determined together that we liked the look of this better. Feedback noted though, we can keep an eye on it going forward.

src/slider/internal.tsx
katiegeorge: Same behavior as Input. I'll add a comment.
```
    // aria-labelledby has precedence over aria-label in accessible name calculation.
    // When aria-label is provided for Input, it should override aria-labelledBy from form-field context.
    // If both aria-label and aria-labelledby come from Input props, aria-labelledby will be used in accessible name
```

src/slider/internal.tsx
katiegeorge: Good catch! Yes!

src/slider/internal.tsx
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    if (value && referenceValues && valueFormatter && !valueFormatter(value)) {
+      const valueArray = [min, ...referenceValues, value, max];
+      const prevAndNext = findLowerAndHigherValues(valueArray, value);
+      const prev = prevAndNext.lower ? valueFormatter(prevAndNext.lower) : valueFormatter(min);
+      const next = prevAndNext.higher ? valueFormatter(prevAndNext.higher) : valueFormatter(max);
+
+      //return `${value}, between ${prev} and ${next}`;
+      return i18n('i18nStrings.valueTextRange', i18nStrings?.valueTextRange(prev, value, next));
```
avinashbot, line 86: Since valueTextRange takes arguments, it needs to be called a little differently:

```suggestion
      return i18n(
        'i18nStrings.valueTextRange',
        i18nStrings?.valueTextRange(prev, value, next),
        (format) => format({ value, prev, next })
      );
```

src/slider/internal.tsx
pan-kot: The context is always defined but when used outside of the form field the individual properties will all be default: undefined or false.

src/slider/slider-labels.tsx
pan-kot: why not?

pan-kot (approved):

pages/slider/basic.page.tsx
```diff
@@ -0,0 +1,178 @@
...
+        onChange={({ detail }) => {
+          setValue(detail.value);
+        }}
+        i18nStrings={i18nStrings}
+      />
+      <FormField
+        label="Slider with input and validation"
+        errorText={
+          error && value2
+            ? value2 % 1 !== 0
```
pan-kot, line 61: nit: this bit does not scale well. Can we make the type of the error to be "string"? Then you can set the specific error upon validation.

src/i18n/messages/all.de.json
```diff
@@ -262,6 +262,9 @@
     "selectedAriaLabel": "Ausgewählt",
     "recoveryText": "Wiederholen"
   },
+  "slider": {
+    "i18nStrings.valueTextRange": "{value}, zwischen {previousValue} und {nextValue}"
```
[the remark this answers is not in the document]
katiegeorge: Resolved in quip.

pages/slider/basic.page.tsx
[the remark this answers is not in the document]
katiegeorge: Decided not to truncate the values anymore, so this shouldn't be a problem!

src/slider/slider-labels.tsx
katiegeorge: Because each value is assigned a column in the grid layout, and that's what determines the order. Added a test for that [here](https://github.com/cloudscape-design/components/pull/2056/files#diff-acb4f1937faef0d28137ce656c4de3e363d2215585461f1fc671869bb3f04367R115).

src/table/__tests__/a11y.test.tsx
```diff
@@ -180,5 +180,28 @@ describe('labels', () => {
         `Displaying items from ${firstIndex} to ${lastIndex} of ${totalItemsCount} items`
       );
     });
+
+    test('The table items live region must not include nested items', () => {
```
pan-kot, line 184: This change is unrelated, looks identical to the test before - likely caused by the merge resolution.

pan-kot (approved):
Great work!

src/slider/internal.tsx
```diff
@@ -61,19 +82,24 @@ export default function InternalSlider({
...
+  if (step && hasValue && (value - min) % step !== 0) {
+    warnOnce('Slider', 'Slider value must be a multiple of the step. The value will round to the nearest step value.');
+  }
+
+  const tooltip = <Tooltip value={valueFormatter ? valueFormatter(getValue()) : getValue()} trackRef={handleRef} />;
 
   const getAriaValueText = () => {
-    if (value !== undefined && valueFormatter && valueFormatter(value)) {
-      return valueFormatter(value);
+    if (valueFormatter && valueFormatter(getValue())) {
```
pan-kot, line 95: Instead of calling the getValue() multiple times can we compute the normalised value once and then use it?

src/slider/styles.scss
```diff
@@ -184,9 +199,12 @@
   appearance: none;
   position: absolute;
   block-size: 0;
-  inline-size: calc(100% + mixins.$thumb-size);
-  outline: 0;
+  // inline-size: calc(100% + mixins.$thumb-size);
+  // margin-inline-start: calc(mixins.$thumb-size / -2);
+  inline-size: calc(100% + 16px);
```
pan-kot, line 204: Is this supposed to be hardcoded to 16px?

katiegeorge: Updated to use variable.

pan-kot (approved):

PR #2068, opened 2024-03-14
fix: hide popover on line charts when relatedTarget is not a DOM node

johannes-weber:
### Description

This PR addresses the issue where an exception was thrown when leaving the popover on a chart by either:
- Focusing another element outside the page (all browsers)
- Right-clicking on a popover (Firefox only)

The root cause was that `event.relatedTarget` pointed to the `window` object in these cases. The `useMouseHover` hook has been updated to account for this scenario.

Related links, issue #, if available: AWSUI-35614

---

Why FF acts differently?
Firefox is firing the `onmousleave` event on a right click, all other browsers don't. Try it on the [mouseleave MDN page's Examples](https://developer.mozilla.org/en-US/docs/Web/API/Element/mouseleave_event#examples). The [onmouseleave specs](https://w3c.github.io/uievents/#event-type-mouseleave) mentions
> A [user agent](https://w3c.github.io/uievents/#user-agent) MUST also dispatch this event when the element or one of its descendants moves to be no longer underneath the primary pointing device.

Not sure how the context menu is handled in this case, it could be interpreted as "no longer underneath the primary pointing device". I'm searching for references and existing FF bug reports, in case I won't find any, I'll create one.

### How has this been tested?

- Verified the fix manually in Firefox, Chrome, and Safari.
- Added unit tests to cover the `useMouseHover` hook's edge cases.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/area-chart/__tests__/area-chart-use-chart-model.test.tsx
+++ b/src/area-chart/__tests__/area-chart-use-chart-model.test.tsx
@@ -303,6 +303,47 @@ describe('useChartModel', () => {
 
         expect(wrapper.findHighlightedX()?.getElement()).toHaveTextContent('');
       });
+
+      test('clear highlighted X when mouse exited from the page', () => {
+        const { wrapper } = renderChartModelHook({
+          height: 0,
+          highlightedSeries: null,
+          setHighlightedSeries: (_series: AreaChartProps.Series<ChartDataTypes> | null) => _series,
+          setVisibleSeries: (_series: readonly AreaChartProps.Series<ChartDataTypes>[]) => _series,
+          width: 0,
+          xDomain: undefined,
+          xScaleType: 'linear',
+          yScaleType: 'linear',
+          externalSeries: series,
+          visibleSeries: series,
+          popoverRef: { current: null },
+        });
+
+        const mouseMoveEvent = {
+          relatedTarget: wrapper.findPlot()?.getElement(),
+          clientX: 100,
+          clientY: 100,
+        } as any;
+
+        act(() => {
+          fireEvent.mouseMove(wrapper.findPlot()!.getElement(), mouseMoveEvent);
+        });
+
+        expect(wrapper.findHighlightedX()?.getElement()).toHaveTextContent('1');
+
+        const mouseLeaveEvent = {
+          relatedTarget: window, // when mouse exited the page, relatedTarget is set to window.
+          clientX: 0,
+          clientY: 0,
+        } as any;
+
+        act(() => {
+          fireEvent.mouseLeave(wrapper.findDetailPopover()!.getElement(), mouseLeaveEvent);
+        });
+
+        expect(wrapper.findHighlightedX()?.getElement()).toHaveTextContent('');
+      });
+
       test('keep highlighted X when mouse leaves popover but in plot', () => {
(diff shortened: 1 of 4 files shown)


avinashbot (commented):
Do we also need changes here:
https://github.com/cloudscape-design/components/blob/main/src/area-chart/model/use-chart-model.ts#L322

avinashbot (approved):
Looks good at first glance, assuming all tests pass :)

src/mixed-line-bar-chart/hooks/use-mouse-hover.ts
```diff
@@ -134,7 +134,11 @@ export function useMouseHover<T>({
   };
 
   const onPopoverLeave = (event: React.MouseEvent) => {
-    if (!isHandlersDisabled && !plotRef.current!.svg.contains(event.relatedTarget as Node)) {
+    if (
```
just-boris, line 137: How about using `nodeContains` utility, similar to this PR: https://github.com/cloudscape-design/components/pull/1374

And avoid fixing the same issue over and over again?

avinashbot: I didn't know we still had this utility - that would be a much better fix for this, yeah

just-boris: It moved to the component-toolkit, but still available and should be used to mitigate all such issues

src/area-chart/model/use-chart-model.ts
```diff
@@ -319,11 +319,7 @@ export default function useChartModel<T extends AreaChartProps.DataTypes>({
     };
 
     const onPopoverLeave = (event: MouseEvent) => {
-      // pointing device exited from the page or from an element within the svg
-      if (
-        (event.relatedTarget !== window && plotRef.current!.svg.contains(event.relatedTarget as Node)) ||
-        interactions.get().isPopoverPinned
-      ) {
+      if (nodeContains(plotRef.current!.svg, event.relatedTarget as Element) || interactions.get().isPopoverPinned) {
```
just-boris, line 322: I think this type-casting is not needed. I specifically updated the utility signature to accept `relatedTarget`: https://github.com/cloudscape-design/component-toolkit/blob/fa9186592934151b720c12070239ec4ded7c34a9/src/dom/node-contains.ts#L9

But thanks for using this utility anyway, I can approve this now

just-boris (approved):

src/area-chart/__tests__/area-chart-use-chart-model.test.tsx
```diff
@@ -303,6 +303,47 @@ describe('useChartModel', () => {
...
+
+        expect(wrapper.findHighlightedX()?.getElement()).toHaveTextContent('1');
+
+        const mouseLeaveEvent = {
+          relatedTarget: window, // when mouse exited the page, relatedTarget is set to window.
+          clientX: 0,
+          clientY: 0,
+        } as any;
+
+        act(() => {
```
just-boris, line 340: `fireEvent` already contains `act` wrapper inside, so this is redundant

just-boris (dismissed):

just-boris (approved):

PR #2078, opened 2024-03-19
chore: Add RTL support for small components

DaemonCahill:
This PR adds RTL support for a list of components categorized as having a small implementation effort.

### DIFF
--- a/src/link/styles.scss
+++ b/src/link/styles.scss
@@ -54,6 +54,11 @@
       @include styles.link-font-size-style(map.get(constants.$link-font-sizes, $font-size));
     }
   }
+
+  @include styles.with-direction('rtl') {
+    align-items: center;
+    display: inline-flex;
+  }
 }
 
 .icon-wrapper {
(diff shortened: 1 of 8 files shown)


src/link/styles.scss
```diff
@@ -54,6 +54,11 @@
       @include styles.link-font-size-style(map.get(constants.$link-font-sizes, $font-size));
     }
   }
+
+  @include styles.with-direction('rtl') {
+    align-items: center;
+    display: inline-flex;
+  }
```
just-boris, line 61: Why is this rtl only? I do not see any direction related behavior in these lines

just-boris: Ok to ship this, but let's do a follow up because it is clearly a debt

just-boris (approved):

src/link/styles.scss
DaemonCahill: Agreed!

PR #2082, opened 2024-03-21
fix: Table inline actions in sticky column to cause no horizontal scroll

pan-kot:
### Description

There is an issue reproducible in Chrome for table sticky columns: when an interactive element inside a sticky column receives focus the table container scrolls so that the column becomes in view as no longer sticky.

Scenarios:
* A button inside a sticky column is clicked - no issue here;
   - no Fix needed.
* Any focusable element inside a sticky column is navigated to with Tab / Shift + Tab - the scroll issue happens;
   - Not fixed. The keyboard focus is taken care of by the browser so no easy way to fix and no strong justification either.
* A button dropdown option is selected - the scroll issue happens;
   - Fixed by preventing scroll when moving focus back to the button dropdown trigger.
* In table with keyboard navigation a cell inside a sticky column is clicked - the scroll issue happens;
   - Fixed by preventing scroll when moving focus inside a cell programmatically.
* A table with inline editing inside a sticky column is activated - the scroll issue happens.
   - Not fixed. No straightforward way to fix because the focusing normally happens due to auto-focus set on the inline editing form input. No strong justification to fix.

https://github.com/cloudscape-design/components/assets/20790937/b423ae8a-95eb-4037-af7c-d0a29c115f3c

Rel: AWSUI-37794

### How has this been tested?

* New integration test

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/table/table-role/grid-navigation.tsx
+++ b/src/table/table-role/grid-navigation.tsx
@@ -127,12 +127,9 @@ class GridNavigationProcessor {
       changeHandler(newIsFocusable);
     }
     // When newly registered element belongs to the focused cell the focus must transition to it.
-    if (
-      this.focusedCell &&
-      this.focusedCell.element.tagName === 'TD' &&
-      this.focusedCell.element.contains(focusableElement)
-    ) {
-      focusableElement.focus();
+    if (this.focusedCell?.element.tagName === 'TD' && this.focusedCell?.element.contains(focusableElement)) {
+      // Scroll is unnecessary when moving focus from a cell to element within the cell.
+      focusableElement.focus({ preventScroll: true });
     }
     return () => this.unregisterFocusable(focusableElement);
   };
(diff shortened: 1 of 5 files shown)


src/table/table-role/grid-navigation.tsx
```diff
@@ -167,10 +164,11 @@ class GridNavigationProcessor {
 
     // Focusing on cell is not eligible when it contains focusable elements in the content.
     // If content focusables are available - move the focus to the first one.
-    const cellElement = getClosestCell(this.focusedCell.element);
-    const nextTarget = this.focusedCell.element === cellElement ? this.getFocusablesFrom(cellElement)[0] : null;
+    const focusedElement = this.focusedCell.element;
```
pan-kot, line 167: Minor refactoring to avoid unnecessary getClosestCell lookup.

src/table/table-role/utils.ts
```diff
@@ -20,7 +20,7 @@ export function defaultIsSuppressed(target: Element) {
   let current: null | Element = target;
   while (current) {
     // Stop checking for parents upon reaching the cell element as the function only aims at the cell content.
-    if (current instanceof HTMLTableCellElement) {
+    if (current.tagName === 'TD') {
```
pan-kot, line 23: Applied suggestion from another PR: https://github.com/cloudscape-design/components/pull/2118#discussion_r1551818397

gethinwebster (approved):

PR #2090, opened 2024-03-25
chore: Add RTL support for medium components.

DaemonCahill:
This PR adds RTL support for a list of components categorized as having a medium implementation effort.

### DIFF
--- a/pages/top-navigation/logos/long-logo.svg
+++ b/pages/top-navigation/logos/long-logo.svg
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<svg height="31px" viewBox="0 0 401 31" version="1.1" xmlns="http://www.w3.org/2000/svg">
+<svg height="31px" width="401px" viewBox="0 0 401 31" version="1.1" xmlns="http://www.w3.org/2000/svg">
     <g stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
         <g transform="translate(-31.000000, -12.000000)">
             <g transform="translate(31.000000, 12.000000)">
--- a/pages/top-navigation/logos/long-logo.svg
+++ b/pages/top-navigation/logos/long-logo.svg
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<svg width="443px" height="31px" viewBox="0 0 443 31" version="1.1" xmlns="http://www.w3.org/2000/svg">
+<svg height="31px" viewBox="0 0 401 31" version="1.1" xmlns="http://www.w3.org/2000/svg">
     <g stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
         <g transform="translate(-31.000000, -12.000000)">
             <g transform="translate(31.000000, 12.000000)">
--- a/src/date-range-picker/__tests__/date-range-picker-absolute.test.tsx
+++ b/src/date-range-picker/__tests__/date-range-picker-absolute.test.tsx
@@ -8,6 +8,7 @@ import { i18nStrings } from './i18n-strings';
 import { isValidRange } from './is-valid-range';
 import '../../__a11y__/to-validate-a11y';
 import TestI18nProvider from '../../../lib/components/i18n/testing';
+import { KeyCode } from '../../../lib/components/internal/keycode';
 
 const defaultProps: DateRangePickerProps = {
   locale: 'en-US',
@@ -168,6 +169,21 @@ describe('Date range picker', () => {
         expect(wrapper.findDropdown()!.findStartTimeInput()!.findNativeInput().getElement()).toHaveValue('00:00:00');
       });
 
+      test('start date can be selected with keyboard', () => {
+        const { wrapper } = renderDateRangePicker({
+          ...defaultProps,
+          value: { type: 'absolute', startDate: '2021-03-02T05:00:00+08:45', endDate: '2021-03-12T13:05:21+08:45' },
+        });
+        wrapper.findTrigger().click();
+
+        wrapper.findDropdown()!.findDateAt('left', 3, 4).focus();
+        wrapper.findDropdown()!.findDateAt('left', 3, 4).keydown(KeyCode.enter);
+
+        expect(wrapper.findDropdown()!.findSelectedStartDate()!.getElement()).toHaveTextContent('17');
+        expect(wrapper.findDropdown()!.findStartDateInput()!.findNativeInput().getElement()).toHaveValue('2021/03/17');
+        expect(wrapper.findDropdown()!.findStartTimeInput()!.findNativeInput().getElement()).toHaveValue('00:00:00');
+      });
+
       test('end date can be selected', () => {
         const { wrapper } = renderDateRangePicker({
           ...defaultProps,
@@ -177,6 +193,22 @@ describe('Date range picker', () => {
 
         wrapper.findDropdown()!.findDateAt('left', 2, 3).click();
         wrapper.findDropdown()!.findDateAt('left', 3, 4).click();
+
+        expect(wrapper.findDropdown()!.findSelectedEndDate()!.getElement()).toHaveTextContent('17');
+        expect(wrapper.findDropdown()!.findEndDateInput()!.findNativeInput().getElement()).toHaveValue('2021/03/17');
+        expect(wrapper.findDropdown()!.findEndTimeInput()!.findNativeInput().getElement()).toHaveValue('23:59:59');
+      });
+
+      test('end date can be selected with keyboard', () => {
+        const { wrapper } = renderDateRangePicker({
+          ...defaultProps,
+          value: { type: 'absolute', startDate: '2021-03-02T05:00:00+08:45', endDate: '2021-03-12T13:05:21+08:45' },
+        });
+        wrapper.findTrigger().click();
+
+        wrapper.findDropdown()!.findDateAt('left', 2, 3).click();
+        wrapper.findDropdown()!.findDateAt('left', 3, 4).focus();
+        wrapper.findDropdown()!.findDateAt('left', 3, 4).keydown(KeyCode.space);
 
         expect(wrapper.findDropdown()!.findSelectedEndDate()!.getElement()).toHaveTextContent('17');
         expect(wrapper.findDropdown()!.findEndDateInput()!.findNativeInput().getElement()).toHaveValue('2021/03/17');
--- a/src/date-range-picker/calendar/grids/index.tsx
+++ b/src/date-range-picker/calendar/grids/index.tsx
@@ -13,6 +13,7 @@ import { hasValue } from '../../../internal/utils/has-value';
 import { useDateCache } from '../../../internal/hooks/use-date-cache';
 import { moveNextDay, movePrevDay, moveNextWeek, movePrevWeek, getBaseDay } from '../../../calendar/utils/navigation';
 import { findDateToFocus } from '../utils';
+import handleKeyDown from '../../../internal/utils/handle-key-down';
 
 function isVisible(date: Date, baseDate: Date, isSingleGrid: boolean) {
   if (isSingleGrid) {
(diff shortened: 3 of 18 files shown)


src/internal/utils/handle-key-down.ts
```diff
@@ -0,0 +1,68 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import React from 'react';
+import { KeyCode } from '../keycode';
+
+export default function handleKeyDown({
```
jperals, line 7: Does the writing direction affect all possible usages of this abstraction? For example, there can be cases where we just want to move the focus one element to the left or to the right, regardless of the writing direction, so handlers like `onLeft` and `onRight` might make sense too. Maybe it makes sense to start only with the handlers that we need, and keep adding the ones that we see necessary. I wouldn't mind mixing logical and directional handlers if we see that both have their sensible use cases.

DaemonCahill: Can you elaborate on circumstances where we would want to go left in the UI for a component in both LTR and RTL writing modes?

jperals: Actually if we manage logical handlers, not sure it is a "keyboard handler" anymore, but some other kind of abstraction. For example, `handleDirection` and it could be built on top of this one?

src/internal/direction.ts
```diff
@@ -0,0 +1,12 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+export function isRtl(element: HTMLElement | EventTarget) {
+  let isRtl = false;
+
+  if (element instanceof HTMLElement) {
+    isRtl = getComputedStyle(element).direction === 'rtl';
+  }
+
+  return isRtl;
```
just-boris, line 11: Are there any possible issues with incorrect default value on global hotkeys (if event target is `window`), which always assumes no-RTL?

src/internal/utils/handle-key-down.ts
```diff
@@ -0,0 +1,63 @@
...
+  onPageDown?: (event: React.KeyboardEvent) => void;
+  onPageUp?: (event: React.KeyboardEvent) => void;
+}) {
+  return function (event: React.KeyboardEvent) {
+    switch (event.keyCode) {
+      case KeyCode.down:
+        onBlockEnd?.(event);
+        break;
+      case KeyCode.end:
+        onEnd?.(event);
```
just-boris, line 35: Are you saying that unlike arrow keys, <kbd>Home</kbd> and <kbd>End</kbd> keys do flip their direction in RTL?

src/internal/utils/handle-key-down.ts
```diff
@@ -0,0 +1,63 @@
...
+      case KeyCode.pageDown:
+        onPageDown?.(event);
+        break;
+      case KeyCode.pageUp:
+        onPageUp?.(event);
+        break;
+      case KeyCode.right:
+        isRtl(event.currentTarget) ? onInlineStart?.(event) : onInlineEnd?.(event);
+        break;
+      case KeyCode.space:
```
just-boris, line 55: How about using `case` stacking instead of repeating things?

```js
case KeyCode.enter:
case KeyCode.space:
  onActivate?.(event);
  break;
```

This is especially confusing when you have some other handlers between these two

src/segmented-control/__tests__/desktop.test.tsx
```diff
@@ -1,14 +1,15 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 import React from 'react';
+import { KeyCode } from '@cloudscape-design/test-utils-core/dist/utils';
```
just-boris, line 4: This should be `@cloudscape-design/test-utils-core/utils`

```suggestion
import { KeyCode } from '@cloudscape-design/test-utils-core/utils';
```

src/segmented-control/__tests__/desktop.test.tsx
```diff
@@ -141,4 +142,18 @@ describe('selected property', () => {
...
   });
+
+  test('does change focus when left/right arrow key is pressed', () => {
+    const { segmentedControlWrapper } = renderSegmentedControl(
+      <SegmentedControl selectedId="seg-1" options={defaultOptions} />
+    );
+
+    getSegmentWrapper(segmentedControlWrapper, 0).getElement().focus();
+    getSegmentWrapper(segmentedControlWrapper, 0).keydown(KeyCode.right);
+    expect(document.activeElement).toHaveAttribute('aria-label', 'Icon for Segment-2');
```
just-boris, line 153: This assertion should be 

```js
expect(getSegmentWrapper(segmentedControlWrapper, 1)).toHaveFocus()
```

And similar for all other assertions here

src/tabs/scroll-utils.ts
```diff
@@ -1,46 +1,57 @@
...
-  const paginationButtonsWidth = leftOverflowButton.current && 2 * leftOverflowButton.current.offsetWidth;
+  const paginationButtonsWidth = inlineStartOverflowButton.current && 2 * inlineStartOverflowButton.current.offsetWidth;
   return paginationButtonsWidth ? scrollWidth > offsetWidth + paginationButtonsWidth : scrollWidth > offsetWidth;
 };
 
-export const hasLeftOverflow = (headerBar: HTMLElement): boolean => {
-  return headerBar.scrollLeft > 0;
+export const hasInlineStartOverflow = (headerBar: HTMLElement): boolean => {
+  // scrollLeft will be a negative number if the direction is RTL
+  const scrollInlineStart = isRtl(headerBar) ? headerBar.scrollLeft * -1 : headerBar.scrollLeft;
```
just-boris, line 43: Do you think it is okay to repeat this multiple time instead of extracting one more function

```js
const scrollInlineStart = getScrollInlineStart(headerBar)
```

which we can reuse?

src/date-range-picker/calendar/grids/index.tsx
```diff
@@ -85,39 +86,27 @@ export const Grids = ({
...
-        return;
+    event.preventDefault();
+
+    handleKeyDown({
+      onActivate: () => focusedDate && onSelectDate(focusedDate),
+      onBlockEnd: () => focusedDate && (updatedFocusDate = moveNextWeek(focusedDate, isDateEnabled)),
+      onBlockStart: () => focusedDate && (updatedFocusDate = movePrevWeek(focusedDate, isDateEnabled)),
+      onInlineEnd: () => focusedDate && (updatedFocusDate = moveNextDay(focusedDate, isDateEnabled)),
+      onInlineStart: () => focusedDate && (updatedFocusDate = movePrevDay(focusedDate, isDateEnabled)),
+    })(event);
```
just-boris, line 106: Not sure about this `)(event)` syntax. Usually people do not read the trailing part and this is where bugs can hide.

This is more error proof

```js
handleKey(error, {
  onActivate,
   ...etc
});
```

Also the code inside has nothing to do with `down` event, so let's call it just `key` so we can reuse it in keyup and keypress

pages/top-navigation/logos/long-logo.svg
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<svg width="443px" height="31px" viewBox="0 0 443 31" version="1.1" xmlns="http://www.w3.org/2000/svg">
+<svg height="31px" viewBox="0 0 401 31" version="1.1" xmlns="http://www.w3.org/2000/svg">
```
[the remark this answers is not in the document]
DaemonCahill: Screenshot 2024-04-03 at 4 10 34 PM
Screenshot 2024-04-03 at 4 10 25 PM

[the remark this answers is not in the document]
DaemonCahill: The width and viewport definitions are larger than the content which creates a misalignment from the inline start position in RTL. When I reduced the size to match the content both directions are aligned correctly. All tests passed so I'm assuming this was not intentional? Let me know if you have any ideas on a better solution.

src/date-range-picker/calendar/grids/index.tsx
```diff
@@ -85,39 +86,27 @@ export const Grids = ({
     }
   }, [baseDate, focusedDate, isSingleGrid, isDateEnabled, onFocusedDateChange]);
 
-  const onGridKeyDownHandler = (e: React.KeyboardEvent) => {
+  const onGridKeyDownHandler = (event: React.KeyboardEvent) => {
```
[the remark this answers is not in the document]
DaemonCahill: Likewise!

src/internal/direction.ts
DaemonCahill: That's an interesting question. Off the top of my head, any callers of this function should be a deliberate request for the computed style of a specified element (and never the window). I'm actually updating the function to only take an Element because that satisfies a standard element, event.currentTarget, and various SVG node types. I checked it out and this out and Typescript will scream if it gets a window object.

pages/top-navigation/logos/long-logo.svg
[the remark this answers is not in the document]
just-boris: I see. Having viewbox and rect width agree is a good idea I think. 

But we still need to keep explicit width declaration. I think it is important.


> All tests passed

This is not how it works. if you change something in the source code and tests are still passing you are safe. But same does not apply to changes in tests

src/internal/utils/handle-key-down.ts
DaemonCahill: My understanding is that the left and right keys should move physically, but the home and end keys (MacOS fn + left/right) should move logically. This means that End should move left in an RTL horizontal overflow even though you are pressing the right key. I am double checking with our resident native Arabic speakers.

src/internal/utils/handle-key-down.ts
DaemonCahill: Agreed!

src/internal/utils/handle-key-down.ts
just-boris: Make sense for me, but can we have a formal confirmation of this somewhere to be extra sure about our decision?

src/segmented-control/__tests__/desktop.test.tsx
DaemonCahill: Good catch, updated.

src/segmented-control/__tests__/desktop.test.tsx
DaemonCahill: Good catch, updated.

src/tabs/scroll-utils.ts
DaemonCahill: I was on the fence about this myself but you pushed me over the line. Adding a new function to the direction file.

src/date-range-picker/calendar/grids/index.tsx
DaemonCahill: I updated the function to `handleKey` across all instances. The `(event`) syntax is effectively calling the `handleKey` function on the same line as the declaration. The reason it is not called directly is because it is nested inside a keydown handler that is trigged by the `onKeyDown` of the component which contains additional logic. I tried the above but that doesn't actually call the `handleKey` with the event. Is there a more appropriate syntax?

DaemonCahill:
```
handleKey({ ... })(event);
let myHandleKey = handleKey({ ... });
myHandleKey(event);
```
To my knowledge, these are the same?

src/internal/direction.ts
DaemonCahill: If I type it as `HTMLElement` then when I have `SVGCircleElement` the function will fail to return RTL as true. Any suggestions here?

pages/top-navigation/logos/long-logo.svg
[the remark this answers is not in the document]
DaemonCahill: Sounds good, added it back.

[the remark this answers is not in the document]
jperals: The explicit width is still missing in the narrow logo.

src/internal/direction.ts
```diff
@@ -0,0 +1,18 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+export function isRtl(element: Element) {
+  let isRtl = false;
+
+  if (element instanceof Element) {
```
jperals, line 7: This `instanceof` check is not really necessary if the type is already declared as `Element` in the function definition, right?

src/internal/direction.ts
DaemonCahill: I'll set it back to `HTMLElement` for the scope of this PR so we can get it merged. When I handle the popover and charts we can continue the conversation on typing as I get a more granular picture of what we need to support.

src/date-range-picker/__tests__/date-range-picker-absolute.test.tsx
```diff
@@ -168,6 +169,21 @@ describe('Date range picker', () => {
...
       });
 
+      test('start date can be selected with keyboard', () => {
+        const { wrapper } = renderDateRangePicker({
+          ...defaultProps,
+          value: { type: 'absolute', startDate: '2021-03-02T05:00:00+08:45', endDate: '2021-03-12T13:05:21+08:45' },
+        });
+        wrapper.findTrigger().click();
+
+        wrapper.findDropdown()!.findDateAt('left', 3, 4).focus();
```
jperals, line 179: Is this line necessary or would `keydown` work all the same without focusing first?

src/internal/direction.ts
DaemonCahill: I need to leave it because I am temporarily setting the type back to `HTMLElement | EventTarget`. Also, there were some tests that broken when I didn't run this check. We can finalize this in the follow up PR for popover.

src/date-range-picker/__tests__/date-range-picker-absolute.test.tsx
DaemonCahill: You need to focus first!

src/internal/direction.ts
jperals: OK, makes sense to me to inly support what is needed at the moment.

src/internal/direction.ts
```diff
@@ -0,0 +1,27 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * Determine if the direction of a given element is left-to-right
+ * or right-to-left. The is the JavaScript equivalent of the
+ * with-direction mixin that implements the dir() pseudoclass.
+ */
+export function isRtl(element: HTMLElement | EventTarget) {
+  let isRtl = false;
```
jperals, line 10: Nitpick: this could be simplified a little bit without the need of the `let` declaration:
```
if (element instanceof HTMLElement) {
  return getComputedStyle(element).direction === 'rtl';
}

return false;
```

jperals (dismissed):

src/internal/direction.ts
just-boris: Unresolving this. @DaemonCahill you said you will set it back to `HTMLElement` but this did not happen

src/internal/direction.ts
```diff
@@ -0,0 +1,27 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * Determine if the direction of a given element is left-to-right
+ * or right-to-left. The is the JavaScript equivalent of the
+ * with-direction mixin that implements the dir() pseudoclass.
+ */
```
just-boris, line 8: I spotted a trend that I had to remove many such comments when I maintain our code.

This function is short and self explanatory. But extra comment adds a liability to the team members to not forget to update this comment if the implementation changes.

Do not add extra liability to the team

just-boris (commented):
All good, except `src/internal/direction.ts` file which needs updates

jperals (approved):

PR #2097, opened 2024-03-27
chore: Add table expandable rows to metrics

pan-kot:
### Description

A follow-up to https://github.com/cloudscape-design/components/pull/2000

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/table/index.tsx
+++ b/src/table/index.tsx
@@ -23,6 +23,7 @@ const Table = React.forwardRef(
         variant,
         wrapLines: props.wrapLines,
         enableKeyboardNavigation: props.enableKeyboardNavigation,
+        expandableRows: !!props.expandableRows,
       },
     });
 


src/table/index.tsx
```diff
@@ -23,6 +23,7 @@ const Table = React.forwardRef(
         variant,
         wrapLines: props.wrapLines,
         enableKeyboardNavigation: props.enableKeyboardNavigation,
+        expandableRows: !!props.expandableRows,
```
johannes-weber, line 26: Is there a reason why we cast this one to a boolean while we hand the other data points as we get them?

pan-kot: The expandable rows feature is active when the expandableRows property is defined. All four properties of expandable rows configuration are required and do not contribute to metrics, but we can extend it in the future if needed, for example:
```
expandableRows: !!props.expandableRows,
expandableRowsProgressiveLoading: !!props.expandableRows?.getItemState
```

johannes-weber (approved):

src/table/index.tsx
just-boris: Very good point in this discussion. We can provide extra metadata, but not in this object. This object is only for vanilla props and I reverted this change: https://github.com/cloudscape-design/components/pull/2102

For custom metadata, you need to update `component-toolkit` to support `meta` field along with the `props`.

@pan-kot will you do that?

PR #2107, opened 2024-04-02
feat: Announce form field errors using ARIA live region

cansuaa:
### Description

Added aria live to form field error messages as part of validation improvements. These changes allow error messages to be announced in form field, attribute editor, tag editor, file upload, etc (components that use FormFieldError component).

Issue: AWSDesignSystem-432

### How has this been tested?

Tested manually (to try yourself; go to pages/form/with-validation, turn on screen reader, type text longer than 4 letters to the input field and click submit)
Also added and ran unit test

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/form-field/internal.tsx
+++ b/src/form-field/internal.tsx
@@ -1,6 +1,6 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
-import React, { useEffect } from 'react';
+import React, { useEffect, useRef } from 'react';
 import clsx from 'clsx';
 
 import { getBaseProps } from '../internal/base-component';
@@ -37,6 +37,7 @@ interface FormFieldErrorProps {
 
 export function FormFieldError({ id, children, errorIconAriaLabel }: FormFieldErrorProps) {
   const i18n = useInternalI18n('form-field');
+  const contentRef = useRef<HTMLDivElement | null>(null);
 
   return (
     <>
@@ -50,12 +51,12 @@ export function FormFieldError({ id, children, errorIconAriaLabel }: FormFieldEr
             <InternalIcon name="status-warning" size="small" />
           </div>
         </div>
-        <span className={styles.error__message}>{children}</span>
+        <span className={styles.error__message} ref={contentRef}>
+          {children}
+        </span>
       </div>
-      <LiveRegion assertive={true}>
-        <span>{errorIconAriaLabel}</span>
-        <span>{children}</span>
-      </LiveRegion>
+
+      <LiveRegion assertive={true} source={[errorIconAriaLabel, contentRef]} />
     </>
   );
 }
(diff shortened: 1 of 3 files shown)


src/form-field/internal.tsx
```diff
@@ -38,18 +39,23 @@ export function FormFieldError({ id, children, errorIconAriaLabel }: FormFieldEr
...
+            <InternalIcon name="status-warning" size="small" />
+          </div>
         </div>
+        <span className={styles.error__message}>{children}</span>
       </div>
-      <span className={styles.error__message}>{children}</span>
-    </div>
+      <LiveRegion assertive={true}>
+        {errorIconAriaLabel} {children}
+      </LiveRegion>
```
michaeldowseza, line 57: Should we consider wrapping both of these in `spans` as per the comments here in the LiveRegion implementation?

https://github.com/cloudscape-design/components/blob/main/src/internal/components/live-region/index.tsx#L37-L48

michaeldowseza:
What happens when you end up with multiple errors at the same time? Would we just be flooding the screen reader? Does the error message that is read out include the FormField label?

src/form-field/internal.tsx
cansuaa: Sure, will update it.

cansuaa:
> What happens when you end up with multiple errors at the same time? Would we just be flooding the screen reader? Does the error message that is read out include the FormField label?

I tested with multiple form fields which I passed errorText to all at the same time. It only announces one of them, usually the first one. It doesn't read out the form field label

src/form-field/internal.tsx
```diff
@@ -38,18 +39,24 @@ export function FormFieldError({ id, children, errorIconAriaLabel }: FormFieldEr
...
+        <div className={styles['error-icon-shake-wrapper']}>
+          <div
+            role="img"
+            aria-label={i18n('i18nStrings.errorIconAriaLabel', errorIconAriaLabel)}
+            className={styles['error-icon-scale-wrapper']}
+          >
+            <InternalIcon name="status-warning" size="small" />
+          </div>
         </div>
+        <span className={styles.error__message}>{children}</span>
```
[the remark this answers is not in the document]
pan-kot: We can wrap the entire thing with `<LiveRegion visible={true} />`. The simple aria-live used to have a bug in Firefox.

pan-kot:
> I tested with multiple form fields which I passed errorText to all at the same time. It only announces one of them, usually the first one. It doesn't read out the form field label

Did you test it with different browsers and screen-readers? I would actually expect multiple errors to get announced. I think the reason only one get announced is because you set the live region type to "assertive". The assertive live regions tend to overshadow one another and I am not sure it is the first one that consistently wins.

cansuaa:
> > I tested with multiple form fields which I passed errorText to all at the same time. It only announces one of them, usually the first one. It doesn't read out the form field label
> 
> Did you test it with different browsers and screen-readers? I would actually expect multiple errors to get announced. I think the reason only one get announced is because you set the live region type to "assertive". The assertive live regions tend to overshadow one another and I am not sure it is the first one that consistently wins.

Not with different browser but I tried with VoiceOver and NVDA. The case you mentioned is also quite possible. I'm currently testing out different implementations and will discuss them with accessibility.

timogasda (approved):

michaeldowseza (approved):

timogasda:
Successful internal dryrun: `6813817202`

PR #2118, opened 2024-04-04
fix: Table navigation to focus element registered inside focused cell

pan-kot:
### Description

In tables with keyboard nav when a cell does not have interactive elements the cell itself is focused.
When there are interactive elements we always expect one of those to be focused instead of the cell.

The PR fixes the issue when table cell content updates dynamically so that the cell receives interactive element. When the cell is focused the focus must transition to the newly added element.

The problem was originally found as part of https://github.com/cloudscape-design/components/pull/2108 testing. When progressive loading transitions from "load more" to "loading" and back to "load more" the focus should transition between the load-more button and the text-only loading cell.

### How has this been tested?

* New unit test
* New integration test as part of https://github.com/cloudscape-design/components/pull/2108

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/context/__tests__/utils.tsx
+++ b/src/internal/context/__tests__/utils.tsx
@@ -19,7 +19,7 @@ const FakeSingleTabStopNavigationProvider = forwardRef(
   ) => {
     const focusablesRef = useRef(new Set<Element>());
     const focusHandlersRef = useRef(new Map<Element, FocusableChangeHandler>());
-    const registerFocusable = useCallback((focusable: Element, changeHandler: FocusableChangeHandler) => {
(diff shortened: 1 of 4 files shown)


avinashbot (commented):
I'm a little confused on how the focusable "registration" system as a whole works. Reached out to you about it, but also dropping this one minor comment.

pan-kot:
> I'm a little confused on how the focusable "registration" system as a whole works. Reached out to you about it, but also dropping this one minor comment.

The registration mechanism makes the table aware of the interactive content inside so that every time the keyboard command it received or the focus is moved in other way all relevant elements are considered.

avinashbot (approved):

PR #2119, opened 2024-04-05
refactor: Table expandable item props internal api rework

pan-kot:
### Description

Updated internal table APIs to support upcoming change https://github.com/cloudscape-design/components/pull/2108.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/table/body-cell/index.tsx
+++ b/src/table/body-cell/index.tsx
@@ -154,14 +154,15 @@ export function TableBodyCell<ItemType>({
   isEditable,
   ...rest
 }: TableBodyCellProps<ItemType> & { isEditable: boolean }) {
+  const isExpandableColumnCell = rest.level !== undefined;
   const editDisabledReason = rest.column.editConfig?.disabledReason?.(rest.item);
 
   // Inline editing is deactivated for expandable column because editable cells are interactive
   // and cannot include interactive content such as expand toggles.
-  if (editDisabledReason && rest.level === undefined) {
+  if (editDisabledReason && !isExpandableColumnCell) {
     return <DisabledInlineEditor editDisabledReason={editDisabledReason} {...rest} />;
   }
-  if ((isEditable || rest.isEditing) && rest.level === undefined) {
+  if ((isEditable || rest.isEditing) && !isExpandableColumnCell) {
     return <TableCellEditable {...rest} />;
   }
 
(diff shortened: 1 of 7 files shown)


src/table/internal.tsx
```diff
@@ -532,9 +534,7 @@ const InternalTable = React.forwardRef(
                                   submitEdit={cellEditing.submitEdit}
                                   columnId={column.id ?? colIndex}
                                   colIndex={colIndex + colIndexOffset}
-                                  // Expandable props only apply to the first data column of the table.
-                                  // When present, the cell content is decorated with expand toggles and extra paddings.
-                                  expandableProps={isExpandable && colIndex === 0 ? expandableProps : undefined}
+                                  {...expandableCellProps}
```
pan-kot, line 537: This change it necessary to not provide empty onExpandableItemToggle when only offset is needed, when a cell does not have the expand toggle but requires an offset.

src/table/body-cell/index.tsx
```diff
@@ -157,10 +156,12 @@ export function TableBodyCell<ItemType>({
 }: TableBodyCellProps<ItemType> & { isEditable: boolean }) {
   const editDisabledReason = rest.column.editConfig?.disabledReason?.(rest.item);
 
-  if (editDisabledReason && !rest.expandableProps) {
+  // Inline editing is deactivated for expandable column because editable cells are interactive
```
Al-Dani, line 159: should we put this in usage guidelines too? I know they're pretty complex already, but i can imagine this case to come up one day

src/table/body-cell/index.tsx
```diff
@@ -157,10 +156,12 @@ export function TableBodyCell<ItemType>({
...
+  // and cannot include interactive content such as expand toggles.
+  if (editDisabledReason && rest.level === undefined) {
```
Al-Dani, line 161: I'd suggest to create a constant with helpful name for the future
Smth like
```js
const cellWithExpandableToggle = rest.level !== undefined
if(editDisabledReason && !cellWithExpandableToggle) {..}
```

src/table/body-cell/index.tsx
pan-kot: We plan to remove this constraint in the near future. It wasn't done as part of the initial release due to design and dev complexity.

Al-Dani (approved):

PR #2121, opened 2024-04-05
chore: Add RTL support for Popover component.

DaemonCahill:
This PR adds RTL support for the `Popover` component.


DaemonCahill:
@jperals For visibility, I did a deep dive into the Codecov patch report with @connorlanigan this morning. The lines that were flagged in the dropdown are extremely difficult to cover in unit tests. We verified that these lines of code are executed and covered by both integration tests and screenshot tests. Our conclusion is that that patch diff block can be overridden for this PR.

src/internal/components/dropdown/dropdown-fit-handler.ts
```diff
@@ -104,28 +115,34 @@ export const getInteriorAvailableSpace = ({
...
+    insetInlineEnd: triggerRight,
+  } = getLogicalBoundingClientRect(trigger);
 
   return overflowParents.reduce(
-    ({ above, below, left, right }, overflowParent) => {
-      const currentAbove = triggerBottom - overflowParent.top - AVAILABLE_SPACE_RESERVE_VERTICAL;
-      const currentBelow = overflowParent.height - triggerTop + overflowParent.top - AVAILABLE_SPACE_RESERVE_VERTICAL;
-      const currentLeft = triggerLeft - overflowParent.left - AVAILABLE_SPACE_RESERVE_HORIZONTAL;
+    ({ blockStart, blockEnd, inlineStart, inlineEnd }, overflowParent) => {
+      const currentAbove = triggerBottom - overflowParent.insetBlockStart - AVAILABLE_SPACE_RESERVE_VERTICAL;
```
jperals, line 126: Since we are refactoring internal APIs already, why not these variable names too to make it consistent?

src/internal/components/dropdown/dropdown-fit-handler.ts
```diff
@@ -141,13 +158,13 @@ export const getWidths = ({
   stretchBeyondTriggerWidth?: boolean;
 }) => {
   // Determine the width of the trigger
-  const triggerWidth = triggerElement.getBoundingClientRect().width;
+  const { inlineSize: triggerWidth } = getLogicalBoundingClientRect(triggerElement);
```
jperals, line 161: Same here: can we make the variable names also logical so they stay up to date with their actual meaning?

pages/popover/container-test.page.tsx
```diff
@@ -33,7 +34,11 @@ export default function () {
     <article>
       <h1>Popover container test</h1>
       <ScreenshotArea>
-        <svg width={500} height={500} style={{ margin: '20px' }}>
+        <svg
+          width={500}
+          height={500}
+          style={{ margin: '20px', transform: `${isRtl(document.body) ? 'scaleX(-1)' : ''}` }}
```
jperals, line 40: this
```
`${isRtl(document.body) ? 'scaleX(-1)' : ''}`
```
could be just
```
isRtl(document.body) ? 'scaleX(-1)' : ''
```

src/internal/components/chart-popover/index.tsx
```diff
@@ -20,7 +20,7 @@ export interface ChartPopoverProps extends PopoverProps {
   title?: React.ReactNode;
 
   /** References the element the container is positioned against. */
-  trackRef: React.RefObject<Element>;
+  trackRef: React.RefObject<HTMLElement | SVGElement>;
```
jperals, line 23: Why are these type changes necessary?

jperals:
The popover in RTL inverts the value given as `position`. Is this intended? RTL means that we read from right to left but not that left is right and right is left.

src/internal/components/dropdown/dropdown-fit-handler.ts
DaemonCahill: Updated!

src/internal/components/dropdown/dropdown-fit-handler.ts
DaemonCahill: Updated!

pages/popover/container-test.page.tsx
DaemonCahill: Updated!

src/internal/components/chart-popover/index.tsx
DaemonCahill: In the previous PR we decided to type the element that is passed to `isRtl` to be `HTMLElement | SVGElement`. I applied the same typing to the `getLogicalBoundingClientRect` function which required a type update.

DaemonCahill:
> The popover in RTL inverts the value given as `position`. Is this intended? RTL means that we read from right to left but not that left is right and right is left.

@jperals Using the popover positioning demo page as the source of truth, this behavior is intended. My understanding is that when a developer implementing a component specifies `right` for the position, they don't intend `right` in all direction contexts but rather the `inline-end` position. Thus, `right` will equate to `left`, and vice versa, based on the value of the position relative to the element direction.

src/internal/components/chart-popover/index.tsx
jperals: I see that this is actually _more_ restrictive now than before, so no problems with this.

jperals (approved):

PR #2123, opened 2024-04-08
fix: Prevent ariaLabel on code editor from conflicting with ace's built-in aria-label

avinashbot:
### Description

We're overriding ace's built-in `aria-label` by using our own `aria-labelledby`, which is a property not touched by ace. Ace's built-in aria-label provides cursor information, so we need to fetch and include that information ourselves. After consulation, using aria-describedby to provide the label has been ruled out, so we have to go with this approach. A number of other alternatives (using `setMessages`, using the `groupAriaLabel` instead of the textarea, etc.) have also been ruled out.

**Important:** The strings are being translated now, and they'll be included in a separate PR, but merged at this same time as this one.

Related links, issue #, if available: AWSUI-24619

#### API changes

- New `i18nStrings` property for code editor: `cursorPositionAriaLabel`
  - Expected usage: `cursorPositionAriaLabel: row => 'Cursor at row ' + row`
  - This string is combined with the ariaLabel.
  - The idea is that we release with `I18nProvider` support from the very beginning, so it's a no-touch upgrade for all internationalization users, and even if they don't, having a working `ariaLabel` is a better alternative than just getting just the cursor position (but obviously, both would be great).

### How has this been tested?

Updated relevant tests.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/code-editor/interfaces.ts
+++ b/src/code-editor/interfaces.ts
@@ -148,6 +148,7 @@ export namespace CodeEditorProps {
 
     editorGroupAriaLabel?: string;
     statusBarGroupAriaLabel?: string;
+    cursorPositionAriaLabel?: (row: number) => string;
 
     cursorPosition?: (row: number, column: number) => string;
     errorsTab?: string;
--- a/src/__tests__/__snapshots__/documenter.test.ts.snap
+++ b/src/__tests__/__snapshots__/documenter.test.ts.snap
@@ -4542,6 +4542,11 @@ The object should contain, among others:
             "type": "(row: number, column: number) => string",
           },
           Object {
+            "name": "cursorPositionAriaLabel",
+            "optional": true,
+            "type": "(row: number) => string",
+          },
+          Object {
             "name": "editorGroupAriaLabel",
             "optional": true,
             "type": "string",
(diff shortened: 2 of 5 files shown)


src/code-editor/interfaces.ts
```diff
@@ -148,6 +148,7 @@ export namespace CodeEditorProps {
 
     editorGroupAriaLabel?: string;
     statusBarGroupAriaLabel?: string;
+    cursorPositionAriaLabel?: (row: number) => string;
```
johannes-weber, line 151: I think this is an opportunity to streamline the docs for the CodeEditors `i18n` property. Currently the API docs for the property contains just a few of them. I think we could add an explanaition for all properties including an example similar to what we do at the [AppLayout's API doc](https://cloudscape.design/components/app-layout/?tabId=api#:~:text=false-,ariaLabels,-AppLayoutProps.Labels).

Let me know what you think about.

avinashbot: Ehhh, we *could*, but how much value is that going to provide? We don't recommend any new users of the component provide `i18nStrings` manually, and to just use `I18nProvider` instead. So it doesn't feel particularly impactful, especially with the content review that all the properties have to go through.

PR #2134, opened 2024-04-11
fix: Do not display popover inline when using portal

jperals:
### Description

Corrected rework of #2116 (reverted by #2133), but having inline popovers into account. The new corrected logic is to add `display: inline` only if not using a portal.

Issue: AWSUI-40397

### How has this been tested?

- Ran in my pipeline, getting the same expected diffs in the Firefox diffs as in #2116
- Added test page to cover inline popovers. Tested manually and will add a visual regression test.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/popover/styles.scss
+++ b/src/popover/styles.scss
@@ -51,6 +51,6 @@
   }
 }
 
-.popover-content {
+.popover-inline-content {
   display: inline;
 }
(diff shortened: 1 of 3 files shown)


src/popover/styles.scss
```diff
@@ -51,6 +51,6 @@
   }
 }
 
-.popover-content {
+.popover-inline-content {
```
jperals, line 54: Changed the class name to something more appropriate to what it does

abdhalees (approved):

PR #2137, opened 2024-04-11
refactor: Merge app layout resize observers

just-boris:
### Description

Another code deduplication between classic and refresh app layouts

### How has this been tested?

* Locally
* Screenshot tests in my dev pipeline
* See comments about updated tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/app-layout/__tests__/split-panel.test.tsx
```diff
@@ -109,11 +109,15 @@ for (const theme of ['refresh', 'classic']) {
...
           />
         );
-        expect(wrapper.findSplitPanel()!.findOpenButton()).not.toBeNull();
-        act(() => wrapper.findSplitPanel()!.findOpenButton()!.click());
-        expect(wrapper.findSplitPanel()!.findOpenButton()).toBeNull();
-        act(() => wrapper.findSplitPanel()!.findCloseButton()!.click());
-        expect(wrapper.findSplitPanel()!.findOpenButton()).not.toBeNull();
+        expect(wrapper.findSplitPanelOpenButton()).not.toBeNull();
+        wrapper.findSplitPanelOpenButton()!.click();
+        if (position === 'side' && theme === 'refresh') {
```
just-boris, line 114: This test was a false positive. It was supposed to test bottom and side split panel position, but because JSDOM defaults it always assumed "forced bottom" position.

I fixed this long ago for classic in #1952 and now I reused some of that code for refresh, which made this test work correctly.

The expected design requirement is that in refresh and side position, the open button _stays visible_. So the test matches the original requirements.

src/app-layout/__tests__/split-panel.test.tsx
```diff
@@ -124,7 +128,7 @@ for (const theme of ['refresh', 'classic']) {
             onSplitPanelPreferencesChange={noop}
           />
         );
-        act(() => wrapper.findSplitPanel()!.findOpenButton()!.click());
```
just-boris, line 127: Just dropping `act` everywhere because they are useless. `.click()` already does `act` when needed.

src/app-layout/__tests__/split-panel.test.tsx
```diff
@@ -136,9 +140,9 @@ for (const theme of ['refresh', 'classic']) {
...
-        act(() => wrapper.findSplitPanel()!.findCloseButton()!.click());
-        expect(wrapper.findSplitPanel()!.findOpenButton()!.getElement()).toHaveFocus();
+        wrapper.findSplitPanelOpenButton()!.click();
```
just-boris, line 143: When split panel is closed, `findSplitPanel` may return null. But the open button is available because it is rendered outside, directly in app layout. To open that panel, need to skip `findSplitPanel` and use `findSplitPanelOpenButton`

We have got a few tests relying on this behavior, so I updated them to use the new test-util

src/app-layout/__tests__/split-panel.test.tsx
```diff
@@ -331,5 +340,5 @@ test('should not set width on split panel drawer when there is no splitPanel', (
       splitPanelPreferences={{ position: 'side' }}
     />
   );
-  expect(wrapper.find('[data-testid="side-split-panel-drawer"]')?.getElement().style.width).toBeFalsy();
+  expect(wrapper.find('[data-testid="side-split-panel-drawer"]')!.getElement().style.width).toBeFalsy();
```
just-boris, line 343: There are no good reasons to use `?` in tests. Always use `!`

src/test-utils/dom/app-layout/index.ts
```diff
@@ -48,6 +49,10 @@ export default class AppLayoutWrapper extends ComponentWrapper {
     return this.findComponent(`.${SplitPanelWrapper.rootSelector}`, SplitPanelWrapper);
   }
 
+  findSplitPanelOpenButton(): ElementWrapper | null {
+    return this.findByClassName(splitPanelStyles['open-button']);
+  }
```
just-boris, line 54: This test-util appeared useful in many tests, so I decided to expose it here as well.

YueyingLu (approved):

src/app-layout/__tests__/split-panel.test.tsx
just-boris: because we already do it inside the test-util: https://github.com/cloudscape-design/test-utils/blob/595baac332382d22b87ad210f26010a02cbc3f67/src/core/dom.ts#L65-L67

No need to duplicate. Also, it is for testing development experience of our consumers, see AWSUI-22860. 

Every time you feel like you need to do act in our tests, it is probably a sign that we should release a test-util giving it to our consumers automatically.

Eventually I expect to remove all manual `act` from our tests, little-by-little, like in this PR. If you can clean our code too when doing your changes, it would be great

PR #2155, opened 2024-04-12
chore: Add RTL support for the Table component.

DaemonCahill:
This PR adds RTL support for the `Table` component.

### DIFF
--- a/src/table/table-role/grid-navigation.tsx
+++ b/src/table/table-role/grid-navigation.tsx
@@ -193,27 +193,19 @@ class GridNavigationProcessor {
       return;
     }
 
-    const keys = [
-      KeyCode.up,
-      KeyCode.down,
-      KeyCode.left,
-      KeyCode.right,
-      KeyCode.pageUp,
-      KeyCode.pageDown,
-      KeyCode.home,
-      KeyCode.end,
-    ];
     const ctrlKey = event.ctrlKey ? 1 : 0;
     const altKey = event.altKey ? 1 : 0;
     const shiftKey = event.shiftKey ? 1 : 0;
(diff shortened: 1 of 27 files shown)


src/table/table-role/grid-navigation.tsx
```diff
@@ -191,69 +192,54 @@ class GridNavigationProcessor {
...
+
+    isEventLike(event) &&
+      handleKey(event, {
+        onBlockStart: () => this.moveFocusBy(from, { y: -1, x: 0 }),
+        onBlockEnd: () => this.moveFocusBy(from, { y: 1, x: 0 }),
+        onInlineStart: () => this.moveFocusBy(from, { y: 0, x: -1 }),
+        onInlineEnd: () => this.moveFocusBy(from, { y: 0, x: 1 }),
+        onPageUp: () => this.moveFocusBy(from, { y: -this.pageSize, x: 0 }),
+        onPageDown: () => this.moveFocusBy(from, { y: this.pageSize, x: 0 }),
+        onHome: () =>
```
pan-kot, line 234: A non-blocking observation here. On a MacBook there is no Home/End buttons on the keyboard. Instead, you can press Fn+Left/Fn+Right. Do you know if the Home/End are reversed when switching to an RTL language? If not, the interaction feels wrong.

pan-kot (approved):
Awesome work! 👍

src/table/table-role/grid-navigation.tsx
DaemonCahill: My understanding is that Home/End are universal, which can make for a confusing experience on Mac because the Fn+Right will go physically Left in RTL. I double checked and this is the intended behavior.

PR #2163, opened 2024-04-16
chore: Add extra metadata to app layout props reporting

just-boris:
### Description

Add additional data to telemetry to better understand our use-cases

### How has this been tested?

PR build

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/index.tsx
+++ b/src/app-layout/index.tsx
@@ -57,6 +57,10 @@ const AppLayout = React.forwardRef(
         stickyNotifications: rest.stickyNotifications,
         disableContentHeaderOverlap: rest.disableContentHeaderOverlap,
       },
+      metadata: {
+        drawersCount: rest.drawers?.length ?? null,
+        hasContentHeader: !!rest.contentHeader,
+      },
     });
     const isRefresh = useVisualRefresh();
     const isMobile = useMobile();


src/app-layout/index.tsx
```diff
@@ -57,6 +57,10 @@ const AppLayout = React.forwardRef(
         stickyNotifications: rest.stickyNotifications,
         disableContentHeaderOverlap: rest.disableContentHeaderOverlap,
       },
+      metadata: {
+        drawersCount: rest.drawers?.length ?? null,
```
jperals, line 61: Any reason why it needs to be `null` and not 0 or `undefined`?

jperals:
Not specific to this PR, but should we start adding test coverage for these metadata metrics at the component level?

src/app-layout/index.tsx
just-boris: They all have different meaning: 

* `<AppLayout drawers={[]}>` this will report zero
* `<AppLayout />` this will report null
* We cannot use `undefined`, because `JSON.stringify({value: undefined})` produces `"{}"`, so we do not know if there are no drawers or an old version without metric is used

just-boris:
When we do not have any conditional reporting, not sure what value you will get from the test. It is a typical snapshot test anti-pattern.

just-boris:
However I added a test on `null` handling in the toolkit, so we can trust that we will see `null` values in our metrics

https://github.com/cloudscape-design/component-toolkit/pull/64

jperals (approved):

PR #2164, opened 2024-04-17
fix: Allow touch devices to resize table columns

avinashbot:
### Description

Does what it says on the tin. I checked it in Chrome's touch "emulation" mode, on Android, and iPhone (through BrowserStack). What actually scares me is whether downstream tests will break after this. But the dry-run seems to look good, so...

### How has this been tested?

Manually. Existing tests should continue to work... I'm really counting on them to.

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/table/__tests__/columns-auto-resize.test.tsx
```diff
@@ -4,7 +4,7 @@ import * as React from 'react';
 import { render, act } from '@testing-library/react';
 import createWrapper from '../../../lib/components/test-utils/dom';
 import Table, { TableProps } from '../../../lib/components/table';
-import { fireMousedown, fireMouseMove, fireMouseup, fakeBoundingClientRect } from './utils/resize-actions';
+import { firePointerdown, firePointermove, firePointerup, fakeBoundingClientRect } from './utils/resize-actions';
```
just-boris, line 7: After changing that many tests, are you not afraid that the tests of our consumers will fail too?

How much do we need this change vs how much do you want to support teams following up their release?

src/table/resizer/styles.scss
```diff
@@ -50,6 +50,12 @@ th:not(:last-child) > .divider {
   inset-inline-end: calc(-1 * #{$handle-width} / 2);
   inline-size: $handle-width;
   z-index: 10;
+
+  // Allow touch devices to use the resizer horizontally.
+  // The lint isn't correct: "pan-y" and "pinch-zoom" values are supported.
```
just-boris, line 55: Is it possible to upgrade the related dependency, instead of writing a comment?

Seeing [how often we ignore that rule already](https://github.com/search?q=repo%3Acloudscape-design%2Fcomponents%20no-unsupported-browser-features&type=code) it is also possible to remove that rule all together

src/table/__tests__/columns-auto-resize.test.tsx
avinashbot: I am most definitely afraid lol - but dry-run passes and I couldn't find anything on the internal code search regarding test with mouse/table terms. I did my precautions here, but of course there's always a chance.

But the balance here is between providing a tangibly better experience for touch devices and get ahead of a mobile bug, or just wait until we get a report.

src/table/resizer/styles.scss
avinashbot: I'll check, but my understanding was is that the rule is only "partially" supported on Safari, which is why the rule is triggering. A reasonable situation for a scanner to flag it, but I'll see if it's fixed more recently.

just-boris (dismissed):

PR #2171, opened 2024-04-17
chore: Apply rounded courners to first and last stacked bars only

pan-kot:
### Description

Updated border rounding for stacked bar charts (horizontal and vertical, classic and VR) so that not every segment but only the edge ones are rounded from the top or from the bottom. That is to avoid an issue when small segments that are a few pixel tall/wide look awkward when rounded.

**Before**
Screenshot 2024-04-18 at 15 21 25

**After**
Screenshot 2024-04-18 at 15 23 03

### How has this been tested?

* Existing permutations for bar charts

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/mixed-line-bar-chart/utils.ts
+++ b/src/mixed-line-bar-chart/utils.ts
@@ -77,20 +77,18 @@ export function calculateStackedBarValues(
   const negativeValues = new Map<string | number, number>();
   const positiveValues = new Map<string | number, number>();
   const values = new Map<string | number, Map<number, number>>();
-  const addValue = (seriesIndex: number, key: string | number, value: number) => {
-    if (value < 0) {
-      negativeValues.set(key, (negativeValues.get(key) ?? 0) + value);
-    } else {
-      positiveValues.set(key, (positiveValues.get(key) ?? 0) + value);
-    }
-    const seriesValue = (value < 0 ? negativeValues.get(key) : positiveValues.get(key)) ?? 0;
-    const valuesByIndex = values.get(key) ?? new Map<number, number>();
-    valuesByIndex.set(seriesIndex, seriesValue);
-    values.set(key, valuesByIndex);
-  };
   for (let seriesIndex = 0; seriesIndex < dataBySeries.length; seriesIndex++) {
     for (const datum of dataBySeries[seriesIndex]) {
-      addValue(seriesIndex, getKeyValue(datum.x), datum.y);
+      const key = getKeyValue(datum.x);
+      if (datum.y < 0) {
+        negativeValues.set(key, (negativeValues.get(key) ?? 0) + datum.y);
+      } else {
+        positiveValues.set(key, (positiveValues.get(key) ?? 0) + datum.y);
+      }
+      const seriesValue = (datum.y < 0 ? negativeValues.get(key) : positiveValues.get(key)) ?? 0;
+      const valuesByIndex = values.get(key) ?? new Map<number, number>();
+      valuesByIndex.set(seriesIndex, seriesValue);
+      values.set(key, valuesByIndex);
     }
   }
   return values;


src/mixed-line-bar-chart/bar-series.tsx
```diff
@@ -126,30 +135,68 @@ export default function BarSeries<T extends ChartDataTypes>({
...
   );
 }
+
+function stretchRect(
+  rect: { x: number; y: number; height: number; width: number },
+  direction: 'left' | 'right' | 'up' | 'down'
+) {
+  switch (direction) {
+    case 'up':
+      return { ...rect, height: rect.height + 10 };
```
just-boris, line 194: Where are these 10 numbers came from? I do not see it in the original code

src/mixed-line-bar-chart/data-series.tsx
```diff
@@ -65,6 +65,13 @@ export default function DataSeries<T extends ChartDataTypes>({
     return calculateOffsetMaps(barData);
   }, [visibleSeries, stackedBars]);
 
+  const getStackedMinimum = (xValue: ChartDataTypes) => {
+    return stackedBarOffsetMaps?.minValues.get(getKeyValue(xValue)) ?? 0;
```
just-boris, line 69: When is this value undefined and why is it safe to assume 0 here?

src/mixed-line-bar-chart/utils.ts
```diff
@@ -95,9 +109,13 @@ export function calculateOffsetMaps(
...
         const lastValue = lastMap?.negativeOffsets.get(key) || 0;
         map.negativeOffsets.set(key, lastValue + y);
+        setMinValue(key, lastValue + y);
+        setMaxValue(key, lastValue + y);
       } else {
         const lastValue = lastMap?.positiveOffsets.get(key) || 0;
         map.positiveOffsets.set(key, lastValue + y);
+        setMinValue(key, lastValue + y);
+        setMaxValue(key, lastValue + y);
       }
```
just-boris, line 119: Don't you see any issues when repeating `lastValue + y` 6 times here?

src/mixed-line-bar-chart/utils.ts
```diff
@@ -108,6 +126,8 @@ export function calculateOffsetMaps(
 
     return acc;
   }, [] as StackedOffsets[]);
+
+  return { minValues, maxValues, seriesOffsets };
```
just-boris, line 130: If `seriesOffsets` already contains sorted data, isn't it self explanatory which item is min and which item is max? Do you really need a separate value here?

pan-kot: The `seriesOffsets` includes data for a single series only. For example:

```
series = "John"
seriesOffsets = {
  positive: { "Apples": 0, "Oranges": 5 },
  negative: { "Pears": 0  }
}
```

The chart is rendered series by series, so the John series will render right above the X axis for "Apples", above the X axis and some other series for "Oranges", and right below the X axis for "Pears". This carries no information about other series.

src/mixed-line-bar-chart/utils.ts
pan-kot: I will refactor this utility, resolving the thread.

src/mixed-line-bar-chart/data-series.tsx
pan-kot: Logically, the min/max values are defined for each value so do not expect any to be undefined. However, I found how to make the overall code simpler - resolving the thread.

src/mixed-line-bar-chart/bar-series.tsx
pan-kot: That's a good question. I will add a code comment to explain that.

src/mixed-line-bar-chart/utils.ts
pan-kot: Thanks for the comment though, looking at it now I realised it worth taking a step back and create a better data structure than series offsets to accommodate all needs.

src/mixed-line-bar-chart/bar-series.tsx
just-boris: See the comment now but still do not understand what is happening? Why is this 10 pixels and not 5 or 100?

src/mixed-line-bar-chart/data-series.tsx
```diff
@@ -62,7 +61,7 @@ export default function DataSeries<T extends ChartDataTypes>({
         barData.push([]);
       }
     });
-    return calculateOffsetMaps(barData);
+    return calculateStackedBarValues([...barData]);
```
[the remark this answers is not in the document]
pan-kot: Sorry, that's a leftover code. I was experimenting with changing the argument type to accept the readonly array. Reverted this back.

PR #2154, opened 2024-04-12
WIP: Tabs Actions

sderacy:
### Description

This is not yet in a state to merge, but this implements actions (button dropdowns here) along with a dismissible prop.

I will follow up with a change to implement a tab focus trap as we agreed the current tab flow w/ actions is not acceptable accessibility wise.

Related links, issue #, if available: n/a

### How has this been tested?

<details>
   <summary>Review checklist</summary>

_The following items are to be evaluated by the author(s) and the reviewer(s)._

#### Correctness

- _Changes include appropriate documentation updates._
- _Changes are backward-compatible if not indicated, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#public-apis)._
- _Changes do not include unsupported browser features, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#browsers-support)._
- _Changes were manually tested for accessibility, see [accessibility guidelines](https://cloudscape.design/foundation/core-principles/accessibility/)._

#### Security

- _If the code handles URLs: all URLs are validated through [the `checkSafeUrl` function](https://github.com/cloudscape-design/components/blob/main/src/internal/utils/check-safe-url.ts)._

#### Testing

- _Changes are covered with new/existing unit tests?_
- _Changes are covered with new/existing integration tests?_
</details>

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2173, opened 2024-04-18
feat: Implements actions and optional dismissible on tabs component

sderacy:
### Description

This is not yet in a state to merge, but this implements actions (button dropdowns here) along with a dismissible prop.

I will follow up with a change to implement a tab focus trap as we agreed the current tab flow w/ actions is not acceptable accessibility wise.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

ref: https://github.com/cloudscape-design/components/pull/2154
---

### DIFF
--- a/src/tabs/interfaces.ts
+++ b/src/tabs/interfaces.ts
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 import { BaseComponentProps } from '../internal/base-component';
 import { NonCancelableEventHandler } from '../internal/events';
+import { ButtonProps } from '../button/interfaces';
 
 export interface TabsProps extends BaseComponentProps {
   /**
@@ -90,6 +91,24 @@ export namespace TabsProps {
      * Whether this tab is disabled.
      */
     disabled?: boolean;
+    /**
+     * (Optional) Determines whether the tab includes a close icon button. By default, the close button is not included.
+     * When a user clicks on this button the `onClose` handler is called.
+     */
+    dismissible?: boolean;
+    /**
+     * (Optional) Specifies an `aria-label` for the close icon button for improved accessibility.
+     */
+    dismissLabel?: string;
+    /**
+     * (Optional) Action for the tab, rendered next to its label. Although it is technically possible to insert any content,
+     * our UX guidelines only allow you to add an icon button or icon button dropdown.
+     */
+    action?: React.ReactNode;
+    /**
+     * (event => void) Called when a user clicks on the close button.
+     */
+    onDismiss?: ButtonProps['onClick'];
     /**
      * You can use this parameter to change the default `href` of the internal tab anchor. The
      * `click` event default behavior is prevented, unless the user clicks the tab with a key modifier (CTRL,
--- a/src/tabs/tab-header-bar.scss
+++ b/src/tabs/tab-header-bar.scss
@@ -70,28 +70,85 @@ $label-horizontal-spacing: awsui.$space-l;
   flex-shrink: 0;
   display: flex;
   max-inline-size: calc(90% - #{$label-horizontal-spacing});
-
-  & > button {
-    background-color: transparent;
-  }
 }
 
 .tabs-tab-label {
   display: flex;
   align-items: center;
-  padding-block: awsui.$space-scaled-2x-xxs;
-  padding-inline: $label-horizontal-spacing;
   text-align: start;
   position: relative;
 
   @include styles.text-wrapping;
 }
 
+.tabs-tab-header-content {
+  position: relative;
+
+  border-block: awsui.$border-divider-section-width solid transparent;
+  border-inline: awsui.$border-divider-section-width solid transparent;
+
+  padding-block: awsui.$space-scaled-2x-xxs;
+  padding-inline: $label-horizontal-spacing;
+
+  &,
+  & > button {
+    background-color: transparent;
+    display: flex;
+    align-items: center;
+    text-align: start;
+  }
+
+  &:focus {
+    outline: none;
+  }
+  @include styles.font-smoothing;
+
+  @include focus-visible.when-visible {
+    z-index: 1;
+    @include styles.focus-highlight(awsui.$space-tabs-focus-outline-gutter);
+    border-inline-end-color: transparent;
+  }
+}
+
+.tabs-tab-header-content:not(.tabs-tab-disabled) {
+  // This is the underline for the currently selected tab.
+  &:after {
+    content: '';
+    position: absolute;
+    inset-inline-start: 0;
+    inline-size: 100%;
+    inset-block-end: calc(-1 * #{awsui.$border-divider-section-width});
+    block-size: awsui.$border-active-width;
+    border-start-start-radius: awsui.$border-radius-tabs-focus-ring;
+    border-start-end-radius: awsui.$border-radius-tabs-focus-ring;
+    border-end-start-radius: awsui.$border-radius-tabs-focus-ring;
+    border-end-end-radius: awsui.$border-radius-tabs-focus-ring;
+    background: awsui.$color-border-tabs-underline;
+    opacity: 0;
+  }
+
+  &.refresh {
+    &:after {
+      @include styles.with-motion {
+        transition: opacity awsui.$motion-duration-refresh-only-medium awsui.$motion-easing-refresh-only-c;
+      }
+    }
+  }
+}
+
 .tabs-tab:not(:last-child) {
-  & > a > .tabs-tab-label,
-  & > button > .tabs-tab-label {
+  & > .tabs-tab-header-content {
     margin-inline-end: calc(-1 * #{awsui.$border-divider-section-width});
-    border-inline-end: awsui.$border-divider-section-width solid $separator-color;
+    &:before {
+      content: '';
+      block-size: 1.7rem;
+      inline-size: #{awsui.$border-divider-section-width};
+      position: absolute;
+      inset-inline-end: 0;
+      margin-inline-end: calc(-1 * #{awsui.$border-divider-section-width});
+      background: $separator-color;
+      opacity: 1;
+    }
   }
 }
 
@@ -125,23 +182,16 @@ $label-horizontal-spacing: awsui.$space-l;
   &:focus {
     outline: none;
   }
-  @include styles.font-smoothing;
+}
 
-  @include focus-visible.when-visible {
-    z-index: 1;
-    @include styles.focus-highlight(awsui.$space-tabs-focus-outline-gutter);
-    & > a > .tabs-tab-label,
-    & > button > .tabs-tab-label {
-      border-inline-end-color: transparent;
-    }
-  }
+.tabs-tab-link-only {
+  margin-inline-end: awsui.$space-scaled-m;
 }
 
 // Remediate focus shadow
 .tabs-tab:first-child {
   margin-inline-start: 1px;
-  & > a > .tabs-tab-label,
-  & > button > .tabs-tab-label {
+  & > .tabs-tab-header-content {
     padding-inline-start: calc(#{$label-horizontal-spacing} - 1px);
   }
 }
@@ -149,8 +199,7 @@ $label-horizontal-spacing: awsui.$space-l;
 // Remediate focus shadow
 .tabs-tab:last-child {
   margin-inline-end: 1px;
-  & > a > .tabs-tab-label,
-  & > button > .tabs-tab-label {
+  & > .tabs-tab-header-content {
     padding-inline-end: calc(#{$label-horizontal-spacing} - 1px);
   }
 }
@@ -162,32 +211,6 @@ $label-horizontal-spacing: awsui.$space-l;
     cursor: default;
     color: awsui.$color-text-interactive-disabled;
     font-weight: awsui.$font-tabs-disabled-weight;
-  }
-}
-
-.tabs-tab-link:not(.tabs-tab-disabled) {
-  // This is the underline for the currently selected tab.
-  &:after {
-    content: '';
-    position: absolute;
-    inset-inline-start: 0;
-    inline-size: 100%;
-    inset-block-end: calc(-1 * #{awsui.$border-divider-section-width});
-    block-size: awsui.$border-active-width;
-    border-start-start-radius: awsui.$border-radius-tabs-focus-ring;
-    border-start-end-radius: awsui.$border-radius-tabs-focus-ring;
-    border-end-start-radius: awsui.$border-radius-tabs-focus-ring;
-    border-end-end-radius: awsui.$border-radius-tabs-focus-ring;
-    background: awsui.$color-border-tabs-underline;
-    opacity: 0;
-  }
-
-  &.refresh {
-    &:after {
-      @include styles.with-motion {
-        transition: opacity awsui.$motion-duration-refresh-only-medium awsui.$motion-easing-refresh-only-c;
-      }
-    }
   }
 }
 
--- a/src/tabs/tab-header-bar.tsx
+++ b/src/tabs/tab-header-bar.tsx
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 import React, { useRef, useState, useEffect } from 'react';
 import { TabsProps } from './interfaces';
+import { ButtonProps } from '../button/interfaces';
 import clsx from 'clsx';
 import styles from './styles.css.js';
 import { InternalButton } from '../button/internal';
@@ -18,6 +19,19 @@ import { hasModifierKeys, isPlainLeftClick } from '../internal/events';
 import { useVisualRefresh } from '../internal/hooks/use-visual-mode';
 import { useInternalI18n } from '../i18n/context';
 import { useContainerQuery } from '@cloudscape-design/component-toolkit';
+
+function dismissButton(dismissLabel: TabsProps.Tab['dismissLabel'], onDismiss: TabsProps.Tab['onDismiss']) {
+  return (
+    <InternalButton
+      onClick={onDismiss}
+      className={styles['dismiss-button']}
+      variant="icon"
+      iconName="close"
+      formAction="none"
+      ariaLabel={dismissLabel}
+    />
+  );
+}
 
 export interface TabHeaderBarProps {
   onChange: (changeDetail: TabsProps.ChangeDetail) => void;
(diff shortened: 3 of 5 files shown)


src/tabs/tab-header-bar.tsx
```diff
@@ -19,6 +20,35 @@ import { useVisualRefresh } from '../internal/hooks/use-visual-mode';
 import { useInternalI18n } from '../i18n/context';
 import { useContainerQuery } from '@cloudscape-design/component-toolkit';
 
+function setTabIndexForTabContainer(
```
timogasda, line 23: I'm not sure if it's really necessary to move this into its own function if it's only called in one place.

src/tabs/tab-header-bar.tsx
```diff
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
```
gethinwebster, line 2: is the plan to add `role="application"` to this PR?

gethinwebster (changes requested):

src/tabs/tab-header-bar.tsx
```diff
@@ -107,6 +138,18 @@ export function TabHeaderBar({
     }
   };
 
+  useEffect(() => {
```
gethinwebster, line 141: Mutating the DOM in `useEffect` can cause many problems with inconsistency, so we should avoid that if at all possible. Focusing probably does need to be in here, but setting the `tabIndex` should be done in the standard render logic.

src/test-utils/dom/tabs/index.ts
```diff
@@ -22,20 +23,69 @@ export default class TabsWrapper extends ComponentWrapper<HTMLButtonElement> {
...
   findTabLinkById(id: string): ElementWrapper<HTMLAnchorElement | HTMLButtonElement> | null {
     return this.find(`.${styles['tabs-tab-link']}[data-testid="${id}"]`);
   }
 
+  /**
+   * Finds the dimissible button by using the tab index
+   *
+   * @param index 1-based index of the clickable element to return
+   */
+  findDimissibleButtonByTabIndex(index: number): ButtonWrapper | null {
```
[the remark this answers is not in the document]
timogasda: we ended up calling the property `dismissible` though?

src/tabs/tab-header-bar.tsx
timogasda: Yes, Sterly wanted a general review first, just to gauge what the extent of the feedback from us will be 😅 So next step would be addressing the comments and implementing the role=application approach.

src/tabs/interfaces.ts
```diff
@@ -90,6 +91,24 @@ export namespace TabsProps {
      * Whether this tab is disabled.
      */
     disabled?: boolean;
+    /**
+     * (Optional) Determines whether the tab includes a close icon button. By default, the close button is not included.
```
gethinwebster, line 95: (here and throughout) maybe just "close button" rather than "close icon button". Or do we want to consistently refer to this as "dismiss button" to be consistent with the API? Depends a bit what terminology we use in the usage guidelines too

src/tabs/interfaces.ts
```diff
@@ -90,6 +91,24 @@ export namespace TabsProps {
...
+     * When a user clicks on this button the `onClose` handler is called.
+     */
+    dismissible?: boolean;
+    /**
+     * (Optional) Specifies an `aria-label` for the close icon button for improved accessibility.
```
gethinwebster, line 100: remove "for improved accessibility"

src/test-utils/dom/tabs/index.ts
```diff
@@ -22,20 +23,69 @@ export default class TabsWrapper extends ComponentWrapper<HTMLButtonElement> {
...
    *
-   * @param index ID of the clickable element to return
+   * @param id ID of the clickable element to return
    */
   findTabLinkById(id: string): ElementWrapper<HTMLAnchorElement | HTMLButtonElement> | null {
     return this.find(`.${styles['tabs-tab-link']}[data-testid="${id}"]`);
   }
 
+  /**
+   * Finds the dimissible button by using the tab index
```
gethinwebster, line 45: let's standardize on either "close" or "dismiss" button across here and the component API docs

.DS_Store
pan-kot, line 1: let's add .DS_Store and .vscode to gitignore

src/tabs/tab-header-bar.tsx
```diff
@@ -286,7 +354,17 @@ export function TabHeaderBar({
         role="presentation"
         key={tab.id}
       >
-        {trigger}
+        <div className={tabHeaderContainerClasses} role="group" aria-labelledby={commonProps.id}>
+          {trigger}
+          {action && (
+            <span className={styles['tabs-tab-action']} aria-label="More Actions" aria-haspopup="true">
```
pan-kot, line 360: Why is there an ARIA label and ARIA has-popup? For custom actions we should expect the consumers to set the ARIA label on the button or button-dropdown they provide. Besides, an ARIA label added to a span element w/o interactive role won't be read by screen readers.

src/tabs/tab-header-bar.scss
```diff
@@ -74,28 +75,75 @@ $label-horizontal-spacing: awsui.$space-l;
...
+
 .tabs-tab:not(:last-child) {
-  & > a > .tabs-tab-label,
-  & > button > .tabs-tab-label {
+  & > .tabs-tab-header-container {
     margin-inline-end: calc(-1 * #{awsui.$border-divider-section-width});
-    border-inline-end: awsui.$border-divider-section-width solid $separator-color;
+    &:before {
+      content: '';
+      block-size: 1.7rem;
```
gethinwebster, line 139: I think this could be a regression: previously if a tab label spanned multiple lines, the divider would be the full height, but now it would be fixed to always `1.7rem`. I'm not sure if this is something we actively want to support or not though. @timogasda @pan-kot can we check with design?

src/tabs/tab-header-bar.tsx
```diff
@@ -223,20 +255,37 @@ export function TabHeaderBar({
...
-          onScroll={onScroll}
-          onKeyDown={onKeyDown}
-          onFocus={onFocus}
-          onBlur={onBlur}
-        >
-          {tabs.map(renderTabHeader)}
-        </ul>
+        {hasActionOrDismissible ? (
+          <div
+            {...tabActionAttributes}
```
pan-kot, line 262: can we reuse more properties between actionable and non-actionable tabs?

src/tabs/tab-header-bar.tsx
```diff
@@ -292,16 +343,36 @@ export function TabHeaderBar({
...
+      [styles.refresh]: isVisualRefresh,
+      [styles['tabs-tab-active']]: activeTabId === tab.id && !tab.disabled,
+      [styles['tabs-tab-disabled']]: tab.disabled,
+    });
+
     const commonProps: (JSX.IntrinsicElements['a'] | JSX.IntrinsicElements['button']) & { 'data-testid': string } = {
       className: classes,
-      role: 'tab',
-      'aria-selected': tab.id === activeTabId,
+      role: hasActionOrDismissible ? undefined : 'tab',
```
pan-kot, line 355: why role is undefined for hasActionOrDismissible?

src/tabs/tab-header-bar.scss
sderacy: Good callout, I feel like they'd lean towards having a dynamic height - I believe I originally did this because the height would be the full height of the new container div which seemed too tall

.DS_Store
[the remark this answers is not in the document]
sderacy: Oh sorry, do we want to keep this file? I didn't intentionally add this so I am considering updating the .gitignore

src/tabs/__integ__/tabs.test.ts
```diff
@@ -275,6 +281,13 @@ test(
         await page.keys('Tab');
       }
       await page.keys(['End']);
+      /*
+       * Navigating to an action or dismissible buttons does not focus the tab
+       * Because the buttons are considered as part of the tab stop navigation
+       * The tab will not be considered active until you navigate to the label.
```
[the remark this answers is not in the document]
sderacy: That's fair, I can look into asserting against the element directly - I got too caught up trying to make it work against `page.isFocused`.

src/tabs/__integ__/tabs.test.ts
```diff
@@ -50,6 +52,18 @@ class TabsPage extends BasePageObject {
     await this.click('#size-toggle');
   }
 
+  async navigateTabList() {
```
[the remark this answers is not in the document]
sderacy: Sure! :)

src/tabs/__integ__/tabs.test.ts
```diff
@@ -302,3 +315,89 @@ test(
...
   })
 );
+
+test(
+  'verifies focus moves to dismissible via keyboard',
+  setupTest(async page => {
+    await page.focusTabHeader();
+    await page.click(wrapper.findTabLinkByIndex(3).toSelector());
+    await page.keys(['ArrowRight']);
+    // Since the focus doesn't actually update on the action/dismissible buttons,
```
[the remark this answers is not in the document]
sderacy: I was trying to use the `page.isDisplayed/isFocused` assertion to evaluate focus here, and `isFocused` evaluates to false in the case of action/dismissible since they don't invoke the `onChange` event to set focus to a tab.

The test here validates you can navigate to dismissible as part of the tab stop (unless we don't think it's needed)

src/tabs/__integ__/tabs.test.ts
```diff
@@ -302,3 +315,89 @@ test(
...
+    await page.click(wrapper.findTabLinkByIndex(3).toSelector());
+    await page.keys(['ArrowRight']);
+    await page.keys(['Tab']);
+    await page.keys(['Enter']);
+    await expect(page.isFocused(wrapper.findTabLinkByIndex(3).toSelector())).resolves.toBe(false);
+  })
+);
+
+test(
+  'verifies focus moves to action via keyboard',
```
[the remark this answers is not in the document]
sderacy: We can, I should've looked at how Cloudscape constructs other tests but I've been told in some cases the tests should be as granular as possible.

It's trivial to combine them so I can most certainly add it to my next commit

src/tabs/__integ__/tabs.test.ts
```diff
@@ -302,3 +315,89 @@ test(
...
+  setupTest(async page => {
+    await page.click(dismissibleWrapper.findTabLinkByIndex(1).toSelector());
+    await page.keys(['ArrowRight']);
+    await page.keys(['Enter']);
+    await expect(page.isFocused(dismissibleWrapper.findTabLinkByIndex(1).toSelector())).resolves.toBe(true);
+  })
+);
+
+test(
+  'prevents user from calling dismiss event twice in tab list of two',
```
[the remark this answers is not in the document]
sderacy: > What do we test here exactly?

We are testing the component does not invoke the onDismiss event if the user's tab component has a length of <= 1. 

> is that our component's limitation or is it the test page?

Component limitation. I added this because I was re-reading the tech-design and saw this: https://quip-amazon.com/jZhtASlAysv3#QYU9DAKyquI

I think there should be a stopgap measure on the Cloudscape measure to prevent this from happening but the user should be handling this case as well

src/tabs/tab-header-bar.tsx
sderacy: For sure! I'll look into how we can make this cleaner

src/tabs/tab-header-bar.tsx
sderacy: This might be moreso a question regarding the implementation but the motivation was this: https://quip-amazon.com/GeTZAiyGeGYS/Tabs-with-Actions#temp:C:EYX9a272942d788438abdc7084b5

I looked into using `role='none` but since it's a synonym for the presentaton role, I didn't think either were a good fit. We do want the button to be interactable but we don't want to have an explicit role for it.

I'm open to suggestion though of course

src/tabs/__integ__/tabs.test.ts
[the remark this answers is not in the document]
sderacy: It was discussed initially to enforce a component limitation on removing the single tab. The user would own the actual removal process so we're preventing the onDismiss callback

src/tabs/tab-header-bar.tsx
sderacy: This may be a question moreso on the implementation but the A11y review advised us to remove any explicit role on this button. I can see if I can change this conditional to something like `!hasActionOrDismissible && role: 'tab'`

sderacy:
Currently the fix for the non a11y bug found is not present here. I will need to add this fix sometime today

pages/tabs/permutations.page.tsx
timogasda, line 1: For the permutations, I would suggest that you don't touch the existing permutation sets and instead create a new one with dismiss and actions. This has two advantages: 1) you can focus on only meaningful permutations and the page doesnt grow as much, and 2) it's easier to compare the before and after of the first few permutations that dont have any actions.

src/tabs/interfaces.ts
```diff
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 import { BaseComponentProps } from '../internal/base-component';
 import { NonCancelableEventHandler } from '../internal/events';
+import { ButtonProps } from '../button/interfaces';
 
 export interface TabsProps extends BaseComponentProps {
```
timogasda, line 7: This is missing the addition to `i18nStrings`

src/tabs/tab-header-bar.tsx
```diff
@@ -64,6 +71,20 @@ export function TabHeaderBar({
   const [horizontalOverflow, setHorizontalOverflow] = useState(false);
   const [inlineStartOverflow, setInlineStartOverflow] = useState(false);
   const [inlineEndOverflow, setInlineEndOverflow] = useState(false);
+  const [previousActiveTabId, setPreviousActiveTabId] = useState<string | undefined>(activeTabId);
+  const hasActionOrDismissible = tabs.some(tab => tab.action || tab.dismissible);
+  const tabsWithActionsAriaRoleDescription = 'Tabs with Actions';
```
timogasda, line 76: There shouldn't be a default string here. Instead it should be using the prop from `i18nStrings`

src/tabs/tab-header-bar.tsx
```diff
@@ -292,16 +348,41 @@ export function TabHeaderBar({
...
+    const tabActionClasses = clsx({
+      [styles['tabs-tab-action']]: true,
+      [styles['tabs-tab-active']]: activeTabId === tab.id && !tab.disabled,
+    });
+
     const commonProps: (JSX.IntrinsicElements['a'] | JSX.IntrinsicElements['button']) & { 'data-testid': string } = {
       className: classes,
-      role: 'tab',
-      'aria-selected': tab.id === activeTabId,
+      // role: 'tab',
```
timogasda, line 365: remove comment

src/tabs/interfaces.ts
```diff
@@ -120,5 +147,10 @@ export namespace TabsProps {
      * ARIA label for the scroll right button that appears when the tab header is wider than the container.
      */
     scrollRightAriaLabel?: string;
+    /**
+     * ARIA role description for the Tabs component when an action or dismissible prop is in use. This is used
+     * with role="application" to provide further information on the purpose of this component
+     */
+    tabsWithActionsAriaRoleDescription?: string;
```
timogasda, line 154: This needs to be wrapped into an `i18nStrings` prop like we do in other components.

src/tabs/interfaces.ts
```diff
@@ -11,6 +12,14 @@ export interface TabsProps extends BaseComponentProps {
    * - `label` (ReactNode) - Tab label shown in the UI.
    * - `content` (ReactNode) - (Optional) Tab content to render in the container.
    * - `disabled` (boolean) - (Optional) Specifies if this tab is disabled.
+   * - `dismissible` (boolean) - (Optional) Determines whether the tab includes a dismiss icon button. By default, the dismiss button is not included.
+   * - `dismissLabel` (boolean) - (Optional) Specifies an aria-label for the dismiss icon button.
+   *    When a user clicks on this button the onDismiss handler is called.
```
gethinwebster, line 17: I think this belongs to `dismissible` not to `dismissLabel`? (or could even be skipped entirely, I think it's fairly self-explanatory)

src/tabs/interfaces.ts
```diff
@@ -11,6 +12,14 @@ export interface TabsProps extends BaseComponentProps {
...
+   * - `dismissDisabled` (boolean) - (Optional) Determines whether the dismissible button is disabled. We recommend using this only when the dismissible is present
```
gethinwebster, line 18: I don't think we need the bit about recommendation: I think it's pretty clear that it's only relevant when the dismiss button is present. And suggest "dismiss button" for consistency.
```suggestion
   * - `dismissDisabled` (boolean) - (Optional) Determines whether the dismiss button is disabled.
```

src/tabs/interfaces.ts
```diff
@@ -90,6 +99,29 @@ export namespace TabsProps {
...
+     * (Optional) Determines whether the tab includes a dismiss icon button. By default, the dismiss button is not included.
+     * When a user clicks on this button the onDismiss handler is called.
+     */
+    dismissible?: boolean;
+    /**
+     * (Optional) Specifies an aria-label for the dismiss icon button.
+     */
+    dismissLabel?: string;
+    /**
+     * (Optional) Determines whether the dismissible button is disabled. We recommend using this only when the dismissible is present
```
gethinwebster, line 112:
```suggestion
     * (Optional) Determines whether the dismiss button is disabled.
```

gethinwebster (changes requested):

src/tabs/interfaces.ts
sderacy: Ah yea that looks like a typo - I removed it in the next commit since I agree it's self-explanatory + is described elsewhere in our guidelines / implementation

src/tabs/interfaces.ts
sderacy: That makes sense -  good catch on the wording for `dismiss`

src/tabs/interfaces.ts
sderacy: Done!

gethinwebster (approved):

PR #2174, opened 2024-04-18
chore: Add RTL support for the AppLayout, Drawers, and SplitPanel components.

DaemonCahill:
This PR adds RTL support for the `AppLayout`, `Drawers`, and `SplitPanel` components.


src/internal/direction.ts
```diff
@@ -1,8 +1,20 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-export function isRtl(element: HTMLElement | SVGElement) {
-  return getComputedStyle(element).direction === 'rtl';
+export type IsRtl = boolean;
```
just-boris, line 4: Typescript uses structural typing, this type alias does not do anything useful

src/internal/direction.ts
```diff
@@ -1,8 +1,20 @@
...
+export type ElementWithDirection = Document | HTMLElement | SVGElement;
+
+export function getIsRtl(element: ElementWithDirection) {
+  if (element instanceof Document) {
+    return document.documentElement.dir === 'rtl';
+  } else {
```
just-boris, line 10: Why is this change happening in this PR?

src/internal/direction.ts
```diff
@@ -1,8 +1,20 @@
...
+    return getComputedStyle(element).direction === 'rtl';
+  }
+}
+
+export function getOffsetInlineStart(element: HTMLElement) {
+  const offsetParentWidth = element.offsetParent?.clientWidth ?? 0;
+  return getIsRtl(element) ? offsetParentWidth - element.clientWidth - element.offsetLeft : element.offsetLeft;
```
just-boris, line 17: Why is this `clientWidth` and not `offsetWidth`?

src/split-panel/__tests__/utils.test.tsx
```diff
@@ -15,68 +16,141 @@ describe('use-keyboard-events', () => {
   });
 
   test('bottom position, up key', () => {
-    const onKeyDown = useKeyboardEvents(sizeControlProps);
-    const event = new KeyboardEvent('keydown', { keyCode: KeyCode.up });
-    onKeyDown(event as any);
+    const onKeyDown = useKeyboardEvents({ ...sizeControlProps });
+    const div = document.createElement('div');
+    div.addEventListener('keydown', event => onKeyDown(event as any));
+    fireEvent.keyDown(div, { keyCode: KeyCode.up });
```
just-boris, line 22: After copy-pasting these lines 12 times, don't you think this could be reused somehow?

src/table/resizer/index.tsx
```diff
@@ -9,7 +9,7 @@ import { useStableCallback } from '@cloudscape-design/component-toolkit/internal
 import { useUniqueId } from '../../internal/hooks/use-unique-id';
 import { getHeaderWidth, getResizerElements } from './resizer-lookup';
 import { useSingleTabStopNavigation } from '../../internal/context/single-tab-stop-navigation-context.js';
-import { isRtl, getLogicalBoundingClientRect, getLogicalPageX } from '../../internal/direction.js';
+import { getIsRtl, getLogicalBoundingClientRect, getLogicalPageX } from '../../internal/direction.js';
```
just-boris, line 12: Changes in these components are not in the scope of this PR. Should go separately.

src/internal/direction.ts
DaemonCahill: I can use offsetWidth for the element but the offsetParent is typed as Element and will only allow for clientWidth.

src/internal/direction.ts
```diff
@@ -15,6 +20,15 @@ export function getScrollInlineStart(element: HTMLElement) {
   return isRtl(element) ? Math.floor(element.scrollLeft) * -1 : Math.ceil(element.scrollLeft);
 }
 
+/**
+ * The clientX position needs to be converted so it is relative to the right of
+ * the document in order for computations to yield the same result in both
+ * element directions.
+ */
+export function getLogicalClientX(event: PointerEvent, IsRtl: boolean) {
```
just-boris, line 28: If you want to separate `isRtl` (a function) and this argument, you could call it `rtlState` or `currentRtl`.

P.S. renaming `isRtl` -> `getIsRtl` still makes sense, just not in this PR. Are you planning to do that still?

src/internal/direction.ts
just-boris: These values are not equivalent, they have different use-cases.

Which one do we need to make it work correctly?

--

P.S. Since this function is used only once and this whole usage is a hack, can you think of a better way to refactor this whole piece in RTL-friendly way?

src/internal/direction.ts
DaemonCahill: I'm still planning to rename it in a separate PR.

src/internal/direction.ts
DaemonCahill: `offsetWidth` is the more accurate measurement because it is inclusive of borders and margins. In the `mainOffsetLeft` use case they are computing the same, but `offsetWidth` will be more resilient when used in max width computations. As far alternatives, I have the same issue with `offsetLeft` as I do with `getBoundingClientRect` in that they are absolute values from the physical x position regardless of document direction. If you have other ideas please share them.

just-boris: Your bet, but I would not introduce a reusable function for that and just do that calculation inline.

Reusable function tells other developers "here is a working implementation, feel free to use without worrying about internals". Can you say this about the function?

DaemonCahill: I have at least one more occurrence that I know of which will require the same computation. I'm going to keep the utility function for now.

just-boris (approved):

src/internal/direction.ts
just-boris: The question if this function is really reusable by anyone without deep understanding of the matter, still stays

PR #2193, opened 2024-04-24
chore: Reintroduce bar chart stack rounding

pan-kot:
### Description

Reintroducing https://github.com/cloudscape-design/components/pull/2171 that was reverted due to performance impact. The new solution uses SVG path to draw rectangles with corners rounded at one side only. That causes no performance impact (it actually improves stats!) compared to the clip path-based solution.

### How has this been tested?

* Screenshot tests;
* Lighthouse tests.

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/mixed-line-bar-chart/bar-series.tsx
```diff
@@ -67,23 +77,30 @@ export default function BarSeries<T extends ChartDataTypes>({
...
       let yValue = d.y;
-
-      if (stackedBarOffsets) {
-        // Stacked bars
-        const offsetMap = d.y < 0 ? stackedBarOffsets.negativeOffsets : stackedBarOffsets.positiveOffsets;
-        yValue = d.y + (offsetMap.get(getKeyValue(d.x)) || 0);
-      } else if (totalSeriesCount > 1) {
-        // Regular grouped bars
+      let isMin = !isStacked;
+      let isMax = !isStacked;
```
just-boris, line 92: If there are multiple isMin or isMax for a single row, this is confusing. 
In this case, `isRoundedStart` `isRoundedEnd` is a better name.

src/mixed-line-bar-chart/bar-series.tsx
```diff
@@ -110,46 +129,86 @@ export default function BarSeries<T extends ChartDataTypes>({
...
-        );
+        if (isMin && isMax) {
+          return <rect key={i} {...placement} {...styleProps} rx={rx} />;
+        }
+        if (!isMin && !isMax) {
+          return <rect key={i} {...placement} {...styleProps} rx={0} />;
+        }
+        const side =
+          !isVertical && !isMax ? 'left' : !isVertical && isMax ? 'right' : isVertical && isMax ? 'top' : 'bottom';
+        return <path key={i} d={createOneSideRoundedRectPath(placement, rx, side)} {...styleProps} {...styleProps} />;
```
just-boris, line 171: Do I miss anything or you just repeat `{...styleProps}` twice?

src/mixed-line-bar-chart/bar-series.tsx
```diff
@@ -110,46 +129,86 @@ export default function BarSeries<T extends ChartDataTypes>({
...
-            height={width - widthOffset}
-            rx={rx}
-            className={className}
-          />
-        );
+        if (isMin && isMax) {
+          return <rect key={i} {...placement} {...styleProps} rx={rx} />;
+        }
+        if (!isMin && !isMax) {
+          return <rect key={i} {...placement} {...styleProps} rx={0} />;
```
just-boris, line 167: Not covered by tests? Why?

src/mixed-line-bar-chart/bar-series.tsx
```diff
@@ -110,46 +129,86 @@ export default function BarSeries<T extends ChartDataTypes>({
...
+        }
+        const side =
+          !isVertical && !isMax ? 'left' : !isVertical && isMax ? 'right' : isVertical && isMax ? 'top' : 'bottom';
+        return <path key={i} d={createOneSideRoundedRectPath(placement, rx, side)} {...styleProps} {...styleProps} />;
       })}
     </g>
   );
 }
+
+function createOneSideRoundedRectPath(
```
just-boris, line 177: 1. Don't you think this code deserves its own unit tests?
2. Or moveover, I think I found the source: https://github.com/honzaap/Systemizer/blob/59f37ae71993d9a28997ccb4b9e81edc4e4e7f8d/src/shared/ExtensionMethods.ts#L105-L123

Don't you want to give it some attribution or find a way to load this as a dependency (does not have to be this package, but maybe there is something)?

src/mixed-line-bar-chart/bar-series.tsx
pan-kot: This part is well covered with screenshot tests. The unit tests do not seem very suitable in this case.

src/mixed-line-bar-chart/bar-series.tsx
pan-kot: I took this from this stack overflow answer actually: https://stackoverflow.com/a/65186378/16648714

just-boris: If a code snippet is large enough, we are required to annotate this. Examples:

* https://github.com/cloudscape-design/components/blob/main/pages/utils/pseudo-random.ts
* https://github.com/cloudscape-design/components/blob/main/src/internal/utils/browser-scrollbar-size.ts

After annotating this, how can you share this information more clearly so it will not depend on me spotting it?

just-boris: The question about writing some standalone unit tests also still stands

src/mixed-line-bar-chart/bar-series.tsx
just-boris: 😕 this code is a pure math, converting the input data into SVG offsets. See no reason why it is not suitable.

But because this is one line only, I would leave the burden of this tech debt on you to decide

src/mixed-line-bar-chart/bar-series.tsx
pan-kot: It is my bad of not sharing this

src/mixed-line-bar-chart/utils.ts
```diff
@@ -130,3 +114,43 @@ export function isDataSeries<T>(
 ): series is MixedLineBarChartProps.DataSeries<T> {
   return series.type === 'line' || series.type === 'bar';
 }
+
+// Source https://stackoverflow.com/a/65186378/16648714
+export function createOneSideRoundedRectPath(
```
just-boris, line 119: The external code should live in a separate file without our standard header: "Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved."

Because technically it is not ours

See the example I shared before: https://github.com/cloudscape-design/components/blob/main/pages/utils/pseudo-random.ts

just-boris (approved):

PR #2165, opened 2024-04-17
feat: Table renderAriaLive receives visibleItemsCount

pan-kot:
### Description

Adding `visibleItemsCount` argument to table's `renderAriaLive` so that the table can announce the difference in visible items when an expandable group is expanded or collapsed. For example:

1. Table renders - `displaying clusters 1 to 10 of 35, 10 resources visible`
2. User expands a group - `displaying clusters 1 to 10 of 35, 13 resources visible`
3. User loads more items for the group - `displaying clusters 1 to 10 of 35, 16 resources visible`
4. User changes filter - `displaying clusters 1 to 10 of 17, 12 resources visible`
5. User changes page - `displaying clusters 11 to 17 of 17, 7 resources visible`

### How has this been tested?

* Unit tests
* Dry-run

<details>
   <summary>Review checklist</summary>

_The following items are to be evaluated by the author(s) and the reviewer(s)._

#### Correctness

- _Changes include appropriate documentation updates._
- _Changes are backward-compatible if not indicated, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#public-apis)._
- _Changes do not include unsupported browser features, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#browsers-support)._
- _Changes were manually tested for accessibility, see [accessibility guidelines](https://cloudscape.design/foundation/core-principles/accessibility/)._

#### Security

- _If the code handles URLs: all URLs are validated through [the `checkSafeUrl` function](https://github.com/cloudscape-design/components/blob/main/src/internal/utils/check-safe-url.ts)._

#### Testing

- _Changes are covered with new/existing unit tests?_
- _Changes are covered with new/existing integration tests?_
</details>

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2203, opened 2024-04-26
chore: Table progressive loading (no public API)

pan-kot:
### Description

The PR introduces progressive loading feature for tables but without any public API updates yet.

The feature is added in its simplest form: the extra rows are added without col-span which is a debatable design choice and is challenging to introduce due to various design and accessibility considerations.

The PR will be merged as soon it is approved. The feature will be released upon its readiness and it will require a follow-up PR to update the public API.

Rel: sEQSANE2RASR

### How has this been tested?

* New unit tests
* New integration tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/table/interfaces.tsx
```diff
@@ -466,6 +466,7 @@ export namespace TableProps {
   export interface LiveAnnouncement {
     firstIndex: number;
     lastIndex: number;
+    visibleItemsCount: number;
```
jperals, line 469: This is a change to the public API, and this new prop is not optional, so not a backwards-compatible change? This is used by `renderAliaLive`.

pages/table/expandable-rows-test.page.tsx
```diff
@@ -225,22 +296,94 @@ function useTableData() {
...
+    setLoadingState(nextLoading(id));
+    if (delay) {
+      setTimeout(() => setLoadingState(settings.emulateServerError ? nextError(id) : nextPending(id)), delay);
+    } else {
+      setLoadingState(nextPending(id));
+    }
+  };
+
+  const getItemChildren = collectionResult.collectionProps.expandableRows
+    ? collectionResult.collectionProps.expandableRows.getItemChildren.bind(null)
```
jperals, line 339: Why do we need to bind to null?

jperals:
> The feature is added in its simplest form: the extra rows are added without col-span which is a debatable design choice and is challenging to introduce due to various design and accessibility considerations

Can you elaborate on these challenges? It seems to me like the UI has suffered quite a bit here:

Screenshot 2024-04-29 at 15 00 28

src/table/interfaces.tsx
pan-kot: Thanks, that is indeed an API change! I will remove it from the current PR, it is meant to be released separately: https://github.com/cloudscape-design/components/pull/2165

pan-kot:
> > The feature is added in its simplest form: the extra rows are added without col-span which is a debatable design choice and is challenging to introduce due to various design and accessibility considerations
> 
> Can you elaborate on these challenges? It seems to me like the UI has suffered quite a bit here:
> 
> Screenshot 2024-04-29 at 15 00 28

The idea here is to add the loader to the first cell only instead of making the first cell span to the entire row. It has a disadvantage of not utilising the available empty space efficiently, but there are advantages as well:
1. It works naturally with sticky columns;
2. We can extend it by allowing custom content in other cells (e.g. we can make the loader row selectable when multi-selection) and allow populating other cells with aggregated values for those not loaded items.

Besides, that approach requires a far simpler implementation. Should we consider the col-span version, I will open a follow-up PR which will be then better straightforward to review.

pan-kot:
> Actually even if using colspan, the loader row could stick (i.e, not be affected by horizontal scroll, and probably wrap if needed) unconditionally?

If using colspan we can make use of all available space so it won't wrap unless the table itself is very narrow. Yes, we can make it sticky too, this actually works this way here: https://github.com/cloudscape-design/components/pull/2131

> OK but these are hypothetical future improvements. I would like to make sure the component is in good shape now regardless of those.

We plan to keep the current behaviour for nested row loaders. The content of the loader might wrap or not wrap depending on the wrapLines property of the table. The loader content must not be too long - it should be designed to fit the first column or at least to not wrap for too many lines.

> Could you double-check with visual design if you haven't already to make sure this is OK?

Yes, I checked already. The current plan is to keep the behaviour as is for the child rows but introduce the colSpan and make the content centered for the root loader. However, this treatment can be revisited again, that's why the current PR features the simple-most approach.

src/table/interfaces.tsx
jperals: Are you still aiming to remove this from this PR? Or rather preferring to merge #2165 first now?

pan-kot: I will merge 2165 first, it is blocked by the dry run atm.

pan-kot: @jperals updated

pages/table/expandable-rows-test.page.tsx
```diff
@@ -228,22 +295,94 @@ function useTableData() {
...
   useEffect(() => {
     setLoading(true);
+    setError(false);
     const timeoutId = setTimeout(() => {
       setLoading(false);
       setReadyItems(memoItems);
+      setError(settings.emulateServerError);
     }, delay);
     return () => clearTimeout(timeoutId);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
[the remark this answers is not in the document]
pan-kot: Right, will add this and other set-state functions.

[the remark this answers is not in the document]
pan-kot: @jperals done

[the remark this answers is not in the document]
pan-kot: Yes, that is due to `settings.emulateServerError`. A change to this setting should not cause the hook to fire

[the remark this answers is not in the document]
pan-kot: It allows to e.g. load the data successfully and then change the flag to false before expanding the first group - that would set its children to error. They you can change back to true and try again.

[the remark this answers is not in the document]
pan-kot: If the flag is added to the dependency array it triggers the effect every time so you get either errors everywhere or no errors at all.

jperals (approved):

PR #2209, opened 2024-04-30
feat: Add wrapTriggerText to popover

avinashbot:
### Description

Added a `wrapTriggerText` property.

Follows the "negative" prop convention we essentially established for button and status indicator components. It has the name `trigger` in it to both follow the pattern with the other props (`triggerType`, `triggerAriaLabel`), and make it clear that this doesn't apply to the body itself. Setting this prop to `false` adds both text wrapping and truncation to the text trigger, and while I messed around with making it default, I've seen enough popovers with "internal" wrapping to realize this is a bad idea (plus it would make it completely unreadable/inaccessible for most cases outside a resizable table).

#### API changes

```typescript
/**
 * Specifies if the text trigger content should wrap. If you set it to false, it prevents the text from
 * wrapping and truncates it with an ellipsis.
 */
wrapTriggerText?: boolean;
```

Copied almost verbatim from the status indicator (except I added the word "trigger").

Related links, issue #, if available: AWSUI-38391

### How has this been tested?

Added permutation tests. Updated snapshots.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/popover/index.tsx
+++ b/src/popover/index.tsx
@@ -18,6 +18,7 @@ export default function Popover({
   triggerType = 'text',
   dismissButton = true,
   renderWithPortal = false,
+  wrapTriggerText = true,
   header,
   ...rest
 }: PopoverProps) {
@@ -40,6 +41,7 @@ export default function Popover({
       triggerType={triggerType}
       dismissButton={dismissButton}
       renderWithPortal={renderWithPortal}
+      wrapTriggerText={wrapTriggerText}
       {...externalProps}
       {...baseComponentProps}
     />
--- a/src/popover/internal.tsx
+++ b/src/popover/internal.tsx
@@ -47,6 +47,7 @@ function InternalPopover(
     content,
     triggerAriaLabel,
 
+    wrapTriggerText = true,
     renderWithPortal = false,
 
     __onOpen,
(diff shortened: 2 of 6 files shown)


src/popover/internal.tsx
```diff
@@ -187,13 +188,14 @@ function InternalPopover(
...
         <button
           {...triggerProps}
+          className={clsx(triggerProps.className, wrapTriggerText === false && styles['overflow-ellipsis'])}
           tabIndex={triggerTabIndex}
           type="button"
           aria-haspopup="dialog"
           id={referrerId}
           aria-label={triggerAriaLabel}
         >
-          <span className={styles['trigger-inner-text']}>{children}</span>
```
avinashbot, line 196: > Prevents trigger from shifting on click in IE11.

I think IE11 users can bear with a little shift when they click on the popover now, if the page even loads for them in the first place...

just-boris:
As discussed on the offline meeting, let's see if we can do `wrapTriggerText="inherit"` to let this component inherit the behavior from the parent (most likely, our table)

src/popover/index.tsx
```diff
@@ -18,6 +18,7 @@ export default function Popover({
   triggerType = 'text',
   dismissButton = true,
   renderWithPortal = false,
+  wrapTriggerText = true,
```
cansuaa, line 21: Could making it default cause any issues for the customers?

avinashbot: It's a little confusing because it's actually changing to `false` that causes changes in behavior. This isn't normally how we do API properties, but this is the one weird exception, for consistency reasons.

cansuaa (approved):

PR #2217, opened 2024-04-30
chore: Use design tokens for shared panel styles

jkuelz:
### Description

This commit is part of an effort to unify the styles of panel components (e.g. drawers, side nav, help panel, split panel).
It also fixes an inconsistency with the divider widths and help panel footer content alignment.  

### How has this been tested?

Ran through dev pipeline.
Everything should be visually backward compatible with the exception of the divider indentation fix, which is intended.

<details>

#### Correctness

#### Security

#### Testing

</details>

---


style-dictionary/classic/spacing.ts
```diff
@@ -40,6 +40,10 @@ const tokens: StyleDictionary.SpacingDictionary = {
   spaceLayoutContentHorizontal: '{spaceScaled2xXxxl}',
   spaceModalContentBottom: '{spaceScaledL}',
   spaceOptionIconBigTop: '{spaceXxxs}',
+  spacePanelContentBottom: '{spaceScaledXxxl}',
```
jperals, line 43: Classic takes the Visual Refresh values as defaults (this is done with `merge` below), so I think only the tokens that differ from Visual Refresh need to be specified here.

style-dictionary/classic/colors.ts
```diff
@@ -71,6 +71,7 @@ const tokens: StyleDictionary.ColorsDictionary = {
   colorBorderItemFocused: { light: '{colorBlue600}' },
   colorBorderItemPlaceholder: '{colorTransparent}',
   colorBorderNotificationStackBar: '{colorGrey700}',
+  colorBorderPanelHeader: '{colorBorderDividerDefault}',
```
jperals, line 74: Same as the comment below: I think adding this one here is unnecessary.

style-dictionary/utils/token-names.ts
```diff
@@ -630,6 +631,10 @@ export type SpacingTokenName =
   | 'spaceModalContentBottom'
   | 'spaceModalHorizontal'
   | 'spaceOptionIconBigTop'
+  | 'spacePanelContentBottom'
+  | 'spacePanelContentTop'
+  | 'spacePanelDividerMarginHorizontal'
```
jperals, line 636: Any specific reason why this token contains the word `Margin` and the others don't? Usually we don't include this word in tokens, and it used for paddings too, so could it be `spacePanelDividerHorizontal` instead?

src/help-panel/styles.scss
```diff
@@ -161,9 +161,9 @@
   color: awsui.$color-text-body-secondary;
   border-block-start: awsui.$border-divider-section-width solid awsui.$color-border-divider-default;
   margin-block: awsui.$space-scaled-xl;
-  margin-inline: calc(-1 * #{awsui.$space-xs});
+  margin-inline: calc(-1 * #{awsui.$space-panel-divider-margin-horizontal});
   padding-block: 0;
-  padding-inline: awsui.$space-s;
+  padding-inline: awsui.$space-panel-divider-margin-horizontal;
```
jperals, line 166: Inline margin and inline padding now have the same amount but with alternating signs. Can they be removed as they cancel each other?

jperals:
> Everything should be visually backward compatible with the exception of the divider indentation fix, which is intended.

The BackstopJs tests don't reveal any diff regarding the dividers but instead they show that the indentation of the help panel footer decreased. Is this the intention?

style-dictionary/classic/spacing.ts
jkuelz: While that's true, these might change in the future (I have another PR prepared where this is the case). So I think a more explicit approach is also fine.

src/help-panel/styles.scss
jkuelz: The margin is being used as negative margin to extend the divider beyond the text. I think this can be improved to not use negative margin in the future, but that refactoring is not the goal of this PR.

Screenshot 2024-05-02 at 8 25 42 AM

jkuelz:
> > Everything should be visually backward compatible with the exception of the divider indentation fix, which is intended.
> 
> The BackstopJs tests don't reveal any diff regarding the dividers but instead they show that the indentation of the help panel footer decreased. Is this the intention? @jperals 

Yes. There are two expected changes. 
1. The help panel footer indentation fix:
Before 
Screenshot 2024-05-02 at 8 34 05 AM
After
Screenshot 2024-05-02 at 8 25 54 AM

2. The other fix is that the side navigation divider width is now the same as the help panel divider width. These were mismatched and have different values in classic and VR. In VR, this results in a change to the side nav divider width, in classic this results in a change to the help panel divider width:
Before
Screenshot 2024-05-02 at 8 43 18 AM
After
Screenshot 2024-05-02 at 8 44 36 AM

jperals:
> In VR, this results in a change to the side nav divider width

Shouldn't BackstopJS have caught a diff in the side nav permutations page then?

jperals (approved):

PR #2220, opened 2024-05-02
feat: Add warning styling to Slider

cansuaa:
### Description

Slider is recently released using FormFieldValidationControlProps which with warning validations includes warning API but the component doesn't have the styling for it yet. This PR addresses this problem.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


pages/slider/permutations.page.tsx
```diff
@@ -17,6 +17,26 @@ const permutations = createPermutations<SliderProps>([
...
+    min: [-10],
+    max: [10],
+    hideFillLine: [false, true],
+    disabled: [false],
+    tickMarks: [false, true],
+    warning: [true],
+    step: [0.5, 1, 5],
+  },
+  {
+    value: [0],
```
katiegeorge, line 31: It seems like this is doing the same as the above. Can you combine them and add `disabled: [false, true]` to the top one?

src/slider/styles.scss
```diff
@@ -173,10 +181,18 @@
...
 
+  &.warning {
+    background: awsui.$color-text-status-warning;
+  }
+
   &.error-active {
     background-color: awsui.$color-background-slider-error-pressed;
   }
 
+  &.warning-active {
```
katiegeorge, line 192: Pressing the slider doesn't seem to be changing the color. It looks like there is supposed to be two different colors for active and non-active warning, is that a bug?

pages/slider/permutations.page.tsx
cansuaa: The one below tests the warning slider with disabled setting, combining them exceeds the page limitation that's why I separated this part.

katiegeorge (approved):
Will resolve pressed warning color discussion in a separate PR. Looks great!

Issue #2181, opened 2024-04-21
[Bug]: Autosuggest's `focus` function does not always work


drewdh:
### Browser

Chrome

### Package version

3.0.615

### React version

18.2.0

### Description

On initial mount, calling the `focus` function of the autosuggest component, even though defined, has no effect. If the component is remounted, calling focus works as expected.

### Source code

See CodeSandbox below.

### Reproduction

https://codesandbox.io/p/sandbox/autosuggest-focus-bug-vdszv6?file=%2Fsrc%2FApp.tsx

1. Open the preview in a wide enough screen that the `search` slot content of the TopNavigation is visible on the first render.
2. Click the "Focus" button. See that the ref's focus method was called, but the component was not focused.
3. Reduce the width of the screen so the `search` slot content is hidden.
4. Widen the screen so the `search` slot content is visible again.
5. Click "Focus" and see the component was focused.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/cloudscape-design/components/blob/main/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/cloudscape-design/components/issues) for duplicate problems

drewdh:
Hi, I’m just checking in on this issue as it’s impacting our current project timeline. Is there any update, or is there any additional information I could provide to assist? I'm also happy to help with debugging or testing if needed.

pan-kot:
Hello,

Apologies for the late reply and many thanks for providing a reproducible example. No additional information is needed - our team will take a look at the problem and keep you posted.

gethinwebster:
A fix for this has now been released

PR #2225, opened 2024-05-03
fix: Make inputs inside top navigation programmatically focusable

pan-kot:
### Description

In Top navigation all content is rendered twice. The second time the content is rendered as "virtual" - it is invisible and is used to measure content elements widths. That creates potential problems, for example, using the `id` attribute on a content node would result in duplicate IDs present in the DOM. A similar issue was with React refs. The ref callback is called twice and the actual node it gets assigned is the one which comes from the callback executed the last. That means the virtual content must come before the actual content and it can't be rendered in our Portal because the portal content is rendered with a delay.

Rel: AWSUI-42554, https://github.com/cloudscape-design/components/issues/2181

### How has this been tested?

* New integration test

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/autosuggest/simple.page.tsx
+++ b/pages/autosuggest/simple.page.tsx
@@ -1,7 +1,7 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
-import React, { useState } from 'react';
-import Autosuggest from '~components/autosuggest';
+import React, { useRef, useState } from 'react';
+import Autosuggest, { AutosuggestProps } from '~components/autosuggest';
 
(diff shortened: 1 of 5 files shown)


pan-kot:
> Oh, huh. What are the chances you think it'll break selectors in use on customer tests?

Hard to tell. The changes are non-zero for sure.

If a selector uses `wrapper.findTopNavigation()` - it won't get broken, because we exclude the virtual part deliberately. However, if a selector just searches for the component like `wrapper.findInput()`, assuming the search input is the first Cloudscape input in the page - that will match the virtual input instead. Previously it matched the correct one because the virtual was rendered after.

I will run a dry-run.

PR #2230, opened 2024-05-08
feat: Add loop arrow key navigation in autosuggest

georgylobko:
### Description

Currently, users can navigate through suggestion list using the keyboard up/down arrow keys. However, when they reach the last position in the suggestion list, pressing the down arrow key does not loop to the first position of the list (and vice versa). Users must repeatedly press the arrow keys to return to the top of the list. This PR addressed this problem.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/autosuggest/internal.tsx
+++ b/src/autosuggest/internal.tsx
@@ -139,10 +139,23 @@ const InternalAutosuggest = React.forwardRef((props: InternalAutosuggestProps, r
   };
 
   const handlePressArrowDown = () => {
+    if (autosuggestItemsState.items.length - 1 === autosuggestItemsState.highlightedIndex) {
+      autosuggestItemsHandlers.goHomeWithKeyboard();
+      return;
+    }
+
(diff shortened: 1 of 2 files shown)


src/autosuggest/internal.tsx
```diff
@@ -139,10 +139,23 @@ const InternalAutosuggest = React.forwardRef((props: InternalAutosuggestProps, r
   };
 
   const handlePressArrowDown = () => {
+    if (autosuggestItemsState.items.length - 1 === autosuggestItemsState.highlightedIndex) {
```
just-boris, line 142: Why did you implement it here and not inside `autosuggestItemsHandlers.moveHighlightWithKeyboard`?

We have some shared dropdown code for all components. It is possible to implement it here and cover ButtonDropdown too? (currently it does not get the chage)

just-boris: Okay, got this. Sounds good by now

just-boris (approved):

PR #2232, opened 2024-05-08
chore: Migrate pages to logical properties.

DaemonCahill:
This PR migrates the pages directory to logical properties and adds a stylelint rule to enforce future usage.

### DIFF
--- a/pages/table/sticky-header-with-actions.page.tsx
+++ b/pages/table/sticky-header-with-actions.page.tsx
@@ -23,10 +23,10 @@ export default function () {
       <h1>Table with a sticky header and actions</h1>
       <ScreenshotArea>
         <div
-          style={{ height: '400px', inlineSize: '500px', overflow: 'auto', padding: '0px 1px' }}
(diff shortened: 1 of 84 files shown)


pages/table/sticky-header-with-actions.page.tsx
```diff
@@ -22,8 +22,11 @@ export default function () {
     <>
       <h1>Table with a sticky header and actions</h1>
       <ScreenshotArea>
-        <div style={{ height: '400px', width: '500px', overflow: 'auto', padding: '0px 1px' }} id="scroll-container">
-          <div style={{ height: '100px' }} />
+        <div
+          style={{ blockSize: '400px', inlineSize: '500px', overflow: 'auto', padding: '0px 1px' }}
```
connorlanigan, line 26: Should we update the `padding` here too?

connorlanigan (changes requested):

pages/table/sticky-header-with-actions.page.tsx
DaemonCahill: Good catch! Updating...

connorlanigan (approved):

PR #2237, opened 2024-05-13
feat: Avatar component

cansuaa:
### Description
Avatar component. DRAFT.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/__integ__/__snapshots__/themes.test.ts.snap
```diff
@@ -22,6 +22,7 @@ Object {
   "border-panel-header-width": "0px",
   "border-panel-top-width": "0px",
   "border-radius-alert": "2px",
+  "border-radius-avatar": "50%",
```
pan-kot, line 25: do we really need this to be a token?

src/avatar/__tests__/avatar.test.tsx
```diff
@@ -0,0 +1,75 @@
...
+function renderAvatar(props: AvatarProps) {
+  const { container } = render(<Avatar {...props} />);
+
+  return createWrapper(container).findAvatar()!;
+}
+
+describe('Avatar component', () => {
+  describe('Basic', () => {
+    (['gen-ai', 'default'] as AvatarProps.Color[]).forEach(color => {
+      test(`renders ${color} color avatar`, () => {
```
pan-kot, line 20: I don't think we need to test against every color. It is enough to add a single test that verifies the color is correctly asigned or even no unit test for it at all because such feature is much better to assure with VR/screenshot tests.

src/avatar/interfaces.ts
```diff
@@ -0,0 +1,70 @@
...
+  /**
+   * Specifies the SVG of a custom icon.
+   *
+   * Use this property if the icon you want isn't available.
+   * If you set both `iconUrl` and `iconSvg`, `iconSvg` will take precedence.
+   */
+  iconSvg?: React.ReactNode;
+}
+
+export namespace AvatarProps {
```
pan-kot, line 68: As the component is focusable we might consider adding the `focus()` function to it same as we add for other focusable components like buttons. I think it is also fine to not introduce the function unless someone asks or unless we make the component clickable.

src/avatar/internal.tsx
```diff
@@ -0,0 +1,117 @@
...
+  __internalRootRef = null,
+  ...rest
+}: InternalAvatarProps) {
+  const baseProps = getBaseProps(rest);
+  const handleRef = useRef<HTMLDivElement>(null);
+  const [showTooltip, _setShowTooltip] = useState(false);
+
+  const setShowTooltip = (value: boolean) => {
+    if (tooltipText) {
+      return _setShowTooltip(value);
```
pan-kot, line 59: Alternatively, we can keep the default setShowTooltip but only render it when `showTooltip && tooltipText`

src/avatar/loading-dots/index.tsx
```diff
@@ -0,0 +1,21 @@
...
+import styles from './styles.css.js';
+import clsx from 'clsx';
+
+interface LoadingDotsProps {
+  color?: string;
+}
+
+export default function LoadingDots({ color }: LoadingDotsProps) {
+  return (
+    <div className={clsx(styles.root, { [styles['gen-ai']]: color === 'gen-ai' })}>
```
pan-kot, line 13: What is the need in having conditional gen-ai style? I thought the loading dots work the same for all colors

src/avatar/styles.scss
```diff
@@ -0,0 +1,45 @@
...
+
+.root {
+  @include styles.styles-reset;
+}
+
+.avatar {
+  display: inline-flex;
+  @include mixins.border-radius-avatar;
+  color: white;
+  block-size: 28px;
```
pan-kot, line 18: The 28px variable is repeating here and in the dancing dots styles. You can move it to a variable or maybe you can use width,height=100% for dancing dots wrapper.

src/avatar/__tests__/avatar.test.tsx
cansuaa: That is correct, I believe I forgot to re-visit this part after updating the API from type/variant to color. I'll do the change.

src/avatar/loading-dots/index.tsx
cansuaa: On top of animated dots, gen-ai background color also animates, you can see [here](https://d21d5uik3ws71m.cloudfront.net/components/ee0a7eadf7a10c26ed248036c485d2a19f1fb86d/dev-pages/index.html#/light/avatar/basic). The solid colors don't really make sense to be animated

src/__integ__/__snapshots__/themes.test.ts.snap
cansuaa: Not really, removed.

src/avatar/internal.tsx
cansuaa: Indeed. Updated.

src/avatar/styles.scss
cansuaa: Using width, height 100& caused disruption in the animation but I changed block and inline size to use inherit instead of 28px

src/avatar/loading-dots/index.tsx
pan-kot: Make sense, I would add a code comment to make this clear

src/avatar/__tests__/avatar.test.tsx
```diff
@@ -0,0 +1,106 @@
...
+}));
+afterEach(() => {
+  (warnOnce as jest.Mock).mockReset();
+});
+
+describe('Avatar component', () => {
+  describe('Basic', () => {
+    test('Renders avatar', () => {
+      const wrapper = renderAvatar({ ...defaultAvatarProps });
+      wrapper.focus();
```
pan-kot, line 30: Why this test is needed?

src/avatar/__tests__/avatar.test.tsx
```diff
@@ -0,0 +1,106 @@
...
+    });
+
+    test('Shows tooltip on hover', () => {
```
pan-kot, line 33: This one tests tooltip on focus, not hover

src/avatar/__tests__/avatar.test.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+  describe('a11y', () => {
+    test('Validates', () => {
+      const wrapper = renderAvatar({
+        color: 'default',
+        initials: 'JD',
+        tooltipText: 'Jane Doe',
+        ariaLabel: 'User avatar',
+      });
+      expect(wrapper.getElement()).toValidateA11y;
```
pan-kot, line 97: This line tests nothing as the method is not executed, it is also async so should be:
```
await  expect(wrapper.getElement()).toValidateA11y();
```

src/avatar/loading-dots/motion.scss
```diff
@@ -0,0 +1,87 @@
...
+    content: '';
+    position: absolute;
+    inline-size: inherit;
+    block-size: inherit;
+    inset-block-start: 50%;
+    inset-inline-start: 50%;
+    transform: translate(-50%, -50%);
+    background: awsui.$color-background-avatar-gen-ai;
+    @include styles.with-motion {
+      animation: gradientMove 3.6s infinite cubic-bezier(0.7, 0, 0.3, 1);
```
pan-kot, line 19: We normally take motion durations from tokens like motion-duration-complex, should we introduce a new one here?

pan-kot: I think one of the reasons is to support the reduced motion mode.

src/avatar/styles.scss
```diff
@@ -0,0 +1,42 @@
...
+    background-color: awsui.$color-background-avatar-default;
+  }
+
+  &.avatar-color-gen-ai {
+    background: awsui.$color-background-avatar-gen-ai;
+  }
+
+  &.initials {
+    font-family: awsui.$font-family-base;
+    font-size: 12px;
```
pan-kot, line 35: should we use tokens for font-size and line height?

src/avatar/styles.scss
```diff
@@ -0,0 +1,42 @@
...
+    line-height: 16px;
+  }
+
+  @include focus-visible.when-visible {
```
pan-kot, line 39: We should be able to use the existing `focus-highlight` mixin to draw the focus outline

PR #2238, opened 2024-05-13
refactor: Reorganise and document code in table use-selection util

pan-kot:
### Description

The refactoring is done to facilitate future planned support for grouped selection so that when a parent row is selected (in table with expandable rows) all child rows are automatically selected too. That different selection mechanism requires a different computation which can be substituted to the useSelection.

### How has this been tested?

Existing tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/table/selection/use-selection.ts
+++ b/src/table/selection/use-selection.ts
@@ -29,92 +29,85 @@ export function useSelection<T>({
   | 'onSelectionChange'
   | 'loading'
 >) {
-  const [shiftPressed, setShiftPressed] = useState(false);
-  const [lastClickedItem, setLastClickedItem] = useState<T | null>(null);
-  const selectionName = useUniqueId();
+  // The name assigned to all radio- controls to combine them in a single group.
+  const selectionControlName = useUniqueId();
+
+  // Selection state for individual items.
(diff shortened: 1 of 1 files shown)


src/table/selection/use-selection.ts
```diff
@@ -29,92 +29,85 @@ export function useSelection<T>({
...
+    const newSelectedItems = allEnabledItemsSelected ? deselectItems(items) : selectItems(items);
+    fireNonCancelableEvent(onSelectionChange, { selectedItems: newSelectedItems });
+  };
+
+  const handleToggleItem = (item: T) => {
+    if (isItemDisabled(item)) {
       return;
     }
-    if (selectionType === 'single') {
+    if (selectionType === 'single' && !isItemSelected(item)) {
```
jperals, line 105: Why was `&& !isItemSelected(item)` added to the condition?

Does `isItemSelected` return the selection state before or after toggling?

pan-kot: That is consistent with the previous line `if (disabled || (selectionType === 'single' && selected))`. The reason behind these conditions is to prevent state update is the user clicks on a disabled control or on a radio button which is already selected. The isItemSelected returns state valid for the current render cycle so before toggling.

jperals (approved):

PR #2242, opened 2024-05-13
feat: Constraint text for Property Filter

youdz:
### Description

This CR adds the ability to display a constraint text for the Property Filter. This can be used for server-side collections when the capabilities of the Search API do not allow every combination of filters offered by Property Filter: limited number of filter tokens, incompatible tokens, etc.

This change includes a new method to the `PropertyFilterWrapper` test utility, to find the new slot container for the constraint text.

See: mMxbAEGGaa5a

### How has this been tested?

1. Tested by updating the permutations page to display constraint text with and without tokens:

Screenshot 2024-05-13 at 17 12 10

2. Updated unit tests with a straightforward test to confirm the new test utility works and that the constraint text is correctly slotted.
3. Ran all unit, integ, a11y and motion tests. Updated documentation snapshots.
4. Ran visual regression testing. Since I updated the Property Filter permutations, the tests for that page failed and the snapshots will need to be updated. Before/after on the page:

https://github.com/cloudscape-design/components/assets/4647197/26b2a1d1-05df-42eb-a30c-214e0361850c

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/property-filter/index.tsx
+++ b/src/property-filter/index.tsx
@@ -37,6 +37,7 @@ import { PropertyFilterOperator } from '@cloudscape-design/collection-hooks';
 import { useInternalI18n } from '../i18n/context';
 import TokenList from '../internal/components/token-list';
 import { SearchResults } from '../text-filter/search-results';
+import { joinStrings } from '../internal/utils/strings';
 
 export { PropertyFilterProps };
 
@@ -94,6 +95,7 @@ const PropertyFilter = React.forwardRef(
       filteringFinishedText,
       filteringErrorText,
       filteringRecoveryText,
+      filteringConstraintText,
       filteringStatusType,
       asyncProperties,
       tokenLimit,
@@ -349,6 +351,10 @@ const PropertyFilter = React.forwardRef(
       parsedText.step === 'property' && parsedText.property.getValueFormRenderer(parsedText.operator);
 
     const searchResultsId = useUniqueId('property-filter-search-results');
+    const constraintTextId = useUniqueId('property-filter-constraint');
+    const textboxAriaDescribedBy = filteringConstraintText
+      ? joinStrings(rest.ariaDescribedby, constraintTextId)
+      : rest.ariaDescribedby;
 
     return (
       <div {...baseProps} className={clsx(baseProps.className, styles.root)} ref={__internalRootRef}>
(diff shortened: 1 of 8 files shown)


src/property-filter/interfaces.ts
```diff
@@ -184,6 +184,11 @@ export interface PropertyFilterProps extends BaseComponentProps, ExpandToViewpor
    * Use the `onLoadItems` event to perform a recovery action (for example, retrying the request).
    **/
   filteringRecoveryText?: string;
+  /**
+   * Constraint text that's displayed below the filtering input and above the tokens.
```
timogasda, line 188: I believe we agreed to drop the "above the tokens" part?

src/property-filter/index.tsx
```diff
@@ -404,6 +405,7 @@ const PropertyFilter = React.forwardRef(
             </div>
           ) : null}
         </div>
+        {filteringConstraintText && <div className={styles.constraint}>{filteringConstraintText}</div>}
```
timogasda, line 408: In a regular form field, constraint text is connected to the input using `aria-describedby`. We should probably do the same here.

timogasda (changes requested):

src/property-filter/index.tsx
youdz: Great call, I completely missed that. Fixed in the latest revision.

src/property-filter/index.tsx
```diff
@@ -361,7 +367,7 @@ const PropertyFilter = React.forwardRef(
             ariaLabel={filteringAriaLabel ?? i18nStrings.filteringAriaLabel}
             placeholder={filteringPlaceholder ?? i18nStrings.filteringPlaceholder}
             ariaLabelledby={rest.ariaLabelledby}
-            ariaDescribedby={rest.ariaDescribedby}
+            ariaDescribedby={textboxAriaDescribedBy}
```
timogasda, line 370: now we just need a test for this and we should be good :)

timogasda (dismissed):

src/property-filter/index.tsx
youdz: 🤦‍♂️ My bad, that's what I get for rushing the second revision.

youdz: Done. Sorry for the extra revisions on this one, I got too comfortable. 😛

pan-kot (approved):

PR #2259, opened 2024-05-16
feat: KeyValuePairs component

georgylobko:
### Description

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


avinashbot (changes requested):
Made a comment on Friday that I forgot to post, so...

src/key-value-pairs/internal.tsx
```diff
@@ -0,0 +1,48 @@
...
+import ColumnLayout from '../column-layout/internal';
+import Box from '../box/internal';
+
+export { KeyValuePairsProps };
+
+const InternalKeyValuePairs = React.forwardRef(
+  ({ columns, className, ...rest }: KeyValuePairsProps, ref: React.Ref<HTMLDivElement>) => {
+    return (
+      <div {...rest} className={clsx(styles['key-value-pairs'], className)} ref={ref}>
+        {!!columns?.length && (
```
just-boris, line 17: If `columns` is required, why do you need `?.`? Why do you need this condition at all?

src/key-value-pairs/internal.tsx
```diff
@@ -0,0 +1,48 @@
...
+                  <Box variant="h3" padding="n">
+                    {column.title}
+                  </Box>
+                )}
+                <dl className={styles.list}>
+                  {column.items.map((item, itemIndex) => (
+                    <div key={columnIndex + itemIndex} className={styles.item}>
+                      <dt className={styles.label}>
+                        <label className={styles['key-label']}>{item.label}</label>
+                        <InfoLinkLabelContext.Provider value="">
```
just-boris, line 31: Why is `value` an empty string?

src/key-value-pairs/internal.tsx
```diff
@@ -0,0 +1,48 @@
...
+            {columns.map((column, columnIndex) => (
+              <div className={styles.column} key={columnIndex}>
+                {column.title && (
+                  <Box variant="h3" padding="n">
+                    {column.title}
+                  </Box>
+                )}
+                <dl className={styles.list}>
+                  {column.items.map((item, itemIndex) => (
+                    <div key={columnIndex + itemIndex} className={styles.item}>
```
just-boris, line 28: Why was `key={itemIndex}` not enough here?

src/key-value-pairs/styles.scss
```diff
@@ -0,0 +1,53 @@
...
+}
+
+.column {
+  display: flex;
+  flex-direction: column;
+  row-gap: awsui.$space-scaled-m;
+}
+
+.list {
+  @include styles.styles-reset;
```
just-boris, line 20: One style reset on the root was enough, wasn't it?

src/test-utils/dom/key-value-pairs/index.ts
```diff
@@ -0,0 +1,37 @@
...
+    return this.findByClassName(styles.description);
+  }
+
+  findInfo(): ElementWrapper | null {
+    return this.findByClassName(styles.info);
+  }
+}
+
+export class KeyValuePairsColumnWrapper extends ComponentWrapper {
+  findTitle(): BoxWrapper | null {
```
just-boris, line 22: This exposes an implementation detail that title uses box. You do not want that. `ElementWrapper` will be enough

src/key-value-pairs/__tests__/key-value-pairs.test.tsx
```diff
@@ -0,0 +1,136 @@
...
+              label: 'Label for key 4',
+              value: 'Value 4',
+            },
+          ],
+        },
+      ],
+    });
+
+    expect(wrapper.findColumns()[0].findTitle()!.getElement().textContent).toBe('Column title 1');
+    expect(wrapper.findColumns()[0].findPairs()[0].findLabel()!.getElement().textContent).toBe('Label for key 1');
```
just-boris, line 109: Did you enjoy writing these selectors? How about introducing a more practical utility? `wrapper.findItem(colIndex, rowIndex)`?

src/key-value-pairs/__tests__/key-value-pairs.test.tsx
```diff
@@ -0,0 +1,136 @@
...
+            {
+              label: 'Label for key 4',
+              value: 'Value 4',
+            },
+          ],
+        },
+      ],
+    });
+
+    expect(wrapper.findColumns()[0].findTitle()!.getElement().textContent).toBe('Column title 1');
```
just-boris, line 108: We have `toHaveTextContent` matcher

src/key-value-pairs/__tests__/key-value-pairs.test.tsx
```diff
@@ -0,0 +1,136 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import * as React from 'react';
+import { render } from '@testing-library/react';
+import createWrapper from '../../../lib/components/test-utils/dom';
+import KeyValuePairs, { KeyValuePairsProps } from '../../../lib/components/key-value-pairs';
+
+function renderKeyValuePairs(props: KeyValuePairsProps) {
+  const { container, debug } = render(<KeyValuePairs {...props} />);
+  return { wrapper: createWrapper(container).findKeyValuePairs()!, debug };
```
just-boris, line 10: Seems like some debug value was left over here

src/key-value-pairs/__tests__/key-value-pairs.test.tsx
```diff
@@ -0,0 +1,136 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import * as React from 'react';
+import { render } from '@testing-library/react';
+import createWrapper from '../../../lib/components/test-utils/dom';
+import KeyValuePairs, { KeyValuePairsProps } from '../../../lib/components/key-value-pairs';
+
+function renderKeyValuePairs(props: KeyValuePairsProps) {
```
just-boris, line 8: We prefer to accept JSX here, not the props object:

```js
renderKeyValuePairs(<KeyValuePairs columns={...} />)
```

This is not much more to write, but the tests become also usage examples, which is nice to have

just-boris (commented):
Not all comments are required to fix, just let me know what you think

src/key-value-pairs/internal.tsx
georgylobko: updated

src/key-value-pairs/internal.tsx
georgylobko: updated

src/key-value-pairs/internal.tsx
georgylobko: Because there might be multiple columns, so `itemIndex` won't be unique

just-boris: The index need to be unique within a single parent, `<dl>` tag in this case. Other columns do not matter

src/key-value-pairs/styles.scss
georgylobko: Not really, root style reset does not reset some specific `dl` styles

src/test-utils/dom/key-value-pairs/index.ts
georgylobko: Technically yes, `ElementWrapper` is enough, but I exposed `BoxWrapper` to keep `findComponent` usage consistent, since every other usage exposes an appropriate component type. Do you think it is necessary to make it differently for this case?

src/key-value-pairs/styles.scss
just-boris: Correct, because the point of `styles-reset` is to cancel all inheritable CSS (colors, font-sizes, etc), not set all possible values to default.

You need to reset browser defaults on your own.

src/test-utils/dom/key-value-pairs/index.ts
just-boris: People may start relying on `BoxWrapper` return type and you will be in problem when there will be a requirement to change this for some reason

src/key-value-pairs/internal.tsx
georgylobko: updated

src/key-value-pairs/__tests__/key-value-pairs.test.tsx
georgylobko: `toHaveTextContent` checks for text entry, while `textContent` along with `toBe` checks the exact string matching

src/key-value-pairs/__tests__/key-value-pairs.test.tsx
georgylobko: I think these selectors allow you to see relationships between columns and pairs, however, you're right, they not very concise. Do you think it worth it to rewrite them?

src/key-value-pairs/__tests__/key-value-pairs.test.tsx
georgylobko: updated

src/key-value-pairs/styles.scss
georgylobko: updated

src/test-utils/dom/key-value-pairs/index.ts
georgylobko: updated

src/key-value-pairs/__tests__/key-value-pairs.test.tsx
georgylobko: updated

src/key-value-pairs/__tests__/key-value-pairs.test.tsx
just-boris: Jest-dom has a guidance for this case: https://github.com/testing-library/jest-dom?tab=readme-ov-file#tohavetextcontent

They recommend using regexp for it

```js
expect(element).toHaveTextContent(/^Text Content$/) // to match the whole content
```

Feel free to apply their suggestion or keep your way, no much difference for me

src/key-value-pairs/__tests__/key-value-pairs.test.tsx
just-boris: Feel free to discuss with other team members what they think. I approve the current PR because this extra utility can be a future addition

just-boris (approved):

just-boris (approved):

PR #2269, opened 2024-05-21
chore: Migrate CodeEditor to native CSS container queries.

DaemonCahill:
This PR migrates the CodeEditor component to using native CSS container queries and removes the need for a virtual status bar measurment.

### DIFF
--- a/src/code-editor/styles.scss
+++ b/src/code-editor/styles.scss
@@ -79,6 +79,11 @@
   }
 }
 
+.tab-list {
+  align-items: center;
+  display: inline-flex;
+}
+
 .tab-button {
   position: relative;
   display: inline-flex;
@@ -159,13 +164,21 @@
     @include styles.focus-highlight(awsui.$space-code-editor-status-focus-outline-gutter);
   }
 
+  &--divider {
+    display: inline-block;
+    block-size: awsui.$line-height-body-m;
+    inline-size: awsui.$border-code-editor-status-divider-width;
+    background: awsui.$color-border-tabs-divider;
+    vertical-align: middle;
+  }
+
   &--errors {
     /* used in test-utils */
   }
 
   /* stylelint-disable plugin/no-unsupported-browser-features */
   @supports (contain: inline-size) {
-    @container code-editor (min-width: 501px) {
+    @container code-editor not (max-width: 500px) {
       > .count {
         display: none;
       }
--- a/src/code-editor/styles.scss
+++ b/src/code-editor/styles.scss
@@ -59,11 +59,6 @@
     border-inline-end: awsui.$border-item-width solid awsui.$color-border-code-editor-default;
   }
 
-  &__left-virtual {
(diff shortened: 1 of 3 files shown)


connorlanigan (dismissed):

just-boris:
To avoid more issues like this one https://github.com/cloudscape-design/components/pull/2352 another rebase before merge is needed

src/code-editor/styles.scss
```diff
@@ -12,7 +12,9 @@
 
 .code-editor {
   @include styles.styles-reset;
-  display: inline-block;
+  container-type: inline-size;
+  container-name: code-editor;
```
just-boris, line 16: How well does this play with scoped CSS?

src/code-editor/styles.scss
```diff
@@ -177,6 +175,30 @@
   &--errors {
     /* used in test-utils */
   }
+
+  /* stylelint-disable plugin/no-unsupported-browser-features */
```
just-boris, line 179: Please check if this is still needed after  https://github.com/cloudscape-design/components/pull/2340

src/code-editor/styles.scss
DaemonCahill: It shouldn't be relevant. Our CSS "scoping" is just creating unique classnames with arbitrary strings. When the class is assigned to a DOM node, the container name will be applied accordingly. That being said, the container name itself is not encapsulated so there is a potential vulnerability similar to our use of internal custom properties. I suppose we could look into "scoping" our container names as well.

src/code-editor/styles.scss
DaemonCahill: Confirmed, this is still required.

src/code-editor/styles.scss
DaemonCahill: FYI I updated the code to remove container name and have the containment context be implicit. No functional changes here. It just removes and concerns about name spacing.

just-boris: Now with the change I am even more confused. A deeper dive needed

1. What happens when there are multiple code editors on the page?
2. Can any 3rd party CSS cancel our implicit containment context?
3. Can we accidentally apply another context in between the root and the status bar and ruin the current setup?

src/code-editor/styles.scss
just-boris: But this could be a smaller scope exception, because `@supports` does not need this anymore

DaemonCahill: Oh, I see. Sure we can do that.

just-boris (dismissed):

src/code-editor/styles.scss
just-boris: To make it more error proof, I would recommend setting container-type on the closest ancestor to avoid accidental conflicts with ace styles which we do not control

image

just-boris (approved):

just-boris (dismissed):

PR #2284, opened 2024-05-23
fix: Remove unexpected space above sticky headers in content layout

jperals:
The sticky position of table sticky headers was incorrect when using content layout inside app layout.

### Description

Issue: `AWSUI-44515`

### How has this been tested?

Added a page for visual regression testing.

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/app-layout/visual-refresh/main.tsx
```diff
@@ -61,7 +62,7 @@ export default function Main() {
           `var(${customCssProps.mobileBarHeight})`,
           !!disableBodyScroll,
           isMobile,
-          hasBackgroundOverlap && !isMobile ? tokens.spaceScaledS : '0px'
+          hasBackgroundOverlap && hasStickyBackground && !isMobile ? tokens.spaceScaledS : '0px'
```
jperals, line 65: `tokens.spaceScaledS` here defines the sticky position, but it is only needed when rendering a full-page table, otherwise it should be 0. Otherwise the same space is already applied by defining [this padding](https://github.com/cloudscape-design/components/blob/main/src/table/styles.scss#L185) inside the header instead of outside.

If adding this space without the sticky background present, the unexpected space will appear and the content will be visible behind.

pages/app-layout/with-sticky-header-table-in-content-layout.page.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { Breadcrumbs } from './utils/content-blocks';
+import labels from './utils/labels';
+import ContentLayout from '~components/content-layout';
+import ScreenshotArea from '../utils/screenshot-area';
+
+const items = generateItems(20);
+
+export default function () {
+  return (
+    <ScreenshotArea gutters={false}>
```
just-boris, line 18: Visual regression testing is fine, but how about adding a normal test with assertions?

Take this one as example: https://github.com/cloudscape-design/components/blob/8b62025c8010376f5c7db65a81581b0bc2a7409a/src/container/__integ__/awsui-applayout-sticky.test.ts#L70-L83

jperals: Added in [9992501](https://github.com/cloudscape-design/components/pull/2284/commits/9992501596cf78d7ab52b41830af3383ea1f80e9).

Checked that it fails before the changes.

just-boris (approved):

PR #2294, opened 2024-05-27
fix: Table grid navigation re-apply focus in Firefox and Safari

pan-kot:
### Description

Fixed a bug in table grid navigation when focus was not re-applied to a cell when the cell is removed from DOM in Firefox and Safari. The logic to make this happen relied on focusout event, however, the event is not firing in Firefox and Safari when a DOM node gets removed.

Ref: TjeaAaLsCsbV

Before (Firefox, the focus is lost and a Tab keypress is needed to restore it):

https://github.com/cloudscape-design/components/assets/20790937/e0c575fa-cebe-42a4-a27c-35caed7169ae

After (Firefox, the focus is not lost):

https://github.com/cloudscape-design/components/assets/20790937/ddf9a34d-9e05-4a06-8cd9-7b8c78ac4cf3

### How has this been tested?

* Existing unit and integration tests
* Screenshot tests
* Manual testing in Chrome, Firefox and Safari
* New behaviour test for progressive loading

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/table/table-role/grid-navigation.tsx
```diff
@@ -88,16 +89,18 @@ class GridNavigationProcessor {
     this._table = table;
 
     this.table.addEventListener('focusin', this.onFocusin);
-    this.table.addEventListener('focusout', this.onFocusout);
     this.table.addEventListener('keydown', this.onKeydown);
 
     this.updateFocusTarget();
 
+    const tableNodesObserver = new MutationObserver(this.onTableNodeMutation);
+    tableNodesObserver.observe(table, { childList: true, subtree: true });
```
Al-Dani, line 97: Can we do a more targeted observation instead of listening to updates in the entire table? using MutationObserver can cost us performance

Al-Dani: also, could we replace `focusout` with `blur` and avoid MutationObserver altogether?

pan-kot: Thanks, there was actually a better way. Since all focusables are registered explicitly we can use the unregister events to capture node removals. I updated my approach so no observer now.

Al-Dani (approved):

PR #2298, opened 2024-05-27
feat: Add PromptInput component

rubencarvalho:
### Description

New PromptInput component
image

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/prompt-input/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+      disableBrowserAutocorrect,
+      disableActionButton,
+      spellcheck,
+      readOnly,
+      ...props
+    }: PromptInputProps,
+    ref: React.Ref<PromptInputProps.Ref>
+  ) => {
+    const baseComponentProps = useBaseComponent('PromptInput', {
+      props: { readOnly, autoComplete, autoFocus, disableBrowserAutocorrect, disableActionButton, spellcheck },
```
gethinwebster, line 25: let's also track these properties: `actionButtonIconName`, `minRows`, `maxRows`

src/prompt-input/internal.tsx
```diff
@@ -0,0 +1,168 @@
...
+  (
+    {
+      value,
+      actionButtonAriaLabel,
+      actionButtonIconName,
+      actionButtonIconUrl,
+      actionButtonIconSvg,
+      actionButtonIconAlt,
+      ariaLabel,
+      autoComplete = true,
```
gethinwebster, line 28: I'm not sure defaulting this to true makes a lot of sense given the sort of context in which this component will be used
```suggestion
      autoComplete,
```

src/prompt-input/internal.tsx
```diff
@@ -0,0 +1,168 @@
...
+      }
+    };
+
+    const handleChange = (event: React.ChangeEvent<HTMLTextAreaElement>) => {
+      fireNonCancelableEvent(onChange, { value: event.target.value });
+      adjustTextareaHeight();
+    };
+
+    const adjustTextareaHeight = useCallback(() => {
+      const PADDING = 4;
```
gethinwebster, line 81: it should be possible to import design tokens as JS values

https://github.com/cloudscape-design/components/blob/main/src/app-layout/visual-refresh/main.tsx#L10

src/prompt-input/internal.tsx
```diff
@@ -0,0 +1,168 @@
...
+      const LINE_HEIGHT = 20;
+      if (textareaRef.current) {
+        textareaRef.current.style.height = 'auto';
+
+        const newHeight = Math.min(
+          textareaRef.current.scrollHeight + PADDING,
+          (maxRows ?? 3) * (LINE_HEIGHT + PADDING / 2) + PADDING
```
gethinwebster, line 88: suggest putting this default value of `3` where we destructure the props rather than inline like this, it makes it easier to see as an overview

src/test-utils/dom/prompt-input/index.ts
```diff
@@ -0,0 +1,38 @@
...
+import styles from '../../../prompt-input/styles.selectors.js';
+
+export default class PromptInputWrapper extends ComponentWrapper {
+  static rootSelector = styles.root;
+
+  findNativeTextarea(): ElementWrapper<HTMLTextAreaElement> {
+    return this.find<HTMLTextAreaElement>(`.${styles.textarea}`)!;
+  }
+
+  findSubmitButton(): ElementWrapper<HTMLButtonElement> {
```
gethinwebster, line 14: suggest calling this `findActionButton` as that's how it's referred to in all docs

gethinwebster (commented):
first some comments on the component itself

src/prompt-input/styles.scss
```diff
@@ -0,0 +1,98 @@
...
+}
+
+.button {
+  position: absolute;
+  inset-inline-end: $send-icon-right-spacing;
+  inset-block-end: 0;
+}
+
+.action-button {
+  /* used in test-utils */
```
Al-Dani, line 97: general thought: we were discussing separation of classes for tests from styling-related to improve stability of selectors. That was done for [copy-to-clipboard](https://github.com/cloudscape-design/components/blob/fcd9ab11a71083f7476ba2c5f0a494ef7c047398/src/copy-to-clipboard/test-classes/styles.scss#L4) and [anchor navigation](https://github.com/cloudscape-design/components/blob/fcd9ab11a71083f7476ba2c5f0a494ef7c047398/src/anchor-navigation/test-classes/styles.scss#L4). I'd suggest we continue doing so for new components

src/prompt-input/__tests__/prompt-input.test.tsx
```diff
@@ -0,0 +1,256 @@
...
+    test('is not added if set to null', () => {
+      const { wrapper } = renderPromptInput({ value: '' });
+      expect(wrapper.findNativeTextarea().getElement()).not.toHaveAttribute('aria-describedby');
+    });
+    test('can be set to custom value', () => {
+      const { wrapper } = renderPromptInput({ value: '', ariaDescribedby: 'my-custom-id' });
+      expect(wrapper.findNativeTextarea().getElement()).toHaveAttribute('aria-describedby', 'my-custom-id');
+    });
+    test('can be customized without controlId', () => {
+      const { wrapper } = renderPromptInput({ value: '', id: undefined, ariaDescribedby: 'my-custom-id' });
```
Al-Dani, line 240: why `id`, if test is about `controlId`?

src/prompt-input/interfaces.ts
```diff
@@ -0,0 +1,98 @@
...
+  InputAutoCorrect,
+  InputAutoComplete,
+  InputKeyEvents,
+  InputSpellcheck,
+  BaseChangeDetail,
+} from '../input/interfaces';
+import { FormFieldValidationControlProps } from '../internal/context/form-field-context';
+import { IconProps } from '../icon/interfaces';
+
+export interface PromptInputProps
```
Al-Dani, line 16: another general thought: for the sake of future archaeologists to follow up in `quip/vQJVAwOM2gv2` with some comments/answers regarding  decisions that were made here. For example, about `select()`

src/prompt-input/internal.tsx
katiegeorge: Its set to true by default for the input and textarea components, do you think it'll be confusing for people to have it as the opposite here?

src/prompt-input/internal.tsx
katiegeorge: I saw that, but noticed that the only tokens in that file are for charts and a few for spacing. I did a surface-level search and found a couple other places where we declare padding as a number value:
- https://github.com/cloudscape-design/components/blob/main/src/code-editor/pane.tsx#L17
- https://github.com/cloudscape-design/components/blob/main/src/mixed-line-bar-chart/bar-series.tsx#L74

But I agree that this seems like it should pull from the token value. Perhaps a larger team thread should be had?

src/prompt-input/__tests__/prompt-input.test.tsx
katiegeorge: Hmm, good catch. This was copied from the test from text area: https://github.com/cloudscape-design/components/blob/main/src/textarea/__tests__/textarea.test.tsx#L295

I'll update it.

src/prompt-input/internal.tsx
katiegeorge: I've updated it and I agree I like it better when both values pull from the same source.

src/prompt-input/interfaces.ts
katiegeorge: This is a great catch! I missed that method in the API doc, since Ruben hadn't implemented it already. I've added it.

katiegeorge: I've also updated the API doc with a couple of the other changes we made, namely changing `submit` to `action` everywhere you see it.

src/prompt-input/internal.tsx
gethinwebster: `autoComplete` makes most sense in a form context (it's e.g. for auto-filling address fields), so given that the PromptInput is designed to be primarily used outside that context it makes sense to be different

gethinwebster (dismissed):

Al-Dani (dismissed):

gethinwebster (approved):

Al-Dani (approved):

PR #2316, opened 2024-05-30
feat: Read-only states for form related components

georgylobko:
### Description

Introduced `readOnly` state for form related components.

The `readonly` attribute is supported by `text`, `search`, `url`, `tel`, `email`, `password`, `date`, `month`, `week`, `time`, `datetime-local`, and number <input> types and the <textarea> form control elements. https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/readonly

Therefore for most elements, the `readOnly` state is handled manually by adding the appropriate style rules and preventing events from being triggered.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/date-range-picker/index.tsx
```diff
@@ -286,9 +286,7 @@ const DateRangePicker = React.forwardRef(
           })}
           hideCaret={true}
           onClick={() => {
-            if (!readOnly && !disabled) {
```
georgylobko, line 289: `ButtonTrigger` doesn't trigger this event when it's in either `readOnly` or `disabled` state anyway.

src/internal/components/button-trigger/index.tsx
```diff
@@ -99,7 +99,7 @@ const ButtonTrigger = (
       {children}
       {!hideCaret && (
         <span className={styles.arrow}>
-          <InternalIcon name="caret-down-filled" variant={disabled ? 'disabled' : 'normal'} />
+          <InternalIcon name="caret-down-filled" variant={disabled || readOnly ? 'disabled' : 'normal'} />
```
georgylobko, line 102: Didn't introduce a new variant because the `InternalIcon` doesn't use any `aria-*` attributes, and `disabled` and `readOnly` share the same style rules

src/token-group/token.tsx
georgylobko, line 1: for this one a11y tests threw an error that `aria-readonly` is not supported. In this case a way how users will be informed about the read-only state is an open question

src/multiselect/internal.tsx
georgylobko, line 1: For this component, the readOnly state applies to the trigger button and token groups. I'm not sure if aria-readonly should be added to the component's root element.

pages/multiselect/permutations.page.tsx
```diff
@@ -82,6 +82,16 @@ const permutations = createPermutations<MultiselectProps & { inlineTokens?: bool
     i18nStrings: [i18nStrings],
     inlineTokens: [true],
   },
+  {
+    placeholder: ['Select an item'],
+    readOnly: [true],
+    tokenLimit: [2],
+    options: [options],
+    selectedOptions: [[options[0], options[1]]],
```
michaeldowseza, line 90: Do we need a permutation for when the number of selected tokens exceeds the `tokenLimit` as that would then show the "Show more" state?

pages/radio-group/permutations.page.tsx
```diff
@@ -92,6 +92,17 @@ const permutations = createPermutations<RadioGroupProps>([
...
     ],
   },
+  {
+    readOnly: [true],
+    value: ['first'],
+    items: [
+      [
+        { value: 'first', label: 'First Button' },
+        { value: 'second', label: 'Second Button' },
+        { value: 'third', label: 'Third Button', disabled: true },
```
michaeldowseza, line 102: Do we need a permutation when a `disabled` radio is selected while in a readOnly state? Maybe an edge case or maybe not even a valid state?

pages/token-group/permutations.page.tsx
```diff
@@ -179,6 +179,68 @@ const permutations = createPermutations<TokenGroupProps>([
       ],
     ],
   },
+  {
+    i18nStrings: [i18nStrings],
+    limit: [undefined],
```
michaeldowseza, line 184: Similar to an earlier comment, do we need a permutation where there is a limit that has been exceeded and we are showing the "Show more" state?

src/checkbox/base-checkbox.tsx
```diff
@@ -24,6 +24,14 @@ export interface BaseCheckboxProps extends BaseComponentProps, FormFieldControlP
    */
   disabled?: boolean;
 
+  /**
+   * Specifies if the control is readOnly, which prevents the
+   * user from modifying the value, but does not prevent the value from
+   * being included in a form submission. A readOnly control is still focusable.
+   * Should not be set together with `disabled: true`
```
michaeldowseza, line 31: Missing full stop.

src/internal/components/abstract-switch/__tests__/abstract-switch.test.tsx
```diff
@@ -92,5 +93,45 @@ describe('Abstract switch', () => {
...
+        ariaLabelledby: 'some-custom-label',
+        ariaDescribedby: 'some-custom-description',
+        label: 'Default label',
+        description: 'Default description',
+        disabled: true,
+        nativeControl: nativeControlProps => <input {...nativeControlProps} className="switch-element" type="radio" />,
+        onClick,
+      });
+
+      wrapper.findByClassName(styles['label-wrapper'])!.click();
```
michaeldowseza, line 113: Can we not use the AbstractSwitch test util wrapper here? 

https://github.com/cloudscape-design/components/blob/c54515793103c374eb72cf76a52a618e262c960a/src/test-utils/dom/internal/abstract-switch.ts#L6

Or similar to above, wrap the render in a label and click that?

src/internal/components/button-trigger/index.tsx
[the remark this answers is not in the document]
michaeldowseza: Is there an ongoing follow up to resolve this?

src/internal/components/button-trigger/index.tsx
```diff
@@ -74,22 +74,28 @@ const ButtonTrigger = (
...
 
+  if (!readOnly) {
+    attributes = {
+      ...attributes,
+      onKeyDown: onKeyDown && (event => fireKeyboardEvent(onKeyDown, event)),
+      onKeyUp: onKeyUp && (event => fireKeyboardEvent(onKeyUp, event)),
+      onMouseDown: onMouseDown && (event => fireCancelableEvent(onMouseDown, {}, event)),
+      onClick: onClick && (event => fireCancelableEvent(onClick, {}, event)),
+      onFocus: onFocus && (event => fireCancelableEvent(onFocus, {}, event)),
+      onBlur: onBlur && (event => fireCancelableEvent(onBlur, { relatedTarget: event.relatedTarget }, event)),
```
michaeldowseza, line 95: From the api documents readOnly controls are still focusable, should these events then still be fired?

src/radio-group/__tests__/radio-group.test.tsx
```diff
@@ -110,6 +110,31 @@ describe('items', () => {
     expect(onChange).not.toHaveBeenCalled();
   });
 
+  test('can be readOnly', () => {
```
michaeldowseza, line 113: The test name does not reflect what the test is doing. Suggestion "Adds aria-readonly when readOnly is set to true".

src/radio-group/__tests__/radio-group.test.tsx
```diff
@@ -110,6 +110,31 @@ describe('items', () => {
...
+  });
+
+  test('does not trigger change handler if readOnly', () => {
+    const onChange = jest.fn();
+    const { wrapper } = renderRadioGroup(
+      <RadioGroup value="val1" name="test" items={defaultItems} readOnly={true} onChange={onChange} />
+    );
+
+    act(() => wrapper.findButtons()[0].findLabel().click());
+    expect(onChange).not.toHaveBeenCalled();
```
michaeldowseza, line 126: I'm not sure this would have fired even if it wasn't readOnly. The test is trying to click the same item that is already selected so the value doesn't change.

src/__tests__/__snapshots__/documenter.test.ts.snap
```diff
@@ -4415,6 +4415,15 @@ in the native control.",
       "optional": true,
       "type": "string",
     },
+    Object {
+      "description": "Specifies if the control is readOnly, which prevents the
+user from modifying the value, but does not prevent the value from
+being included in a form submission. A readOnly control is still focusable.
```
michaeldowseza, line 4421: From a content point of view "readOnly" is in camel case which would refer to the `readOnly` property but in a description as an adjective should this be "readonly" or "read-only"?

 (The internet is fairly divided how this is spelt)

src/select/interfaces.ts
```diff
@@ -142,6 +142,13 @@ export interface BaseSelectProps
    * Called when input focus is set onto the UI control.
    */
   onFocus?: NonCancelableEventHandler;
+
+  /**
+   * Specifies if the control is readOnly, which prevents the
+   * user from modifying the value and opening the dropdown, but does not prevent the value from
+   * being included in a form submission. A readOnly control is still focusable.
```
michaeldowseza, line 149: "but does not prevent the value from being included in a form submission" - This seems to be a semi generic description for all the docs but components like Select currently does not submit any values if they are in a form. With this description could it not give the impression that it would be included?

Sandbox example: https://codesandbox.io/p/sandbox/sad-wildflower-dwd9n3

src/slider/__tests__/slider.test.tsx
```diff
@@ -444,4 +444,26 @@ describe('Slider a11y', () => {
...
+    const onChange = jest.fn();
+    const wrapper = renderSlider({
+      min: 0,
+      max: 100,
+      value: 50,
+      onChange: event => onChange(event.detail),
+      readOnly: true,
+    });
+
+    wrapper.setInputValue('60');
```
michaeldowseza, line 465: This test is a bit difficult for me. By [definition](https://github.com/cloudscape-design/components/blob/c54515793103c374eb72cf76a52a618e262c960a/src/test-utils/dom/input/base-input.ts#L37-L47) this test util calls `onChange`.

So in theory the test tests what it needs to but it feels like its being tested in an unnatural way. I don't have any better suggestion for you though so we can leave it as is.

src/tiles/styles.scss
```diff
@@ -63,6 +63,10 @@ $gutter: awsui.$space-grid-gutter;
     background: awsui.$color-background-tiles-disabled;
   }
 
+  &.read-only {
```
michaeldowseza, line 66: Why did this change to `read-only` here while all the previous files had it as `readonly`?

src/token-group/dismiss-button.tsx
```diff
@@ -6,20 +6,24 @@ import InternalIcon from '../icon/internal';
...
+  ref: Ref<HTMLButtonElement>
+) {
   return (
     <button
       ref={ref}
       type="button"
       className={styles['dismiss-button']}
       aria-disabled={disabled ? true : undefined}
-      onClick={!disabled && onDismiss ? () => onDismiss() : undefined}
+      onClick={!disabled && !readOnly && onDismiss ? () => onDismiss() : undefined}
```
michaeldowseza, line 26: Non-blocking: Just noticing how inconsistent we are in code for handling this. Some cases we return and then like here and a few other places we assign undefined.

pages/multiselect/permutations.page.tsx
georgylobko: I don't think we need this example, because the "Show more" behavior does not change when `readOnly` is set to true

pages/radio-group/permutations.page.tsx
georgylobko: I don't think we need it. The example in permutation page represents a state with `readOnly` selected radio, which is, in our case, pretty much the same as `disabled`, because both of them could not be changed.

This behavior is described in [general guidelines](https://cloudscape.design/components/radio-group?tabId=usage): `If one item is selected and inactive, always deactivate the other items in the radio group.`

pages/token-group/permutations.page.tsx
georgylobko: "Show more" behavior remains the same in this case, same as in the earlier comment

src/checkbox/base-checkbox.tsx
georgylobko: updated

src/internal/components/abstract-switch/__tests__/abstract-switch.test.tsx
georgylobko: good catch. yes, we can. updated

src/internal/components/button-trigger/index.tsx
georgylobko: No, they should not. The element's focus is managed by css, not by these events. They can cause unexpected behavior for `readOnly` state

src/radio-group/__tests__/radio-group.test.tsx
georgylobko: agree! updated

src/radio-group/__tests__/radio-group.test.tsx
georgylobko: It actually fires when it's not `readOnly`, the component does not check if an element is selected, but I agrees that it might cause confusion, so changed `value` to `val2`

src/__tests__/__snapshots__/documenter.test.ts.snap
georgylobko: replaced every entry with `read-only`

src/select/interfaces.ts
georgylobko: yes, agree, this behavior applies to native form controls like `input` type `checkbox`, `radio` etc. removed this statement

src/tiles/__tests__/tiles.test.tsx
```diff
@@ -110,6 +110,20 @@ describe('items', () => {
     expect(onChange).not.toHaveBeenCalled();
   });
 
+  test('can be readOnly', () => {
```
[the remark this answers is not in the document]
georgylobko: updated

src/tiles/styles.scss
georgylobko: updated

src/token-group/dismiss-button.tsx
georgylobko: Changed to

`onClick={() => {
  if (disabled || readOnly || !onDismiss) {
    return;
  }

  onDismiss();
}}`

pages/multiselect/permutations.page.tsx
michaeldowseza: Is that not then what the permutation would be asserting? so that in future if a change is made to the readonly styles of a button/link that it doesn't show up here. Are we also maybe missing a unit test that in a readonly state we can still click the "Show more" button?

pages/radio-group/permutations.page.tsx
michaeldowseza: Its maybe why I consider it an edge case, what happens if the styling for disabled is changed? Do we expect to see the disable or readonly styles while it is in readonly mode?

src/radio-group/interfaces.ts
```diff
@@ -55,6 +55,13 @@ export interface RadioGroupProps extends BaseComponentProps, FormFieldControlPro
    * @deprecated Has no effect.
    */
   controlId?: string;
+
+  /**
+   * Specifies if the whole group is readOnly, which prevents the
```
michaeldowseza, line 60: "readOnly" -> "read-only"

src/select/interfaces.ts
```diff
@@ -142,6 +142,12 @@ export interface BaseSelectProps
    * Called when input focus is set onto the UI control.
    */
   onFocus?: NonCancelableEventHandler;
+
+  /**
+   * Specifies if the control is readOnly, which prevents the
```
michaeldowseza, line 147: "readOnly" -> "read-only"

src/tiles/interfaces.ts
```diff
@@ -59,6 +59,13 @@ export interface TilesProps extends BaseComponentProps, FormFieldControlProps {
    * If the component controls any secondary content (for example, another form field), use this to provide an ID referring to the secondary content.
    */
   ariaControls?: string;
+
+  /**
+   * Specifies if the control is readOnly, which prevents the
```
michaeldowseza, line 64: "readOnly" -> "read-only"

src/token-group/interfaces.ts
```diff
@@ -59,6 +59,12 @@ export interface TokenGroupProps extends BaseComponentProps {
    * Use to assign unique labels when there are multiple token groups with the same `limitShowMore` label on one page.
    */
   limitShowMoreAriaLabel?: string;
+  /**
+   * Specifies if the control is readOnly, which prevents the
```
michaeldowseza, line 63: "readOnly" -> "read-only"

michaeldowseza (commented):
Minor comments.

pages/multiselect/permutations.page.tsx
georgylobko: added u test and permutation example

pages/radio-group/permutations.page.tsx
georgylobko: when both `disabled` and `readOnly` are true, `disabled` takes precedence

src/token-group/interfaces.ts
georgylobko: updated

src/tiles/interfaces.ts
georgylobko: updated

src/slider/interfaces.ts
```diff
@@ -75,6 +75,13 @@ export interface SliderProps extends BaseComponentProps, FormFieldValidationCont
    * @i18n
    */
   i18nStrings?: SliderProps.I18nStrings;
+
+  /**
+   * Specifies if the control is readOnly, which prevents the
```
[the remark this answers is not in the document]
georgylobko: updated

src/select/interfaces.ts
georgylobko: updated

src/radio-group/interfaces.ts
georgylobko: updated

src/__tests__/__snapshots__/documenter.test.ts.snap
```diff
@@ -4415,6 +4415,15 @@ in the native control.",
       "optional": true,
       "type": "string",
     },
+    Object {
+      "description": "Specifies if the control is readOnly, which prevents the
```
michaeldowseza, line 4419: I think this snapshot should have been updated?

src/checkbox/base-checkbox.tsx
```diff
@@ -24,6 +24,14 @@ export interface BaseCheckboxProps extends BaseComponentProps, FormFieldControlP
    */
   disabled?: boolean;
 
+  /**
+   * Specifies if the control is readOnly, which prevents the
```
michaeldowseza, line 28: "readOnly" -> "read-only"?

michaeldowseza (changes requested):
Last description changes still needed.

michaeldowseza (approved):
LGTM!

michaeldowseza (approved):

PR #2336, opened 2024-06-05
fix: Fix table cell focus outline

pan-kot:
### Description

This PR mainly changes how paddings are applied to table cells. Previously the paddings were applied to the cell directly (TD or TH element) and now all paddings are applied to the nested content element (the DIV element). The content element also receives the `overflow: none` from the cell. This way, the content DIV does not allow for the content to overflow but the cell element does. As result, we can render the cell focus outline which exceeds the dimensions of the cell on the left.

Because the borders are still applied to the cells and not the content, there is a 1px difference in cell size compared to the previous version. That difference is compensated by subtracting border width from padding, but due to conditional padding application in certain cases there is a small pixel shift in some table screenshots.

Rel: AWSUI-36760

### How has this been tested?

* Existing unit- and integration tests;
* VR tests;
* Screenshot tests and Lighthouse tests (run accfb3a0-c67a-47a5-8381-2925dc27e2a7);
* Dry-tun to live (build 6888143054).

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/styles/forms/mixins.scss
+++ b/src/internal/styles/forms/mixins.scss
@@ -50,6 +50,19 @@
     $gutter-vertical: map.get($gutter, 'vertical');
     $gutter-horizontal: map.get($gutter, 'horizontal');
   }
+  $gutter-top: $gutter-vertical;
+  $gutter-bottom: $gutter-vertical;
+  @if type-of($gutter-vertical) == 'map' {
+    $gutter-top: map.get($gutter-vertical, 'top');
+    $gutter-bottom: map.get($gutter-vertical, 'bottom');
+  }
+  $gutter-left: $gutter-horizontal;
+  $gutter-right: $gutter-horizontal;
+  @if type-of($gutter-horizontal) == 'map' {
+    $gutter-left: map.get($gutter-horizontal, 'left');
+    $gutter-right: map.get($gutter-horizontal, 'right');
+  }
+
   position: relative;
   // Add a special outline for Window High Contrast Mode.
   // This mode will remove all box shadows from the side and can only use outline
@@ -58,7 +71,7 @@
   // are transparent.
   & {
     outline: 2px dotted transparent;
-    outline-offset: calc(#{$gutter-horizontal} - 1px);
+    outline-offset: calc(#{$gutter-left} - 1px);
   }
 
   // Regular rounded outline for all other browsers and modes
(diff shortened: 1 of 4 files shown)


src/internal/styles/forms/mixins.scss
```diff
@@ -50,6 +50,19 @@
...
+    $gutter-top: map.get($gutter-vertical, 'top');
+    $gutter-bottom: map.get($gutter-vertical, 'bottom');
+  }
+  $gutter-left: $gutter-horizontal;
+  $gutter-right: $gutter-horizontal;
+  @if type-of($gutter-horizontal) == 'map' {
+    $gutter-left: map.get($gutter-horizontal, 'left');
+    $gutter-right: map.get($gutter-horizontal, 'right');
+  }
+
```
pan-kot, line 65: The code above allows to define outline gutters for individual sides - this way we can define a different offset on the left (inline-start, in fact) for table's first column.

src/table/body-cell/styles.scss
```diff
@@ -44,24 +58,47 @@ $interactive-column-padding-inline-end: calc(#{$cell-horizontal-padding} + #{aws
...
 
 .body-cell {
   box-sizing: border-box;
-  padding-block-start: $cell-vertical-padding;
-  padding-block-end: $cell-vertical-padding-w-border;
-  padding-inline-end: $cell-horizontal-padding;
+  @include cell-padding-inline-start($cell-horizontal-padding);
+  @include cell-padding-block-start($cell-vertical-padding);
+  @include cell-padding-block-end($cell-vertical-padding-w-border);
+  @include cell-padding-inline-end($cell-horizontal-padding);
```
pan-kot, line 101: The mixins are not strictly necessary but they allow to make the code difference minimal and arguable improve readability by avoiding too much nesting. There is an opportunity to refactor the styles further by making body and body content styles less coupled.

src/table/body-cell/td-element.tsx
```diff
@@ -122,17 +122,19 @@ export const TableTdElement = React.forwardRef<HTMLTableCellElement, TableTdElem
...
-              isExpanded={isExpanded}
-              onExpandableItemToggle={onExpandableItemToggle}
-              expandButtonLabel={expandButtonLabel}
-              collapseButtonLabel={collapseButtonLabel}
-            />
-          </div>
-        )}
-        <span className={styles['body-cell-content']}>{children}</span>
+        <div className={styles['body-cell-content']}>
+          {level !== undefined && isExpandable && (
```
pan-kot, line 126: It was necessary to move the expand toggle to inside the content to keep the changes to body-cell CSS minimal. In the future, we will need to move it back and adjust styles or maybe introduce another wrapper over `{children}`. That will be necessary to allow having both the expandable toggle and editable cell content in a single cell.

Al-Dani (approved):

PR #2345, opened 2024-06-06
chore: Connect performance marks to DOM elements

connorlanigan:
### Description

This PR extends the existing performance mark mechanism with a mechanism to identify the corresponding DOM element. Each performance mark already contains an ID (in `detail.instanceIdentifier`) that is tied to the lifetime of the component.

With this PR, this same ID is now attached to a DOM element inside the component (not necessarily the _root_ element), in the data attribute `data-analytics-performance-mark`. Thus, the elements for each performance mark can be found like this:
```js
performance
  .getEntriesByType('mark')
  .filter(mark => mark.detail?.source === 'awsui')
  .map(mark => ({
    mark,
    element: document.querySelector(
      `[data-analytics-performance-mark="${mark.detail.instanceIdentifier}"]`
    ),
  }));


```

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/hooks/use-performance-marks/index.ts
+++ b/src/internal/hooks/use-performance-marks/index.ts
@@ -2,19 +2,20 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { useEffect, useRef } from 'react';
-import { useIdFallback } from '../use-unique-id';
+import { useRandomId } from '../use-unique-id';
 import { useEffectOnUpdate } from '../use-effect-on-update';
 import { useModalContext } from '../../context/modal-context';
 
 /*
 This hook allows setting an HTML attribute after the first render, without rerendering the component.
 */
-function useAttributeWithoutRerendering(elementRef: React.RefObject<HTMLElement>, value: string) {
+function usePerformanceMarkAttribute(elementRef: React.RefObject<HTMLElement>, value: string) {
   const attributeName = 'data-analytics-performance-mark';
 
   const attributeValueRef = useRef<string | undefined>();
 
   useEffect(() => {
+    // With this effect, we apply the attribute only on the client, to avoid hydration errors.
     attributeValueRef.current = value;
     elementRef.current?.setAttribute(attributeName, value);
   }, [value, elementRef]);
@@ -36,9 +37,9 @@ export function usePerformanceMarks(
   getDetails: () => Record<string, string | boolean | number | undefined>,
   dependencies: React.DependencyList
 ) {
-  const id = useIdFallback();
+  const id = useRandomId();
   const { isInModal } = useModalContext();
-  const attributes = useAttributeWithoutRerendering(elementRef, id);
+  const attributes = usePerformanceMarkAttribute(elementRef, id);
 
   useEffect(() => {
     if (!enabled || !elementRef.current || isInModal) {
--- a/src/internal/hooks/use-performance-marks/index.ts
+++ b/src/internal/hooks/use-performance-marks/index.ts
@@ -1,11 +1,34 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-import { useEffect } from 'react';
-import { useUniqueId } from './use-unique-id';
-import { useEffectOnUpdate } from './use-effect-on-update';
-import { useModalContext } from '../context/modal-context';
+import { useEffect, useRef } from 'react';
+import { useIdFallback } from '../use-unique-id';
+import { useEffectOnUpdate } from '../use-effect-on-update';
+import { useModalContext } from '../../context/modal-context';
 
+/*
+This hook allows setting an HTML attribute after the first render, without rerendering the component.
+*/
+function useAttributeWithoutRerendering(elementRef: React.RefObject<HTMLElement>, value: string) {
+  const attributeName = 'data-analytics-performance-mark';
+
+  const attributeValueRef = useRef<string | undefined>();
+
+  useEffect(() => {
+    attributeValueRef.current = value;
+    elementRef.current?.setAttribute(attributeName, value);
+  }, [value, elementRef]);
+
+  return {
+    [attributeName]: attributeValueRef.current,
+  };
+}
+
+/**
+ * This function returns an object that needs to be spread onto the same
+ * element as the `elementRef`, so that the data attribute is applied
+ * correctly.
+ */
 export function usePerformanceMarks(
   name: string,
   enabled: boolean,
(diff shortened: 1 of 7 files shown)


src/internal/hooks/use-performance-marks/index.ts
```diff
@@ -1,20 +1,44 @@
...
+ */
 export function usePerformanceMarks(
   name: string,
   enabled: boolean,
   elementRef: React.RefObject<HTMLElement>,
   getDetails: () => Record<string, string | boolean | number | undefined>,
   dependencies: React.DependencyList
 ) {
-  const id = useUniqueId();
+  const id = useIdFallback();
```
taheramr, line 39: Why this change is needed?

src/internal/hooks/use-performance-marks/index.ts
```diff
@@ -1,20 +1,44 @@
...
-import { useModalContext } from '../context/modal-context';
+import { useEffect, useRef } from 'react';
+import { useIdFallback } from '../use-unique-id';
+import { useEffectOnUpdate } from '../use-effect-on-update';
+import { useModalContext } from '../../context/modal-context';
 
+/*
+This hook allows setting an HTML attribute after the first render, without rerendering the component.
+*/
+function useAttributeWithoutRerendering(elementRef: React.RefObject<HTMLElement>, value: string) {
```
taheramr, line 12: Can we change the naming to relate it more to the data analytics performance mark? to me it felt like a generic function but then we use inside of it `data-analytics-performance-mark`.

If it is expected to be used somewhere else in the fututre for a different usecase we can then make the attribute name as an input param

src/internal/hooks/use-performance-marks/index.ts
taheramr: Apart from the why, looking at the `useIdFallback()` hook it generates a random id per usage using its `counter`, does this mean for every component that gets rendered we'll have a different data analytics attribute?

How does it generally work when we try to collect performance about all of our buttons or tables for example and each one of those has a different analytics id?

connorlanigan: `useUniqueId` uses React's `useId` hook when available, which creates an ID that is unique only for that React tree. If there are multiple React trees on a page (e.g. a micro-frontend architecture), these IDs can collide. With the fallback hook, we generate a random ID, which is very unlikely to collide.

Yes, each component instance gets its own ID here, i.e. if there are two buttons on the page they will have different attribute values. The ID is generally intended to be used only in the direction from a performance mark to a DOM node. So if you want to find the performance of all Buttons and Tables, you would first query all performance marks from the browser, then for each mark you lookup the DOM node.

src/internal/hooks/use-unique-id/index.ts
```diff
@@ -3,7 +3,7 @@
 import React, { useRef } from 'react';
 
 let counter = 0;
-const useIdFallback = () => {
+export const useIdFallback = () => {
```
just-boris, line 6: If we are going to use it this way, we should rename it to `useRandomId`. We want to use it because it is "random", not because it is "fallback"

src/internal/hooks/use-performance-marks/index.ts
```diff
@@ -1,20 +1,44 @@
...
+  const attributeName = 'data-analytics-performance-mark';
+
+  const attributeValueRef = useRef<string | undefined>();
+
+  useEffect(() => {
+    attributeValueRef.current = value;
+    elementRef.current?.setAttribute(attributeName, value);
```
just-boris, line 19: How about assigning `__awsuiMetadata__` property here instead of an attribute? Would it make it easier to maintain?

src/internal/hooks/use-performance-marks/index.ts
```diff
@@ -1,20 +1,44 @@
...
+import { useEffectOnUpdate } from '../use-effect-on-update';
+import { useModalContext } from '../../context/modal-context';
 
+/*
+This hook allows setting an HTML attribute after the first render, without rerendering the component.
+*/
+function useAttributeWithoutRerendering(elementRef: React.RefObject<HTMLElement>, value: string) {
+  const attributeName = 'data-analytics-performance-mark';
+
+  const attributeValueRef = useRef<string | undefined>();
```
just-boris, line 15: Why is this ref initially empty and does not use the `value` argument?

src/internal/hooks/use-performance-marks/index.ts
```diff
@@ -1,20 +1,44 @@
...
+
+  useEffect(() => {
```
just-boris, line 17: Is this effect really needed? Both `value` and `elementRef` seem to never change

src/internal/hooks/use-performance-marks/index.ts
just-boris: Actually, if we eliminate useEffect completely (https://github.com/cloudscape-design/components/pull/2345#discussion_r1631105798), the attribute is fine

src/internal/hooks/use-performance-marks/index.ts
connorlanigan: If the ref were set to the attribute value initially, the attribute value would be included in the first render, which means that it would be present in server-side-rendered markup (and cause a hydration mismatch). By setting it only immediately after the first render, the attribute is present on any render except the first one.

src/internal/hooks/use-performance-marks/index.ts
connorlanigan: If we assigned it to a property instead of a HTML attribute, it would be more expensive to query the DOM for the element. With the attribute we can run `querySelector`, but with a property we would probably have to loop through every DOM element until we find the one with the property set to the right value.

src/internal/hooks/use-performance-marks/index.ts
connorlanigan: The effect is needed so that the value is set only after the first render. `value` and `elementRef` will indeed not change, so we can make the dependency array empty - I'll update the code.

src/internal/hooks/use-performance-marks/index.ts
just-boris: "expensive to query" needs some supportive data. For instance, [TreeWalker DOM API](https://developer.mozilla.org/en-US/docs/Web/API/Treewalker) exists.

I can approve the version with attribute, just saying that some assumptions need validation

src/internal/hooks/use-performance-marks/index.ts
just-boris: Got it, so we apply `useEffect` as "client-side only" technique. I think it would be useful to add a comment with this info. Leaving dependencies is fine, because it makes the eslint rule happy

src/internal/hooks/use-performance-marks/index.ts
connorlanigan: Interesting, I did not know about the TreeWalker API! I've created a small benchmark (on jsbench.me) using the HTML from [our Table demo](https://cloudscape.design/examples/react/table.html), and it confirms that `querySelector` is much faster (or that I'm using the TreeWalker inefficiently)

benchmark

just-boris (approved):

src/internal/hooks/use-performance-marks/index.ts
just-boris: 1. Even if it is slower, it is unclear if it matters for this use-case (depends how often this lookup is called)
2. We can also fine tune the benchmark, but this is not important for the PR, the attributes can stay

PR #2359, opened 2024-06-12
chore: Property filter custom forms input dropdown

pan-kot:
### Description

Updates property filter so that tokens with custom forms have their value editable with a custom input dropdown to ensure the value input takes less space.

Rel: Nu26A4Kr5E80

### How has this been tested?

* Unit tests and manual testing
* Consultation with accessibility team

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/property-filter/__tests__/property-filter-extended-operators.test.tsx
+++ b/src/property-filter/__tests__/property-filter-extended-operators.test.tsx
@@ -3,8 +3,13 @@
 
 import React from 'react';
 import { act, render } from '@testing-library/react';
-import createWrapper, { PropertyFilterWrapper } from '../../../lib/components/test-utils/dom';
+import createWrapper, {
+  ElementWrapper,
+  FormFieldWrapper,
+  PropertyFilterWrapper,
+} from '../../../lib/components/test-utils/dom';
 import PropertyFilter from '../../../lib/components/property-filter';
+import DropdownWrapper from '../../../lib/components/test-utils/dom/internal/dropdown';
 import styles from '../../../lib/components/property-filter/styles.selectors.js';
 import { FilteringProperty, PropertyFilterProps, Ref } from '../../../lib/components/property-filter/interfaces.js';
 import { createDefaultProps } from './common';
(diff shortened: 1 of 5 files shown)


connorlanigan:
The dropdown trigger for the custom property value behaves inconsistently with the dropdown triggers for the property name and the operator. When I click the trigger to open the dropdown and then click the trigger again, the dropdown stays open. However, it should close.

Another inconsistency is with the keyboard interaction: when I focus the custom property value trigger and press Space, the dropdown opens. When I then press Shift-Tab to move to the operator trigger, the dropdown of the custom property value stays open. For the other Select elements, the dropdown closes in that case. I recommend that we align this with the behaviour of the Date Range Picker component, which traps the focus in the dropdown.

I wonder if we can potentially also reuse logic here from other components, e.g. [the Select component's `useSelect` hook](https://github.com/cloudscape-design/components/blob/main/src/select/utils/use-select.ts#L166-L185), which also addresses focus handling and keyboard interaction.

pan-kot:
> The dropdown trigger for the custom property value behaves inconsistently with the dropdown triggers for the property name and the operator. When I click the trigger to open the dropdown and then click the trigger again, the dropdown stays open. However, it should close.

When there is a custom form we can consider it a different type of an input. By default the value is edited with an autosuggest but if, say, we added a date picker instead there will be the same inconsistency.

> Another inconsistency is with the keyboard interaction: when I focus the custom property value trigger and press Space, the dropdown opens. When I then press Shift-Tab to move to the operator trigger, the dropdown of the custom property value stays open. For the other Select elements, the dropdown closes in that case. I recommend that we align this with the behaviour of the Date Range Picker component, which traps the focus in the dropdown.

Thanks for that observation. Turns out the tab trap that we already have inside the dropdown is not enough. I added a focus lock same as in date- and date-range pickers.

> I wonder if we can potentially also reuse logic here from other components, e.g. [the Select component's `useSelect` hook](https://github.com/cloudscape-design/components/blob/main/src/select/utils/use-select.ts#L166-L185), which also addresses focus handling and keyboard interaction.

I think the select/autosuggest components are very different. E.g. for the custom form dropdown there are explicit Cancel/Apply buttons and more.

connorlanigan (changes requested):

pan-kot:
Closed in favour of Qeb3AbRK3bNw

PR #2368, opened 2024-06-13
fix: CopyToClipboard is stuck in insecure contexts

connorlanigan:
### Description

When a page runs in an insecure context (e.g. plain HTTP), the browser's clipboard API is not available. In this case, the CopyToClipboard component used to get stuck in the pending state. With this change, the component shows the error state instead.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/copy-to-clipboard/internal.tsx
+++ b/src/copy-to-clipboard/internal.tsx
@@ -29,20 +29,25 @@ export default function InternalCopyToClipboard({
 
   const baseProps = getBaseProps(restProps);
   const onClick = () => {
-    if (navigator.clipboard) {
-      setStatus('pending');
-      setStatusText('');
-      navigator.clipboard
-        .writeText(textToCopy)
-        .then(() => {
-          setStatus('success');
-          setStatusText(copySuccessText);
-        })
-        .catch(() => {
-          setStatus('error');
-          setStatusText(copyErrorText);
-        });
+    if (!navigator.clipboard) {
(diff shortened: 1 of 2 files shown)


gethinwebster (approved):

just-boris (approved):

src/copy-to-clipboard/internal.tsx
```diff
@@ -29,20 +29,25 @@ export default function InternalCopyToClipboard({
...
-          setStatusText(copySuccessText);
-        })
-        .catch(() => {
-          setStatus('error');
-          setStatusText(copyErrorText);
-        });
+    if (!navigator.clipboard) {
+      // The clipboard API is not available in insecure contexts.
+      setStatus('error');
+      setStatusText(copyErrorText);
```
just-boris, line 35: It seems like we have a UX problem of showing the same error text for all cases. This does not allow users to understand how to fix the copying (clicking the button again obviously won't help here)

connorlanigan: Yes, this is a general problem in this component. We also don't expose the error reasonto the developers (whether it's the missing API or a thrown error) . Maybe we should add `console.error` commands in this component (although we're not doing that in any other components, as far as I know).

just-boris: We can merge this as is and raise this question with our team (open a ticket if this use-case is accepted)

PR #2373, opened 2024-06-14
feat: button-group component

spugachev:
### Description

Button Group component is a collection of contextual and persistent triggers. The primary use case is to enable users to perform actions on generative AI outputs. The button triggers all have a consistent variant, might have separators and a button dropdown at the end. Momentarily we only need the icon variant but we expect a future extension with the normal variant. The icon triggers can have tooltips activated with hover or focus and popovers indicating status (like Copy / Copied).

---


pages/button-group/simple.page.tsx
pan-kot, line 1: Please add permutations page. We have it for about every component to run screenshot tests on.

pan-kot: The pages simple and test are very similar - would it make sense to only keep the test one (the name test suggests it is used for integ testing)?

In the page you can feature a few different usage examples to cover the majority of use cases. As the most of the examples will be covered by the permutations page it only worths adding those that are used for integ testing or feature some interactive behaviours that can be used for manually checking the component or used for bug bush.

src/button-group/__tests__/icon-button-item.test.tsx
```diff
@@ -0,0 +1,79 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import React from 'react';
+import { fireEvent, render, screen } from '@testing-library/react';
+import { ButtonGroupProps } from '../../../lib/components/button-group/interfaces';
+import IconButtonItem from '../../../lib/components/button-group/icon-button-item';
```
pan-kot, line 6: It would be great to test all items types: button, group, and menu. It is better to test the all via the root component, for example:

```
test('sample', () => {
  const wrapper = renderButtonGroup({ items });
  const button = wrapper.findButtonById('test-button');
  // test button
})
```

src/button-group/__tests__/menu-dropdown-item.test.tsx
```diff
@@ -0,0 +1,33 @@
...
+        text: 'Group',
+        items: [
+          { id: '5', text: 'Item 5' },
+          { id: '6', text: 'Item 6' },
+        ],
+      },
+    ],
+  };
+
+  test('renders the button', () => {
```
pan-kot, line 28: Let's add more tests to cover the unique behaviours of the menu. It does not make sense to test what is already tested in the button dropdown, so there can be just a simple test to verify that an in-menu item is clickable and fires the onItemClicked as expected. We also need to check the tooltip.

src/button-group/icon-button-item.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+    const onClickHandler = (event: CustomEvent<ClickDetail>) => {
+      if (item.actionPopoverText) {
+        setIsActionPopover(true);
+        setPopoverOpen(true);
+      } else {
+        setPopoverOpen(false);
+      }
+
+      if (onItemClick) {
```
pan-kot, line 36: This IF condition is redundant - the fireCancellableEvent accepts nullable handler.

src/button-group/internal.tsx
```diff
@@ -0,0 +1,81 @@
...
+      items = [],
+      onItemClick,
+      __internalRootRef = null,
+      dropdownExpandToViewport,
+      ...props
+    }: InternalButtonGroupProps,
+    ref: React.Ref<ButtonGroupProps.Ref>
+  ) => {
+    if (!variant) {
+      throw new Error('Invariant violation: ButtonGroup variant is not set.');
```
pan-kot, line 25: I am thinking that maybe it is better to return `null` here and add a dev warnings (with `warnOnce` function) here. This will limit the blast radius in case the variant property will be missed on some accident in prod version. The fallback needs to be covered with unit tests.

src/button-group/menu-dropdown-item.tsx
```diff
@@ -0,0 +1,84 @@
...
+    }: {
+      item: ButtonGroupProps.MenuDropdown;
+      onItemClick?: CancelableEventHandler<ButtonGroupProps.ItemClickDetails>;
+      dropdownExpandToViewport?: boolean;
+    },
+    ref: React.Ref<ButtonDropdownProps.Ref>
+  ) => {
+    const containerRef = useRef<HTMLDivElement>(null);
+    const [popoverOpen, setPopoverOpen] = useState(false);
+    const [isActionPopover, setIsActionPopover] = useState(false);
```
pan-kot, line 26: The popover code is similar between button and menu dropdown, maybe we can move it to the item-element or reuse in some other way?

src/button-group/styles.scss
```diff
@@ -0,0 +1,26 @@
...
+.root {
+  @include styles.styles-reset;
+}
+
+.item {
+  /* used in test-utils */
+}
+
+.group {
+  display: flex;
```
pan-kot, line 17: This one does not maintain the consistent spacing between the items. We can add the matching gap and no-wrap but can try use `display: contents` instead. The group div is needed to assign the ARIA role and label, but we can exclude it from the layout effectively.

pan-kot: Alternatively, we can keep it and use to display dividers which can be defined as inline-end border on the group. In that case we won't need a separate divider element.

src/button-group/tooltip/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+export interface TooltipProps {
+  trackKey: string;
+  trackRef: React.RefObject<HTMLElement | SVGElement>;
+  content: React.ReactNode;
+  open: boolean;
+  close?: () => void;
+}
+
+export default function Tooltip({ trackKey, trackRef, content, open, close }: TooltipProps) {
```
pan-kot, line 20: How is that different compared to the tooltip from src/internal/components? Can we improve that one instead and reuse?

pan-kot (commented):
Awesome work 👍

src/button-group/tooltip/index.tsx
spugachev: - Internal Tooltip accepts only value: number | string and this Tooltip accepts React.ReactNode
 - Internal Tooltip has no "open" property. So it's behavior can't be controlled. 
 
 This component has advantages of  InternalPopover component and internal Tooltip componet. I was thinking of using this two components together but it's hard to make smooth transition between tooltip text and actionPopoverText when using diffrenet components.

spugachev: There is a reason to accept ReactNode. We need to show StatusIndicator component when actionPopoverText is shown and just text when it's regular tooltip

spugachev: There is also incompatibility between components. If value can be of type ReactNode than we need to specify explicit trackKey. Because it can be only string or integer.

PR #2379, opened 2024-06-17
feat: button-group component

spugachev:
### Description

Button Group component is a collection of contextual and persistent triggers. The primary use case is to enable users to perform actions on generative AI outputs. The button triggers all have a consistent variant, might have separators and a button dropdown at the end. Momentarily we only need the icon variant but we expect a future extension with the normal variant. The icon triggers can have tooltips activated with hover or focus and popovers indicating status (like Copy / Copied).

---


pages/button-group/permutations.page.tsx
```diff
@@ -0,0 +1,146 @@
...
+import React from 'react';
+import ScreenshotArea from '../utils/screenshot-area';
+import PermutationsView from '../utils/permutations-view';
+import createPermutations from '../utils/permutations';
+import { ButtonGroup, ButtonGroupProps } from '~components';
+
+const permutations = createPermutations<ButtonGroupProps>([
+  {
+    variant: ['icon'],
+    dropdownExpandToViewport: [true, false],
```
pan-kot, line 12: This is not really needed because in permutation tests we do not run any interactions and the dropdown is only shown upon clicking. Besides, the it is the button dropdown responsible for the dropdown presentation and not button group.

pages/button-group/permutations.page.tsx
```diff
@@ -0,0 +1,146 @@
...
+    items: [
```
pan-kot, line 13: I would create a few separate arrays of items and create permutations with different items. The things to check are e.g.: item-only button group, button group with adjacent groups, with a group separated with actions on both sides, with one show-more dropdown, with multiple show-more dropdowns, with show-more dropdown inside a group, etc.

PR #2382, opened 2024-06-17
chore: remove redundant act from unit tests in a* components

just-boris:
### Description

Remove not needed code

### How has this been tested?

PR build passes

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/__tests__/desktop.test.tsx
+++ b/src/app-layout/__tests__/desktop.test.tsx
@@ -1,7 +1,7 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 import React from 'react';
-import { act, screen, within } from '@testing-library/react';
+import { act, fireEvent, screen, within } from '@testing-library/react';
 
 import {
   describeEachThemeAppLayout,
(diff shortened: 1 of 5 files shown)


src/app-layout/__tests__/desktop.test.tsx
```diff
@@ -253,17 +253,17 @@ describe('Classic only features', () => {
     const { wrapper } = renderComponent(
       <AppLayout toolsHide={true} drawers={[testDrawer]} ariaLabels={{ drawers: 'Drawers' }} />
     );
-    act(() => screen.getByLabelText('Drawers').click());
+    fireEvent.click(screen.getByLabelText('Drawers'));
```
georgylobko, line 256: What is the benefit of using `fireEvent.click` over `click` in this case?

BTW, it might worth to consider `user-event` lib as an alternative for `fireEvent`, but it's definitely out of this PR scope
https://testing-library.com/docs/guide-events/#alternatives

georgylobko (approved):
Good job! Left a minor comment

src/app-layout/__tests__/desktop.test.tsx
just-boris: `.click()` is `HTMLElement.prototype.click` without any extra logic. Both our test-utils' `wrapper.click()` and `fireEvent.click()` provide built-in `act` integration so extra wrapper is not needed here.

Using `fireEvent.click` is not ideal, because we recommend everyone using our test utils and here we undermine our own recommendation. Using `userEvent.click` moves us even further from our own test-utils, so definitely not recommended

just-boris: If you have more questions, I think we can chat more in our team channel for better knowledge sharing

PR #2409, opened 2024-06-25
feat: Disabled reason for Button and ButtonDropdown

georgylobko:
### Description

Introduced disabled reason for Button and ButtonDropdown components

Related links, issue #, if available: n/a

### How has this been tested?

<details>
   <summary>Review checklist</summary>

_The following items are to be evaluated by the author(s) and the reviewer(s)._

#### Correctness

- _Changes include appropriate documentation updates._
- _Changes are backward-compatible if not indicated, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#public-apis)._
- _Changes do not include unsupported browser features, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#browsers-support)._
- _Changes were manually tested for accessibility, see [accessibility guidelines](https://cloudscape.design/foundation/core-principles/accessibility/)._

#### Security

- _If the code handles URLs: all URLs are validated through [the `checkSafeUrl` function](https://github.com/cloudscape-design/components/blob/main/src/internal/utils/check-safe-url.ts)._

#### Testing

- _Changes are covered with new/existing unit tests?_
- _Changes are covered with new/existing integration tests?_
</details>

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2412, opened 2024-06-25
feat: Disabled reason for Select and Multiselect

georgylobko:
### Description

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/multiselect/disabled-reason.page.tsx
+++ b/pages/multiselect/disabled-reason.page.tsx
@@ -1,0 +1,40 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import * as React from 'react';
+import Multiselect, { MultiselectProps } from '~components/multiselect';
+import Box from '~components/box';
+import ScreenshotArea from '../utils/screenshot-area';
+
+const options: MultiselectProps.Options = [
+  { value: 'first', label: 'Simple' },
+  { value: 'second', label: 'With small icon', iconName: 'folder' },
+  {
+    value: 'third',
+    label: 'With big icon icon',
+    description: 'Very big option',
+    iconName: 'heart',
+    disabled: true,
+    disabledReason: 'disabled reason',
+  },
+  {
+    label: 'Option group',
+    options: [{ value: 'forth', label: 'Nested option' }],
+    disabled: true,
+    disabledReason: 'disabled reason',
+  },
+];
+
+export default function SelectPage() {
+  return (
+    <article>
+      <ScreenshotArea>
+        <Box variant="h1">Multiselect with disabled reason</Box>
+        <Box padding="l">
+          <div style={{ width: '400px' }}>
+            <Multiselect placeholder="Choose option" selectedOptions={[]} options={options} />
+          </div>
+        </Box>
+      </ScreenshotArea>
+    </article>
+  );
+}
--- a/src/internal/components/selectable-item/index.tsx
+++ b/src/internal/components/selectable-item/index.tsx
@@ -22,6 +22,7 @@ export type SelectableItemProps = BaseComponentProps & {
   ariaPosinset?: number;
   ariaSetsize?: number;
   highlightType?: HighlightType['type'];
+  'aria-describedby'?: string;
 } & ({ ariaSelected?: boolean; ariaChecked?: never } | { ariaSelected?: never; ariaChecked?: boolean | 'mixed' });
 
 const SelectableItem = (
@@ -108,6 +109,10 @@ const SelectableItem = (
   if (ariaPosinset && ariaSetsize) {
     a11yProperties['aria-posinset'] = ariaPosinset;
     a11yProperties['aria-setsize'] = ariaSetsize;
+  }
+
+  if (restProps['aria-describedby']) {
+    a11yProperties['aria-describedby'] = restProps['aria-describedby'];
   }
 
   return (
--- a/src/internal/components/tooltip/index.tsx
+++ b/src/internal/components/tooltip/index.tsx
@@ -9,15 +9,23 @@ import PopoverBody from '../../../popover/body';
 import Portal from '../portal';
 import popoverStyles from '../../../popover/styles.css.js';
 import styles from './styles.css.js';
+import { PopoverProps } from '../../../../lib/components';
 
 export interface TooltipProps {
   value: number | string;
   trackRef: React.RefObject<HTMLElement | SVGElement>;
+  position?: PopoverProps.Position;
   className?: string;
   contentAttributes?: React.HTMLAttributes<HTMLDivElement>;
 }
 
-export default function Tooltip({ value, trackRef, className, contentAttributes = {} }: TooltipProps) {
+export default function Tooltip({
+  value,
+  trackRef,
+  className,
+  contentAttributes = {},
+  position = 'top',
+}: TooltipProps) {
   return (
     <Portal>
       <div className={clsx(styles.root, className)} {...contentAttributes}>
@@ -28,7 +36,7 @@ export default function Tooltip({ value, trackRef, className, contentAttributes 
               trackKey={value}
               size="small"
               fixedWidth={false}
-              position="top"
+              position={position}
               arrow={position => (
                 <div className={clsx(popoverStyles.arrow, popoverStyles[`arrow-position-${position}`])}>
                   <div className={popoverStyles['arrow-outer']} />
--- a/src/internal/components/tooltip/index.tsx
+++ b/src/internal/components/tooltip/index.tsx
@@ -7,9 +7,9 @@ import { Transition } from '../transition';
 import PopoverContainer from '../../../popover/container';
 import PopoverBody from '../../../popover/body';
 import Portal from '../portal';
+import { PopoverProps } from '../../../popover/interfaces';
 import popoverStyles from '../../../popover/styles.css.js';
 import styles from './styles.css.js';
-import { PopoverProps } from '../../../../lib/components';
 
 export interface TooltipProps {
   value: number | string;
--- a/src/internal/components/tooltip/index.tsx
+++ b/src/internal/components/tooltip/index.tsx
@@ -13,13 +13,14 @@ import styles from './styles.css.js';
 export interface TooltipProps {
   value: number | string;
   trackRef: React.RefObject<HTMLElement | SVGElement>;
+  className?: string;
   contentAttributes?: React.HTMLAttributes<HTMLDivElement>;
 }
 
-export default function Tooltip({ value, trackRef, contentAttributes = {} }: TooltipProps) {
+export default function Tooltip({ value, trackRef, className, contentAttributes = {} }: TooltipProps) {
   return (
     <Portal>
-      <div className={styles.root} {...contentAttributes}>
+      <div className={clsx(styles.root, className)} {...contentAttributes}>
         <Transition in={true}>
           {() => (
             <PopoverContainer
--- a/src/select/__tests__/select.test.tsx
+++ b/src/select/__tests__/select.test.tsx
@@ -1,6 +1,7 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 import * as React from 'react';
+import ReactDOM from 'react-dom';
 import { render, waitFor } from '@testing-library/react';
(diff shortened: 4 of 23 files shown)


pages/multiselect/disabled-reason.page.tsx
```diff
@@ -0,0 +1,40 @@
...
+    label: 'Option group',
+    options: [{ value: 'forth', label: 'Nested option' }],
+    disabled: true,
+    disabledReason: 'disabled reason',
+  },
+];
+
+export default function MultiselectPage() {
+  return (
+    <article>
```
just-boris, line 29: I do not think this element is needed. Same on other pages

src/internal/components/tooltip/index.tsx
```diff
@@ -7,27 +7,36 @@ import { Transition } from '../transition';
 import PopoverContainer from '../../../popover/container';
 import PopoverBody from '../../../popover/body';
 import Portal from '../portal';
+import { PopoverProps } from '../../../popover/interfaces';
```
just-boris, line 10: Depending internal component on public feels wrong.

How about moving the type here and re-export it in popover? See [`OptionDefinition`](https://github.com/search?q=repo%3Acloudscape-design%2Fcomponents%20optiondefinition&type=code) for a similar treatment

src/select/__tests__/select.test.tsx
```diff
@@ -450,6 +451,162 @@ describe.each([false, true])('expandToViewport=%s', expandToViewport => {
       wrapper.openDropdown();
       expect(wrapper.findDropdown({ expandToViewport })?.findOpenDropdown()).toBeFalsy();
     });
+
+    describe('Disabled item with reason', () => {
+      beforeEach(() => {
+        jest.spyOn(ReactDOM, 'createPortal').mockImplementation((element: any) => element);
```
just-boris, line 457: Same as in another thread https://github.com/cloudscape-design/components/pull/2409#discussion_r1656631973

Need a better solution

src/select/__tests__/select.test.tsx
```diff
@@ -450,6 +451,162 @@ describe.each([false, true])('expandToViewport=%s', expandToViewport => {
...
+            if (index === 0) {
+              return {
+                ...item,
+                disabled: true,
+              };
+            }
+
+            return item;
+          }),
+        });
```
just-boris, line 494: I think at this point it is simpler to create your own copy of props than deriving them from defailt.

Compare:

```js
renderSelect({
   option: [{ label: 'Disabled for no reason ', value: 'foo', disabled: true }]
})
``` 

Why is extra complexity?

(same applies to all `defaultOptions.map` in this file)

src/select/__tests__/select.test.tsx
```diff
@@ -450,6 +451,162 @@ describe.each([false, true])('expandToViewport=%s', expandToViewport => {
...
+        wrapper.openDropdown();
+
+        expect(wrapper.findDropdown().findSelectableItem(1)!.getElement()).toHaveAttribute('aria-describedby');
+      });
+
+      test('has hidden element (linked to aria-describedby) with disabledReason', () => {
```
just-boris, line 569: Can this be merged with another test rendering the same props configuration?

src/select/__tests__/select.test.tsx
```diff
@@ -450,6 +451,162 @@ describe.each([false, true])('expandToViewport=%s', expandToViewport => {
...
+                disabledReason: 'disabled reason',
+              };
+            }
+
+            return item;
+          }),
+        });
+        wrapper.openDropdown();
+        wrapper.selectOptionByValue('1');
+        expect(onChange).not.toHaveBeenCalled();
```
just-boris, line 607: Should we also assert if the dropdown closes or not?

src/select/parts/item.tsx
```diff
@@ -50,6 +53,14 @@ const Item = (
   const isChild = option.type === 'child';
   const wrappedOption: OptionDefinition = option.option;
   const disabled = option.disabled || wrappedOption.disabled;
+  const disabledReason =
+    disabled && (option.disabledReason || wrappedOption.disabledReason)
+      ? option.disabledReason || wrappedOption.disabledReason
+      : '';
```
just-boris, line 59: Why do you need to check both `option` and `wrappedOption`?

src/test-utils/dom/internal/dropdown-host.ts
```diff
@@ -154,6 +155,18 @@ export class DropdownContentWrapper extends ComponentWrapper {
...
+   * Returns a selectable item (option wrapper) from the dropdown.
+   *
+   * @param optionIndex 1-based index of the option to select.
+   */
+  findSelectableItem(optionIndex: number): SelectableItemWrapper | null {
+    return this.findComponent(
+      `.${selectableStyles['selectable-item']}[data-test-index="${optionIndex}"]`,
+      SelectableItemWrapper
+    );
+  }
```
just-boris, line 168: There is already `.findDropdown().findOptionByValue(value)` and `.findDropdown().findOption(index)` utils.

Why does this code not use the same finders?

src/internal/components/tooltip/index.tsx
georgylobko: We use public component `Popover` inside the internal and the appropriate interface for it. How it could be beneficial to duplicate the props?

just-boris: This looks like a circular dependency, popover extends the tooltip, so all dependencies between these two components should be one direction

pages/multiselect/disabled-reason.page.tsx
georgylobko: updated

src/select/__tests__/select.test.tsx
georgylobko: updated

src/select/__tests__/select.test.tsx
georgylobko: Technically yes, but I followed "one test one assertion" pattern to keep things easier to debug if something goes wrong

src/select/__tests__/select.test.tsx
georgylobko: No. If the dropdown is closed, the select option is not shown in the screen anymore

src/select/parts/item.tsx
georgylobko: Good catch. Only `wrappedOption` could be checked. Updated

src/select/__tests__/select.test.tsx
just-boris: Not a strong opinion here, but I think deduplicating tests actually helps than reviewing multiple failed tests with the same setup config

src/select/__tests__/select.test.tsx
just-boris: Can we write an explicit assertion (or a separate test maybe): "after clicking on the disabled option, the dropdown stays open"?

src/test-utils/dom/internal/dropdown-host.ts
georgylobko: The selectable item structure is the following:
```javascript
<SelectableItem>
  <Option />
  ...
  disabled state related content
</SelectableItem>
```

These methods find child component `Option`, while the new one finds the root element to have access to disabled state content. It was implemented this way to avoid props drilling to `Option` component

just-boris: Need to look at it from customer perspective.

Imagine you already have a bunch of tests using `findOptionByValue`. How much will they be surprised that `findOptionByValue().findDisabledReason()` does not work?

I think for this part we need to get back and revisit the API design to offer something more convenient to our users.

src/internal/components/tooltip/index.tsx
georgylobko: It would be a circular dependency if the `Tooltip` component was used within the `Popover`, which is not a case. Nevertheless, moved the type in-place:

```javascript
position?: 'top' | 'right' | 'bottom' | 'left';
```

src/select/__tests__/select.test.tsx
georgylobko: We can, but how it could help us? If disabled option closes the dropdown, it means that `onChange` handler triggers as well, so this tests fails.

src/test-utils/dom/internal/dropdown-host.ts
georgylobko: I forgot to mention one important detail: `Option` component renders conditionally within `SelectableItem`, this way:
```javascript
{isParent ? (
  <span>{wrappedOption.label || wrappedOption.value}</span>
) : (
  <Option
    option={{ ...wrappedOption, disabled }}
    highlightedOption={highlighted}
    selectedOption={selected}
    highlightText={filteringValue}
  />
)}
```
If we put disabled content inside `Option` and make `findOptionByValue().findDisabledReason()` work, disabled reason won't be shown in the parent item, which is not a case. 

So we have 2 options here:
1. change `findOptionByValue` implementation to find `SelectableItem` instead of `Option`. Might make sense, but at the same time exposes additional risks, because it's a customer facing API
2. leave things as is, which is less risky. we don't know is customers will be surprised, even if they do, they can refer to testing documentation and solve this problem.

WDYT?

just-boris: I recommend putting yourself in shoes of customer and try writing some sample tests in a way how our customers use that (not for release in this PR, but just to get a sense of the API you are proposing).

When you have these tests we can look again at your PR.

georgylobko: After some research came to realize that `findDisabledReason()` could be implemented within `OptionWrapper`, because they are requested via `createWrapper()` anyway (they are rendered through react portals)

So now test utils look the following way:

```javascript
expect(
  wrapper.findDropdown().findOption(1)!.findDisabledReason()!.getElement()
).toHaveTextContent('disabled reason');
```

src/select/__tests__/select.test.tsx
just-boris: My question is not much connected to `onChange` function but testing clicking on a disabled item in general.

src/internal/components/tooltip/index.tsx
just-boris: I see that there were popover imports in this file before, but it is still good to not exacerbate the issue.

You may also want to talk with other team members about possible refactoring here

src/internal/components/selectable-item/index.tsx
```diff
@@ -22,6 +22,7 @@ export type SelectableItemProps = BaseComponentProps & {
   ariaPosinset?: number;
   ariaSetsize?: number;
   highlightType?: HighlightType['type'];
+  'aria-describedby'?: string;
```
just-boris, line 25: There is a type safety issue: https://github.com/microsoft/TypeScript/issues/32447

`aria-` and `data-` attributes are not type checked. This is why we use camel case names for all other aria props

Sorry for not spotting it earlier, but I think you want to address this issue

just-boris (commented):
Found one more issue, but other than that, can approve: https://github.com/cloudscape-design/components/pull/2412/files?w=1#r1658432407

src/internal/components/selectable-item/index.tsx
georgylobko: updated

src/select/__tests__/select.test.tsx
georgylobko: added a new test

src/select/__tests__/select.test.tsx
```diff
@@ -450,6 +450,87 @@ describe.each([false, true])('expandToViewport=%s', expandToViewport => {
...
+      });
+
+      test('click on disabled with reason option does not close dropdown', () => {
+        const { wrapper } = renderSelect({
+          options: [{ label: 'First', value: '1', disabled: true, disabledReason: 'disabled reason' }],
+        });
+        wrapper.openDropdown();
+        wrapper.selectOptionByValue('1', { expandToViewport });
+        expect(wrapper.findDropdown({ expandToViewport }).getElement()).not.toBe(null);
```
just-boris, line 531: `wrapper.findDropdown()` returns true even if it closed. There is a separate `findOpenDropdown` for that

just-boris (dismissed):

src/select/__tests__/select.test.tsx
georgylobko: updated

just-boris (approved):

timogasda (approved):

PR #2413, opened 2024-06-25
feat: Disabled reason for Tabs

georgylobko:
### Description

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/tabs/__tests__/tabs.test.tsx
```diff
@@ -752,6 +796,272 @@ describe('Tabs', () => {
...
+              }
+
+              return item;
+            })}
+            activeTabId={firstTabId}
+            onChange={() => void 0}
+          />
+        );
+
+        expect(wrapper.findTabLinkById('second')!.find('span[hidden]')!.getElement()).toHaveTextContent(
```
pan-kot, line 1036: Would it make sense to merge this test with previous one?

Instead of checking the presence of aria-describedby you can do:
```
const describedBy = wrapper.findTabLinkById('second').getElement().getAttribute('aria-describedby');
const description = wrapper.findTabLinkById('second').find(`#${descrbedBy}`).getElement().textContent;
```

src/tabs/__tests__/tabs.test.tsx
```diff
@@ -752,6 +796,272 @@ describe('Tabs', () => {
...
+          'disabled reason'
+        );
+      });
+
+      test('show previously selected tab content when disabled with reason tab is focused', () => {
```
pan-kot, line 1041: nit: I'd phrase it as "shows active tab content when disabled with reason tab is focused" (it if actually works that way). Otherwise, the previously selected tab can be another tab with disabled reason.

src/tabs/tab-header-bar.scss
```diff
@@ -162,7 +162,6 @@ $label-horizontal-spacing: awsui.$space-l;
 .tabs-tab-disabled {
   &,
   &:hover {
-    pointer-events: none;
```
pan-kot, line 165: Does this also add pointer events to disabled tabs with no disabled reason? Is that expected?

src/tabs/tab-header-bar.tsx
```diff
@@ -194,13 +201,18 @@ export function TabHeaderBar({
...
       return navigationAPI.current?.isRegistered(element) ?? false;
     }
-    function isElementDisabled(element: HTMLElement) {
+    function isElementFocusable(element: HTMLElement) {
       if (element instanceof HTMLButtonElement) {
         return element.disabled && element.getAttribute('aria-selected') !== 'true';
       }
+      if (element instanceof HTMLAnchorElement) {
+        // aria-describedby is present only if the element is disabled with reason. in this case it should be focusable, otherwise no
+        return element.getAttribute('aria-disabled') === 'true' && !element.hasAttribute('aria-describedby');
```
pan-kot, line 210: Would it make sense to add some `tab-focusable` style to focusable tabs and simplify the check here?

src/tabs/tab-header-bar.tsx
```diff
@@ -194,13 +201,18 @@ export function TabHeaderBar({
     function isElementRegistered(element: HTMLElement) {
       return navigationAPI.current?.isRegistered(element) ?? false;
     }
-    function isElementDisabled(element: HTMLElement) {
+    function isElementFocusable(element: HTMLElement) {
```
pan-kot, line 204: The function name suggests it returns `true` when the tab is focusable but it does the opposite. Either the function name or the logic inside should be changed.

src/tabs/tab-header-bar.tsx
```diff
@@ -299,11 +313,13 @@ export function TabHeaderBar({
       'aria-controls': `${idNamespace}-${tab.id}-panel`,
       'data-testid': tab.id,
       id: getTabElementId({ namespace: idNamespace, tabId: tab.id }),
-      children: <span className={styles['tabs-tab-label']}>{tab.label}</span>,
     };
 
     if (tab.disabled) {
       commonProps['aria-disabled'] = 'true';
+      commonProps.onClick = (event: React.MouseEvent) => {
+        event.preventDefault();
```
pan-kot, line 321: why this is needed?

src/tabs/tab-header-bar.tsx
```diff
@@ -332,12 +348,55 @@ const TabTrigger = forwardRef(
     ref: React.Ref<HTMLElement>
   ) => {
     const refObject = useRef<HTMLElement>(null);
+    const childRefObject = useRef<HTMLElement>(null);
```
pan-kot, line 351: Would it make sense to rename the refs to e.g. `tabRefObject` and `tabLabelRefObject` for better semantic meaning?

pan-kot: Why an extra ref is needed at all, can the refObject be used instead?

src/tabs/tab-header-bar.tsx
```diff
@@ -332,12 +348,55 @@ const TabTrigger = forwardRef(
...
+    const handlers = {
+      onFocus: isDisabledWithReason ? () => setShowTooltip(true) : undefined,
+      onBlur: isDisabledWithReason ? () => setShowTooltip(false) : undefined,
+      onMouseEnter: isDisabledWithReason ? () => setShowTooltip(true) : undefined,
+      onMouseLeave: isDisabledWithReason ? () => setShowTooltip(false) : undefined,
+    };
+
+    const commonProps = {
+      ...elementProps,
+      ...(isDisabledWithReason ? targetProps : {}),
```
pan-kot, line 386: Can we add handlers the same way for consistency? Like:

```
const commonProps = {
  ...elementProps,
  ...(isDisabledWithReason ? targetProps : {}),
  ...(isDisabledWithReason ? handlers : {}),
  ref: mergedRef,
  tabIndex
};
```

src/tabs/__tests__/tabs.test.tsx
georgylobko: Makes sense. Updated

src/tabs/__tests__/tabs.test.tsx
georgylobko: Updated

src/tabs/tab-header-bar.scss
georgylobko: No, it does not, because tabs with no disabled reason has attribute `disabled`, which prevents events by default.

src/tabs/tab-header-bar.tsx
georgylobko: Checked once again, it is not needed. Updated

georgylobko: upd: it's needed when a tab with href is disabled

pan-kot: let's add a code comment then to make it transparent

src/tabs/tab-header-bar.tsx
georgylobko: Negate the logic inside the function and left name as `isElementFocusable`. Moreover, added new conditions, because the recently merged feature "tabs with actions" changed the rules

src/tabs/tab-header-bar.tsx
georgylobko: made this even simpler by removing this additional handler and attaching existing handler unconditionally (regardless of being `disabled`), because it prevents default for disabled tabs anyway

https://github.com/cloudscape-design/components/blob/feat/disabled-reasons-tabs/src/tabs/tab-header-bar.tsx#L401

src/tabs/tab-header-bar.tsx
georgylobko: updated

src/tabs/tab-header-bar.tsx
georgylobko: updated

src/tabs/__tests__/tabs.test.tsx
pan-kot: I don't see this change in code, could you share a link to the line please?

georgylobko: Sure, here you go:

https://github.com/cloudscape-design/components/pull/2413/files#diff-fc85e49981e81cc2dbe0ff658349f22e0fdb03d247a6aa3370b4fd828c8cbb6eR1133

But there are more test suites for these a11y attributes above which could be combined.

pan-kot: Ah, so what I meant is that not only the tests can be merged but you can also ensure that the describedby points to the correct element. See code above

src/tabs/tab-header-bar.tsx
```diff
@@ -254,13 +261,20 @@ export function TabHeaderBar({
     function isElementRegistered(element: HTMLElement) {
       return navigationAPI.current?.isRegistered(element) ?? false;
     }
-    function isElementDisabled(element: HTMLElement) {
+    function isElementFocusable(element: HTMLElement) {
       if (element instanceof HTMLButtonElement) {
-        return element.disabled || element.closest(`.${styles['tabs-tab-disabled']}`);
+        return !(
+          element.disabled ||
+          (element.closest(`.${styles['tabs-tab-disabled']}`) && !element.closest(`.${styles['tabs-tab-focusable']}`))
```
pan-kot, line 268: Would it make sense to attach the disabled style whenever the element is disabled regardless of whether it has a reason or not? Then it should be enough to only check styles and do not check the disabled attribute.

src/tabs/tab-header-bar.tsx
```diff
@@ -254,13 +261,20 @@ export function TabHeaderBar({
     function isElementRegistered(element: HTMLElement) {
       return navigationAPI.current?.isRegistered(element) ?? false;
     }
-    function isElementDisabled(element: HTMLElement) {
+    function isElementFocusable(element: HTMLElement) {
```
pan-kot, line 264: I think the purpose of the code inside the function is not quite apparent and it can be difficult to relate it to the disabled reasons feature. I suggest adding a small code comment to explain the need.

src/tabs/__tests__/tabs.test.tsx
georgylobko: now I got it! tbh, I don't think it makes much sense, because the purpose of this test is to check presence of the hidden element with certain content, and the way how it's associated with the id is handled by `useHiddenDescription` hook, so this case is rather internal implementation of it

pan-kot: I would not agree that it is an internal implementation because the only presence of the hidden element does not ensure this information is announced. The element must be connected with the aria-describedby. The hook helps with establishing that connection but does not guarantee that (e.g. if you don't pass the ID to the correct element or accidentally override it with another ID).

src/tabs/tab-header-bar.tsx
pan-kot: The tab can be disabled when there is no reason, but it can have `tabs-tab-disabled` in both cases. Yet, it should only have `tabs-tab-focusable` when it is either not disabled or disabled with disabled reason.

pan-kot: Which actually means that it should be enough to use `tabs-tab-focusable` if it is only present when the tab is actually focusable no matter why.

georgylobko: Made it even simpler 

```javascript
return !element.disabled || element.closest(focusableTabSelector);
```

src/tabs/__tests__/tabs.test.tsx
georgylobko: updated. introduced `data-describedby-id`, because requesting elements via `.find('#some-id')` is not a valid way 

> accidentally override it with another ID
I would not consider it, because the dom element representing the description is encapsulated within `useHiddenDescription` hook, so there is not accidental way to override it

pan-kot: Then you can use `.find('[id="your-id"]')`

georgylobko: Didn't know it was possible :)
Updated

pan-kot (approved):

PR #2414, opened 2024-06-25
feat: Disabled reason for SegmentedControl

georgylobko:
### Description

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/segmented-control/internal-segmented-control.tsx
```diff
@@ -21,7 +21,9 @@ export default function InternalSegmentedControl({
     return option.id === selectedId;
   });
   const currentSelectedOption = selectedOptions.length ? selectedOptions[0] : null;
-  const enabledSegments = (options || []).filter(option => !option.disabled);
+  const enabledSegments = (options || []).filter(
```
jperals, line 24: Suggestion: rename `enabledSegments` to `focusableSegments`

src/test-utils/dom/segmented-control/index.ts
```diff
@@ -13,4 +19,13 @@ export default class SegmentedControlWrapper extends ComponentWrapper {
   findSelectedSegment(): ElementWrapper | null {
     return this.findByClassName(styles.selected);
   }
+
+  /**
+   * Finds the segment with the given ID and returns the element.
```
jperals, line 24: Nitpick, non-blocking but I find "the element" a bit unnecessary. What about:

```Finds the segment with the given ID and returns it.```

or

```Returns the segment with the given ID.```

or just

```Finds the segment with the given ID.```

File: src/segmented-control/__tests__/desktop.test.tsx [hunk 3]
```diff
@@ -115,6 +116,295 @@ describe('Segment disabled property', () => {
     getSegmentWrapper(segmentedControlWrapper, 0).getElement().click();
     expect(onChange).toHaveBeenCalledWith(expect.objectContaining({ detail: { selectedId: 'seg-1' } }));
   });
+
+  describe('Disabled with reason', () => {
```
jperals, line 120: What about the case of using `disabledReason` without `disabled` set to `true`? Although this is not a "valid" use case, we want to have an established behavior for it (which to my understanding is the same as not passing any `disabledReason`) which shouldn't change, so it might be a good idea to cover this case as well.

And maybe we also want to emit a dev warning in such case?

src/segmented-control/internal-segmented-control.tsx
```diff
@@ -49,7 +51,7 @@ export default function InternalSegmentedControl({
       {options &&
         options.map((option: SegmentedControlProps.Option, index) => {
           const isActive = selectedId === option.id;
-          const enabledSegmentIndex = enabledSegments.indexOf(option);
+          const enabledSegmentIndex = focusableSegments.indexOf(option);
```
jperals, line 54: Guess it would be good to rename `enabledSegmentIndex` to `focusableSegemtIndex` too :)

src/segmented-control/__tests__/desktop.test.tsx
jperals: I see that the PR was updated but this in particular wasn't addressed

georgylobko: I introduced this test case: https://github.com/cloudscape-design/components/pull/2414/commits/ee0e639c21aeff83e5395125342cd387ab41737b#diff-7d2444d48fe7dd36eb8a4e838395cff5269219a361dca51ea45a9e5c572c02dbR121

jperals: Oh right, my bad

jperals (approved):

jperals (approved):

PR #2419, opened 2024-06-26
feat: Support breadcrumb instances deduplication

just-boris:
### Description

Deduplicate multiple `BreadcrumbGroup` instances across the page and render only one inside the respective app layout slot. Only behind `appLayoutWidget` feature flag, not for production use.

```js
<AppLayout
   breadcrumbs={null}
   content={<>
       <BreadcrumbGroup /> // this one will be rendered in the intended place, even if it did not use the slot
   </>}
/>
```

See the tests for more details about order resolution and other corner-cases.

### How has this been tested?

Added unit and integration tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/app-layout/global-breadcrumbs.page.tsx
+++ b/pages/app-layout/global-breadcrumbs.page.tsx
@@ -1,0 +1,45 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import React, { useState } from 'react';
+import AppLayout from '~components/app-layout';
+import SpaceBetween from '~components/space-between';
+import labels from './utils/labels';
+import BreadcrumbGroup from '~components/breadcrumb-group';
+
+export default function () {
+  const [extraBreadcrumb, setExtraBreadcrumb] = useState(false);
+  return (
+    <AppLayout
+      ariaLabels={labels}
+      breadcrumbs={
+        <BreadcrumbGroup
+          items={[
+            { text: 'Default', href: '#' },
+            { text: 'one', href: '#' },
+          ]}
+        />
+      }
+      content={
+        <SpaceBetween size="m">
+          <h1>Page with additional breadcrumbs</h1>
+          <label>
+            <input
+              type="checkbox"
+              data-testid="toggle-extra-breadcrumb"
+              onChange={event => setExtraBreadcrumb(event.target.checked)}
+            />{' '}
+            Extra breadcrumb
+          </label>
+          {extraBreadcrumb && (
+            <BreadcrumbGroup
+              items={[
+                { text: 'Dynamic', href: '#' },
+                { text: 'instance', href: '#' },
+              ]}
+            />
+          )}
+        </SpaceBetween>
+      }
+    />
+  );
+}
--- a/src/app-layout/visual-refresh-toolbar/index.tsx
+++ b/src/app-layout/visual-refresh-toolbar/index.tsx
@@ -24,6 +24,7 @@ import {
   AppLayoutToolbar,
 } from './internal';
 import { AppLayoutInternals } from './interfaces';
+import { useGetGlobalBreadcrumbs } from '../../internal/plugins/helpers/use-global-breadcrumbs';
 
 const AppLayoutVisualRefreshToolbar = React.forwardRef(
   (
@@ -176,9 +177,16 @@ const AppLayoutVisualRefreshToolbar = React.forwardRef(
       splitPanelPosition: splitPanelPreferences?.position,
     });
 
-    const hasToolbar =
+    const discoveredBreadcrumbs = useGetGlobalBreadcrumbs();
+
+    const hasToolbar = Boolean(
       !embeddedViewMode &&
-      (!!resolvedNavigation || !!breadcrumbs || splitPanelToggleConfig.displayed || drawers!.length > 0);
+        (resolvedNavigation ||
+          breadcrumbs ||
+          discoveredBreadcrumbs ||
+          splitPanelToggleConfig.displayed ||
+          drawers!.length > 0)
+    );
 
     const verticalOffsets = computeVerticalLayout({
       topOffset: placement.insetBlockStart,
@@ -193,6 +201,7 @@ const AppLayoutVisualRefreshToolbar = React.forwardRef(
       headerVariant,
       isMobile,
       breadcrumbs,
+      discoveredBreadcrumbs,
       stickyNotifications,
       navigationOpen,
       navigation: resolvedNavigation,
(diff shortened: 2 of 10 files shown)


src/app-layout/visual-refresh-toolbar/index.tsx
```diff
@@ -176,9 +177,15 @@ const AppLayoutVisualRefreshToolbar = React.forwardRef(
       splitPanelPosition: splitPanelPreferences?.position,
     });
 
+    const discoveredBreadcrumbs = useGetGlobalBreadcrumbs();
+
     const hasToolbar =
       !embeddedViewMode &&
-      (!!resolvedNavigation || !!breadcrumbs || splitPanelToggleConfig.displayed || drawers!.length > 0);
+      (!!resolvedNavigation ||
+        !!breadcrumbs ||
```
just-boris, line 185: F

src/app-layout/visual-refresh-toolbar/toolbar/index.tsx
```diff
@@ -106,8 +108,11 @@ export function AppLayoutToolbarImplementation({ appLayoutInternals }: AppLayout
             />
           </nav>
         )}
-        {breadcrumbs && (
-          <div className={clsx(styles['universal-toolbar-breadcrumbs'], testutilStyles.breadcrumbs)}>{breadcrumbs}</div>
+        {(breadcrumbs || discoveredBreadcrumbs) && (
+          <div className={clsx(styles['universal-toolbar-breadcrumbs'], testutilStyles.breadcrumbs)}>
+            {breadcrumbs}
+            {discoveredBreadcrumbs && <InternalBreadcrumbGroup {...discoveredBreadcrumbs} />}
+          </div>
```
just-boris, line 115: The `breadcrumbs` value here is also deduplicated into `discoveredBreadcrumbs` same way as any other instance on the page. No special treatment for "own breadcrumbs"

src/breadcrumb-group/index.tsx
```diff
@@ -5,14 +5,21 @@ import { BreadcrumbGroupProps } from './interfaces';
...
+import { useSetGlobalBreadcrumbs } from '../internal/plugins/helpers/use-global-breadcrumbs';
 
 export { BreadcrumbGroupProps };
 
 export default function BreadcrumbGroup<T extends BreadcrumbGroupProps.Item = BreadcrumbGroupProps.Item>({
   items = [],
   ...props
 }: BreadcrumbGroupProps<T>) {
+  const registeredGlobally = useSetGlobalBreadcrumbs({ items, ...props });
   const baseComponentProps = useBaseComponent('BreadcrumbGroup');
```
just-boris, line 17: Usage metrics are still emitted for each instance individually

src/internal/plugins/helpers/use-global-breadcrumbs.ts
```diff
@@ -0,0 +1,51 @@
...
+  });
+
+  return registered;
+}
+
+export function useSetGlobalBreadcrumbs<T extends BreadcrumbGroupProps.Item>(props: BreadcrumbGroupProps<T>) {
+  // avoid additional side effects when this feature is not active
+  if (!getGlobalFlag('appLayoutWidget')) {
+    return false;
+  }
```
just-boris, line 35: This guard prevents any changes for the current production behavior

src/internal/plugins/helpers/use-global-breadcrumbs.ts
```diff
@@ -0,0 +1,51 @@
...
+    registrationRef.current = registration;
+
+    return () => {
+      registration.cleanup();
+    };
+    // subsequent prop changes are handled by another effect
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  useLayoutEffect(() => {
```
connorlanigan, line 24: Since `useLayoutEffect` can have a negative performance impact, (why) do we need to use it?

src/app-layout/visual-refresh-toolbar/index.tsx
```diff
@@ -176,9 +177,15 @@ const AppLayoutVisualRefreshToolbar = React.forwardRef(
...
+        !!discoveredBreadcrumbs ||
```
connorlanigan, line 186: I think we don't need the double exclamation marks here, and they make the code difficult to read. We could make this block more readable by removing them, and converting only the result to a boolean:
```
const hasToolbar = Boolean(
      !embeddedViewMode &&
      (resolvedNavigation ||
        breadcrumbs ||
        discoveredBreadcrumbs ||
        splitPanelToggleConfig.displayed ||
        drawers!.length > 0)
);
```

pages/app-layout/global-breadcrumbs.page.tsx
```diff
@@ -0,0 +1,45 @@
...
+// SPDX-License-Identifier: Apache-2.0
+import React, { useState } from 'react';
+import AppLayout from '~components/app-layout';
+import SpaceBetween from '~components/space-between';
+import labels from './utils/labels';
+import BreadcrumbGroup from '~components/breadcrumb-group';
+
+export default function () {
+  const [extraBreadcrumb, setExtraBreadcrumb] = useState(false);
+  return (
```
connorlanigan, line 11: Maybe we could add a `<p>` to inform that this page only works when the `appLayoutWidget` flag is enabled (and maybe its current status?) - it could be confusing otherwise, making people think the feature is broken

src/internal/plugins/controllers/breadcrumbs.ts
```diff
@@ -0,0 +1,87 @@
...
+    this.#appLayoutUpdateCallback(latestBreadcrumb?.props ?? null);
+  }, 0);
+
+  #notifyBreadcrumbs = debounce(() => {
+    if (!this.#appLayoutUpdateCallback) {
+      return;
+    }
+    this.#breadcrumbRegistrations.forEach(listener => listener());
+  }, 0);
+
```
connorlanigan, line 41: We handle multiple breadcrumb instances, do we also need to handle multiple AppLayouts?

just-boris: Handled below. Only one app layout can register as breadcrumbs holder

src/internal/plugins/helpers/use-global-breadcrumbs.ts
just-boris: It causes impact when touches the hosting component:


```
useLayoutEffect(() => {
   setState(something); // this is bad
}, [...])
```

Here we call an API outside React, no concerns here.

src/app-layout/visual-refresh-toolbar/index.tsx
just-boris: Applied

connorlanigan (approved):

PR #2424, opened 2024-06-27
fix: Full-page table uses container background colours instead of page background colours

connorlanigan:
There was a previous PR for this, which was later reverted: #2398 

### Description

This PR updates the colour tokens used for the Table header background. It used to use the Container tokens, even if the table was in the `full-page` variant, which does not have a (visible) container. After this PR, the table uses the page-level colour tokens when it is in the `full-page` variant.

#### Screenshots

Before
themed before

After
themed-after

### How has this been tested?

Manual testing in the dev pages

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/container/internal.tsx
+++ b/src/container/internal.tsx
@@ -143,6 +143,7 @@ export default function InternalContainer({
                   [styles['with-toolbar']]: hasToolbar,
                   [styles['with-hidden-content']]: !children || __hiddenContent,
                   [styles['header-with-media']]: hasMedia,
+                  [styles['header-full-page']]: __fullPage && isRefresh,
                 })}
                 {...stickyStyles}
                 ref={headerMergedRef}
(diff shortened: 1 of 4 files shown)


src/container/internal.tsx
```diff
@@ -143,6 +143,7 @@ export default function InternalContainer({
                   [styles['with-toolbar']]: hasToolbar,
                   [styles['with-hidden-content']]: !children || __hiddenContent,
                   [styles['header-with-media']]: hasMedia,
+                  [styles['header-full-page']]: __fullPage && isRefresh,
```
connorlanigan, line 146: This line is the only change compared to the previous PR. Since the full-page variant only exists in VR, we only apply this class here when VR is enabled.

YueyingLu (approved):

PR #2453, opened 2024-07-10
fix: Removed gap from the normal variant of the button-dropdown

orangevolon:
### Description

There was an unwanted gap in drop down button which was only visible in "normal" variant.
**This PR removes the gap.**

Related links, issue #, if available: AWSUI-48343

### How has this been tested?

Tested and verified using the dev run with Chrome, Firefox and Safari.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/button-dropdown/styles.scss
+++ b/src/button-dropdown/styles.scss
@@ -58,7 +58,8 @@ $dropdown-trigger-icon-offset: 2px;
   display: flex;
 
   & > .trigger-item {
-    & > button:focus {
+    & > .trigger-button:focus,
+    & > .trigger-button:hover {
       z-index: 1;
     }
   }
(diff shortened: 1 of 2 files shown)


avinashbot (changes requested):
Couple of minor visual issues I noticed:

- Disabled dropdown in classic shows its left border on hover. *Feels* like a bug to me, but you could also argue that it's intentional, and I'd be okay with that. If it's disabled, I feel like it shouldn't really have a visual change on focus or hover at all.
Screenshot 2024-07-10 at 17 09 08

- In both themes, when both buttons are active, the focus-visible ring for the main action goes behind the dropdown button
Screenshot 2024-07-10 at 17 12 16

orangevolon:
@avinashbot:

> Disabled dropdown

Like you mentioned this behavior is logically correct but might seem weird from the user's perspective. We certainly can get it done but it comes with the cost of making the styles yet more complicated than it is right now.

One can argue that it's a diminishing return territory, but let's ask for more input and see if this is something that we need to address or not.

> he focus-visible ring

✅ Great catch! Fixed.

src/button-dropdown/styles.scss
```diff
@@ -58,7 +58,8 @@ $dropdown-trigger-icon-offset: 2px;
   display: flex;
 
   & > .trigger-item {
-    & > button:focus {
```
orangevolon, line 61: In some permutations, the trigger is actually not a button but a hyperlink. So it's better to match by the class name rather than the element type.

avinashbot (approved):

PR #2469, opened 2024-07-15
feat: Button Group component

spugachev:
### Description

Button Group component development

Rel: vAZbObs0rwNs

### How has this been tested?

* New unit- and integration tests
* New screenshot tests on permutations

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/button-group/internal.tsx
```diff
@@ -0,0 +1,227 @@
...
+                <ItemElement
+                  item={itemOrGroup}
+                  onItemClick={onItemClick}
+                  ref={element => onSetButtonRef(itemOrGroup, element)}
+                />
+              );
+
+            const isGroupBefore = items[index - 1]?.type === 'group';
+            const isGroupNow = items[index]?.type === 'group';
+            const shouldAddDivider = isGroupBefore || (!isGroupBefore && isGroupNow && index !== 0);
```
pan-kot, line 212: Let's add a dev warning (with warnOnce) for empty groups

src/button-group/internal.tsx
```diff
@@ -0,0 +1,227 @@
...
+        } else {
+          return [item.id];
+        }
+      };
+
+      const getAllIds = (item: ButtonGroupProps.Item) => {
+        if (item.type === 'icon-button') {
+          return [item.id];
+        } else {
+          const values = getAllIdsItemOrGroup(item);
```
pan-kot, line 71: Why is it necessary to collect IDs of the menu items?

src/button-group/item-element.tsx
```diff
@@ -0,0 +1,131 @@
...
+      return () => {
+        window.removeEventListener('pointerdown', handlePointerDownEvent);
+        window.removeEventListener('keydown', handleKeyDownEvent);
+        window.removeEventListener('btngroup-tooltip:toggle', handleTooltipToogleEvent as any);
+      };
+    }, [item.id, showTooltip]);
+
+    useEffect(() => {
+      window.dispatchEvent(
+        new CustomEvent('btngroup-tooltip:toggle', { detail: { open: showTooltip, trackKey: item.id } })
```
pan-kot, line 101: The custom event inside an effect logic is hard to follow and test against. Is it possible to make all checks in the click handler itself?

Instead of having tooltip show and tooltip feedback states here you can lift them to the button group level so that there is only one state for the entire component as there is only one tooltip that can be shown at a time. The state can be like tooltipItemId (can be null which means no tooltip or an item ID which defines for which item to show the tooltip) and tooltipFeedback (a boolean, when true - show feedback instead of text).

src/button-group/menu-dropdown-item.tsx
```diff
@@ -0,0 +1,47 @@
...
+      <ButtonDropdown
+        ref={ref}
+        variant="icon"
+        loading={item.loading}
+        loadingText={item.loadingText}
+        disabled={item.disabled}
+        items={item.items}
+        onItemClick={onClickHandler}
+        expandToViewport={expandToViewport}
+        ariaLabel={item.text}
```
pan-kot, line 39: When doing manual testing I noticed that text is not shown in a tooltip for menu buttons, could you verify please?

src/button-group/__tests__/icon-button-item.test.tsx
```diff
@@ -0,0 +1,32 @@
...
+  text: 'Test Button',
+  iconName: 'add-plus',
+  loading: false,
+  loadingText: 'Loading...',
+  disabled: false,
+  popoverFeedback: 'Action Popover',
+};
+
+describe('IconButtonItem', () => {
+  test('renders the button', () => {
```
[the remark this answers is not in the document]
spugachev: Removed

src/button-group/__tests__/menu-dropdown-item.test.tsx
```diff
@@ -0,0 +1,44 @@
...
+      items: [
+        { id: '5', text: 'Item 5' },
+        { id: '6', text: 'Item 6' },
+      ],
+    },
+  ],
+};
+
+describe('MenuDropdownItem', () => {
+  test('renders the button', () => {
```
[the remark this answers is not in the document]
spugachev: Removed

src/button-group/icon-button-item.tsx
```diff
@@ -0,0 +1,46 @@
...
+    {
+      item,
+      onItemClick,
+    }: {
+      item: ButtonGroupProps.IconButton;
+      onItemClick?: (event: CustomEvent<ClickDetail>) => void;
+    },
+    ref: React.Ref<ButtonProps.Ref>
+  ) => {
+    const hasIcon = item.iconName || item.iconUrl || item.iconSvg;
```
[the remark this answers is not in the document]
spugachev: Added

src/button-group/interfaces.ts
```diff
@@ -0,0 +1,115 @@
...
+import { IconProps } from '../icon/interfaces';
+import { NonCancelableEventHandler } from '../internal/events';
+import { ButtonDropdownProps } from '../button-dropdown/interfaces';
+
+export interface ButtonGroupProps extends BaseComponentProps {
+  /**
+   * Adds `aria-label` to the button group toolbar element.
+   */
+  ariaLabel: string;
+  ariaLabelledby?: string;
```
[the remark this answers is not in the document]
spugachev: ariaLabelledby property is removed

src/button-group/interfaces.ts
```diff
@@ -0,0 +1,115 @@
...
+     * Enabling this property will allow the dropdown to extend beyond that container by using fixed positioning and
+     * [React Portals](https://reactjs.org/docs/portals.html).
+     *
+     * Set this property if the dropdown would otherwise be constrained by a scrollable container,
+     * for example inside table and split view layouts.
+     *
+     * However, use discretion. We recommend you don't enable this property unless necessary
+     * because fixed positioning results in a slight, visible lag when scrolling complex pages.
+     */
+    expandToViewport?: boolean;
```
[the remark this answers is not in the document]
spugachev: Fixed

src/button-group/menu-dropdown-item.tsx
spugachev: Currently it's not shown. When it is shown it looks weird. Let's discuss disign of this

src/button-group/item-element.tsx
spugachev: The puppose of this code is to make synchronize toolpips beetween different ButtonGroup components on the same page. I agree that it's better to lift the code. The only question - should we keep global event taht closes other ButtonGroup's tooltips?

pan-kot: I don't see how the global event is useful. The tooltip of a particular button group instance must close when there is a click outside

spugachev: https://github.com/user-attachments/assets/d2d0f5c8-286c-471a-a73e-000828820f0f

Here is a demo

src/button-group/icon-button-item.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+    if (isDevelopment && !hasIcon) {
```
pan-kot, line 27: I believe isDevelopment is already checked by warnOnce

pan-kot: We should also add a unit test for the warning

src/button-group/interfaces.ts
```diff
@@ -0,0 +1,114 @@
...
+
+export interface InternalButtonGroupProps extends ButtonGroupProps, InternalBaseComponentProps {}
+
+export namespace ButtonGroupProps {
+  // Fixes documenter output.
+  // When a union only has one element the resulting type for it is not string but object.
+  export type Variant = 'icon' | 'icon';
+
+  export type ItemOrGroup = Item | Group;
+  export type Item = IconButton | MenuDropdown;
```
pan-kot, line 75: Let's make the types Item and ItemOrGroup private (move outside the namespace)

src/button-group/item-element.tsx
pan-kot: I think this one is probably fine - this is consistent with how our popover works so if we decide to fix that it is better to invent a holistic solution.

src/button-group/index.tsx
```diff
@@ -0,0 +1,21 @@
...
+import useBaseComponent from '../internal/hooks/use-base-component';
+import { ButtonGroupProps } from './interfaces';
+import InternalButtonGroup from './internal';
+import { getBaseProps } from '../internal/base-component';
+
+export { ButtonGroupProps };
+
+const ButtonGroup = React.forwardRef((props: ButtonGroupProps, ref: React.Ref<ButtonGroupProps.Ref>) => {
+  const baseProps = getBaseProps(props);
+  const baseComponentProps = useBaseComponent('ButtonGroup');
```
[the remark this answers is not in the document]
pan-kot: Yes!

src/button-group/internal.tsx
spugachev: Added

src/button-group/internal.tsx
spugachev: Thes code was removed

src/button-group/interfaces.ts
spugachev: Why do you want to do it? It ssems that users should be able to create items of these types

pan-kot: Ah, I forgot the type is used in the interface above. Nevermind - let's keep both Item and ItemOrGroup

src/button-group/internal.tsx
```diff
@@ -0,0 +1,297 @@
...
+    const baseProps = getBaseProps(props);
+    const [focusedItemId, setFocusedItemId] = useState<null | string>(null);
+    const navigationAPI = useRef<SingleTabStopNavigationAPI>(null);
+    const containerObjectRef = useRef<HTMLDivElement>(null);
+    const containerRef = useMergeRefs(containerObjectRef, __internalRootRef);
+    const itemsRef = useRef<Record<string, ButtonProps.Ref | null>>({});
+    const itemWrappersRef = useRef<Record<string, HTMLDivElement | null>>({});
+    const [tooltipItemId, setTooltipItemId] = useState<string | null>(null);
+    const [isFeedbackTooltip, setIsFeedbackTooltip] = useState(false);
+    const allItems = useMemo(
```
pan-kot, line 48: nit: I'd call the variable itemsById or similar to emphasise it is a dictionary type

src/button-group/internal.tsx
```diff
@@ -0,0 +1,297 @@
...
+              <React.Fragment key={itemOrGroup.type === 'group' ? index : itemOrGroup.id}>
+                {shouldAddDivider && <div className={styles.divider} />}
+                {itemOrGroup.type === 'group' ? (
+                  <div key={index} role="group" aria-label={itemOrGroup.text} className={styles.group}>
+                    {itemOrGroup.items.map(item => (
+                      <React.Fragment key={item.id}>{itemContent(item)}</React.Fragment>
+                    ))}
+                  </div>
+                ) : (
+                  <React.Fragment key={itemOrGroup.id}>{itemContent(itemOrGroup)}</React.Fragment>
```
pan-kot, line 271: This fragment is probably not needed here

src/button-group/item-element.tsx
```diff
@@ -0,0 +1,38 @@
...
+      onItemClick,
+    }: {
+      item: ButtonGroupProps.Item;
+      dropdownExpandToViewport?: boolean;
+      onItemClick?: CancelableEventHandler<ButtonGroupProps.ItemClickDetails | ClickDetail>;
+    },
+    ref: React.Ref<HTMLDivElement>
+  ) => {
+    return (
+      <>
```
pan-kot, line 23: nit: you can remove the fragment in favour of having a switch or if/else in the component's function body

src/button-group/internal.tsx
spugachev: Renamed

src/button-group/internal.tsx
spugachev: Removed

src/button-group/item-element.tsx
spugachev: Using switch now

src/button-group/internal.tsx
```diff
@@ -0,0 +1,191 @@
...
+        onHome: () => focusElement(focusables[0]),
+        onEnd: () => focusElement(focusables[focusables.length - 1]),
+        onInlineStart: () => focusElement(focusables[circleIndex(activeIndex - 1, [0, focusables.length - 1])]),
+        onInlineEnd: () => focusElement(focusables[circleIndex(activeIndex + 1, [0, focusables.length - 1])]),
+      });
+    }
+
+    function focusElement(element: HTMLElement) {
+      setFeedbackItemId(null);
+      setTimeout(() => {
```
pan-kot, line 111: what it the purpose of this timeout?

src/button-group/item-element.tsx
```diff
@@ -0,0 +1,142 @@
...
+      };
+    }, [item.id, setFeedbackItemId, showTooltip]);
+
+    const onShowTooltip = (show: boolean) => {
+      if (feedbackItemId && (feedbackItemId !== item.id || (feedbackItemId && showTooltip))) {
+        return;
+      }
+
+      if (show) {
+        buttonRef.current?.focus();
```
pan-kot, line 88: This line steals the focus from the dropdown when it is open. The user moves focus to the menu button and presses Enter. The expectation is that the first item of the menu dropdown is focused but instead the focus comes back to the menu item.

pan-kot: @spugachev  I created a PR with a fix: https://github.com/cloudscape-design/components/pull/2491

src/button-group/icon-button-item.tsx
spugachev: How do you test warnOnce?

pan-kot: We mock it and assert calls, see example: https://github.com/cloudscape-design/components/blob/2c3bfbe12bfe8773f12896a09664f911d468bd85/src/expandable-section/__tests__/expandable-section.test.tsx#L295

pan-kot:
@spugachev I contributed a fix for tooltip conditions: https://github.com/cloudscape-design/components/pull/2496

src/button-group/interfaces.ts
```diff
@@ -0,0 +1,121 @@
...
+    iconUrl?: string;
+    iconSvg?: React.ReactNode;
+    popoverFeedback?: React.ReactNode;
+  }
+
+  export interface MenuDropdown {
+    type: 'menu-dropdown';
+    id: string;
+    text: string;
+    disabled?: boolean;
```
johannes-weber, line 99: We recently introduced the disabled with reason feature (https://github.com/cloudscape-design/components/pull/2409). I can imagine we'll get some why they are not available on the ButtonGroup component.

Is there a specific reason we do not add the disabled reason here as well?

src/button-group/interfaces.ts
```diff
@@ -0,0 +1,121 @@
...
+import { IconProps } from '../icon/interfaces';
+import { BaseComponentProps } from '../internal/base-component';
+import { NonCancelableEventHandler } from '../internal/events';
+import { InternalBaseComponentProps } from '../internal/hooks/use-base-component';
+
+export interface ButtonGroupProps extends BaseComponentProps {
+  /**
+   * Adds `aria-label` to the button group toolbar element.
+   */
+  ariaLabel: string;
```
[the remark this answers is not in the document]
pan-kot: Right, good catch 👍 

@spugachev , could you please update? Reference: 9Tj7AP2BUVuJ

src/button-group/interfaces.ts
pan-kot: Yes, the disabled reasons are not supported for icon buttons as of yet.

pan-kot:
Submitted a fix for the failing test and more tests: https://github.com/cloudscape-design/components/pull/2516

src/button-group/interfaces.ts
```diff
@@ -0,0 +1,121 @@
...
+  /**
+   * Called when the user clicks on an item, and the item is not disabled. The event detail object contains the id of the clicked item.
+   */
+  onItemClick?: NonCancelableEventHandler<ButtonGroupProps.ItemClickDetails>;
+}
+
+export interface InternalButtonGroupProps extends ButtonGroupProps, InternalBaseComponentProps {}
+
+export namespace ButtonGroupProps {
+  // Fixes documenter output.
```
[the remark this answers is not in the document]
pan-kot: Good question. Here is the fix: https://github.com/cloudscape-design/documenter/pull/42

src/button-group/interfaces.ts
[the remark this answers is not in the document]
pan-kot: Update: I fixed the property type, resolving

src/button-group/__tests__/button-group.test.tsx
```diff
@@ -0,0 +1,253 @@
...
+    expect(wrapper.findMenuById('misc')!.getElement().getElementsByTagName('button')[0]).toHaveFocus();
+  });
+
+  test('focuses the correct item with keyboard', async () => {
+    const ref: { current: ButtonGroupProps.Ref | null } = { current: null };
+    const wrapper = renderButtonGroup({ items: items1 }, ref);
+    ref.current?.focus('copy');
+
+    fireEvent.keyDown(wrapper.getElement(), { keyCode: KeyCode.right });
+    await timeout();
```
johannes-weber, line 123: Why do we need the timeouts? I assume it's because of waiting for the react updates. If so, why not using `act` instead?

src/button-group/icon-button-item.tsx
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    return (
+      <div ref={containerRef}>
+        <InternalButton
+          variant="icon"
+          loading={item.loading}
+          loadingText={item.loadingText}
+          disabled={item.disabled}
+          iconName={hasIcon ? item.iconName : 'close'}
```
johannes-weber, line 47: What's the motivation behind using the `close` icon as default variant? I'd like to better understand why the `iconName` property is optional and not required.

src/button-group/interfaces.ts
```diff
@@ -0,0 +1,121 @@
...
+
+import { ButtonDropdownProps } from '../button-dropdown/interfaces';
+import { IconProps } from '../icon/interfaces';
+import { BaseComponentProps } from '../internal/base-component';
+import { NonCancelableEventHandler } from '../internal/events';
+import { InternalBaseComponentProps } from '../internal/hooks/use-base-component';
+
+export interface ButtonGroupProps extends BaseComponentProps {
+  /**
+   * Adds `aria-label` to the button group toolbar element.
```
johannes-weber, line 12: Let's add a sentence about when to use this optional property as we have it in other components. Suggestion:
Use to assign unique labels when there are multiple button groups on one page.

src/button-group/item-element.tsx
```diff
@@ -0,0 +1,128 @@
...
+    const onShowTooltipHard = (show: boolean) => {
+      if (!show && item.id !== tooltip?.item) {
+        return;
+      }
+
+      setTooltip(show ? { item: item.id, feedback: false } : null);
+    };
+
+    const onClickHandler = (event: CustomEvent<ButtonGroupProps.ItemClickDetails | ClickDetail>) => {
+      const popoverFeedback = 'popoverFeedback' in item && item.popoverFeedback;
```
johannes-weber, line 86: Minor naming suggestion:
```suggestion
      const hasPopoverFeedback = 'popoverFeedback' in item && item.popoverFeedback;
```

src/button-group/item-element.tsx
```diff
@@ -0,0 +1,128 @@
...
+      tooltip,
+      setTooltip,
+      onItemClick,
+    }: {
+      item: ButtonGroupProps.Item;
+      dropdownExpandToViewport?: boolean;
+      tooltip: null | { item: string; feedback: boolean };
+      setTooltip: (tooltip: null | { item: string; feedback: boolean }) => void;
+      onItemClick?: NonCancelableEventHandler<ButtonGroupProps.ItemClickDetails> | undefined;
+    },
```
johannes-weber, line 27: Suggestion (readability): You could extract the prop definitions in a separtate `ItemElementProps` interface.

src/button-group/menu-dropdown-item.tsx
```diff
@@ -0,0 +1,77 @@
...
+      item,
+      showTooltip,
+      onItemClick,
+      expandToViewport,
+    }: {
+      item: ButtonGroupProps.MenuDropdown;
+      showTooltip: boolean;
+      onItemClick?: CancelableEventHandler<ButtonGroupProps.ItemClickDetails>;
+      expandToViewport?: boolean;
+    },
```
johannes-weber, line 28: Suggestion (readability): You could extract the prop definitions in a separtate `MenuDropdownItemProps` interface.

src/button-group/menu-dropdown-item.tsx
```diff
@@ -0,0 +1,77 @@
...
+                className={clsx(styles.tooltip, testUtilStyles['button-group-tooltip'])}
+              />
+            )}
+            <InternalButton
+              ref={triggerRef}
+              variant="icon"
+              ariaLabel={ariaLabel}
+              ariaExpanded={ariaExpanded}
+              className={clsx(styles.item, testUtilsClass)}
+              data-testid={item.id}
```
johannes-weber, line 65: We are setting the same `data-test-id` twice (in this line and above, in line 48) - can we avoid setting the same ID twice?

src/internal/hooks/use-focus-visible-state/index.ts
```diff
@@ -0,0 +1,31 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import { useEffect, useState } from 'react';
+
+import { isModifierKey } from '@cloudscape-design/component-toolkit/internal';
+
+export function useFocusVisibleState() {
```
johannes-weber, line 7: I think we should add unit tests for this hook as it's now a standalone hook outside of the internal chart-plot component.

src/internal/context/single-tab-stop-navigation-context.tsx
```diff
@@ -97,7 +97,7 @@ export const SingleTabStopNavigationProvider = forwardRef(
       const isUnregisteringFocusedNode = nodeBelongs(focusableElement, document.activeElement);
       if (isUnregisteringFocusedNode) {
         // Wait for unmounted node to get removed from the DOM.
-        requestAnimationFrame(() => onUnregisterActive?.(focusableElement));
+        setTimeout(() => onUnregisterActive?.(focusableElement), 0);
```
johannes-weber, line 100: Why did we change from `requestAnimationFrame` to `setTimeout`?

johannes-weber (commented):
Thanks for the contribution, @spugachev! I've included some thoughts and suggestions.

src/button-group/__tests__/button-group.test.tsx
```diff
@@ -0,0 +1,263 @@
...
+const renderButtonGroup = (props: Partial<ButtonGroupProps>, ref?: React.Ref<ButtonGroupProps.Ref>) => {
+  const renderResult = render(<ButtonGroup ref={ref} {...defaultProps} {...props} />);
+  return createWrapper(renderResult.container).findButtonGroup()!;
+};
+
+const timeout = async (duration = 0) => {
+  await new Promise(resolve => setTimeout(resolve, duration));
+};
+
+const items1: ButtonGroupProps.ItemOrGroup[] = [
```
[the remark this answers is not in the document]
spugachev: Changed it to "items"

src/button-group/__tests__/button-group.test.tsx
```diff
@@ -0,0 +1,253 @@
...
+    items: [],
+  },
+];
+
+test('renders all items', () => {
+  const wrapper = renderButtonGroup({ items: items1 });
+
+  expect(wrapper.findItems()).toHaveLength(6);
+  expect(wrapper.findButtonById('edit')).not.toBe(null);
+  expect(wrapper.findMenuById('misc')).not.toBe(null);
```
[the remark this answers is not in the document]
spugachev: Yes, it would be enough. Removed checks

src/internal/hooks/use-focus-visible-state/index.ts
spugachev: We are not using this anymore, removed the hook

src/button-group/styles.scss
```diff
@@ -0,0 +1,37 @@
...
+  display: inline-block;
+}
+
+.item {
+  /* used in test-utils */
+}
+
+.tooltip {
+  /* used in test-utils */
+}
```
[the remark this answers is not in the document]
spugachev: Moved

src/internal/context/single-tab-stop-navigation-context.tsx
pan-kot: That is due to Wait for unmounted node to get removed from the DOM comment. The requestAnimationFrame fires before the element is removed and the focus is casted on the element which is then removed.

The code is used in button group, table, and tabs. Using the set timeout does not affect the other component but using the animation frame in button group will fail an integration test designed to capture this behavior.

src/button/internal.tsx
```diff
@@ -45,6 +45,7 @@ export type InternalButtonProps = Omit<ButtonProps, 'variant'> & {
   __iconClass?: string;
   __focusable?: boolean;
   __injectAnalyticsComponentMetadata?: boolean;
+  __title?: string;
```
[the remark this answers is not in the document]
spugachev: We already have a tooltip. Adding a title creates a second tooltip. We want this title to always be empty.
Screenshot 2024-07-16 at 7 50 25 PM

pages/button-group/item-permutations.page.tsx
```diff
@@ -0,0 +1,100 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import React from 'react';
+
+import { ButtonGroup, ButtonGroupProps, SpaceBetween, StatusIndicator } from '~components';
```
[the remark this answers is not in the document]
spugachev: Done

pages/button-group/permutations.page.tsx
```diff
@@ -0,0 +1,106 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import React from 'react';
+
+import { ButtonGroup, ButtonGroupProps } from '~components';
```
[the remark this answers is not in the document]
spugachev: Done

src/internal/context/single-tab-stop-navigation-context.tsx
johannes-weber: Got it, thanks for clarification!

src/button-group/__tests__/button-group.test.tsx
spugachev: Don't need it anymore. Removed

src/button-group/icon-button-item.tsx
spugachev: Good question. It can be iconName, iconAlt or iconSvg. We can't make iconName requiried. 
@pan-kot waht do you think, should we display nothing if no icon defined?

src/button-group/interfaces.ts
spugachev: I cahnged it to:
  /**
   * Adds `aria-label` to the button group toolbar element.
   * Use this to provide a unique accessible name for buttons groups on the page.
   */

src/button-group/menu-dropdown-item.tsx
pan-kot: Do you see a risk with that?

These test-IDs are always used in combination with class names. We need the one on the button-dropdown so that the test-utils elector can point to the component's root so that we can wrap it with ButtonDropdownWrapper.

The second one is used in the internal.tsx to find the focus targets that are the actual focusable elements so we need it to point to the trigger. We can technically find the button dropdown and use a nested selector to find a trigger but that would increase those queries complexity by much as we'd need to assert item types to figure out the selector to use.

src/button-group/menu-dropdown-item.tsx
spugachev: Extracteed MenuDropdownItemProps to interfaces.ts

src/button-group/item-element.tsx
spugachev: Extracted ItemElementProps to interfaces.ts

src/button-group/menu-dropdown-item.tsx
spugachev: @pan-kot I think we can keep only one. Am I right?

src/button-group/__tests__/button-group.test.tsx
```diff
@@ -0,0 +1,253 @@
...
+  const onItemClick = jest.fn();
+  const wrapper = renderButtonGroup({ items: items1, onItemClick });
+  const buttonDropdown = wrapper.findMenuById('misc')!;
+  buttonDropdown.openDropdown();
+  buttonDropdown.findItemById('menu-open')!.click();
+
+  expect(onItemClick).toHaveBeenCalledWith(expect.objectContaining({ detail: { id: 'menu-open' } }));
+});
+
+describe('focus', () => {
```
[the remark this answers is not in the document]
spugachev: Added a test that focus returned to the dropdown button

src/button-group/icon-button-item.tsx
johannes-weber: What about a `warnOnce` if none of them is provided?

spugachev: Yes, we already have warnOnce

src/button-group/menu-dropdown-item.tsx
pan-kot: Addressed here: https://github.com/cloudscape-design/components/pull/2469/commits/4a40ee45296b5b217aeb03a14abe0156f51feaab

johannes-weber (approved):

pan-kot (dismissed):
Great work @spugachev !

pan-kot (approved):

pan-kot (approved):

PR #2475, opened 2024-07-17
feat: add loading a11y to s3 selectors for screen readers

orangevolon:
### Description

Adds `Last Updated` to the S3 resource selector modal.
The addition of this text should help the screen readers to inform the users once the fetch process is finished.

Related links, issue #, if available: AWSUI-41521

### How has this been tested?

Manually tested on Chrome, Firefox & Safari.
Automated testing _in progress_.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/s3-resource-selector/interfaces.ts
+++ b/src/s3-resource-selector/interfaces.ts
@@ -218,6 +218,7 @@ export namespace S3ResourceSelectorProps {
     modalCancelButton?: string;
     modalSubmitButton?: string;
     modalBreadcrumbRootItem?: string;
+    modalLastUpdatedText?: string;
 
     selectionBuckets?: string;
     selectionObjects?: string;
--- a/src/s3-resource-selector/s3-modal/basic-table.tsx
+++ b/src/s3-resource-selector/s3-modal/basic-table.tsx
@@ -8,15 +8,20 @@ import { useStableCallback } from '@cloudscape-design/component-toolkit/internal
 import { InternalButton } from '../../button/internal';
 import InternalHeader from '../../header/internal';
 import { ComponentFormatFunction } from '../../i18n/context';
+import LiveRegion from '../../internal/components/live-region';
 import useForwardFocus, { ForwardFocusRef } from '../../internal/hooks/forward-focus';
+import formatDateLocalized from '../../internal/utils/date-time/format-date-localized';
 import { PaginationProps } from '../../pagination/interfaces';
 import InternalPagination from '../../pagination/internal';
+import InternalSpaceBetween from '../../space-between/internal';
 import { TableProps } from '../../table/interfaces';
 import InternalTable from '../../table/internal';
 import { TextFilterProps } from '../../text-filter/interfaces';
 import InternalTextFilter from '../../text-filter/internal';
 import { S3ResourceSelectorProps } from '../interfaces';
 import { EmptyState } from './empty-state';
+
+import styles from './styles.css.js';
 
 interface BasicS3TableStrings<T> {
   labelRefresh?: string;
@@ -27,6 +32,7 @@ interface BasicS3TableStrings<T> {
   filteringAriaLabel?: string;
   filteringClearAriaLabel?: string;
   filteringCounterText?: S3ResourceSelectorProps.I18nStrings['filteringCounterText'];
+  lastUpdatedText?: string;
   emptyText?: string;
   noMatchTitle?: string;
   noMatchSubtitle?: string;
@@ -51,7 +57,7 @@ interface BasicS3TableProps<T> {
 export function getSharedI18Strings(
   i18n: ComponentFormatFunction<'s3-resource-selector'>,
   i18nStrings: S3ResourceSelectorProps.I18nStrings | undefined
-) {
+): BasicS3TableStrings<unknown> {
   return {
     filteringCounterText: i18n(
       'i18nStrings.filteringCounterText',
@@ -64,6 +70,7 @@ export function getSharedI18Strings(
     noMatchSubtitle: i18n('i18nStrings.filteringCantFindMatch', i18nStrings?.filteringCantFindMatch),
     clearFilterButtonText: i18n('i18nStrings.clearFilterButtonText', i18nStrings?.clearFilterButtonText),
     filteringClearAriaLabel: i18nStrings?.labelClearFilter,
+    lastUpdatedText: i18n('i18nStrings.modalLastUpdatedText', i18nStrings?.modalLastUpdatedText),
   };
 }
 
@@ -80,12 +87,13 @@ export function BasicS3Table<T>({
 }: BasicS3TableProps<T>) {
   const [loading, setLoading] = useState(false);
   const [allItems, setAllItems] = useState<ReadonlyArray<T>>([]);
+  const [lastUpdated, setLastUpdated] = useState<Date>();
   const textFilterRef = useRef<TextFilterProps.Ref>(null);
   const onSelectLatest = useStableCallback(onSelect);
 
   function loadData() {
     setLoading(true);
-    fetchData()
+    return fetchData()
       .then(items => {
         setAllItems(items);
         setLoading(false);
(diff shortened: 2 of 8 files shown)


src/s3-resource-selector/interfaces.ts
```diff
@@ -137,6 +137,8 @@ export interface S3ResourceSelectorProps extends BaseComponentProps, BaseModalPr
    */
   fetchBuckets: () => Promise<ReadonlyArray<S3ResourceSelectorProps.Bucket>>;
 
+  fetchStatus?: S3ResourceSelectorProps.FetchStatus;
```
avinashbot, line 140: We should be able to determine this internally using the state of the `fetchBuckets`/`fetchObjects`/`fetchVersions` function, can't we? 'not-initiated' by default, 'pending' when the function is called, 'failed' or 'success' depending on the return value of the promise.

Another note here is that we should only show the "pending" text [after loading is taking a while](https://cloudscape.design/patterns/general/loading-and-refreshing/), so maybe another internal state value needs to be present for when the state is stuck in pending for a while:

> When refreshing can’t be completed within 1-10 seconds, communicate that the process is taking longer by using a status indicator next to the refresh button.

src/s3-resource-selector/s3-modal/basic-table.tsx
```diff
@@ -170,3 +191,53 @@ export function BasicS3Table<T>({
...
+
+      case 'not-initiated':
+      default:
+        return null;
+    }
+  }
+
+  return (
+    <InternalSpaceBetween size="xs" direction="horizontal" alignItems="center">
+      {/* Q: should we make it assertive for warning cases? */}
```
avinashbot, line 238: `"assertive"` is only needed for rare cases, for when we're certain we want to interrupt the user's flow. Polite is fine here :)

I think you also need to pass in `visible={true}` to the live region component for the children to be visible.

src/s3-resource-selector/interfaces.ts
orangevolon: I also had the [same question here](https://github.com/cloudscape-design/components/pull/2475/files#diff-f931890bbb8a1cba1cbfebdf90ffce8ba49ee3caafb7986cab8035f06a20d775R221), but from a different perspective. Let's refine this.

src/s3-resource-selector/s3-modal/basic-table.tsx
```diff
@@ -172,3 +180,39 @@ export function BasicS3Table<T>({
...
+export function InternalHeaderActions<T>({ i18nStrings, loadData, lastUpdated }: InternalHeaderActionsProps<T>) {
+  function getLastUpdated() {
+    if (!lastUpdated || !i18nStrings.lastUpdatedText) {
+      return null;
+    }
+
+    return (
+      <InternalBox textAlign="right" color="text-status-inactive">
+        <LiveRegion visible={true}>
+          <span role="status">
```
orangevolon, line 199: Ideally `<InternalBox />`, `<LiveRegion />` and `<span role="status" />` would be one element but `<LiveRegion />` doesn't receive a `role` prop.

Any suggestion?

jperals (dismissed):

src/s3-resource-selector/s3-modal/basic-table.tsx
```diff
@@ -172,3 +182,38 @@ export function BasicS3Table<T>({
...
+      return null;
+    }
+
+    const formattedDate = formatDateLocalized({
+      date: lastUpdated.toString(),
+      isDateOnly: false,
+    });
+
+    return (
+      <InternalBox className={styles['last-updated-caption']}>
```
avinashbot, line 204: If we're styling using CSS anyway, this can just be a span/div.

src/s3-resource-selector/s3-modal/basic-table.tsx
```diff
@@ -64,6 +71,7 @@ export function getSharedI18Strings(
     noMatchSubtitle: i18n('i18nStrings.filteringCantFindMatch', i18nStrings?.filteringCantFindMatch),
     clearFilterButtonText: i18n('i18nStrings.clearFilterButtonText', i18nStrings?.clearFilterButtonText),
     filteringClearAriaLabel: i18nStrings?.labelClearFilter,
+    lastUpdatedText: i18n('i18nStrings.modalLastUpdatedText', i18nStrings?.modalLastUpdatedText),
```
avinashbot, line 74: Would be good to write a test for this actually - a two-line addition [here](https://github.com/cloudscape-design/components/blob/main/src/s3-resource-selector/__tests__/main.test.tsx#L202) should be good enough.

avinashbot (changes requested):

src/s3-resource-selector/s3-modal/basic-table.tsx
```diff
@@ -96,6 +104,11 @@ export function BasicS3Table<T>({
       });
   }
 
+  async function reloadData() {
+    await loadData();
+    setLastUpdated(new Date());
```
avinashbot, line 109: Turns out this doesn't work, because `loadData` doesn't actually return the promise it creates. You need to add a `return` to the promise in `loadData` so you can `await` it here.

src/s3-resource-selector/s3-modal/basic-table.tsx
```diff
@@ -172,3 +187,38 @@ export function BasicS3Table<T>({
...
+        {formattedDate}
+        <LiveRegion visible={false} source={[i18nStrings.lastUpdatedText, formattedDate]} />
+      </div>
+    );
+  }
+
+  return (
+    <InternalSpaceBetween size="s" direction="horizontal" alignItems="center">
+      {getLastUpdated()}
+      <InternalButton iconName="refresh" ariaLabel={i18nStrings.labelRefresh} onClick={reloadData} />
```
avinashbot, line 221: Just a random comment - it's crazy to me that we don't disable the reload button while a fetch is taking place. But at this point, I'm a little worried that switching it to a loading state is a breaking change, so... 😔

src/s3-resource-selector/s3-modal/basic-table.tsx
```diff
@@ -172,3 +187,38 @@ export function BasicS3Table<T>({
...
+  lastUpdated: Date | undefined;
+}
+
+export function InternalHeaderActions<T>({ i18nStrings, reloadData, lastUpdated }: InternalHeaderActionsProps<T>) {
+  function getLastUpdated() {
+    if (!lastUpdated || !i18nStrings.lastUpdatedText) {
+      return null;
+    }
+
+    const formattedDate = formatDateLocalized({
```
avinashbot, line 203: Just realized this doesn't meet the [right format](https://cloudscape.design/patterns/general/timestamps/) for the timestamp (month first, UTC offset is HH:MM) - I guess the date range picker broke from the format a little, so we'll have to customize the function to fit the timestamp format.

We also include seconds here, but I think we _should_ break from our format here so that the live region reliably announces the timestamp (at least for now until we [improve the live region](https://github.com/cloudscape-design/components/pull/2128))

jperals: The examples in our date formatting guidelines use month-first format because they are assume "en-US" locale. `formatDateLocalized` uses the browser native `Intl.DateTimeFormat` API to generate the formatted dates, which will use month-first format for "en-US" and day-first format for "en" or "en-UK".

If you set the locale to "en-US" to the [date range picker with absolute human-readable format](https://cloudscape.design/components/date-range-picker/?tabId=playground&example=with-absolute-human-readable-format), it will show the month-first format.

jperals: Long story short: the apparent mismatch with the documentation is known and only because of different locale. I think `formatDateLocalized` works correctly and that we should use it here in the same way.

avinashbot: Makes sense, thanks for the context!

src/s3-resource-selector/s3-modal/basic-table.tsx
orangevolon: It certainly looks weird not receiving a feedback when clicking on the button. I think here the other parts of the "loading & refreshing" guideline also comes into play.

Cases for when it's loading, it's taking longer that usual or it's failing for some reason.
I think it makes sense to bucket them together.

avinashbot (approved):

jperals (approved):

PR #2481, opened 2024-07-18
chore: Property filter group tokens

pan-kot:
### Description

Updated filtering token to accommodate grouped tokens. Added a permutation page for tokens.
The grouped tokens to be used by the property filter component once support is added.

Rel: 38LiO4aQfWDx, https://github.com/cloudscape-design/collection-hooks/pull/74

### How has this been tested?

* New unit tests
* Screenshot tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/popover/internal.tsx
```diff
@@ -26,6 +26,7 @@ import { useSingleTabStopNavigation } from '../internal/context/single-tab-stop-
 
 export interface InternalPopoverProps extends PopoverProps, InternalBaseComponentProps {
   __onOpen?: NonCancelableEventHandler<null>;
+  __triggerTypeFlex?: boolean;
```
pan-kot, line 29: This change is needed so that a popover can be added to the new edit button of the grouped filtering token. Without that the popover trigger styles don't allow that custom button to grow and render correctly inside the token.

src/property-filter/styles.scss
```diff
@@ -104,7 +112,6 @@
 }
 
 .remove-all,
-.token-label,
```
pan-kot, line 107: This was not actually used in test utils.

src/popover/internal.tsx
georgylobko: I would suggest to extend `triggerType` internally instead (same way as `variant` for button: https://github.com/cloudscape-design/components/blob/main/src/button/internal.tsx#L40) and introduce a new type `groupedToken`, because `flex` does not say much about type of content, but rather describes style rule.

georgylobko: Moreover, this approach allows you to get rid of the conditions in the code below

src/property-filter/filtering-token/__tests__/filtering-token.test.tsx
```diff
@@ -1,65 +1,176 @@
...
+  token.findRemoveButton().click();
+  expect(onDismissToken).toHaveBeenCalledTimes(1);
+  expect(onDismissToken).toHaveBeenCalledWith(0);
+});
+
+test('fires grouped onDismiss when the remove button is pressed', () => {
+  const onDismissToken = jest.fn();
+  const token = renderToken({ tokens: [token1, token2], onDismissToken });
+  token.findGroupTokens()[1].findRemoveButton().click();
+  expect(onDismissToken).toHaveBeenCalledTimes(1);
```
georgylobko, line 98: [minor] do we need to check how many times this handler has been called? looks like the statement below is enough for this use case

src/property-filter/filtering-token/styles.scss
```diff
@@ -8,13 +8,32 @@
...
+.root,
+.inner-root {
+  @include styles.styles-reset;
+
   display: flex;
   align-content: stretch;
+
+  // Make tokens larger so that individual and group tokens appear of the same size.
+  &.has-groups {
+    min-block-size: calc(2px + #{awsui.$size-vertical-input});
```
georgylobko, line 27: is there any way to introduce a token instead of the 'magic number' for this value instead?

src/property-filter/filtering-token/styles.scss
```diff
@@ -8,13 +8,32 @@
...
+
+    &.compact-mode {
+      min-block-size: calc(2px + 2 * #{styles.$control-border-width} + #{awsui.$size-vertical-input});
```
georgylobko, line 30: same here

pages/property-filter/property-filter-tokens-permutations.page.tsx
```diff
@@ -0,0 +1,159 @@
...
+  ariaLabel: 'filter property3 = value',
+  dismissAriaLabel: 'remove filter property3 = value',
+} as const;
+
+const tokenProps: FilteringTokenProps = {
+  tokens: [],
+  operation: 'and',
+  groupOperation: 'or',
+  showOperation: false,
+  andText: 'und',
```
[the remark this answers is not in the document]
pan-kot: Makes sense! I copied it from the unit test where I decided to use other language so that the label cannot be mistaken with the default value

pages/property-filter/property-filter-tokens-permutations.page.tsx
```diff
@@ -0,0 +1,159 @@
...
+    ],
+  },
+]);
+
+export default function () {
+  return (
+    <>
+      <h1>Property filter tokens permutations</h1>
+      <ScreenshotArea disableAnimations={true}>
+        <PermutationsView
```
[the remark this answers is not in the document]
pan-kot: I would keep it in one page as closely related. This helps ensuring the permutations cover all states and there is no unnecessary duplication.

src/property-filter/filtering-token/__tests__/filtering-token.test.tsx
pan-kot: True, but w/o the counter check the test feels incomplete to me - I always add the counter check as a habit.

Say, we might extract the dismiss handler from a lambda into a function and introduce some logic that would handle a dismiss of a single token differently:

```
function handleDismiss(index) {
  if (tokens.length === 1) {
    onDismiss(index);
    // do something else and forget to return
  }
  onDismiss(index);
}
```

In this case the onDimiss will be called twice with the same index which might come unnoticed.

src/property-filter/filtering-token/styles.scss
pan-kot: I see no value in having a token here at this point because the adjustment does not need to be configurable and there is no token which can account for possible future changes of the spacing that might affect the formula.

In either case, this particular style might get updated after a VR review. At the moment it is not available publicly as the property filter is not yet updated to use the grouped tokens.

@katiegeorge , what do you think?

georgylobko (approved):

PR #2490, opened 2024-07-22
chore: Property filter group editor

pan-kot:
### Description

The PR introduces an alternative property filter token editor that supports grouped tokens as per [38LiO4aQfWDx].

See related PR that introduces group tokens: https://github.com/cloudscape-design/components/pull/2481

### How has this been tested?

* New unit tests
* Screenshot tests

Note: the alternative component is not used in the property filter yet. It will be taken once we add support for grouped token to the property filter.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/property-filter/token-editor.scss
+++ b/pages/property-filter/token-editor.scss
@@ -1,0 +1,17 @@
+/*
+ Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ SPDX-License-Identifier: Apache-2.0
+*/
+
+// Imitate popover container
+.token-editor-container {
+  max-inline-size: 912px;
+
+  @media (max-width: 912px) {
+    max-inline-size: 310px;
+  }
+
+  &.fixed-width {
+    inline-size: 912px;
+  }
+}
--- a/src/property-filter/__tests__/property-filter-token-editor-grouped.test.tsx
+++ b/src/property-filter/__tests__/property-filter-token-editor-grouped.test.tsx
@@ -306,4 +306,19 @@ describe.each(['xs', 'default'] as const)('breakpoints = %s', breakpoint => {
     expect(onChangeStandalone).toHaveBeenCalledTimes(1);
     expect(onChangeStandalone).toHaveBeenCalledWith([]);
   });
+
+  test('add standalone property menu items have indices as IDs', () => {
+    const wrapper = renderComponent({
+      tempGroup: [],
+      standaloneTokens: [
+        { property: nameProperty, operator: '=', value: '1' },
+        { property: nameProperty, operator: '=', value: '2' },
+      ],
+    });
+    const tokenActions = wrapper.findTokenAddActions()!;
+
+    tokenActions.openDropdown();
+    expect(tokenActions.findItemById('0')!.getElement().textContent).toBe('Add filter Name = 1 to group');
+    expect(tokenActions.findItemById('1')!.getElement().textContent).toBe('Add filter Name = 2 to group');
+  });
 });
--- a/src/property-filter/__tests__/property-filter-token-editor-grouped.test.tsx
+++ b/src/property-filter/__tests__/property-filter-token-editor-grouped.test.tsx
@@ -91,193 +91,210 @@ function renderComponent(props?: Partial<TokenEditorProps>) {
   return new InternalPropertyFilterEditorDropdownWrapper(container);
 }
 
-test('renders token editor with a single property', () => {
-  const wrapper = renderComponent({});
+describe.each([500, 1000])('window size = %spx', width => {
+  function findRemoveAction(wrapper: InternalPropertyFilterEditorDropdownWrapper, index: number) {
+    wrapper.findTokenRemoveActions(index).openDropdown();
+    return width === 500
+      ? wrapper.findTokenRemoveActions(index).findMainAction()!
+      : wrapper.findTokenRemoveActions(index).findItems()[0];
+  }
+  function findRemoveFromGroupAction(wrapper: InternalPropertyFilterEditorDropdownWrapper, index: number) {
+    wrapper.findTokenRemoveActions(index).openDropdown();
+    return wrapper.findTokenRemoveActions(index).findItems()[width === 500 ? 0 : 1];
+  }
 
-  // Not supported yet because the standalone editor does not include the popover
-  // This can be uncommented when defining the test against the property filter
-  // which is not possible for the internal-only token-editor-grouped yet.
-  // expect(wrapper.findHeader().getElement()).toHaveTextContent('Edit token');
-  // expect(wrapper.findDismissButton().getElement()).toHaveAccessibleName('Close token editor');
-
-  // Token inputs
-  expect(wrapper.findPropertyField().findControl()!.find('button')!.getElement()).toHaveAccessibleName('Property Name');
-  expect(wrapper.findOperatorField().findControl()!.find('button')!.getElement()).toHaveAccessibleName('Operator =');
-  expect(wrapper.findValueField().findControl()!.find('input')!.getElement()).toHaveAccessibleName('Value');
-  expect(wrapper.findValueField().findControl()!.find('input')!.getElement()).toHaveValue('John');
-
-  // Remove actions
-  expect(wrapper.findTokenRemoveActions().findNativeButton().getElement()).toHaveAccessibleName(
-    'Filter remove actions for Name = John'
-  );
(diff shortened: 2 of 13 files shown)


src/property-filter/token-editor-grouped.tsx
```diff
@@ -0,0 +1,421 @@
...
+    render: (index: number) => React.ReactNode;
+  };
+  value: {
+    label: string;
+    render: (index: number) => React.ReactNode;
+  };
+}
+
+function TokenEditorFields({ fields, groupLabel, removeButton, property, operator, value }: TokenEditorLayout) {
+  const breakpoint = 912;
```
katiegeorge, line 245: Where is 912 coming from? Should we export this as a constant somewhere?

src/property-filter/styles.scss
```diff
@@ -83,6 +83,111 @@
...
+      gap: awsui.$space-xs;
+      padding-block-end: awsui.$space-xs;
+      border-block-end: 1px solid awsui.$color-border-divider-default;
+    }
+  }
+
+  &-table {
+    @include styles.styles-reset;
+
+    display: table;
```
katiegeorge, line 147: This is interesting. Why not use grid here?

katiegeorge: With grid, you can add minmax to the column definition, which will let you more easily constrain the widths of the content (and therefore, the width of the popover)

src/property-filter/styles.scss
```diff
@@ -83,6 +83,111 @@
...
+      padding-block: awsui.$space-scaled-xxs;
+      text-align: start;
+    }
+
+    &-cell {
+      padding-block-end: awsui.$space-s;
+    }
+
+    @media (max-width: 912px) {
+      //
```
katiegeorge, line 173: Should this be here?

src/property-filter/styles.scss
```diff
@@ -83,6 +83,111 @@
...
+    }
+  }
+
+  &-operation {
+    align-self: flex-start;
+  }
+
+  &-add-token {
+    // margin-block-start: awsui.$space-xxs;
```
katiegeorge, line 182: This too

src/property-filter/token-editor-grouped.tsx
pan-kot: yes, we have a utility we can use - will update

src/property-filter/styles.scss
pan-kot: Updated

pages/property-filter/token-editor.scss
```diff
@@ -0,0 +1,17 @@
+/*
+ Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ SPDX-License-Identifier: Apache-2.0
+*/
+
+// Imitate popover container
+.token-editor-container {
+  max-inline-size: 912px;
```
katiegeorge, line 8: Can we make this a variable? This feels like a very random value without any context.

pan-kot: This is defined for test pages only, I'd keep it like that rather, at least until the popover size for the new token editor is figured out.

src/property-filter/__tests__/property-filter-token-editor-grouped.test.tsx
```diff
@@ -0,0 +1,309 @@
...
+  function findRemoveFromGroupAction(wrapper: InternalPropertyFilterEditorDropdownWrapper, index: number) {
+    wrapper.findTokenRemoveActions(index)!.openDropdown();
+    return wrapper.findTokenRemoveActions(index)!.findItems()[isNarrow ? 0 : 1];
+  }
+
+  beforeEach(() => {
+    jest.mocked(useContainerBreakpoints).mockReturnValue([breakpoint, () => {}]);
+  });
+
+  test.each([false, true])('renders token editor with a single property, supportsGroups=%s', supportsGroups => {
```
georgylobko, line 116: does it make sense to run such tests in the loop with the condition below? I would suggest to split them up in this case

src/property-filter/styles.scss
```diff
@@ -83,6 +85,91 @@
...
 
+.token-editor-grouped {
+  display: flex;
+  flex-direction: column;
+  gap: awsui.$space-s;
+  user-select: text;
+  margin-block: awsui.$space-xxs;
+  margin-inline: awsui.$space-xxs;
+
+  &-form {
```
georgylobko, line 96: I'm not sure if we have any agreements on that, but from my perspective writing css classes this way makes it difficult to find and debug them in future.

WDYT?

src/property-filter/styles.scss
```diff
@@ -83,6 +85,91 @@
...
+  &-field-value {
+    flex-grow: 2;
+  }
+
+  &-cancel {
+    margin-inline-end: awsui.$space-xs;
+  }
+
+  &-submit {
+    /* used in test-utils */
```
georgylobko, line 117: I think this class should be moved to `test-classes` folder

src/property-filter/token-editor-grouped.tsx
```diff
@@ -0,0 +1,407 @@
...
+        tokens={groups.map(group => getFormattedToken(group.token, i18nStrings))}
+        onRemove={index => {
+          const copy = [...tempGroup];
+          copy.splice(index, 1);
+          onChangeTempGroup(copy);
+        }}
+        onRemoveFromGroup={index => {
+          const removedToken = tempGroup[index];
+          const copy = [...tempGroup];
+          copy.splice(index, 1);
```
georgylobko, line 124: `.filter`?

src/property-filter/token-editor-grouped.tsx
```diff
@@ -0,0 +1,407 @@
...
+        >
+          <InternalButtonDropdown
+            variant="normal"
+            ariaLabel={i18nStrings.tokenEditorAddTokenActionsLabel}
+            items={standaloneTokens.map((token, index) => ({
+              id: index.toString(),
+              text: i18nStrings.tokenEditorAddExistingTokenLabel(getFormattedToken(token, i18nStrings)),
+            }))}
+            onItemClick={({ detail }) => {
+              const index = parseInt(detail.id);
```
georgylobko, line 177: if `detail.id` contains anything but not stringified number, `index` will be declared as `NaN`. I would suggest to handle this case and write unit tests for that

src/property-filter/token-editor-grouped.tsx
```diff
@@ -0,0 +1,407 @@
...
+              const addedToken = standaloneTokens[index];
+              const copy = [...standaloneTokens];
+              copy.splice(index, 1);
```
georgylobko, line 180: wouldn't it be better to use immutable methods for that? `.filter` is a good fit

src/property-filter/__tests__/property-filter-token-editor-grouped.test.tsx
pan-kot: I think it does. There are certain fields that are supported when the groups are enabled and when they are not. Enabling groups unblocks more elements and those are tested separately with dedicated test cases.

src/property-filter/styles.scss
pan-kot: I am fine with either keeping that style or banning it, but there are currently 300+ usages of this style in the code. I don't see it as a big concern as hovering over that line in most editors would show the complete class name.

src/property-filter/styles.scss
pan-kot: The current plan is to replace the existing token editor with the grouped one. In that case we would remove the token-editor-grouped styles and update the existing ones instead and so that the existing &-submit would stay unchanged, lowering the risk of test classes mismatch.

src/property-filter/token-editor-grouped.tsx
pan-kot: I personally find the .splice approach better explicit but don't mind either. Updated to .filter in both handlers

src/property-filter/token-editor-grouped.tsx
pan-kot: That is an interesting one. I will add a check to avoid NaN and index overflow errors as executing no action is likely better than throwing an exception in this particular case. However, the index being NaN or overflowing is an invariant as it is controlled by us just a few lines above - I can't even define a test to verify the onItemClick logic in this case but I will add one to ensure the assigned IDs are indices.

georgylobko (approved):

PR #2492, opened 2024-07-22
feat: Toggle button component

georgylobko:
### Description

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/toggle-button/internal.tsx
```diff
@@ -0,0 +1,53 @@
...
+      onClick,
+      ...rest
+    }: ToggleButtonProps,
+    ref: React.Ref<ToggleButtonProps.Ref>
+  ) => {
+    return (
+      <InternalButton
+        className={clsx(className, styles[`variant-${variant}`], { [styles.pressed]: pressed })}
+        variant={variant}
+        iconName={getToggleIcon(pressed, defaultIconName, pressedIconName)}
```
georgylobko, line 35: Introduced `getToggleIcon` to make it more testable

style-dictionary/classic/metadata.ts
```diff
@@ -13,6 +13,7 @@ const updatedDescriptions: StyleDictionary.TokenIndex<string> = {
     'The color of clickable elements in their default state. *For example: expandable sections, tabs, and icons.*',
   colorTextInteractiveHover:
     'The color of clickable elements on hover. *For example: expandable sections, and icons on hover.*',
+  colorTextInteractivePressed: 'The color of clickable elements in pressed state.',
```
jperals, line 16: Adding this here will make this a public token. Just to make sure, is this intentional?

src/__tests__/outer-form-submit.test.tsx
```diff
@@ -6,7 +6,7 @@ import { render } from '@testing-library/react';
 import { getRequiredPropsForComponent } from './required-props-for-components';
 import { getAllComponents, requireComponent } from './utils';
 
-const skippedComponents = ['button'];
+const skippedComponents = ['button', 'toggle-button'];
```
jperals, line 9: Why is this change necessary?

This test was introduced to make sure that there was no element with `type="submit"` which would accidentally cause form submission. The button component can render that attribute in the native button if using `formAction="submit"`, but the new ToggleButton component doesn't use `formAction="submit"` in the InternalButton props so it should not have that issue. So is there any reason to not cover this new component here along with the rest of components?

pages/toggle-button/permutations.page.tsx
```diff
@@ -0,0 +1,40 @@
...
+// SPDX-License-Identifier: Apache-2.0
+import React from 'react';
+
+import ToggleButton, { ToggleButtonProps } from '~components/toggle-button';
+
+import createPermutations from '../utils/permutations';
+import PermutationsView from '../utils/permutations-view';
+import ScreenshotArea from '../utils/screenshot-area';
+
+const permutations = createPermutations<ToggleButtonProps>([
```
jperals, line 11: Could we add coverage for using SVG and even URL for the icon?

src/toggle-button/internal.tsx
```diff
@@ -0,0 +1,53 @@
...
+        iconUrl={getToggleIcon(pressed, defaultIconUrl, pressedIconUrl)}
+        iconSvg={getToggleIcon(pressed, defaultIconSvg, pressedIconSvg)}
+        aria-pressed={pressed}
+        onClick={event => {
+          if (onClick) {
```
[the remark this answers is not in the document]
jperals: But they can use `onChange` all the same, right?

src/button/interfaces.ts
```diff
@@ -124,11 +91,6 @@ export interface ButtonProps extends BaseComponentProps {
    */
   variant?: ButtonProps.Variant;
```
jperals, line 92: Nit: this could be in the Button interfaces instead of here as ToggleButton overrides it with different values

src/button/interfaces.ts
```diff
@@ -146,11 +108,6 @@ export interface ButtonProps extends BaseComponentProps {
    */
   onFollow?: CancelableEventHandler<ButtonProps.FollowDetail>;
```
jperals, line 109: Do we want to inherit `onFollow`? To my understanding the reason to have it in the Button component is for cases where using `href` and therefore clicking the button navigates to a different route, which is not a use case for the Toggle Button

src/button/interfaces.ts
```diff
@@ -162,6 +119,53 @@ export interface ButtonProps extends BaseComponentProps {
   fullWidth?: boolean;
```
jperals, line 119: Do we have use cases for full width Toggle Buttons?

src/toggle-button/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+      props: { fullWidth, iconName, pressedIconName, pressed, variant, wrapText },
+    });
+    const baseProps = getBaseProps(props);
+
+    return (
+      <InternalToggleButton
+        {...baseProps}
+        {...baseComponentProps}
+        ref={ref}
+        iconName={iconName}
```
jperals, line 52: Have we considered setting the `fill` property of the SVG on our end instead of requiring the `pressed` icon counterpart? This would not only make its usage easier but also reduce the risk of erroneous usage from customers.

And we could constrain the set of icon names that can be used to a few that we know would work well with this approach (this feature only makes sense with a very reduced number of icons anyway).

src/toggle-button/internal.tsx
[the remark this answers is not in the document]
georgylobko: yes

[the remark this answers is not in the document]
jperals: So why should `onClick` be necessary then?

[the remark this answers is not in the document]
georgylobko: Good question. In this case they are interchangeable, so the `onClick` could be removed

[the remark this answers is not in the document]
jperals: We can always add it later if we need it, but in case of doubt, it's usually preferable to lean towards not adding new props because removing them is harder than adding them (or impossible)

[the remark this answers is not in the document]
georgylobko: updated

src/button/interfaces.ts
georgylobko: updated

src/__tests__/outer-form-submit.test.tsx
georgylobko: Fair enough, removed it

src/button/interfaces.ts
georgylobko: No. Updated

src/toggle-button/index.tsx
georgylobko: Yes, we considered giving customers more control over this component in the doc `quip/ZwfaAocOlIHK/API-Proposal-Toggle-button#temp:C:LXedc371330557f46ffb874e0dc5` but decided to introduce new states for pressed icons instead to avoid possible confusions. 

> And we could constrain the set of icon names that can be used to a few that we know would work well with this approach (this feature only makes sense with a very reduced number of icons anyway).

yes, it's already described in the doc `quip/wWzGA4MOBt9D/Component-article-Toggle-Button#temp:C:KZU2646763fff38442aad70540b9`

pages/toggle-button/permutations.page.tsx
georgylobko: I think adding such examples does not bring much value, because:
1. Such examples are already provided in button permutation component, and toggle button inherits these props
2. The switching logic is the same as for icon name and it is already covered with u tests

src/button/interfaces.ts
georgylobko: updated

src/toggle-button/index.tsx
jperals: My concern is that some customers might use the component e.g, like this:

```
<ToggleButton pressed={pressed} iconName="star" onChange={...}>Favourite</ToggleButton>
```

without realizing they also need to set `iconNamePressed` (or `iconSVGPressed` or `iconURLPressed`). Until they manually interact with it, the component will good to them and TypeScript will not yield any error either.

So the question is whether the component could actually be used like that without the `pressed` props, e.g, by us internally adding the `-filled` icon to the icon name, or by setting the `fill` attribute in the SVG.

georgylobko: This behavior will be described in the usage guideline for this component (currently drafted in the doc `quip/wWzGA4MOBt9D/Component-article-Toggle-Button#temp:C:KZU2646763fff38442aad70540b9`)

> So the question is whether the component could actually be used like that without the pressed props, e.g, by us internally adding the -filled icon to the icon name, or by setting the fill attribute in the SVG.

We considered this option, but came to realize that we have no consistency over names in our iconography, e.g. filled icon might have either `-filled` or `-action` postfix, so we cannot rely on them.

jperals: > We considered this option, but came to realize that we have no consistency over names in our iconography, e.g. filled icon might have either -filled or -action postfix, so we cannot rely on them.

But we have control over which icon names to use internally, and we could constrain the accepted icon names to a subset of those which have a filled version.

Anyway, if that decision is already taken, what about at least issuing dev warnings if the corresponding `pressed` icon is not defined?

georgylobko: Sounds reasonable. Added

jperals (approved):

src/toggle-button/index.tsx
jperals: Thanks. Now you just need to add coverage to those warnings to let the coverage check pass :)

jperals (dismissed):

style-dictionary/classic/metadata.ts
```diff
@@ -13,6 +13,7 @@ const updatedDescriptions: StyleDictionary.TokenIndex<string> = {
     'The color of clickable elements in their default state. *For example: expandable sections, tabs, and icons.*',
   colorTextInteractiveHover:
     'The color of clickable elements on hover. *For example: expandable sections, and icons on hover.*',
+  colorTextToggleButtonIconPressed: 'The pressed text color of icon toggle buttons.',
```
timogasda, line 16: Why does this token have a separate entry in the `classic` style-dictionary?

georgylobko: Good catch, we don't need it. Removed.

src/toggle-button/interfaces.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  /**
+   * Sets the toggle button to pressed state.
+   */
+  pressed: boolean;
+
+  /**
+   * Provides a reason why the button is disabled (only when `disabled` is `true`).
+   * If provided, the button becomes focusable.
+   * Applicable only for the normal toggle-button variant.
```
jperals, line 66: Could this sentence be just "Applicable only for the normal variant"?

Otherwise I read "toggle-button" as the name of the variant, but the variant name is "normal". "toggle-button" refers to the component but it is not necessary to mention it (besides, kebab-case is rather technical rather than human-friendly writing).

georgylobko: updated

style-dictionary/visual-refresh/metadata/colors.ts
```diff
@@ -609,6 +624,11 @@ const metadata: StyleDictionary.MetadataIndex = {
     public: true,
     themeable: true,
   },
+  colorTextToggleButtonIconPressed: {
+    description: 'The pressed text color of icon toggle buttons.',
```
jperals, line 628: But toggle buttons of icon variant don't have text, do they? What is this token for?

georgylobko: We normally use colorText tokens for icons' colors

style-dictionary/visual-refresh/metadata/colors.ts
```diff
@@ -23,6 +23,11 @@ const metadata: StyleDictionary.MetadataIndex = {
     themeable: true,
     public: true,
   },
+  colorBackgroundToggleButtonNormalPressed: {
```
jperals, line 26: Why is only the pressed state of the toggle button themeable?

georgylobko: because they inherit all tokens from the button component

src/toggle-button/styles.scss
```diff
@@ -0,0 +1,18 @@
+/*
+ Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ SPDX-License-Identifier: Apache-2.0
+*/
+
+@use '../internal/styles/tokens' as awsui;
+
+.variant-normal.pressed {
```
jperals, line 8: I have some concerns about relying on selector specificity here. If adding for example a selector like `variant-normal.something-else` somewhere in the button component, specificity will start to get uncertain and hard to manage.

Usually I try to avoid CSS overrides altogether by using the `:not` pseudo-selector, like
```
.variant-normal.pressed {...}
.variant-normal:not(.pressed) {...}
```

Although the `pressed` class doesn't conceptually belong to the Button component (i.e, not Toggle Button) strictly speaking, maybe it is still reasonable to add such pseudo-selectors in there if this avoids bigger refactors.

georgylobko: the toggle-button does not have any specific styles for not pressed state. what is the benefit of introducing additional complexity with preudo-selectors?

jperals: > the toggle-button does not have any specific styles for not pressed state

No but the button component [does](https://github.com/cloudscape-design/components/blob/dbcc7a18b8d7847c4ccc64d38fab6d85684ead42/src/button/styles.scss), that is the problem. And actually I think _we rely_ on subtle CSS specificity calculations because for example for the hover background color, we rely on this selector generated in the button component
```
.button.variant-normal:hover
```

to be more specific than this one generated in the toggle button component:
```
.variant-normal.pressed
```

Anyway, let's do the following: can you make sure that our pipeline integration tests include visual regression tests for hover states? Some of our components already have such tests so it shouldn't be difficult.

jperals (approved):

PR #2498, opened 2024-07-23
chore: Add additional metadata metrics to table and collection preferences

avinashbot:
### Description

Supporting an automated internal use of the component metrics. We can use these metrics to find more complex components/patterns that don't match up with our usage guidelines and UX best practices (like uses of sticky headers, or not allowing disabling table features enabled by default). Keeping details out of this public PR, but feel free to reach out. Adding another context felt safe because it's only used internally, there isn't a wide gap between the components (as in, unlikely to be split across bundles), and it doesn't break any end-user experience if it doesn't match up. Maybe long-term, this could be "generified" into a linting context or something, but not for now.

### How has this been tested?

No coded test, just manually made sure the shape/values align with expectations.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/table/index.tsx
+++ b/src/table/index.tsx
@@ -3,6 +3,7 @@
 import React from 'react';
 
 import { AnalyticsFunnelSubStep } from '../internal/analytics/components/analytics-funnel';
+import { CollectionPreferencesMetadata } from '../internal/context/collection-preferences-metadata-context';
 import useBaseComponent from '../internal/hooks/use-base-component';
 import { applyDisplayName } from '../internal/utils/apply-display-name';
 import { TableForwardRefType, TableProps } from './interfaces';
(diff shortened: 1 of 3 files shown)


src/table/index.tsx
```diff
@@ -37,6 +42,12 @@ const Table = React.forwardRef(
         progressiveLoading: !!props.getLoadingStatus,
         inlineEdit: props.columnDefinitions.some(def => !!def.editConfig),
         disabledInlineEdit: props.columnDefinitions.some(def => !!def.editConfig?.disabledReason),
+        hasSortableColumns: props.columnDefinitions.some(def => def.sortingField || def.sortingComparator),
+        hasHiddenColumns,
+        hasStickyColumns,
+        hasFilterSlot: !!props.filter,
+        hasPaginationSlot: !!props.pagination,
+        itemCount: items.length,
```
abdhalees, line 50: how useful is this? since this is the item count in current page and not the full item count

avinashbot: I was going to say this is the most reliable data we have, but I realized we can actually get the full item count 😅

abdhalees: I feel like this makes the data more unreliable, I wouldn't have a way to detrmine if this number is total items count or a page count

avinashbot: Hmm... it doesn't really matter for my case, but you're right, we can't be shoving unreliable values into the analytics. I can split it up into two values and provide both and pick out what I need at the consuming end.

abdhalees (approved):

PR #2502, opened 2024-07-23
feat: Add analytics metadata to Link

fralongo:
### Description

Similar to https://github.com/cloudscape-design/components/commit/0d1ae085dcf1985c414ac145f30562dc4be29e7c

### How has this been tested?

Additional unit tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/link/analytics-metadata/interfaces.ts
+++ b/src/link/analytics-metadata/interfaces.ts
@@ -1,0 +1,25 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { LabelIdentifier } from '@cloudscape-design/component-toolkit/internal/analytics-metadata/interfaces';
+
+interface GeneratedAnalyticsMetadataLinkClick {
+  action: 'click';
+  detail: {
+    label: string | LabelIdentifier;
+    href?: string;
+    external: string;
+  };
+}
+
+interface GeneratedAnalyticsMetadataLinkComponent {
+  name: 'awsui.Link';
+  label: string | LabelIdentifier;
+  properties: {
+    variant: string;
+  };
+}
+
+export interface GeneratedAnalyticsMetadataLinkFragment extends GeneratedAnalyticsMetadataLinkClick {
+  component?: GeneratedAnalyticsMetadataLinkComponent;
+}
--- a/src/link/index.tsx
+++ b/src/link/index.tsx
@@ -2,8 +2,11 @@
 // SPDX-License-Identifier: Apache-2.0
 import React from 'react';
 
+import { getAnalyticsMetadataAttribute } from '@cloudscape-design/component-toolkit/internal/analytics-metadata';
+
 import useBaseComponent from '../internal/hooks/use-base-component';
 import { applyDisplayName } from '../internal/utils/apply-display-name';
+import { GeneratedAnalyticsMetadataLinkFragment } from './analytics-metadata/interfaces';
 import { LinkProps } from './interfaces';
 import InternalLink from './internal';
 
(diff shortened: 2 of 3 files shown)


src/link/index.tsx
```diff
@@ -14,6 +17,21 @@ const Link = React.forwardRef(
     const baseComponentProps = useBaseComponent('Link', {
       props: { color, external, fontSize, rel: props.rel, target: props.target, variant: props.variant },
     });
+
+    const analyticsMetadata: GeneratedAnalyticsMetadataLinkFragment = {
+      action: 'click',
+      detail: {
+        label: '',
```
just-boris, line 24: The more I look at these empty strings, the more I dislike them. Same suggestion as the previous PR: https://github.com/cloudscape-design/components/pull/2479#discussion_r1686203948

You will be grateful yourself for the fix, as it makes your future investigations of corner-cases easier

src/link/index.tsx
```diff
@@ -14,6 +17,21 @@ const Link = React.forwardRef(
...
+        href: props.href || '',
```
just-boris, line 25: Why do you cast to an empty string here? Missing href and empty attribute are both valid and distinct use-cases. I think they should be reported respectively

src/link/analytics-metadata/interfaces.ts
```diff
@@ -0,0 +1,23 @@
...
+
+interface GeneratedAnalyticsMetadataLinkComponent {
+  name: 'awsui.Link';
+  label: string;
+  properties: {
+    variant: string;
+  };
+}
+
+export interface GeneratedAnalyticsMetadataLinkFragment extends Partial<GeneratedAnalyticsMetadataLinkClick> {
```
just-boris, line 21: What is the point of this `Partial`? All fields are set anyway

src/link/index.tsx
```diff
@@ -14,6 +17,21 @@ const Link = React.forwardRef(
...
+        external: `${external}`,
+      },
+      component: {
+        name: 'awsui.Link',
+        label: '',
+        properties: { variant: props.variant || 'secondary' },
```
just-boris, line 31: If `secondary` is our default variant, can we resolve it in a single place instead of passing a fictional value here?

src/link/index.tsx
fralongo: What is your suggestion to address this? setting it to undefined will not work

just-boris: Why will this not work?

src/link/analytics-metadata/interfaces.ts
fralongo: Removed

src/link/index.tsx
fralongo: I made it optional and set it only if present

src/link/index.tsx
fralongo: The default variant is a bit tricky, as it depends on a React Context. I decided to keep this simpler. https://github.com/cloudscape-design/components/blob/main/src/link/internal.tsx#L59

src/link/index.tsx
fralongo: because the lack of `label` indicates to the metadata resolution algorithm (https://github.com/cloudscape-design/component-toolkit/blob/main/src/internal/analytics-metadata/utils.ts#L11) that there is no label to resolve.

fralongo: There is an option to use selector :scope, but it would make the object more complicated

just-boris: So, there are 3 possible use-cases

1. `label` is `null` which prevents any resolution
1. `label` is a string which is treated as CSS-selector
1. If `label` is an empty string or no match found, resort to data attribute

I see a solution with extending LabelIdentifier interface https://github.com/cloudscape-design/component-toolkit/blob/main/src/internal/analytics-metadata/interfaces.ts#L37-L40

We can add `label: { disabled: true }` value instead of `null`  and then we will avoid null and empty string as different values

src/link/index.tsx
just-boris: This way you will not be able to tell apart explicit "secondary" variant and assumed value. Up to you, but the current reported value cannot be used for any decisions, IMO

just-boris (dismissed):

src/link/index.tsx
fralongo: Will go for `label: { root: "self" }`after merging https://github.com/cloudscape-design/component-toolkit/pull/84

just-boris (approved):

PR #2506, opened 2024-07-23
feat: Allow to hide entered text label in autosuggest

gloncakv:
### Description

Allow the hideEnteredTextLabel to be set for Autosuggest. This will allow to hide the default _"Use: user input"_ option that's displayed when typing into Autosuggest. This is useful when Autosuggest is used to suggest options from a closed set.

The default behavior is the same as the current behavior, i.e. the entered text label is visible.

### How has this been tested?

- Added unit tests
- Added a test page

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/autosuggest/hide-entered-text-label.page.tsx
+++ b/pages/autosuggest/hide-entered-text-label.page.tsx
@@ -1,0 +1,65 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import React, { useContext, useRef, useState } from 'react';
+
+import Autosuggest, { AutosuggestProps } from '~components/autosuggest';
+import Toggle from '~components/toggle';
+
+import AppContext, { AppContextType } from '../app/app-context';
+
+const empty = <span>Nothing found</span>;
+
+const options = [
+  { value: 'Option 0', tags: ['tag1', 'tag2'], filteringTags: ['bla', 'opt'], description: 'description1' },
+  { value: 'Option 1', labelTag: 'This is a label tag' },
+  { value: 'Option 2' },
+  { value: 'Option', description: 'description2' },
+];
+
+const enteredTextLabel = (value: string) => `Use: ${value}`;
+
+type PageContext = React.Context<
+  AppContextType<{
+    hasOptions?: boolean;
+    hasEmpty?: boolean;
+    hasEnteredTextLabel?: boolean;
+  }>
+>;
+
+export default function AutosuggestPage() {
+  const [value, setValue] = useState('');
+  const {
+    urlParams: { hasOptions = true, hasEmpty = true, hasEnteredTextLabel = true },
+    setUrlParams,
+  } = useContext(AppContext as PageContext);
+  const ref = useRef<AutosuggestProps.Ref>(null);
+  return (
+    <div style={{ padding: 10 }}>
+      <h1>Hide entered text labebel in autosuggest</h1>
+      <Autosuggest
+        ref={ref}
+        value={value}
+        readOnly={false}
+        options={hasOptions ? options : []}
+        onChange={event => setValue(event.detail.value)}
+        enteredTextLabel={enteredTextLabel}
+        hideEnteredTextLabel={!hasEnteredTextLabel}
+        ariaLabel={'hide entered text label in autosuggest'}
+        selectedAriaLabel="Selected"
+        empty={hasEmpty ? empty : null}
+      />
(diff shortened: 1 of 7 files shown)


pages/autosuggest/hide-entered-text-label.page.tsx
```diff
@@ -0,0 +1,47 @@
...
+  { value: 'Option 1', labelTag: 'This is a label tag' },
+  { value: 'Option 2' },
+  { value: 'Option', description: 'description2' },
+];
+const enteredTextLabel = (value: string) => `Use: ${value}`;
+export default function AutosuggestPage() {
+  const [value, setValue] = useState('');
+  const [hasOptions, setHasOptions] = useState(true);
+  const [hasEmpty, setHasEmpty] = useState(true);
+  const [hideEnteredTextLabel, setHideEnteredTextLabel] = useState(false);
```
pan-kot, line 19: nit: for settings it is better to use page context so that the configuration stays in the url and can be easily used in integ tests, see: https://github.com/cloudscape-design/components/blob/main/pages/autosuggest/async.page.tsx#L37

pages/autosuggest/hide-entered-text-label.page.tsx
```diff
@@ -0,0 +1,47 @@
...
+        readOnly={false}
+        options={hasOptions ? options : []}
+        onChange={event => setValue(event.detail.value)}
+        enteredTextLabel={enteredTextLabel}
+        hideEnteredTextLabel={hideEnteredTextLabel}
+        ariaLabel={'hide entered text label in autosuggest'}
+        selectedAriaLabel="Selected"
+        empty={hasEmpty ? empty : null}
+      />
+      <button id="hide-entered-text-label" onClick={() => setHideEnteredTextLabel(true)}>
```
pan-kot, line 36: nit: these can be toggle buttons, you can use the Cloudscape Toggle component for this. This way one can see the current state and change it w/o need to reload page

src/autosuggest/internal.tsx
```diff
@@ -179,7 +180,7 @@ const InternalAutosuggest = React.forwardRef((props: InternalAutosuggestProps, r
   const highlightedOptionIdSource = useUniqueId();
   const highlightedOptionId = autosuggestItemsState.highlightedOption ? highlightedOptionIdSource : undefined;
 
-  const isEmpty = !value && !autosuggestItemsState.items.length;
+  const isEmpty = !autosuggestItemsState.items.length;
```
pan-kot, line 183: should this change be made only for the case when hideEnteredTextLabel=true?

pages/autosuggest/hide-entered-text-label.page.tsx
gloncakv: This will be much easier to understand, thanks! I'll update this PR to add it.

pages/autosuggest/hide-entered-text-label.page.tsx
gloncakv: Makes sense! I'll add the page context in the next PR update

src/autosuggest/internal.tsx
gloncakv: Based on my testing and understanding of the [options controller](https://github.com/cloudscape-design/components/blob/main/src/autosuggest/options-controller.ts#L70), I've found that when `value` is defined and `hideEnteredTextLabel` is false, `autosuggestItemsState.items` always includes at least one item—the entered text label. Therefore, I've removed the `!value`. Please let me know if this works.

src/autosuggest/interfaces.ts
```diff
@@ -80,6 +80,11 @@ export interface AutosuggestProps
    */
   enteredTextLabel?: AutosuggestProps.EnteredTextLabel;
 
+  /**
+   * Specifies whether to display the custom value indicator (for example, `Use "${value}"`).
+   */
+  hideEnteredTextLabel?: boolean;
```
timogasda, line 86: Do we really need a new property? Or could we just hide the entered text option if `enteredTextLabel` is undefined/empty ?

ethangeyer: @timogasda wouldn't that be less backwards compatible, as it would impact the existing default? Additionally, we expect the majority of autosuggest customers to continue with the current default, wheras this property would be a smaller optional subset of use cases.

gloncakv:
Reopening this from the original repository: https://github.com/cloudscape-design/components/pull/2526

just-boris:
Opened to check something. please ignore. I will continue in #2526

PR #2513, opened 2024-07-24
feat: Add analytics metadata to Table

fralongo:
### Description

Similar to https://github.com/cloudscape-design/components/commit/0d1ae085dcf1985c414ac145f30562dc4be29e7c

### How has this been tested?

Additional unit tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/table/header-cell/index.tsx
+++ b/src/table/header-cell/index.tsx
@@ -3,11 +3,14 @@
 import React, { useRef } from 'react';
 import clsx from 'clsx';
 
+import { getAnalyticsMetadataAttribute } from '@cloudscape-design/component-toolkit/internal/analytics-metadata';
+
 import { useInternalI18n } from '../../i18n/context';
 import InternalIcon from '../../icon/internal';
 import { useSingleTabStopNavigation } from '../../internal/context/single-tab-stop-navigation-context';
 import { useUniqueId } from '../../internal/hooks/use-unique-id';
 import { KeyCode } from '../../internal/keycode';
+import { GeneratedAnalyticsMetadataTableSort } from '../analytics-metadata/interfaces';
 import { TableProps } from '../interfaces';
 import { Divider, Resizer } from '../resizer';
 import { StickyColumnsModel } from '../sticky-columns';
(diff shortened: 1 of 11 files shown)


src/table/selection/selection-control.tsx
```diff
@@ -98,6 +105,14 @@ export function SelectionControl({
         className={clsx(styles.label, styles.root)}
         aria-label={ariaLabel}
         title={ariaLabel}
+        {...(rowIndex !== undefined && !sharedProps.disabled
+          ? getAnalyticsMetadataAttribute({
+              detail: {
+                position: `${rowIndex + 1}`,
+                item: itemKey || '',
+              },
+            } as Partial<GeneratedAnalyticsMetadataTableSelect>)
```
avinashbot, line 114: Why is this only partial? I can understand `label`, but why is `selected` omitted?

src/table/header-cell/index.tsx
```diff
@@ -105,6 +109,17 @@ export function TableHeaderCell<ItemType>({
       columnId={columnId}
       stickyState={stickyState}
       tableRole={tableRole}
+      {...(sortingDisabled
+        ? {}
+        : getAnalyticsMetadataAttribute({
+            action: 'sort',
+            detail: {
+              position: `${colIndex + 1}`,
```
avinashbot, line 117: Not a strong blocker, but out of curiosity, why are these 1-indexed, if they're presumably meant to be consumed by programmatic clients or a technical audience?

src/table/header-cell/index.tsx
```diff
@@ -105,6 +109,17 @@ export function TableHeaderCell<ItemType>({
...
+              columnId: column.id ? `${column.id}` : '',
+              label: `.${analyticsSelectors['header-cell-text']}`,
+              sortingDescending: `${!sortingDescending}`,
+            },
+          } as GeneratedAnalyticsMetadataTableSort))}
```
avinashbot, line 122: In the long-term, using type parameters on the `getAnalyticsMetadataAttribute` (and related) function might be less easy to forget than checking types using `as`.

src/table/selection/selection-control.tsx
fralongo: because it comes from the Checkbox inside of it

src/table/header-cell/index.tsx
fralongo: I find that to be more intuitive. the 'first' column is column N. 1. This data might end up in dashboards that non-tech audience looks at

avinashbot: I can see someone unaware of 1-indexing being tripped up by this, but at the same time, we're not using a term like `index` here, so I'm not extremely worried.

avinashbot (approved):

PR #2526, opened 2024-07-26
feat: Allow to hide entered text label in autosuggest

gloncakv:
### Description

This is a copy of https://github.com/cloudscape-design/components/pull/2506 from another fork.

Allow the hideEnteredTextLabel to be set for Autosuggest. This will allow to hide the default _"Use: user input"_ option that's displayed when typing into Autosuggest. This is useful when Autosuggest is used to suggest options from a closed set.

The default behavior is the same as the current behavior, i.e. the entered text label is visible.

### How has this been tested?

- Added unit tests
- Added a test page

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/autosuggest/interfaces.ts
+++ b/src/autosuggest/interfaces.ts
@@ -81,6 +81,11 @@ export interface AutosuggestProps
   enteredTextLabel?: AutosuggestProps.EnteredTextLabel;
 
   /**
+   * Specifies whether to display the custom value indicator (for example, `Use "${value}"`).
+   */
+  hideEnteredTextLabel?: boolean;
+
+  /**
    * Specifies the text to display with the number of matches at the bottom of the dropdown menu while filtering.
    */
   filteringResultsText?: (matchesCount: number, totalCount: number) => string;
--- a/src/autosuggest/interfaces.ts
+++ b/src/autosuggest/interfaces.ts
@@ -81,7 +81,7 @@ export interface AutosuggestProps
   enteredTextLabel?: AutosuggestProps.EnteredTextLabel;
 
   /**
-   * Specifies whether to display the custom value indicator (for example, `Use "${value}"`).
+   * Hides the custom value indicator controlled with the `enteredTextLabel` property.
    */
   hideEnteredTextLabel?: boolean;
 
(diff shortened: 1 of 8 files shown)


src/autosuggest/interfaces.ts
```diff
@@ -80,6 +80,11 @@ export interface AutosuggestProps
    */
   enteredTextLabel?: AutosuggestProps.EnteredTextLabel;
 
+  /**
+   * Specifies whether to display the custom value indicator (for example, `Use "${value}"`).
```
pan-kot, line 84: I am not sure if this line will be correctly presented in the website markdown.

How about we change the description and say that the value indicator is controlled by the `enteredTextLabel` property?

src/autosuggest/internal.tsx
```diff
@@ -179,7 +180,7 @@ const InternalAutosuggest = React.forwardRef((props: InternalAutosuggestProps, r
   const highlightedOptionIdSource = useUniqueId();
   const highlightedOptionId = autosuggestItemsState.highlightedOption ? highlightedOptionIdSource : undefined;
 
-  const isEmpty = !value && !autosuggestItemsState.items.length;
```
pan-kot, line 182: I understood why you added this change after playing with the new playground page (which is great 👍).

However, I think it would be great to have the interaction between hideEnteredTextLabel, empty text, and options be explicitly covered with unit tests.

I recommend designing 7 test cases that represent all possible combination of the states and where each test asserts the presence of the dropdown, the presence of the "Use: ..." option, the presence of Empty text, and the presence of options. Like:

```
it('hideEnteredTextLabel=true, emptyText=true, options=false', () => {
  // render
  // assert
})
```

What do you think?

src/autosuggest/interfaces.ts
pan-kot: Oh, nevermind - I just figured out this text is used in the enteredTextLabel already.

Still, it might be more clear if we say like: 
```
Hides the custom value indicator controlled with `enteredTextLabel` property
```

gloncakv: Thanks, that wording is clearer, will update.

src/autosuggest/internal.tsx
gloncakv: Sounds like a good idea! I added some unit tests in [66e066b](https://github.com/cloudscape-design/components/pull/2526/commits/66e066b57fc71f21c98863ffb297ee38dbd5004b), let me know what you think

johannes-weber:
This PR is being closed because of inactivity.

PR #2543, opened 2024-08-05
fix: Handle dynamic cells within table

gethinwebster:
### Description

** Currently in draft - still experimenting **

This improves handling of cells with dynamic content, to ensure that column widths are recalculated when this content changes.

Related links, issue #, if available: AWSUI-52826

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/table/dynamic-content.page.tsx
+++ b/pages/table/dynamic-content.page.tsx
@@ -1,0 +1,117 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import React, { useEffect, useState } from 'react';
+
+import { useCollection } from '@cloudscape-design/collection-hooks';
+
+import Button from '~components/button';
+import CollectionPreferences, { CollectionPreferencesProps } from '~components/collection-preferences';
+import Header from '~components/header';
+import Pagination from '~components/pagination';
+import Table from '~components/table';
+import TextFilter from '~components/text-filter';
+
+import { contentDisplayPreferenceI18nStrings } from '../common/i18n-strings';
+import ScreenshotArea from '../utils/screenshot-area';
+import { generateItems, Instance } from './generate-data';
+import {
+  columnsConfig,
+  contentDisplayPreference,
+  defaultPreferences,
+  EmptyState,
+  getMatchesCountText,
+  pageSizeOptions,
+  paginationLabels,
+} from './shared-configs';
+
+const allItems = generateItems();
+
+const DynamicContent = ({ text }: { text?: string }) => {
+  const [loaded, setLoaded] = useState(false);
+  useEffect(() => {
+    setTimeout(() => setLoaded(true), 2000);
+  }, []);
+  return <div>{loaded ? text : <span>Empty</span>}</div>;
+};
+
+const dynamicColumnsConfig = columnsConfig.map<(typeof columnsConfig)[0]>(c =>
+  c.id === 'dnsName'
+    ? {
+        ...c,
+        hasDynamicContent: true,
+        cell: i => <DynamicContent text={i.dnsName} />,
+      }
+    : c
+);
+
+export default function App() {
+  const [preferences, setPreferences] = useState<CollectionPreferencesProps.Preferences>(defaultPreferences);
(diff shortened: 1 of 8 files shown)


src/internal/hooks/use-mutation-observer/index.ts
```diff
@@ -1,32 +1,18 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
-import React, { useEffect } from 'react';
-
-import { createSingletonHandler, useStableCallback } from '@cloudscape-design/component-toolkit/internal';
-
-const useMutationSingleton = createSingletonHandler<void>(handler => {
-  const observer = new MutationObserver(() => handler());
-  observer.observe(document.body, { attributes: true, subtree: true });
```
gethinwebster, line 9: this wasn't being actively used anywhere as far as I could tell, and I couldn't understand why it observes on `document.body` (and then does no subtree filtering to actually limit to `elementRef` in the main hook below

src/table/use-column-widths.tsx
```diff
@@ -140,11 +148,24 @@ export function ColumnWidthsProvider({ visibleColumns, resizableColumns, contain
...
   // Observes container size and requests an update to the last cell width as it depends on the container's width.
-  useResizeObserver(containerRef, ({ contentBoxWidth: containerWidth }) => {
+  useResizeObserver(measureRef, ({ contentBoxWidth: containerWidth }) => {
     containerWidthRef.current = containerWidth;
     updateColumnWidths();
   });
 
+  useMutationObserver(
+    containerRef,
+    () => {
```
pan-kot, line 158: This can not only impact the rendering performance but also the runtime performance of features such as resizable columns and sticky columns, where the updates to styles are bound to the pointer movements.

gethinwebster: Good point. I wonder whether observing each cell individually might have some benefits in that sense. I'll continue investigating and experimenting

pan-kot: Why the mutation observer is needed? Theoretically, any update to the table cell should cause browser to update its width accordingly which should cause the measure div to grow and trigger the resize observer.

src/internal/hooks/use-mutation-observer/index.ts
pan-kot: This is likely a leftover and the current use cases in components if any probably do not require such implementation. However, the same approach is still used in the toolkit (https://github.com/cloudscape-design/component-toolkit/blob/0dd16b944b60f25ec39b4354513fff3009122f6e/src/internal/visual-mode/index.ts#L54) and it is actually needed because we use it in the tools like useCurrentMode that ensure a class-name to be set. As we don't require the class-name to target a specific component, the update fires on any attribute change of any node to inform the requester which then tries to find a parent with the required class name. It is true that the approach is far from being efficient.

src/table/use-column-widths.tsx
pan-kot: I don't think observing each cell is needed. Whenever the browser increases size of a single cell it also affect the size of the entire column and entire table. It should be enough to only observe the table (but not the container which is scrollable and does not increase its width).

gethinwebster: ah yes, got what you mean now about the measure div. I'll look into this. Thanks for the insight!

pages/table/dynamic-content.page.tsx
```diff
@@ -0,0 +1,116 @@
...
+  pageSizeOptions,
+  paginationLabels,
+} from './shared-configs';
+
+const allItems = generateItems();
+
+const DynamicContent = ({ text }: { text?: string }) => {
+  const [loaded, setLoaded] = useState(false);
+  useEffect(() => {
+    setTimeout(() => setLoaded(true), 2000);
```
pan-kot, line 32: nit: Would it make sense to replace the timeout activation with a button activation? That would require to decouple the state so that clicking a button won't cause the re-render of the component that renders a table. An advantage is that manual and integration tests will be better straightforward.

pan-kot: to make this work you can use e.g. a custom window event so that React subscriptions and memos can be avoided

PR #2545, opened 2024-08-05
fix: Fix tabs dismiss button test-utils

pan-kot:
### Description

Fixing tabs test utils for dismiss button so that the selectors return the button and not its wrapper.

Rel: [AWSUI-54108]

### How has this been tested?

* Updated existing unit tests
* Dry-run to live

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/tabs/tab-header-bar.scss
+++ b/src/tabs/tab-header-bar.scss
@@ -32,10 +32,14 @@ $label-horizontal-spacing: awsui.$space-xs;
   // do not use pointer-events none because it disables scroll by sliding
 
   // Hide scrollbar in all browsers
-  -ms-overflow-style: none; /* Internet Explorer 10+ */
-  scrollbar-width: none; /* Firefox */
+  -ms-overflow-style: none;
+  /* Internet Explorer 10+ */
+  scrollbar-width: none;
+
+  /* Firefox */
   &::-webkit-scrollbar {
-    display: none; /* Safari and Chrome */
+    display: none;
+    /* Safari and Chrome */
   }
 }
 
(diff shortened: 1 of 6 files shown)


src/tabs/tab-header-bar.scss
```diff
@@ -87,11 +87,6 @@ $label-horizontal-spacing: awsui.$space-xs;
   @include styles.text-wrapping;
 }
 
-.tabs-tab-dismiss,
-.tabs-tab-action {
```
just-boris, line 91: What happened to this class? Why is it just gone without replacement?

src/test-utils/dom/tabs/index.ts
```diff
@@ -75,7 +76,7 @@ export default class TabsWrapper extends ComponentWrapper<HTMLButtonElement> {
    */
   findDismissibleButtonByTabId(id: string): ButtonWrapper | null {
     return this.findComponent(
-      `.${styles['tabs-tab-link']}[data-testid="${id}"] ~ .${styles['tabs-tab-dismiss']}`,
+      `.${styles['tabs-tab-link']}[data-testid="${id}"] ~ .${styles['tabs-tab-dismiss']} .${testUtilStyles['tab-dismiss-button']}`,
```
just-boris, line 79: How about simplifying this whole chain and introduce a separate data-attribute `data-testid="awsui-tab-dismiss-button-${id}"` to find this button directly?

pan-kot: Will update

src/tabs/tab-header-bar.scss
just-boris: 😕 it seems like it also could be simplified same way with a direct test id.

But if it does not cause issues right now, I will approve this revision

just-boris (approved):

PR #2553, opened 2024-08-07
feat: Support percentage widths in resizable table columns

gethinwebster:
### Description

Support percentage (rather than pixel) widths in resizable table columns, so that more responsive table layouts can be created.

Related links, issue #, if available: AWSUI-22206

### How has this been tested?

New integ tests. Will add screenshot tests post-release.

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/table/use-column-widths.tsx
```diff
@@ -16,15 +16,18 @@ export interface ColumnWidthDefinition {
   width?: string | number;
 }
 
+const treatAsNumber = (value: number | string | undefined) =>
+  typeof value === 'number' || (value && parseInt(value, 10).toString() === value);
```
georgylobko, line 20: Wouldn't it be better to check if `parseInt` returns `NaN` using the built-in `isNaN` function?

gethinwebster: that wouldn't work because `parseInt('10%')` returns `10` rather than `NaN`, and we explicitly want to treat that sort of string as a string, not a number

georgylobko: Got it. In this case, I would put this function in utils and cover it with unit tests. WDYT?

gethinwebster: sounds a good plan!

pages/table/resizable-columns-percentage.page.tsx
```diff
@@ -0,0 +1,233 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
pan-kot, line 1: Would it make sense to modify the existing resizable-columns page with an ability to define column widths using some settings?

src/table/column-widths-utils.ts
```diff
@@ -40,3 +40,6 @@ function checkProperty(column: TableProps.ColumnDefinition<any>, name: 'width' |
     );
   }
 }
+
+export const treatAsNumber = (value: number | string | undefined) =>
```
pan-kot, line 44: Not sure if it is an issue in this context but `treatAsNumber(NaN)` and `treatAsNumber('NaN')` produce true.

pages/table/resizable-columns-percentage.page.tsx
gethinwebster: makes sense, I was worried about making it overly complex, but I think it works better overall having it in one place

src/table/__integ__/resizable-columns.test.ts
```diff
@@ -373,3 +379,42 @@ test.each([false, true])(
...
+      { percentageWidths: true },
+      { width: 1200 }
+    )
+  );
+
+  test(
+    'percentage column widths do not change on browser resize',
+    setupTest(
+      async page => {
+        await page.setWindowSize({ ...defaultScreen, width: 1400 });
```
pan-kot, line 400: Why is width set here and also on line 405, as an argument?

gethinwebster: the second one is the "on browser resize" part (i.e. the page has already been rendered in `setupTest`, this is now changing the width with a fully-rendered table

src/table/column-widths-utils.ts
gethinwebster: good catch, fixed for completeness

pan-kot:
@gethinwebster when running the test page in dev mode I am getting the warnings like:

```
[AwsUi] [Table] resizableColumns feature requires width property to be a number, got 20%. The component may work incorrectly.
```

pan-kot:
@gethinwebster I found some weird behaviour reproducible in a slightly updated version of the test page: https://github.com/cloudscape-design/components/pull/2560

As you can see in the screen recording, adding "Description" columns also increases the width of the first column when percentages are used.

https://github.com/user-attachments/assets/dee5e3a8-0a66-4e3e-9708-4e86482f790d

pages/table/resizable-columns.page.tsx
```diff
@@ -215,7 +229,9 @@ export default function App() {
           key={renderKey}
           header={<Header>Simple table</Header>}
           stickyHeader={stickyHeader}
-          columnDefinitions={columns.map(col => (withColumnIds ? col : { ...col, id: undefined }))}
+          columnDefinitions={columns
+            .map(col => (withColumnIds ? col : { ...col, id: undefined }))
+            .map(col => ({ ...col, width: percentageWidths ? col.percentageWidth ?? col.width : col.width }))}
```
jperals, line 234: Nit: you are passing the `percentageWidth` attribute here which is not part of the `ColumnDefinition` interface. I guess TypeScript does not complain because of the spread syntax

src/table/use-column-widths.tsx
```diff
@@ -23,8 +23,8 @@ function readWidths(
   const result = new Map<PropertyKey, number>();
   for (let index = 0; index < visibleColumns.length; index++) {
     const column = visibleColumns[index];
-    let width = (column.width as number) || 0;
-    const minWidth = (column.minWidth as number) || width || DEFAULT_COLUMN_WIDTH;
+    let width = (treatAsNumber(column.width) && (column.width as number)) || 0;
```
jperals, line 26: I don't see this logic using the percentage, am I missing something? If `column.width` is a percentage like `3%`, `treatAsNumber` will return false so `width` will be 0.

gethinwebster: As it is 0 it then hits the `!width` condition below, and reads the current width from the DOM element. So basically, this part of code 'freezes' the percentage-calculated width

https://github.com/cloudscape-design/components/pull/2553/files#diff-1c6f5a8d52d6e96b2165e4eeb6d8bcbe2674cf66f0e03f078a1e69e8767f5c11R29

jperals:
If this feature is aimed to work together with resizable columns, should we update the interface docs for Table.ColumnDefinition? Currently they state:

>  When `resizableColumns` property is set to `true`, additional constraints apply: 1) string values are not allowed

gethinwebster:
> If this feature is aimed to work together with resizable columns, should we update the interface docs for Table.ColumnDefinition? Currently they state:
> 
> >

good point, I thought this was only mentioned in the Development Guidelines, but will check through and make updates to API docs too.

gethinwebster:
> @gethinwebster when running the test page in dev mode I am getting the warnings like:
> 
> ```
> [AwsUi] [Table] resizableColumns feature requires width property to be a number, got 20%. The component may work incorrectly.
> ```

Updated to remove these

gethinwebster:
> If this feature is aimed to work together with resizable columns, should we update the interface docs for Table.ColumnDefinition? Currently they state:
> 
> >

Updated with fix for this.

src/table/interfaces.tsx
```diff
@@ -440,6 +439,7 @@ export namespace TableProps {
     isRowHeader?: boolean;
     verticalAlign?: VerticalAlign;
     cell(item: ItemType): React.ReactNode;
+    hasDynamicContent?: boolean;
```
pan-kot, line 442: This one is probably accidental

src/table/use-column-widths.tsx
```diff
@@ -180,7 +180,7 @@ export function ColumnWidthsProvider({ visibleColumns, resizableColumns, contain
     setColumnWidths(() => readWidths(getCell, visibleColumns));
     // This code is intended to run only at the first render and should not re-run when table props change
```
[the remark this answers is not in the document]
gethinwebster: This fixes the issue you identified about dynamically adding a column, so I think it makes sense to keep in in this PR. But good point on the comment.

src/table/use-column-widths.tsx
pan-kot: Is there a chance to make this code easier to follow? We can maybe add a helper function for width and min-width extraction

src/table/use-column-widths.tsx
```diff
@@ -178,9 +178,10 @@ export function ColumnWidthsProvider({ visibleColumns, resizableColumns, contain
       return;
     }
     setColumnWidths(() => readWidths(getCell, visibleColumns));
-    // This code is intended to run only at the first render and should not re-run when table props change
+    // This code is intended to run only at the first render (or when rendering a column that was not previously rendered)
+    // and should otherwise not re-run when table props change
```
pan-kot, line 182: The `visibleColumns` is not memoized so now this effect that was previously used in the initial render only is executed for every render, e.g. when a column was resized or a row was selected.

gethinwebster: Simplified this logic to combine the two `useEffect` hooks that were doing similar but subtly different things.

pan-kot:
Is it expected that the width is applied differently for when resizable columns is on or off?

In my example, I am only setting width to the first and second column (300% and 300px accordingly).

https://github.com/user-attachments/assets/ee08d33e-b263-4e28-a0ae-8eb34e7d86e7

src/table/use-column-widths.tsx
```diff
@@ -145,8 +139,7 @@ export function ColumnWidthsProvider({ visibleColumns, resizableColumns, contain
     updateColumnWidths();
   });
 
-  // The widths of the dynamically added columns (after the first render) if not set explicitly
-  // will default to the DEFAULT_COLUMN_WIDTH.
+  // Re-calculate widths of any "new" columns - either on first render, or dynamically added
   useEffect(() => {
```
pan-kot, line 143: It would be great if we can add more code comments explaining what is actually happening inside as it is quite hard to read. For instance, there are calls like updateColumnWidths(), setColumnWidth(...), readWidth(...) - and what exactly happening is not obvious.

src/table/use-column-widths.tsx
```diff
@@ -16,25 +16,19 @@ export interface ColumnWidthDefinition {
...
-      const colEl = getCell(column.id);
-      width = colEl ? getLogicalBoundingClientRect(colEl).inlineSize : DEFAULT_COLUMN_WIDTH;
-    }
-    result.set(column.id, Math.max(width, minWidth));
+  let width = (treatAsNumber(column.width) && (column.width as number)) || 0;
+  const minWidth = (treatAsNumber(column.minWidth) && (column.minWidth as number)) || width || DEFAULT_COLUMN_WIDTH;
+  // read width from the DOM if it is missing in the config
+  if (!width && !isLastColumn) {
+    const colEl = getCell(column.id);
+    width = colEl ? getLogicalBoundingClientRect(colEl).inlineSize : DEFAULT_COLUMN_WIDTH;
```
pan-kot, line 29: If I understood correctly, we call readWidth on each column to compare it with the prev state to know if the configuration changed, like a new column was added or its width/minWidth/maxWidth change. In that case, would it be possible to avoid reading the actual element width?

gethinwebster:
Closing, will not address immediately.

PR #2374, opened 2024-06-14
chore: Add app layout toolbar internal component

just-boris:
### Description

Adding new version of app layout to the mainline to avoid permanent rebases

Related links, issue #, if available: n/a

### How has this been tested?

Currently, only Typescript build. Making existing app layout tests passing against all design variants will be a separate task

<details>
   <summary>Review checklist</summary>

_The following items are to be evaluated by the author(s) and the reviewer(s)._

#### Correctness

- _Changes include appropriate documentation updates._
- _Changes are backward-compatible if not indicated, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#public-apis)._
- _Changes do not include unsupported browser features, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#browsers-support)._
- _Changes were manually tested for accessibility, see [accessibility guidelines](https://cloudscape.design/foundation/core-principles/accessibility/)._

#### Security

- _If the code handles URLs: all URLs are validated through [the `checkSafeUrl` function](https://github.com/cloudscape-design/components/blob/main/src/internal/utils/check-safe-url.ts)._

#### Testing

- _Changes are covered with new/existing unit tests?_
- _Changes are covered with new/existing integration tests?_
</details>

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2559, opened 2024-08-08
chore: Restore panel animations with toolbar

jkuelz:
### Description

Animations for panels got deprioritized and removed when merging the toolbar code into mainline (#2374).
This commit adds them back. It also fixes a bug (?) where VR-only animations would appear in classic.
Prior to this commit, VR only had animations on side nav, tools (help), and bottom positioned split panel. Drawers and side split panel did not have animations.

### How has this been tested?

Manually checked VR and classic. Ran through dev pipeline as well. 
Classic doesn't have animation requirements.

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/app-layout/styles.scss
```diff
@@ -7,11 +7,25 @@
...
+    // stylelint-disable-next-line @cloudscape-design/no-implicit-descendant
+    .with-motion {
+      // stylelint-disable-next-line @cloudscape-design/no-motion-outside-of-mixin
+      transition: none;
+      // stylelint-disable-next-line @cloudscape-design/no-motion-outside-of-mixin
+      animation: none;
+    }
   }
   &-active iframe {
     // Prevents iframes from eating the mouseup event if the cursor is on top of it.
```
just-boris, line 31: If we are going to reuse these styles, let's extract them in a separate file, `src/app-layout/resize/styles.scss` for example

src/app-layout/visual-refresh-toolbar/drawer/index.tsx
```diff
@@ -100,17 +101,9 @@ export function AppLayoutDrawerImplementation({ appLayoutInternals }: AppLayoutD
...
           />
         </div>
-        {toolsContent && (
-          <div
-            className={clsx(
-              styles['drawer-content'],
-              activeDrawerId !== TOOLS_DRAWER_ID && styles['drawer-content-hidden']
-            )}
-          >
-            {toolsContent}
```
just-boris, line 110: This special treatment for the `toolsContent` has a reason. I think this page is broken with the change: https://github.com/cloudscape-design/components/blob/main/pages/app-layout/runtime-drawers-imperative.page.tsx

src/app-layout/visual-refresh-toolbar/drawer/styles.scss
```diff
@@ -4,6 +4,7 @@
 */
 @use '../../../internal/styles/tokens' as awsui;
 @use '../../../internal/styles' as styles;
+@use '../../../internal/generated/custom-css-properties/index.scss' as custom-props;
```
just-boris, line 7: Is this needed? I see no related usage

src/app-layout/visual-refresh-toolbar/index.tsx
```diff
@@ -303,8 +304,7 @@ const AppLayoutVisualRefreshToolbar = React.forwardRef<AppLayoutProps.Ref, AppLa
           toolsOpen={!!activeDrawer}
           toolsWidth={activeDrawerSize}
           sideSplitPanel={
-            splitPanelPosition === 'side' &&
-            splitPanel && (
```
just-boris, line 307: Why is this removed from the condition?

src/app-layout/visual-refresh-toolbar/skeleton/index.tsx
```diff
@@ -57,6 +59,7 @@ export function SkeletonLayout({
 }: SkeletonLayoutProps) {
   const isMaxWidth = maxContentWidth === Number.MAX_VALUE || maxContentWidth === Number.MAX_SAFE_INTEGER;
   const anyPanelOpen = navigationOpen || toolsOpen;
+  const isMobile = useMobile();
```
just-boris, line 62: There was a plan to avoid using `useMobile` hook in this code to allow server-side-rendering. Is it possible to keep mobile overrides with CSS media queries only?

src/split-panel/bottom.tsx
```diff
@@ -59,23 +53,29 @@ export function SplitPanelContentBottom({
   return (
     <div
       {...baseProps}
-      className={clsx(baseProps.className, styles.drawer, styles['position-bottom'], testUtilStyles.root, {
-        [testUtilStyles['open-position-bottom']]: isOpen,
-        [styles['drawer-closed']]: !isOpen,
-        [styles['drawer-mobile']]: isMobile,
-        [styles['drawer-disable-content-paddings']]: disableContentPaddings,
-        [styles.animating]: isRefresh && (state === 'entering' || state === 'exiting'),
```
just-boris, line 67: This change impacts the non-toolbar version as well. Is it expected?

src/split-panel/implementation.tsx
```diff
@@ -185,77 +185,71 @@ export const SplitPanelImplementation = React.forwardRef<HTMLElement, SplitPanel
...
      * of the following code.
      */
-    if (isRefresh && !isOpen && position === 'side') {
+    if (isRefresh && !isAppLayoutToolbarEnabled() && !isOpen && position === 'side') {
       return <></>;
     }
 
     return (
-      <Transition in={isOpen ?? false}>
-        {(state, transitioningElementRef) => (
```
just-boris, line 194: This change impacts even the classic design? is it expected too?

src/app-layout/visual-refresh-toolbar/drawer/styles.scss
jkuelz: Nope, removed

src/app-layout/visual-refresh-toolbar/index.tsx
jkuelz: I think this was to make it consistent with the bottom orientation, because it seemed redundant with [this line](https://github.com/cloudscape-design/components/blob/main/src/app-layout/visual-refresh-toolbar/skeleton/index.tsx#L100). Is this condition doing more than that?

src/split-panel/bottom.tsx
jkuelz: Refresh (non-toolbar) version now gets animations via the `.with-motion` class too. I checked manually that the split panel still has motion upon entering with and without the toolbar in VR. You can verify on [this dev page](https://d21d5uik3ws71m.cloudfront.net/components/bed78a5d67a934450dbdc2ed470148103d9bdd22/dev-pages/index.html#/light/app-layout/with-split-panel). Is there somewhere you saw the animations not being applied as expected?

src/split-panel/implementation.tsx
jkuelz: Classic isn't expected to have any animations. The [current classes](https://github.com/cloudscape-design/components/blob/main/src/split-panel/motion.scss#L9-L21) only apply to refresh and I also checked manually that this was the case. It shouldn't have any impact.

src/app-layout/visual-refresh-toolbar/index.tsx
just-boris: Okay, consistency is a good reason. I also cannot recall reasons why this condition was added

src/split-panel/bottom.tsx
just-boris: If you tested this in non-toolbar-VR, it is all good. Just needed to call out the impact radius of this PR

src/app-layout/visual-refresh-toolbar/drawer/index.tsx
```diff
@@ -100,7 +103,7 @@ export function AppLayoutDrawerImplementation({ appLayoutInternals }: AppLayoutD
             variant="icon"
           />
         </div>
-        {toolsContent && (
+        {
```
just-boris, line 106: If we removed the condition, we do not need the braces `{}` anymore

src/app-layout/visual-refresh-toolbar/skeleton/index.tsx
```diff
@@ -102,7 +101,11 @@ export function SkeletonLayout({
           {sideSplitPanel}
         </div>
       )}
-      {tools && <div className={clsx(styles.tools, !toolsOpen && styles['panel-hidden'])}>{tools}</div>}
+      {
```
just-boris, line 104: Also not needed `{}` here

src/split-panel/side.tsx
```diff
@@ -37,13 +38,21 @@ export function SplitPanelContentSide({
...
+        {
+          [testUtilStyles['open-position-side']]: isOpen,
+          [styles['drawer-closed']]: !isOpen,
+          [styles['with-toolbar']]: isAppLayoutToolbarEnabled(),
+          [styles.refresh]: isRefresh,
+        }
+      )}
       style={{
-        width: isOpen && isRefresh ? cappedSize : undefined,
+        width: isOpen ? cappedSize : '0px',
```
just-boris, line 55: With this change, the width for classic becomes `0` too, but in fact the closed drawer has some width there.

image


---

It looks working because of some CSS magic, but still looks very sus? Can we keep it unchanged for classic?

just-boris (commented):
There is some weird motion happening when resizing the screen with the open side split panel:

https://github.com/user-attachments/assets/7691730a-1885-4dd2-964a-eb435a2a41f2

The same does not happen in the previous VR app layoit

just-boris (approved):

just-boris (approved):

PR #2563, opened 2024-08-12
fix: Consistently fire onSelect even when closing the dropdown

just-boris:
### Description

Fixing an inconsistency in the events handling

1. When the input field is focused, pressing enter closes the dropdown, but no `onSelect` event fired
2. When the "Use current value" option is focused, the dropdown closes, with an `onSelect` event

Both interactions are the same for the end user – the dropdown closes while keeping the current value, so they should produce the same event.

For the context, this is a long way to fix AWSUI-53412. In that issue on our website there is a bug with the search autosuggest. It uses a workaround for the missing event. After this event is added, we can remove the workaround

### How has this been tested?

Locally + updated existing events test to include the new events.

Dry-runs if anybody relied on the broken behavior: 6982713945, 6982715033, 6982780091

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/property-filter/index.tsx
+++ b/src/property-filter/index.tsx
@@ -13,7 +13,6 @@ import TokenList from '../internal/components/token-list';
 import { fireNonCancelableEvent } from '../internal/events';
 import useBaseComponent from '../internal/hooks/use-base-component';
 import { useUniqueId } from '../internal/hooks/use-unique-id/index';
-import { KeyCode } from '../internal/keycode';
 import { applyDisplayName } from '../internal/utils/apply-display-name';
 import { joinStrings } from '../internal/utils/strings';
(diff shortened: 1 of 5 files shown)


src/property-filter/index.tsx
```diff
@@ -235,12 +234,6 @@ const PropertyFilter = React.forwardRef(
       addToken(newToken);
       setFilteringText('');
     };
-    const ignoreKeyDown = useRef<boolean>(false);
```
just-boris, line 238: This code should not exist in the first place. It was added to get around the missing `onSelect` event, which is fixed now

johannes-weber (approved):
Great simplification, thanks for fixing!

PR #2574, opened 2024-08-16
chore: Added Trusted Types to iconSVG

Hoxmot:
I expanded the type of `components/src/internal/plugins/controllers/drawers.ts` to also support the [Trusted Types API](https://developer.mozilla.org/en-US/docs/Web/API/Trusted_Types_API). In short, Trusted Types is a browser API, which allows the developers to write the code with a better awareness of the injection sinks (e.g., `dangerouslySetInnerHTML` in React) usage. With the Trusted Types enforcement enabled, each injection sinks requires the usage of a trusted object, which can have a sanitization or other kinds of allowlisting in place.

With this new change, the `iconSvg` in `DrawerConfig` supports `string | TrustedHTML`.

As a result, the developer experience has improved, since now they can use something as simple as:

```ts
const trustedPolicy: TrustedTypePolicy = ...;
awsuiPlugins.appLayout.registerDrawer({
  ...
  trigger: {
    // ✅ preferable
    iconSvg: trustedPolicy.createHTML(svgHtml),
  },
});
```

Instead of type casting just to make it compliant:

```ts
const trustedPolicy: TrustedTypePolicy = ...;
appLayout.registerDrawer({
  ...
  trigger: {
    // ❌ cumbersome
    iconSvg: trustedPolicy.createHTML(svgHtml) as any as string,
  },
});
```

The already existing implementation of `AppLayout` implemented in `components/src/app-layout/runtime-api.tsx` supports the trusted types, since `dangerouslySetInnerHTML` API already accepts type `string | TrustedHTML` (types update might be required though).

⚠️ However, if someone implements the consumer themselves, this _might be_ a breaking change.

### How has this been tested?

I ran `npm test` locally.

* `npm run test:unit`:
```
Test Suites: 369 passed, 369 total
Tests:       46 skipped, 6881 passed, 6927 total
Snapshots:   96 passed, 96 total
Time:        72.377 s
```
* `npm run test:integ`:
```
Test Suites: 141 passed, 141 total
Tests:       2 skipped, 1180 passed, 1182 total
Snapshots:   8 passed, 8 total
Time:        369.777 s
```
* `npm run test:a11y`:
```
Test Suites: 6 passed, 6 total
Tests:       8 skipped, 1894 passed, 1902 total
Snapshots:   0 total
Time:        503.107 s, estimated 533 s
```
* `npm run test:motion`:
```
Test Suites: 1 passed, 1 total
Tests:       2 passed, 2 total
Snapshots:   0 total
Time:        26.34 s
```

<details>

#### Correctness

- 🤷 _Changes include appropriate documentation updates._
- ❌ _Changes are backward-compatible if not indicated, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#public-apis)._
- ✅ _Changes do not include unsupported browser features, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#browsers-support)._
- ✅ _Changes were manually tested for accessibility, see [accessibility guidelines](https://cloudscape.design/foundation/core-principles/accessibility/)._

#### Security

- ✅ _If the code handles URLs: all URLs are validated through [the `checkSafeUrl` function](https://github.com/cloudscape-design/components/blob/main/src/internal/utils/check-safe-url.ts)._

#### Testing

- ✅ _Changes are covered with new/existing unit tests?_
- ✅ _Changes are covered with new/existing integration tests?_
</details>

---


just-boris:
Did a bunch of test builds on the internal infra, this fails too many consumers because of `@types/react` version bump. This needs an alternative solution

Hoxmot:
I updated the branch with an additional commit to pin the `@types/react` version to `16.14.40`. This should minimise the scope of the changes. The version in the current `main` is `16.14.34` (based on [package-lock.json](https://github.com/cloudscape-design/components/blob/main/package-lock.json#L3608-L3616)) 

Previously, I used `^16.14.40`, which bumped the version to `16.14.60` (newest `16.14.x` version), which included a change of moving `JSX` from the global scope to the `React` scope.

just-boris:
Completed several dry-runs (6988134134, 6988133849, 6988133252) the change seems compatible. Now let's clean up some things and we can merge this

package.json
```diff
@@ -71,12 +71,13 @@
...
     "@types/lodash": "^4.14.176",
     "@types/node": "^18.18.5",
-    "@types/react": "^16.14.20",
+    "@types/react": "16.14.40",
     "@types/react-dom": "^16.9.14",
     "@types/react-resizable": "^1.7.4",
     "@types/react-router": "^5.1.18",
     "@types/react-router-dom": "^5.3.2",
     "@types/react-transition-group": "^4.4.4",
+    "@types/trusted-types": "^2.0.7",
```
just-boris, line 80: This dependency is problematic. It is supposed to be a bult-in `dom.d.ts` type, but temporarily it exists as a library.

When it becomes built-in, our consumers may have problems upgrading to the latest typescript because of that. 

Is type-safety here really that much important, or can we get away with type casting

```js
iconSvg: trustedPolicy.createHTML(svgHtml) as any as string,
```

Hoxmot:
After investigating a bit further, you're right. I'm not sure I can make to 100% working correctly with all the consumers.

I guess I can go back to it, once the trusted types related types (incl. `TrustedHTML`) are added to TypeScript. Thanks for your time.

PR #2585, opened 2024-08-19
chore: Property filter token editor integration

pan-kot:
### Description

Pass internal token editor props via property filter internal to feature the new editor in the editor test page.

Added a change to popover internal to allow no size setting: that is something we implicitly support as passing a size not matching small, medium or large does the same.

### How has this been tested?

* New test page asset
* Existing token editor permutations
* Existing unit- and integration tests
* Screenshot tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/popover/container.scss
+++ b/src/popover/container.scss
@@ -57,7 +57,7 @@
 }
 
 .container-body-size-content {
-  // no styles
+  /* no styles */
 }
 
 .container-arrow {
--- a/src/popover/container.scss
+++ b/src/popover/container.scss
@@ -56,6 +56,10 @@
   }
 }
 
+.container-body-size-content {
+  // no styles
+}
+
 .container-arrow {
   $arrow-offset: 12px;
   $arrow-width: 20px;
--- a/src/popover/container.scss
+++ b/src/popover/container.scss
@@ -56,10 +56,6 @@
   }
 }
 
-.container-body-size-content {
-  /* no styles */
-}
-
 .container-arrow {
   $arrow-offset: 12px;
   $arrow-width: 20px;
--- a/src/popover/container.tsx
+++ b/src/popover/container.tsx
@@ -30,7 +30,7 @@ export interface PopoverContainerProps {
   arrow: (position: InternalPosition | null) => React.ReactNode;
   children: React.ReactNode;
   renderWithPortal?: boolean;
-  size: PopoverProps.Size;
+  size: PopoverProps.Size | 'content';
   fixedWidth: boolean;
   variant?: 'annotation';
   // When keepPosition is true, the popover will not recalculate its position when it resizes nor when it receives clicks.
--- a/src/popover/container.tsx
+++ b/src/popover/container.tsx
@@ -30,6 +30,8 @@ export interface PopoverContainerProps {
   arrow: (position: InternalPosition | null) => React.ReactNode;
   children: React.ReactNode;
   renderWithPortal?: boolean;
+  // The "content" size is a special one that does not have the associated container-body-size-{size} class.
+  // It can be used to create larger popovers where the responsiveness is handled by the content.
   size: PopoverProps.Size | 'content';
   fixedWidth: boolean;
   variant?: 'annotation';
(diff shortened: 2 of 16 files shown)


src/popover/container.scss
```diff
@@ -56,6 +56,10 @@
   }
 }
 
+.container-body-size-content {
+  // no styles
+}
```
just-boris, line 61: 1. What is happening here?
2. When using `//` comment (instead of `/**/`), SASS removes them during compilation, so this entire thing is erased. Whatever you wanted to do here, it does not work

src/property-filter/internal.tsx
```diff
@@ -0,0 +1,456 @@
...
+export type PropertyFilterInternalProps = SomeRequired<
+  PropertyFilterProps,
+  'filteringOptions' | 'customGroupsText' | 'disableFreeTextFiltering'
+> &
+  InternalBaseComponentProps & {
+    enableTokenGroups?: boolean;
+    i18nStringsExt?: I18nStringsExt;
+  };
+
+const PropertyFilterInternal = React.forwardRef(
```
just-boris, line 62: Am I supposed to review this whole file, or is there anything in particular changed after moving from `property-filter/index.ts`?

src/property-filter/internal.tsx
```diff
@@ -0,0 +1,456 @@
...
+} from './interfaces';
+import { PropertyEditor } from './property-editor';
+import PropertyFilterAutosuggest, { PropertyFilterAutosuggestProps } from './property-filter-autosuggest';
+import { TokenButton } from './token';
+import { useLoadItems } from './use-load-items';
+import { matchTokenValue } from './utils';
+
+import styles from './styles.css.js';
+
+export interface I18nStringsExt {
```
just-boris, line 44: What is this object? Why are these not a part of the public API?

src/popover/container.scss
pan-kot: We add size styles as `container-body-size-${size}` and the style defined the responsiveness of the popover.

For the property filter tokens we want the popover to grow larger and a different approach is used to change layouts for smaller screens.

src/property-filter/internal.tsx
pan-kot: There is nothing special but extending the internal props with `enableTokenGroups` and `i18nStringsExt`. The first prop is a flag that is to become public once we are ready to release token groups. The second prop is a set of extra properties to be added to i18nStrings.

src/popover/container.scss
just-boris: You should use `clsx` library which automatically skip undefined class names. We only render empty test-utils classes, all others should not exist

pan-kot: My idea was to add this class as a code documentation in a way, but I can replace it with a code comment. Will update

just-boris (approved):

src/popover/container.tsx
```diff
@@ -30,7 +30,9 @@ export interface PopoverContainerProps {
   arrow: (position: InternalPosition | null) => React.ReactNode;
   children: React.ReactNode;
   renderWithPortal?: boolean;
-  size: PopoverProps.Size;
+  // The "content" size is a special one that does not have the associated container-body-size-{size} class.
+  // It can be used to create larger popovers where the responsiveness is handled by the content.
```
just-boris, line 34: I do not think this is something that needs explanation. Not every `variant-${variantName}` class in any other component has this comment. 

Leaving this comment you add more confusion than clarity.

src/property-filter/filtering-token/index.tsx
```diff
@@ -79,7 +81,7 @@ const FilteringToken = forwardRef(
       content: editorContent,
       triggerType: 'text',
       header: editorHeader,
-      size: 'large',
+      size: popoverSize,
```
pan-kot, line 84: @just-boris I added this change so that the token editor popover size depends on the enableTokenGroups flag, that is to limit the effect to the new editor variant that does require a change in size.

I noticed the effect in screenshot tests.

src/popover/container.tsx
pan-kot: removed

src/property-filter/filtering-token/index.tsx
just-boris: ok

just-boris (approved):

PR #2594, opened 2024-08-21
chore: independently testing toolbar trigger buttons

dpitcock:
…rigger-buttons indepenently

### Description

Consolidating the logic for the toolbar trigger buttons and adding unit tests for them

- in Visual Refresh this combines the logic for the mobile or the desktop toolbar trigger button into its own component
- this tests the components for the toolbar trigger buttons in both visual refresh and visual refresh toolbar app layout modes
- additional props were added to the internal components

This also sets up the components nicely for follow up commits that :
- resolution of toolbar focus control issues for both visual refresh and visual refresh toolbar modes
- addition of tooltips to the trigger buttons

### How has this been tested?

- No change in functionality or UI
- Human tested the functionality
- added unit tests

<details>

#### Correctness

#### Security

#### Testing

- _Changes are covered with new/existing unit tests?_ Yes new and existing
- _Changes are covered with new/existing integration tests?_ Yes, existing
</details>

---


src/app-layout/visual-refresh/drawers.tsx
```diff
@@ -19,6 +19,13 @@ import splitPanelTestUtilStyles from '../../split-panel/test-classes/styles.css.
 import testutilStyles from '../test-classes/styles.css.js';
 import styles from './styles.css.js';
 
+/**
+ * The VISIBLE_MOBILE_TOOLBAR_TRIGGERS_COUNT is used to reduce the number
+ * of triggers that are initially visible on the mobile toolbar, the rest
+ * are then placed into an overflow menu
+ */
+export const VISIBLE_MOBILE_TOOLBAR_TRIGGERS_COUNT = 2;
```
georgylobko, line 27: I would rather name it `VISIBLE_MOBILE_TOOLBAR_TRIGGERS_LIMIT`

src/app-layout/visual-refresh/trigger-button.tsx
```diff
@@ -33,37 +52,69 @@ function TriggerButton(
...
-        data-testid={testId}
-      >
-        <span className={clsx(badge && styles['trigger-badge-wrapper'])}>
-          <Icon name={iconName} svg={iconSvg} />
-        </span>
-      </button>
-      {badge && <div className={styles.dot} />}
+    <div
+      className={clsx(styles['trigger-wrapper'], !highContrastHeader && styles['remove-high-contrast-header'])}
+      data-testid={`${testId ? `${testId}-wrapper` : 'awsui-app-layout-trigger-wrapper'}`}
```
georgylobko, line 67: We typically query DOM selectors in our test utils/test cases using CSS classes to avoid potential name collisions with the consumers' code

src/app-layout/visual-refresh/trigger-button.tsx
```diff
@@ -33,37 +52,69 @@ function TriggerButton(
...
+    >
+      {isMobile ? (
```
georgylobko, line 69: Why do we need to render the button differently for mobile view? The current implementation achieves this without any special conditions, so what's the purpose of making it more complicated?

dpitcock: I'm under the assumption there is a reason for it, so I kept it https://github.com/cloudscape-design/components/blob/main/src/app-layout/visual-refresh/drawers.tsx#L342

src/app-layout/visual-refresh/trigger-button.tsx
dpitcock: ok.  I can add classes and change this

src/app-layout/visual-refresh/drawers.tsx
```diff
@@ -237,22 +244,21 @@ function DesktopTriggers() {
...
               ariaControls={activeDrawerId === item.id ? item.id : undefined}
               className={clsx(
                 styles['drawers-trigger'],
                 testutilStyles['drawers-trigger'],
                 item.id === TOOLS_DRAWER_ID && testutilStyles['tools-toggle']
               )}
+              ref={item.id === previousActiveDrawerId.current ? drawersRefs.toggle : undefined}
               iconName={item.trigger.iconName}
               iconSvg={item.trigger.iconSvg}
-              key={item.id}
```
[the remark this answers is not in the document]
dpitcock: No I'm picking out of another pull request, so there can be incremental deployments.  I copy and pasted from there.  Key should be in there,  I will move it back to its original place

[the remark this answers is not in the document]
georgylobko: Which PR did you pick the changes from?

src/app-layout/visual-refresh/trigger-button.tsx
georgylobko: What does the drawers component from the link above do with this change? Also, what is the purpose of the isMobile condition?

src/app-layout/visual-refresh/drawers.tsx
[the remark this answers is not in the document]
dpitcock: My own for tooltip which will be discarded, replaced with a much smaller one that only has logic for tooltips

src/app-layout/visual-refresh/trigger-button.tsx
dpitcock: i removed the testid from here, if needed for the tooltip I will add there.  The isMobile is used to determine which component to render. There is a different layout on mobile. When the drawer is open it covers the toolbar

georgylobko: If the layout is different on mobile, how is it currently being handled without this condition? What caught my attention is this sentence in the PR description:

> No change in functionality or UI

Could you clarify how the mobile layout is managed in the current implementation?

src/app-layout/__tests__/trigger-button.test.tsx
```diff
@@ -0,0 +1,357 @@
...
+      <VFTBTriggerButton {...(renderProps as VFTBTriggerButtonProps)} ref={ref} />
+      <span className={mockOtherEl.class}>{mockOtherEl.text}</span>
+    </div>
+  );
+  const wrapper = createWrapper(container).findByClassName(toolbarTriggerButtonStyles['trigger-wrapper'])!;
+  await delay();
+  return { wrapper, rerender, getByTestId, getByText };
+};
+
+describe('Visual refresh trigger-button (not in appLayoutWidget toolbar)', () => {
```
georgylobko, line 94: What are the advantages of testing these buttons separately, considering they are not self-sufficient components but rather part of layouts?

src/app-layout/__tests__/trigger-button.test.tsx
```diff
@@ -0,0 +1,357 @@
...
+  relatedTarget: {
+    dataset: {
+      shiftFocus: 'last-opened-toolbar-trigger-button',
+    },
+  },
+};
+
+const delay = () => act(() => new Promise(resolve => setTimeout(resolve)));
+
+const renderVFTriggerButton = async (
```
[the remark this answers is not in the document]
dpitcock: There are two components imported with the same name. VF is VisualRefresh, as in the abbreviated folder it was imported from

src/app-layout/__tests__/trigger-button.test.tsx
dpitcock: Isolation of the component, knowing it works as expected in isolation so easier to pinpoint failures from the parent. Also keeps coverage high.  This will be needed later as we add the tooltip functionality to the trigger button, we want to make sure the tooltip is working correctly insolation.

src/app-layout/__tests__/trigger-button.test.tsx
```diff
@@ -0,0 +1,357 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+/* eslint simple-import-sort/imports: 0 */
```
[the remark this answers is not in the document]
dpitcock: It was used in the other tests I modeled this after (and copied from). I assumed it was there for a reason.

src/app-layout/__tests__/trigger-button.test.tsx
```diff
@@ -0,0 +1,357 @@
...
+      <VFTriggerButton {...(renderProps as VFTriggerButtonProps)} ref={ref} />
+      <span className={mockOtherEl.class}>{mockOtherEl.text}</span>
+    </div>
+  );
+  const wrapper = createWrapper(container).findByClassName(visualRefreshStyles['trigger-wrapper'])!;
+  await delay();
+  return { wrapper, rerender, getByTestId, getByText };
+};
+
+const renderVFTBTriggerButton = async (
```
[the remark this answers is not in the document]
dpitcock: would renderVisualRefreshToolbarTriggerButton be better?

dpitcock (commented):
I was more explicit in the naming convention for imports and render functions in the test

src/app-layout/__tests__/trigger-button.test.tsx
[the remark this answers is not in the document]
dpitcock: Updated naming convention

src/app-layout/__tests__/trigger-button.test.tsx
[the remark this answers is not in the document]
dpitcock: removed

src/app-layout/__tests__/trigger-button.test.tsx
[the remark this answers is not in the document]
dpitcock: Updated naming convention

src/app-layout/__tests__/trigger-button.test.tsx
```diff
@@ -0,0 +1,357 @@
...
+        const ref = React.createRef<React.Ref<ButtonProps.Ref>>();
+        const { wrapper, getByTestId } = await renderVisualRefreshTriggerButton(
+          {
+            iconSvg: icon,
+          },
+          {
+            isMobile,
+            hasOpenDrawer,
+          },
+          ref as any
```
georgylobko, line 175: Why?

src/app-layout/visual-refresh/trigger-button.tsx
georgylobko: I see. But you didn't replace the existing trigger button in the mobile triggers. Is that intentional?

Also, can we use InternalButton consistently and remove the implementation based on <button>?

src/app-layout/visual-refresh/trigger-button.tsx
```diff
@@ -1,24 +1,43 @@
...
+   * A svg node for a custom icon.
+   *
+   * An icon is neccessary for this component to render. If the SVG is not given there must be a iconName prop passed or the component will not render
+   */
   iconSvg?: React.ReactNode;
   ariaExpanded: boolean | undefined;
   ariaControls?: string;
+  disabled?: boolean;
+  /**
+   * Ovewrwrites any internal testIds when provided
```
georgylobko, line 33: Not sure if we need all of the comments for every property, as they are already described in the public API documentation for the button component.

src/app-layout/__tests__/trigger-button.test.tsx
dpitcock: `Argument of type 'RefObject<Ref<Ref>>' is not assignable to parameter of type 'Ref<Ref>'.
      Type 'RefObject<Ref<Ref>>' is not assignable to type 'RefObject<Ref>'.`
      
      the other alternative to any is to cast as `ref as unknown as React.Ref<ButtonProps.Ref>`

src/app-layout/visual-refresh/trigger-button.tsx
dpitcock: These don't use the public API though. This is internal and either passes the props onto an InternalButton or to a html button. The Icon comments need to be updated

src/app-layout/visual-refresh/trigger-button.tsx
dpitcock: The tooltips will eventually be rendered as a sibiling element to the trigger button. The trigger buttons are intended to work as they do currently, but the logic for the tooltip and its handlers will be added here and rendered when called for based on props that will be added here.

src/app-layout/__tests__/trigger-button.test.tsx
```diff
@@ -0,0 +1,357 @@
...
+  } as AllContext.AppLayoutInternals);
+  const renderProps = { ...mockProps, ...props };
+  const { container, rerender, getByTestId, getByText } = render(
+    <div>
+      <VisualRefreshTriggerButton {...(renderProps as VisualRefreshTriggerButtonProps)} ref={ref} />
+      <span className={mockOtherEl.class}>{mockOtherEl.text}</span>
+    </div>
+  );
+  const wrapper = createWrapper(container).findByClassName(visualRefreshStyles['trigger-wrapper'])!;
+  await delay();
```
pan-kot, line 74: why is this delay needed?

src/app-layout/visual-refresh/trigger-button.tsx
```diff
@@ -33,37 +52,66 @@ function TriggerButton(
...
+              {
+                [styles.selected]: selected,
+                [styles.badge]: badge,
+              },
+              className
+            )}
+            onClick={onClick}
+            ref={ref as Ref<HTMLButtonElement>}
+            type="button"
+            data-testid={testId}
```
pan-kot, line 106: The data-testid is not set on the mobile version - why is that?

nit: would it make sense to create some commonProps and pass it to each button? That can include ARIA props, onClick, etc. This way it will be easier to see what is actually different.

src/app-layout/visual-refresh/trigger-button.tsx
```diff
@@ -1,24 +1,43 @@
...
+   *
+   * An icon is neccessary for this component to render. If the name is not given there must be a iconSvg prop passed or the component will not render
+   */
   iconName?: IconProps.Name;
+  /**
+   * A svg node for a custom icon.
+   *
+   * An icon is neccessary for this component to render. If the SVG is not given there must be a iconName prop passed or the component will not render
+   */
   iconSvg?: React.ReactNode;
```
pan-kot, line 28: What is the fallback for the case an icon is missing? Should we add a warning?

src/app-layout/visual-refresh/trigger-button.tsx
pan-kot: oh, I see it is actually provided as a native prop, but why?

src/app-layout/visual-refresh/trigger-button.tsx
```diff
@@ -1,24 +1,43 @@
...
   ariaExpanded: boolean | undefined;
   ariaControls?: string;
+  disabled?: boolean;
+  /**
+   * Ovewrwrites any internal testIds when provided
+   */
   testId?: string;
-  onClick: React.MouseEventHandler<HTMLButtonElement>;
+  /**
+   * If button is selected. Used only for desktop and applies a selected class
```
pan-kot, line 37: why selected only applies for desktop?

src/app-layout/__tests__/trigger-button.test.tsx
```diff
@@ -0,0 +1,357 @@
...
+      ref as any
+    );
+    expect(wrapper).not.toBeNull();
+    const button = wrapper.find('button')!;
+    expect(getByTestId(mockTestId)).toBeTruthy();
+    button.click();
+    expect(mockClickSpy).toBeCalledTimes(disabledValue ? 0 : 1);
+  });
+
+  test('renders an empty button when no iconName and iconSVG prop', async () => {
```
pan-kot, line 344: would it be nicer to fallback to some icon instead?

src/app-layout/__tests__/trigger-button.test.tsx
dpitcock: I was getting network errors, 404 of I did not have delays. I will see if it still applies

src/app-layout/visual-refresh/trigger-button.tsx
dpitcock: it gets added above

```
 __nativeAttributes={{
            'aria-haspopup': true,
            ...(testId && {
              'data-testid': testId,
            }),
          }}
```

src/app-layout/visual-refresh/trigger-button.tsx
dpitcock: Good question.  I am removing those comments. The button will render with no icon if none provided.

src/app-layout/visual-refresh/trigger-button.tsx
dpitcock: because when the drawer is opened it covers the buttons on mobile. So the styling for selected will not be seen

src/app-layout/__tests__/trigger-button.test.tsx
dpitcock: This can be added in a followup

src/app-layout/__tests__/trigger-button.test.tsx
dpitcock: I did not need. If i need later for the tooltip I will add a comment as to why. Thanks for pointing out

src/app-layout/visual-refresh/trigger-button.tsx
dpitcock: That is how it is currently done. Not sure as to why.

src/app-layout/visual-refresh/trigger-button.tsx
dpitcock: removed docstrings for now

src/app-layout/visual-refresh/trigger-button.tsx
dpitcock: updated docstring

src/app-layout/__tests__/trigger-button.test.tsx
```diff
@@ -0,0 +1,353 @@
...
+    </div>
+  );
+  const wrapper = createWrapper(container).findByClassName(toolbarTriggerButtonStyles['trigger-wrapper'])!;
+  return { wrapper, rerender, getByTestId, getByText };
+};
+
+describe('Visual refresh trigger-button (not in appLayoutWidget toolbar)', () => {
+  beforeEach(() => jest.clearAllMocks());
+
+  //tests for both desktop and mobile
```
pan-kot, line 93: nit: This code comment looks redundant

src/app-layout/__tests__/trigger-button.test.tsx
```diff
@@ -0,0 +1,353 @@
...
+  describe.each([true, false])('Toolbar trigger-button with isMobile=%s', isMobile => {
+    //tests for both drawer being opend and closed
```
pan-kot, line 95: This as well - the variables isMobile and hasOpenDrawer convey enough info, I think.

src/app-layout/__tests__/trigger-button.test.tsx
```diff
@@ -0,0 +1,353 @@
...
+    describe.each([true, false])('AppLayoutInternals with hasOpenDrawer=%s', hasOpenDrawer => {
+      test('renders correctly with without badge', () => {
```
pan-kot, line 97:
```suggestion
      test('renders correctly without badge', () => {
```

src/app-layout/__tests__/trigger-button.test.tsx
```diff
@@ -0,0 +1,353 @@
...
+        const button = wrapper.find('button');
+        expect(getByTestId(mockTestId)).toBeTruthy();
+        expect(button).toBeTruthy();
+        expect(button!.getElement().getAttribute('aria-label')).toEqual(mockProps.ariaLabel);
+        expect(wrapper!.findIcon()).toBeTruthy();
+        expect(wrapper.findByClassName(visualRefreshStyles.dot)).toBeNull();
+        expect(wrapper.findBadge()).toBeNull();
+      });
+
+      test('renders correctly with aria controls', () => {
```
pan-kot, line 120: Can this test be merged with the previous one? It looks like it only verifies if the ariaControls property is assigned.

pan-kot: If not merge, I would at least change the description to something like "aria controls is assigned on the button"

src/app-layout/__tests__/trigger-button.test.tsx
```diff
@@ -0,0 +1,353 @@
...
+
+        expect(wrapper).not.toBeNull();
+        const button = wrapper.find('button');
+        expect(getByTestId(mockTestId)).toBeTruthy();
+        expect(button).toBeTruthy();
+        expect(button!.getElement().getAttribute('aria-label')).toEqual(mockProps.ariaLabel);
+        expect(button!.getElement().getAttribute('aria-controls')).toEqual(mockAriaControls);
+      });
+
+      test('renders correctly with iconName', () => {
```
pan-kot, line 143: I didn't get this test. Is icon name assigned implicitly when the badge is not provided?

pan-kot: The asserts also look very similar to the previous tests, I wonder if there is a value of keeping this one.

src/app-layout/__tests__/trigger-button.test.tsx
```diff
@@ -0,0 +1,353 @@
...
+            hasOpenDrawer,
+          },
+          ref
+        );
+
+        expect(wrapper).not.toBeNull();
+        const button = wrapper.find('button');
+        expect(getByTestId(mockTestId)).toBeTruthy();
+        expect(button).toBeTruthy();
+        expect(button!.getElement().getAttribute('aria-label')).toEqual(mockProps.ariaLabel);
```
pan-kot, line 178: why do we verify aria label in this test?

src/app-layout/__tests__/trigger-button.test.tsx
```diff
@@ -0,0 +1,353 @@
...
+        expect(wrapper!.findIcon()).toBeTruthy();
```
pan-kot, line 179: Is it possible to create some generic test, like:

```
test.each([{ iconName }, { iconSvg }, { iconUrl }])('renders with icon', iconProps => {
  // render with icon props
  // assert the icon is defined
})
```

?

src/app-layout/__tests__/trigger-button.test.tsx
```diff
@@ -0,0 +1,353 @@
...
+        expect(getByTestId(iconTestId)).toBeTruthy();
+      });
+
+      test.each([true, false])('renders correctly and with when disabled is %s', disabledValue => {
```
pan-kot, line 183: The description is unclear: with what it renders when disabled?

pan-kot: nit: I would avoid saying "renders correctly" because that is vague. It is better to express the clear expectation like "aria label is assigned" or "disabled reasons is shown" etc.

src/app-layout/__tests__/trigger-button.test.tsx
```diff
@@ -0,0 +1,348 @@
...
+        expect(button).toBeTruthy();
+        expect(button!.getElement().getAttribute('aria-label')).toEqual(mockProps.ariaLabel);
+        expect(button!.getElement().getAttribute('aria-controls')).toEqual(mockAriaControls);
+      });
+
+      test('renders correctly with iconName', () => {
+        const ref: React.MutableRefObject<ButtonProps.Ref | null> = React.createRef();
+        const { wrapper, getByTestId } = renderVisualRefreshTriggerButton({}, {}, ref);
+
+        expect(wrapper).not.toBeNull();
```
georgylobko, line 143: could it be combined with a previous test? the assertions look pretty much similar

src/app-layout/__tests__/trigger-button.test.tsx
```diff
@@ -0,0 +1,348 @@
...
+    const button = wrapper.find('button');
+    expect(getByTestId(mockTestId)).toBeTruthy();
+    expect(button).toBeTruthy();
+    expect(button!.getElement().getAttribute('aria-label')).toEqual(mockProps.ariaLabel);
+    expect(wrapper!.findIcon()).toBeTruthy();
+    expect(wrapper.findByClassName(visualRefreshStyles.dot)).toBeTruthy();
+  });
+
+  test.each([true, false] as const)('Is focusable using the forwarded ref with mobile is %s', isMobile => {
+    const ref: React.MutableRefObject<ButtonProps.Ref | null> = React.createRef();
```
georgylobko, line 242: why is this type ButtonProps.Ref | null? In this case, it should be guaranteed that this ref is always assigned, shouldn't it?

georgylobko (approved):

PR #2595, opened 2024-08-21
fix: Fixed property filter tokens e2e selector

pan-kot:
### Description

Changed token selector to work consistently in unit- and e2e tests. A downside is that the updated selector matches an element higher in the hierarchy so that e.g. asserting the ARIA label won't work.

An alternative solution is available requiring changes to test-utils code, see:
* https://github.com/cloudscape-design/test-utils/pull/70
* [4QRKAKIJD3RX]

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/property-filter/__integ__/property-filter.test.ts
+++ b/src/property-filter/__integ__/property-filter.test.ts
@@ -1,0 +1,47 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import { BasePageObject } from '@cloudscape-design/browser-test-tools/page-objects';
+import useBrowser from '@cloudscape-design/browser-test-tools/use-browser';
+
+import createWrapper from '../../../lib/components/test-utils/selectors';
+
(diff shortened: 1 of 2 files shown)


src/property-filter/__integ__/property-filter.test.ts
```diff
@@ -0,0 +1,47 @@
...
+  'can create an edit a second token',
+  setupTest(async page => {
+    // Create second token
+    await page.click(wrapper.findNativeInput().toSelector());
+    await page.keys(['i', 'd']);
+    await page.keys(['ArrowDown', 'ArrowDown', 'Enter']);
+    await page.keys(['=']);
+    await page.keys(['ArrowDown', 'ArrowDown', 'Enter']);
+
+    console.log('selector: ', token2Wrapper.toSelector());
```
dpitcock, line 31: Nit. Is the condole.log needed here?

pan-kot: It is not :)

PR #2606, opened 2024-08-22
chore: Add support for ESM in jest

gethinwebster:
### Description

Adjust NodeJS config so that jest tests can cope with ESM (see https://jestjs.io/docs/ecmascript-modules). This is so that we can upgrade webdriverio, which is distributed as ESM in recent versions.

Related links, issue #, if available: AWSUI-51276

### How has this been tested?

Tested in local workspace with upgraded version of webdriverio (https://github.com/cloudscape-design/browser-test-tools/pull/86)

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -162,6 +162,12 @@ npx jest -c jest.motion.config.js src/flashbar
 npx jest -c jest.build-tools.config.js build-tools/stylelint
 ```
 
+Note: when running jest directly you may see errors about `--experimental-vm-modules`, to fix this you can set this NodeJS flag as follows:
+
+```
+export NODE_OPTIONS="$NODE_OPTIONS --experimental-vm-modules"
+```
+
 ### Run visual regression tests
 
 Visual regression tests for the permutation pages are automatically run when opening a pull request in GitHub.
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -12,9 +12,10 @@
     "build": "cross-env NODE_ENV=production gulp build",
     "test": "TZ=UTC gulp test",
     "test:unit": "TZ=UTC gulp test:unit",
-    "test:a11y": "gulp test:a11y",
-    "test:integ": "gulp test:integ",
-    "test:motion": "gulp test:motion",
+    "gulp:integ": "NODE_OPTIONS=\"$NODE_OPTIONS --experimental-vm-modules\" gulp",
```
just-boris, line 15: I already added this flag before: https://github.com/cloudscape-design/components/blob/a35b753b47172f9b51123fc3f7c9a627f647112b/build-tools/tasks/unit.js#L19-L23

How about doing the same here, so even when running `npx gulp ...` it still sets the expected options?

Also, I am not sure this line in our contributing docs will remain working: https://github.com/cloudscape-design/components/blob/a35b753b47172f9b51123fc3f7c9a627f647112b/CONTRIBUTING.md?plain=1#L155-L156

gethinwebster: Nice approach, thanks. Updated the docs too.

just-boris (approved):

PR #2637, opened 2024-08-30
fix: Fixes table sticky col padding in vr

pan-kot:
### Description

Fixing a bug when first sticky column padding is missing in VR

Rel: AWSUI-57462

### How has this been tested?

* Manual tests in Classic, VR, with compact mode, with striped rows
* New screenshot tests, see: CR-146133400

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/table/sticky-columns.page.tsx
+++ b/pages/table/sticky-columns.page.tsx
@@ -22,6 +22,7 @@ type DemoContext = React.Context<
     resizableColumns: boolean;
     stickyHeader: boolean;
     sortingDisabled: boolean;
+    stripedRows: boolean;
     selectionType: undefined | 'single' | 'multi';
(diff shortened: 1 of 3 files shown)


just-boris (approved):
Such issues can and should be tested using an integration test ("expect text left offset in plain and sticky state is the same").

But if you would like to risk and count on screenshot tests only, your call

pan-kot:
> Such issues can and should be tested using an integration test ("expect text left offset in plain and sticky state is the same").
> 
> But if you would like to risk and count on screenshot tests only, your call

That is not quite true for VR, where there is no padding when non-sticky but we add padding when sticky. I am going to proceed with the screenshot tests for now.

PR #2644, opened 2024-09-02
feat: Filter options in Content Display Options

E-Kaese:
### Description

Original ticket ID: AWSUI-22334

This PR adds the support for filtering down the options in the content display options in collection preferences.
The reordering gets disabled when a search starts and the drag handles show the disabled state.

There are 1 new property and a set of i18n strings to support this change:
```
columnFiltering?: boolean;
i18nStrings: {
  columnFilteringPlaceholder?: string;
  columnFilteringAriaLabel?: string;
  columnFilteringCountText?: (count: number) => string;
  columnFilteringNoMatchText?: string;
  columnFilteringClearFilterText?: string;
}
```

[Original filtering request ticket](https://issues.amazon.com/issues/AWSUI-22334)

### How has this been tested?

Manual and unit tests

- Enabled the search filter by setting `searchableColumns` to `true` and setting the supporting text properties.
- Start typing in the text filter input above the reorderable options, the list will filter down and the drag handles will be disabled.
- Clear button will clear the filter and enable the dragging again.
- Filtering by a non-existant option will show an empty state
- The matches count next to the filter input reflects the count of the filtered options

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/collection-preferences/content-display/index.tsx
```diff
@@ -98,6 +113,33 @@ export default function ContentDisplayPreference({
       <p className={getClassName('description')} id={descriptionId}>
         {i18n('contentDisplayPreference.description', description)}
       </p>
+
+      {/* Filter input */}
+      {searchableColumns && (
+        <div className={getClassName('text-filter')}>
+          <InternalTextFilter
+            filteringText={columnFilteringText}
```
pan-kot, line 121: I believe these messages can be standard as the target resource is always columns. In that case there is a value of using the built-in i18n mechanism, see e.g. the 'contentDisplayPreference.description' from a few lines above.

For that to work we'd need to provide translations but as a POC it is fine to modify the messages types files and en translations under src/i18n.

src/collection-preferences/content-display/__tests__/content-display.test.tsx
```diff
@@ -174,6 +174,71 @@ describe('Content Display preference', () => {
     });
   });
 
+  describe('Filtering', () => {
```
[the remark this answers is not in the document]
pan-kot: Also need to add tests to ensure the i18n strings can be injects. We use a TestI18nProvider helper for that. You can maybe even do something like: `describe.each([{ withI18nProvider: false }, { withI18nProvider: true }])(...)` and inject provider in the renderContentDisplay conditionally.

src/i18n/messages-types.ts
```diff
@@ -142,6 +142,10 @@ export interface I18nFormatArgTypes {
       "total": string | number;
       "finalPosition": string | number;
     }
+    "contentDisplayPreference.columnFilteringPlaceholderText": never;
+    "contentDisplayPreference.columnFilteringAriaLabel": never;
+    "contentDisplayPreference.columnFilteringEmptyText": never;
+    "contentDisplayPreference.columnFilteringClearFilterText": never;
   }
```
pan-kot, line 149: What about columnFilteringCountText?

src/internal/components/drag-handle/styles.scss
```diff
@@ -16,3 +16,7 @@
 .handle:active {
   cursor: grabbing;
 }
+
+.handle-disabled {
+  cursor: default !important;
```
pan-kot, line 21: Let's try to avoid `!important` directive. You can achieve the same effect with classes specificity.

src/collection-preferences/content-display/__integ__/pages/content-display-page.ts
```diff
@@ -28,8 +28,13 @@ export default class ContentDisplayPageObject extends CollectionPreferencesPageO
...
   }
 
-  focusDragHandle(index = 0) {
-    return this.keys(new Array(5 + index * 2).fill('Tab'));
+  /**
+   * @param index The index of the drag handle to focus on. (Optional, default 0)
+   * @param offset The number of times to press the tab key to focus on the drag handle.
+   * Use 5 for collection preferences without searchable columns enabled, and 6 as the offset for searchable columns enabled. (Optional, default 5)
+   */
+  focusDragHandle(index = 0, offset = 5) {
```
pan-kot, line 36: nit: Introducing the offset API exposes the detail on how this focusing is done and the consumer has to read the description before they can use it. I'd recommend having something like `{ hasSearchableColumns: boolean }` so that the offset choice can be done inside the method. In that case the added comments might not even be needed.

Alternatively, you can get the info of whether the search is enabled by querying the filter inside the method, like:
```
focusDragHanlde(index) {
  const isSearchable = !!this.wrapper.findModal().findContentDisplayPreference().findTextFilter():
  const offset = isSearchable ? 6 : 5;
  ...
}
```

src/collection-preferences/content-display/__tests__/content-display.test.tsx
```diff
@@ -174,6 +174,107 @@ describe('Content Display preference', () => {
...
+      expect(filterInput?.findInput().findClearButton()?.getElement()).toHaveAccessibleName('Clear filter');
+
+      expect(filterInput!.findResultsCount().getElement()).toHaveTextContent('1 match');
+
+      const options = wrapper.findOptions();
+      expect(options).toHaveLength(1);
+      expect(options[0].findLabel().getElement()).toHaveTextContent('Item 1');
+    });
+
+    it('clears filter and shows all options', () => {
```
pan-kot, line 211: let's remove tests that do not depend on withI18nProvider outside of the describe.each block

src/collection-preferences/content-display/index.tsx
```diff
@@ -51,7 +58,13 @@ export default function ContentDisplayPreference({
   const titleId = `${idPrefix}-title`;
   const descriptionId = `${idPrefix}-description`;
 
-  const sortedOptions = getSortedOptions({ options, contentDisplay: value });
+  const sortedOptions = useMemo(
+    () =>
+      getSortedOptions({ options, contentDisplay: value }).filter(option =>
+        option.label.toLowerCase().includes(columnFilteringText.toLowerCase())
```
pan-kot, line 64: Should we trim filtering text when checking? Also in the condition below: `disabled={columnFilteringText.length > 0}`

src/internal/components/drag-handle/styles.scss
pan-kot: Hey @E-Kaese , any chance we can avoid the `!important` here?

src/collection-preferences/utils.tsx
```diff
@@ -68,21 +68,12 @@ export const ModalContentLayout = ({ left, right }: ModalContentLayoutProps) =>
...
-
   const columns = left && right ? 2 : 1;
   return (
     <div ref={ref}>
-      <InternalColumnLayout columns={columns} variant="text-grid">
+      <InternalColumnLayout columns={smallContainer ? 1 : columns} variant="text-grid">
         {left && <div>{left}</div>}
-        {right && <div>{right}</div>}
+        {right && <div className={clsx(left && smallContainer && styles['second-column-small'])}>{right}</div>}
       </InternalColumnLayout>
```
E-Kaese, line 77: These two return statements were causing the state to be lost in the filtering, updated this section to only have on return statement. When `smallContainer` is true the `columnLayout` is reduced to 1 column to mimic the original layout from the `return` above, it is also used to conditionally apply the relevant class when true

pan-kot (dismissed):

pan-kot (approved):

PR #2648, opened 2024-09-03
feat: Support icon only main action in button dropdown

abdhalees:
### Description

- make `text` property in `mainAction` optional
- remove extra padding when there is no text in main action using the same padding as dropdown trigger

Related links, issue #, if available: HgxDAw1K2GBU

### How has this been tested?

added new permutations

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/button-dropdown/internal.tsx
+++ b/src/button-dropdown/internal.tsx
@@ -210,7 +210,7 @@ const InternalButtonDropdown = React.forwardRef(
       const mainActionAriaLabel = externalIconAriaLabel
         ? `${mainAction.ariaLabel ?? mainAction.text} ${mainAction.externalIconAriaLabel}`
         : mainAction.ariaLabel;
-
+      const hasNoText = !text;
       trigger = (
         <div role="group" aria-label={ariaLabel} className={styles['split-trigger-wrapper']}>
           <div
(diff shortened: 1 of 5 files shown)


src/button-dropdown/styles.scss
```diff
@@ -70,6 +70,14 @@ $dropdown-trigger-icon-offset: 2px;
       border-end-end-radius: 0;
       padding-inline-end: awsui.$space-m;
       margin-inline-end: awsui.$space-xxxs;
+
+      &.has-no-text {
+        padding-inline: awsui.$space-button-icon-only-horizontal;
+
+        &.visual-refresh {
+          padding-inline-start: calc(#{awsui.$space-s} - #{$dropdown-trigger-icon-offset});
+        }
```
just-boris, line 79: Why is this spacing difference is not built into the design token by default?

src/button-dropdown/internal.tsx
```diff
@@ -210,7 +210,7 @@ const InternalButtonDropdown = React.forwardRef(
       const mainActionAriaLabel = externalIconAriaLabel
         ? `${mainAction.ariaLabel ?? mainAction.text} ${mainAction.externalIconAriaLabel}`
         : mainAction.ariaLabel;
-
+      const hasNoText = !text;
```
just-boris, line 213: I would not introduce a constant when it is just used once, but okay

just-boris (approved):

src/button-dropdown/styles.scss
abdhalees: good point, I just reused the existing value

PR #2650, opened 2024-09-03
feat: adding tooltips and tests to visual-refresh-toolbar

dpitcock:
This adds a visual tooltip to the toolbar on visual-refresh-toolbar variant by adding event listeners to the toolbar-trigger-wrapper elements that shows the tooltip when focused or hovered.

Screenshot 2024-09-09 at 10 59 58

### Description

The trigger-button takes new props to determine if a tooltip should be possible. 

### How has this been tested?

-  confirm UI testing on https://d21d5uik3ws71m.cloudfront.net/components/dc5256b023200217c47b78be1b0393963f173c85/dev-pages/index.html#/light/app-layout/with-drawers?appLayoutWidget=true
- unit test for the trigger-button and the drawers added to assert the tooltips work
- integ tests for the tooltips is added

<details>

#### Correctness

#### Security

#### Testing

- _Changes are covered with new/existing unit tests?_ Y
- _Changes are covered with new/existing integration tests?_ Y
</details>

---


package-lock.json
jperals, line 1: These changes in `package-lock.json` are not necessary, right?

src/app-layout/__integ__/app-layout-toolbar-tooltips.test.ts
```diff
@@ -0,0 +1,337 @@
...
+  visualRefresh?: string;
+}
+
+const drawerIds = Object.values(drawerIdObj);
+const VISIBLE_MOBILE_TOOLBAR_TRIGGERS_LIMIT = 2; //must match the number in  '../../../lib/components/app-layout/visual-refresh/drawers';
+const mobileDrawerTriggerIds = drawerIds.slice(0, VISIBLE_MOBILE_TOOLBAR_TRIGGERS_LIMIT);
+
+const setupTest = (
+  {
+    splitPanelPosition = 'bottom',
```
jperals, line 70: Looks like all these params are always called with the same values except for `size`, so they are not necessary?

src/app-layout/__integ__/app-layout-toolbar-tooltips.test.ts
```diff
@@ -0,0 +1,337 @@
...
+        await expect(page.isExisting(`.${visualRefreshStyles['trigger-tooltip']}`)).resolves.toBeFalsy();
+        await expect(
+          page.isExisting(`.${visualRefreshStyles[`drawers-desktop-triggers-container`]}`)
+        ).resolves.toBeTruthy();
+        await expect(page.isExisting(`.${visualRefreshStyles['drawers-trigger-overflow']}`)).resolves.toBeFalsy();
+        await page.hoverElement(wrapper.findDrawerTriggerById(drawerIds[0]).toSelector());
+        await expect(page.isExisting(`.${visualRefreshStyles['trigger-tooltip']}`)).resolves.toBeTruthy();
+
+        for (const drawerId of drawerIds) {
+          async () => {
```
jperals, line 105: Doesn't this `async () { ... }` create a function? Is the block inside actually called? Same in line 120

src/app-layout/__integ__/app-layout-toolbar-tooltips.test.ts
```diff
@@ -0,0 +1,337 @@
...
+            ).resolves.toBe(1);
+            await page.hoverElement(`.${visualRefreshStyles[`drawers-desktop-triggers-container`]}`);
+            await expect(page.isExisting(`.${visualRefreshStyles['trigger-tooltip']}`)).resolves.toBeFalsy();
+            await page.click(`button[data-testid='awsui-app-layout-trigger-${drawerId}']`);
+            await expect(page.isExisting(`.${visualRefreshStyles['trigger-tooltip']}`)).resolves.toBeFalsy();
+            await expect(page.isExisting(wrapper.findActiveDrawer().toSelector())).resolves.toBeTruthy();
+            await page.hoverElement(`.${visualRefreshStyles[`drawers-desktop-triggers-container`]}`);
+            await expect(page.isExisting(wrapper.findActiveDrawer().toSelector())).resolves.toBeTruthy();
+
+            for (const nestedDrawerId of drawerIds) {
```
jperals, line 166: Isn't this looping over the same elements as the parent loop? Is this intended?

src/app-layout/visual-refresh/trigger-button.tsx
```diff
@@ -46,13 +64,119 @@ function TriggerButton(
...
+
+      const handleKeyDownEvent = (event: KeyboardEvent) => {
+        if (event.key === 'Escape') {
+          close();
+        }
+      };
+
+      window.addEventListener('pointerdown', handlePointerDownEvent);
+      window.addEventListener('keydown', handleKeyDownEvent);
+      // console.log('listeners added', {hasTooltip, tooltipVisible})
```
jperals, line 158: Remove this line :)

jperals:
Note that if i blur and focus the browser again, I can get into a situation where multiple tooltips are visible:

Screenshot 2024-09-03 at 15 35 54

package-lock.json
dpitcock: no.  a pull from main and new build resolved them

src/app-layout/__integ__/app-layout-toolbar-tooltips.test.ts
dpitcock: yes. confirming the tooltips work when the drawer is each drawer is opened

src/app-layout/__integ__/app-layout-toolbar-tooltips.test.ts
```diff
@@ -0,0 +1,337 @@
...
+    await testFn(page);
+  });
+
+describe(`theme='visual-refresh'`, () => {
+  describe(`desktop`, () => {
+    const size = 'desktop';
+
+    test(
+      'Shows tooltip correctly for mouse interactions on desktop',
+      setupTest({ disableContentPaddings: 'true', visualRefresh: 'true', size }, async page => {
```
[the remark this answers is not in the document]
dpitcock: I am assuming it sets expectations on the screen. I will try without it. I took from an existing working test.

src/app-layout/__integ__/app-layout-toolbar-tooltips.test.ts
dpitcock: removed the redundant code

src/app-layout/__integ__/app-layout-toolbar-tooltips.test.ts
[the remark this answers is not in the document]
dpitcock: it was not needed. Removed

src/app-layout/__integ__/app-layout-toolbar-tooltips.test.ts
dpitcock: Fixed this. But the tests were timing out so I had to simplify them

src/app-layout/__integ__/app-layout-toolbar-tooltips.test.ts
dpitcock: I was a nice to have but caused tests to time out.  It was robust, signaling that an open drawer did not have an effect on the tooltips but testing against ever trigger was too much. Simplified now

src/app-layout/visual-refresh-toolbar/toolbar/trigger-button/styles.scss
```diff
@@ -106,6 +106,10 @@ handleSplitPanelMaxWidth function in the context.
   border-end-end-radius: 50%;
 }
 
+.trigger-wrapper-tooltip-visible {
+  /* used in test-utils*/
```
jperals, line 110: To be precise, we are using these classes in our component tests but not in test utils (i.e, we are not providing test util wrappers for these elements)

src/app-layout/visual-refresh-toolbar/toolbar/trigger-button/index.tsx
```diff
@@ -44,11 +89,103 @@ function TriggerButton(
...
+  };
+
+  const handleOnFocus = useCallback(() => {
+    if (!(isForPreviousActiveDrawer && hideTooltipOnFocus)) {
+      setShowTooltip(true);
+    }
+  }, [hideTooltipOnFocus, isForPreviousActiveDrawer]);
+
+  const tooltipVisible = useMemo(() => {
+    // console.log({hasTooltip, showTooltip, containerRef, tooltipValue, isMobile, hasOpenDrawer, final: hasTooltip && showTooltip && !!containerRef?.current && tooltipValue && !(isMobile && hasOpenDrawer)})
```
jperals, line 137: To be removed :)

src/app-layout/visual-refresh-toolbar/toolbar/trigger-button/index.tsx
```diff
@@ -44,11 +89,103 @@ function TriggerButton(
...
+  };
+
+  const onShowTooltipHard = (show: boolean) => {
+    if (!showTooltipOverride) {
+      setShowTooltip(show);
+    }
+  };
+
+  const handleOnClick = (event: React.MouseEvent<HTMLButtonElement>) => {
+    console.log('handle on click');
```
jperals, line 119: To be removed

src/app-layout/visual-refresh/trigger-button.tsx
```diff
@@ -103,6 +227,14 @@ function TriggerButton(
           {badge && <div className={styles.dot} />}
         </>
       )}
+      {tooltipVisible && (
+        <Tooltip
+          trackRef={containerRef}
+          value={tooltipValue}
+          className={styles['trigger-tooltip']}
+          {...(!isMobile && { position: 'left' })}
```
jperals, line 235: Is `position: left` necessary? The tooltip uses the popover internally, which manages its position automatically according to the space available.

src/app-layout/__integ__/app-layout-toolbar-tooltips.test.ts
```diff
@@ -0,0 +1,244 @@
...
+  async getElementCenter(selector: string) {
+    const targetRect = await this.getBoundingBox(selector);
+    const x = Math.round(targetRect.left + targetRect.width / 2);
+    const y = Math.round(targetRect.top + targetRect.height / 2);
+    return { x, y };
+  }
+
+  /**
+   * This holds the pointer down for an longer than the buttonDownOnElement
+   * @param selector
```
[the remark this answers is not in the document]
dpitcock: it explains why the function is needed compared to a the button down. There is a grammar error in there I will fix, but if someone later visits the code and wonders why, this will explain it

src/app-layout/__integ__/app-layout-toolbar-tooltips.test.ts
```diff
@@ -0,0 +1,244 @@
...
+class AppLayoutDrawersPage extends BasePageObject {
+  async getElementCenter(selector: string) {
+    const targetRect = await this.getBoundingBox(selector);
+    const x = Math.round(targetRect.left + targetRect.width / 2);
+    const y = Math.round(targetRect.top + targetRect.height / 2);
+    return { x, y };
+  }
+
+  /**
+   * This holds the pointer down for an longer than the buttonDownOnElement
```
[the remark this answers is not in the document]
dpitcock: a simple press opens the drawer, but holding it seems to stop the propagation but opens the tooltip. It must be a browser feature

src/app-layout/visual-refresh-toolbar/toolbar/trigger-button/index.tsx
dpitcock: i will add an eslint rule for these in a separate PR

src/app-layout/visual-refresh/trigger-button.tsx
dpitcock: yes. for UI consistency so all tooltips show in a consistent way

src/app-layout/__integ__/app-layout-focus-delegation.test.ts
```diff
@@ -138,7 +138,10 @@ describe.each(['classic', 'visual-refresh', 'visual-refresh-toolbar'] as const)(
         'focuses split panel preferences button when its position changes from side to bottom',
         setupTest(
           async page => {
+            await page.isExisting(wrapper.findSplitPanel().findOpenButton().toSelector());
```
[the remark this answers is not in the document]
dpitcock: Removed

src/app-layout/visual-refresh-toolbar/toolbar/trigger-button/index.tsx
dpitcock: removed

src/app-layout/visual-refresh/trigger-button.tsx
```diff
@@ -46,13 +64,119 @@ function TriggerButton(
...
+  const { hasOpenDrawer, isMobile } = useAppLayoutInternals();
+
+  const tooltipVisible =
+    containerRef && containerRef?.current && tooltipValue && !(isMobile && hasOpenDrawer) && showTooltip;
+
+  const onShowTooltipSoft = (show: boolean) => {
+    setShowTooltip(show);
+  };
+
+  const onShowTooltipHard = (show: boolean) => {
```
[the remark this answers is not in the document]
dpitcock: removed on this trigger button but left on the other

src/app-layout/__integ__/app-layout-toolbar-tooltips.test.ts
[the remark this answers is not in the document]
dpitcock: I was able to get buttonUp to work

src/app-layout/__integ__/app-layout-toolbar-tooltips.test.ts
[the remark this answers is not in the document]
dpitcock: no longer needed

src/app-layout/visual-refresh-toolbar/toolbar/trigger-button/index.tsx
```diff
@@ -28,6 +29,50 @@ export interface TriggerButtonProps {
...
+   */
+  tooltipText?: string;
+  /**
+   * If the AppLayout has a drawer that is open
+   * Used to hide tooltips in certian conditions
+   */
+  /**
+   * set to true if the trigger button was used to open the last active drawer
+   */
+  isForPreviousActiveDrawer?: boolean;
```
jperals, line 51: Just to make sure I understand, this is to prevent the tooltip from showing up again when it is focused after closing the drawer?

Could we add an explanation in the comment?

src/app-layout/__integ__/app-layout-toolbar-tooltips.test.ts
```diff
@@ -0,0 +1,236 @@
...
+    VISIBLE_MOBILE_TOOLBAR_TRIGGERS_LIMIT + (theme === 'visual-refresh-toolbar' ? 1 : 0)
+  );
+
+  const appliedThemeStyles = theme === 'visual-refresh' ? visualRefreshStyles : toolbarStyles;
+
+  describe.each(['desktop', 'mobile'] as const)('%s', size => {
+    const drawersTriggerContainerClassKey = `drawers-${size === 'desktop' ? 'desktop' : 'mobile'}-triggers-container`;
+    const drawerIdsToTest = size === 'mobile' ? mobileDrawerTriggerIds : toolbarDrawerIds;
+
+    testIf(size === 'desktop')(
```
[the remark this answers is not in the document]
dpitcock: Mainly because the tests are failing in mobile in dry-run but not on my local

src/app-layout/visual-refresh-toolbar/toolbar/trigger-button/index.tsx
dpitcock: Yes. Will do. this part of the logic will need to be revisited as it does not work when focus and hover events are combined via a bug bash finding.

src/__tests__/__snapshots__/documenter.test.ts.snap
```diff
@@ -729,7 +729,7 @@ Each Drawer is an item in the drawers wrapper with the following properties:
 - \`iconSvg\` (React.ReactNode) - (Optional) Specifies the SVG of a custom icon. For more information, see [SVG icon guidelines](/components/icon/?tabId=api#slots)
 
 #### DrawerAriaLabels
-- \`drawerName\` (string) - Label for the drawer itself.
+- \`drawerName\` (string) - Label for the drawer itself, and for the drawer trigger button tooltip text.
```
jperals, line 732: I think "label" so far meant "ARIA label", but now the newly added part reads like "label for the drawer button tooltip text", which makes the meaning of "label" no more clear. To make it clear that in the case of the drawer trigger it is used as an ARIA label, and in the case of the tooltip it is used for the text, what about something like

> Label for the drawer itself. Also used for the trigger button tooltip text.

dpitcock: I'll follow up on this

src/app-layout/visual-refresh-toolbar/toolbar/trigger-button/index.tsx
```diff
@@ -44,11 +75,145 @@ function TriggerButton(
...
+      setShowTooltip(true);
+    },
+    [
+      // To assert reference equality check
+      isForPreviousActiveDrawer,
+      isForSplitPanel,
+    ]
+  );
+
+  const tooltipVisible = useMemo(() => {
```
jperals, line 145: Minor: I don't think `useMemo` really has any value here. This is not an expensive calculation and this might be adding more computation (and complexity) than it saves.

src/app-layout/visual-refresh-toolbar/toolbar/trigger-button/index.tsx
```diff
@@ -44,11 +75,145 @@ function TriggerButton(
...
+        if (breadcrumbsComponent && relatedTarget && breadcrumbsComponent.contains(relatedTarget)) {
+          //show tooltip when using keys to navigate to it from the breadcrumbs
+          shouldShowTooltip = true;
+        } else {
+          shouldShowTooltip = isFromAnotherTrigger;
+        }
+      } else if (!isForPreviousActiveDrawer) {
+        shouldShowTooltip = true; //for keyed navigation inside the toolbar
+      } else if (isFromAnotherTrigger) {
+        //needed for safari which doesn't read the relatedTarget when drawer closed via
```
jperals, line 131: If Safari doesn't read the relatedTarget, how do we enter this condition if `isFromAnotherTrigger` is only true when `relatedTarget?.dataset?.shiftFocus === 'awsui-layout-drawer-trigger'`?

src/app-layout/visual-refresh-toolbar/toolbar/trigger-button/index.tsx
jperals: Oh I see, it's so that you can use it as a dependency in a hook further down (line 202)?

src/app-layout/visual-refresh-toolbar/toolbar/trigger-button/index.tsx
```diff
@@ -44,11 +75,145 @@ function TriggerButton(
...
+    hasOpenDrawer = false,
+    isMobile = false,
+    isForPreviousActiveDrawer = false,
+    isForSplitPanel = false,
   }: TriggerButtonProps,
   ref: React.Ref<ButtonProps.Ref>
 ) {
+  const containerRef = useRef<HTMLDivElement | null>(null);
+  const tooltipValue = tooltipText ? tooltipText : ariaLabel ? ariaLabel : '';
+  const [showTooltip, setShowTooltip] = useState<boolean>(false);
```
jperals, line 89: What about calling this one something like `[isFocused, setIsFocused]`?

src/app-layout/visual-refresh-toolbar/toolbar/trigger-button/index.tsx
```diff
@@ -44,11 +75,145 @@ function TriggerButton(
...
+  const [suppressTooltip, setSupressTooltip] = useState<boolean>(false);
+
+  const handleTriggerClick = (event: React.MouseEvent<HTMLButtonElement>) => {
+    event.stopPropagation(); // Stop the event from propagating to the badge icon
```
jperals, line 93: What is the "badge" icon? If this refers to the icon inside the trigger, it is inside the trigger and not the other way around, so what is the concern?

src/app-layout/visual-refresh/drawers.tsx
```diff
@@ -122,7 +122,11 @@ function ActiveDrawer() {
         }
       }}
     >
-      {!isMobile && activeDrawer?.resizable && <div className={styles['drawer-slider']}>{resizeHandle}</div>}
+      {!isMobile && activeDrawer?.resizable && (
+        <div className={styles['drawer-slider']} data-shift-focus="drawer-resize-button-handle">
```
jperals, line 126: This `data-shift-focus` attribute is not used?

src/app-layout/visual-refresh/drawers.tsx
```diff
@@ -139,6 +143,7 @@ function ActiveDrawer() {
             }}
             ref={drawersRefs.close}
             variant="icon"
+            data-shift-focus="last-opened-toolbar-trigger-button"
```
jperals, line 146: This `data-shift-focus` attribute is not used?

src/app-layout/visual-refresh-toolbar/toolbar/trigger-button/index.tsx
dpitcock: tooltip shows when the button is not focused, such as when hovered

src/app-layout/visual-refresh-toolbar/toolbar/trigger-button/index.tsx
```diff
@@ -44,11 +75,145 @@ function TriggerButton(
...
+  const handleOnFocus = useCallback(
+    (event: FocusEvent) => {
+      const eventWithRelatedTarget = event as any;
+      const relatedTarget = eventWithRelatedTarget?.relatedTarget;
+      const isFromAnotherTrigger = relatedTarget?.dataset?.shiftFocus === 'awsui-layout-drawer-trigger';
+      let shouldShowTooltip = false;
+      if (isForSplitPanel) {
+        const breadcrumbsComponent = document.querySelector('nav[data-awsui-discovered-breadcrumbs="true"]');
+        if (breadcrumbsComponent && relatedTarget && breadcrumbsComponent.contains(relatedTarget)) {
+          //show tooltip when using keys to navigate to it from the breadcrumbs
```
jperals, line 123: Why is there a need to be so specific to the breadcrumbs? If the intention is to show the tooltip on focus except right after closing the split panel, can't we do something similar to what we do with the drawer triggers with `isForPreviousActiveDrawer`?

src/app-layout/visual-refresh-toolbar/toolbar/trigger-button/index.tsx
dpitcock: without this, somehow the relatedTarget is seen as the badge icon, that little red dot indicating something and that throws off the focus event handling causing the tooltip to flash when not expected to

src/app-layout/visual-refresh-toolbar/toolbar/trigger-button/index.tsx
dpitcock: I can do a follow up where i add a data-shift-focus to the breadcrumbs but I'd rather not touch anther part of this tooltip project.

src/app-layout/__integ__/app-layout-toolbar-tooltips.test.ts
```diff
@@ -0,0 +1,361 @@
...
+        await page.keys(['Tab']);
+        await expect(page.isFocused(secondDrawerTriggerSelector)).resolves.toBeTruthy();
+        await expect(page.getElementsCount(`.${tooltipStyles.root}`)).resolves.toBe(1);
+        await page.keys(['Shift', 'Tab', 'Null']);
+        await expect(page.isFocused(firstDrawerTriggerSelector)).resolves.toBeTruthy();
+        await expect(page.getElementsCount(`.${tooltipStyles.root}`)).resolves.toBe(1);
+      })
+    );
+
+    testIf(theme === 'visual-refresh-toolbar')(
```
jperals, line 277: Minor: by the `describe.each` above, `theme` can only be `visual-refresh-toolbar`. Same for the next `testIf` blocks.

dpitcock: there is a followup to add visual-refresh to this as it was in previous commits.

dpitcock: removed the code and added a comment for when visual refresh is returned

src/app-layout/visual-refresh-toolbar/toolbar/trigger-button/index.tsx
jperals: My concern is that this is a bit too specific, and might lead to cases where the tooltip should show and it doesn't. See, for example: if I click on the last breadcrumb, which does not have a link, the focus is still on the body but by pressing Tab the focus will move to the split panel trigger but the tooltip doesn't show up. This is because technically (i.e, according to this logic) the focus did not come from the breadcrumbs.

dpitcock: thanks for the comment. Removing the check works

src/app-layout/visual-refresh-toolbar/toolbar/trigger-button/index.tsx
```diff
@@ -44,11 +75,137 @@ function TriggerButton(
...
+  };
+
+  /**
+   * Takes the drawer being closed and the data-shift-focus value from a close button on that drawer that persists
+   * on the event relatedTarget to determine not to show the tooltip
+   * @param event
+   */
+  const handleOnFocus = useCallback(
+    (event: FocusEvent) => {
+      let shouldShowTooltip = false;
```
jperals, line 116: Minor: no need to declare a `let`, can be declared as `const` as it is assigned only once

src/app-layout/visual-refresh-toolbar/toolbar/trigger-button/index.tsx
```diff
@@ -44,11 +75,137 @@ function TriggerButton(
...
+      const eventWithRelatedTarget = event as any;
+      const relatedTarget = eventWithRelatedTarget?.relatedTarget;
+      const isFromAnotherTrigger = relatedTarget?.dataset?.shiftFocus === 'awsui-layout-drawer-trigger';
+      if (
+        isForSplitPanel || //for tab/key navigation to button from breadcrume
```
jperals, line 121: Nitpick: typo in comment (`breadcrume`)

jperals: Is it ever possible that `isForSplitPanel` and `isForPreviousActiveDrawer` are true at the same time? If not, there is no need to check for `isForSplitPanel` in this condition.

dpitcock: yes. when the split panel is open that both would be true

src/app-layout/visual-refresh-toolbar/toolbar/trigger-button/index.tsx
dpitcock: It gets muted though in the conditional below it

jperals: But it's the only place where it is assigned, so it could be declared just there. Anyway not blocking.

jperals (approved):

PR #2652, opened 2024-09-03
fix: Improve progress bar color contrast in warning flash message

avinashbot:
We had to revert this PR because Firefox didn't like it. Now fixed (see second commit) and tested in my dev pipeline, so we're good to go.

+ Reverts cloudscape-design/components#2518
+ Refactors the CSS to make it work with Firefox (split the -moz and -webkit rules into separate CSS declarations)

### DIFF
--- a/src/progress-bar/styles.scss
+++ b/src/progress-bar/styles.scss
@@ -37,7 +37,7 @@
 
 .label {
   &-flash {
-    color: 'inherit';
+    color: inherit;
     font-weight: styles.$font-weight-bold;
   }
   &-key-value {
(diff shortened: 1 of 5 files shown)


src/progress-bar/styles.scss
```diff
@@ -94,14 +94,11 @@ $background-color-in-flash: awsui.$color-background-progress-bar-layout-in-flash
 $bar-color: awsui.$color-background-progress-bar-content-default;
 $bar-color-in-flash: awsui.$color-background-progress-bar-content-in-flash;
 
-// Current version of Edge has a known bug with CSS variables used in pseudo elements.
```
jperals, line 97: Do we know that this is not necessary anymore?

avinashbot: "Current version of Edge" here is referring to EdgeHTML, before Edge switched to using Chromium as its base.

jperals (approved):

PR #2654, opened 2024-09-04
feat: Add analytics metadata to internal Selectable item component

fralongo:
### Description

Needed to introduce analytics metadata for Select, Multiselect, Autosuggest, Property filter.
Note that it depends on https://github.com/cloudscape-design/component-toolkit/pull/94

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/__tests__/analytics-metadata-test-utils.ts
+++ b/src/internal/__tests__/analytics-metadata-test-utils.ts
@@ -25,7 +25,8 @@ const validateLabels = (labelSelectors: Array<string>, labelsClassNames: Record<
     const parts = current
       .split(/\s+/)
       .filter(part => part.startsWith('.'))
-      .map(part => part.replace('.', ''));
+      .map(part => part.replace('.', ''))
+      .map(part => part.replace(/\[.*\]/, ''));
     acc = [...acc, ...parts];
     return acc;
   }, []);
--- a/src/internal/components/selectable-item/index.tsx
+++ b/src/internal/components/selectable-item/index.tsx
@@ -3,29 +3,16 @@
 import React, { useLayoutEffect, useRef } from 'react';
 import clsx from 'clsx';
 
-import { BaseComponentProps, getBaseProps } from '../../base-component';
-import { HighlightType } from '../options-list/utils/use-highlight-option.js';
+import { getAnalyticsMetadataAttribute } from '@cloudscape-design/component-toolkit/internal/analytics-metadata';
 
+import { getBaseProps } from '../../base-component';
+import { getAnalyticsSelectActionMetadata } from './analytics-metadata/utils';
+import { SelectableItemProps } from './interfaces';
+
+import analyticsSelectors from './analytics-metadata/styles.css.js';
 import styles from './styles.css.js';
 
-export type SelectableItemProps = BaseComponentProps & {
-  children: React.ReactNode;
-  selected?: boolean;
-  highlighted?: boolean;
-  disabled?: boolean;
-  hasBackground?: boolean;
-  isParent?: boolean;
-  isChild?: boolean;
-  virtualPosition?: number;
-  padBottom?: boolean;
-  isNextSelected?: boolean;
-  useInteractiveGroups?: boolean;
-  screenReaderContent?: string;
-  ariaPosinset?: number;
-  ariaSetsize?: number;
-  highlightType?: HighlightType['type'];
-  ariaDescribedby?: string;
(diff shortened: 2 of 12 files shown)


src/internal/__tests__/analytics-metadata-test-utils.ts
```diff
@@ -25,7 +25,8 @@ const validateLabels = (labelSelectors: Array<string>, labelsClassNames: Record<
     const parts = current
       .split(/\s+/)
       .filter(part => part.startsWith('.'))
-      .map(part => part.replace('.', ''));
+      .map(part => part.replace('.', ''))
+      .map(part => part.replace(/\[.*\]/, ''));
```
amrtaher1234, line 29: why did we add this regex?

src/internal/components/selectable-item/index.tsx
```diff
@@ -3,11 +3,63 @@
...
 import styles from './styles.css.js';
 
+export interface ItemDataAttributes {
+  'data-group-index'?: string;
+  'data-child-index'?: string;
+  'data-in-group-index'?: string;
+  'data-test-index'?: string;
+}
+
+const getAnayticsMetadata = ({
```
amrtaher1234, line 23: typo in the naming, should be "getAnalyticsMetadata" instead

src/internal/components/selectable-item/index.tsx
```diff
@@ -3,11 +3,63 @@
...
+
 import { BaseComponentProps, getBaseProps } from '../../base-component';
 import { HighlightType } from '../options-list/utils/use-highlight-option.js';
+import { GeneratedAnalyticsMetadataSelectableItemSelect } from './analytics-metadata/interfaces';
 
+import optionAnalyticsSelectors from './../option/analytics-metadata/styles.css.js';
+import analyticsSelectors from './analytics-metadata/styles.css.js';
 import styles from './styles.css.js';
 
+export interface ItemDataAttributes {
```
amrtaher1234, line 16: For this part of the code would it make since to move it in its own file (like a utility file or so) ?

Reasoning behind this: As an open-source contributor I will get confused by trying to read and understand the analytics logic and how it flows.

src/internal/components/selectable-item/index.tsx
```diff
@@ -3,11 +3,63 @@
...
+  'data-group-index'?: string;
+  'data-child-index'?: string;
+  'data-in-group-index'?: string;
+  'data-test-index'?: string;
```
amrtaher1234, line 20: These data attributes only exists inside the  autosuggest [option](https://github.com/cloudscape-design/components/blob/09aa16f2933bcd5cf004dad5f5c82cedb5a93251/src/autosuggest/autosuggest-option.tsx#L83-L85) `<SelectableItem/>` component, the other references does not have any data attributes, is this expected?
- [Item](https://github.com/cloudscape-design/components/blob/09aa16f2933bcd5cf004dad5f5c82cedb5a93251/src/select/parts/item.tsx#L65)
- [Multiselect](https://github.com/cloudscape-design/components/blob/09aa16f2933bcd5cf004dad5f5c82cedb5a93251/src/select/parts/multiselect-item.tsx#L56)

src/internal/__tests__/analytics-metadata-test-utils.ts
fralongo: This test ensures that all classNames used in these objects come from specific files that we will not change for stability.

In this component, the selectors for the group label contain a `.className[data-group-index="n"] .another-className`. We need to remove all attribute-based CSS selectors to be able to run this test.

src/internal/components/selectable-item/index.tsx
fralongo: Addressed

src/internal/components/selectable-item/index.tsx
fralongo: Great idea, I have refactored the code

src/internal/components/selectable-item/index.tsx
fralongo: They appear in many places:
- https://github.com/search?q=repo%3Acloudscape-design%2Fcomponents%20data-in-group-index&type=code
- https://github.com/search?q=repo%3Acloudscape-design%2Fcomponents%20data-group-index&type=code
- https://github.com/search?q=repo%3Acloudscape-design%2Fcomponents%20data-child-index&type=code

It is actually inconsistent, because sometimes we use `data-child-index` and `data-in-group-index` for the same purpose. However, I didn't want to refactor those attributes because we use them in test utilities, so there might be broken builds.

taheramr (approved):

PR #2666, opened 2024-09-05
chore: Fixes flaky app layout tests that depend on timeout

pan-kot:
### Description

Use waitFor instead of delay for global-breadcrumbs.test.tsx tests.

See failing test: https://github.com/cloudscape-design/components/actions/runs/10718898212/job/29722075662

### How has this been tested?

* Confirmed flakiness with multiple test run
* Verified not a single test in the fail fails after 1500 test runs

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/__tests__/global-breadcrumbs.test.tsx
+++ b/src/app-layout/__tests__/global-breadcrumbs.test.tsx
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 /* eslint simple-import-sort/imports: 0 */
 import React, { useState } from 'react';
-import { render, cleanup, waitFor } from '@testing-library/react';
+import { render, cleanup, waitFor, act } from '@testing-library/react';
 import { describeEachAppLayout } from './utils';
 import createWrapper, { BreadcrumbGroupWrapper } from '../../../lib/components/test-utils/dom';
 import AppLayout from '../../../lib/components/app-layout';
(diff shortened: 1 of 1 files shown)


src/app-layout/__tests__/global-breadcrumbs.test.tsx
```diff
@@ -60,67 +50,86 @@ afterEach(() => {
 
 describeEachAppLayout({ themes: ['refresh-toolbar'], sizes: ['desktop'] }, () => {
   test('renders normal breadcrumbs when no app layout is present', async () => {
-    await renderAsync(<BreadcrumbGroup items={defaultBreadcrumbs} />);
-    expect(findAllBreadcrumbsInstances()).toHaveLength(1);
-    expect(wrapper.findBreadcrumbGroup()!.findBreadcrumbLinks()).toHaveLength(2);
+    render(<BreadcrumbGroup items={defaultBreadcrumbs} />);
+    await waitFor(() => {
```
just-boris, line 54: This change breaks the intention of the tests. `waitFor` runs the first check without any delay and when it passes it moves forward.

The point of this test was to assert there is still only one breadcrumb instance in a long term, even after a timeout.


We can talk about individual flaky tests and how to make them more stable, but applying waitFor for every test in this file is incorrect.

pan-kot: Makes total sense!

I added a delay before each wait for to that the first check is delayed as it was before the change.

just-boris: Now the setup is even more far from original.

How about keep everything unchanged, and only add a couple `waitFor` to potentially flaky tests?

pan-kot: The first one was different. This one is not. I am replacing this:
```
render()
delay()
assert()
```

with this:
```
render()
delay()
assertUntilItWorks()
```

The latter is equivalent of having a larger delay, but better efficient. I also run it 1500 times.

The change can be applied not to all tests but only to some. However, it looks like they all can fail. I reproduced failures in 2 or 3 different tests in that file.

just-boris: I still do not understand what you disliked so much about `renderAsync`, so you decided to remove it.

I think the correct fix would be to improve `renderAsync` to make it more stable. It will happen here: https://github.com/cloudscape-design/components/pull/2676

I will close this PR as not needed. Thanks for taking your time

PR #2668, opened 2024-09-05
feat: Add analytics metadata to Select component

fralongo:
### Description

Similar to https://github.com/cloudscape-design/components/commit/dde24620bceb97c2fd3e59aa45b0e95af8402313

#### Correctness

- _Changes include appropriate documentation updates._
- _Changes are backward-compatible if not indicated, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#public-apis)._
- _Changes do not include unsupported browser features, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#browsers-support)._
- _Changes were manually tested for accessibility, see [accessibility guidelines](https://cloudscape.design/foundation/core-principles/accessibility/)._

#### Security

- _If the code handles URLs: all URLs are validated through [the `checkSafeUrl` function](https://github.com/cloudscape-design/components/blob/main/src/internal/utils/check-safe-url.ts)._

#### Testing

- _Changes are covered with new/existing unit tests?_
- _Changes are covered with new/existing integration tests?_
</details>

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2667, opened 2024-09-05
feat: Add analytics metadata to Autosuggest component

fralongo:
### Description

Similar to https://github.com/cloudscape-design/components/commit/dde24620bceb97c2fd3e59aa45b0e95af8402313

<details>
   <summary>Review checklist</summary>

_The following items are to be evaluated by the author(s) and the reviewer(s)._

#### Correctness

- _Changes include appropriate documentation updates._
- _Changes are backward-compatible if not indicated, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#public-apis)._
- _Changes do not include unsupported browser features, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#browsers-support)._
- _Changes were manually tested for accessibility, see [accessibility guidelines](https://cloudscape.design/foundation/core-principles/accessibility/)._

#### Security

- _If the code handles URLs: all URLs are validated through [the `checkSafeUrl` function](https://github.com/cloudscape-design/components/blob/main/src/internal/utils/check-safe-url.ts)._

#### Testing

- _Changes are covered with new/existing unit tests?_
- _Changes are covered with new/existing integration tests?_
</details>

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2670, opened 2024-09-05
feat: Add analytics metadata to Property filter

fralongo:
### Description

Similar to https://github.com/cloudscape-design/components/pull/2669
Depends on https://github.com/cloudscape-design/components/pull/2667, https://github.com/cloudscape-design/components/pull/2668, https://github.com/cloudscape-design/components/pull/2669

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/popover/internal.tsx
+++ b/src/popover/internal.tsx
@@ -27,6 +27,7 @@ export interface InternalPopoverProps extends Omit<PopoverProps, 'triggerType' |
   __onOpen?: NonCancelableEventHandler<null>;
   triggerType?: PopoverProps.TriggerType | 'filtering-token';
   size: PopoverProps.Size | 'content';
+  __closeAnalyticsAction?: string;
 }
 
 export interface InternalPopoverRef {
@@ -54,6 +55,7 @@ function InternalPopover(
 
     __onOpen,
     __internalRootRef = null,
+    __closeAnalyticsAction,
     ...restProps
   }: InternalPopoverProps,
   ref: React.Ref<InternalPopoverRef>
@@ -166,6 +168,7 @@ function InternalPopover(
               header={header}
               onDismiss={onDismiss}
               overflowVisible="both"
+              closeAnalyticsAction={__closeAnalyticsAction}
             >
               {content}
             </PopoverBody>
--- a/src/property-filter/filtering-token/index.tsx
+++ b/src/property-filter/filtering-token/index.tsx
@@ -15,6 +15,7 @@ import { useListFocusController } from '../../internal/hooks/use-list-focus-cont
 import { useMergeRefs } from '../../internal/hooks/use-merge-refs';
 import InternalPopover, { InternalPopoverProps, InternalPopoverRef } from '../../popover/internal';
 import InternalSelect from '../../select/internal';
+import { GeneratedAnalyticsMetadataPropertyEditStart } from '../analytics-metadata/interfaces';
 
 import testUtilStyles from '../test-classes/styles.css.js';
 import styles from './styles.css.js';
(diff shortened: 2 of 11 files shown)


src/popover/internal.tsx
```diff
@@ -27,6 +27,7 @@ export interface InternalPopoverProps extends Omit<PopoverProps, 'triggerType' |
   __onOpen?: NonCancelableEventHandler<null>;
   triggerType?: PopoverProps.TriggerType | 'filtering-token';
   size: PopoverProps.Size | 'content';
+  __closeAnalyticsAction?: string;
```
pan-kot, line 30: Shouldn't popover emit its own metric on dismiss, why is it coming from the property filter?

src/property-filter/filtering-token/index.tsx
```diff
@@ -315,6 +324,7 @@ function TokenDismissButton({
       aria-label={ariaLabel}
       onClick={onClick}
       disabled={disabled}
+      {...getAnalyticsMetadataAttribute({ action: 'dismiss' })}
```
pan-kot, line 327: can this action be more specific, like: "tokenRemove"?

src/property-filter/token.tsx
```diff
@@ -92,7 +97,12 @@ export const TokenButton = ({
         const formattedToken = i18nStrings.formatToken(token);
         return {
           content: (
-            <span className={styles['token-trigger']}>
+            <span
+              className={clsx(styles['token-trigger'], analyticsSelectors['token-trigger'])}
+              {...getAnalyticsMetadataAttribute({
+                action: 'editStart',
```
pan-kot, line 103: We already support grouped tokens internally. For grouped tokens, the edit operation originates from a dedicated button. I think this action should be defined in the filtering-token/index.tsx

src/property-filter/token-editor.tsx
```diff
@@ -199,21 +205,33 @@ export function TokenEditor({
       )}
 
       <div className={styles['token-editor-actions']}>
-        <InternalButton
-          formAction="none"
-          variant="link"
-          className={clsx(styles['token-editor-cancel'], testUtilStyles['token-editor-cancel'])}
-          onClick={onDismiss}
+        <span
+          {...getAnalyticsMetadataAttribute({
```
pan-kot, line 209: Can we pass this as an internal prop to the button, similar to how it is done for popover?

src/popover/internal.tsx
fralongo: Internal components do not emit their own metrics because consumers do not explicitly add the internal component, so the metric would generate noise.

src/property-filter/filtering-token/index.tsx
fralongo: I prefer to keep it aligned with Multiselect and Tokengroup

src/property-filter/token.tsx
fralongo: Since grouped tokens are not in prod today, can I move that logic at a later stage?

src/property-filter/token-editor.tsx
fralongo: I prefer as mush as possible to avoid adding new properties to components, if there is another solution. Is there any risk in adding this span?

fralongo: The exception is rather the Popover. I didn't manage to avoid it there because wrapping the whole popover in another element with the metadata would lead to overriding the `action` of every single event coming from the popover, instead of the close action only.

src/property-filter/token.tsx
fralongo: Implemented the change. I still think it is not very convenient, because it requires the addition of an unnecessary span, but I am fine with that

pan-kot (approved):

PR #2673, opened 2024-09-05
feat: itemsControl property of Side Navigation

Manzurka:
### Description

Items Control property allows customers to define the component of their chose to be added to a slot below the header and above the items of Side Navigation Panel. For example, customers can pass Select or Multi-select components with custom attribute values (ex. options, selectedOption, triggerVariant, label, loadingText). Use case: This allows customers to define a callback function onChange event of the component, for example, to allow users filter out navigation menu items.

Related links
AWSUI-54774
API Proposal ID bG8cAsCSosVm

### How has this been tested?

Unit tests, snapshot test, integ test have been added.

<details>

#### Correctness

#### Security

- _If the code handles URLs: all URLs are validated through [the `checkSafeUrl` function](https://github.com/cloudscape-design/components/blob/main/src/internal/utils/check-safe-url.ts)._ n/a

#### Testing

- _Changes are covered with new/existing unit tests?_ yes
- _Changes are covered with new/existing integration tests?_ yes
</details>

---


src/side-navigation/__integ__/side-navigation.test.ts
```diff
@@ -39,4 +39,16 @@ describe('SideNavigation', () => {
       })
     );
   });
+
+  describe('Items Header', () => {
+    it(
```
pan-kot, line 44: This test is supposed to fail because the `#/light/side-navigation/integration` page does not, in fact, have the items header. It fails because of incorrectly written assertion on the line 50: the page.isExisting is asynchronous.

I propose to remove this test though because the logic is covered with unit tests. We usually create integration test for behaviours depending on browser in a way that cannot be well replicated in unit tests.

Nevertheless, it would be useful to add use the new property in one of the side navigation test pages just to showcase it and allow for quick manual checks.

src/side-navigation/__tests__/side-navigation.test.tsx
```diff
@@ -3,6 +3,8 @@
 import * as React from 'react';
 import { render } from '@testing-library/react';
 
+import Multiselect from '../../../lib/components/multiselect';
+import Select from '../../../lib/components/select';
```
pan-kot, line 7: The new property is just a slot - there is no need to test if it can render different components. The test can use a fake component or maybe just select so that we can ensure the `wrapper.findItemsHeader()!.findSelect()` chaining works.

src/side-navigation/__tests__/side-navigation.test.tsx
```diff
@@ -1219,6 +1221,88 @@ describe('SideNavigation', () => {
       expect(logo).toHaveAttribute('alt', 'logo');
     });
   });
+
+  describe('Items Header', () => {
+    it('renders select component as items header', () => {
+      const options = [
```
pan-kot, line 1227: I propose to keep it simple and render an empty select component. In the test it is enough to assert the select component can be found when it is passed as itemsHeader.

You can also add a second small test to ensure the findItemsHeader() returns null when it is not provided.

pan-kot (approved):
Great work!

Manzurka:
Screenshot 2024-09-11 at 11 36 20 AM

pan-kot (approved):

PR #2660, opened 2024-09-05
refactor: Update integration tests to resolve chromium upgrade 2

pan-kot:
### Description

Related to https://github.com/cloudscape-design/browser-test-tools/pull/105 (see dry-run test failures).

Switching to Chromium --headless=new causes failures to our tests that depend on screen size. The refactoring fixes the issue for some of the app layout tests to either remove the dependency completely or make assertions depend on the measured viewport height.

Previous PR: https://github.com/cloudscape-design/components/pull/2657

### How has this been tested?

* Verified the tests work with old and new headless mode

<details>
   <summary>Review checklist</summary>

_The following items are to be evaluated by the author(s) and the reviewer(s)._

#### Correctness

- _Changes include appropriate documentation updates._
- _Changes are backward-compatible if not indicated, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#public-apis)._
- _Changes do not include unsupported browser features, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#browsers-support)._
- _Changes were manually tested for accessibility, see [accessibility guidelines](https://cloudscape.design/foundation/core-principles/accessibility/)._

#### Security

- _If the code handles URLs: all URLs are validated through [the `checkSafeUrl` function](https://github.com/cloudscape-design/components/blob/main/src/internal/utils/check-safe-url.ts)._

#### Testing

- _Changes are covered with new/existing unit tests?_
- _Changes are covered with new/existing integration tests?_
</details>

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2679, opened 2024-09-06
refactor: Update integration tests to resolve chromium upgrade 3

pan-kot:
### Description

Related to https://github.com/cloudscape-design/browser-test-tools/pull/105 (see dry-run test failures).

Switching to Chromium --headless=new causes failures to our tests that depend on screen size. The refactoring fixes the issue for some of the app layout tests to either remove the dependency completely or make assertions depend on the measured viewport height.

Previous PR: https://github.com/cloudscape-design/components/pull/2660

### How has this been tested?

* Verified the tests work with old and new headless mode

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/__integ__/app-layout-split-panel.test.ts
+++ b/src/app-layout/__integ__/app-layout-split-panel.test.ts
@@ -224,11 +224,13 @@ describe.each(['classic', 'visual-refresh', 'visual-refresh-toolbar'] as const)(
       // Simulating 200% zoom on medium screens (1366x768 / 2 ~= 680x360 ).
       await page.setWindowSize({ width: 680, height: 360 });
       await page.openPanel();
-      const { height } = await page.getViewportSize();
-      await page.dragResizerTo({ x: 0, y: height });
+      const { height: screenHeight } = await page.getViewportSize();
+      const headerRect = await page.getBoundingBox('#h');
+      await page.dragResizerTo({ x: 0, y: screenHeight });
       expect((await page.getSplitPanelSize()).height).toEqual(160);
       await page.dragResizerTo({ x: 0, y: 0 });
-      expect(Math.round((await page.getSplitPanelSize()).height)).toEqual(277);
+      const splitPanelMaxSize = screenHeight - 40 - headerRect.height;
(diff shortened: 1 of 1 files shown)


src/app-layout/__integ__/app-layout-split-panel.test.ts
```diff
@@ -224,11 +224,13 @@ describe.each(['classic', 'visual-refresh', 'visual-refresh-toolbar'] as const)(
...
       await page.openPanel();
-      const { height } = await page.getViewportSize();
-      await page.dragResizerTo({ x: 0, y: height });
+      const { height: screenHeight } = await page.getViewportSize();
+      const headerRect = await page.getBoundingBox('#h');
+      await page.dragResizerTo({ x: 0, y: screenHeight });
       expect((await page.getSplitPanelSize()).height).toEqual(160);
       await page.dragResizerTo({ x: 0, y: 0 });
-      expect(Math.round((await page.getSplitPanelSize()).height)).toEqual(277);
+      const splitPanelMaxSize = screenHeight - 40 - headerRect.height;
```
cansuaa, line 232: Why 40?

pan-kot: It is the value of CONSTRAINED_MAIN_PANEL_MIN_HEIGHT, used here: https://github.com/cloudscape-design/components/blob/main/src/app-layout/classic.tsx#L270

cansuaa: Then should we use that variable directly here? or at least add a comment?

cansuaa (approved):

src/app-layout/__integ__/app-layout-split-panel.test.ts
pan-kot: I hesitate to do either, to be honest, as both solutions create a binding between the test and the implementation. I think the fact the assertion is now a formula is good because it makes the intent better clear and we can even define the test against multiple screen heights to ensure the formula holds. The constant 40 is just a part of that. Technically, the implementation can change, e.g. we can move it to the different place in code or compute the constant as a sum of two smaller ones - but that should not matter unless the actual value changes, so it is the actual value that is here important.

PR #2684, opened 2024-09-09
feat: Steps

Revuj:
### Description

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/steps/styles.scss
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    > .step-header {
+      grid-row: 1;
+      grid-column: 2;
+    }
+
+    > .step-details {
+      align-items: center;
+      font-size: awsui.$font-size-body-s;
```
Revuj, line 36: font-size and color should not be set here. I'll remove this

src/test-utils/dom/steps/index.ts
```diff
@@ -0,0 +1,39 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import { ComponentWrapper, ElementWrapper } from '@cloudscape-design/test-utils-core/dom';
+
+import statusIconStyles from '../../../status-indicator/styles.selectors.js';
+import testUtilStyles from '../../../steps/test-classes/styles.selectors.js';
+
+export class StepWrapper extends ComponentWrapper {
```
Revuj, line 8: This is WIP but any feedback regarding what selectors we should have is appreciated

src/test-utils/dom/steps/index.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  /**
+   * Finds the details of a step
+   */
+  findDetails(): ElementWrapper | null {
+    return this.findByClassName(testUtilStyles['step-details']);
+  }
+}
+export default class StepsWrapper extends ComponentWrapper {
+  static rootSelector: string = testUtilStyles.steps;
```
cansuaa, line 31: The same case with my comment above, you don't need the testUtilStyles, here you can use `styles.root` which is usually how it's defined for the other components as well.

src/steps/internal.tsx
```diff
@@ -0,0 +1,52 @@
...
+import styles from './styles.css.js';
+import testUtilStyles from './test-classes/styles.css.js';
+
+type InternalStepsProps = SomeRequired<StepsProps, 'steps'> & InternalBaseComponentProps<HTMLDivElement>;
+
+const InternalStep = ({ status, statusIconAriaLabel, header, details }: StepsProps.Step) => {
+  return (
+    <li className={clsx(styles['step-container'], testUtilStyles['step-container'])}>
+      <div className={clsx(styles['step-status'], testUtilStyles['step-status'])}>
+        <StatusIndicator type={status} iconAriaLabel={statusIconAriaLabel} />
```
cansuaa, line 20: It seems that only icon is being used from status indicator? If we only want to get the icon then why not use the icon component? 
Another question, do we expect the text to be the same color as the status icon? e.g. green when status is success?

package.json
```diff
@@ -117,6 +117,7 @@
     "prettier": "^3.2.5",
     "react": "^16.14.0",
     "react-dom": "^16.14.0",
+    "react-markdown": "^8.0.7",
```
cansuaa, line 120: Do we really need this?

src/test-utils/dom/steps/index.ts
```diff
@@ -0,0 +1,39 @@
...
+import statusIconStyles from '../../../status-indicator/styles.selectors.js';
+import testUtilStyles from '../../../steps/test-classes/styles.selectors.js';
+
+export class StepWrapper extends ComponentWrapper {
+  /**
+   * Finds the status of a step
+   */
+  findStatusIndicator(): ElementWrapper | null {
+    return this.findByClassName(statusIconStyles.icon);
+  }
```
cansuaa, line 14: I'm not sure how helpful this would be given it only is an icon

cansuaa (commented):
As a side note, for type changes to be reflected in documentation don't forget to build the package and then update the documenter snapshot

package.json
Revuj: Not really! I mostly added it because in Q we are gonna be using this so I wanted to validate it works as expected but I can replace it by something else and avoid adding a devDependency here

src/steps/internal.tsx
Revuj: > Another question, do we expect the text to be the same color as the status icon? e.g. green when status is success?

That is really up to the consumer. There are use cases where consumers want to do that and others where they don't

cansuaa: Let's discuss switching it to icon with Kashi as well. If it's an icon, we could even expose the other available icon names and custom icons through url or svg but there could be a specific reason why status indicator was chosen (which could be the point I brought up whether we expect text to be green, red, etc.)

Revuj: >  If we only want to get the icon then why not use the icon component?

I was looking into this and one issue is that the https://refresh.cloudscape.aws.dev/components/icon does not have Loading, so we would need to provide some custom logic for that. Because of that I would propose continuing with using the StatusIndicator. WDYT?

src/test-utils/dom/steps/index.ts
Revuj: Isn't it useful so that consumers can test that they expected status is there?

src/steps/internal.tsx
cansuaa: Hmm true that is lacking but we can easily switch it with a spinner. Though it'd mean some API changes and implementation changes. I'd say you leave it as is right now and then we discuss with Kashi

src/test-utils/dom/steps/index.ts
cansuaa: Not really, this would only return an icon element which is alone not enough to communicate the status, besides it won't really indicate the status unless they depend on the class name which is not recommended. If you check status indicator, it also doesn't have a selector for the icon.

pages/steps/permutations-with-updates.page.tsx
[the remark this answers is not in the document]
Revuj: So I should just leave the regular `permutations.page` right?

src/steps/styles.scss
```diff
@@ -0,0 +1,53 @@
...
+        block-size: auto;
+        min-block-size: awsui.$space-static-xs;
+        position: relative;
+        inset-inline-end: awsui.$space-static-xxxs;
+      }
+    }
+
+    > :last-of-type > .connector {
+      display: none;
+    }
```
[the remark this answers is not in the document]
Revuj: Confirmed that it is not able to find it :)

pages/steps/permutations-with-updates.page.tsx
```diff
@@ -0,0 +1,160 @@
...
+    setStepIndex4(0);
+    if (timeoutRef4.current !== undefined) {
+      clearTimeout(timeoutRef1.current);
+      timeoutRef4.current = undefined;
+    }
+  };
+
+  return (
+    <ScreenshotArea disableAnimations={false}>
+      <article aria-live="polite">
```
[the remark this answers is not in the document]
Revuj: Confirmed :)

pages/steps/styles.scss
jperals, line 1: What is this file for?

Revuj: It contains the styles being used in the permutations page. I can remove it and make it inline css if you'd prefer

src/steps/interfaces.ts
```diff
@@ -0,0 +1,43 @@
...
+import { type StatusIndicatorProps } from '../status-indicator/internal';
+
+export interface StepsProps extends BaseComponentProps {
+  /**
+   * An array of individual steps
+   *
+   * Each step definition has the following properties:
+   *  * `status` (string) - Status of the step corresponding to a status indicator.
+   *  * `statusIconAriaLabel` - (string) - (Optional) Text alternative for the status icon.
+   *  * `header` (react.ReactNode) - Area next to the status indicator of the step.
```
gethinwebster, line 13: just `(ReactNode)`
```suggestion
   *  * `header` (ReactNode) - Area next to the status indicator of the step.
```

gethinwebster: I think the description of `header` could be improved too, maybe work with Kashi to align how it will be referred to in the usage guidelines

src/steps/internal.tsx
```diff
@@ -0,0 +1,54 @@
...
+      </div>
+      <hr className={styles.connector} role="none" />
+      {details && <div className={styles.details}>{details}</div>}
+    </li>
+  );
+};
+
+export const InternalSteps = ({ steps, __internalRootRef, ...props }: InternalStepsProps) => {
+  return (
+    <div
```
gethinwebster, line 31: shouldn't these `aria-*` properties be on the semantic element, i.e. the `ol`?

pages/steps/permutations-utils.tsx
```diff
@@ -0,0 +1,774 @@
...
+    statusIconAriaLabel: 'loading',
+    header: (
+      <span style={{ margin: '0' }}>
+        Update <Link href="#">Lambda-1</Link>
+      </span>
+    ),
+  },
+];
+
+export const stepsPermutationsInteractive = createPermutations<StepsProps>([
```
jperals, line 759: This export (and the steps within it) doesn't seem to be used

src/steps/interfaces.ts
```diff
@@ -0,0 +1,43 @@
...
+   *
+   * Each step definition has the following properties:
+   *  * `status` (string) - Status of the step corresponding to a status indicator.
+   *  * `statusIconAriaLabel` - (string) - (Optional) Text alternative for the status icon.
+   *  * `header` (ReactNode) - Summary corresponding to the step.
+   *  * `details` (ReactNode) - (Optional) Additional information corresponding to the step.
+   */
+  steps: ReadonlyArray<StepsProps.Step>;
+  /**
+   * Provides an `aria-label` to progress steps container.
```
jperals, line 18: I would add "the" before "progress steps container": 

> Provides an `aria-label` to **the** progress steps container.

(as in line 23)

src/steps/internal.tsx
```diff
@@ -0,0 +1,59 @@
...
+export const InternalSteps = ({
+  steps,
+  ariaLabel,
+  ariaLabelledby,
+  ariaDescribedby,
+  __internalRootRef,
+  ...props
+}: InternalStepsProps) => {
+  return (
+    <div {...props} className={clsx(styles.root, props.className)} ref={__internalRootRef}>
```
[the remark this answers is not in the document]
jperals: Although it is deprecated, we still support the `className` prop to allow adding custom class names to the component's outermost HTML element -even for new components. But I think we should use `getBaseProps` ([link](https://github.com/cloudscape-design/components/blob/4b05e5df6472e4cfcae14cd994bee61d65dcaaec/src/internal/base-component/index.ts#L32)) to support `id` and data attributes as well. Check for example the prompt input component: https://github.com/cloudscape-design/components/blob/main/src/prompt-input/internal.tsx#L57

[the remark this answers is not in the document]
cansuaa: Hmm I thought we didn't want to include it in new components. Okay then.

Base props are already added to props in the index file.

pages/steps/permutations-utils.tsx
```diff
@@ -715,7 +698,7 @@ export const failedStepsWithRetryButtonInteractive: ReadonlyArray<StepsProps.Ste
     header: <span style={{ margin: '0' }}>Could not fetch security groups</span>,
     details: (
       <>
-        <span style={{ margin: '0' }}>The request expired due to a server issue.</span>
+        <span style={{ marginBlock: 0, fontSize: '12px' }}>The request expired due to a server issue.</span>
```
cansuaa, line 701: Why font size and the margin block? Where is the margin coming from that we need to set it to 0? For font size maybe you can check Box and its settings

pages/steps/permutations-utils.tsx
Revuj: You were right! But now it's used again :) sorry

pages/steps/with-updates.page.tsx
```diff
@@ -0,0 +1,170 @@
...
+  const timeoutRef1 = useRef<ReturnType<typeof setTimeout>>();
+  const stepsExecution1 = [
+    initialStepsInteractive,
+    loadingStepsInteractive,
+    loadingSteps2Interactive,
+    loadingSteps3Interactive,
+    successfulStepsInteractive,
+  ];
+
+  const [stepIndex2, setStepIndex2] = useState(0);
```
jperals, line 32: This can probably be refactored with a component that manages its own internal state independently, instead of managing 4 almost-identical states in a single component.

Revuj: That is true, but is it worth it at this point? :) It's just an example page

jperals: Let's call this non-blocking but a suggestion. Always good to keep things simple for the next one who might work on it.

package-lock.json
```diff
@@ -6817,108 +6817,6 @@
       "integrity": "sha512-9rTIZ4ZjWwalCPiaY+kPiALLfOKgAz5CTi/Zb1L+qSZ8PH3zVo1T8JcgXIIqg1iM3pZ6hF+n9xO5r2jZ/SF+jg==",
       "dev": true
     },
-    "node_modules/devtools/node_modules/debug": {
```
jperals, line 6820: Do you know why these changes in the `package-lock.json` popped up? If not, can we exclude them from the PR?

Revuj: Excluded :)

src/steps/interfaces.ts
```diff
@@ -0,0 +1,43 @@
...
+import { BaseComponentProps } from '../internal/base-component';
+import { StatusIndicatorProps } from '../status-indicator/internal';
+
+export interface StepsProps extends BaseComponentProps {
+  /**
+   * An array of individual steps
+   *
+   * Each step definition has the following properties:
+   *  * `status` (string) - Status of the step corresponding to a status indicator.
+   *  * `statusIconAriaLabel` - (string) - (Optional) Text alternative for the status icon.
```
jperals, line 12: "Text alternative" or "Alternative text"?

jperals (approved):

cansuaa (approved):

PR #2687, opened 2024-09-10
feat: Add analytics metadata to Flashbar

fralongo:
### Description

Similar to https://github.com/cloudscape-design/components/commit/9c1752711f80122519564dbbe65249f15ca1da1c
<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/flashbar/collapsible-flashbar.tsx
+++ b/src/flashbar/collapsible-flashbar.tsx
@@ -4,6 +4,8 @@ import React, { ReactNode, useCallback, useLayoutEffect, useMemo, useRef, useSta
 import { TransitionGroup } from 'react-transition-group';
 import clsx from 'clsx';
 
+import { getAnalyticsMetadataAttribute } from '@cloudscape-design/component-toolkit/internal/analytics-metadata';
+
 import { useInternalI18n } from '../i18n/context';
 import { IconProps } from '../icon/interfaces';
 import InternalIcon from '../icon/internal';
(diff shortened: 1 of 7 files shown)


src/flashbar/collapsible-flashbar.tsx
```diff
@@ -322,6 +328,13 @@ export default function CollapsibleFlashbar({ items, ...restProps }: FlashbarPro
           )}
           onClick={toggleCollapseExpand}
           ref={notificationBarRef}
+          {...getAnalyticsMetadataAttribute({
```
orangevolon, line 331: 1. This one doesn't get any analytics data
2. Why this one is not using `getComponentsAnalyticsMetadata`?

src/flashbar/flash.tsx
```diff
@@ -41,7 +44,12 @@ function dismissButton(
   onDismiss: FlashbarProps.MessageDefinition['onDismiss']
 ) {
   return (
-    <div className={styles['dismiss-button-wrapper']}>
+    <div
+      className={styles['dismiss-button-wrapper']}
+      {...getAnalyticsMetadataAttribute({
+        action: 'dismiss',
+      } as Partial<GeneratedAnalyticsMetadataFlashbarDismiss>)}
```
orangevolon, line 51: Two questions about this line:

1. We defined `GeneratedAnalyticsMetadataFlashbarDismiss` interface but its only use case is in this line with `Partial<...>`. Why we don't just mark the `label` as optional?
2. Why we don't use a creator functions like the other cases? something like `getDismissButtonAnalyticsMetadata` which returns this type like others?

fralongo: Thank you for the comments:
1. The output the [function that collects the metadata](https://github.com/cloudscape-design/component-toolkit/blob/main/src/internal/analytics-metadata/utils.ts#L11) will have a label. We use this attribute here to override the `action` part of it, which would be a generic `click`. I prefer to keep the type aligned with the outcome of the processing, so that I can automatically generate documentation in the future.
2. More than 15 components are already instrumented, and I normally didn't need to create specific functions to collect the metadata. The Flashbar was an exception because of the double implementation (collapsible and non-collapsible. However, if you prefer I can create a function. Let me know.

orangevolon (approved):

PR #2690, opened 2024-09-11
test: Enable more toolbar unit tests

jkuelz:
### Description

Enables tests in `common.tsx`, `runtime-drawers.tsx`, and `multi-layout.tsx`.
Fixes focus delegation for navigation 
Ensures tools remain mounted even when closed

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/__tests__/common.test.tsx
+++ b/src/app-layout/__tests__/common.test.tsx
@@ -11,7 +11,7 @@ jest.mock('@cloudscape-design/component-toolkit', () => ({
   useContainerQuery: () => [100, () => {}],
 }));
 
-describeEachAppLayout({ themes: ['classic', 'refresh'] }, ({ size }) => {
+describeEachAppLayout(({ theme, size }) => {
   test('Default state', () => {
     const { wrapper } = renderComponent(<AppLayout />);
 
@@ -43,7 +43,8 @@ describeEachAppLayout({ themes: ['classic', 'refresh'] }, ({ size }) => {
       openProp: 'navigationOpen',
       hideProp: 'navigationHide',
       handler: 'onNavigationChange',
-      expectedCallsOnMobileToggle: 2,
+      // TODO: Figure out why this needs to be called twice on mobile
+      expectedCallsOnMobileToggle: theme === 'refresh-toolbar' ? 1 : 2,
       findLandmarks: (wrapper: AppLayoutWrapper) => wrapper.findAll('nav'),
       findElement: (wrapper: AppLayoutWrapper) => wrapper.findNavigation(),
       findToggle: (wrapper: AppLayoutWrapper) => wrapper.findNavigationToggle(),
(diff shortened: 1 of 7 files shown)


src/app-layout/visual-refresh-toolbar/index.tsx
```diff
@@ -305,7 +306,11 @@ const AppLayoutVisualRefreshToolbar = React.forwardRef<AppLayoutProps.Ref, AppLa
           navigation={resolvedNavigation && <AppLayoutNavigation appLayoutInternals={appLayoutInternals} />}
           navigationOpen={navigationOpen}
           navigationWidth={navigationWidth}
-          tools={activeDrawer && <AppLayoutDrawer appLayoutInternals={appLayoutInternals} />}
+          tools={
+            ((drawers!.length > 0 && !toolsHide) || activeDrawer) && (
```
jkuelz, line 310: Need tools to mount even if the drawer is not active (open).

src/app-layout/__tests__/common.test.tsx
```diff
@@ -43,7 +43,8 @@ describeEachAppLayout({ themes: ['classic', 'refresh'] }, ({ size }) => {
       openProp: 'navigationOpen',
       hideProp: 'navigationHide',
       handler: 'onNavigationChange',
-      expectedCallsOnMobileToggle: 2,
+      // TODO: Figure out why this needs to be called twice on mobile
```
just-boris, line 46: Checked the test where this change was introduced: https://github.com/cloudscape-design/components/commit/9cd41aa0deab4c0759c613da1dbe8c9791ce2804

In general, twice events are not expected. We used to receive such bug reports before: https://github.com/cloudscape-design/components/pull/928

Good thing we fixed it on the toolbar now

src/app-layout/__tests__/common.test.tsx
```diff
@@ -142,10 +144,10 @@ describeEachAppLayout({ themes: ['classic', 'refresh'] }, ({ size }) => {
...
 
-          if (landmarks[0].getElement().contains(toggleElement)) {
+          if (theme !== 'refresh-toolbar' && landmarks[0].getElement().contains(toggleElement)) {
             expect(landmarks[0].getElement()).toHaveAttribute('aria-hidden', 'true');
             expect(landmarks[1].getElement()).toHaveAttribute('aria-hidden', 'false');
-          } else {
+          } else if (theme !== 'refresh-toolbar') {
             expect(landmarks[1].getElement()).toContainElement(toggleElement);
             expect(landmarks[1].getElement()).toHaveAttribute('aria-hidden', 'true');
             expect(landmarks[0].getElement()).toHaveAttribute('aria-hidden', 'false');
```
just-boris, line 153: Can we bubble up this repeating part

```js
if(theme === 'refresh-toolbar') {
   // assertions for the toolbar design
} else {
   // current assertions
}
```

I think it would be easier to follow the expected state in this format

src/app-layout/__tests__/multi-layout.test.tsx
```diff
@@ -84,10 +84,11 @@ describeEachAppLayout({ themes: ['refresh-toolbar'], sizes: ['desktop'] }, () =>
         content={<AppLayout data-testid="second" navigationHide={true} tools="testing tools" />}
       />
     );
-    expect(firstLayout.findTools()).toBeFalsy();
-    expect(secondLayout.findTools()).toBeFalsy();
+    expect(isDrawerClosed(firstLayout.findTools())).toEqual(true);
```
just-boris, line 87: This code was meant to assert there is no tools panel on that instance at all, because `toolsHide={true}`. Not sure if `isDrawerClosed` is correct here

src/app-layout/visual-refresh-toolbar/index.tsx
just-boris: That's unfortunate, because I was hoping to make the behavior more straightforward. But if our tests rely on it very much, than our customers do it too and we cannot break them 🥲

src/app-layout/__tests__/multi-layout.test.tsx
jkuelz: Ok, I think there is something going on with this test suite or the current implementation. For a sanity check, I switched to the main branch and added `expect(firstLayout.findTools()).toBeFalsy();` after `expect(secondLayout.findTools()).toBeTruthy();`, after the toggle is clicked, and it fails returning a panel with the same panel content as the `secondLayout`. I've noticed a couple flakey assertions like that in this file, which makes me think there are some other issues at play.

PR #2694, opened 2024-09-11
chore: Add component render metric hook

michaeldowseza:
### Description

Adds a Component Metric hook that will be used to support the funnel metrics through pattern identification. In this revision Table has been instrumented to emit a metric on render with some additional metadata.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/__tests__/__snapshots__/documenter.test.ts.snap
+++ b/src/__tests__/__snapshots__/documenter.test.ts.snap
@@ -14860,6 +14860,11 @@ scroll parent scrolls to reveal the first row of the table.",
         "name": "TableProps.AnalyticsMetadata",
         "properties": [
           {
+            "name": "flowType",
+            "optional": true,
+            "type": ""view-resource"",
+          },
+          {
             "name": "instanceIdentifier",
             "optional": true,
             "type": "string",
--- a/src/internal/analytics/index.ts
+++ b/src/internal/analytics/index.ts
@@ -2,13 +2,16 @@
 // SPDX-License-Identifier: Apache-2.0
 /* istanbul ignore file */
 
-import { IFunnelMetrics, IPerformanceMetrics } from './interfaces';
+import { IComponentMetrics, IFunnelMetrics, IPerformanceMetrics } from './interfaces';
 
 export function setFunnelMetrics(funnelMetrics: IFunnelMetrics) {
   FunnelMetrics = funnelMetrics;
 }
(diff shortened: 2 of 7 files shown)


src/internal/hooks/use-component-metrics/index.ts
```diff
@@ -0,0 +1,22 @@
...
+
+import { ComponentMetrics } from '../../analytics';
+
+export function useComponentMetrics(
+  componentName: string,
+  elementRef: React.RefObject<HTMLElement>,
+  getDetails: () => Record<string, string | boolean | number | undefined>
+) {
+  useEffect(() => {
+    if (!elementRef.current) {
```
orangevolon, line 14: Just a curios question, why we check for the ref when is not needed by the `ComponentMetrics.componentMounte(...)` at all?

pages/table/simple.page.tsx
```diff
@@ -0,0 +1,89 @@
...
+import React, { useState } from 'react';
+
+import { Button, Header, SpaceBetween, Table, TableProps } from '~components';
+import { setComponentMetrics } from '~components/internal/analytics';
+
+import ScreenshotArea from '../utils/screenshot-area';
+setComponentMetrics({
+  componentMounted: event => {
+    console.log(`${event.componentName}Rendered`, event);
+  },
```
orangevolon, line 12: I assume this is left on purpose, right?

orangevolon (dismissed):

pages/table/simple.page.tsx
michaeldowseza: So-so. This page isn't used for testing yet but for now it makes it easy for demo/sharing purposes.

src/internal/hooks/use-component-metrics/index.ts
michaeldowseza: Originally I was going to something similar to the [performance hook](https://github.com/cloudscape-design/components/blob/main/src/internal/hooks/use-performance-marks/index.ts#L50-L57) to check if the element was visible on page. I opted for just firing the event regardless and ended up leaving this behind as a bit of a defensive check that this should only run once it's been attached in the dom. It could possibly be removed though.

connorlanigan (approved):

orangevolon (approved):

orangevolon (approved):

orangevolon (approved):

connorlanigan (approved):

PR #2708, opened 2024-09-13
fix: Focus color contrast on active runtime drawer trigger

johannes-weber:
### Description

Adjust color contrast for the active runtime drawer trigger buttons.
The color contrast (Button background color vs focus ring color) changed:
- light mode:
  -  1:1 before this change
  -  4.82:1 after this change

Before/after this change:

|pr-2708-before-light   |pr-2708-after-light   |
|---|---|

- dark mode: 
  -  1.31:1 before this change
  -  5.95:1 after this change

Before/after this change:

|pr-2708-before-dark   |pr-2708-after-dark   |
|---|---|

Related links, issue #, if available: AWSUI-58442

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/__tests__/__snapshots__/design-tokens.test.ts.snap
+++ b/src/__tests__/__snapshots__/design-tokens.test.ts.snap
@@ -2410,6 +2410,2413 @@ exports[`Design tokens artifacts Design tokens JSON for classic matches the snap
         },
       },
     },
+    "app-layout-tools-drawer-trigger": {
+      "tokens": {
+        "border-radius-alert": {
+          "$description": "The border radius of alerts.",
+          "$value": "2px",
+        },
+        "border-radius-badge": {
+          "$description": "The border radius of badges.",
+          "$value": "16px",
+        },
+        "border-radius-button": {
+          "$description": "The border radius of buttons and segmented control's segments.",
+          "$value": "2px",
+        },
+        "border-radius-calendar-day-focus-ring": {
+          "$description": "The border radius of the focused day in date picker and date range picker.",
+          "$value": "2px",
+        },
+        "border-radius-container": {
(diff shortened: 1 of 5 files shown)


avinashbot (commented):
Why create a whole new context rather than update the styles for the toggle button directly? I'm not fully opposed to this solution, but it also feels like another context for us to maintain and for customers to possibly think about when theming.

Maybe a custom focus style in `src/app-layout/toggles/styles.scss` is the more lightweight alternative (you can provide a custom box-shadow to the focus-highlight mixin)? This is also a hack, but has a very smal impact.

johannes-weber:
> Why create a whole new context rather than update the styles for the toggle button directly? I'm not fully opposed to this solution, but it also feels like another context for us to maintain and for customers to possibly think about when theming.
> 
> Maybe a custom focus style in `src/app-layout/toggles/styles.scss` is the more lightweight alternative (you can provide a custom box-shadow to the focus-highlight mixin)? This is also a hack, but has a very smal impact.

I've created the new context in favor of flexibility: the colors tokens for the toggle selected state are already themeable ($color-background-layout-toggle-selected-default). Because of that, I think having the related focus color in classic themeable provides the flexibility to adjust the focus color contrast for customers as well - if needed.

I tried adding the custom focus style to the toggle buttons directly - in this case I needed to take care about dark/light mode and classic (should only apply there) / VR in the scss/js code. Having the context which already handles the work felt less complex to me.

Curious about your thoughts about.

avinashbot (dismissed):
Given the longetivity of "classic", I wonder whether it can be deprecated in a backward compatible way - it looks like we have more visual contexts than the ones we publicly "document" on our website, so I guess that's the contract.

But the response makes sense to me, happy that the "internal" approach was considered.

avinashbot (approved):

PR #2720, opened 2024-09-16
feat: Prompt input secondary actions and content slots

katiegeorge:
### Description

This adds four new properties to the prompt input:
- `secondaryContent`
- `secondaryActions`
- `disableSecondaryContentPaddings`
- `disableSecondaryActionsPaddings`

The disable paddings properties are necessary because many of our components come with their own padding (including the icon button and the button group, two components that will commonly be used inside these slots).

image

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/prompt-input/__tests__/prompt-input.test.tsx
```diff
@@ -104,18 +110,44 @@ describe('action button', () => {
...
+      actionButtonIconName: 'send',
+      disabled: true,
+    });
     expect(wrapper.findActionButton().getElement()).toHaveAttribute('disabled');
   });
 
-  test('disabled when in read-only state', () => {
-    const { wrapper } = renderPromptInput({ value: '', actionButtonIconName: 'send', readOnly: true });
-    expect(wrapper.findActionButton().getElement()).toHaveAttribute('disabled');
+  test('focusable when in read-only state', () => {
```
connorlanigan, line 143: This test does not seem to actually test that the element is focusable (which as far as I know cannot be done in a unit test using JSDOM, but has to be done in a browser, i.e. an integ test)

src/prompt-input/styles.scss
```diff
@@ -95,14 +131,71 @@ $send-icon-right-spacing: awsui.$space-static-xxs;
...
   }
-  &.textarea-with-button {
-    padding-inline-end: calc(styles.$control-padding-horizontal + $send-icon-right-spacing + awsui.$size-icon-normal);
+}
+
+.secondary-content {
+  @include styles.styles-reset;
+  @include styles.control-border-radius-full();
+  padding-block-start: styles.$control-padding-vertical;
+  padding-block-end: awsui.$space-scaled-s;
```
connorlanigan, line 163: Shouldn't this padding also be conditional, i.e. inside the `.with-paddings` block?

src/prompt-input/__tests__/prompt-input.test.tsx
katiegeorge: Good point. I should update the name of this test to be what it should be testing for, which is the addition of the disabled aria tag.

connorlanigan (approved):

DaemonCahill (approved):

PR #2739, opened 2024-09-19
feat: Property filter enum tokens

pan-kot:
### Description

Adds first-class support for multi-choice (enum) tokens in property filter.

Rel: [Tz3OAr3i2ni1], [0aVwAjMOimcN]

Depends on:
* https://github.com/cloudscape-design/collection-hooks/pull/85
* https://github.com/cloudscape-design/components/pull/2789
* https://github.com/cloudscape-design/components/pull/2818
* https://github.com/cloudscape-design/components/pull/2835
* https://github.com/cloudscape-design/components/pull/2840
* https://github.com/cloudscape-design/components/pull/2841
* https://github.com/cloudscape-design/components/pull/2896
* https://github.com/cloudscape-design/components/pull/2897

### How has this been tested?

* New unit tests
* Screenshot tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/property-filter/common-props.tsx
+++ b/pages/property-filter/common-props.tsx
@@ -158,6 +158,7 @@ export const labels = {
   filteringErrorText: 'Error fetching results.',
   filteringRecoveryText: 'Retry',
   filteringFinishedText: 'End of results',
+  filteringEmpty: 'No suggestions found',
 };
 
 export const i18nStrings: PropertyFilterProps.I18nStrings = {
--- a/pages/property-filter/common-props.tsx
+++ b/pages/property-filter/common-props.tsx
@@ -5,16 +5,7 @@ import React from 'react';
 import { Badge, SpaceBetween } from '~components';
 import { PropertyFilterProps } from '~components/property-filter';
 
-import {
-  DateForm,
-  DateTimeForm,
-  DateTimeFormLegacy,
-  formatDateTime,
-  formatOwners,
-  OwnerMultiSelectForm,
-  YesNoForm,
-  yesNoFormat,
-} from './custom-forms';
+import { DateForm, DateTimeForm, DateTimeFormLegacy, formatDateTime, YesNoForm, yesNoFormat } from './custom-forms';
 import { states, TableItem } from './table.data';
 
 const getStateLabel = (value: TableItem['state'], fallback = 'Invalid value') =>
@@ -34,7 +25,8 @@ export const columnDefinitions = [
     sortingField: 'state',
     header: 'State',
     type: 'enum',
-    getLabel: getStateLabel,
+    getLabel: (value: any) =>
+      Array.isArray(value) ? value.map(v => getStateLabel(v)).join(', ') : getStateLabel(value, value),
     propertyLabel: 'State',
     cell: (item: TableItem) => getStateLabel(item.state),
   },
@@ -259,7 +251,20 @@ export const filteringProperties: readonly PropertyFilterProps.FilteringProperty
   let groupValuesLabel = `${def.propertyLabel} values`;
 
   if (def.type === 'enum') {
-    operators = ['=', '!='].map(operator => ({ operator, format: def.getLabel }));
+    operators = [
+      ...['=', '!='].map(operator => ({ operator, format: def.getLabel, tokenType: 'enum' })),
+      ...[':', '!:'].map(operator => ({ operator, format: def.getLabel, tokenType: 'value' })),
+    ];
+  }
+  if (def.id === 'tags') {
+    const format = (value: string[]) =>
+      value.length <= 5 ? value.join(', ') : [...value.slice(0, 5), `${value.length - 5} more`].join(', ');
+    operators = [
+      { operator: '=', tokenType: 'enum', format, match: (v: unknown[], t: unknown[]) => checkArrayMatches(v, t) },
+      { operator: '!=', tokenType: 'enum', format, match: (v: unknown[], t: unknown[]) => !checkArrayMatches(v, t) },
+      { operator: ':', tokenType: 'enum', format, match: (v: unknown[], t: unknown[]) => checkArrayContains(v, t) },
+      { operator: '!:', tokenType: 'enum', format, match: (v: unknown[], t: unknown[]) => !checkArrayContains(v, t) },
+    ];
   }
 
   if (def.type === 'text') {
@@ -302,19 +307,6 @@ export const filteringProperties: readonly PropertyFilterProps.FilteringProperty
     ];
   }
 
-  // This is not recommended as it nests
-  if (def.id === 'owner') {
-    operators = [
-      {
-        operator: '=',
-        form: OwnerMultiSelectForm,
-        format: formatOwners,
-        match: (itemValue: string, tokenValue: string[]) =>
-          Array.isArray(tokenValue) && tokenValue.some(value => itemValue === value),
-      },
-    ];
-  }
-
   return {
     key: def.id,
     operators: operators,
@@ -323,3 +315,35 @@ export const filteringProperties: readonly PropertyFilterProps.FilteringProperty
     groupValuesLabel,
   };
 });
+
+function checkArrayMatches(value: unknown[], token: unknown[]) {
+  if (!Array.isArray(value) || !Array.isArray(token) || value.length !== token.length) {
+    return false;
+  }
+  const valuesMap = value.reduce<Map<unknown, number>>(
+    (map, value) => map.set(value, (map.get(value) ?? 0) + 1),
+    new Map()
+  );
+  for (const tokenEntry of token) {
+    const count = valuesMap.get(tokenEntry);
+    if (count) {
+      count === 1 ? valuesMap.delete(tokenEntry) : valuesMap.set(tokenEntry, count - 1);
+    } else {
+      return false;
+    }
+  }
+  return valuesMap.size === 0;
+}
+
+function checkArrayContains(value: unknown[], token: unknown[]) {
+  if (!Array.isArray(value) || !Array.isArray(token)) {
+    return false;
+  }
+  const valuesSet = new Set(value);
+  for (const tokenEntry of token) {
+    if (!valuesSet.has(tokenEntry)) {
+      return false;
+    }
+  }
+  return true;
+}
--- a/src/__tests__/__snapshots__/documenter.test.ts.snap
+++ b/src/__tests__/__snapshots__/documenter.test.ts.snap
@@ -12667,7 +12667,7 @@ The \`operation\` property has two valid values: "and", "or", and controls the j
 The \`tokens\` property is an array of objects that will be displayed to the user beneath the filtering input. When \`enableTokenGroups=true\`, the
 \`tokenGroups\` property is used instead, which supports nested tokens.
 Each token has the following properties:
-* value [string]: The string value of the token to be used as a filter.
+* value [unknown]: The value of the token to be used as a filter. Can be null or string for default tokens, string[] for enum tokens, and anything for tokens with custom forms.
 * propertyKey [string]: The key of the corresponding property in filteringProperties.
 * operator ['<' | '<=' | '>' | '>=' | ':' | '!:' | '=' | '!=' | '^' | '!^']: The operator which indicates how to filter the dataset using this token.
 ",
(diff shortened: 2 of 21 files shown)


src/button-dropdown/internal.tsx
```diff
@@ -230,9 +230,7 @@ const InternalButtonDropdown = React.forwardRef(
           {text}
         </InternalButton>
       );
-      trigger = showMainActionOnly ? (
```
pan-kot, line 233: This change is needed so that onKeyDown and onKeyUp handlers are attached that prevent Enter keypress propagation. Without that, the main action cannot be activated with Enter.

src/multiselect/embedded.tsx
```diff
@@ -77,12 +74,7 @@ const EmbeddedMultiselect = React.forwardRef(
     const status = multiselectProps.dropdownStatus;
 
     return (
-      <div
-        role="group"
```
pan-kot, line 81: The wrapper does not need to be a semantic element, the ARIA label is already attached to the list element.

pages/property-filter/common-props.tsx
```diff
@@ -34,7 +25,8 @@ export const columnDefinitions = [
     sortingField: 'state',
     header: 'State',
     type: 'enum',
-    getLabel: getStateLabel,
+    getLabel: (value: any) =>
+      Array.isArray(value) ? value.map(v => getStateLabel(v)).join(', ') : getStateLabel(value, value),
```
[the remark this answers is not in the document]
pan-kot: Yes, that is done for the case when the value is not an array but a string (when the contains / not contains operator is used). In that case, the value can be partial e.g. "Sto" instead of "Stopped" and it should be accepted just fine.

src/button-dropdown/internal.tsx
pan-kot: I am not 100% sure, but I think that is because those should work for both the trigger and the dropdown: if the listeners are added to the button dropdown trigger alone, that might change the current behaviour. I can experiment with that but in a standalone PR.

jperals: Is this fixing an existing bug? If so, can we add coverage for the fix? I'd be OK with having this in a separate PR (actually even better).

pan-kot: Created a separate PR: https://github.com/cloudscape-design/components/pull/2897

src/__tests__/__snapshots__/documenter.test.ts.snap
```diff
@@ -12636,7 +12636,7 @@ The \`operation\` property has two valid values: "and", "or", and controls the j
 The \`tokens\` property is an array of objects that will be displayed to the user beneath the filtering input. When \`enableTokenGroups=true\`, the
 \`tokenGroups\` property is used instead, which supports nested tokens.
 Each token has the following properties:
-* value [string]: The string value of the token to be used as a filter.
+* value [unknown]: The value of the token to be used as a filter. Can be null or string for default tokens, string[] for enum tokens, and anything for tokens with custom forms.
```
jperals, line 12639: Just to double-check that this API change is OK (the doc text update suggests yes), and this will not cause issues to existing customers (do we want to run a dry run?)

pan-kot: I did make a dry-run for the type update in collection hooks. In this PR no type was updated, but will do a dry-run as well, just in case.

The value has been essentially unknown since introduction of custom forms. The current PR should have no impact to the existing customers but those making something with the query and wanting to adopt the feature would need to ensure the type is correctly handled at runtime.

src/property-filter/__tests__/property-filter-enum-tokens.test.tsx
```diff
@@ -0,0 +1,550 @@
...
+    getTokensContent,
+    getOptionsContent,
+    findEditorMultiselect,
+    getEditorOptionsContent,
+    onChange,
+    onLoadItems,
+  };
+};
+
+test('formats tokens using custom and default formatters', () => {
```
jperals, line 100: What do we mean by "custom and default formatters"? Are we indeed testing both?

src/property-filter/__tests__/property-filter-enum-tokens.test.tsx
```diff
@@ -0,0 +1,550 @@
...
+
+    expect(getTokensContent()).toEqual(['State = Stopped, Stopping']);
+    expect(onChange).toHaveBeenCalledWith(
+      expect.objectContaining({
+        detail: { operation: 'and', tokens: [{ propertyKey: 'state', operator: '=', value: ['0', '1'] }] },
+      })
+    );
+  });
+
+  test('keeps value when changing operator so that value type preserves', () => {
```
jperals, line 276: Why not also cover the complementary case: that the value is cleared when switching from an enum-type to a value-type operator and vice versa?

src/property-filter/property-editor.tsx
```diff
@@ -43,6 +57,55 @@ export function PropertyEditorContent<TokenValue = any>({
...
+  onChange: (value: null | string[]) => void;
+  asyncProps: DropdownStatusProps;
+  filteringOptions: readonly InternalFilteringOption[];
+  onLoadItems?: NonCancelableEventHandler<LoadItemsDetail>;
+}) {
+  const valueOptions: readonly { value: string; label: string }[] = filteringOptions
+    .filter(option => option.property?.propertyKey === property.propertyKey)
+    .map(({ label, value }) => ({ label, value }));
+  const valueHandlers = useLoadItems(onLoadItems, '', property.externalProperty);
+  const value = !unknownValue ? [] : Array.isArray(unknownValue) ? unknownValue : [unknownValue];
```
jperals, line 81: According to the function interface, `value` can only be `null` or `string[]`, so if we already checked against falsy value with `!unknownValue`, why do we need to check that it's an array?

src/property-filter/property-editor.tsx
```diff
@@ -43,6 +57,55 @@ export function PropertyEditorContent<TokenValue = any>({
...
+  const selectedOptions = valueOptions.filter(option => value.includes(option.value));
```
jperals, line 82: Why not make value a set instead of an array so that we don't need to iterate it with `value.includes` for every element in `valueOptions`?

src/property-filter/__tests__/property-filter-enum-tokens.test.tsx
jperals: Oh I see, this is already done further up in `can replace value token with enum token` and `can replace enum token with value token`. My bad

src/property-filter/property-editor.tsx
```diff
@@ -43,6 +57,55 @@ export function PropertyEditorContent<TokenValue = any>({
...
+  const filteredOptions = filterOptions(valueOptions, filter);
+  return (
+    <div className={clsx(styles['property-editor'], styles['property-editor-enum'])}>
+      {filteredOptions.length === 0 && (
+        <div className={styles['property-editor-header-enum']}>
+          <InternalCheckbox checked={false} readOnly={true} />
+          {property.groupValuesLabel}
+        </div>
+      )}
```
jperals, line 91: Why is this section not already handled by the embedded multiselect?

src/property-filter/property-editor.tsx
pan-kot: Oh, true. I will change it so that it is unknown in the interface

src/property-filter/property-editor.tsx
pan-kot: The value is casted to an array if it is not, but it is expected to be an array. It is possible to introduce another var like `const valueSet = new Set(value)`, but I don't think it is much necessary in this case.

src/property-filter/property-editor.tsx
pan-kot: Very good question!

We need to ensure the group option always exists - that's why it is excluded from filtering and then inserted separately. However, that does not work well in case there are no options or no matched options. In such case the group text is greyed out not meeting color contrast and the empty text is not shown (because the group option still counts). 

Screenshot 2024-10-21 at 10 53 41

src/property-filter/styles.scss
```diff
@@ -98,6 +116,17 @@ $operator-field-width: 120px;
     flex-grow: 2;
   }
 
+  &-multiselect-wrapper {
+    position: relative;
+    block-size: awsui.$size-vertical-input;
+    min-inline-size: 200px;
+
+    &-inner {
+      position: absolute;
```
jperals, line 125: Why do you need to have an extra inner wrapper with absolute position? A quick check removing this rule in the browser made no difference

src/property-filter/styles.scss
```diff
@@ -75,6 +87,12 @@ $operator-field-width: 120px;
   }
 }
 
+.property-editor-enum {
+  display: flex;
+  flex-direction: column;
+  overflow-y: hidden;
+}
```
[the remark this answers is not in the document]
pan-kot: The embedded multiselect takes care of the overflow. I will check if maybe the overflow can be removed from this class - this does not seem to have any effect now.

Screenshot 2024-10-21 at 10 55 54

src/property-filter/styles.scss
```diff
@@ -98,6 +116,17 @@ $operator-field-width: 120px;
     flex-grow: 2;
   }
 
+  &-multiselect-wrapper {
+    position: relative;
+    block-size: awsui.$size-vertical-input;
+    min-inline-size: 200px;
```
jperals, line 122: - Does 200px need to be hardcoded or can it be related to something else?
- Note that the dropdown is 7 pixels wider than the controls above, is this OK?

Screenshot 2024-10-21 at 10 55 03

src/property-filter/styles.scss
pan-kot: That is to avoid the input growing when more options are added

https://github.com/user-attachments/assets/1a2fcdbc-6acf-4a49-9ac7-53efe0da23fe

src/property-filter/styles.scss
jperals: I also notice a slight difference in the color behavior of this embedded select in comparison with the other selects in this form (in the other ones, the selected option has blue outline only when the dropdown is hovered). But this might be out of scope

pan-kot: The value 200px is not related to anything in particular - it is just somewhat reasonable width for the input.

It is ok that the dropdown is wider - it can be even wider depending on the options width.

pan-kot: Yes, the hover is rather related to the multiselect and is out of scope. We will probably look deeper into that after the release.

src/property-filter/property-editor.tsx
jperals: OK. Can we have a visual regression test for this case?

pan-kot: Will do!

There is a test page already: /light/property-filter/token-editor, and I will add a screenshot test.

src/property-filter/token-editor-inputs.tsx
```diff
@@ -168,3 +181,42 @@ export function ValueInput({
...
+  asyncProps,
+  filteringOptions,
+  onLoadItems,
+}: ValueInputPropsEnum) {
+  const valueOptions = filteringOptions
+    .filter(option => option.property?.propertyKey === property.propertyKey)
+    .map(({ label, value }) => ({ label, value }));
+  const valueAutosuggestHandlers = useLoadItems(onLoadItems, '', property.externalProperty, undefined, operator);
+  const asyncValueAutosuggestProps = { statusType: 'finished' as const, ...valueAutosuggestHandlers, ...asyncProps };
+  const value = Array.isArray(unknownValue) ? unknownValue : [unknownValue];
```
jperals, line 204: Why do we check against `!unknownValue` [here](https://github.com/cloudscape-design/components/pull/2739/files#diff-20f3d037528e34ddb0726f12a7de9c8adf043d278644c88d29c4485259d0a605R81) in `property-editor.tsx` but here we don't?

pan-kot: Good call, I will update!

In the token editor we generally ensure the temp value for enum tokens is created as an empty array instead of null, however it can still be null when coming from the customer-provided query token.

src/property-filter/styles.scss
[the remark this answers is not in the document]
pan-kot: Update: that was needed for the prev design with paddings around the embedded multiselect. Now it does not contribute any value so I removed it.

jperals (approved):

PR #2743, opened 2024-09-20
chore: Enhances property filter test page

pan-kot:
### Description

Adding tags column and page settings to property filter testing page

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


pages/property-filter/table.data.ts
```diff
@@ -16,18 +16,20 @@ export interface TableItem {
...
   lasteventat?: Date;
+  tags?: string[];
+  tagsIndex?: string;
 }
 
-export const states: Record<number, string> = {
-  0: 'Stopped',
-  1: 'Stopping',
-  2: 'Pending',
+export const states: Record<string, string> = {
```
orangevolon, line 23: This to me looks like an `enum`. Any reason why we're not using an enum instead?

pan-kot: This is not an enum - that is a mapping between the enum key like STOPPED, PENDING to a label. The keys can be made an enum but I see no much value in it as the mapping is still to exist.

orangevolon: If it's a mapper then I don't think it would make sense as well.

orangevolon (approved):

PR #2744, opened 2024-09-20
fix: Fix split panel vertical offset calculation

just-boris:
### Description

We already do that for the navigation and tools panel, but not the split panel

https://github.com/cloudscape-design/components/blob/8ca4c6a55dd13eb572df364cb420527fa1366035/src/app-layout/visual-refresh-toolbar/navigation/index.tsx#L31

### How has this been tested?

1. Manually
2. Added an integration test

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/__integ__/app-layout-split-panel.test.ts
+++ b/src/app-layout/__integ__/app-layout-split-panel.test.ts
@@ -313,6 +313,19 @@ describe.each(['classic', 'visual-refresh', 'visual-refresh-toolbar'] as const)(
     })
   );
 
+  test(
+    'side panel stays in the viewport when scrolling down',
+    setupTest(async page => {
+      const splitPanelSelector = wrapper.findSplitPanel().findOpenPanelSide().toSelector();
+      await page.openPanel();
+      await page.switchPosition('side');
+      const { top: offsetBefore } = await page.getBoundingBox(splitPanelSelector);
+      await page.windowScrollTo({ top: 500 });
+      const { top: offsetAfter } = await page.getBoundingBox(splitPanelSelector);
+      expect(offsetAfter).toEqual(offsetBefore);
+    })
+  );
+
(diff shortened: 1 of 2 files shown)


src/app-layout/__integ__/app-layout-split-panel.test.ts
```diff
@@ -313,6 +317,21 @@ describe.each(['classic', 'visual-refresh', 'visual-refresh-toolbar'] as const)(
...
+  test(
+    'side panel stays in the viewport when scrolling down',
+    setupTest(async page => {
+      await page.openPanel();
+      await page.switchPosition('side');
+      await expect(
+        page.isDisplayedInViewport(wrapper.findSplitPanel().findOpenPanelSide().toSelector())
+      ).resolves.toEqual(true);
+      await page.windowScrollTo({ top: 500 });
+      await expect(
```
jperals, line 329: Does this assertion fail without the fix? `findOpenPanelSide` returns the entire split panel, and `isDisplayedInViewport` will return `true` even if the element is only partially visible: https://webdriver.io/docs/api/element/isDisplayedInViewport/

just-boris: Passes, so I would not call this a real regression test. Do you have any suggestions for a better test?

jperals: You could take the vertical position of the panel before and after scrolling and assert that it is the same.

Alternatively, you could verify that some internal element at the top (such as the header) is visible instead of the entire panel. But this approach is more dependent on element sizes so it feels a bit more fragile.

just-boris: Thanks for the suggestion. Now I got the test failing before and working after

jperals (approved):

PR #2782, opened 2024-09-27
fix: Pass empty string to onLoadItems after clicking Clear button

Al-Dani:
### Description

Currently, in the Property Filter there is a specific corner case: with `asyncProperties === true` when moving focus away from the input and then clicking on Clear button, `onLoadItems` is invoked with only previous `filteringText` value. That means, callback for `onLoadItems` won't load items for a new, empty string. The issue happens when filter by free text, not when specific property was detected

This is not expected and different from similar behaviour in autosuggest.

With this fix, there will be multiple subsequent `onLoadItems` calls:
1. with the previous `filteringText` value on input focus, which happens before Clear button callback is invoked;
2. with empty `filteringText`

Related links, issue #, if available: AWSUI-59621

### How has this been tested?

* update unit test to move focus away before clicking on Clear
* (not related to this specific issue) clearing input by selecting the text and deleting with keyboard

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/autosuggest/load-more-controller.ts
+++ b/src/autosuggest/load-more-controller.ts
@@ -36,7 +36,12 @@ export const useAutosuggestLoadMore = ({
     samePage: boolean;
     filteringText?: string;
   }) => {
-    if (filteringText === undefined || lastFilteringText.current !== filteringText) {
+    if (
+      filteringText === undefined ||
+      lastFilteringText.current !== filteringText ||
+      // special handling for clear input click
+      (filteringText === '' && lastFilteringText.current === '')
+    ) {
       if (filteringText !== undefined) {
         lastFilteringText.current = filteringText;
       }
--- a/src/autosuggest/load-more-controller.ts
+++ b/src/autosuggest/load-more-controller.ts
@@ -36,12 +36,7 @@ export const useAutosuggestLoadMore = ({
     samePage: boolean;
     filteringText?: string;
   }) => {
-    if (
-      filteringText === undefined ||
-      lastFilteringText.current !== filteringText ||
-      // special handling for clear input click
-      (filteringText === '' && lastFilteringText.current === '')
-    ) {
+    if (filteringText === undefined || lastFilteringText.current !== filteringText) {
       if (filteringText !== undefined) {
         lastFilteringText.current = filteringText;
       }
@@ -55,7 +50,8 @@ export const useAutosuggestLoadMore = ({
 
   const fireLoadMoreOnRecoveryClick = () => fireLoadMore({ firstPage: false, samePage: true });
 
-  const fireLoadMoreOnInputFocus = () => fireLoadMore({ firstPage: true, samePage: false, filteringText: '' });
+  const fireLoadMoreOnInputFocus = () =>
+    fireLoadMore({ firstPage: true, samePage: false, filteringText: lastFilteringText.current ?? '' });
 
   const fireLoadMoreOnInputChange = (filteringText: string) =>
     fireLoadMore({ firstPage: true, samePage: false, filteringText });
--- a/src/property-filter/__integ__/async-loading.test.ts
+++ b/src/property-filter/__integ__/async-loading.test.ts
@@ -58,6 +58,9 @@ class AsyncPropertyFilterPage extends BasePageObject {
     await this.openValueEdit();
     await this.keys(value);
   };
+  clearFilteringInput = async () => {
+    await this.click(wrapper.findClearButton().toSelector());
+  };
 }
 
 function setupTest(
@@ -205,4 +208,25 @@ test.each<TestCase>(testCases)('%p', (_, asyncProperties, token, scenario) =>
       await page.expectLoadItemsEvents(result);
     }
   })()
+);
+
+test(
+  'calls onLoadItems with an empty string if Clear button is clicked',
+  setupTest(true, 'property', async page => {
+    await page.typeInFilteringInput('1');
+    await page.expectLoadItemsEvents([
+      { firstPage: true, samePage: false, filteringText: '' },
+      { filteringText: '1', firstPage: true, samePage: false },
+    ]);
+
+    await page.click('#focus-target');
+    await page.clearFilteringInput();
+    await page.waitForJsTimers(500);
+    await page.expectLoadItemsEvents([
+      { firstPage: true, samePage: false, filteringText: '' },
+      { filteringText: '1', firstPage: true, samePage: false },
+      { filteringText: '1', firstPage: true, samePage: false },
+      { filteringText: '', firstPage: true, samePage: false },
+    ]);
+  })
 );
--- a/src/property-filter/__integ__/async-loading.test.ts
+++ b/src/property-filter/__integ__/async-loading.test.ts
@@ -58,9 +58,20 @@ class AsyncPropertyFilterPage extends BasePageObject {
     await this.openValueEdit();
     await this.keys(value);
   };
-  clearFilteringInput = async () => {
-    await this.click(wrapper.findClearButton().toSelector());
-  };
+
+  setCursorPosition(startPosition: number, endPosition: number) {
+    return this.browser.execute(
+      (defaultSelector, startPosition, endPosition) => {
+        return (document.querySelector(defaultSelector) as HTMLInputElement).setSelectionRange(
+          startPosition,
+          endPosition
+        );
+      },
+      inputSelector,
+      startPosition,
+      endPosition
+    );
+  }
 }
 
 function setupTest(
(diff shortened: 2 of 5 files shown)


src/property-filter/__integ__/async-loading.test.ts
```diff
@@ -58,6 +58,9 @@ class AsyncPropertyFilterPage extends BasePageObject {
     await this.openValueEdit();
     await this.keys(value);
   };
+  clearFilteringInput = async () => {
```
pan-kot, line 61: Would it be possible to use a unit test instead of the integ test?

src/autosuggest/load-more-controller.ts
```diff
@@ -36,7 +36,12 @@ export const useAutosuggestLoadMore = ({
     samePage: boolean;
     filteringText?: string;
   }) => {
-    if (filteringText === undefined || lastFilteringText.current !== filteringText) {
+    if (
+      filteringText === undefined ||
+      lastFilteringText.current !== filteringText ||
+      // special handling for clear input click
+      (filteringText === '' && lastFilteringText.current === '')
```
pan-kot, line 43: This looks very hacky. We have fireLoadMoreOnScroll, fireLoadMoreOnRecoveryClick, fireLoadMoreOnInputFocus, fireLoadMoreOnInputChange. Would it make sense to introduce one for clear button click specifically?

src/property-filter/__integ__/async-loading.test.ts
```diff
@@ -206,3 +209,24 @@ test.each<TestCase>(testCases)('%p', (_, asyncProperties, token, scenario) =>
...
+      { filteringText: '1', firstPage: true, samePage: false },
+    ]);
+
+    await page.click('#focus-target');
+    await page.clearFilteringInput();
+    await page.waitForJsTimers(500);
+    await page.expectLoadItemsEvents([
+      { firstPage: true, samePage: false, filteringText: '' },
+      { filteringText: '1', firstPage: true, samePage: false },
+      { filteringText: '1', firstPage: true, samePage: false },
```
pan-kot, line 228: why is this event repeating?

src/property-filter/__integ__/async-loading.test.ts
```diff
@@ -206,3 +209,24 @@ test.each<TestCase>(testCases)('%p', (_, asyncProperties, token, scenario) =>
...
+      { firstPage: true, samePage: false, filteringText: '' },
+      { filteringText: '1', firstPage: true, samePage: false },
+    ]);
+
+    await page.click('#focus-target');
+    await page.clearFilteringInput();
+    await page.waitForJsTimers(500);
+    await page.expectLoadItemsEvents([
+      { firstPage: true, samePage: false, filteringText: '' },
+      { filteringText: '1', firstPage: true, samePage: false },
```
pan-kot, line 227: Why is there an event with filteringText: "1" following one with filteringText: ""?

src/autosuggest/load-more-controller.ts
Al-Dani: I agree, and I thought about that. The problem is, `onFocus` will be triggered before on click, and it fires `onLoadItems` with `currentFilteringText` being non-empty. I couldn't find a good way, other than maybe start checking click target, but that sounds like a overcomplicating things

src/property-filter/__integ__/async-loading.test.ts
Al-Dani: first two calls for the first `onLoadItems`, consequent for the load after clicking on clear. So, when you click on clear, `onLoadItems` is called twice, because of `focus` event first, and afterwards after clearing the input

src/property-filter/__integ__/async-loading.test.ts
Al-Dani: events after the interactive part didn't work well with the unit test for some reason

src/property-filter/__integ__/async-loading.test.ts
pan-kot: I did some testing and here is what I found:

When using same test configuration and selecting `label =` as input and then clearing it - the following events are produced:
```
{filteringText: '', samePage: false, firstPage: true} // focus
{filteringText: 'label', samePage: false, firstPage: true} // enter label
{filteringText: '', samePage: false, firstPage: true, filteringProperty: {…}, filteringOperator: '='} // enter operator
{filteringText: '', samePage: false, firstPage: false} // clear 1
{filteringText: '', samePage: false, firstPage: true} // clear 2
```

I think only one is expected (that is what happens with clearing with keyboard).

When using test configuration with asyncProperties=false, there is no even when the search is cleared.

pan-kot: Also, let's add the test for clearing the input with the keyboard (by selecting and clearing the text). In the current mainline there is no event triggered in that case but that is not supported with the current fix.

pan-kot: Idea: can we resolve the issue by sending no event when the clear button is clicked? When the search input is empty the event is not really needed as there is nothing to load.

src/property-filter/__integ__/async-loading.test.ts
pan-kot: I added some tests, all seems to be working: https://github.com/cloudscape-design/components/pull/2917

pan-kot:
@Al-Dani , attached a screen recording from main: it looks like the issue or not reproducible or I should use different steps to reproduce it?

https://github.com/user-attachments/assets/73dc2bd0-c72c-45d1-af88-83655e254a4e

Al-Dani:
@pan-kot no, the issue still persists, but after merging latest changes from main, the fix stopped working 😅
I guess that's the best proof that it wasn't a good fix from the beginning

before clicking on the "clear" button, you need to move focus away from the property filter

Al-Dani:
I agree, two `onLoadItems` calls could be (and should be) resolved on the customer end. But we need to provide them with this second call (that has empty string value) in the first place

src/property-filter/__tests__/property-filter.test.tsx
```diff
@@ -303,12 +308,54 @@ describe('property filter parts', () => {
...
+      createWrapper(container).find('#focus-target')!.focus();
+
+      wrapper.findClearButton()!.click();
+
+      expect(loadItemCalls).toEqual([
+        expect.objectContaining({ filteringText: '' }),
+        expect.objectContaining({ filteringText: '' }),
+        expect.objectContaining({ filteringText: 'stat' }),
+        expect.objectContaining({ filteringText: 'stat' }),
         expect.objectContaining({ filteringText: '' }),
```
pan-kot, line 359: What is different between free-text and parsed option? Since we receive the expected load-more call for the latter, that should be possible for the free-text, too, but w/o causing the second call of it in the other test.

src/property-filter/__tests__/property-filter.test.tsx
```diff
@@ -303,12 +308,54 @@ describe('property filter parts', () => {
...
+      expect(loadItemCalls).toEqual([
+        expect.objectContaining({ filteringText: '' }),
+        expect.objectContaining({ filteringText: '' }),
+        expect.objectContaining({ filteringText: 'stat' }),
+      ]);
+
+      createWrapper(container).find('#focus-target')!.focus();
+
+      wrapper.findClearButton()!.click();
+
```
pan-kot, line 353: If you add `jest.advanceTimersByTime(1000);` - the expected load-items event is sent w/o any changes to the component's code.

Al-Dani: true, added and updated the test. The expected event was there because `filteringStatusType` was still pending. That triggered `fireLoadMoreOnScroll`. But it should have been `finished`, as there are no more results to load

src/property-filter/__tests__/property-filter.test.tsx
```diff
@@ -303,12 +308,17 @@ describe('property filter parts', () => {
...
       wrapper.findClearButton()!.click();
       jest.advanceTimersByTime(1000);
+
       expect(loadItemCalls).toEqual([
         expect.objectContaining({ filteringText: '' }),
         expect.objectContaining({ filteringText: '' }),
         expect.objectContaining({ filteringText: '123', filteringProperty: stateProperty, filteringOperator: '=' }),
+        expect.objectContaining({ filteringText: '123', filteringProperty: stateProperty, filteringOperator: '=' }),
+        expect.objectContaining({ filteringText: '' }),
         expect.objectContaining({ filteringText: '' }),
```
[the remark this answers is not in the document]
Al-Dani: added a new test that describes problematic use-case: user filters by an arbitrary string that isn't parsed to an existing property.

src/property-filter/__tests__/property-filter.test.tsx
Al-Dani: this is actually a matter of handling `filteringStatusType`. If we update the status to `finished` in the example with parsed option, the desired event won't be called either

src/property-filter/__tests__/property-filter.test.tsx
pan-kot: Should we then update the "calls onLoadItems on clear when filter by property and asyncProperties=true" test accordingly? Rn it looks like the behaviour is different between parsed and free-text input.

pan-kot (approved):

PR #2795, opened 2024-10-01
fix: Fix area-chart zero height when statusType is loading then finished with  true

taheramr:
### Description
When an area-chart is rendered with `fitHeight` true and has an initial `statusType` of loading then transitions to `finished` the height of the chart becomes zero.

We calculate the chart height given its ref to `useHeightMeasure` and some deps, this is solved by also passing to it the `statusType` as a dependency.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/area-chart/__integ__/area-chart.test.ts
+++ b/src/area-chart/__integ__/area-chart.test.ts
@@ -2,7 +2,10 @@
 // SPDX-License-Identifier: Apache-2.0
 import useBrowser from '@cloudscape-design/browser-test-tools/use-browser';
 
+import { AreaChartWrapper } from '../../../lib/components/test-utils/selectors';
 import AreaChartPageObject from './page-objects/area-chart-page';
+
+const computedDomainChartWrapper = new AreaChartWrapper('#chart');
 
 function setupTest(url: string, name: string, testFn: (page: AreaChartPageObject) => Promise<void>) {
   return useBrowser(async browser => {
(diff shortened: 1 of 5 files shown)


src/area-chart/__integ__/area-chart.test.ts
```diff
@@ -488,7 +491,23 @@ describe('Controlled', () => {
     })
   );
 });
-
+describe.only('Loading state', () => {
```
YueyingLu, line 494: `only` needs to be removed

YueyingLu (changes requested):

src/area-chart/__integ__/area-chart.test.ts
```diff
@@ -488,7 +491,23 @@ describe('Controlled', () => {
...
+    setupTest('#/light/area-chart/loading', 'chart', async page => {
+      await page.setWindowSize({ width: 500, height: 800 });
+      // the loading is about 1000ms in the loading page and then the
+      // statusType becomes 'finished'. This just waits for it.
+      await page.waitForJsTimers(2000);
+
+      // we only need to validate one of the series here, looping over all
+      // of the series will cause a failure since we might have one series with height 0 (threshold one)
+      const seriesBox = await page.getBoundingBox(computedDomainChartWrapper.findSeries().get(1).toSelector());
+      expect(seriesBox.height).toBeGreaterThan(100);
```
YueyingLu, line 506: Minor: The area chart has height as 200, can we make it check it is greater than 200?

src/area-chart/__integ__/area-chart.test.ts
taheramr: good catch, addressed.

PR #2797, opened 2024-10-01
fix: Shaking split panel

georgylobko:
### Description

This issue occurs when the 'Show scroll bars' option is set to 'Always' in the macOS Appearance System preferences. 

The problem arises when the side split panel is opened at desktop screen size, causing the scroll bar to appear briefly and then hide because of the animation. This problem is fixed via setting `overflow: hidden` to side-panel's parent element.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/visual-refresh-toolbar/index.tsx
+++ b/src/app-layout/visual-refresh-toolbar/index.tsx
@@ -1,6 +1,6 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
-import React, { useEffect, useImperativeHandle, useState } from 'react';
+import React, { useEffect, useImperativeHandle, useMemo, useState } from 'react';
 
 import ScreenreaderOnly from '../../internal/components/screenreader-only';
 import { SplitPanelSideToggleProps } from '../../internal/context/split-panel-context';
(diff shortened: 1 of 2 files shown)


src/app-layout/visual-refresh-toolbar/index.tsx
```diff
@@ -202,10 +202,16 @@ const AppLayoutVisualRefreshToolbar = React.forwardRef<AppLayoutProps.Ref, AppLa
       fireNonCancelableEvent(onSplitPanelPreferencesChange, detail);
     };
 
+    const splitPanelDefaultSize = useMemo(
+      () => getSplitPanelDefaultSize(splitPanelPreferences?.position ?? 'bottom'),
```
pan-kot, line 206: Should the same be done for classic and vr variants? Also, the getSplitPanelDefaultSize can be reused in all variants I think

georgylobko: A fun fact is while answering your question I understood the real root cause of this problem: this "shaking" is caused by split panel animation for width. This bug reproduces only in refresh-toolbar theme, because refresh and classic don't have such an animation. 

This problem is solved by simply applying `overflow: hidden` for the split panel parent element.

pan-kot (approved):

PR #2798, opened 2024-10-01
chore: Added modal performance metrics

guptbhum:
Description
The change involves adding new analytics metric - _timeToContentReadyInModal_  when a modal is loaded. 

### How has this been tested?
- manually tested the change
- integration test added
- unit tests added

<details>

#### Correctness

#### Security

#### Testing

- _Changes are covered with new/existing unit tests?_ new 
- _Changes are covered with new/existing integration tests?_ new
</details>

---


src/modal/internal.tsx
```diff
@@ -82,27 +83,49 @@ function PortaledModal({
...
+  const performanceMetricLogged = useRef<boolean>(false);
+  const loadStartTime = useRef<number>(0);
+  const componentLoadingCount = useRef<number>(0);
   // enable body scroll and restore focus if unmounting while visible
   useEffect(() => {
     return () => {
       enableBodyScrolling();
+      resetModalPerformanceData();
     };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
abdhalees, line 98: is this needed here?

src/modal/internal.tsx
```diff
@@ -82,27 +83,49 @@ function PortaledModal({
...
   // enable / disable body scroll
   useEffect(() => {
     if (visible) {
       disableBodyScrolling();
+      loadStartTime.current = performance.now();
     } else {
       enableBodyScrolling();
+      resetModalPerformanceData();
     }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
abdhalees, line 110: same as above, it's safer to include the functions in the deps array

src/modal/__integ__/modal.test.ts
```diff
@@ -114,3 +115,26 @@ test(
...
+    await browser.url('#/light/modal/with-component-load');
+    const getModalPerformanceMetrics = () =>
+      browser.execute(() => ((window as any).modalPerformanceMetrics ?? []) as ModalPerformanceDataProps[]);
+    await page.click('[data-testid="modal-trigger"]');
+    let metrics = await getModalPerformanceMetrics();
+
+    //verify metrics are not emitted until all the components are loaded
+    expect(metrics.length).toBe(0);
+
+    await delay(2000);
```
abdhalees, line 132: can we have a more specific wait? 
what components are you waiting for to be ready? you can use `await page.waitForVisible` or `await waitFor` instead of specifying a time

src/internal/hooks/use-modal-component-analytics/index.ts
```diff
@@ -0,0 +1,90 @@
...
+
+import { useEffect } from 'react';
+
+import { PerformanceMetrics } from '../../analytics';
+import { ModalContextProps, useModalContext } from '../../context/modal-context';
+import { useEffectOnUpdate } from '../use-effect-on-update';
+
+/**
+ *
+ */
```
abdhalees, line 12: nit: not needed

src/button/internal.tsx
```diff
@@ -121,6 +122,7 @@ export const InternalButton = React.forwardRef(
       }),
       [loading, disabled]
     );
+    usePrimaryButtonModalComponentAnalytics(variant === 'primary', buttonRef, [loading, disabled]);
```
abdhalees, line 125: on line 116 we use `__emitPerformanceMarks`, should we also use it here?

src/spinner/internal.tsx
```diff
@@ -18,7 +19,7 @@ export default function InternalSpinner({
   ...props
 }: InternalSpinnerProps) {
   const baseProps = getBaseProps(props);
-
+  useModalComponentAnalytics();
```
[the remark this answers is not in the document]
guptbhum: I have verified this scenario. It is okay if we include the spinners which are part of other components like button.

src/modal/internal.tsx
```diff
@@ -96,12 +98,35 @@ function PortaledModal({
...
   useEffect(() => {
+    /**
+     * Resets performanceMetricLogged
+     * PerformanceMetricLogged false and componentLoadingCount 0,
+     * indicates that there were no components loading in the modal and it was loaded instantly.
+     * In that case emit 0 as timeToContentReadyInModal
+     */
+    const resetModalPerformanceData = () => {
+      if (loadStartTime.current !== 0 && componentLoadingCount.current === 0 && !performanceMetricLogged.current) {
+        PerformanceMetrics.modalPerformanceData({
```
connorlanigan, line 109: The function name is a bit misleading, since it does not only reset data, it actually _emits_ data. I would recommend to keep only the resetting-related part in this function.

src/modal/internal.tsx
```diff
@@ -96,12 +98,35 @@ function PortaledModal({
...
     } else {
       enableBodyScrolling();
+      resetModalPerformanceData();
     }
-  }, [visible]);
+    return () => {
+      if (!visible) {
+        resetModalPerformanceData();
+      }
+    };
```
connorlanigan, line 128: I think I'm not quite following why we're running this here. This function runs when the Modal component either leaves the invisible state, or when it unmounts. Couldn't we instead just reset the data when the Modal _opens_, i.e. when the effect runs and `visible` is true? That way the (visible) modal will always start  from a clean state

src/internal/hooks/use-modal-component-analytics/index.ts
```diff
@@ -0,0 +1,87 @@
...
+      return;
+    }
+
+    modalContext.componentLoadingCount.current--;
+    emitModalContentReadyMetric(modalContext);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, dependencies);
+};
+
+export const useModalComponentAnalytics = () => {
```
connorlanigan, line 60: To me, this function name doesn't make it entirely clear that it reports the current component as a loading component. Maybe we could make this a bit clearer with a more explicit name.

A suggestion:
```suggestion
export const useModalContextLoadingComponent = () => {
```

src/internal/hooks/use-modal-component-analytics/index.ts
```diff
@@ -0,0 +1,87 @@
...
+    const elementVisible =
+      elementRef.current.offsetWidth > 0 &&
+      elementRef.current.offsetHeight > 0 &&
+      getComputedStyle(elementRef.current).visibility !== 'hidden';
+
+    if (!elementVisible) {
+      return;
+    }
+
+    modalContext.componentLoadingCount.current--;
```
connorlanigan, line 54: We're unconditionally decrementing the counter here, and I don't see that we're preventing this hook from decrementing the counter more than once per component instance. So whenever the dependencies of the hook change, the counter is decremented, which (if I'm not mistaken) can lead to a wrong value over time. It effectively would undo the counting of _other_ elements in the Modal

src/internal/hooks/use-modal-component-analytics/index.ts
```diff
@@ -0,0 +1,87 @@
...
+    modalContext.componentLoadingCount.current++;
+    return () => {
+      modalContext.componentLoadingCount.current--;
+      emitModalContentReadyMetric(modalContext);
+    };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+};
+
+const emitModalContentReadyMetric = (modalContext: ModalContextProps) => {
```
connorlanigan, line 75: We could potentially move this function into the file `modal/internal.tsx`. That way, the function has direct access to e.g. `loadStartTime` and `performanceMetricLogged`, and we don't have to expose those values in the context at all. I think that this could lead to a clearer separation between what the non-Modal components do (report their loading state, and inform the modal that it may emit a metric now) and what the Modal component does (collect the loading states, decide whether to emit a metric)

connorlanigan (changes requested):

src/internal/hooks/use-modal-component-analytics/index.ts
guptbhum: Will update to add a function that relies on the loading state

src/internal/hooks/use-modal-component-analytics/index.ts
guptbhum: will update

src/internal/hooks/use-modal-component-analytics/index.ts
guptbhum: added to modal/internal.tsx

src/modal/__integ__/modal.test.ts
```diff
@@ -114,3 +115,34 @@ test(
...
+    const buttonLoadingState = createWrapper()
+      .findModal()
+      .findFooter()
+      .findButton('#primary-button')
+      .findLoadingIndicator()
+      .toSelector();
+    const checkInterval = 2000;
+    while (await browser.$(buttonLoadingState).isExisting()) {
+      await delay(checkInterval);
+    }
```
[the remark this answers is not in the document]
guptbhum: does not work- will keep the current logic

src/modal/__integ__/modal.test.ts
```diff
@@ -114,3 +115,34 @@ test(
...
+    metrics = await getModalPerformanceMetrics();
+    expect(metrics[0].instanceIdentifier).not.toBeNull();
+    expect(metrics[0].timeToContentReadyInModal).toBeGreaterThanOrEqual(2000);
+    expect(metrics[0].timeToContentReadyInModal).toBeLessThan(2100);
```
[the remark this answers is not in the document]
guptbhum: will update to 3000

src/modal/internal.tsx
guptbhum: updated

src/modal/internal.tsx
guptbhum: Will add a comment and a timer to emit the metric

connorlanigan (approved):

PR #2807, opened 2024-10-04
fix: Display all options for CollectionPreferences content display

youdz:
### Description

Currently in the `CollectionPreferences`, if the `preferences.contentDisplay` prop includes 3 values but the `contentDisplayPreference` lists 50 options, we only display the 3 options. This seems like a clear bug, as the list of full options with label should probably be the source of truth on which ones to allow the customer to toggle and order.

This PR ensures options present in `contentDisplayPreference` but not in `preferences.contentDisplay` are displayed last and made `visible: false`. Given that this is a cheap bug fix, I figured I could open a PR and discuss directly here, and if I missed something and it is expected behavior then I lost very little time.

Note that in order to ensure the "missing" options can be toggled and reordered, the `preferences.contentDisplay` value needs to be updated. A common solution for this is to use an effect, but I believe this is an anti-pattern that results in effect hell. Instead, I decided to only add the missing options to the value right before calling `onChange`, so this is easily traceable and debuggable.

Related links, issue #, if available: None

### How has this been tested?

Tested using the `longOptionsList` example in the CollectionPreferences demo page passing
```
contentDisplay: [
  { id: 'id_3', visible: true },
  { id: 'id_1', visible: true },
]
```

**Before**
Screenshot 2024-10-03 at 20 49 10

**After**
All 50 options displayed, toggleable and reorderable

I also ran unit tests to ensure backward compatibility, and added five new tests for the expected behavior as it's very easy to miss. The fact that no test failed seems to confirm the current behavior is not intended.

<details>

#### Correctness

#### Security

#### Testing

- _Changes are covered with new/existing unit tests?_ ✅
- _Changes are covered with ~~new~~/existing integration tests?_ ✅
</details>

---


src/collection-preferences/content-display/index.tsx
```diff
@@ -52,20 +52,30 @@ export default function ContentDisplayPreference({
...
-  );
+  const [sortedOptions, sortedAndFilteredOptions] = useMemo(() => {
+    const sorted = getSortedOptions({ options, contentDisplay: value });
+    const filtered = sorted.filter(option =>
+      option.label.toLowerCase().trim().includes(columnFilteringText.toLowerCase().trim())
+    );
+    return [sorted, filtered];
+  }, [columnFilteringText, options, value]);
+
+  const addMissingOptionsInValue = () => {
```
gethinwebster, line 66: I'm not sure about this pattern of having to call this `addMissingOptionsInValue` function in multiple places before calling `onChange`, could we instead use `sortedOptions` directly in place of `value` throughout this component? Or create a function that wraps `onChange` and does the addMissing in just one place?

youdz: Good call, I was deep in the weeds when I wrote this and reading it again, it does look weird. I'll fix that and submit a new revision right now.

gethinwebster (approved):

gethinwebster (approved):

PR #2810, opened 2024-10-04
fix: Empty notifications height

georgylobko:
### Description

This issue arises when the notifications in the refresh-toolbar theme are empty. The current implementation relies on the :not(:empty) CSS property, which doesn't work because the notifications DOM node is always present, even when there are no notifications. In this PR, I updated the calculation to match how it's handled in the refresh theme

image

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/visual-refresh-toolbar/index.tsx
+++ b/src/app-layout/visual-refresh-toolbar/index.tsx
@@ -326,6 +326,7 @@ const AppLayoutVisualRefreshToolbar = React.forwardRef<AppLayoutProps.Ref, AppLa
       setToolbarState,
       verticalOffsets,
       drawersOpenQueue,
+      notificationsHeight,
       setToolbarHeight,
       setNotificationsHeight,
       onSplitPanelToggle: onSplitPanelToggleHandler,
--- a/src/app-layout/visual-refresh-toolbar/index.tsx
+++ b/src/app-layout/visual-refresh-toolbar/index.tsx
@@ -5,6 +5,7 @@ import React, { useEffect, useImperativeHandle, useState } from 'react';
 import ScreenreaderOnly from '../../internal/components/screenreader-only';
 import { SplitPanelSideToggleProps } from '../../internal/context/split-panel-context';
 import { fireNonCancelableEvent } from '../../internal/events';
+import * as tokens from '../../internal/generated/styles/tokens';
 import { useControllable } from '../../internal/hooks/use-controllable';
 import { useMobile } from '../../internal/hooks/use-mobile';
 import { useUniqueId } from '../../internal/hooks/use-unique-id';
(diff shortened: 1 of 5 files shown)


src/app-layout/visual-refresh-toolbar/notifications/styles.scss
```diff
@@ -2,7 +2,13 @@
...
 */
+@use '../../../internal/styles/tokens' as awsui;
+
 .sticky-notifications {
   position: sticky;
   z-index: 850;
 }
+
+.has-notifications-content {
+  padding-block-start: awsui.$space-scaled-xs;
```
dpitcock, line 13: I'm seeing the padding still persist in https://d21d5uik3ws71m.cloudfront.net/components/bb292b4cc55f3ea99461ce066e41c6e117158b2c/dev-pages/index.html#/light/app-layout/with-table?appLayoutToolbar=true

```
.awsui_has-notifications-content_154bi_1h0eg_14:not(#\9) {
    padding-block-start: var(--space-scaled-xs-t3am3w, 8px);
}
```

src/app-layout/visual-refresh-toolbar/index.tsx
```diff
@@ -377,7 +379,7 @@ const AppLayoutVisualRefreshToolbar = React.forwardRef<AppLayoutProps.Ref, AppLa
         {!hasToolbar && breadcrumbs ? <ScreenreaderOnly>{breadcrumbs}</ScreenreaderOnly> : null}
         <SkeletonLayout
           style={{
-            [globalVars.stickyVerticalTopOffset]: `${verticalOffsets.header}px`,
+            [globalVars.stickyVerticalTopOffset]: `${verticalOffsets.header}px + ${resolvedStickyNotifications && notificationsHeight > 0 ? tokens.spaceXs : '0px'}`,
```
georgylobko, line 382: You may ask why this space is calculated here, not in the `computeVerticalLayout` or not just calculate it as part of notifications height.

There are 2 reasons:
1. Since now notifications height includes only actual content size, not the _wrapper_ size, we need to take this space somewhere
2. It's not feasible to include this to `computeVerticalLayout`, because `tokens.spaceXs` is not a `number`, but css token (string)

If you have better suggestion how to implement it, let's discuss it in this thread

src/app-layout/visual-refresh-toolbar/notifications/styles.scss
georgylobko: fixed

dpitcock (approved):

just-boris:
Temporarily addressed here https://github.com/cloudscape-design/components/pull/2821, closing this one for now as no urgent need

PR #2822, opened 2024-10-08
fix: Split panel appearance based on splitPanel prop

georgylobko:
### Description

Deactivated split panel should not appear. It works as expected in classic and VR, but visual-refresh-toolbar.

image

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/__tests__/split-panel.test.tsx
+++ b/src/app-layout/__tests__/split-panel.test.tsx
@@ -1,7 +1,7 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 /* eslint simple-import-sort/imports: 0 */
-import React from 'react';
+import React, { useState } from 'react';
 import { screen } from '@testing-library/react';
 import AppLayout from '../../../lib/components/app-layout';
 import { AppLayoutProps } from '../../../lib/components/app-layout/interfaces';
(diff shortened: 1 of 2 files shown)


src/split-panel/implementation.tsx
```diff
@@ -58,6 +58,10 @@ export const SplitPanelImplementation = React.forwardRef<HTMLElement, SplitPanel
     const openButtonAriaLabel = i18nStrings.openButtonAriaLabel;
     useEffect(() => {
       setSplitPanelToggle({ displayed: closeBehavior === 'collapse', ariaLabel: openButtonAriaLabel });
+
```
dpitcock, line 61: Two questions:

1. Should the split panel state be reset upon reenabling?  Currently if you open drawer, then disable it, then re-enable it the split panel is showing in either the bottom and side position. Same for if you change the positioning. Is this what the cusomers expect? OR should the values be reset? 

2. Can we add an integ test for this? perhaps in the `app-layout-with-split-panel.test.tsx` or perhaps add its own test such as `runtime-split-panel-controllability.test.tsx` if applicable.  Here we start with it enabled, and disable it, and also enable it from it being disabled?

src/app-layout/__tests__/split-panel.test.tsx
```diff
@@ -226,6 +226,31 @@ describeEachAppLayout({ sizes: ['desktop'] }, ({ theme }) => {
...
+      return (
+        <AppLayout
+          splitPanel={splitPanelEnabled && defaultSplitPanel}
+          content={
+            <>
+              <button data-testid="toggle-split-panel" onClick={() => setSplitPanelEnabled(!splitPanelEnabled)}>
+                toggle
+              </button>
+            </>
+          }
```
just-boris, line 242:
```suggestion
          content={
            <button data-testid="toggle-split-panel" onClick={() => setSplitPanelEnabled(!splitPanelEnabled)}>
              toggle
            </button>
          }
```

just-boris (dismissed):

just-boris (dismissed):

src/split-panel/implementation.tsx
georgylobko: 1. Yes, I think this behavior is expected
2. I don't see much value adding an integ tests for that. But I agree that enabling split panel after disabling is worth adding to the existing u test

dpitcock (approved):
looks good

PR #2825, opened 2024-10-08
chore: Activate common test for toolbar

fralongo:
<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/__tests__/multi-layout.test.tsx
+++ b/src/app-layout/__tests__/multi-layout.test.tsx
@@ -84,11 +84,11 @@ describeEachAppLayout({ themes: ['refresh-toolbar'], sizes: ['desktop'] }, () =>
         content={<AppLayout data-testid="second" navigationHide={true} tools="testing tools" />}
       />
     );
-    expect(firstLayout.findTools()).toBeFalsy();
-    expect(secondLayout.findTools()).toBeFalsy();
+    expect(isDrawerClosed(firstLayout.findTools())).toEqual(true);
+    expect(isDrawerClosed(secondLayout.findTools())).toEqual(true);
 
     firstLayout.findToolsToggle().click();
-    expect(secondLayout.findTools()).toBeTruthy();
+    expect(isDrawerClosed(secondLayout.findTools())).toEqual(false);
   });
 
   test('merges split panel from two instances', async () => {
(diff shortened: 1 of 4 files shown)


src/app-layout/__tests__/multi-layout.test.tsx
```diff
@@ -84,11 +84,11 @@ describeEachAppLayout({ themes: ['refresh-toolbar'], sizes: ['desktop'] }, () =>
         content={<AppLayout data-testid="second" navigationHide={true} tools="testing tools" />}
       />
     );
-    expect(firstLayout.findTools()).toBeFalsy();
-    expect(secondLayout.findTools()).toBeFalsy();
+    expect(isDrawerClosed(firstLayout.findTools())).toEqual(true);
```
jkuelz, line 87: Forwarding a comment thread from when I made this same change on [the draft PR](https://github.com/cloudscape-design/components/pull/2690/files#r1754089742), so it doesn't get lost. 

> This code was meant to assert there is no tools panel on that instance at all, because toolsHide={true}. Not sure if `isDrawerClosed` is correct here.

I see you disabled a test above in the `common.test.tsx` file that tests for this case, so maybe you plan to address it later?

fralongo: Yeah, `toolsHide` doesn't work. In this specific test the `toolsHide` property is not set

fralongo: I have addressed the `toolsHide` issue

jkuelz: Tools on the first layout, have `toolsHide={true}`, doesn't that mean tools shouldn't exist on that layout? On the second layout, the behavior you described is expected. But here we are asserting on both layouts, which feels incorrect.

fralongo: My understanding is that only one applayout is actually displayed, and the `toolsHide` property of the first is overridden by the `tools` property of the second. But I will wait for @just-boris to confirm

jkuelz: That was my understanding initially as well. I thought deduplication was sending all props to the parent. But if you inspect the DOM in the [multi-layout test page](https://d21d5uik3ws71m.cloudfront.net/components/94661a13c78657add42de9bde8cce7a3c0d9cc6b/dev-pages/index.html#/light/app-layout/multi-layout-iframe), you'll see that the deduplication is only on the toolbar itself. The tools panel still exists in the second (inner) app layout.

just-boris (approved):
All good, only test improvement suggestions

PR #2816, opened 2024-10-07
fix: preserves tools inner in local drawer in AppLayout toolbar variant

dpitcock:
### Description

- preserves tools inner content in local drawer in AppLayout toolbar variant
- removes skipping condition for unit test for runtime drawers
- adds integ test to the stateful page

- not relevant to this fix but in the pull request there is also the addition of a testing util `testOnlyIf()` to be used when debugging integ tests locally.

Human testing Go to https://d21d5uik3ws71m.cloudfront.net/components/ca007ca1f62430c066401d68d11d4ded65cbe88c/dev-pages/index.html#/light/app-layout/stateful?appLayoutToolbar=true then set the param to false and confirm, then set to classic and confirm. 

Related links, issue #, if available: n/a

### How has this been tested?

<details>
   <summary>Review checklist</summary>

_The following items are to be evaluated by the author(s) and the reviewer(s)._

#### Correctness

- _Changes include appropriate documentation updates._
- _Changes are backward-compatible if not indicated, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#public-apis)._
- _Changes do not include unsupported browser features, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#browsers-support)._
- _Changes were manually tested for accessibility, see [accessibility guidelines](https://cloudscape.design/foundation/core-principles/accessibility/)._

#### Security

- _If the code handles URLs: all URLs are validated through [the `checkSafeUrl` function](https://github.com/cloudscape-design/components/blob/main/src/internal/utils/check-safe-url.ts)._

#### Testing

- _Changes are covered with new/existing unit tests?_
- _Changes are covered with new/existing integration tests?_
</details>

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2826, opened 2024-10-08
revert: preserves tools inner in local drawer in AppLayout toolbar variant

just-boris:
Reverts cloudscape-design/components#2816

There is a regression: https://d21d5uik3ws71m.cloudfront.net/components/2dc1dbac64920a30fa81c761749f6ca825d93410/dev-pages/index.html#/light/app-layout/with-split-panel?appLayoutToolbar=true

This page renders an empty space on the right now 

image


fralongo:
Can we merge this instead? https://github.com/cloudscape-design/components/pull/2828

just-boris:
We found a fix: https://github.com/cloudscape-design/components/pull/2828

PR #2831, opened 2024-10-09
chore: adding assertion tools are hidden in own test

dpitcock:
### Description

This test asserts the tools are hidden when expected in all variants of AppLayout

this added test will pick up on unexpected changes before they get to regression as they did here https://github.com/cloudscape-design/components/pull/2826

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/__integ__/app-layout-tools.test.ts
+++ b/src/app-layout/__integ__/app-layout-tools.test.ts
@@ -1,0 +1,48 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import { BasePageObject } from '@cloudscape-design/browser-test-tools/page-objects';
+import useBrowser from '@cloudscape-design/browser-test-tools/use-browser';
+
+import createWrapper from '../../../lib/components/test-utils/selectors';
+import { viewports } from './constants';
+
+const wrapper = createWrapper().findAppLayout();
+class AppLayoutToolsViewPage extends BasePageObject {
+  async openTools() {
+    await this.click(wrapper.findToolsToggle().toSelector());
+  }
+
+  async closeTools() {
+    await this.click(wrapper.findToolsClose().toSelector());
+  }
+}
+
+describe.each(['classic', 'refresh', 'refresh-toolbar'] as const)('%s', theme => {
+  function setupTest(testFn: (page: AppLayoutToolsViewPage) => Promise<void>, path: string) {
+    return useBrowser(async browser => {
+      const page = new AppLayoutToolsViewPage(browser);
+      const url = `#/light/app-layout/${path}`;
+      await page.setWindowSize(viewports.desktop);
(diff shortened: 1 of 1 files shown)


src/app-layout/__integ__/app-layout-tools.test.ts
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  async closeTools() {
+    await this.click(wrapper.findToolsClose().toSelector());
+  }
+}
+
+describe.each(['classic', 'refresh', 'refresh-toolbar'] as const)('%s', theme => {
+  function setupTest(testFn: (page: AppLayoutToolsViewPage) => Promise<void>, path: string) {
+    return useBrowser(async browser => {
```
just-boris, line 22: You can put viewport size here as an extra argument and save one resize operation

```suggestion
    return useBrowser({ ...viewports.desktop }, async browser => {
```

src/app-layout/__integ__/app-layout-tools.test.ts
```diff
@@ -0,0 +1,48 @@
...
+      const page = new AppLayoutToolsViewPage(browser);
+      const url = `#/light/app-layout/${path}`;
+      await page.setWindowSize(viewports.desktop);
+      const params = new URLSearchParams({
+        visualRefresh: `${theme.startsWith('refresh')}`,
+        appLayoutToolbar: `${theme === 'refresh-toolbar'}`,
+      });
```
just-boris, line 29: We have a shared utility for this now: https://github.com/cloudscape-design/components/blob/5b0e892123f96f8004d146d11251dee27894d481/src/app-layout/__integ__/utils.ts#L22-L29

src/app-layout/__integ__/app-layout-tools.test.ts
```diff
@@ -0,0 +1,48 @@
...
+      await browser.url(`${url}?${params.toString()}`);
+      await page.waitForVisible(wrapper.findContentRegion().toSelector());
+      await testFn(page);
+    });
+  }
+
+  test(
+    'tools is displayed properly with split panel',
```
just-boris, line 37: If this was a regression test to prevent: https://github.com/cloudscape-design/components/pull/2826, I do not think it will help.

Tools element was hidden, but still caused the content overflow.

To catch that regression we need a test like this: https://github.com/cloudscape-design/components/blob/5b0e892123f96f8004d146d11251dee27894d481/src/app-layout/__integ__/mobile.test.ts#L52-L53

Or, actually extend that existing test to cover desktop sizes too.

---

P.S. also there is nothing app-layout-specific in that issue. Should we have a global check that none of our dev pages have horizontal scroll?

dpitcock: nice testing method. I haven't came across that. I looked at the html and body with but they remained the same. I do like the horizontal scroll. That is how the regression picked it up

PR #2839, opened 2024-10-10
fix: AppLayout horizontal scroll

georgylobko:
### Description

When 3 drawers are open on a page with the app layout toolbar, resizing the page eventually causes a horizontal scrollbar to appear (see the picture below), even though all drawers are resized to their minimum size. A solution for that is to check for the horizontal scroll on resize and close the drawers following the priority: navigation bar first, then the first opened drawer.

Screenshot 2024-10-09 at 19 08 30

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/app-layout/__integ__/runtime-drawers.test.ts
```diff
@@ -154,7 +154,8 @@ describe('Visual refresh toolbar only', () => {
       })
     );
 
-    test('first opened drawer should be closed when active drawers can not be shrunk to accommodate it (1400px)', () => {
+    test(
```
georgylobko, line 157: I just realized that this test suite didn't run earlier due to an extra wrap with an arrow function

src/app-layout/visual-refresh-toolbar/index.tsx
```diff
@@ -227,6 +223,14 @@ const AppLayoutVisualRefreshToolbar = React.forwardRef<AppLayoutProps.Ref, AppLa
     const navigationFocusControl = useFocusControl(navigationOpen);
     const splitPanelFocusControl = useSplitPanelFocusControl([splitPanelPreferences, splitPanelOpen]);
 
+    const onNavigationToggle = useCallback(
```
georgylobko, line 226: Moved this handler below because it relies on navigationFocusControl, which is declared earlier

src/app-layout/__integ__/runtime-drawers.test.ts
```diff
@@ -77,7 +77,7 @@ describe.each(['classic', 'refresh', 'refresh-toolbar'] as const)('%s', theme =>
 });
 
 describe('Visual refresh toolbar only', () => {
-  function setupTest(testFn: (page: BasePageObject) => Promise<void>) {
+  function setupTest(testFn: (page: BasePageObject, browser: WebdriverIO.Browser) => Promise<void>) {
```
just-boris, line 80: This typing is not cross-version compatible. You will make Andrei's life upgrading webdriverio version easier, if you use a more portable version

src/app-layout/visual-refresh-toolbar/index.tsx
```diff
@@ -369,6 +373,42 @@ const AppLayoutVisualRefreshToolbar = React.forwardRef<AppLayoutProps.Ref, AppLa
...
+      onActiveGlobalDrawersChange,
+      onNavigationToggle,
+    ]);
+
+    useEffect(() => {
+      if (isMobile) {
+        return;
+      }
+      const throttledOnPageResize = throttle(onPageResize, 500);
+      window.addEventListener('resize', throttledOnPageResize);
```
georgylobko, line 408: You might wonder why I didn't use useResizeObserver—the reason is that this approach is easier to mock and test

src/app-layout/__integ__/runtime-drawers.test.ts
just-boris: Yes, this type is not even imported here in this file

georgylobko: because this type is globally available: https://webdriver.io/docs/api/globals/

what exactly do you want me to change to make it forward compatible?

just-boris: Follow this pattern: https://github.com/search?q=repo%3Acloudscape-design%2Fcomponents%20browser%3A%20ConstructorParameters&type=code

georgylobko: done

src/app-layout/__integ__/runtime-drawers.test.ts
```diff
@@ -77,9 +77,14 @@ describe.each(['classic', 'refresh', 'refresh-toolbar'] as const)('%s', theme =>
 });
 
 describe('Visual refresh toolbar only', () => {
-  function setupTest(testFn: (page: BasePageObject) => Promise<void>) {
+  class PageObject extends BasePageObject {
+    getBrowser() {
+      return this.browser;
```
georgylobko, line 82: the `browser` property is protected in the `BasePageObject` class and it needs to be public here. `get browser` does not work as typescript considers it as overriding and throws `TS2611`

src/app-layout/__integ__/runtime-drawers.test.ts
just-boris: Thanks

src/app-layout/visual-refresh-toolbar/index.tsx
just-boris: Sounds like a potential for future issues. 

Also, this makes app layout depending on user interactions – open and then resize renders a different UI than opening on a small screen initially. We should not allow such deviations.

src/app-layout/__integ__/runtime-drawers.test.ts
```diff
@@ -179,7 +186,41 @@ describe('Visual refresh toolbar only', () => {
...
-    });
+      })
+    );
   });
+
+  test(
+    'should prevent the horizontal page scroll from appearing during resize',
+    setupTest(async page => {
+      const hasHorizontalScroll = () =>
+        page.getBrowser().execute(() => document.body.scrollWidth - document.body.clientWidth > 0);
```
just-boris, line 197: NB, we have another similar check here: https://github.com/cloudscape-design/components/blob/50a9c9e9bd86ddddb072be2fc87e2950ea190f68/src/app-layout/__integ__/mobile.test.ts#L52-L53

After merging this, we should follow up and unify detection logic. (Discussed it DM, your version is better, because it properly counts for the scrollbar widths)

src/app-layout/__tests__/runtime-drawers-layout.test.tsx
```diff
@@ -195,5 +195,64 @@ describe('toolbar mode only features', () => {
...
+      wrapper.findDrawerTriggerById('security')!.click();
+      wrapper.findDrawerTriggerById('global-drawer-1')!.click();
+      wrapper.findDrawerTriggerById('global-drawer-2')!.click();
+      expect(wrapper.findNavigation().getElement()).toHaveAttribute('aria-hidden', 'false');
+      expect(wrapper.findActiveDrawer()).toBeTruthy();
+      expect(globalDrawersWrapper.findDrawerById('global-drawer-1')!.isActive()).toBe(true);
+      expect(globalDrawersWrapper.findDrawerById('global-drawer-2')!.isActive()).toBe(true);
+
+      act(() => {
+        Object.defineProperty(HTMLElement.prototype, 'scrollWidth', {
```
just-boris, line 224: 1. If we check only the `document` values, can we only override these values?
2. Why do I not see a reset of this value after the test?

src/app-layout/__integ__/runtime-drawers.test.ts
georgylobko: addressed

src/app-layout/visual-refresh-toolbar/index.tsx
georgylobko: What is your suggestion here?

just-boris: Do something similar to `computeLayout` function. The decision for drawers visibility should be a pure function too.

src/app-layout/visual-refresh-toolbar/index.tsx
```diff
@@ -369,6 +386,36 @@ const AppLayoutVisualRefreshToolbar = React.forwardRef<AppLayoutProps.Ref, AppLa
...
 
+    useEffect(() => {
+      if (isMobile) {
+        return;
+      }
+
+      const totalActiveDrawersMinSize = getActiveDrawersTotalMinSize();
+      const activeNavigationWidth = navigationOpen ? navigationWidth : 0;
+      // collapsed content width is $space-layout-content-horizontal * 2 = 48px
+      const minContentVisibleWidth = 48;
```
georgylobko, line 397: There’s no way to calculate this directly from the DOM tree since `$space-layout-content-horizontal` is applied in the grid-template to `.` area, specifically in this line: https://github.com/cloudscape-design/components/blob/main/src/app-layout/visual-refresh-toolbar/skeleton/styles.scss#L56

src/app-layout/visual-refresh-toolbar/index.tsx
```diff
@@ -116,27 +111,15 @@ const AppLayoutVisualRefreshToolbar = React.forwardRef<AppLayoutProps.Ref, AppLa
       //   and compare a given number with the new drawer id min size
 
       // the total size of all global drawers resized to their min size
-      let totalActiveDrawersMinSize = activeGlobalDrawersIds
-        .map(
-          activeDrawerId => combinedDrawers.find(drawer => drawer.id === activeDrawerId)?.defaultSize ?? MIN_DRAWER_SIZE
```
georgylobko, line 121: I realized that this calculation works incorrectly, as it returns min size only if `defaultSize` is not defined, but it should compare them and return min of them. Fixed, see below.

src/app-layout/visual-refresh-toolbar/index.tsx
just-boris: We should use our `useStableCallback` util.

There is no promise that `onNavigationChange` is a stable callback, so your code does not really do what you expect.

To an extent, the same applies to other `useCallback` additions here. Make sure their dependencies are really stable (`drawers` array, for example is not)

src/app-layout/visual-refresh-toolbar/index.tsx
just-boris: We already have this constant: https://github.com/cloudscape-design/components/blob/cc75463135eb8fa98a746cebdb6d3776673b934f/src/app-layout/visual-refresh-toolbar/compute-layout.ts#L32

I suggest to reuse it somehow

src/app-layout/visual-refresh-toolbar/index.tsx
```diff
@@ -369,6 +386,36 @@ const AppLayoutVisualRefreshToolbar = React.forwardRef<AppLayoutProps.Ref, AppLa
...
+
+      const scrollWidth = activeNavigationWidth + minContentVisibleWidth + totalActiveDrawersMinSize;
+      const hasHorizontalScroll = scrollWidth > placement.inlineSize;
+      if (hasHorizontalScroll) {
+        if (navigationOpen) {
+          onNavigationToggle(false);
+          return;
+        }
+
+        closeFirstDrawer();
```
just-boris, line 407: Didn't we want to have an iteration here, "close until hasHorizontalScroll is false"?

just-boris (commented):
All good overall, only minor comments

src/app-layout/visual-refresh-toolbar/index.tsx
georgylobko: No, we don't, because closing a drawer triggers a change in the dependency array, which leads to recalculations until the horizontal scrollbar is no longer present

georgylobko:
https://github.com/cloudscape-design/components/pull/2839#discussion_r1806489760

Tried. It leads to unexpected results

src/app-layout/visual-refresh-toolbar/index.tsx
just-boris: Not instantly obvious, but okay

just-boris (approved):

PR #2841, opened 2024-10-10
fix: Property filter editor async props

pan-kot:
### Description

This fixes an issue when load items detail from token editor does not include info of selected operator and filtering text, which makes it inconsistent with the same event issued on token creation.

### How has this been tested?

* New unit test

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/common/options-loader.ts
+++ b/pages/common/options-loader.ts
@@ -64,8 +64,9 @@ export function useOptionsLoader<Item>({ pageSize = 25, timeout = 1000, randomEr
         if (randomErrors && Math.random() < 0.3) {
           reject();
         } else {
-          const nextItems = sourceItems.slice(pageNumber * pageSize, (pageNumber + 1) * pageSize);
-          resolve({ items: nextItems, hasNextPage: items.length + nextItems.length < sourceItems.length });
+          const newItems = sourceItems.slice(pageNumber * pageSize, (pageNumber + 1) * pageSize);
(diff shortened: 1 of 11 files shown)


pages/common/options-loader.ts
```diff
@@ -65,7 +73,7 @@ export function useOptionsLoader<Item>({ pageSize = 25, timeout = 1000, randomEr
           reject();
         } else {
           const nextItems = sourceItems.slice(pageNumber * pageSize, (pageNumber + 1) * pageSize);
-          resolve({ items: nextItems, hasNextPage: items.length + nextItems.length < sourceItems.length });
+          resolve({ items: nextItems, hasNextPage: itemsRef.current.length + nextItems.length < sourceItems.length });
```
pan-kot, line 76: This fixes an issue when hasNextPage is set to false because the check is done against the prev state. As result, there is sometimes a bug in test pages when only 1st page of items is loaded.

src/multiselect/styles.scss
```diff
@@ -12,10 +12,7 @@
 
 .embedded {
   @include styles.styles-reset;
-
-  display: flex;
-  flex-direction: column;
-  inline-size: 100%;
+  display: contents;
```
pan-kot, line 15: This is a small fix for embedded multiselect to be used in the property filter (already used in the test page).

avinashbot (approved):

PR #2842, opened 2024-10-10
feat: Allow table column to be marked as dynamic

gethinwebster:
### Description

Allow table columns to be marked as containing dynamic content. When this flag is set,
a resize observer is attached to auto-update the column widths based on new content.

(Resize observer is needed because the dynamic update might happen nested down the React tree,
in which case no React state change occurs in the table component itself).

Related links, issue #, if available: AWSUI-52826

### How has this been tested?

- New tests
- Ran performance tests locally to ensure no negative impact for non-dynamic cases

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/table/__integ__/dynamic-content.test.ts
+++ b/src/table/__integ__/dynamic-content.test.ts
@@ -1,0 +1,33 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import { BasePageObject } from '@cloudscape-design/browser-test-tools/page-objects';
+import useBrowser from '@cloudscape-design/browser-test-tools/use-browser';
+
+import createWrapper from '../../../lib/components/test-utils/selectors/index.js';
+
+const tableWrapper = createWrapper().findTable();
+
+class TablePage extends BasePageObject {
+  async getColumnWidth(columnIndex: number) {
+    const columnSelector = tableWrapper.findColumnHeaders().get(columnIndex).toSelector();
+    const element = await this.browser.$(columnSelector);
+    const size = await element.getSize();
+    return size.width;
+  }
+}
+
+test(
+  'something',
+  useBrowser(async browser => {
+    await browser.url('#/light/table/dynamic-content');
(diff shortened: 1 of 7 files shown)


src/table/__integ__/dynamic-content.test.ts
```diff
@@ -0,0 +1,33 @@
...
+  async getColumnWidth(columnIndex: number) {
+    const columnSelector = tableWrapper.findColumnHeaders().get(columnIndex).toSelector();
+    const element = await this.browser.$(columnSelector);
+    const size = await element.getSize();
+    return size.width;
+  }
+}
+
+test(
+  'something',
```
connorlanigan, line 20:
```suggestion
  'column width adjusts when the column content loads',
```

connorlanigan (changes requested):

connorlanigan (approved):

src/table/thead.tsx
```diff
@@ -176,6 +176,7 @@ const Thead = React.forwardRef(
                 // Expandable option is only applicable to the first data column of the table.
                 // When present, the header content receives extra padding to match the first offset in the data cells.
                 isExpandable={colIndex === 0 && isExpandable}
+                hasDynamicContent={column.hasDynamicContent}
```
pan-kot, line 179: should we ignore it when sticky header is not enabled?

gethinwebster: yes, have now moved around the logic to make this clearer and more concise

pan-kot: Makes sense, but should we still check for stickyHeader flag so that we don't attach observers if the feature is used in a table w/o sticky header, e.g. if the stickiness is conditional.

gethinwebster: `hidden` on [this line](https://github.com/cloudscape-design/components/pull/2842/files#diff-abe617070411a68e1443642bc6854265273a9e0d1452fc449b65a3d0a9355c8dR179) is only set if stickyHeader is true

pan-kot: Ah, I see now.

pan-kot (approved):

PR #2856, opened 2024-10-11
fix: Empty notifications height

just-boris:
### Description

Widget friendly version of #2810. Thanks @georgylobko for the original code.

See inline comments for more details

### How has this been tested?

* Locally
* Updated integration test
* New screenshot test CR-153803954

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/app-layout/utils/content-blocks.tsx
+++ b/pages/app-layout/utils/content-blocks.tsx
@@ -7,7 +7,7 @@ import range from 'lodash/range';
 import BreadcrumbGroup from '~components/breadcrumb-group';
 import Button from '~components/button';
 import Container from '~components/container';
-import Flashbar from '~components/flashbar';
+import Flashbar, { FlashbarProps } from '~components/flashbar';
 import Header from '~components/header';
 import HelpPanel from '~components/help-panel';
 import SideNavigation from '~components/side-navigation';
@@ -66,23 +66,18 @@ export function Navigation() {
 
 export function Notifications() {
   const [visible, setVisible] = useState(true);
-  if (!visible) {
-    return null;
+  const notifications: Array<FlashbarProps.MessageDefinition> = [];
+  if (visible) {
+    notifications.push({
+      type: 'success',
+      header: 'Success message',
+      statusIconAriaLabel: 'success',
+      dismissLabel: 'Dismiss notification',
+      dismissible: true,
+      onDismiss: () => setVisible(false),
+    });
   }
-  return (
-    <Flashbar
-      items={[
-        {
-          type: 'success',
-          header: 'Success message',
-          statusIconAriaLabel: 'success',
-          dismissLabel: 'Dismiss notification',
-          dismissible: true,
-          onDismiss: () => setVisible(false),
-        },
-      ]}
-    />
-  );
+  return <Flashbar items={notifications} />;
 }
 
 export function Footer({ legacyConsoleNav }: { legacyConsoleNav: boolean }) {
--- a/pages/app-layout/utils/content-blocks.tsx
+++ b/pages/app-layout/utils/content-blocks.tsx
@@ -66,18 +66,15 @@ export function Navigation() {
 
 export function Notifications() {
   const [visible, setVisible] = useState(true);
-  const notifications: Array<FlashbarProps.MessageDefinition> = [];
-  if (visible) {
-    notifications.push({
-      type: 'success',
-      header: 'Success message',
-      statusIconAriaLabel: 'success',
-      dismissLabel: 'Dismiss notification',
-      dismissible: true,
-      onDismiss: () => setVisible(false),
-    });
-  }
-  return <Flashbar items={notifications} />;
+  const demoNotification: FlashbarProps.MessageDefinition = {
+    type: 'success',
+    header: 'Success message',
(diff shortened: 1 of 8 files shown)


src/app-layout/__integ__/awsui-applayout.test.ts
```diff
@@ -167,11 +167,18 @@ describe.each(['classic', 'refresh', 'refresh-toolbar'] as Theme[])('%s', theme
   test(
     'does not render notifications slot when it is empty',
```
just-boris, line 168: We already had a test which was supposed to catch the issue, but it was asserting size of the wrong element. Refactored the test. It is now more strict and may require an update even on unrelated design changes, but this is still better than silently passing

src/app-layout/visual-refresh-toolbar/notifications/index.tsx
```diff
@@ -23,8 +24,9 @@ export function AppLayoutNotificationsImplementation({
   children,
 }: AppLayoutNotificationsImplementationProps) {
   const { ariaLabels, stickyNotifications, setNotificationsHeight, verticalOffsets } = appLayoutInternals;
-  const ref = useRef<HTMLElement>(null);
-  useResizeObserver(ref, entry => setNotificationsHeight(entry.borderBoxHeight));
+  const [hasNotificationsContent, contentRef] = useContainerQuery(rect => rect.borderBoxHeight > 0);
```
just-boris, line 27: an extra resize observer may seem like a waste, but it allows us to resolve this rendering logic without talking outside this component, which is useful in the widget environment (compatibility with existing code)

pages/app-layout/utils/content-blocks.tsx
```diff
@@ -66,23 +66,18 @@ export function Navigation() {
 
 export function Notifications() {
   const [visible, setVisible] = useState(true);
-  if (!visible) {
-    return null;
+  const notifications: Array<FlashbarProps.MessageDefinition> = [];
+  if (visible) {
+    notifications.push({
```
georgylobko, line 71: Why do you push an element to the array on the rendering condition? Why not directly provide an array with a single item based on `visible`? At least it would be more readable

just-boris: Do you want something like this

```js
return <Flashbar items={visible ? [demoNotification] : []}>
```

georgylobko: yes

just-boris: Done

georgylobko (dismissed):

src/app-layout/__integ__/awsui-applayout.test.ts
just-boris: This test relied on hard-coded numbers which are different on local and CI environment.

Let's keep the current test and rely on screenshot tests to capture this regression

src/app-layout/__integ__/awsui-applayout.test.ts
```diff
@@ -167,11 +167,19 @@ describe.each(['classic', 'refresh', 'refresh-toolbar'] as Theme[])('%s', theme
...
     setupTest({ pageName: 'with-notifications' }, async page => {
-      const { height: originalHeight } = await page.getBoundingBox(wrapper.findNotifications().toSelector());
-      expect(originalHeight).toBeGreaterThan(0);
+      const { height: headerHeight } = await page.getBoundingBox('#h');
```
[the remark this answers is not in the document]
georgylobko: Could you elaborate what is "dynamic" in this code? And why the height is different in different environments?

src/app-layout/__integ__/awsui-applayout.test.ts
```diff
@@ -167,11 +167,19 @@ describe.each(['classic', 'refresh', 'refresh-toolbar'] as Theme[])('%s', theme
...
+      await expect(
+        page.getElementsCount(wrapper.findNotifications().findFlashbar().findItems().toSelector())
+      ).resolves.toBe(0);
+      const { top: contentTop } = await page.getBoundingBox('[data-testid="content-root"]');
+      const expectedOffset = {
+        classic: headerHeight + 16,
+        refresh: headerHeight + 12,
+        'refresh-toolbar': headerHeight + 54,
+      }[theme];
+      expect(contentTop).toEqual(expectedOffset);
```
georgylobko, line 182: Isn't it be easier to only check notification container height and make sure it equals 0?

just-boris: This is how original test was structured and it still allowed this regression to happen.

The new approach makes sure this test fails every time there is an extended top offset of the main content, regardless how it is implemented internally

src/app-layout/__integ__/awsui-applayout.test.ts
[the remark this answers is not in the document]
just-boris: The header height depends on system font rendering which is not deterministic and varies between Mac/Linux.

Tried to make it more deterministic: https://github.com/cloudscape-design/components/commit/14de51b628a9c09a826fb3db30c9564f42ad7845

It broke other tests, so I abandoned this path

georgylobko (approved):

PR #2746, opened 2024-09-20
fix: restore focus delegation to split panel in visual-refresh-toolbar

dpitcock:
### Description

- works for both spit panel positions
- set focus to split panel resize button when split panel opened via toolbar trigger button
- set focus back to split panel trigger button on the toolbar once it is closed from the close button on the split panel
-  returns split panel interactions in integ tests

Related links, issue #, if available: n/a

### How has this been tested?
- update to app layout integ tests where:
   - removed redundant classes
   - moved test added testutil functions and updated the tests to use those util functions
   - separated tooltip tests into multiple files with a util

got to deployment with the following path & params - /light/app-layout/with-split-panel?appLayoutWidget=true
- open split panel via the first button in the toolbar
- close split panel via its close button
- note the focus
Confirm /light/app-layout/with-split-panel also works in classic and visual refresh

- compare to current https://d21d5uik3ws71m.cloudfront.net/components/3f188c630a8446f5776683b9dffc0f2b6ce157ef/dev-pages/index.html#/light/app-layout/with-split-panel?appLayoutWidget=true

<details>
   <summary>Review checklist</summary>

_The following items are to be evaluated by the author(s) and the reviewer(s)._

#### Correctness

- _Changes include appropriate documentation updates._
- _Changes are backward-compatible if not indicated, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#public-apis)._
- _Changes do not include unsupported browser features, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#browsers-support)._
- _Changes were manually tested for accessibility, see [accessibility guidelines](https://cloudscape.design/foundation/core-principles/accessibility/)._

#### Security

- _If the code handles URLs: all URLs are validated through [the `checkSafeUrl` function](https://github.com/cloudscape-design/components/blob/main/src/internal/utils/check-safe-url.ts)._

#### Testing

- _Changes are covered with new/existing unit tests?_
- _Changes are covered with new/existing integration tests?_
</details>

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2868, opened 2024-10-14
chore: Cleanup and enable remaining toolbar app layout focus delegation tests 

jkuelz:
### Description

- Removed duplicated tests
- Removed unnecessary test id
- Edited and enabled remaining toolbar test. The tab order is different in vr-toolbar on split panel blur because the tab order follows DOM reading order, and toolbar comes before the main content in the toolbar version, thus focus should not go back to tools trigger.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/__integ__/app-layout-focus-delegation.test.ts
+++ b/src/app-layout/__integ__/app-layout-focus-delegation.test.ts
@@ -5,13 +5,12 @@ import useBrowser from '@cloudscape-design/browser-test-tools/use-browser';
 
 import createWrapper from '../../../lib/components/test-utils/selectors';
 import { viewports } from './constants';
-
-const testIf = (condition: boolean) => (condition ? test : test.skip);
+import { getUrlParams, testIf, Theme } from './utils';
 
 const wrapper = createWrapper().findAppLayout();
 
 interface SetupTestObj {
-  theme: string;
+  theme: Theme;
   pageName: string;
   splitPanelPosition?: string;
   mobile: boolean;
@@ -23,17 +22,16 @@ function setupTest(
 ) {
   return useBrowser(async browser => {
     const page = new BasePageObject(browser);
-    const params = new URLSearchParams({
-      visualRefresh: `${theme.startsWith('refresh')}`,
-      appLayoutWidget: `${theme === 'refresh-toolbar'}`,
-      ...(splitPanelPosition
+    const params = getUrlParams(
+      theme,
+      splitPanelPosition
         ? {
-            splitPanelPosition,
+            splitPanelPosition: splitPanelPosition,
           }
-        : {}),
-    });
+        : {}
+    );
     await page.setWindowSize(mobile ? viewports.mobile : viewports.desktop);
-    await browser.url(`#/light/app-layout/${pageName}?${params.toString()}`);
+    await browser.url(`#/light/app-layout/${pageName}?${params}`);
     await page.waitForVisible(wrapper.findContentRegion().toSelector());
     await testFn(page);
   });
@@ -53,19 +51,17 @@ describe.each(['classic', 'refresh', 'refresh-toolbar'] as const)('%s', theme =>
       );
 
       test(
-        'split panel focus moves to slider on open and open button on close',
+        'split panel focus moves to slider on open, and open button on close',
         setupTest(
           async page => {
-            const splitPanelOpenActionEl =
-              theme === 'refresh-toolbar'
-                ? wrapper.findDrawerTriggerById('slide-panel').toSelector()
-                : wrapper.findSplitPanel().findOpenButton().toSelector();
-            await page.click(splitPanelOpenActionEl);
+            await page.click(wrapper.findSplitPanel().findOpenButton().toSelector());
             await expect(page.isFocused(wrapper.findSplitPanel().findSlider().toSelector())).resolves.toBe(true);
             await page.keys(['Tab', 'Tab']);
             await expect(page.isFocused(wrapper.findSplitPanel().findCloseButton().toSelector())).resolves.toBe(true);
             await page.keys('Enter');
-            await expect(page.isFocused(splitPanelOpenActionEl)).resolves.toBe(true);
+            await expect(page.isFocused(wrapper.findSplitPanel().findOpenButton().toSelector())).resolves.toBe(true);
+            await page.keys('Enter');
+            await expect(page.isFocused(wrapper.findSplitPanel().findSlider().toSelector())).resolves.toBe(true);
           },
           { pageName: 'with-split-panel', theme, mobile }
         )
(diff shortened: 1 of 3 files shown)


src/app-layout/__integ__/app-layout-focus-delegation.test.ts
```diff
@@ -53,19 +51,17 @@ describe.each(['classic', 'refresh', 'refresh-toolbar'] as const)('%s', theme =>
...
       test(
-        'split panel focus moves to slider on open and open button on close',
+        'split panel focus moves to slider on open, and open button on close',
         setupTest(
           async page => {
-            const splitPanelOpenActionEl =
-              theme === 'refresh-toolbar'
-                ? wrapper.findDrawerTriggerById('slide-panel').toSelector()
-                : wrapper.findSplitPanel().findOpenButton().toSelector();
-            await page.click(splitPanelOpenActionEl);
```
jkuelz, line 63: The `findOpenButton()` utility works in both versions, no need for separate selectors.

src/app-layout/__integ__/app-layout-focus-delegation.test.ts
```diff
@@ -241,47 +240,6 @@ describe.each(['classic', 'refresh', 'refresh-toolbar'] as const)('%s', theme =>
...
-            await page.keys(['Tab', 'Tab']);
-            await expect(page.isFocused(wrapper.findSplitPanel().findCloseButton().toSelector())).resolves.toBe(true);
-            await page.keys('Enter');
-            await expect(page.isFocused(triggerSelector)).resolves.toBe(true);
-            await page.keys('Enter');
-            await expect(page.isFocused(wrapper.findSplitPanel().findSlider().toSelector())).resolves.toBe(true);
-          },
-          { pageName: 'with-drawers', theme, mobile }
-        )
-      );
```
jkuelz, line 283: Duplicate test (already covered by the first test in this file)

src/app-layout/__integ__/app-layout-focus-delegation.test.ts
```diff
@@ -241,47 +240,6 @@ describe.each(['classic', 'refresh', 'refresh-toolbar'] as const)('%s', theme =>
         );
       });
 
-      test(
-        'drawers focus toggles between open and close buttons',
-        setupTest(
-          async page => {
-            //Altermatomg between triggers because test-1 trigger hidden in overflow menu on mobile,
-            //security has resize button on desktop
-            const triggerSelector = wrapper.findDrawerTriggerById(mobile ? 'security' : 'test-1').toSelector();
```
jkuelz, line 250: Replaced with a drawer which is always visible and is not resizable ('pro-help').

src/app-layout/__integ__/app-layout-focus-delegation.test.ts
```diff
@@ -107,37 +120,23 @@ describe.each(['classic', 'refresh', 'refresh-toolbar'] as const)('%s', theme =>
...
-                : wrapper.findSplitPanel().findOpenButton().toSelector();
-            await page.setWindowSize({ width: 1000, height: 800 });
-            await page.click(triggerSelector);
-            await page.keys(['Tab', 'Tab', 'Enter']);
-            await expect(page.isFocused(triggerSelector)).resolves.toBe(true);
-          },
-          { pageName: 'with-split-panel', theme, mobile, splitPanelPosition: 'side' }
-        )
-      );
-
```
jkuelz, line 140: This test wasn't actually testing what it was describing. This was testing split panel open/close behavior, which is already covered in other tests. Combined the changing position case with the test above.

src/app-layout/__integ__/app-layout-focus-delegation.test.ts
```diff
@@ -107,37 +120,23 @@ describe.each(['classic', 'refresh', 'refresh-toolbar'] as const)('%s', theme =>
...
 
-      //todo tools functionality needs to be added to toolbar
-      testIf(theme !== 'refresh-toolbar')(
-        'focuses tools panel closed button when it is opened using keyboard and caused split panel to change position',
+      test(
+        'focuses tools panel closed button when it is opened and caused split panel to change position',
         setupTest(
           async page => {
-            await page.setWindowSize({ width: 1000, height: 800 });
+            await page.setWindowSize({ width: 878, height: 800 });
```
just-boris, line 127: There is open navigation on that page. On such small screen size, the split panel is in bottom position because of that anyway. Are you sure the test is still authentic in that case?

src/app-layout/visual-refresh/drawers.tsx
```diff
@@ -300,7 +300,6 @@ function DesktopTriggers() {
             selected={hasSplitPanel && isSplitPanelOpen}
             ref={splitPanelRefs.toggle}
             highContrastHeader={headerVariant === 'high-contrast'}
-            testId="awsui-app-layout-trigger-slide-panel"
```
just-boris, line 303: 👀 this may come up as a breaking change (even if undocumented). I will keep an eye how this gets released

src/app-layout/__integ__/app-layout-focus-delegation.test.ts
```diff
@@ -107,37 +120,23 @@ describe.each(['classic', 'refresh', 'refresh-toolbar'] as const)('%s', theme =>
...
             await page.click(wrapper.findSplitPanel().findOpenButton().toSelector());
-            await page.keys(['Tab', 'Tab', 'Tab', 'Tab', 'Enter']);
+            if (theme !== 'refresh-toolbar') {
+              await page.keys(['Tab', 'Tab', 'Tab', 'Tab', 'Enter']);
+            } else {
+              await page.click(wrapper.findToolsToggle().toSelector());
```
just-boris, line 132: Should we use the click for all variants for consistency? Or if using keyboard is important, can we click on the breadcrumbs first and then Tab -> Tab -> Enter to reach the tools panel in that design

src/app-layout/visual-refresh/drawers.tsx
jkuelz: This was only added 2 weeks ago in [#2746](https://github.com/cloudscape-design/components/pull/2746), so should be relatively contained. But still a good call out.

src/app-layout/__integ__/app-layout-focus-delegation.test.ts
jkuelz: This was interesting. When I tested it locally, these sizes seemed to work fine. But just tested again and even 1000 is too small for VR. Increased the width and added assertions to ensure the panel is in the proper positions.

src/app-layout/__integ__/app-layout-focus-delegation.test.ts
jkuelz: Makes sense. Updated

src/app-layout/__integ__/app-layout-focus-delegation.test.ts
jkuelz: Update, I think it had to do with loading the page on mobile size vs desktop. Also added a reset for mobile, where nav is closed at page load.

src/app-layout/__integ__/app-layout-focus-delegation.test.ts
```diff
@@ -107,37 +120,33 @@ describe.each(['classic', 'refresh', 'refresh-toolbar'] as const)('%s', theme =>
...
-      testIf(theme !== 'refresh-toolbar')(
-        'focuses tools panel closed button when it is opened using keyboard and caused split panel to change position',
+      test(
+        'focuses tools panel closed button when it is opened via keyboard and caused split panel to change position',
         setupTest(
           async page => {
-            await page.setWindowSize({ width: 1000, height: 800 });
+            // Mobile nav is closed on page load
+            mobile && (await page.click(wrapper.findNavigationToggle().toSelector()));
+            await page.setWindowSize({ width: 1100, height: 800 });
```
just-boris, line 129: So, in this file we have some desktop/mobile dual tests. But this test sets the fixed size here, which makes it run twice as "mobile" and "desktop" with the same actual screen size.

No questions for the test flow, but we can be more frugal and extract it outside  that `[true, false].forEach(mobile =>` loop

just-boris (approved):

src/app-layout/__integ__/app-layout-focus-delegation.test.ts
jkuelz: That was my original approach, but because it did fail on mobile otherwise (because it has this different navigation behavior), I felt it was still providing value.

PR #2870, opened 2024-10-15
fix: Runtime drawers default active

georgylobko:
### Description

Runtime drawers with `defaultActive: true` were not opening when there were already open drawers on the page. This PR fixes that issue by updating the condition in the `onDrawersRegistered` event. Previously, runtime drawers with `defaultActive: true` would only set active drawers if none were open. Now, the condition has been adjusted to check if drawers with `defaultActive` are not currently open, allowing them to open regardless of other active drawers.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/app-layout/utils/use-drawers.ts
```diff
@@ -74,23 +76,25 @@ function useRuntimeDrawers(
...
         }
-
-        const defaultActiveGlobalDrawers = sortByPriority(globalDrawers).filter(drawer => drawer.defaultActive);
-        defaultActiveGlobalDrawers.forEach(drawer => {
-          onGlobalDrawersChangeStable(drawer.id);
-        });
       }
+
+      const defaultActiveGlobalDrawers = sortByPriority(globalDrawers).filter(
+        drawer => !activeGlobalDrawersIdsStable.current.includes(drawer.id) && drawer.defaultActive
```
georgylobko, line 87: yes, this way, using the ref. otherwise, if we check `activeGlobalDrawersIds` and add it to the deps array, we fall into infinite loop

src/app-layout/utils/use-drawers.ts
```diff
@@ -74,23 +76,25 @@ function useRuntimeDrawers(
...
+      defaultActiveGlobalDrawers.forEach(drawer => {
+        onGlobalDrawersChangeStable(drawer.id);
+      });
     });
     return () => {
       unsubscribe();
       setRuntimeDrawers({ localBefore: [], localAfter: [], global: [] });
     };
-  }, [disableRuntimeDrawers, onGlobalDrawersChangeStable, onLocalDrawerChangeStable]);
+  }, [disableRuntimeDrawers, onGlobalDrawersChangeStable, onLocalDrawerChangeStable, activeGlobalDrawersIdsStable]);
```
ajsuvarna6, line 97: nit: since `activeGlobalDrawersIdsStable` is a ref, do we need to add it to dependency deps?

src/app-layout/utils/use-drawers.ts
```diff
@@ -63,8 +63,10 @@ function useRuntimeDrawers(
   const onLocalDrawerChangeStable = useStableCallback(onActiveDrawerChange);
   const onGlobalDrawersChangeStable = useStableCallback(onActiveGlobalDrawersChange);
 
-  const drawersWereOpenRef = useRef(false);
-  drawersWereOpenRef.current = drawersWereOpenRef.current || !!activeDrawerId || !!activeGlobalDrawersIds.length;
+  const localDrawerWasOpenRef = useRef(false);
+  localDrawerWasOpenRef.current = localDrawerWasOpenRef.current || !!activeDrawerId;
+  const activeGlobalDrawersIdsStable = useRef<Array<string>>([]);
```
ajsuvarna6, line 68: nit: `activeGlobalDrawersIdsStable` could be named as `activeGlobalDrawersIdsStableRef` postfix it with `Ref`.

src/app-layout/__tests__/runtime-drawers.test.tsx
```diff
@@ -1042,5 +1042,36 @@ describe('toolbar mode only features', () => {
 
       expect(wrapper.findActiveDrawer()!.getElement()).toHaveTextContent('runtime drawer content');
     });
+
+    test('dynamically registered drawers with defaultActive: true should open even when there are already open drawer(s) on the page', async () => {
```
just-boris, line 1046: Do we have negative case scenarios?

1. After max drawers already open, `defaultActive` does not open a new one
2. After user opened a global drawer manually, late registered `defaultActive` don't open

georgylobko: No for both. Do we have any specific requirements for this? If so, could you point me to where I can check them?

just-boris: That's my understanding of our discussion for this bug yesterday. I do not think there is any written down text

just-boris (approved):

PR #2872, opened 2024-10-15
fix: hide navigation button when toolbar variant of AppLayout has hideNavigation

dpitcock:
### Description

Overview 
- this hides the navigation toggle button when toolbar variant of AppLayout has hideNavigation prop of true.
- it also insures there is never a situation where a toolbar of AppLayout will exist with navigationOpen is true when navigationHide is true which is the case in some consoles. This insures the rendering is as expected in those consoles
- New page added in separate PR - https://github.com/cloudscape-design/components/pull/2898 and problem can be reproduced at its deployment https://d21d5uik3ws71m.cloudfront.net/components/35d3bf2eab5555a1a8ec4b7eb6c7c36f2d378fa4/dev-pages/index.html#/light/app-layout/controlled-navigation/?appLayoutToolbar=true&navigationHide=false&applyNavOpenOverride=true&navOpenOverrideValue=false
- This also resolves the issue of a blank screen when resizing to mobile when onNavigationChange has its own component

   start with `<AppLayout navigationOpen={true} navigationHide={true} onNavigationChange={() => {/*decline built-in navigation controlled close feature*/}} />` then resize the screen on mobile mode and note the content is no longer visible

Other
- testing the mergeProps in isolation on own unit test file
- testing the specific case of AppLayout where the navigationHide prop is true with the assertion that no navigation-panel nor navigation toggle is visible.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/app-layout/controlled-navigation.page.tsx
+++ b/pages/app-layout/controlled-navigation.page.tsx
@@ -18,9 +18,8 @@ type ControlledNavigationDemoContext = React.Context<
   AppContextType<{
     navigationOpen: AppLayoutProps['navigationOpen'];
     navigationHide: AppLayoutProps['navigationHide'];
-    navOpenOverrideValue?: boolean;
-    navHideOverrideValue?: boolean;
-    applyNavOverrides: boolean;
+    navOpenOverrideValue?: boolean | null;
+    applyNavOpenOverride: boolean;
   }>
 >;
 
@@ -29,22 +28,20 @@ export default function () {
     urlParams: {
       navigationOpen = true,
       navigationHide = false,
-      navOpenOverrideValue = false,
-      navHideOverrideValue = false,
-      applyNavOverrides = false,
+      navOpenOverrideValue = null,
+      applyNavOpenOverride = false,
     },
     setUrlParams,
   } = useContext(AppContext as ControlledNavigationDemoContext);
   const [resetNeeded, setResetNeeded] = useState(false);
   const [navigationEmpty, setNavigationEmpty] = useState(false);
-  const resolvedNavOpen = applyNavOverrides ? navOpenOverrideValue : navigationOpen;
-  const resolvedNavHide = applyNavOverrides ? navHideOverrideValue : navigationHide;
+  const [resolvedNavOpen, setResolvedNavOpen] = useState(navigationOpen);
 
   const Content = (
     <>
       <div style={{ marginBottom: '1rem' }}>
         <Header variant="h1" description="Basic demo with split panel">
-          AppLayout where NavigationHide resolves to {resolvedNavHide.toString()}
+          AppLayout where NavigationHide is {navigationHide.toString()}
         </Header>
       </div>
 
@@ -61,33 +58,28 @@ export default function () {
           </Toggle>
         </SpaceBetween>
         <SpaceBetween size="s" direction="horizontal">
-          <Checkbox onChange={e => setUrlParams({ applyNavOverrides: e.detail.checked })} checked={applyNavOverrides}>
+          <Checkbox
+            onChange={e => setUrlParams({ applyNavOpenOverride: e.detail.checked })}
+            checked={applyNavOpenOverride}
+          >
             Apply override
           </Checkbox>
           <Toggle
             checked={!!navOpenOverrideValue}
-            disabled={!applyNavOverrides}
+            disabled={!applyNavOpenOverride}
             onChange={e => setUrlParams({ navOpenOverrideValue: e.detail.checked })}
           >
-            Navigation Open Override
+            override Navigation Open
           </Toggle>
-          <Toggle
-            checked={!!navHideOverrideValue}
-            disabled={!applyNavOverrides}
-            onChange={e => setUrlParams({ navHideOverrideValue: e.detail.checked })}
+          <Button
+            onClick={() => setResetNeeded(true)}
+            disabled={resetNeeded}
+            loading={resetNeeded}
+            data-testid="reset-app-layout"
           >
-            Navigation Hide Override
-          </Toggle>
+            Reset - Force rerender
+          </Button>
         </SpaceBetween>
-        <Button
-          onClick={() => setResetNeeded(true)}
-          disabled={resetNeeded}
-          loading={resetNeeded}
-          data-testid="reset-app-layout"
-        >
-          Reset - Force rerender
-        </Button>
-
         <Containers />
       </SpaceBetween>
     </>
@@ -95,12 +87,18 @@ export default function () {
 
   useEffect(() => {
     if (resetNeeded) {
-      setUrlParams({ applyNavOverrides: false, navOpenOverrideValue: false, navHideOverrideValue: false });
+      setUrlParams({ applyNavOpenOverride: false, navOpenOverrideValue: null });
       setTimeout(() => {
         setResetNeeded(false);
       }, 200);
     }
   }, [resetNeeded, setUrlParams]);
+
+  useEffect(() => {
+    if (applyNavOpenOverride && navOpenOverrideValue !== null) {
+      setResolvedNavOpen(navOpenOverrideValue);
+    }
+  }, [applyNavOpenOverride, navOpenOverrideValue]);
 
   return (
     <ScreenshotArea gutters={false}>
@@ -110,7 +108,7 @@ export default function () {
         <AppLayout
           data-testid="main-layout"
           ariaLabels={labels}
-          navigationHide={resolvedNavHide}
+          navigationHide={navigationHide}
           navigation={navigationEmpty ? <></> : <Navigation />}
           navigationOpen={resolvedNavOpen}
           onNavigationChange={e => setUrlParams({ navigationOpen: e.detail.open })}
--- a/pages/app-layout/controlled-navigation.page.tsx
+++ b/pages/app-layout/controlled-navigation.page.tsx
@@ -50,25 +50,13 @@ export default function () {
 
       <SpaceBetween size="l">
         <SpaceBetween size="s" direction="horizontal">
-          <Toggle
-            id="control-navigation-open"
-            checked={navigationOpen}
-            onChange={e => setUrlParams({ navigationOpen: e.detail.checked })}
-          >
+          <Toggle checked={navigationOpen} onChange={e => setUrlParams({ navigationOpen: e.detail.checked })}>
             Navigation Open
           </Toggle>
-          <Toggle
-            id="control-navigation-hide"
-            checked={navigationHide}
-            onChange={e => setUrlParams({ navigationHide: e.detail.checked })}
-          >
+          <Toggle checked={navigationHide} onChange={e => setUrlParams({ navigationHide: e.detail.checked })}>
             Navigation Hide
           </Toggle>
-          <Toggle
-            id="control-navigation-empty"
-            checked={navigationEmpty}
-            onChange={e => setNavigationEmpty(e.detail.checked)}
-          >
+          <Toggle checked={navigationEmpty} onChange={e => setNavigationEmpty(e.detail.checked)}>
             Navigation Empty
           </Toggle>
         </SpaceBetween>
@@ -77,7 +65,6 @@ export default function () {
             Apply override
           </Checkbox>
           <Toggle
-            id="override-nav-open"
             checked={!!navOpenOverrideValue}
             disabled={!applyNavOverrides}
             onChange={e => setUrlParams({ navOpenOverrideValue: e.detail.checked })}
@@ -85,7 +72,6 @@ export default function () {
             Navigation Open Override
           </Toggle>
           <Toggle
-            id="override-nav-open"
             checked={!!navHideOverrideValue}
             disabled={!applyNavOverrides}
             onChange={e => setUrlParams({ navHideOverrideValue: e.detail.checked })}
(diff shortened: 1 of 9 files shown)


src/app-layout/__tests__/multi-layout.test.tsx
```diff
@@ -46,6 +50,115 @@ async function renderAsync(jsx: React.ReactElement) {
...
+        controlId: 'test',
+        ariaLabel: 'test',
+      },
+    },
+  ];
+
+  it('should merge ownProps and additionalProps correctly', () => {
+    const result = mergeMultiAppLayoutProps(ownProps, additionalPropsBase);
+
+    expect(result).toEqual({
```
dpitcock, line 118: is this the wrong assertion?

src/app-layout/visual-refresh-toolbar/toolbar/index.tsx
```diff
@@ -54,6 +54,7 @@ interface AppLayoutToolbarImplementationProps {
 
 // support compatibility with changes before this commit: cf0f2b0755af1a28ac7c3c9476418a7ea807d0fd
 function convertLegacyProps(toolbarProps: ToolbarProps, legacyProps: AppLayoutInternals): ToolbarProps {
+  console.log({ tbp: toolbarProps.hasNavigation, lp: !!legacyProps.navigation });
```
dpitcock, line 57: will remove

src/app-layout/visual-refresh-toolbar/multi-layout.ts
```diff
@@ -49,6 +52,7 @@ function mergeProps(ownProps: SharedProps, additionalProps: ReadonlyArray<Partia
       toolbar.drawersFocusRef = props.drawersFocusRef;
       toolbar.onActiveDrawerChange = props.onActiveDrawerChange;
     }
+    //there is never a case where navigation will exist and a toggle will not
```
dpitcock, line 55: this is where legacy props can overwrite a navigationHide

src/app-layout/visual-refresh-toolbar/toolbar/index.tsx
```diff
@@ -64,7 +65,8 @@ function convertLegacyProps(toolbarProps: ToolbarProps, legacyProps: AppLayoutIn
     globalDrawers: toolbarProps.globalDrawers ?? legacyProps.globalDrawers,
     activeGlobalDrawersIds: toolbarProps.activeGlobalDrawersIds ?? legacyProps.activeGlobalDrawersIds,
     onActiveGlobalDrawersChange: toolbarProps.onActiveGlobalDrawersChange ?? legacyProps.onActiveGlobalDrawersChange,
-    hasNavigation: toolbarProps.hasNavigation ?? !!legacyProps.navigation,
+    hasNavigation:
+      typeof toolbarProps.hasNavigation === 'boolean' ? toolbarProps.hasNavigation : !!legacyProps.navigation,
```
dpitcock, line 69: this type check does not help,   somehow we need to stop legacyProps.navigation from overwriting hasNavigation when navigationHide is true.

src/app-layout/visual-refresh-toolbar/index.tsx
```diff
@@ -299,7 +300,7 @@ const AppLayoutVisualRefreshToolbar = React.forwardRef<AppLayoutProps.Ref, AppLa
       breadcrumbs,
       discoveredBreadcrumbs,
       stickyNotifications: resolvedStickyNotifications,
-      navigationOpen,
+      navigationOpen: !!resolvedNavigation && navigationOpen,
```
dpitcock, line 303: resolvedNavigation value of null will ensure the navigation is not opened when navigationHidden.

src/app-layout/visual-refresh-toolbar/index.tsx
```diff
@@ -235,8 +235,11 @@ const AppLayoutVisualRefreshToolbar = React.forwardRef<AppLayoutProps.Ref, AppLa
       focusSplitPanel: () => splitPanelFocusControl.refs.slider.current?.focus(),
     }));
 
-    const resolvedNavigation = navigationHide ? null : navigation ?? <></>;
     const resolvedStickyNotifications = !!stickyNotifications && !isMobile;
+    //navigation must be null if hidden so toolbar knows to hide the toggle button
+    const resolvedNavigation = navigationHide ? null : navigation ?? <></>;
+    //navigation must not be open if navigationHide is true
+    const resolvedNavigationOpen = !!resolvedNavigation && navigationOpen;
```
dpitcock, line 242: Even if no navigation provided, the fragment passed will be truthy, therefor resolvedNavigation will only be falsy if navigationHide is true

pages/app-layout/controlled-navigation.page.tsx
```diff
@@ -0,0 +1,103 @@
...
+type ControlledNavigationDemoContext = React.Context<
+  AppContextType<{
+    navigationOpen: AppLayoutProps['navigationOpen'];
+    navigationHide: AppLayoutProps['navigationHide'];
+  }>
+>;
+
+export default function () {
+  const {
+    urlParams: { navigationOpen = true, navigationHide = false },
```
dpitcock, line 25: so we can easily set expectations in regression tests

src/app-layout/__tests__/multi-layout.test.tsx
```diff
@@ -75,6 +75,25 @@ describeEachAppLayout({ themes: ['refresh-toolbar'], sizes: ['desktop'] }, () =>
     expect(isDrawerClosed(firstLayout.findNavigation())).toEqual(false);
   });
 
+  test('merges navigation from two instances with navigation hidden in primary', async () => {
+    const { firstLayout, secondLayout } = await renderAsync(
+      <AppLayout
+        {...defaultAppLayoutProps}
+        data-testid="first"
+        navigation="testing nav"
+        navigationHide={true}
```
abdhalees, line 84: The second layout using navigationHide is a more common case.

dpitcock: I updated the test name and added some assertions below to cover this case more thoroughly.

src/app-layout/__tests__/multi-layout-props.test.tsx
```diff
@@ -0,0 +1,109 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+/* eslint-disable simple-import-sort/imports */
```
just-boris, line 3: We only do this in files which use `src/app-layout/__tests__/utils.tsx` because it contains side effects.

All other tests can benefit from the linter rule

src/app-layout/__tests__/multi-layout.test.tsx
```diff
@@ -75,7 +75,26 @@ describeEachAppLayout({ themes: ['refresh-toolbar'], sizes: ['desktop'] }, () =>
...
+        toolsHide={true}
+        content={
+          <AppLayout {...defaultAppLayoutProps} data-testid="second" navigationHide={true} tools="testing tools" />
+        }
+      />
+    );
+    expect(firstLayout.findNavigation()).toBeFalsy();
+    expect(firstLayout.findNavigationToggle()).toBeFalsy();
+    expect(secondLayout.findNavigation()).toBeFalsy();
+    expect(secondLayout.findNavigationToggle()).toBeFalsy();
```
just-boris, line 94: What do we test here?

Is it about `navigationHide` is respected, even if `navigation` is defined? Then we should call it in the test name

just-boris (commented):
small comments only

just-boris (dismissed):

src/app-layout/visual-refresh-toolbar/index.tsx
```diff
@@ -224,8 +224,20 @@ const AppLayoutVisualRefreshToolbar = React.forwardRef<AppLayoutProps.Ref, AppLa
       focusSplitPanel: () => splitPanelFocusControl.refs.slider.current?.focus(),
     }));
 
-    const resolvedNavigation = navigationHide ? null : navigation ?? <></>;
     const resolvedStickyNotifications = !!stickyNotifications && !isMobile;
+    //navigation must be null if hidden so toolbar knows to hide the toggle button
+    const resolvedNavigation = useMemo(() => {
```
dpitcock, line 229: linter recommending I useMemo this hook to prevent exhaustive hooks

just-boris: This makes the linter happy, but does not address the root cause. Why do we even pass `resolvedNavigation` into the `useEffect` below on the line 425?

pages/app-layout/controlled-navigation.page.tsx
```diff
@@ -28,92 +29,79 @@ export default function () {
...
     </>
   );
 
   useEffect(() => {
     if (resetNeeded) {
-      setUrlParams({ applyNavOpenOverride: false, navOpenOverrideValue: null });
+      setUrlParams({ applyNavOverrides: false, navOpenOverrideValue: false, navHideOverrideValue: false });
       setTimeout(() => {
         setResetNeeded(false);
       }, 200);
```
just-boris, line 101: Instead of effect, this whole async logic could be inside the reset button onClick handler

pages/app-layout/controlled-navigation.page.tsx
```diff
@@ -28,92 +29,79 @@ export default function () {
...
+      navHideOverrideValue = false,
+      applyNavOverrides = false,
     },
     setUrlParams,
   } = useContext(AppContext as ControlledNavigationDemoContext);
   const [resetNeeded, setResetNeeded] = useState(false);
   const [navigationEmpty, setNavigationEmpty] = useState(false);
-  const [resolvedNavOpen, setResolvedNavOpen] = useState(navigationOpen);
+  const resolvedNavOpen = applyNavOverrides ? navOpenOverrideValue : navigationOpen;
+  const resolvedNavHide = applyNavOverrides ? navHideOverrideValue : navigationHide;
```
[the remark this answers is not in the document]
dpitcock: they don't. It is an attempt to recreate the bug as seen before.  I will return its page to it's past state but that still shows the bug

just-boris:
Steps to reproduce the issue

1. Have this props combination 

```js
<AppLayout navigationOpen={true} navigationHide={true} onNavigationChange={() => {/*decline built-in navigation controlled close feature*/}} />
```

2. Resize the screen on mobile mode

Expected – the content is visible
Actual – the content is not visible

src/app-layout/__tests__/multi-layout.test.tsx
```diff
@@ -89,7 +108,9 @@ describeEachAppLayout({ themes: ['refresh-toolbar'], sizes: ['desktop'] }, () =>
     expect(createWrapper().findAllByClassName(testUtilStyles.tools)).toHaveLength(1);
 
     firstLayout.findToolsToggle().click();
-    expect(secondLayout.findOpenToolsPanel()).toBeTruthy();
+    expect(secondLayout.findNavigation()).toBeFalsy();
+    expect(secondLayout.findNavigationToggle()).toBeFalsy();
+    expect(secondLayout.findOpenToolsPanel).toBeTruthy();
```
just-boris, line 113: Missed to call the method

```suggestion
    expect(secondLayout.findOpenToolsPanel()).toBeTruthy();
```

src/app-layout/__tests__/app-layout-navigation-resize.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+    expect(mockOnNavigationChange).toHaveBeenCalledTimes(0);
+    expect(wrapper).not.toBeNull();
+    expect(wrapper.findNavigation()).toBeFalsy();
+    expect(wrapper.findNavigationToggle()).toBeFalsy();
+    expect(wrapper.findByClassName(visualRefreshToolbarStyles['main-landmark'])).not.toBeNull();
+    expect(wrapper.findByClassName(visualRefreshToolbarStyles['unfocusable-mobile'])).toBeNull();
+    expect(wrapper.findByClassName(visualRefreshToolbarStyles.content)?.getElement()).toBeVisible();
+
+    (useMobile as jest.Mock).mockReturnValue(true);
```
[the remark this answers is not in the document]
dpitcock: We are testing the resizing from desktop to mobile. Would integ tests be better for this?

src/app-layout/__tests__/app-layout-navigation.test.tsx
```diff
@@ -0,0 +1,31 @@
...
+      />
+    );
+
+    expect(mockOnNavigationChange).toHaveBeenCalledTimes(size === 'mobile' ? 1 : 0);
+    expect(wrapper).not.toBeNull();
+    expect(wrapper.findNavigation()).toBeFalsy();
+    expect(wrapper.findNavigationToggle()).toBeFalsy();
+    expect(wrapper.findByClassName(visualRefreshToolbarStyles['main-landmark'])).not.toBeNull();
+    expect(wrapper.findByClassName(visualRefreshToolbarStyles['unfocusable-mobile'])).toBeNull();
+    expect(wrapper.findByClassName(visualRefreshToolbarStyles.content)?.getElement()).toBeVisible();
```
just-boris, line 29: This check does not do much useful, because CSS is not loaded into JSDOM. Under the hood, it checks `getComputedStyle(element).display !== 'none'` which does not work in JSDOM. This assertion will be always passing, even if there are some obstructing styles

I would remove this assertion completely, asserting no `unfocusable-mobile` is enough

just-boris (approved):

PR #2875, opened 2024-10-15
chore: Upgrade axe-core

pan-kot:
### Description

Upgrading axe-core to improve performance in unit tests.

However, one test and one new rule were ignored as now reporting new violations that require a validation. This is to be addressed as a follow-up.

Rel: AWSUI-59822

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/button-dropdown/__tests__/button-dropdown-accessibility.test.tsx
+++ b/src/button-dropdown/__tests__/button-dropdown-accessibility.test.tsx
@@ -177,7 +177,10 @@ it('a11y: default', async () => {
   await expect(container).toValidateA11y();
 });
 
-it('a11y: default, opened', async () => {
+// This fails with AXE aria-required-children version since v4.5.0,
+// See: https://github.com/dequelabs/axe-core/blob/develop/CHANGELOG.md#450-2022-10-17
+// Need to address or exclude the rule.
+it.skip('a11y: default, opened', async () => {
   const { container } = render(<ButtonDropdown items={items}>Open dropdown</ButtonDropdown>);
   const wrapper = createWrapper(container).findButtonDropdown()!;
   wrapper.openDropdown();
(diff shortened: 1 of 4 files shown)


src/button-dropdown/__tests__/button-dropdown-accessibility.test.tsx
```diff
@@ -177,7 +177,10 @@ it('a11y: default', async () => {
   await expect(container).toValidateA11y();
 });
 
-it('a11y: default, opened', async () => {
+// This fails with AXE aria-required-children version since v4.5.0,
+// See: https://github.com/dequelabs/axe-core/blob/develop/CHANGELOG.md#450-2022-10-17
+// Need to address or exclude the rule.
+it.skip('a11y: default, opened', async () => {
```
avinashbot, line 183: I don't like that we're skipping accessibility checks without a plan to re-enable. I can accept disabling `link-in-text-block`, but I feel like we either need to address it now, or treat it as a customer issue level, or it'll stay skipped for a long time.

avinashbot (commented):
I'm okay with skipping the global rule for now, but the component-specific skip is a little worrying to me.

src/button-dropdown/__tests__/button-dropdown-accessibility.test.tsx
avinashbot: > If an element is focusable, or otherwise interactive, user agents MUST ignore the presentation role and expose the element with its implicit role, in order to ensure that the element is [operable](https://www.w3.org/TR/wai-aria-1.2/#dfn-operable).

Oh, interesting. We have an element with a `role="menuitem" tabindex="0"` inside `ul`s and `li`s marked with  `role="presentation"`. This is definitely wrong. Since it's a pre-existing issue, could we turn this into a ticket? If so, I'm good with merging it.

pan-kot: Sure, I already opened a ticket mentioning both issues: AWSUI-59822

avinashbot (approved):

PR #2880, opened 2024-10-16
chore: Fixes flakiness in resizable columns integ test

pan-kot:
### Description

Increasing delays to make sure resize events fire

### How has this been tested?

* Validated by running the test over 100 times in GitHub actions

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/table/__integ__/resizable-columns.test.ts
+++ b/src/table/__integ__/resizable-columns.test.ts
@@ -91,6 +91,7 @@ class TablePage extends BasePageObject {
   async resizeBeyondTableWidth(columnIndex: number) {
     const resizerSelector = tableWrapper.findColumnResizer(columnIndex).toSelector();
     const resizerBox = await this.getBoundingBox(resizerSelector);
+    const { width: windowWidth } = await this.browser.getWindowSize();
     await this.browser.performActions([
       {
         type: 'pointer',
@@ -102,7 +103,7 @@ class TablePage extends BasePageObject {
           { type: 'pointerDown', button: 0 },
           { type: 'pause', duration: 50 },
           // move cursor to screen edge to activate auto-growing behavior
-          { type: 'pointerMove', duration: 10, x: Math.round(resizerBox.left) + 15, y: Math.round(resizerBox.top) },
+          { type: 'pointerMove', duration: 10, x: windowWidth, y: Math.round(resizerBox.top) },
           // pause to let resizing interval fire a few times
           { type: 'pause', duration: 2000 },
           { type: 'pointerUp', button: 0 },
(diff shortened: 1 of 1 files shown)


src/table/__integ__/resizable-columns.test.ts
```diff
@@ -91,26 +91,20 @@ class TablePage extends BasePageObject {
...
-            x: prepareCoordinate(resizerBox.left),
-            y: prepareCoordinate(resizerBox.top),
-          },
+          { type: 'pointerMove', x: Math.round(resizerBox.left), y: Math.round(resizerBox.top) },
           { type: 'pointerDown', button: 0 },
-          { type: 'pause', duration: 30 },
+          { type: 'pause', duration: 50 },
           // move cursor to screen edge to activate auto-growing behavior
-          { type: 'pointerMove', duration: 0, x: windowWidth, y: 0 },
+          { type: 'pointerMove', duration: 10, x: Math.round(resizerBox.left) + 15, y: Math.round(resizerBox.top) },
```
just-boris, line 105: Why was `windowWidth` removed? Now it is a very different test

pan-kot: I removed it as part of my experiments, that also allows to avoid querying the window width.

That does not make the test any different though.

just-boris: The comment says: "move cursor to screen edge". `x: windowWidth` is definitely a screen edge. For `x: Math.round(resizerBox.left) + 15` is it unclear. Even if it coincidentally corresponds to the screen edge , nobody promises you that.

pan-kot: Alright, I revert this bit and re-validate.

src/table/__integ__/resizable-columns.test.ts
```diff
@@ -19,7 +20,6 @@ const wrapper = createWrapper();
 const tableWrapper = wrapper.findTable();
 // All the columns fit in the viewport, which make it easier to test the columns' widths
 const defaultScreen = { width: 1680, height: 800 };
-const prepareCoordinate = (num: number) => Math.round(num);
```
just-boris, line 22: I am also not quite following why you disliked this function and replaced with inline.

For the context, rounding is needed, because WebDriver API expects integer numbers only and fails when a floating value is provided.

just-boris (approved):

src/table/__integ__/resizable-columns.test.ts
just-boris: ok, without that contextual comment it is not very useful, I agree

src/table/__integ__/resizable-columns.test.ts
michaeldowseza: Does the same comment apply to the change to the change in y co-ordinate?

pan-kot: The `y` coordinate should be irrelevant here: only horizontal movement matter. I replaced 0 with resizer coordinate as it represents the expected user interaction better.

PR #2895, opened 2024-10-17
fix: Prevent breadcrumbs from flickering when rendered inside app layout toolbar breadcrumbs slot

just-boris:
### Description

When breadcrumbs are already rendered in the expected slot, there is no need for the runtime API interactions. This allows for a fast track without additional re-renders

### How has this been tested?

1. Updated relevant unit tests
2. Tested the actual flickering locally on the dev page

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/__tests__/global-breadcrumbs.test.tsx
+++ b/src/app-layout/__tests__/global-breadcrumbs.test.tsx
@@ -29,6 +29,10 @@ function findAllBreadcrumbsInstances() {
   return wrapper.findAllByClassName(BreadcrumbGroupWrapper.rootSelector);
 }
 
+function findDiscoveredBreadcrumbs() {
+  return wrapper.findAppLayout()!.find('[data-awsui-discovered-breadcrumbs="true"]');
+}
+
 function findAppLayoutBreadcrumbItems() {
   return wrapper.findAppLayout()!.findBreadcrumbs()!.findBreadcrumbGroup()!.findBreadcrumbLinks();
 }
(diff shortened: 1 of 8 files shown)


src/app-layout/visual-refresh-toolbar/contexts.ts
```diff
@@ -0,0 +1,12 @@
...
+import React from 'react';
+
+import { awsuiPluginsInternal } from '../../internal/plugins/api';
+
+interface BreadcrumbsSlotContextType {
+  isInToolbar: boolean;
+}
+
+export const BreadcrumbsSlotContext =
+  awsuiPluginsInternal.sharedReactContexts.createContext<BreadcrumbsSlotContextType>(React, 'BreadcrumbsSlotContext');
```
just-boris, line 12: This is needed, because by default `React.createContext` produces a new context for each call. In our case, there could be multiple copies of this file, which will not allow context providers connect to its consumers. 

We solve it by caching context instances by their name

src/app-layout/__tests__/global-breadcrumbs.test.tsx
```diff
@@ -108,15 +115,17 @@ describeEachAppLayout({ themes: ['refresh-toolbar'], sizes: ['desktop'] }, () =>
     );
   });
 
-  test('when breadcrumbs are rendered in multiple slots, the last one takes precedence', async () => {
-    await renderAsync(
+  test('when breadcrumbs are rendered in multiple slots, the one in the breadcrumbs slot takes precedence', async () => {
```
just-boris, line 118: Technically, this is a breaking change. But this the new expected one, we should be fine.

src/app-layout/visual-refresh-toolbar/contexts.ts
just-boris: Yes, because `awsuiPluginsInternal` object is deduplicated. https://github.com/cloudscape-design/components/blob/01aecc9aefe0c976f3a586e8ba6917d98540c237/src/internal/plugins/api.ts#L62-L63

jperals (approved):

PR #2897, opened 2024-10-18
fix: Makes button dropdown main action clickable with Enter

pan-kot:
### Description

When a button dropdown is rendered with main action only (for test utils consistency) the main action wasn't clickable with enter. The PR fixes that.

Rel: [baiAA7LWUcov]

### How has this been tested?

* New integration test

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/button-dropdown/__integ__/button-dropdown-events.test.ts
+++ b/src/button-dropdown/__integ__/button-dropdown-events.test.ts
@@ -1,6 +1,8 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
+import { BasePageObject } from '@cloudscape-design/browser-test-tools/page-objects';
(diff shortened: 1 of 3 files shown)


avinashbot (approved):

src/button-dropdown/__integ__/button-dropdown-events.test.ts
```diff
@@ -76,4 +78,21 @@ describe('clicking on a ButtonDropdown item', () => {
       await expect(page.getLocation()).resolves.toEqual(oldLocation);
     })
   );
+  test(
```
jperals, line 81: Since this feature is meant for an internal usage of property filter only, I would be more inclined to test this as part of property filter functionality, rather than making a case for a standalone button dropdown with only main action which in reality we don't expose as a user experience —it is an implementation detail. But this is not a blocker

jperals (approved):

PR #2899, opened 2024-10-18
feat: Add scroll snapping to responsive tabs

timogasda:
### Description

This PR includes two changes to the tabs component:

1. Adds CSS scroll snapping to the responsive tabs, which improves the experience on touch devices and reduces the risks of tab titles being "skipped" when using the left/right navigation buttons.
2. Removed the custom implementation of smooth scrolling, since all our supported browsers can do this natively now.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/tabs/__integ__/tabs.test.ts
+++ b/src/tabs/__integ__/tabs.test.ts
@@ -223,6 +223,7 @@ test(
     await page.setWindowSize({ width: 500, height: 1000 });
     await page.click('#add-tab');
     await page.click(page.paginationButton('right', true));
+    await page.click(page.paginationButton('right', true));
     await page.click(wrapper.findTabLinkByIndex(7).toSelector());
     await page.waitForAssertion(async () =>
       expect(await page.isExisting(page.paginationButton('right', true))).toBe(false)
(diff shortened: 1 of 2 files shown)


src/tabs/smooth-scroll.ts
```diff
@@ -2,62 +2,16 @@
...
-
 const smoothScroll = (element: HTMLElement, to: number) => {
-  if (isMotionDisabled(element)) {
+  if (isMotionDisabled(element) || !element.scrollTo) {
     element.scrollLeft = to;
-    return;
-  }
-  if (isNativeSmoothScrollingSupported() && element.scrollTo) {
+  } else {
+    // istanbul ignore next: unit tests always have motion disabled
```
timogasda, line 9: I don't understand why this is not working properly. If I move the `ignore` comment above line 8, it still complains. I think we might just need to override the coverage check for this PR.

timogasda: (if this becomes a blocker, I'm happy to separate this out into a second PR. I just really need to deliver the scrolling update)

connorlanigan: Have you tried using `/* istanbul ignore else  */`? https://github.com/gotwarlost/istanbul/blob/master/ignoring-code-for-coverage.md#ignore-an-else-path

connorlanigan (approved):

src/tabs/smooth-scroll.ts
timogasda: I didn't, but according to the docs that's for cases where an `else` path might be required/expected but it doesn't exist at all.

PR #2901, opened 2024-10-18
chore: Emit component updated metrics during table interactions

michaeldowseza:
### Description

- Emit component update metrics for table interactions.
- Refactor useComponentAnalytics hook and migrate code into useTableInteraction: We do not have plans to extend this to additional components yet and it was cleaner to combine these hooks as they have the same purpose.

### How has this been tested?

- Added additional unit tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/internal/hooks/use-table-interaction-metrics/index.ts
```diff
@@ -3,8 +3,26 @@
 
 import { useEffect, useRef } from 'react';
 
-import { PerformanceMetrics } from '../../analytics';
+import { ComponentMetrics, PerformanceMetrics } from '../../analytics';
+import { JSONObject } from '../../analytics/interfaces';
 import { useEffectOnUpdate } from '../use-effect-on-update';
+import { useRandomId } from '../use-unique-id';
+
+function useTaskInteractionAttribute(elementRef: React.RefObject<HTMLElement>, value: string) {
```
connorlanigan, line 11: We already have this code in another place:

https://github.com/cloudscape-design/components/blob/56aa09ea8c55009dd7d07f51ef21243da63a29d3/src/internal/hooks/use-performance-marks/index.ts#L13-L27

Can we merge these two into one, to reduce the risk of diverging implementations (e.g. when we make bug fixes)?

src/internal/analytics/interfaces.ts
```diff
@@ -182,12 +182,25 @@ export interface IPerformanceMetrics {
   taskCompletionData: TaskCompletionDataMethod;
 }
 
+type JSONValue = string | number | boolean | null | undefined;
```
connorlanigan, line 185: When the value of an object's property is `undefined`, the property will not appear in the JSONified string at all - is that okay for our metrics? Should we enforce values being `null` instead?

src/internal/hooks/use-table-interaction-metrics/index.ts
```diff
@@ -13,26 +31,41 @@ to be the cause of the current loading state.
...
   const capturedUserAction = useRef<string | null>(null);
   const loadingStartTime = useRef<number | null>(null);
 
-  const metadata = useRef({ itemCount, getComponentIdentifier, interactionMetadata });
-  metadata.current = { itemCount, getComponentIdentifier, interactionMetadata };
+  const metadata = useRef({ itemCount, getComponentIdentifier, getComponentConfiguration, interactionMetadata });
+  metadata.current = { itemCount, getComponentIdentifier, getComponentConfiguration, interactionMetadata };
+
+  useEffect(() => {
+    ComponentMetrics.componentMounted({
```
orangevolon, line 62: With every `taskInteractionId` change, do we wanna track mount?

src/internal/hooks/use-table-interaction-metrics/__tests__/use-table-interaction-metrics.test.tsx
```diff
@@ -1,52 +1,128 @@
...
+  });
+
+  test('should not render the attribute during server-side rendering', () => {
+    const markup = renderToStaticMarkup(<TestComponent />);
+
+    expect(markup).toBe('<div data-testid="element"></div>');
+  });
+
   describe('Interactions', () => {
+    let componentUpdated: jest.Mock;
```
orangevolon, line 105: I don't see assertion against this line anywhere. 🤔

src/internal/hooks/use-table-interaction-metrics/__tests__/use-table-interaction-metrics.test.tsx
```diff
@@ -1,52 +1,128 @@
...
+
+    beforeEach(() => {
+      componentUpdated = jest.fn();
+
+      setComponentMetrics({
+        componentMounted: jest.fn(),
+        componentUpdated,
+      });
+    });
```
orangevolon, line 114: Why do we even need this? we are already resetting the global stubs in line 51. We can assert against the same stub in each test case?

src/internal/hooks/use-table-interaction-metrics/index.ts
michaeldowseza: Fair point. Mine was based off of this one. Will consolidate into a `useDOMAttribute` hook.

src/internal/analytics/interfaces.ts
michaeldowseza: I had mostly tried to align with this [type interface](https://github.com/cloudscape-design/component-toolkit/blob/8ef08a77a38f6c49a2c30751f1fe01f4124345de/src/internal/base-component/metrics/interfaces.ts#L4). I think not appearing in the string is fine.

src/internal/hooks/use-table-interaction-metrics/__tests__/use-table-interaction-metrics.test.tsx
michaeldowseza: Moved variables to the top.

src/internal/hooks/use-table-interaction-metrics/index.ts
orangevolon: I think as long as it's a ref it should be fine, but I guess that also relies on the assumption that `useRandomId` is always going to return a ref.

orangevolon (dismissed):

orangevolon (dismissed):

PR #2906, opened 2024-10-21
fix: Multi layout register

georgylobko:
### Description

The problem occurs when a console has nested App layouts and they are kept mounted in the DOM when navigating through app to different routes. 

The route of the problem is that [multi-layout](https://github.com/cloudscape-design/components/blob/main/src/app-layout/visual-refresh-toolbar/multi-layout.ts) combines props from nested and parent app layouts via `awsuiPluginsInternal.appLayoutWidget.register` api and rely upon nested layout unmount to cleanup unnecessary props for the nested app layout. 

This PR addresses this issue by checking the nested app layout visibility and cleaning up its props if not.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/app-layout/visual-refresh-toolbar/multi-layout.ts
```diff
@@ -69,15 +69,34 @@ function mergeProps(ownProps: SharedProps, additionalProps: ReadonlyArray<Partia
...
-    return awsuiPluginsInternal.appLayoutWidget.register(forceDeduplicationType, props =>
-      setRegistration(props as RegistrationState<SharedProps>)
-    );
-  }, [forceDeduplicationType]);
+    if (!ref.current) {
+      return;
+    }
+    let unregister: () => void | undefined;
+
+    const observer = new IntersectionObserver(([entry]) => {
```
georgylobko, line 82: Our `useIntersectionObserver` hook is not suitable here, because it initially (on the first render) sets visibility state to `false` and then, after first rerender, sets it to true, while here we can initially set the correct value

src/internal/hooks/use-intersection-observer/index.ts
```diff
@@ -3,13 +3,19 @@
...
+  initialState?: boolean;
+}
+
 /**
  * A hook that uses an Intersection Observer on the target element ref
  * and detects if the element is intersecting with its parent.
  */
-export function useIntersectionObserver<T extends HTMLElement>() {
+export function useIntersectionObserver<T extends HTMLElement>({
+  initialState = false,
```
just-boris, line 15: Made this configurable, because there are different use-cases

1. For lazy-loading something you want the initial value to be `false`
2. For suspending some operations when hiding you want the initial value to be `true`

src/app-layout/visual-refresh-toolbar/multi-layout.ts
just-boris: Update: updated the shared hook to make use of it here

src/app-layout/visual-refresh-toolbar/multi-layout.ts
```diff
@@ -69,15 +69,18 @@ function mergeProps(ownProps: SharedProps, additionalProps: ReadonlyArray<Partia
   return Object.keys(toolbar).filter(key => key !== 'ariaLabels').length > 0 ? toolbar : null;
 }
 
-export function useMultiAppLayout(props: SharedProps) {
+export function useMultiAppLayout(props: SharedProps, isEnabled: boolean) {
   const [registration, setRegistration] = useState<RegistrationState<SharedProps> | null>(null);
   const { forceDeduplicationType } = props;
 
   useLayoutEffect(() => {
+    if (!isEnabled) {
```
just-boris, line 77: Yes, not covered in unit tests (because there is no intersection observer API in JSDOM). But there are integration tests, so we are safe

pages/app-layout/multi-layout-iframe.page.tsx
```diff
@@ -17,6 +17,7 @@ import * as toolsContent from './utils/tools-content';
 function InnerApp() {
   return (
     <AppLayout
+      {...{ __disableRuntimeDrawers: true }}
```
just-boris, line 20: Not relevant to the matter of the fix, just suppress warnings in browser console, which distract from debugging

src/__a11y__/a11y-page-object.ts
```diff
@@ -113,7 +113,7 @@ function ariaLevelViolationsFilter(violation: Axe.Result) {
 // There is a known issue when multiple app layout instances rendered on the page. Skip them for pages matching this pattern
 function landmarkViolationFilter(violation: Axe.Result, currentUrl: string) {
   return (
-    !currentUrl.includes('app-layout/multi-layout-simple') ||
+    !currentUrl.includes('app-layout/multi') ||
```
just-boris, line 116: This is a known issue that multiple app layouts cause this landmark violation. Extended the pattern to also cover the new pages

src/app-layout/visual-refresh-toolbar/multi-layout.ts
```diff
@@ -69,15 +69,18 @@ function mergeProps(ownProps: SharedProps, additionalProps: ReadonlyArray<Partia
   return Object.keys(toolbar).filter(key => key !== 'ariaLabels').length > 0 ? toolbar : null;
 }
 
-export function useMultiAppLayout(props: SharedProps) {
+export function useMultiAppLayout(props: SharedProps, isEnabled: boolean) {
```
gethinwebster, line 72: why add this as a new argument rather than to `props`? as far as I see we already have values within props (e.g. `forceDeduplicationType`) that have a similar type of function

just-boris: Because props are "props" they are coming as component configuration `<AppLayout forceDeduplicationType="primary">`. `isEnabled` is a state, so it is natural to keep it separate

gethinwebster (approved):

PR #2914, opened 2024-10-22
chore: Pre-release the live region component API internally

avinashbot:
### Description

*58 files, but they're mostly just import path changes and renames!*

The important changes are in [`src/internal/components/live-region`](https://github.com/cloudscape-design/components/pull/2914/files#diff-0bd914bc4a690dca7f4cdf1732d17728fc131eff66e565d59d38e9a3d37276f9).

- Split LiveRegion into a "public" LiveRegion and an internal "InternalLiveRegion" as we do with other components, along with documented API definitions.
- Update tests that expect `[aria-live]` inside the component.
- There's also another crucial improvement: the aria-live element is rendered in a separate element *at the end of the DOM* so that it doesn't cause screen readers to read live region content twice (once as an announcement, and again as they navigate inside the hidden element). The previous implementation used a centralized singleton to manage announcements globally, but this was backwards incompatible with many consumer's tests (e.g. those that rely on React's lifecycle management and test for text in DOM, like `screen.getByText()`).

There will be a pure path update PR when the component should _actually_ be merged.

### How has this been tested?

Updated unit tests in the live region component, and where components expect a "local" aria-live region.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/components/live-region/index.tsx
+++ b/src/internal/components/live-region/index.tsx
@@ -1,176 +1,31 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-/* eslint-disable @cloudscape-design/prefer-live-region */
+import React from 'react';
 
-import React, { memo, useEffect, useRef } from 'react';
-import clsx from 'clsx';
+// import useBaseComponent from '../../hooks/use-base-component';
+import { applyDisplayName } from '../../utils/apply-display-name';
+import { LiveRegionProps } from './interfaces';
+import InternalLiveRegion from './internal';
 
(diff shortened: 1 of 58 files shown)


src/internal/components/live-region/index.tsx
```diff
@@ -1,176 +1,31 @@
...
-      updateLiveRegion();
-    }
-
-    return () => {
-      if (timeoutId) {
-        clearTimeout(timeoutId);
-      }
-    };
-  });
+function LiveRegion({ assertive = false, hidden = false, tagName = 'div', ...restProps }: LiveRegionProps) {
```
orangevolon, line 13: Do we forward any html attributes to the underlying `div` or `span`, if so I don't think `LiveRegionProps` have the base html props type.

orangevolon (approved):
Clarified the points in person. Thanks!

PR #2921, opened 2024-10-22
feat: Release the LiveRegion component

avinashbot:
### Description

The live region is a non-visual component that can be used to announce page changes to assistive technology, and exposes best practices for [ARIA live regions](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions).

Also includes one crucial change from the sign-off, which is to remove the `message` prop and always use `children` instead. The `sources` prop is still kept for internal use.

### How has this been tested?

Unit tests, dry run builds, and manual checks with internal usages of the component.

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/live-region/controller.ts
```diff
@@ -33,7 +33,6 @@ export class LiveRegionController {
     this._element.className = styles.announcer;
     this._element.setAttribute('aria-live', this.politeness);
     this._element.setAttribute('aria-atomic', 'true');
-    this._element.setAttribute('data-awsui-live-announcer', 'true');
```
johannes-weber, line 36: Not a blocker, out of curiosity: do you remember why we had this data attribute?

johannes-weber: Thanks for explaining! :)

johannes-weber (changes requested):

src/live-region/index.tsx
```diff
@@ -3,17 +3,15 @@
...
 import { LiveRegionProps } from './interfaces';
 import InternalLiveRegion from './internal';
 
 export { LiveRegionProps };
 
 function LiveRegion({ assertive = false, hidden = false, tagName = 'div', ...restProps }: LiveRegionProps) {
-  // TODO: Switch this out when moving this component out of internal
-  // const baseComponentProps = useBaseComponent('LiveRegion');
-  const baseComponentProps = { __internalRootRef: React.useRef() };
+  const baseComponentProps = useBaseComponent('LiveRegion');
```
Al-Dani, line 14: do we want to log any props

src/live-region/internal.tsx
```diff
@@ -106,7 +105,14 @@ export default React.forwardRef(function InternalLiveRegion(
...
 
   return (
-    <TagName ref={mergedRef} {...baseProps} className={clsx(styles.root, className)} hidden={hidden}>
+    <TagName
+      ref={mergedRef}
+      {...baseProps}
+      className={clsx(styles.root, testUtilStyles.root, className)}
+      hidden={hidden}
+      // https://github.com/facebook/react/issues/17157#issuecomment-2127180687
+      {...{ inert: hidden ? '' : undefined }}
```
Al-Dani, line 114: you had a discussion on that topic, what is the outcome? We're keeping it as is?

src/form/__tests__/form.test.tsx
```diff
@@ -7,7 +7,7 @@ import Form, { FormProps } from '../../../lib/components/form';
 import createWrapper from '../../../lib/components/test-utils/dom';
 
 import alertStyles from '../../../lib/components/alert/styles.selectors.js';
-import liveRegionStyles from '../../../lib/components/internal/components/live-region/styles.selectors.js';
+import liveRegionStyles from '../../../lib/components/live-region/styles.selectors.js';
```
Al-Dani, line 10: shouldn't it be?
```suggestion
import liveRegionStyles from '../../../lib/components/live-region/test-classes/styles.selectors.js';
```

src/live-region/index.tsx
avinashbot: I don't see a strong reason for any specific prop, but why not, since I'm fixing other things, I'll just add `assertive` and `hidden` 🤷

src/form/__tests__/form.test.tsx
avinashbot: Yep, you're right, looks like I missed it during find-and-replace.

src/live-region/internal.tsx
avinashbot: Removing this - after the discussion about performance concerns, I added both (initial state won't work on React 19, but the effect will correct it if that happens), but it looks like it was flakiness, so 🤷 

I'll remove it. Less code is better.

Al-Dani (approved):

PR #2923, opened 2024-10-22
Bluedialog Initial commit

laveshv:
### Description

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


pages/bluedialog/bluedialog.page.tsx
```diff
@@ -0,0 +1,98 @@
...
+  const [selectedService, setSelectedService] = React.useState({ label: 'Cloudwatch', value: 'Cloudwatch' });
+
+  function submitData() {
+    console.log({ dateValue, amountValue, descriptionValue, selectedService });
+  }
+  function skipDialog() {
+    setShowDialog(false);
+  }
+
+  return showDialog ? (
```
cansuaa, line 36: Let's not remove the dialog at all, we can console.log instead. I think overall the example here could be simpler.

pages/bluedialog/bluedialog.page.tsx
```diff
@@ -0,0 +1,98 @@
...
+    <div className={styles['blue-dialog-box']}>
```
cansuaa, line 37: I don't see the a11y requirements being addressed.

pages/bluedialog/bluedialog.page.tsx
laveshv: Done in latest revision.

pages/bluedialog/bluedialog.page.tsx
laveshv: I added some a11y specific code. Can you please review?

pages/bluedialog/styles.scss
```diff
@@ -0,0 +1,39 @@
...
+ Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ SPDX-License-Identifier: Apache-2.0
+*/
+@use '~design-tokens' as awsui;
+
+.blue-dialog-box {
+  margin-block: 20px;
+  margin-inline: 20px;
+  border-block: 1px solid awsui.$color-border-status-info;
+  border-inline: 1px solid awsui.$color-border-status-info;
```
cansuaa, line 11: Why not just margin and border? Is margin even needed?

pages/bluedialog/styles.scss
```diff
@@ -0,0 +1,39 @@
...
+.blue-dialog-box__content {
+  padding-block-start: 0;
+  padding-block-end: 20px;
+  padding-inline: 20px;
+}
+
+.blue-dialog-box__footer {
+  border-block-start: 1px solid awsui.$color-border-divider-default;
+  padding-block: 10px;
+  padding-inline: 10px;
```
cansuaa, line 35: Same here

pages/bluedialog/bluedialog.page.tsx
```diff
@@ -0,0 +1,127 @@
...
+    console.log('skip dialog'); // can be used by the parent component to dismiss/ hide this dialog box
+  }
+
+  return (
+    <div
+      className={styles['blue-dialog-box']}
+      role={'dialog'}
+      aria-labelledby="dialog-title"
+      aria-modal="false" // Maintains natural focus flow since it's an inline dialog
+      tabIndex={-1} // This allows the dialog to receive focus
```
cansuaa, line 58: I believe the value should be 0 not -1

pages/bluedialog/bluedialog.page.tsx
```diff
@@ -0,0 +1,127 @@
...
+    >
+      <Form
```
pan-kot, line 60: Should we also use a semantic form element and support onSubmit?

pages/bluedialog/bluedialog.page.tsx
```diff
@@ -0,0 +1,127 @@
...
+  const triggerRef = useRef<HTMLElement | null>(null);
+  // Inputs (these following can be clubbed into one state object)
+  const [dateValue, setDateValue] = React.useState('');
+  const [amountValue, setAmountValue] = React.useState('0');
+  const [descriptionValue, setDescriptionValue] = React.useState('');
+  const [selectedService, setSelectedService] = React.useState({ label: 'Cloudwatch', value: 'Cloudwatch' });
+
+  useEffect(() => {
+    // Store the element that had focus before dialog opened
+    triggerRef.current = document.activeElement as HTMLElement;
```
pan-kot, line 32: If the dialog is opened programmatically this transition can be weird. Would it be better to handle the focus transition on the outside? If the parent component provides onSubmit/onCancel then the focus transition can be done as part of it:

```
function ParentComponent() {
  // ...
  
  function returnFocus() {
    sendButtonRef.current?.focus()
  }
  
  function onSubmit() {
    // submit
    returnFocus();
  }
  
  function onCancel() {
    // cancel
    returnFocus();
  }
  
  return (
    <div>
      // ...
      <Bluedialog onSubmit={onSubmit} onCancel={onCancel} />
    </div>
  );
}
```

pages/bluedialog/bluedialog.page.tsx
pan-kot: -1 means the dialog can only be focused programmatically which is good for one-time focus when it appears. I think making it user-focusable is not needed.

pan-kot: In either case, it would be great to show this focusing behaviour in action.

pages/bluedialog/styles.scss
laveshv: `margin` and `border` are disallowed in the lint settings -> https://github.com/cloudscape-design/components/blob/main/.stylelintrc#L59C39-L59C74

Actually, we can remove the margin. Just that it wasn't looking very good on my whole page. It was spreading on the whole page. So, I introduced it.

pages/bluedialog/styles.scss
laveshv: Refer here -> https://github.com/cloudscape-design/components/blob/main/.stylelintrc#L59C39-L59C74

pages/bluedialog/bluedialog.page.tsx
```diff
@@ -0,0 +1,127 @@
...
+  Header,
+  Input,
+  Select,
+  SpaceBetween,
+  Textarea,
+} from '~components';
+
+import styles from './styles.scss';
+
+export default function Bluedialogbox() {
```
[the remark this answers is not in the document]
laveshv: Actually, the approach that we decided was that we will not create any built-in component for this request. This whole tasks would be just a code snippet that the users can copy and use, with  minimum changes from their end.

Also, I think this folder, might not be the right place for this code. So, please feel free to suggest if there is a better place to put these files

pages/bluedialog/bluedialog.page.tsx
laveshv: Actually we can, but then I did not see the `onSubmit` event in the <Form> component of clouscape, so wanted to navigate away from that practice. Also, it adds some over burden of styling that semantic element.

pages/bluedialog/bluedialog.page.tsx
laveshv: Like I mentioned previously, this is just a code snippet. It is upto the individual users on how they want to use it.

laveshv:
Screenshot 2024-12-19 at 6 01 58 PM
 Attching screenshot for reference

pages/bluedialog/dialog.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+  function submitData() {
+    onSubmit({ feedbackOptions, feedbackText });
+  }
+
+  return (
+    <div
+      className={styles['blue-dialog-box']}
+      role={'dialog'}
+      aria-labelledby="dialog-title"
```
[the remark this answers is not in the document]
laveshv: Yes, I think I forgot to edit it, from previous suggested change. Made the change in new revision.

pages/bluedialog/dialog.tsx
```diff
@@ -0,0 +1,98 @@
...
+      aria-modal="false" // Maintains natural focus flow since it's an inline dialog
+      tabIndex={-1} // This allows the dialog to receive focus
+    >
+      <Form>
+        <div className={styles['blue-dialog-box__content']}>
```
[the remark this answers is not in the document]
laveshv: I thought of using `Box`, but doesn't allow me add css class to it. It gives me more control to use <div> here.

pages/bluedialog/dialog.tsx
```diff
@@ -0,0 +1,98 @@
...
+          <SpaceBetween direction="vertical" size="l">
+            <FormField label="What did you dislike about the response?">
```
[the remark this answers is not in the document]
laveshv: Done in the new Revision. Thanks for pointing it out

pages/bluedialog/dialog.tsx
```diff
@@ -0,0 +1,98 @@
...
+    return () => {
+      if (triggerRef.current) {
+        triggerRef.current.focus();
+      }
+    };
+  }, []);
+
+  function submitData() {
+    onSubmit({ feedbackOptions, feedbackText });
+  }
```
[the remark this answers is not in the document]
laveshv: I was not aware of this use case. Hence avoided it to leave it on the user's choice. I will implement it in the new revision.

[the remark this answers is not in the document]
laveshv: Actually, I am not sure how we want to handle it. This needs to be discussed if there can be a use case where we don't want to implement it, but just want to submit the 'Tell us more' section. Will implement it after discussion.

pages/bluedialog/dialog.tsx
```diff
@@ -0,0 +1,98 @@
...
+export default function Bluedialog({ onSubmit }: any) {
+  const dialogRef = useRef<HTMLDivElement>(null);
+  const triggerRef = useRef<HTMLElement | null>(null);
+  // Inputs (these following can be clubbed into one state object)
+  const [feedbackOptions, setFeedbackOptions] = React.useState({
+    harmful: false,
+    incomplete: false,
+    inaccurate: false,
+    other: false,
+  });
```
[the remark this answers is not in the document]
laveshv: I think this needs to be a multi select, where one or more options can be selected at a time. Hence, I chose to maintain a separate property for each option.

laveshv (commented):
Addressed comments on previous revision

pages/bluedialog/dialog.tsx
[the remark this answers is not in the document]
cansuaa: But "Tell us more" is optional. I believe at least one of the boxes should be checked but we can discuss it.

pages/bluedialog/dialog.tsx
```diff
@@ -0,0 +1,105 @@
...
+  }
+  function skipDialog() {
+    onSkip(); // can be used by the parent component to dismiss/ hide this dialog box
+  }
+
+  return (
+    <div
+      className={styles['blue-dialog-box']}
+      role={'dialog'}
+      aria-labelledby="feedback dialog"
```
cansuaa, line 48: feedback-dialog is not defined anywhere, could you replace this with aria-label?

pages/bluedialog/dialog.tsx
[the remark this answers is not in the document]
cansuaa: Oh right multi choice makes sense in this case.

PR #2928, opened 2024-10-23
feat: Add new app layout test-utils

just-boris:
### Description

Add new test utils in app layout

* `findOpenNavigationPanel` – returns navigation panel only when it is open
* `findOpenToolsPanel` – returns tools panel only when it is open
* `findDrawerTriggerById(id: string, { hasBadge?: boolean; })` returns drawer trigger only when it has a badge

### How has this been tested?

Updated related unit tests to use the new test-utils instead of custom code

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/app-layout/visual-refresh/drawers.tsx
```diff
@@ -356,12 +355,10 @@ export function MobileTriggers() {
               ariaExpanded={item.id === activeDrawerId}
               ariaLabel={item.ariaLabels?.triggerButton}
               ariaControls={activeDrawerId === item.id ? item.id : undefined}
-              className={clsx(
-                `awsui-app-layout-trigger-${item.id}`,
```
just-boris, line 360: Oops, an upscoped random class name. Unused and can be safely removed

src/app-layout/__integ__/awsui-applayout.test.ts
```diff
@@ -51,7 +51,9 @@ describe.each(['classic', 'refresh', 'refresh-toolbar'] as Theme[])('%s', theme
         await expect(page.isDisplayed(wrapper.findToolsToggle().toSelector())).resolves.toBe(true);
       }
       await expect(page.isDisplayed(wrapper.findNavigationClose().toSelector())).resolves.toBe(true);
+      await expect(page.isExisting(wrapper.findOpenNavigationPanel().toSelector())).resolves.toBe(true);
```
pan-kot, line 54: why do you use isExisting and not isDisplayed?

src/app-layout/__tests__/desktop.test.tsx
```diff
@@ -235,8 +233,18 @@ describeEachAppLayout({ sizes: ['desktop'] }, ({ theme }) => {
...
 
-    expect(isDrawerTriggerWithBadge(wrapper, manyDrawers[0].id)).toEqual(true);
-    expect(isDrawerTriggerWithBadge(wrapper, manyDrawers[1].id)).toEqual(false);
+    expect(wrapper.findDrawerTriggerById(manyDrawers[0].id)).toBeTruthy();
+    expect(wrapper.findDrawerTriggerWithBadgeById(manyDrawers[0].id)).toBeTruthy();
+    expect(wrapper.findDrawerTriggerById(manyDrawers[1].id)).toBeTruthy();
+    expect(wrapper.findDrawerTriggerWithBadgeById(manyDrawers[1].id)).toBeFalsy();
+  });
+
+  test('show throw an error when searching for a badge on non-existing drawer', () => {
```
jperals, line 242: Nitpick: double verb ("show throw"), you probably meant just "throw"

src/test-utils/dom/app-layout/index.ts
```diff
@@ -71,6 +87,14 @@ export default class AppLayoutWrapper extends ComponentWrapper {
     return this.find(`.${testutilStyles['drawers-trigger']}[data-testid="awsui-app-layout-trigger-${id}"]`);
   }
 
+  findDrawerTriggerWithBadgeById(id: string): ElementWrapper | null {
```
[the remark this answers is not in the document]
pan-kot: Why not `findDrawerTriggerById(id).findBadge()` ?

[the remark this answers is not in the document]
just-boris: Because "badge" is not a DOM element. In some design variations, it is only a pseudo-element

src/app-layout/__tests__/desktop.test.tsx
just-boris: Nice catch, fixed

src/app-layout/__integ__/awsui-applayout.test.ts
just-boris: Because `isExisting` is cheaper than `isDisplayed` and works for our case. This selector only exists when the panel is open.

* isExisting is [a direct webdriver API call](https://github.com/webdriverio/webdriverio/blob/main/packages/webdriverio/src/commands/element/isExisting.ts#L58)
* isDisplayed however is [a whole lot of Javascript code](https://github.com/webdriverio/webdriverio/blob/main/packages/webdriverio/src/scripts/isElementDisplayed.ts)

If you think symmetry is very important, I can change to `isDisplayed`, but I try to be frugal when possible

pan-kot: No, it is fine like that - thanks for the explanation 👍

src/app-layout/__tests__/desktop.test.tsx
```diff
@@ -235,8 +233,14 @@ describeEachAppLayout({ sizes: ['desktop'] }, ({ theme }) => {
...
-    expect(isDrawerTriggerWithBadge(wrapper, manyDrawers[0].id)).toEqual(true);
-    expect(isDrawerTriggerWithBadge(wrapper, manyDrawers[1].id)).toEqual(false);
+    expect(wrapper.findDrawerTriggerById(manyDrawers[0].id, { hasBadge: true })).toBeTruthy();
+    expect(wrapper.findDrawerTriggerById(manyDrawers[1].id, { hasBadge: false })).toBeTruthy();
+  });
+
+  test('should return null when finding a non-existing drawer with hasBadge condition', () => {
+    const { wrapper } = renderComponent(<AppLayout drawers={manyDrawers} />);
+
+    expect(wrapper.findDrawerTriggerById('non-existing', { hasBadge: true })).toBeNull();
```
pan-kot, line 243: nit: should we add another line and ensure it is null when hasBase is false?

pan-kot (approved):

src/app-layout/__tests__/desktop.test.tsx
just-boris: Codecov was okay without this check, but fine, I added it

src/app-layout/__tests__/desktop.test.tsx
```diff
@@ -235,8 +233,15 @@ describeEachAppLayout({ sizes: ['desktop'] }, ({ theme }) => {
   test('should render badge when defined', () => {
     const { wrapper } = renderComponent(<AppLayout drawers={manyDrawers} />);
 
-    expect(isDrawerTriggerWithBadge(wrapper, manyDrawers[0].id)).toEqual(true);
-    expect(isDrawerTriggerWithBadge(wrapper, manyDrawers[1].id)).toEqual(false);
+    expect(wrapper.findDrawerTriggerById(manyDrawers[0].id, { hasBadge: true })).toBeTruthy();
+    expect(wrapper.findDrawerTriggerById(manyDrawers[1].id, { hasBadge: false })).toBeTruthy();
+  });
+
+  test('should return null when finding a non-existing drawer with hasBadge condition', () => {
```
jperals, line 240: Minor: this test description confused me as it says "when finding..." but it doesn't find it

src/test-utils/dom/app-layout/index.ts
```diff
@@ -15,6 +15,14 @@ export default class AppLayoutWrapper extends ComponentWrapper {
     return this.findByClassName(testutilStyles.navigation)!;
   }
 
+  findOpenNavigationPanel(): ElementWrapper | null {
+    const navigation = this.findNavigation();
+    if (!navigation) {
+      throw new Error('App Layout does not have navigation content');
+    }
```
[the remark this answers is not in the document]
orangevolon: This is certainly more convenient, but is it consistent with the rest of the selectors? Should the builders usually expect a `null` return if something is not found or an exception is also possible?

We only have few other selectors that throw exceptions, but those happen when the argument is out of range.
https://github.com/cloudscape-design/components/blob/3914399081f77ec84e0095d1724f143a9022be9e/src/test-utils/dom/internal/dropdown-host.ts#L118

src/test-utils/dom/app-layout/index.ts
```diff
@@ -39,6 +47,14 @@ export default class AppLayoutWrapper extends ComponentWrapper {
     return this.findByClassName(testutilStyles.tools)!;
   }
 
+  findOpenToolsPanel(): ElementWrapper | null {
```
orangevolon, line 50: **This is non blocking:** Should we generally consider making conditions into parameters?
```tsx
findToolsPanel({ isOpen: true });
```

src/app-layout/__tests__/desktop.test.tsx
just-boris: Renamed to "when searching for..."

jperals (approved):

src/test-utils/dom/app-layout/index.ts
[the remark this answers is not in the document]
just-boris: I think this throw exception is justified. If you use  `findOpenNavigationPanel` against an app layout instance without navigation at all, it is probably a mistake and you want to be informed about it with a thrown error

Removing an error in the future in favor of `null` is a backward compatible change, so let' go with the error for now

src/test-utils/dom/app-layout/index.ts
just-boris: `findTools` does not have `ElementWrapper  | null` in the return value type. `findToolsPanel({ isOpen: true })` may return null. 

If I am going to build this test util API from scratch, I could go your suggestion, but not in the existing code

orangevolon: I don't think we should fix this here either 👍

PR #2932, opened 2024-10-24
feat: Adds extra selectors to the test utils

orangevolon:
### Description

Adds two new test utils selectors to each component wrapper.

1. `find[COMPONENT_NAME]ByTestId(testId: string)`
2. `findAll[COMPONENT_NAME]s()`

Depends on https://github.com/cloudscape-design/test-utils/pull/74 (This is why the Github checks are failing)
Followed by #2944 

Related links, issue #, if available: **Test utils API improvements project**

### How has this been tested?

✅  This PR: #2944 
💹  Copies including the JSDocs and the pluralizations have been reviewed by the content team.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/build-tools/tasks/test-utils.js
+++ b/build-tools/tasks/test-utils.js
@@ -51,7 +51,7 @@ const configs = {
        find${componentName}(selector?: string): ${toWrapper(componentName)} | null;`,
     buildExtraFinderInterfaces: ({ componentName, componentNamePlural }) => `
        /**
-        * Returns the wrappers of all ${componentNamePlural} that match the specified CSS selector.
+        * Returns an array of ${componentName} wrapper that matches the specified CSS selector.
         * If no CSS selector is specified, returns all of the ${componentNamePlural} inside the current wrapper.
         * If no matching ${componentName} is found, returns an empty array.
         *
--- a/src/__tests__/functional-tests/test-utils.test.tsx
+++ b/src/__tests__/functional-tests/test-utils.test.tsx
@@ -11,13 +11,10 @@ import { pascalCase } from 'change-case';
 
 import { Modal } from '../../../lib/components';
 import Button from '../../../lib/components/button';
-import * as dom from '../../../lib/components/test-utils/dom';
-import * as selectors from '../../../lib/components/test-utils/selectors';
+import createWrapperDom, { ElementWrapper as DomElementWrapper } from '../../../lib/components/test-utils/dom';
+import createWrapperSelectors from '../../../lib/components/test-utils/selectors';
 import { getRequiredPropsForComponent } from '../required-props-for-components';
 import { getAllComponents, requireComponent } from '../utils';
-
-const createWrapperDom = dom.default;
-const createWrapperSelectors = selectors.default;
 
 const componentsWithNoExtraFinders = ['top-navigation', 'app-layout'];
 const componentsWithExceptions = ['annotation-context', ...componentsWithNoExtraFinders];
@@ -70,7 +67,7 @@ function getComponentSelectors(componentName: string) {
 
   // The same set of selector functions are present in both dom and selectors.
   // For this reason, looking into DOM is representative of both groups.
-  const wrapperPropsList = Object.keys(dom.ElementWrapper.prototype);
+  const wrapperPropsList = Object.keys(DomElementWrapper.prototype);
 
   // Every component has the same set of selector functions.
   // For this reason, casting the function names into the Alert component.
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,7 +16,6 @@
         "@dnd-kit/sortable": "^7.0.2",
         "@dnd-kit/utilities": "^3.2.1",
         "@juggle/resize-observer": "^3.3.1",
-        "@types/pluralize": "^0.0.33",
         "ace-builds": "^1.34.0",
         "balanced-match": "^1.0.2",
         "clsx": "^1.1.0",
@@ -3948,11 +3947,6 @@
       "version": "2.4.1",
       "dev": true,
       "license": "MIT"
-    },
-    "node_modules/@types/pluralize": {
-      "version": "0.0.33",
-      "resolved": "https://registry.npmjs.org/@types/pluralize/-/pluralize-0.0.33.tgz",
-      "integrity": "sha512-JOqsl+ZoCpP4e8TDke9W79FDcSgPAR0l6pixx2JHkhnRjvShyYiAYw2LVsnA7K08Y6DeOnaU6ujmENO4os/cYg=="
     },
     "node_modules/@types/pngjs": {
       "version": "6.0.5",
--- a/package.json
+++ b/package.json
@@ -33,7 +33,6 @@
     "@dnd-kit/sortable": "^7.0.2",
     "@dnd-kit/utilities": "^3.2.1",
     "@juggle/resize-observer": "^3.3.1",
-    "@types/pluralize": "^0.0.33",
     "ace-builds": "^1.34.0",
     "balanced-match": "^1.0.2",
     "clsx": "^1.1.0",
--- a/src/__tests__/snapshot-tests/__snapshots__/test-utils-wrappers.test.tsx.snap
+++ b/src/__tests__/snapshot-tests/__snapshots__/test-utils-wrappers.test.tsx.snap
@@ -322,7 +322,7 @@ declare module '@cloudscape-design/test-utils-core/dist/dom' {
        findAlert(selector?: string): AlertWrapper | null;
 
        /**
-        * Returns the wrappers of all Alerts that match the specified CSS selector.
+        * Returns an array of Alert wrapper that matches the specified CSS selector.
         * If no CSS selector is specified, returns all of the Alerts inside the current wrapper.
         * If no matching Alert is found, returns an empty array.
         *
@@ -351,7 +351,7 @@ declare module '@cloudscape-design/test-utils-core/dist/dom' {
        findAnchorNavigation(selector?: string): AnchorNavigationWrapper | null;
 
        /**
-        * Returns the wrappers of all AnchorNavigations that match the specified CSS selector.
+        * Returns an array of AnchorNavigation wrapper that matches the specified CSS selector.
         * If no CSS selector is specified, returns all of the AnchorNavigations inside the current wrapper.
         * If no matching AnchorNavigation is found, returns an empty array.
         *
@@ -380,7 +380,7 @@ declare module '@cloudscape-design/test-utils-core/dist/dom' {
        findAnnotation(selector?: string): AnnotationWrapper | null;
 
        /**
-        * Returns the wrappers of all Annotations that match the specified CSS selector.
+        * Returns an array of Annotation wrapper that matches the specified CSS selector.
         * If no CSS selector is specified, returns all of the Annotations inside the current wrapper.
         * If no matching Annotation is found, returns an empty array.
         *
@@ -419,7 +419,7 @@ declare module '@cloudscape-design/test-utils-core/dist/dom' {
        findAreaChart(selector?: string): AreaChartWrapper | null;
 
        /**
-        * Returns the wrappers of all AreaCharts that match the specified CSS selector.
+        * Returns an array of AreaChart wrapper that matches the specified CSS selector.
         * If no CSS selector is specified, returns all of the AreaCharts inside the current wrapper.
         * If no matching AreaChart is found, returns an empty array.
         *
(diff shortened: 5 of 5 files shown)


build-tools/tasks/test-utils.js
```diff
@@ -57,12 +105,22 @@ function generateIndexFileContent(testUtilType, testUtilMetaData) {
...
-          // https://github.com/microsoft/TypeScript/issues/29132
-          return (this as any).findComponent(selector ? appendSelector(selector, rootSelector) : rootSelector, ${wrapperName});
-      };`;
+      return `
+          ElementWrapper.prototype.find${componentName} = function(selector) {
+            const rootSelector = \`.$\{${wrapperName}.rootSelector}\`;
+            return this.findComponent(selector ? appendSelector(selector, rootSelector) : rootSelector, ${wrapperName});
+          };
+
+          ElementWrapper.prototype.findAll${pluralize(componentName)} = function(selector) {
```
just-boris, line 114: How does it work with the Cards, Steps, Tabs components? Does it print `findAllCards` (which is confusing if it is an array of `CardWrapper` or `CardsWrapper`).

If you want my opinion, any auto-plurlalization is a dead end which does not scale. The method should include a the canonical name  in its original form, for example `findCardsAll`

orangevolon: > How does it work with the Cards, Steps, Tabs components?

Some output
```tsx
pluralize('cards')  // cards
pluralize('steps')  // steps
pluralize('tabs')  // tabs
pluralize('canvas')  // canvases
plurazlie('plus')  // pluses
pluralize('watch')  // watches
```

> If you want my opinion, any auto-plurlalization is a dead end which does not scale.

How far do we want to scale this? We have 82 components in the meantime and it covers all of those as well as the components in the future to come. Also looking at [its source code](https://github.com/plurals/pluralize/blob/master/pluralize.js#L269), they have a quite elaborate grammar ruleset for pluralization which should be more than enough for our use case.

> The method should include the canonical name in its original form, for example findCardsAll

This is a good suggestion too. `findAllXX` is closer to RTL's API.

src/__tests__/test-utils-wrapper.test.tsx
```diff
@@ -0,0 +1,16 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import { ElementWrapper as DomElementWrapper } from '../../lib/components/test-utils/dom';
+import { ElementWrapper as SelectorElementWrapper } from '../../lib/components/test-utils/selectors';
```
just-boris, line 4: Did you explore an option of just reading this file and taking it whole as a snapshot?

just-boris: If you heard "snapshot testing is bad", it was about React testing.

For code generation snapshot testing is the correct way to test. We do in fact have some snapshot tests for test-utils already: https://github.com/cloudscape-design/test-utils/blob/main/src/converter/test/__snapshots__/converter.test.ts.snap

So, adding more tests like this would make sense

orangevolon: We should be keeping the contract for TS, so we should expect no API signature change. Aside from the function names, this would only control the comments but it would be sensitive to the whitespace changes.

I think we can get more value if we write assertions for each function rather than a simple snapshot test, but it might be also a bit slower.

e.g.
```tsx
jest.mock('@cloudscape/test-utils')

test.each(functionsInWrapper)(func => {
  func(); 
  expect(testUtils.findAllComponents).toHaveBeenCalledWith(CardWrapper.rootSelector)
})
```

src/__tests__/test-utils-wrapper.test.tsx
```diff
@@ -0,0 +1,99 @@
...
+        testWrapper: new selectors.ElementWrapper('div'),
+        testUtilImport: selectors,
+      },
+    ] as const;
+
+    describe.each(testUtilTypes)('$type wrapper', ({ testWrapper, testUtilImport }) => {
+      const componentNamePluralized = pluralize(componentName);
+
+      test(`find${componentName} calls findComponents with ${componentName}'s root selector`, () => {
+        const WrapperClass = testUtilImport[`${componentName}Wrapper` as 'AlertWrapper'];
```
orangevolon, line 66: Why as `AlertWrapper`? There's nothing special about `AlertWrapper` just that the key is generated dynamically and to keep TS satisfied, we cast it to one of the components arbitrarily. For the purpose of this test file, all of the wrappers must have the exact same behavior.

src/__tests__/test-utils-wrapper.test.tsx
orangevolon: To update this thread:

1. Snapshot tests are not performed on the entire file.
2. I added a few functional tests to make sure that the statically generated files are passing the params correctly.
3. We put the `pluralize.js` out and used a static map instead.

build-tools/utils/pluralize.js
```diff
@@ -0,0 +1,90 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+const pluralizationMap = {
```
just-boris, line 3: I recommend checking this mapping with our tech writer

src/__tests__/test-utils-wrapper.test.tsx
```diff
@@ -0,0 +1,99 @@
...
+    .map(component => exceptions[component] ?? component)
+    .filter(Boolean);
+}
+
+describe('Test utils element wrapper', () => {
+  describe('Snapshot tests', () => {
+    const importPaths = [
+      {
+        type: 'dom',
+        relativePath: '../../lib/components/test-utils/dom/index.js',
```
just-boris, line 29: Let's take a snapshot of `src/test-utils/dom/index.ts` file to have cleaner snapshot (raw typescript instead of compiled JS)

src/__tests__/test-utils-wrapper.test.tsx
```diff
@@ -0,0 +1,99 @@
...
+      },
+      {
+        type: 'selectors',
+        relativePath: '../../lib/components/test-utils/selectors/index.js',
+      },
+    ] as const;
+
+    test.each(importPaths)('$type wrapper stays unchanged', ({ relativePath }) => {
```
just-boris, line 37: I recommend extracting this test separately from the rest

1. Snapshot tests should be in one group with `documenter.test.ts` and others
2. Functional tests may go into `__tests__/test-utils.test.tsx`

src/__tests__/test-utils-wrapper.test.tsx
```diff
@@ -0,0 +1,99 @@
...
+import path from 'path';
+import pluralize from 'pluralize';
+
+import * as components from '../../lib/components';
+import * as dom from '../../lib/components/test-utils/dom';
+import * as selectors from '../../lib/components/test-utils/selectors';
+
+jest.mock('@cloudscape-design/test-utils-core/dom');
+jest.mock('@cloudscape-design/test-utils-core/selectors');
+
```
just-boris, line 13: Are these still needed?

src/__tests__/test-utils-wrapper.test.tsx
```diff
@@ -0,0 +1,99 @@
...
+function getComponentsList() {
```
just-boris, line 14: This function already exists: https://github.com/cloudscape-design/components/blob/20fb84171d38c5ca44959cb5c70d516c1eddf3d2/src/__tests__/utils.tsx#L16C1-L30C1

just-boris: Also, you may find this one useful: https://github.com/cloudscape-design/components/blob/20fb84171d38c5ca44959cb5c70d516c1eddf3d2/src/__tests__/required-props-for-components.ts#L155

src/__tests__/test-utils-wrapper.test.tsx
```diff
@@ -0,0 +1,99 @@
...
+      test(`find${componentName}ByTestId calls findComponent with data-testid appended to ${componentName}'s root selector`, () => {
+        const WrapperClass = testUtilImport[`${componentName}Wrapper` as 'AlertWrapper'];
+
+        testWrapper[`find${componentName}ByTestId` as 'findAlertByTestId']('test-id');
+
+        expect(testUtilImport.ElementWrapper.prototype.findComponent).toHaveBeenCalledWith(
+          `.${WrapperClass.rootSelector}[data-testid="test-id"]`,
+          WrapperClass
+        );
+      });
```
just-boris, line 96:
```js
test('selector utility works', () => {
   const selector = createSelectorWrapper().findAllAlerts().get(1).toSelector()
   expect(document.querySelector(selector)).toBeInDocument();
});
```

src/__tests__/test-utils-wrapper.test.tsx
orangevolon: Removed and replaced with RTL render.

src/__tests__/test-utils-wrapper.test.tsx
orangevolon: Done in https://github.com/cloudscape-design/components/pull/2980

src/__tests__/test-utils-wrapper.test.tsx
orangevolon: Done in https://github.com/cloudscape-design/components/pull/2980

src/__tests__/test-utils-wrapper.test.tsx
orangevolon: The latest change includes this suggestion.

src/__tests__/functional-tests/test-utils.test.tsx
```diff
@@ -7,9 +7,59 @@
...
+import * as dom from '../../../lib/components/test-utils/dom';
+import * as selectors from '../../../lib/components/test-utils/selectors';
+import { getRequiredPropsForComponent } from '../required-props-for-components';
+import { getAllComponents, requireComponent } from '../utils';
+
+const createWrapperDom = dom.default;
+const createWrapperSelectors = selectors.default;
+
+const components = getAllComponents();
+const skipComponents = ['annotation-context', 'top-navigation', 'app-layout', 'modal'];
```
orangevolon, line 22: Still don't know why `top-navigation`, `app-layout` and `modal` are failing. Will continue investigating.

build-tools/tasks/test-utils.js
```diff
@@ -15,24 +16,132 @@ function toWrapper(componentClass) {
 
 const testUtilsSrcDir = path.resolve('src/test-utils');
 const configs = {
+  common: {
+    // These components are not meant to be present in multiple instances in a single app.
+    // For this reason no findAll and findByTestId finders will be generated for them.
+    noExtraFinders: ['AppLayout', 'TopNavigation'],
```
orangevolon, line 22: A bit more context on this: The "root" element of these two components are buried below the top level element:
```tsx
<div data-testid="assiged-by-user">
  <div>
    <div class="awsui-hashed-root-class">
      ...
    </div>
  </div>
</div>
```

So `data-testid` is not passed to the root element with class name hash, but the top `<div />`. This means `findAppLayoutByTestId` will never return any result and might make the users confused.

Since these two components are also not meant to be multiple in a single app, we can skip adding the extra finders (`findAppLayoutByTestId` and `findAllAppLayouts`) for these ones, at least in the meantime.

build-tools/tasks/test-utils.js
```diff
@@ -15,24 +16,132 @@ function toWrapper(componentClass) {
...
+        * If no CSS selector is specified, returns the wrapper of the first ${componentName}.
+        * If no matching ${componentName} is found, returns \`null\`.
+        *
+        * @param {string} [selector] CSS Selector
+        * @returns {${toWrapper(componentName)} | null}
+        */
+       find${componentName}(selector?: string): ${toWrapper(componentName)} | null;`,
+    buildExtraFinderInterfaces: ({ componentName, componentNamePlural }) => `
+       /**
+        * Returns the wrappers of all ${componentNamePlural} that match the specified CSS selector.
```
johannes-weber, line 54: Wording suggestion (not  blocking), I think "an array" is more specific:
```suggestion
        * Returns an array of ${componentNamePlural} wrapper that matches the specified CSS selector.
```

src/__tests__/functional-tests/test-utils.test.tsx
```diff
@@ -7,9 +7,79 @@
 import React from 'react';
 import { render } from 'react-dom';
 import { render as renderTestingLibrary } from '@testing-library/react';
+import { pascalCase } from 'change-case';
 
+import { Modal } from '../../../lib/components';
 import Button from '../../../lib/components/button';
-import createWrapper from '../../../lib/components/test-utils/dom';
+import * as dom from '../../../lib/components/test-utils/dom';
+import * as selectors from '../../../lib/components/test-utils/selectors';
```
johannes-weber, line 15: What's the reason for using the wildcard export instead of the named export?

build-tools/tasks/test-utils.js
orangevolon: That's good suggestion. I just need to double check this later with content 👍

src/__tests__/functional-tests/test-utils.test.tsx
orangevolon: Initially we were iterating through the `dom` and `selectors` object, but with the latest changes we don't need that anymore and we use the default export only. So removing these 👍

johannes-weber (approved):

PR #2936, opened 2024-10-25
chore: Reset panel header spacing and fix make content alignment more consistent

jkuelz:
### Description

Adjusts the panel header padding to match original VR, in order to fix the misalignment of the Q settings icon and the panel close button. Originally the panel header height was reduced for the toolbar so that the headers and close icons could be aligned with the content in the toolbar. This was when the placement of the toolbar was between the panels, which is not the case anymore. This resets the headers to align with the rest of the elements on the page (flashbars, page headers, table headers, etc).
This change also reduces the space below panel headers so it is more uniform, which fixes a visual bug in side nav where control item had too much space around it and sections that were first in the nav list had too much space above them (in both VR and toolbar).

### How has this been tested?

Ran through dev pipeline screenshot tests. Expected screenshot failures can be seen in pipeline approval attempts: 
1. e644f799-c0a2-404e-97f2-a087b7b71a70
2. 989fe3d1-f018-4e29-a5ba-e318c856a4da

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/help-panel/implementation.tsx
+++ b/src/help-panel/implementation.tsx
@@ -35,7 +35,7 @@ export function HelpPanelImplementation({
       baseProps.className,
       styles['help-panel'],
       isToolbar && styles['with-toolbar'],
-      loading && styles.content
+      loading && styles.loading
     ),
   };
   return loading ? (
--- a/src/help-panel/implementation.tsx
+++ b/src/help-panel/implementation.tsx
@@ -31,7 +31,12 @@ export function HelpPanelImplementation({
   const i18n = useInternalI18n('help-panel');
   const containerProps = {
     ...baseProps,
-    className: clsx(baseProps.className, styles['help-panel'], isToolbar && styles['with-toolbar']),
+    className: clsx(
+      baseProps.className,
+      styles['help-panel'],
+      isToolbar && styles['with-toolbar'],
+      loading && styles.content
+    ),
   };
   return loading ? (
     <div {...containerProps} ref={__internalRootRef}>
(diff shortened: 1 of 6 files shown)


src/help-panel/implementation.tsx
```diff
@@ -31,7 +31,12 @@ export function HelpPanelImplementation({
   const i18n = useInternalI18n('help-panel');
   const containerProps = {
     ...baseProps,
-    className: clsx(baseProps.className, styles['help-panel'], isToolbar && styles['with-toolbar']),
+    className: clsx(
+      baseProps.className,
+      styles['help-panel'],
+      isToolbar && styles['with-toolbar'],
+      loading && styles.content
```
georgylobko, line 38: the class name seems not clear if it only applies when `loading` is true. maybe `styles.loading` then?

src/side-navigation/styles.scss
```diff
@@ -94,6 +90,20 @@
   padding-block: 0;
   padding-inline: 0;
   list-style: none;
+
+  @supports selector(:has(*)) {
```
georgylobko, line 94: Do the `@supports` and `@supports not` selectors exclude each other? If so, could we simplify this by keeping only the `@supports not` and removing the additional rules and stylelint suppressions?

jkuelz: It would be preferred to use the `:has` selector here and keep the margins all on the outer `.list-container`, so we don't end up with double margins around sections like we do sometimes today. The `@supports not` is only kept for backwards compatibility, ideally once Safari 15.1, 15.2, and 15.3 supports the `:has` selector, the `@supports not` lines can be removed.

src/help-panel/implementation.tsx
jkuelz: Good call out, I see how this is confusing. Technically the class is used for both, but it is only conditional for loading here because the `.content` class is already provided below in the non-loading version on lines [46/51](https://github.com/cloudscape-design/components/pull/2936/files#diff-64a1630d1728d18eb8064186efc30a870f7ee2e799cd5d7894c7c3f33b09d838R51). However, for clarity sake I created a separate class in the new revision.

georgylobko (approved):

PR #2938, opened 2024-10-25
feat: Updates colors for severity medium and low badges

katiegeorge:
### Description

Text on top of badges with colors `severity medium` and `severity low` were hard to read, and we were receiving negative feedback from users to that point. 

This adds a new yellow color and updates the orange color to be `orange-500`.

Before            |  After
:-------------------------:|:-------------------------:
image  |  image

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


at-susie (dismissed):

style-dictionary/visual-refresh/metadata/color-severity.ts
```diff
@@ -0,0 +1,86 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import { StyleDictionary } from '../../utils/interfaces';
+
+const metadata: StyleDictionary.MetadataIndex = {
+  colorSeverityDarkRed: {
+    description: "Color from the 'dark red' severity palette",
+    public: true,
```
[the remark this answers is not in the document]
katiegeorge: True, we don't have to make it public right now, but I was treating it the same way as the [charts colors](https://github.com/cloudscape-design/components/blob/main/style-dictionary/visual-refresh/metadata/color-charts.ts), where we make both the primitive public, and the semantic token public.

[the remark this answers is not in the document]
katiegeorge: Is there a particular reason why we wouldn't want these to be public?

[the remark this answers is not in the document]
katiegeorge: Updated to leave them out - we can always make them public later 😄

style-dictionary/visual-refresh/colors.ts
```diff
@@ -266,16 +266,6 @@ const tokens: StyleDictionary.ColorsDictionary = {
   colorDropzoneBackgroundHover: { light: '{colorBlue200}', dark: '{colorBlue600}' },
   colorDropzoneTextActive: { light: '{colorGrey550}', dark: '{colorGrey800}' },
   colorDropzoneTextHover: { light: '{colorBlue800}', dark: '{colorWhite}' },
-  colorBackgroundSeverityCritical: '{colorChartsStatusCritical}',
-  colorTextSeverityCritical: { light: '{colorGrey100}', dark: '{colorBlack}' },
-  colorBackgroundSeverityHigh: '{colorChartsStatusHigh}',
-  colorTextSeverityHigh: { light: '{colorGrey100}', dark: '{colorGrey900}' },
-  colorBackgroundSeverityMedium: '{colorChartsStatusMedium}',
-  colorTextSeverityMedium: '{colorGrey900}',
-  colorBackgroundSeverityLow: '{colorChartsStatusLow}',
```
jkuelz, line 275: Since these tokens were assigned to aliases that are themeable (colorChartsStatusX), and the new tokens use different value assignments, this means anyone who previously themed the chart severity colors will get overridden by this change, and need to reassign those in order to see them reflected in the badge component. This could be seen as a feature or a bug/visually breaking change, depending on how you look at it. 

Why are we not updating the severity colors at the source (colorChartStatusX)?

style-dictionary/utils/token-names.ts
```diff
@@ -51,6 +51,7 @@ export type ColorPaletteTokenName =
   | 'colorYellow100'
   | 'colorYellow600'
   | 'colorYellow700'
+  | 'colorYellow750'
```
jkuelz, line 54: Where is this being used? I only see it defined here.

style-dictionary/visual-refresh/color-severity.ts
```diff
@@ -0,0 +1,29 @@
...
+// SPDX-License-Identifier: Apache-2.0
+import { expandColorDictionary } from '../utils';
+import { StyleDictionary } from '../utils/interfaces';
+
+const tokens: StyleDictionary.ColorSeverityDictionary = {
+  colorSeverityDarkRed: { light: '#870303', dark: '#d63f38' },
+  colorSeverityRed: { light: '#ce3311', dark: '#fe6e73' },
+  colorSeverityOrange: { light: '#f89256', dark: '#f89256' },
+  colorSeverityYellow: { light: '#f2cd54', dark: '#f2cd54' },
+  colorSeverityGrey: '{colorGrey550}',
```
jkuelz, line 11: nit: Since we are only using each of these in one place (and they are in this file), and have no intention of exposing these tokens, I think we could consider simplifying by removing this layer. The risk is that we accidentally use one of these internally when we should've used the semantic token and it makes finding dependencies more difficult. 
Not a huge risk. If you think it is helpful to have the color in the name, you can leave it as is. And just add a comment and/or include it in the token metadata specifying that these should not be used directly.

style-dictionary/utils/token-names.ts
katiegeorge: Good catch - was leftover from an earlier version. Will remove

style-dictionary/visual-refresh/colors.ts
katiegeorge: We can't use the `colorChartStatusX` here, since those colors were not meant to have text on top of them. So this purposefully separates the chart severity colors (standalone) with notification severity colors (used with text).

Interesting point about the potential breaking change, though. Any recommendations on how to proceed here?

style-dictionary/visual-refresh/color-severity.ts
katiegeorge: I removed these from the metadata, so I'll add a comment here. I think it could be helpful in the future in case we add other reds / yellows / etc. to the palette to have the two layers.

style-dictionary/visual-refresh/metadata/color-severity.ts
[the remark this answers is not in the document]
abdhalees: not really, it's just safer if they're not used so it's easier for us to change later

style-dictionary/visual-refresh/colors.ts
jkuelz: Investigated offline. The reason chart colors need to remain separate is for a couple reasons: 
1. chart segments are considered graphical objects where the surface area of the background or boundaries of the segments communicate essential information, and
2. while we recommend it, we cannot guarantee that a visible label _and_ value will always be present on charts at all times, especially for line/bar charts.

This means, the chart colors need to meet 3:1 against the background color of containers, according to [WCAG SC 1.4.11](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html). And the new colors being introduced here for low and medium severity (yellow and orange) would not meet the 3:1 contrast ratio with pure white. 
Badges are exempt from this because the boundaries of the badge is not conveying essential information, and thus does not need to meet 3:1 with the background. The only ratio requirements that needs to be met are the badge text with the badge background, which passes for all new colors.

jkuelz (approved):

PR #2952, opened 2024-10-30
fix: Fixes button group item states

pan-kot:
### Description

The PR includes the following fixes:
1. Menu items can be set as loading or disabled
2. Popover feedback can be shown for disabled and loading items

### How has this been tested?

* New unit tests
* Improved test page and adjusted integ tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/button-group/test.page.tsx
+++ b/pages/button-group/test.page.tsx
@@ -3,177 +3,239 @@
 
 import React, { useContext, useState } from 'react';
 
-import Box from '~components/box';
-import Button from '~components/button';
-import ButtonGroup, { ButtonGroupProps } from '~components/button-group';
-import SpaceBetween from '~components/space-between';
-import StatusIndicator from '~components/status-indicator';
+import { Box, Button, ButtonGroup, ButtonGroupProps, Header, SpaceBetween, StatusIndicator } from '~components';
 
 import AppContext, { AppContextType } from '../app/app-context';
+import { enhanceWindow, WindowWithFlushResponse } from '../common/flush-response';
 import ScreenshotArea from '../utils/screenshot-area';
+
+declare const window: WindowWithFlushResponse;
+enhanceWindow();
 
 type PageContext = React.Context<
   AppContextType<{
     dropdownExpandToViewport?: boolean;
+    manualServerMock?: boolean;
   }>
 >;
 
-const likeButton: ButtonGroupProps.Item = {
-  type: 'icon-button',
-  id: 'like',
-  iconName: 'thumbs-up',
-  text: 'Like',
-  popoverFeedback: <StatusIndicator type="success">Liked</StatusIndicator>,
-};
-
-const dislikeButton: ButtonGroupProps.Item = {
-  type: 'icon-button',
-  id: 'dislike',
-  iconName: 'thumbs-down',
-  text: 'Dislike',
-  popoverFeedback: <StatusIndicator type="error">Disliked</StatusIndicator>,
-};
-
-const feedbackGroup: ButtonGroupProps.Group = {
-  type: 'group',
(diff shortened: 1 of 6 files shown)


src/button-group/menu-dropdown-item.tsx
```diff
@@ -60,6 +57,9 @@ const MenuDropdownItem = React.forwardRef(
               ariaExpanded={ariaExpanded}
               className={clsx(testUtilStyles.item, testUtilsClass)}
               iconName="ellipsis"
+              loading={item.loading}
+              loadingText={item.loadingText}
+              disabled={item.disabled}
```
pan-kot, line 62: I created a new unit test for this change but it is also covered with item-permutations. Currently the menu item is rendered the same even when it is set as loading or disabled.

pages/button-group/test.page.tsx
```diff
@@ -3,177 +3,239 @@
 
 import React, { useContext, useState } from 'react';
 
-import Box from '~components/box';
-import Button from '~components/button';
-import ButtonGroup, { ButtonGroupProps } from '~components/button-group';
-import SpaceBetween from '~components/space-between';
-import StatusIndicator from '~components/status-indicator';
+import { Box, Button, ButtonGroup, ButtonGroupProps, Header, SpaceBetween, StatusIndicator } from '~components';
```
orangevolon, line 6: Aren't we encouraging the direct imports because of the treeshaking? If so, wouldn't it be better to keep it as before since our code base is being used as an example.

pages/button-group/test.page.tsx
```diff
@@ -3,177 +3,239 @@
...
-  const onFocusOnMoreActionsButtonClick = () => {
-    ref.current?.focus('more-actions');
+    switch (detail.id) {
+      case 'like':
+      case 'dislike':
+        return syncAction(() => setFeedback(prev => (prev !== detail.id ? (detail.id as 'like' | 'dislike') : 'none')));
+      case 'favorite':
+        return asyncAction(() => setFavorite(prev => !prev));
+      case 'send':
+        return syncAction(() => setCanSend(false));
```
orangevolon, line 198: Do we want the button to disappear after clicking the send, or should we show another state of the same button?
| Before click | After click |
| --- | --- |
| image | image |

src/button-group/__integ__/button-group.test.ts
```diff
@@ -119,7 +120,7 @@ test(
     await page.keys(['Tab']);
     await expect(page.isFocused(likeButton.toSelector())).resolves.toBe(true);
 
-    await page.keys(['ArrowRight', 'ArrowRight']);
+    await page.keys(['ArrowRight', 'ArrowRight', 'ArrowRight', 'ArrowRight']);
```
orangevolon, line 123: np: since your already used `range()` in line `55` maybe we can repeat it here too?

pages/button-group/test.page.tsx
pan-kot: If we want to enforce this I think we should use an ESLint rule. Rn there are different import styles used across dev pages.

pages/button-group/test.page.tsx
pan-kot: The demo features two different behaviours. For the send button we hide it and for the redo button we make it disabled. I think the disappearing button is a very corner case but I think it is still worths having for completeness.

src/button-group/__integ__/button-group.test.ts
pan-kot: I think here it is alright this way because 4 actions is easy to count anyways. I'd use range from 5-6 repeated actions personally.

orangevolon (approved):

PR #2960, opened 2024-10-31
feat: Adds test-id to the grid columns

orangevolon:
### Description

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/grid/interfaces.tsx
```diff
@@ -65,5 +65,11 @@ export namespace GridProps {
      * The number of grid elements to push the column to the right by.
      */
     push?: number | GridProps.BreakpointMapping;
+
+    /**
+     * Test ID of the grid column.
+     * Assigns this value to the `data-testid` attribute of the column's root element.
```
pan-kot, line 71: Should we instead say like: "so that in can be found with findColumnByTestId(testId) test util"? The current one rather describes the internal implementation as we don't expect someone to rely on the attribute in with some custom queries.

src/test-utils/dom/grid/index.ts
```diff
@@ -14,4 +15,17 @@ export default class GridWrapper extends ComponentWrapper<HTMLDivElement> {
   findColumn(columnIndex: number): ElementWrapper<HTMLDivElement> | null {
     return this.find(`.${styles['grid-column']}:nth-child(${columnIndex}) > div`);
   }
+
+  /**
+   * Returns the wrapper of the first column that matches the specified test ID.
+   * Looks for the `data-testid` attribute that is assigned via `gridDefinition` prop.
```
pan-kot, line 21: Let's rather say that is looks for the testId property added to the gridDefinition.

src/grid/__tests__/grid.test.tsx
```diff
@@ -50,4 +50,61 @@ describe('Grid component', () => {
     });
     expect(wrapper.getElement().childElementCount).toBe(2);
   });
+
+  test('assigns data-testid to the columns', () => {
```
pan-kot, line 54: Why do we need this test?

src/grid/interfaces.tsx
orangevolon: We use a template for this prop JSDoc and the problem with mentioning a test util finder here is that: there are interfaces like `Option` where it is being used by different components and they don't necessarily implement the selectors exactly the same way.

> describes the internal implementation as we don't expect someone to rely on the attribute

I think we should expect the customers to do that. Why `data-testid` should be an internal implementation detail? The customers can either query directly or use test utils finders.

To which element `data-testid` will be assigned is an internal detail, but the attribute it self should not be so.

src/grid/__tests__/grid.test.tsx
orangevolon: This is also related to your second comment.
`data-testid` is not  an internal implementation details and we should test that the attribute is assigned correctly and is accessible to the customers.

src/test-utils/dom/grid/index.ts
orangevolon: Like above comments, I don't think we should keep `data-testid` as an internal detail.

PR #2965, opened 2024-11-01
fix: Respect toolsOpen after opening Tools via drawer

Al-Dani:
### Description
Currently, if we have controllable tools and runtime drawers, after clicking on the trigger of the info panel, it ignores updates to the `toolsOpen` value. That happens because `activeDrawerId` is controlled separately and it still points to the tools drawer.

Related links, issue #, if available: AWSUI-57748

### How has this been tested?

- new unit test added

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/__tests__/runtime-drawers.test.tsx
+++ b/src/app-layout/__tests__/runtime-drawers.test.tsx
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 /* eslint simple-import-sort/imports: 0 */
 import React, { useState } from 'react';
-import { act, fireEvent, render } from '@testing-library/react';
+import { act, fireEvent, render, waitFor } from '@testing-library/react';
 import {
   describeEachAppLayout,
   findActiveDrawerLandmark,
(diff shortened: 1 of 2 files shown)


src/app-layout/__tests__/runtime-drawers.test.tsx
```diff
@@ -290,6 +290,37 @@ describeEachAppLayout(({ theme, size }) => {
     expect(onToolsChange).toHaveBeenCalledWith({ open: false });
   });
 
+  test('respect controlled toolsOpen with runtime drawers after clicking on tools drawer', async () => {
```
georgylobko, line 293: This fix also resolves an additional issue with partially controllable tools (`toolsOpen` is provided while `onToolsChange` not), such as in the following case:

```js
<AppLayout tools="Tools content" toolsOpen={false} />
```

Here, clicking the toolbar trigger should not open the tools panel. Let's add a unit test to cover this scenario as well

georgylobko (approved):

PR #2970, opened 2024-11-01
chore: Disable animations until the first user interaction

georgylobko:
### Description

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/__tests__/runtime-drawers.test.tsx
+++ b/src/app-layout/__tests__/runtime-drawers.test.tsx
@@ -980,6 +980,8 @@ describe('toolbar mode only features', () => {
 
       wrapper.findDrawerTriggerById('global-drawer-1')!.click();
 
+      await delay();
+
       expect(globalDrawersWrapper.findDrawerById('global-drawer-1')!.isActive()).toBe(true);
       wrapper.findDrawerTriggerById('global-drawer-1')!.click();
 
--- a/src/app-layout/interfaces.ts
+++ b/src/app-layout/interfaces.ts
@@ -306,7 +306,6 @@ export namespace AppLayoutProps {
     badge?: boolean;
     resizable?: boolean;
     defaultSize?: number;
-    defaultActive?: boolean;
     onResize?: NonCancelableEventHandler<{ size: number }>;
     preserveInactiveContent?: boolean;
   }
--- a/src/app-layout/interfaces.ts
+++ b/src/app-layout/interfaces.ts
@@ -306,6 +306,7 @@ export namespace AppLayoutProps {
     badge?: boolean;
     resizable?: boolean;
     defaultSize?: number;
+    defaultActive?: boolean;
     onResize?: NonCancelableEventHandler<{ size: number }>;
     preserveInactiveContent?: boolean;
   }
--- a/src/app-layout/visual-refresh-toolbar/drawer/global-drawer.tsx
+++ b/src/app-layout/visual-refresh-toolbar/drawer/global-drawer.tsx
@@ -1,6 +1,6 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
-import React, { useRef } from 'react';
+import React, { useEffect, useRef } from 'react';
 import { Transition } from 'react-transition-group';
 import clsx from 'clsx';
 
@@ -24,14 +24,16 @@ interface AppLayoutGlobalDrawerImplementationProps {
   activeGlobalDrawer:
     | (AppLayoutProps.Drawer & { onShow?: NonCancelableEventHandler; onHide?: NonCancelableEventHandler })
     | undefined;
-  firstRender: boolean;
+  animationDisabled?: boolean;
+  onMount?: () => void;
 }
 
 function AppLayoutGlobalDrawerImplementation({
   appLayoutInternals,
   show,
   activeGlobalDrawer,
-  firstRender,
+  animationDisabled,
+  onMount,
 }: AppLayoutGlobalDrawerImplementationProps) {
   const {
     ariaLabels,
@@ -70,7 +72,11 @@ function AppLayoutGlobalDrawerImplementation({
   const size = getLimitedValue(minDrawerSize, activeDrawerSize, maxDrawerSize);
   const lastOpenedDrawerId = drawersOpenQueue.length ? drawersOpenQueue[0] : null;
   const hasTriggerButton = !!activeGlobalDrawer?.trigger;
-  const animationDisabled = firstRender && activeGlobalDrawer?.defaultActive;
+
+  useEffect(() => {
+    onMount?.();
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
 
   return (
     <Transition nodeRef={drawerRef} in={show} appear={show} timeout={0}>
--- a/src/app-layout/visual-refresh-toolbar/drawer/global-drawer.tsx
+++ b/src/app-layout/visual-refresh-toolbar/drawer/global-drawer.tsx
@@ -31,7 +31,7 @@ function AppLayoutGlobalDrawerImplementation({
   appLayoutInternals,
   show,
   activeGlobalDrawer,
-  firstRender = false,
+  firstRender,
 }: AppLayoutGlobalDrawerImplementationProps) {
   const {
     ariaLabels,
--- a/src/app-layout/visual-refresh-toolbar/drawer/global-drawer.tsx
+++ b/src/app-layout/visual-refresh-toolbar/drawer/global-drawer.tsx
@@ -8,7 +8,7 @@ import { InternalButton } from '../../../button/internal';
 import PanelResizeHandle from '../../../internal/components/panel-resize-handle';
 import customCssProps from '../../../internal/generated/custom-css-properties';
 import { getLimitedValue } from '../../../split-panel/utils/size-utils';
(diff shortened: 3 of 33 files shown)


src/app-layout/visual-refresh-toolbar/index.tsx
```diff
@@ -424,12 +427,31 @@ const AppLayoutVisualRefreshToolbar = React.forwardRef<AppLayoutProps.Ref, AppLa
...
 
+    useEffect(() => {
+      // The disabledMotionGlobalClass class is initially applied to the root element.
+      // It disables all animations until the first user interaction,
+      // which prevents dynamic elements intended as "progressive enhancement"
+      // from animating prematurely. This approach ensures animations
+      // are triggered only by user interactions.
+      function enableAnimations() {
+        rootRef.current?.classList.remove(disabledAnimationsGlobalClass);
+        document.removeEventListener('click', enableAnimations);
```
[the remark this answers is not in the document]
georgylobko: 1. Yes, it's fine, since the css class removal happens imperatively, so we don't need to wait the next react update cycle. Tested under multiple conditions manually.
2. Keyboard events trigger click events when called on CTAs (like trigger buttons)
3. Transitions might happen only if animations are enabled, which is not a case here

src/app-layout/visual-refresh-toolbar/index.tsx
```diff
@@ -76,6 +77,8 @@ const AppLayoutVisualRefreshToolbar = React.forwardRef<AppLayoutProps.Ref, AppLa
     const [toolbarState, setToolbarState] = useState<'show' | 'hide'>('show');
     const [toolbarHeight, setToolbarHeight] = useState(0);
     const [notificationsHeight, setNotificationsHeight] = useState(0);
+    const rootRef = useRef<HTMLDivElement>(null);
+    const disabledAnimationsGlobalClass = 'awsui-motion-disabled';
```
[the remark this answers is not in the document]
georgylobko: Why flashbar/alert should be treated differently here? What is the benefit of such an inconsistency?

src/app-layout/visual-refresh-toolbar/index.tsx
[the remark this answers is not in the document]
just-boris: 1. Yes, it seems to work fine for the current setup, but I see some future proof issues. At the very least I suggest to switch from capturing to the bubble mode
2. Ok, got it
3. Imagine this use-case. 1) separate app layout instances 2) you interacted with the first app layout, enabled animations 3) you switched on a different page in SPA, no page reload 3) the app layout on the next page is suffering from the original problem

src/app-layout/visual-refresh-toolbar/index.tsx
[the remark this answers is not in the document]
just-boris: Maybe you are right, but I suggest conservative approach. Only disable the animations we do not want for sure. By default we want to have as many animations as possible

src/app-layout/visual-refresh-toolbar/index.tsx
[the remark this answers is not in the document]
georgylobko: 1. Switching to capture mode sounds reasonable. Addressed
2. -
3. Fair point. Technically I can take into account the current page visibility and add/remove motion disabled class.  But now I have doubts if we even need this animation disabling. I would consider an option to talk with console home to simplify their `navigationOpen` definition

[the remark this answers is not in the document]
just-boris: Summary from the sync today.

3. For this one I still recommend keeping animations only active some some short duration. Mathematically, there is a bigger part of the page lifetime when animations are not needed, then when needed.

it could be something like this

```js
document.addEventListener('click', () => {
 enableAnimations();
 setTimeout(() => disableAnimations(), 250)
}, true);
```

4. For the iframe use-case. We probably should walk up window parents to capture all possible interactions (or maybe there is some convenient API in browsers for that)

```js
while(parent = window.parent) {
   parent.document.addEventListener('click')
}
```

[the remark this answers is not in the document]
georgylobko: Here's what I tried: enabling and disabling animations right after user interaction led to a race condition, making it difficult to determine the correct times for toggling animations during route navigation. Moreover, the following approach doesn't work:

```js
while (parent = window.parent) {
   parent.document.addEventListener('click')
}
```

It requires the use of the postMessage API, which could lead to significant maintenance challenges afterwards.

So, I introduced a different approach. I checked `defaultActive` in drawers and the first user interaction with the navigation component. From my perspective, it simplifies implementation and maintenance

src/app-layout/__tests__/runtime-drawers.test.tsx
```diff
@@ -939,6 +939,8 @@ describe('toolbar mode only features', () => {
 
       wrapper.findDrawerTriggerById('global-drawer-1')!.click();
 
+      await delay();
```
georgylobko, line 942: the test does not work without this condition anymore, because now setting `openDrawersHistory` works asynchronously

src/app-layout/visual-refresh-toolbar/drawer/global-drawers.tsx
```diff
@@ -29,12 +29,13 @@ export function AppLayoutGlobalDrawersImplementation({
             (drawer.preserveInactiveContent && openDrawersHistory.current.has(drawer.id))
         )
         .map(drawer => {
-          openDrawersHistory.current.add(drawer.id);
+          setTimeout(() => openDrawersHistory.current.add(drawer.id), 0);
```
georgylobko, line 32: I'm not 100% sure if it's the best possible way to implement it. the point of this condition is to add drawerId to the open history only after first render. Technically it's possible to expose a callback like `onMount` from `AppLayoutGlobalDrawer` and set value inside of it

src/app-layout/interfaces.ts
```diff
@@ -306,6 +306,7 @@ export namespace AppLayoutProps {
     badge?: boolean;
     resizable?: boolean;
     defaultSize?: number;
+    defaultActive?: boolean;
```
just-boris, line 309: This interface is our public API. We should not leak such things here.

src/app-layout/visual-refresh-toolbar/drawer/global-drawer.tsx
```diff
@@ -69,6 +73,11 @@ function AppLayoutGlobalDrawerImplementation({
   const lastOpenedDrawerId = drawersOpenQueue.length ? drawersOpenQueue[0] : null;
   const hasTriggerButton = !!activeGlobalDrawer?.trigger;
 
+  useEffect(() => {
+    onMount?.();
```
just-boris, line 77: It looks like a literal implementation of this anti-pattern: https://react.dev/learn/you-might-not-need-an-effect#notifying-parent-components-about-state-changes

src/app-layout/visual-refresh-toolbar/drawer/local-drawer.tsx
```diff
@@ -20,9 +20,13 @@ import styles from './styles.css.js';
 
 interface AppLayoutDrawerImplementationProps {
   appLayoutInternals: AppLayoutInternals;
```
just-boris, line 22: Not a blocker for this PR, but there is an API design issue. We are not supposed to pass down `appLayoutInternals` more than one level deep. Compare with this example: https://github.com/cloudscape-design/components/blob/e57c945bbfb8b04840ec4af904e72e86f6866378/src/app-layout/visual-refresh-toolbar/toolbar/index.tsx#L213-L227

When passing `appLayoutInternals` all over the place, it makes refactoring impossible, because you cannot track what is used where and how

src/app-layout/visual-refresh-toolbar/skeleton/index.tsx
```diff
@@ -39,6 +39,7 @@ interface SkeletonLayoutProps
   bottomSplitPanel?: React.ReactNode;
   globalTools?: React.ReactNode;
   globalToolsOpen?: boolean;
+  navigationAnimationEnabled?: boolean;
```
just-boris, line 42: Why there is a different approach for tools and navigation panel? How is this scaling? What about split panel? Why did it not receive any fix at all?

src/app-layout/visual-refresh-toolbar/drawer/global-drawer.tsx
georgylobko: addressed

src/app-layout/visual-refresh-toolbar/skeleton/index.tsx
georgylobko: addressed

src/app-layout/visual-refresh-toolbar/split-panel/index.tsx
```diff
@@ -12,17 +12,19 @@ interface AppLayoutSplitPanelDrawerSideImplementationProps {
   appLayoutInternals: AppLayoutInternals;
   splitPanelInternals: SplitPanelProviderProps;
   children: React.ReactNode;
+  animationDisabled?: boolean;
```
just-boris, line 15: Let's avoid adding more properties on these widget-non-widget boundaries. If we need to pass something down, it should go via `appLayoutInternals` (or ideally find a way to avoid this, for example, applying a class name on a DOM element above that)

just-boris (approved):

PR #2976, opened 2024-11-04
feat: Adds test id to cards

orangevolon:
### Description

Adds `testId` extractor function to cards inside the card component. Unlike other components, cards can't have a single `testId` prop as the `items` are generic with no base types. For this reason we use a function to extract test id for every individual item.

Collective PR: https://github.com/cloudscape-design/components/pull/2985
Project: Improving test utils API

### How has this been tested?

Tests added to cover the change.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/test-utils/dom/cards/index.ts
+++ b/src/test-utils/dom/cards/index.ts
@@ -1,6 +1,7 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 import { ComponentWrapper, ElementWrapper } from '@cloudscape-design/test-utils-core/dom';
+import { escapeSelector } from '@cloudscape-design/test-utils-core/utils';
 
 import CollectionPreferencesWrapper from '../collection-preferences';
 import ContainerWrapper from '../container';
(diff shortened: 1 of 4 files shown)


src/test-utils/dom/cards/index.ts
```diff
@@ -30,6 +31,19 @@ export class CardWrapper extends ComponentWrapper {
     return this.findAllByClassName(styles.section).map(c => new CardSectionWrapper(c.getElement()));
   }
 
+  /**
+   * Returns the wrapper of the first card section that matches the specified test ID.
+   * Looks for the `data-testid` attribute that is assigned via `sections.testId` prop.
```
avinashbot, line 36: Minor, since I don't want to derail all the other PRs: why do we even have this line? From a builder perspective, all they care about is that they provided a `testId` and this util finds it based on what you provided. Why would I need to care about the fact that it's using `data-testid`?

avinashbot (approved):
Approved, but left a non-blocking doc comment.

src/test-utils/dom/cards/index.ts
orangevolon: `data-testid` is now part of our public API. This is for keeping the compatibility with RTL and other tools that look for `data-testid`.

This is actually one of the reasons why we decided to use this attribute rather than internal ones like `data-awsui-id` or others.

You can check the conversation in [this PR](https://github.com/cloudscape-design/components/pull/2960/files) or comments in related API proposal.

PR #2977, opened 2024-11-04
fix: S3 resource selector findModal selector test util returning incorrect modal

michaeldowseza:
### Description

When rendering an S3 Resource Selector with a regular Modal using the visible property set to false the test util selector would return the first Modal it finds. This change makes the test util selector more specific to be able to find the associated S3 Modal.

E.g.
```
<Modal visible={false} />
<S3ResourceSelector {...s3ResourceSelectorProps} />
```

### How has this been tested?

Modified a dev page to include a Modal and checked before and after the change. Since its just a once off fix, it does not make sense to add this in as a regression test. This can better be tested in a package like demos where various components are tested together and the test utils are used from a customer's perspective.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/s3-resource-selector/s3-modal/styles.scss
+++ b/src/s3-resource-selector/s3-modal/styles.scss
@@ -5,10 +5,6 @@
 @use '../../internal/styles/tokens.scss' as awsui;
 @use '../../internal/styles' as styles;
 
-.modal-root {
-  /* used in test-utils */
-}
-
 .modal-actions {
   justify-content: flex-end;
 }
--- a/src/s3-resource-selector/s3-modal/styles.scss
+++ b/src/s3-resource-selector/s3-modal/styles.scss
@@ -5,6 +5,10 @@
 @use '../../internal/styles/tokens.scss' as awsui;
 @use '../../internal/styles' as styles;
 
+.modal-root {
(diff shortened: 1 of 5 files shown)


src/s3-resource-selector/s3-modal/styles.scss
```diff
@@ -5,6 +5,10 @@
 @use '../../internal/styles/tokens.scss' as awsui;
 @use '../../internal/styles' as styles;
 
+.modal-root {
+  /* used in test-utils */
+}
```
just-boris, line 10: You added this test-utility in a file which may generate unstable selectors when we refactor the code.

Did you want to put it under a new file `src/s3-resource-selector/test-classes/styles.scss` which is going to be a stable location for all utils moving forward?

michaeldowseza: First time seeing the pattern. Do we want to move all the test util classes here as part of this PR or follow up PR?

just-boris: Well,

image

I think our current strategy is to make sure all new classes get added according to this convention. Updating past selectors is something that @orangevolon is looking at.

PR #2986, opened 2024-11-06
chore: Allow drawer content to use full height

jkuelz:
### Description

This is a follow up from https://github.com/cloudscape-design/components/pull/2924 that addresses double scrollbar issue in Q chat panel (AWSUI-59823).
It also addresses an issue in Firefox where side nav would have a scrollbar when there was no side nav header present, the original reason for revert. The other issue regarding flashing scrollbar has been extracted from this PR.

Related links, issue #, if available: AWSUI-59823, TT: V1549419424, TT: D169004346, 

### How has this been tested?
Related screenshot tests added in CR-158701022
Ran through dev pipeline screenshot approvals: 
- 1st: 833b9bfc-f87e-462e-b8e9-3d6f48d73c2a
- 2nd: 42b2f287-623e-4d54-ba8d-29edf66bea4d

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/__integ__/awsui-applayout.test.ts
+++ b/src/app-layout/__integ__/awsui-applayout.test.ts
@@ -189,7 +189,7 @@ describe.each(['classic', 'refresh', 'refresh-toolbar'] as Theme[])('%s', theme 
     })
   );
 
-  test.only(
+  test(
     'Side nav does not display a scrollbar when there is no header',
     setupTest({ pageName: 'with-drawers-scrollable' }, async page => {
       const navBefore = await page.getNavPosition();
--- a/src/app-layout/__integ__/awsui-applayout.test.ts
+++ b/src/app-layout/__integ__/awsui-applayout.test.ts
@@ -19,7 +19,7 @@ class AppLayoutPage extends BasePageObject {
   }
 
   getNavPosition() {
-    return this.getBoundingBox(wrapper.findNavigation().findSideNavigation().findHeaderLink().toSelector());
+    return this.getBoundingBox(wrapper.findNavigation().findSideNavigation().findItemByIndex(1).toSelector());
   }
 
   getContentPosition() {
@@ -186,6 +186,15 @@ describe.each(['classic', 'refresh', 'refresh-toolbar'] as Theme[])('%s', theme 
       ).resolves.toBe(0);
       const { top: contentTop } = await page.getBoundingBox('[data-testid="content-root"]');
       expect(contentTop).toEqual(expectedOffset);
+    })
+  );
+
+  test.only(
+    'Side nav does not display a scrollbar when there is no header',
+    setupTest({ pageName: 'with-drawers-scrollable' }, async page => {
+      const navBefore = await page.getNavPosition();
+      await page.elementScrollTo(wrapper.findNavigation().toSelector(), { top: 100 });
+      await expect(page.getNavPosition()).resolves.toEqual(navBefore);
     })
   );
 });
--- a/src/app-layout/drawer/index.tsx
+++ b/src/app-layout/drawer/index.tsx
@@ -123,7 +123,7 @@ export const Drawer = React.forwardRef(
             aria-label={mainLabel}
             aria-hidden={!isOpen}
             style={{
-              ['blockSize']: `calc(100vh - ${(topOffset || 0) + (bottomOffset || 0)}px)`,
+              blockSize: `calc(100vh - ${(topOffset || 0) + (bottomOffset || 0)}px)`,
             }}
           >
             {!isMobile && isOpen && <div className={styles['resize-handle-wrapper']}>{resizeHandle}</div>}
--- a/src/app-layout/drawer/index.tsx
+++ b/src/app-layout/drawer/index.tsx
@@ -119,9 +119,12 @@ export const Drawer = React.forwardRef(
         >
           {!isMobile && !hideOpenButton && regularOpenButton}
           <TagName
-            className={clsx(resizeHandle && styles['drawer-resize-content'])}
+            className={clsx(resizeHandle && styles['drawer-resize-content'], styles['drawer-content-wrapper'])}
             aria-label={mainLabel}
             aria-hidden={!isOpen}
+            style={{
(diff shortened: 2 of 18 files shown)


src/app-layout/__integ__/awsui-applayout.test.ts
```diff
@@ -188,4 +188,13 @@ describe.each(['classic', 'refresh', 'refresh-toolbar'] as Theme[])('%s', theme
       expect(contentTop).toEqual(expectedOffset);
     })
   );
+
+  test.only(
```
just-boris, line 192:
```suggestion
  test(
```

src/app-layout/drawer/index.tsx
```diff
@@ -119,9 +119,12 @@ export const Drawer = React.forwardRef(
         >
           {!isMobile && !hideOpenButton && regularOpenButton}
           <TagName
-            className={clsx(resizeHandle && styles['drawer-resize-content'])}
+            className={clsx(resizeHandle && styles['drawer-resize-content'], styles['drawer-content-wrapper'])}
             aria-label={mainLabel}
             aria-hidden={!isOpen}
+            style={{
+              ['blockSize']: `calc(100vh - ${(topOffset || 0) + (bottomOffset || 0)}px)`,
```
just-boris, line 126: Same, but cleaner/shorter

```suggestion
              blockSize: `calc(100vh - ${(topOffset || 0) + (bottomOffset || 0)}px)`,
```

src/app-layout/visual-refresh-toolbar/navigation/styles.scss
```diff
@@ -28,6 +28,10 @@
     display: none;
   }
 
+  &.is-navigation-open > .content-container {
+    display: contents;
```
just-boris, line 32: When applying `display: contents` to an element, all other styles applied to this element are ignored

Screenshot 2024-11-06 at 12 15 37

Can we clean up the currently unused styles (if they do nothing) or reassign them to different elements (if they still have some use)

just-boris: P.S. If `display: contents` does the job, it seems like we can simply remove that DOM node (because this is what `display: contents` effectively does)

src/app-layout/drawer/styles.scss
```diff
@@ -58,6 +58,9 @@
     block-size: 100%;
     position: relative;
   }
+  > .drawer-content-wrapper[aria-hidden='false'] {
```
just-boris, line 61: I did not find any condition when `aria-hidden=true` is applied on this selector. Does it mean we can remove that part of the selector?

just-boris (dismissed):
Approved, but can be simplified further

src/app-layout/__integ__/awsui-applayout.test.ts
jkuelz: whoops forgot to remove this from testing, thanks

src/app-layout/drawer/styles.scss
jkuelz: Interesting, that feels unexpected because `aria-hidden` is set to `!isOpen` [here](https://github.com/cloudscape-design/components/blob/main/src/app-layout/drawer/index.tsx#L124). I would've expected it to change when the drawer is opened/closed, like it does for help panel and side nav. 

It appears that `isOpen` always returns true because it is [hardcoded this way](https://github.com/cloudscape-design/components/blob/main/src/app-layout/classic.tsx#L550). Not sure exactly when or what changed to make the `isOpen` condition obsolete. But I have removed this condition from the style.

src/app-layout/visual-refresh-toolbar/navigation/styles.scss
jkuelz: We need that node to prevent the text from scrolling in when the drawer is opened with motion. Updated to display flex instead of contents.

src/app-layout/__integ__/awsui-applayout.test.ts
```diff
@@ -188,4 +188,13 @@ describe.each(['classic', 'refresh', 'refresh-toolbar'] as Theme[])('%s', theme
       expect(contentTop).toEqual(expectedOffset);
     })
   );
+
+  test(
+    'Side nav does not display a scrollbar when there is no header',
+    setupTest({ pageName: 'with-drawers-scrollable' }, async page => {
+      const navBefore = await page.getNavPosition();
+      await page.elementScrollTo(wrapper.findNavigation().toSelector(), { top: 100 });
```
just-boris, line 196: Isn't this going to fail when I merge this one: https://github.com/cloudscape-design/browser-test-tools/pull/118

src/app-layout/visual-refresh-toolbar/navigation/styles.scss
```diff
@@ -20,6 +20,7 @@
   overscroll-behavior-y: contain;
   word-wrap: break-word;
   pointer-events: auto;
+  display: flex;
```
just-boris, line 23: Why is this needed?

The default is `flex-direction: row` and I do not think we want this

src/app-layout/visual-refresh/navigation.scss
```diff
@@ -93,6 +93,7 @@ nav.navigation {
   word-wrap: break-word;
   pointer-events: auto;
   border-inline-end: solid awsui.$border-divider-section-width awsui.$color-border-divider-default;
+  display: flex;
```
just-boris, line 96: Same question

just-boris (approved):

src/app-layout/__integ__/awsui-applayout.test.ts
jkuelz: No, it shouldn't. I made it fail with the scrollbar to ensure it was working properly.

src/app-layout/visual-refresh-toolbar/navigation/styles.scss
jkuelz: There is only one child, so it's not actually aligning anything horizontally or vertically. But I can add a more explicit definition.

just-boris: Thank makes it easier to follow, thanks

just-boris (approved):

PR #2988, opened 2024-11-06
feat: Adds test id to the anchor list component

orangevolon:
### Description

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/test-utils/dom/anchor-navigation/index.ts
```diff
@@ -32,6 +32,18 @@ export default class AnchorNavigationWrapper extends ComponentWrapper {
...
+
+  /**
+   * Returns the wrapper of the first anchor that matches the specified test ID.
+   * Looks for the `data-testid` attribute that is assigned via `anchors` prop.
+   * If no matching anchor is found, returns `null`.
+   *
+   * @param {string} testId
+   * @returns {ElementWrapper<HTMLAnchorElement> | null}
+   */
+  findAnchorByTestId(testId: string): ElementWrapper<HTMLAnchorElement> | null {
```
georgylobko, line 44: We expect anchor element as the return type, but it's assigned to `li`. Either return type or data-testid assignment should be adjusted

orangevolon: Good callout, it actually needs to change a bit more to return the correct wrapper as well. 
To do that I added a new selector css class. It should be pointing to "Anchor Item" now rather than "Anchor Link" now. This is consistent with `findAnchorByIndex` and `findAnchors`.

PR #2998, opened 2024-11-11
fix: Sticky header offset

georgylobko:
### Description

Keep table's sticky header within a nested AppLayout on scroll by resetting `globalVars.stickyVerticalTopOffset` and `globalVars.stickyVerticalBottomOffset` for nested layout.

Tested in my dev pipeline, everything seems expected.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/app-layout/visual-refresh-toolbar/index.tsx
```diff
@@ -431,9 +431,13 @@ const AppLayoutVisualRefreshToolbar = React.forwardRef<AppLayoutProps.Ref, AppLa
...
           style={{
-            [globalVars.stickyVerticalTopOffset]: `${verticalOffsets.header}px`,
-            [globalVars.stickyVerticalBottomOffset]: `${placement.insetBlockEnd}px`,
             paddingBlockEnd: splitPanelOpen && splitPanelPosition === 'bottom' ? splitPanelReportedSize : '',
+            ...(hasToolbar
+              ? {
+                  [globalVars.stickyVerticalTopOffset]: `${verticalOffsets.header}px`,
+                  [globalVars.stickyVerticalBottomOffset]: `${placement.insetBlockEnd}px`,
+                }
+              : {}),
```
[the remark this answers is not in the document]
georgylobko: > Why with a toolbar the sticky position is not needed anymore?

It's needed, but when the top offset is added for a nested app layout, it add this offset twice, because this calculation is based on the global header, which results to doubled offset

> it doesn't seem that removing this condition would have any effect on the sticky behavior.

This is what is happens when this condition is removed:
image
And this is what it's present:
image

Could you share the exact steps you did when removing this condition did not change the behavior?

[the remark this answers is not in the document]
orangevolon: Alright, I dug a bit deeper. The reason seems to be that `useAppLayoutToolbarEnabled` doesn't return `true` so what I'm getting in my dev env is not `<ToolbarAppLayout />` but `<RefreshedAppLayout />`:

https://github.com/cloudscape-design/components/blob/31cd27dc010abe8a73286baf449f499cc7d46360/src/app-layout/internal.tsx#L16

The reason is that `getGlobalFlag` here 👇  is `undefined`.

https://github.com/cloudscape-design/components/blob/31cd27dc010abe8a73286baf449f499cc7d46360/src/app-layout/utils/feature-flags.ts#L9

orangevolon (approved):

src/app-layout/__integ__/awsui-applayout.test.ts
```diff
@@ -197,4 +197,15 @@ describe.each(['classic', 'refresh', 'refresh-toolbar'] as Theme[])('%s', theme
       await expect(page.getNavPosition()).resolves.toEqual(navBefore);
     })
   );
+
+  testIf(theme === 'refresh-toolbar')(
```
just-boris, line 201: Can we run it for other designs too?

just-boris (approved):

src/app-layout/__integ__/awsui-applayout.test.ts
georgylobko: No, VR and classic have an additional gap that prevents the header from becoming sticky immediately. We’d need additional conditions for them, which isn’t the purpose of this test. See the attached video for reference.

https://github.com/user-attachments/assets/6d03be33-cecd-4d2a-9dd8-c27ef7d5265d

just-boris: I think we could still run the test, just have a different assertions for each mode. But up to you

PR #3001, opened 2024-11-11
feat: Files

katiegeorge:
**Do not merge**


src/file-upload/internal.tsx
```diff
@@ -134,7 +135,8 @@ function InternalFileUpload(
             multiple={multiple}
             onChange={event => handleFilesChange(event.detail.value)}
             value={value}
-            {...restProps}
+            ariaLabelledby={restProps.ariaLabelledby}
+            controlId={restProps.controlId}
```
pan-kot, line 139: What is the reason of having these 2 props passed down explicitly?

src/file-token-group/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+import InternalFileTokenGroup from './internal';
+
+export { FileTokenGroupProps };
+
+const FileTokenGroup = ({ showFileLastModified, showFileSize, showFileThumbnail, ...props }: FileTokenGroupProps) => {
+  const baseComponentProps = useBaseComponent('FileTokenGroup', {
+    props: {
+      showFileLastModified,
+      showFileSize,
+      showFileThumbnail,
```
pan-kot, line 17: Let's also add alignment and limit as props and something like useFilesLoading as metadata.

src/file-input/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { applyDisplayName } from '../internal/utils/apply-display-name';
+import { FileInputProps } from './interfaces';
+import InternalFileInput from './internal';
+
+export { FileInputProps };
+
+const FileInput = React.forwardRef(({ multiple, ...props }: FileInputProps, ref: React.Ref<FileInputProps.Ref>) => {
+  const baseComponentProps = useBaseComponent('FileInput', {
+    props: {
+      multiple,
```
pan-kot, line 15: let's also add variant

src/file-dropzone/index.tsx
```diff
@@ -0,0 +1,18 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import React from 'react';
+
+import useBaseComponent from '../internal/hooks/use-base-component';
+import { applyDisplayName } from '../internal/utils/apply-display-name';
+import { FileDropzoneProps } from './interfaces';
+import InternalFileDropzone from './internal';
+import { useFilesDragging } from './use-files-dragging';
```
pan-kot, line 9: Can we add metrics instrumentation to useFilesDragging (as a follow-up, I'd say)?

pages/file-input/simple.page.tsx
```diff
@@ -0,0 +1,40 @@
...
+// SPDX-License-Identifier: Apache-2.0
+import React, { useState } from 'react';
+
+import { Box, ColumnLayout, FileInput } from '~components';
+
+export default function DateInputScenario() {
+  const [files, setFiles] = useState<File[]>([]);
+  const [files2, setFiles2] = useState<File[]>([]);
+
+  return (
```
pan-kot, line 11: This page and the one for dropzone do no include screenshot area. I recommend we add it at least to the dropzone and use for screenshot tests.

pages/file-dropzone/container.page.tsx
```diff
@@ -0,0 +1,120 @@
...
+          }
+        >
+          <SpaceBetween size="l">
+            {/* {!onlyVisibleOnDrag && (
+              <FileDropzone onChange={(event: any) => handleFilesChange(event.detail.value)}>
+                <Box fontWeight="bold" color="inherit">
+                  Drop files here
+                </Box>
+              </FileDropzone>
+            )} */}
```
pan-kot, line 64:
```suggestion
```

pages/file-dropzone/container.page.tsx
```diff
@@ -0,0 +1,120 @@
...
+                  },
+                ]}
+                enableKeyboardNavigation={true}
+                ariaLabels={{
+                  selectionGroupLabel: 'group label',
+                  allItemsSelectionLabel: ({ selectedItems }) => `${selectedItems.length} item selected`,
+                  itemSelectionLabel: ({ selectedItems }, item) =>
+                    `${item.name} is ${selectedItems.indexOf(item) < 0 ? 'not ' : ''}selected`,
+                }}
+                items={files.map(file => ({
```
pan-kot, line 106: Why this mapping is needed? It will be the same if the files array is passed as is.

pages/file-dropzone/container.page.tsx
```diff
@@ -0,0 +1,120 @@
...
+  const { urlParams, setUrlParams } = useContext(AppContext as DemoContext);
+  const { onlyVisibleOnDrag } = urlParams;
+
+  const handleFilesChange = (newFiles: File[]) => {
+    const newValue = [...files, ...newFiles];
+    setFiles(newValue);
+  };
+
+  const removeFiles = () => {
+    const newValue = files.filter(file => !selectedItems.map((item: any) => item.name).includes(file.name));
```
pan-kot, line 29: The files names might not be unique. It is better to compare the files by reference. If you pass files as table items you can just do includes(). Otherwise, you can include file as a property of an item and do `files.filter(file => !selectedItems.some(item => item.file === file))`

pages/file-dropzone/container.page.tsx
katiegeorge: Old commented code 😬

src/file-upload/internal.tsx
katiegeorge: It was picking up className and id, and then a test was failing that className & id should only apply to a single component. We don't want className to apply to both the file input and file upload wrappers

src/file-token-group/index.tsx
katiegeorge: What do you mean by useFilesLoading? Is there an example you can share?

pan-kot (approved):

PR #3013, opened 2024-11-14
fix: Sticky offset in nested AppLayout

georgylobko:
### Description

Keep table's sticky header within a nested AppLayout on scroll by resetting globalVars.stickyVerticalTopOffset and globalVars.stickyVerticalBottomOffset for nested layout.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/app-layout/visual-refresh-toolbar/index.tsx
```diff
@@ -432,16 +437,41 @@ const AppLayoutVisualRefreshToolbar = React.forwardRef<AppLayoutProps.Ref, AppLa
...
 
+    /**
+     * Returns true if the AppLayout is nested
+     * Does not apply to iframe
+     */
+    const getIsNestedInAppLayout = (element: HTMLElement | null, rootClassName: string): boolean => {
+      let currentElement: Element | null = element?.parentElement ?? null;
+
+      while (currentElement) {
+        if (currentElement.classList.contains(rootClassName)) {
```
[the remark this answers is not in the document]
georgylobko: > There is no guarantees rootClassName will remain unchanged across versions

In this case the test I introduce will fail. Isn't it enough to avoid such a scenario?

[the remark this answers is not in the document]
just-boris: We cannot get multiple app layouts of different versions in our integration testing.

Also, when the test will fail, it will be too late, because we want to ship forward compatible code now, instead of doing weird backward-compatibility tricks in the future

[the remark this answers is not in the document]
georgylobko: > We cannot get multiple app layouts of different versions in our integration testing.

True, but if anyone changes this classname, the integration test will fail anyway.

> getComputedStyle(rootRef.current)[globalVars.stickyVerticalTopOffset] 

I would not say it's more reliable way as this global var might be (in theory) changed / removed / whatever.

So I would keep this condition as simple as possible.

src/app-layout/visual-refresh-toolbar/index.tsx
```diff
@@ -432,16 +437,41 @@ const AppLayoutVisualRefreshToolbar = React.forwardRef<AppLayoutProps.Ref, AppLa
...
+          return true;
+        }
+        currentElement = currentElement.parentElement;
+      }
+
+      return false;
+    };
+
+    useEffect(() => {
```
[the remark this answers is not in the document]
georgylobko: Fair enough

src/app-layout/visual-refresh-toolbar/index.tsx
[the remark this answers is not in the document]
just-boris: I still do not get how this test "should keep header visible and in position while scrolling" will catch multi version environment.

> this global var might be (in theory) changed / removed / whatever

No it won't. It is a part of our stable API (this is why it is called a "global var" and has no hash suffix, unlike other things)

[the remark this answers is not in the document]
just-boris: Generated class name includes a hash sum of the content. So if any of these styles change (which is plausible), you will get a new class and this code will break: https://github.com/cloudscape-design/components/blob/dc7000bf1f341c3da62ac9228d7ef86fdf587ddb/src/app-layout/visual-refresh-toolbar/skeleton/styles.scss#L30-L46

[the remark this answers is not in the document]
georgylobko: Oh, makes sense now. Thanks for the explanation

src/app-layout/visual-refresh-toolbar/index.tsx
```diff
@@ -432,16 +435,41 @@ const AppLayoutVisualRefreshToolbar = React.forwardRef<AppLayoutProps.Ref, AppLa
...
+        }
+        currentElement = currentElement.parentElement;
+      }
+
+      return false;
+    };
+
+    useLayoutEffect(() => {
+      setIsNested(getIsNestedInAppLayout(rootRef.current));
+    }, []);
```
georgylobko, line 457: @just-boris you previously suggested to skip this calculation if toolbar is empty, but it turns out this performance optimization might even backfire if we put it as a dependency here, since it will be recalculated every time this prop changes, so I decided to leave it as is

src/app-layout/visual-refresh-toolbar/index.tsx
```diff
@@ -432,16 +435,41 @@ const AppLayoutVisualRefreshToolbar = React.forwardRef<AppLayoutProps.Ref, AppLa
...
+     * Does not apply to iframe
+     */
+    const getIsNestedInAppLayout = (element: HTMLElement | null): boolean => {
+      let currentElement: Element | null = element?.parentElement ?? null;
+
+      while (currentElement) {
+        if (getComputedStyle(currentElement).getPropertyValue(globalVars.stickyVerticalTopOffset)) {
+          return true;
+        }
+        currentElement = currentElement.parentElement;
```
just-boris, line 449: CSS var values propagate. You do not need to manually walk up

src/app-layout/visual-refresh-toolbar/index.tsx
just-boris: I do not understand this comment. Could you post here in the comments a possible code how you planned to handle the toolbar and which dependency concerned you?

pages/app-layout/multi-layout-with-table-sticky-header.page.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { Breadcrumbs, Navigation } from './utils/content-blocks';
+import { Tools } from './utils/content-blocks';
+import labels from './utils/labels';
+import * as toolsContent from './utils/tools-content';
+
+const items = generateItems(100);
+
+export default function () {
+  return (
+    <ScreenshotArea gutters={false}>
```
just-boris, line 22: Need to ensure this dev page passes new integration tests added there: https://github.com/cloudscape-design/components/pull/3063

just-boris (dismissed):
Approved, but it would be great to address the comments

src/app-layout/visual-refresh-toolbar/index.tsx
georgylobko:
```javascript
useLayoutEffect(() => {
      if (toolbarProps) {
            setIsNested(getIsNestedInAppLayout(rootRef.current));
      }
    }, [toolbarProps]);
```

src/app-layout/visual-refresh-toolbar/index.tsx
georgylobko: Technically, the decision suggested by you is better, but it does not work in u tests env so I turned it back

just-boris: Okay, so `getComputedStyle` implementation is JSDOM is more naive.

Let's leave a comment for the future readers this loop is needed for JSDOM. 

Fine to keep it because in the real browsers it will exit after the first iteration anyway

src/app-layout/visual-refresh-toolbar/index.tsx
just-boris: How about using `hasToolbar` value. It is boolean and rarely changes: https://github.com/cloudscape-design/components/blob/d1ff27eaa1896e77cdc87fd7b4c3c67528962133/src/app-layout/visual-refresh-toolbar/index.tsx#L291

pages/app-layout/multi-layout-with-table-sticky-header.page.tsx
georgylobko: It does

src/app-layout/visual-refresh-toolbar/index.tsx
```diff
@@ -432,16 +435,45 @@ const AppLayoutVisualRefreshToolbar = React.forwardRef<AppLayoutProps.Ref, AppLa
...
         {/* Rendering a hidden copy of breadcrumbs to trigger their deduplication */}
         {!hasToolbar && breadcrumbs ? <ScreenreaderOnly>{breadcrumbs}</ScreenreaderOnly> : null}
         <SkeletonLayout
-          ref={intersectionObserverRef}
+          ref={useMergeRefs(intersectionObserverRef, rootRef)}
           style={{
-            [globalVars.stickyVerticalTopOffset]: `${verticalOffsets.header}px`,
-            [globalVars.stickyVerticalBottomOffset]: `${placement.insetBlockEnd}px`,
             paddingBlockEnd: splitPanelOpen && splitPanelPosition === 'bottom' ? splitPanelReportedSize : '',
+            ...(toolbarProps || !isNested
```
just-boris, line 471:
```suggestion
            ...(hasToolbar || !isNested
```

just-boris (commented):
Last one and then approved

just-boris (approved):

PR #3020, opened 2024-11-15
feat: Support theming border width and button font weight.

DaemonCahill:
This PR adds support for theming field components border width and button font weight.


jkuelz:
You need to regenerate the design token and theming snapshots to make the tests pass, run the following:
`npx jest -u -c jest.integ.config.js src/__integ__/themes.test.ts`
`npx jest -u -c jest.unit.config.js src/__tests__/design-tokens.test.ts`

src/popover/arrow.scss
```diff
@@ -71,10 +71,10 @@
   }
 
   &-inner {
-    inset-block-start: 2px;
+    inset-block-start: awsui.$border-width-field;
```
gethinwebster, line 74: classic `border-width-field` is 1px as far as I can see, so I think both should be `calc(awsui.$border-width-field + 1px);`?

style-dictionary/visual-refresh/metadata/borders.ts
```diff
@@ -109,6 +109,16 @@ const metadata: StyleDictionary.MetadataIndex = {
     public: true,
     themeable: true,
   },
+  borderWidthAlert: {
+    description: 'The border width of alerts.',
+    public: true,
+    themeable: true,
+  },
+  borderWidthField: {
+    description: 'The border width of fields.',
```
gethinwebster, line 118: other design tokens typically refer to "form fields" rather than just "fields"

gethinwebster (changes requested):

style-dictionary/visual-refresh/metadata/borders.ts
DaemonCahill: Updated!

src/popover/arrow.scss
DaemonCahill: Good catch. They can both share the same inset value now that it is tokenized.

gethinwebster (approved):

src/internal/styles/forms/mixins.scss
```diff
@@ -238,8 +238,8 @@
     inset-inline-start: 0px;
     inset-inline-end: 0px;
     background-color: transparent;
-    border-block: awsui.$border-field-width solid awsui.$color-border-dropdown-container;
-    border-inline: awsui.$border-field-width solid awsui.$color-border-dropdown-container;
+    border-block: awsui.$border-width-field solid awsui.$color-border-dropdown-container;
```
jkuelz, line 241: I would expect dropdown container and popover to have their own token, not use form field.

src/popover/arrow.scss
```diff
@@ -71,11 +71,7 @@
   }
 
   &-inner {
-    inset-block-start: 2px;
-
-    &.refresh {
-      inset-block-start: 3px;
-    }
+    inset-block-start: calc(awsui.$border-width-field + 1px);
```
jkuelz, line 74: I would not consider popover a form field component. I would treat it similar to alert, and create its own token or use a generic dropdown/popup token for dropdown and popover.

src/select/parts/styles.scss
```diff
@@ -135,7 +135,7 @@ $inlineLabel-border-radius: 2px;
   // Stepped gradient background for inline label overlap between input and background.
   background-image: linear-gradient(
     to bottom,
-    transparent calc(100% - (awsui.$border-field-width + awsui.$border-control-focus-ring-shadow-spread + 5px)),
+    transparent calc(100% - (awsui.$border-width-field + awsui.$border-control-focus-ring-shadow-spread + 5px)),
```
jkuelz, line 138: It looks like this brakes in [collection select filters](https://d21d5uik3ws71m.cloudfront.net/components/ac6b34785a9febf09da61c39f79b6166c6f38235/dev-pages/index.html#/light/app-layout/with-table-collection-select-filter), if you change the field width. It strikes through the input label.

DaemonCahill: This needs more investigation I'll make a note to follow up.

jkuelz (dismissed):

jkuelz (dismissed):

src/anchor-navigation/styles.scss
```diff
@@ -117,7 +117,7 @@ $guide-line-offset: -2px; // Negative to expand 2px beyond the element
   margin-inline-start: awsui.$space-xs;
   @include styles.font-body-s;
   @include styles.font-smoothing;
-  font-weight: awsui.$font-button-weight;
+  font-weight: awsui.$font-weight-button;
```
gethinwebster, line 120: Especially now that `font-weight-button` is public/themable, I don't think it makes sense to use it here: it would be unexpected for changing the button weight to impact this "info" text. Suggest using `styles.$font-weight-bold;` instead

src/flashbar/styles.scss
```diff
@@ -80,7 +80,7 @@
 }
 
 .flash-text {
-  margin-block: awsui.$border-field-width;
+  margin-block: awsui.$border-width-field;
```
gethinwebster, line 83: similar: I don't see any reason why this should semantically be linked to form field border width (also affects the related `calc` in the file above)

src/flashbar/collapsible.scss
```diff
@@ -241,7 +241,7 @@ the grid layout will be:
     row-gap: 0;
 
     > .header {
-      font-weight: awsui.$font-button-weight;
+      font-weight: awsui.$font-weight-button;
```
gethinwebster, line 244: Screenshot 2024-11-19 at 09 20 06

unsure on the semantics here, I guess it _kinda_ is a button?

src/anchor-navigation/styles.scss
gethinwebster: ("Updated" and "New" in this screenshot:)

Screenshot 2024-11-19 at 09 20 56

src/flashbar/styles.scss
gethinwebster: Screenshot 2024-11-19 at 09 22 02

src/link/constants.scss
```diff
@@ -71,7 +71,7 @@ $link-variants: (
     'text-color-default': awsui.$color-text-link-default,
     'text-color-hover': awsui.$color-text-link-hover,
     'text-color-active': awsui.$color-text-link-hover,
-    'font-weight': awsui.$font-button-weight,
+    'font-weight': awsui.$font-weight-button,
```
gethinwebster, line 74: don't think that this should inherit the button weight: info link is visually styled more like a link not a button

src/link/constants.scss
```diff
@@ -118,7 +118,7 @@ $link-styles: (
       'text-color-default': awsui.$color-text-link-button-normal-default,
       'text-color-hover': awsui.$color-text-link-button-normal-hover,
       'text-color-active': awsui.$color-text-link-button-normal-active,
-      'font-weight': awsui.$font-button-weight,
+      'font-weight': awsui.$font-weight-button,
```
gethinwebster, line 121: Uncertain here. I could see arguments both ways, but I think this is still semantically more a link than a button, and like for `info` the weight is an important affordance as there is no border/background

src/internal/styles/typography/mixins.scss
```diff
@@ -94,7 +94,7 @@
 }
 
 @mixin font-button {
-  font-weight: awsui.$font-button-weight;
+  font-weight: awsui.$font-weight-button;
```
gethinwebster, line 97: I think we also need to consider what the expected behavior is for `<Button href="#" variant="link">`, and even more so `<Button href="#" variant="inline-link">`: should the font-weight of those be tied to that of normal buttons, or are they visually distinct enough entities that they should have their own design token(s)?

gethinwebster (changes requested):
Sorry for re-review, took a look at this again with more of a focus on "we're making more tokens themable" and what the possible unexpected impacts of that are, and I think there are important things that I missed first time round.

(in a nutshell, right now I think theming `font-weight-button: normal` would have unexpected side-effects, including introducing some potentially serious accessibility concerns)

src/anchor-navigation/styles.scss
DaemonCahill: Updated!

src/flashbar/collapsible.scss
DaemonCahill: Agreed I'm inclined to leave this as button.

src/internal/styles/typography/mixins.scss
DaemonCahill: Per offline discussion leaving this as is.

src/link/constants.scss
DaemonCahill: Resolved!

src/link/constants.scss
DaemonCahill: Resolved!

src/flashbar/styles.scss
DaemonCahill: Per offline discussion leaving this as is.

gethinwebster (approved):

PR #3025, opened 2024-11-18
feat: Improve breadcrumb group responsivness

gethinwebster:
### Description

Improves breadcrumb group responsiveness so that:
1. Links are not truncated (only the final, non-link item can be truncated) due to challenges with tooltips on interactive elements
2. All items are collapsed into a dropdown when insufficient space is available

Related links, issue #, if available: 3SICARmRNK32

### How has this been tested?

Updated unit and integ tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/breadcrumb-group/full-dropdown.tsx
```diff
@@ -0,0 +1,88 @@
...
+        }
+        if (metadata.detail?.position) {
+          metadata.detail.position = `${parseInt(metadata.detail.position as string, 10) + 1}`;
+        }
+        return metadata;
+      }}
+    />
+    {items.length >= 1 && (
+      <span className={styles.hidden} {...metadataTypeAttribute}>
+        {items[1].text}
```
pan-kot, line 55: Should this be `items[items.length - 1].text`?

src/breadcrumb-group/full-dropdown.tsx
```diff
@@ -0,0 +1,88 @@
...
+        const isCurrentPage = index === items.length - 1;
+        return {
+          id: index.toString(),
+          text: item.text,
+          href: isCurrentPage ? undefined : item.href,
+          disabled: isCurrentPage,
+          isCurrentPage,
+        };
+      })}
+      className={styles['collapsed-button-dropdown']}
```
pan-kot, line 38: Should we add an internal prop to the button dropdown instead? The risk of having a custom style is that it is hard to secure no regressions when making changes to the button dropdown.

src/breadcrumb-group/full-dropdown.tsx
```diff
@@ -0,0 +1,88 @@
...
+export const FullDropdown = ({ items, onClick, onFollow }: FullCollapsedDropdownProps) => (
+  <>
+    <InternalButtonDropdown
+      items={items.map((item, index) => {
+        const isCurrentPage = index === items.length - 1;
+        return {
+          id: index.toString(),
+          text: item.text,
+          href: isCurrentPage ? undefined : item.href,
+          disabled: isCurrentPage,
```
pan-kot, line 34: Would it make sense to add the disabled attribute inside the button dropdown for items with isCurrentPage=true?

src/breadcrumb-group/full-dropdown.tsx
```diff
@@ -0,0 +1,88 @@
...
+  ({ ariaLabel, triggerRef, disabled, testUtilsClass, isOpen, onClick }: CustomTriggerProps) => {
+    const metadataAttributes = {
+      [DATA_ATTR_FUNNEL_KEY]: FUNNEL_KEY_FUNNEL_NAME,
+    };
+    return (
+      <button
+        ref={triggerRef as Ref<HTMLButtonElement>}
+        {...metadataAttributes}
+        className={clsx(styles['collapsed-button'], testUtilsClass)}
+        disabled={disabled}
```
pan-kot, line 71: Can this actually be disabled?

src/breadcrumb-group/implementation.tsx
```diff
@@ -254,7 +254,15 @@ export function BreadcrumbGroupImplementation<T extends BreadcrumbGroupProps.Ite
           }
         : {})}
     >
-      <ol className={styles['breadcrumb-group-list']}>{breadcrumbItems}</ol>
+      {collapsed === breadcrumbItems.length - 2 ? (
+        <FullDropdown
```
pan-kot, line 258: nit: Would it make sense to call this component AllItemsDropdown?

src/breadcrumb-group/styles.scss
```diff
@@ -64,3 +70,58 @@
...
+    white-space: nowrap;
+    overflow: hidden;
+    text-overflow: ellipsis;
+    &:hover {
+      color: awsui.$color-text-interactive-hover;
+    }
+  }
+}
+
+.button-icon-open {
```
pan-kot, line 111: idea: Can we create some mixins to reuse the rotating icon styles between button dropdown, expandable sections, expandable rows, and this new component?

src/breadcrumb-group/__tests__/utils.test.ts
```diff
@@ -6,94 +6,24 @@ describe('getItemsDisplayProperties', () => {
...
-      });
-    });
-    test('first and last smaller than default min width and bigger than available width', () => {
-      expect(getItemsDisplayProperties([140, 200, 20, 140, 140], 320)).toEqual({
-        shrinkFactors: [140, 200, 0, 140, 140],
-        minWidths: [135, 135, 0, 135, 135],
-        collapsed: 3,
-      });
-    });
+  test.each([
```
pan-kot, line 12: nice test design 👍

src/breadcrumb-group/__tests__/full-dropdown.test.tsx
```diff
@@ -0,0 +1,45 @@
...
+  },
+  {
+    text: 'Last item',
+    href: '#',
+  },
+];
+describe('FullDropdown', () => {
+  let wrapper: ButtonDropdownWrapper;
+  beforeEach(() => {
+    const { container } = render(<FullDropdown items={items} onClick={jest.fn()} onFollow={jest.fn()} />);
```
pan-kot, line 26: I think it would make sense to define the full-dropdown tests in context of the breadcrumb group: render the breadcrumb group with all items collapsed and then run the tests. This will ensure that the selectors to find the full dropdown from the button group root wrapper are covered, too.

Btw, should we add more test utils?

gethinwebster: yeah, I think that makes sense, i'll consolidate.

I don't think we need more test utils: this new dropdown is still findable via the existing `findDropdown` util, and items within the new and old dropdowns can be interacted with in the same way. I could _potentially_ see the need to check which mode the breadcrumbs overall is in, but didn't need that for any of the tests so far, we can always add in the future if requested.

src/breadcrumb-group/full-dropdown.tsx
gethinwebster: not sure I understand what you mean here

src/breadcrumb-group/full-dropdown.tsx
gethinwebster: No, it's supposed to specifically be the second item, it corresponds to this line https://github.com/cloudscape-design/components/blob/49a178d8f7fa35206645d217c949ecb180dc322d/src/breadcrumb-group/item/funnel.tsx#L29

Will add a comment to clarify

src/breadcrumb-group/full-dropdown.tsx
gethinwebster: good point, it shouldn't ever be, so I think we can remove this

src/breadcrumb-group/full-dropdown.tsx
```diff
@@ -0,0 +1,88 @@
...
+import InternalIcon from '../icon/internal';
+import { DATA_ATTR_FUNNEL_KEY, DATA_ATTR_RESOURCE_TYPE, FUNNEL_KEY_FUNNEL_NAME } from '../internal/analytics/selectors';
+import { CancelableEventHandler } from '../internal/events';
+import { BreadcrumbGroupProps } from './interfaces';
+
+import styles from './styles.css.js';
+
+interface FullCollapsedDropdownProps {
+  items: ReadonlyArray<BreadcrumbGroupProps.Item>;
+  onClick: CancelableEventHandler<{ id: string }>;
```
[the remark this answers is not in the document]
gethinwebster: will update

src/button-dropdown/item-element/styles.scss
```diff
@@ -63,7 +63,13 @@
 
   &.link-style {
     padding-block-end: calc(#{styles.$option-padding-vertical} + #{awsui.$space-xxxs});
-    @include styles.link-inline('body-m');
+    &:not(.disabled > &) {
+      @include styles.link-inline('body-m');
+    }
+    &.current-page {
```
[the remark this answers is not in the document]
gethinwebster: agree, makes sense to make this more specific. It's a bit strange that we're adding _such_ specific things to button dropdown, but I think it's best to be transparent that that's what we're doing.

src/breadcrumb-group/utils.ts
```diff
@@ -30,11 +27,11 @@ const computeNumberOfCollapsedItems = (collapsedWidths: Array<number>, navWidth:
   }
   let collapsed = 0;
```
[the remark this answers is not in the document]
gethinwebster: rewrote this whole bit to get rid of bugs...

src/breadcrumb-group/full-dropdown.tsx
pan-kot: I mean that instead of passing down the `disabled: isCurrentPage` we can only pass `isCurrentPage`. In the button dropdown we can then set: `disabled: disabled || isCurrentPage`.

gethinwebster: got it, thanks! after further discussion, disabled has gone entirely in the most recent revision :D

src/breadcrumb-group/styles.scss
gethinwebster: Updated, at least for those that have a 180 spin. For those with 90degrees I think it's better to keep them separate from this as there's not so much direct overlap and they're a bit more complex with RTL.

src/button-dropdown/interfaces.ts
```diff
@@ -258,6 +258,7 @@ export interface ItemProps {
 
 export interface InternalItem extends ButtonDropdownProps.Item {
   badge?: boolean;
+  isCurrentPage?: boolean;
```
pan-kot, line 261: nit: I think it worths adding a doc comment to the new isCurrentPage and fullWidth to explain the purpose.

pan-kot: update: is isCurrentPage used? I think it was replaced by the isCurrentBreadcrumb, right?

src/internal/styles/motion/utils.ts
```diff
@@ -0,0 +1,6 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import clsx from 'clsx';
+
+export const spinWhenOpen = (styles: Record<string, string>, className: string, open: boolean) =>
```
pan-kot, line 5: nice 👍

pan-kot (dismissed):

pan-kot (approved):

PR #3038, opened 2024-11-20
fix: Prevent Chrome crash when changing date range type

jperals:
### Description

Ticket: D179344381

Chrome crashes when replacing some DOM nodes. Reusing the same node fixes it.

### How has this been tested?

Tested manually on my local machine, where the issue could be reproduced before the changes:
- Change from null to relative date range ("Apply")
- Change from null to absolute date range ("Apply")
- Change from relative date range to absolute range ("Apply")
- Change from absolute date range to relative date range ("Apply")
- Change from relative date range to null ("Clear and dismiss")
- Change from absolute date range to null ("Clear and dismiss")

I could not catch the crash with integration tests, so I did not add automated coverage for this fix.

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/internal/utils/date-time/__tests__/format-date-time-with-offset.test.ts
```diff
@@ -0,0 +1,168 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
jperals, line 1: This file is an adaptation of `format-date-range.test.ts` because `formatDateRange` doesn't exist anymore; instead, we call `formatDateTimeWithOffset` twice, once for the start date and once for the end date, which are rendered together at the component level.

src/date-range-picker/index.tsx
```diff
@@ -54,48 +54,35 @@ function renderDateRange({
...
-  const formatted =
-    range.type === 'relative' ? (
-      formatRelativeRange?.(range) ?? ''
-    ) : (
-      <BreakSpaces
-        text={formatDateRange({
-          startDate: range.startDate,
-          endDate: range.endDate,
-          timeOffset,
+  const firstPart = range
```
jperals, line 57: I also tried the approach below where every span had a separate purpose, I found it a bit more elegant but unfortunately still crashed when changing an absolute value for a relative value. It seems like the content if the first inner span can't be set to an empty string:

```
  const formattedRelativeRange = (range?.type === 'relative' && formatRelativeRange?.(range)) ?? '';

  const formattedAbsoluteStartDate =
    range?.type === 'absolute'
      ? formatDateTimeWithOffset({
          date: range.startDate,
          timeOffset: timeOffset.startDate,
          hideTimeOffset,
          format: absoluteFormat,
          locale,
        })
      : '';

  const formattedAbsoluteEndDate =
    range?.type === 'absolute'
      ? formatDateTimeWithOffset({
          date: range.endDate,
          timeOffset: timeOffset.endDate,
          hideTimeOffset,
          format: absoluteFormat,
          locale,
        })
      : '';

  return (
    <span className={(!range && styles['label-text']) || undefined} aria-disabled={!range}>
      <span>{range?.type === 'relative' ? formattedRelativeRange : range ? '' : placeholder}</span>
      <span className={styles['label-token-nowrap']}>{formattedAbsoluteStartDate}</span>
      <span>{formattedAbsoluteEndDate && ' — '}</span>
      <span className={styles['label-token-nowrap']}>{formattedAbsoluteEndDate}</span>
    </span>
```

YueyingLu (approved):

PR #3039, opened 2024-11-20
fix: Table scrollbar overflows container

connorlanigan:
### Description

When a table has a horizontal scrollbar, this scrollbar is placed at the bottom of the table's container. There, it overlaps with the rounded corners and overflows them, hiding the border. After this change, the scrollbar adopts the same rounding as the container, so that it does not hide the border anymore.

The rounding is achieved with the `clip-path` property. It limits the rendering area of the element, but does not affect any content of the table (like `overflow` potentially would), since it can be applied directly on the element that needs to be clipped.

Related links, issue #, if available: AWSUI-43012

#### Before

image

#### After
image

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/table/sticky-scrollbar/sticky-scrollbar.tsx
+++ b/src/table/sticky-scrollbar/sticky-scrollbar.tsx
@@ -3,6 +3,7 @@
 import React, { forwardRef } from 'react';
 import clsx from 'clsx';
 
+import { useIntersectionObserver } from '../../internal/hooks/use-intersection-observer';
 import { useMergeRefs } from '../../internal/hooks/use-merge-refs';
 import { useVisualRefresh } from '../../internal/hooks/use-visual-mode';
 import { browserScrollbarSize } from '../../internal/utils/browser-scrollbar-size';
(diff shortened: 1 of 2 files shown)


src/table/sticky-scrollbar/styles.scss
```diff
@@ -14,6 +14,10 @@
   inset-block-end: 0;
   inline-size: 100%;
 
+  &[data-stuck='false'] {
+    clip-path: inset(-9999px 0 0 0 round awsui.$border-radius-container);
```
jperals, line 18: Instead of avoiding the top cropping with a the negative top coordinate, the border radius can be specified for only for the bottom corners:

```
inset(0 round 0 0 awsui.$border-radius-container awsui.$border-radius-container)
```

WDYT?

connorlanigan: That would generally work and in our currently supported browsers it would not make a visible difference. However, it is based on the assumption that a browser will never render anything above the area that we calculate for the scrollbar. If there is ever a bug in our calculation, or a browser introduces additional UI elements for their scroll interactions in the future (compare e.g. [scroll progress indicator bubbles on Android](https://i.sstatic.net/A9nfP.png)), this would limit the rendering area unnecessarily. With the negative top coordinate, we reduce the risk of clipping where we don't need it.

src/table/sticky-scrollbar/sticky-scrollbar.tsx
```diff
@@ -35,17 +36,24 @@ function StickyScrollbar(
   const offsetScrollbar = hasStickyColumns || browserScrollbarSize().height === 0;
 
   useStickyScrollbar(scrollbarRef, scrollbarContentRef, tableRef, wrapperRef, offsetScrollbar);
+
+  const { ref: stickyDetectionRef, isIntersecting: isStickyDetectionVisible } = useIntersectionObserver();
```
jperals, line 40: Is this check equivalent to the one being done [here](https://github.com/cloudscape-design/components/blob/main/src/table/sticky-scrollbar/use-sticky-scrollbar.ts#L31)?

If so, can we reuse the existing `sticky-scrollbar-visible` class instead of introducing `data-stuck`?

Or, do we need any class or attribute at all? Does this clipping style need to be conditional?

connorlanigan: No, that check you linked is responsible for determining whether there should be a scrollbar at all. The `sticky-scrollbar-visible` class is added if and only if the table content is wider than the available space.

The clipping style needs to be conditional, because the clipped scrollbar would otherwise show the content underneath when it is stuck:

image

jperals (approved):

PR #3056, opened 2024-11-25
fix: Copy to clipboard misleading popover when clicking outside the corner

georgylobko:
### Description

AWSUI-60050

The root of this issue is that we have two trigger handlers for the copy-to-clipboard component: one attached to a custom trigger wrapper (within the popover component) and another to the trigger itself. Since these triggers have slightly different visual layouts - the outer wrapper lacks a border radius while the inner trigger has one, clicking just outside the inner button corners opens the popover but does not activate the copy-to-clipboard handler. This behavior led to the issue described in the ticket.  

This PR resolves the issue by attaching the existing copy-to-clipboard handler to the popover trigger event.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/copy-to-clipboard/__tests__/copy-to-clipboard.test.tsx
+++ b/src/copy-to-clipboard/__tests__/copy-to-clipboard.test.tsx
@@ -85,6 +85,21 @@ describe('CopyToClipboard', () => {
       );
     });
 
+    test('copies to clipboard and shows success message when clicking on popover trigger', async () => {
+      const { container } = render(
+        <CopyToClipboard {...defaultProps} popoverRenderWithPortal={popoverRenderWithPortal} />
+      );
+      const copyToClipboardWrapper = createWrapper(container).findCopyToClipboard()!;
+      const popoverWrapper = createWrapper(container).findPopover();
+
+      popoverWrapper!.findTrigger().click();
+      await waitFor(() =>
+        expect(copyToClipboardWrapper.findStatusText({ popoverRenderWithPortal })!.getElement().textContent).toBe(
+          'Copied to clipboard'
+        )
+      );
+    });
+
     test('fails to copy to clipboard and shows error message', async () => {
       const { container } = render(
         <CopyToClipboard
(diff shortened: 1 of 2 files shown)


gethinwebster (approved):

jperals:
On [this deployed page](https://d21d5uik3ws71m.cloudfront.net/components/c673e36af9ad4162938252d3ec4639a32dfcc412/dev-pages/index.html#/light/copy-to-clipboard/scenario-split-panel) I can still reproduce the issue.

I think what is happening now is that if I click on the corner, the text is not copied, _but_ the popover is opened anyway, which is misleading.

georgylobko:
> On [this deployed page](https://d21d5uik3ws71m.cloudfront.net/components/c673e36af9ad4162938252d3ec4639a32dfcc412/dev-pages/index.html#/light/copy-to-clipboard/scenario-split-panel) I can still reproduce the issue.
> 
> I think what is happening now is that if I click on the corner, the text is not copied, _but_ the popover is opened anyway, which is misleading.

The issue is that this click caused an unexpected behavior (showed the popover without actually copying) so this PR addresses it. Preventing click on the button's corners is much more tricky, because this wrapper is coming from popover component. I considered an option to prevent it from triggering, but it caused more problems than solved. If you have better suggestions, let's discuss them in this thread.

georgylobko:
> Would it help to apply border-radius to the trigger?

Unfortunately, no, because it might prevent clicking on it only in combination with `overflow: hidden`, which breaks the layout.

jperals:
To me it seems to work: if I apply a 20px border radius to the button's immediate parent, the click does not do anything anymore.

The reason I'm insisting is that I don't see the fix makes things better as it currently stands, rather I'm afraid it makes things a bit worse:

- In the current state, clicking on the corners of the popover trigger will show a popover which shows the "in-progress" state, without ever moving further. Confusing but nothing happened
- After these changes, doing the same action (clicking on the corners) shows the success popover, indicating that the text was copied to the clipboard, but it is not true, it is misleading

jperals:
> After these changes, doing the same action (clicking on the corners) shows the success popover, indicating that the text was copied to the clipboard, ~~but it is not true~~

Sorry I see that it does copy to clipboard. Then the change does things a bit better I guess, although ideally clicking on the corners would do nothing.

jperals (approved):

jperals:
POC of the border-radius approach: https://github.com/cloudscape-design/components/pull/3059. Tested locally, clicking outside the borders does nothing.

PR #3060, opened 2024-11-26
chore: Underline styles for popover text trigger

georgylobko:
### Description

AWSUI-59855
Replaced border bottom with underline dotted styles.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


michaeldowseza:
Are all visual differences to be expected from this? e.g. Slight change in dotted line + slight shift in where the popover appears from?

image

georgylobko (commented):
@michaeldowseza 
I’ve addressed this discrepancy and reviewed the screenshot tests in my dev pipeline. Everything looks good now.

georgylobko (commented):
Posting all main diffs with live (upd 16.01.2025):

live:
image
current (underline applies only to the text, not to the icon):
image

live:
image
current (multi-line underline, which is the whole point of this PR):
image

src/popover/styles.scss
```diff
@@ -12,6 +12,14 @@
 @import './container';
 @import './motion';
 
+$trigger-underline-offset: 0.25em;
```
georgylobko, line 15: 0.25 is used instead of 0.2 because an underline at 0.2 overlaps with the text

src/popover/styles.scss
```diff
@@ -46,7 +54,24 @@
...
+
+  &.overflow-ellipsis {
+    /*
+      This line-height needs to be overridden because the overflow: hidden would otherwise conceal the underline styles.
+    */
+    line-height: calc(1 + #{$trigger-underline-offset} + #{awsui.$border-divider-list-width});
+  }
+
+  /* stylelint-disable selector-combinator-disallowed-list, selector-max-type, selector-max-universal */
+  & span {
```
pan-kot, line 72: Would it make sense to add a span for both in-place and portal variants of the popover to avoid the condition in styles?

I also see no reason to violate the stylelint rules: you can add a class name to the span element and target it directly.

georgylobko: Unfortunately, that is not a case, because this styles are applied when we are dealing with 'external' component which we have no control over. See the example below:

```typescript
<Popover
  size="medium"
  position="right"
  header="Memory error"
  content={<DynamicContent />}
  dismissAriaLabel="Close"
  renderWithPortal={renderWithPortal}
>
   <StatusIndicator type="error">Error</StatusIndicator>
</Popover>
```

The current implementation applies border styles directly to the popover trigger, which is a block element, while underline styles are not inherited down the DOM tree. Therefore, I realized that the only way to apply the styles to the inline element is by using this selector.

pan-kot: What if the nested element is not a span?

georgylobko: Since this change specifically applies to `text` type of triggers, I think it's a reasonable assumption that only span elements will be used. This code specifically targets our components and use cases. If something doesn't work well for users, they always have the option to introduce a custom trigger

src/popover/styles.scss
```diff
@@ -46,7 +54,24 @@
...
+  @include trigger-underline-styles;
+
+  &.overflow-ellipsis {
+    /*
+      This line-height needs to be overridden because the overflow: hidden would otherwise conceal the underline styles.
+    */
+    line-height: calc(1 + #{$trigger-underline-offset} + #{awsui.$border-divider-list-width});
+  }
+
+  /* stylelint-disable selector-combinator-disallowed-list, selector-max-type, selector-max-universal */
```
pan-kot, line 71: I would add a code comment explaining that the combinator actually targets the custom content

pan-kot (approved):

PR #3061, opened 2024-11-26
fix: Fixes table editable cell styles and simplifies it

pan-kot:
### Description

The PR refactors table body cell styles with the focus on editable cells. That is needed for the follow-up task to move editable cell style from the cell element to the cell content instead. This would allow having both row expand toggle and editable cell trigger in a single cell (in 1st column).

### How has this been tested?

* Screenshot tests
* Manual testing on the cell-permutations page

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/table/body-cell/styles.scss
```diff
@@ -25,11 +25,10 @@ $edit-button-padding-right: calc(#{awsui.$space-xs} + #{awsui.$space-xxs});
 // Cell vertical padding + xxs space that would normally come from the button.
 $success-icon-padding-right: calc(#{$edit-button-padding-right} + #{$icon-width-with-spacing});
 $interactive-column-padding-inline-end: calc(#{$cell-horizontal-padding} + #{awsui.$space-l});
+$editing-cell-padding-inline: awsui.$space-xxs;
+$editing-cell-padding-block: awsui.$space-scaled-xxxs;
 $cell-offset: calc(#{awsui.$space-m} + #{awsui.$space-xs});
 
-// Ensuring enough space for absolute-positioned focus outlines of focus-able cell content elements.
-$cell-negative-space-vertical: 2px;
```
pan-kot, line 31: This negative space was used to virtually enlarge the cell-content element. That is because the element has `overflow: hidden` but when there are interactive elements inside such a link or a button dropdown - the focus outline is being cut off.

The negative space is no longer needed because now the cell content size matches cell size. This way the outline can use the available cell space while the content is still truncated correctly.

src/table/body-cell/styles.scss
```diff
@@ -131,11 +127,18 @@ $cell-negative-space-vertical: 2px;
...
-    overflow: hidden;
-    text-overflow: ellipsis;
+    block-size: 100%;
+    align-content: center;
+
+    &.body-cell-align-top {
+      align-content: baseline;
+    }
+
+    &:not(.body-cell-wrap) {
```
pan-kot, line 137: The `.body-cell-wrap` only takes effect on the cell-content element so attaching the style to it directly.

src/table/body-cell/styles.scss
```diff
@@ -177,20 +180,10 @@ $cell-negative-space-vertical: 2px;
     &:not(.has-selection):not(.body-cell-editable) {
       border-inline-start: none;
     }
-
-    // Give extra space on the edge to allow slight content overflow.
-    // That is to prevent focus outline on cell content from being cut out.
-    > .body-cell-content {
-      padding-inline-start: awsui.$border-divider-list-width;
-      margin-inline-start: calc(-1 * #{awsui.$border-divider-list-width});
```
pan-kot, line 185: This negative space shenanigans are no longer needed because there is plenty of space for the focus outline.

src/table/body-cell/styles.scss
```diff
@@ -380,19 +373,6 @@ $cell-negative-space-vertical: 2px;
       color: awsui.$color-text-button-normal-active;
     }
 
-    &-form {
```
pan-kot, line 383: For editing cells we used `overflow: visible` on cell content together with negative margins so that the editor can take more space inside a cell and its focus outline is not cut off. Instead, in the refactored versions we just use different padding for editing cells and there is no longer an issue with the outline.

src/table/body-cell/td-element.tsx
```diff
@@ -90,6 +90,7 @@ export const TableTdElement = React.forwardRef<HTMLTableCellElement, TableTdElem
       collapseButtonLabel,
       verticalAlign,
       resizableColumns,
+      resizableStyle,
```
pan-kot, line 93: The `resizableStyle` replaces `style` to make the cell API better predictable so that no unexpected attributes can be assigned from the outside.

src/table/styles.scss
```diff
@@ -53,6 +53,7 @@
 
 .table {
   inline-size: 100%;
+  block-size: 100%;
```
pan-kot, line 56: This change does not affect the table itself - the size of the table is determined by the size of its content.

However, the table element needs the height set for cells height to take an effect, see discussion: https://stackoverflow.com/questions/3215553/make-a-div-fill-an-entire-table-cell.

src/table/body-cell/td-element.tsx
```diff
@@ -101,6 +102,8 @@ export const TableTdElement = React.forwardRef<HTMLTableCellElement, TableTdElem
     const Element = isRowHeader ? 'th' : 'td';
     const isVisualRefresh = useVisualRefresh();
 
+    resizableStyle = resizableColumns ? {} : resizableStyle;
```
georgylobko, line 105: this condition looks counterintuitive. you apply `resizableStyle` only when `resizableColumns` is false. why is that?

pan-kot: That is taken as is from the current impl: https://github.com/cloudscape-design/components/blob/main/src/table/internal.tsx#L627

When resizable columns is on, the styles are assigned directly to cell nodes in `src/table/column-widths-utils.ts` and I think the properties like min-width and max-width are ignored.

I think there are further opportunities for refactoring and maybe this condition can be just removed but I'd not include this into the current PR.

YueyingLu:
Found 2 issues with manual testing:
1. can not open select dropdown 

https://github.com/user-attachments/assets/6682b553-b18c-44bd-89eb-2d8266ff114a



2. the input position seems a bit off 
Screenshot 2024-12-09 at 22 28 15

pan-kot:
@georgylobko, @YueyingLu 

I added a new commit with the following changes:
1. Reverted changes to negative vertical space. That is because in classic+compact the focus outline can be still cut off if the button is rendered inside content and the rows only have text content. The vertical paddings we use for table cell are under 2px whilst the focus outline offset for buttons is 2px - hence we need to do a trick so that the focus outline can fit.
2. Reverted changes to visible overflow in editable cells. When a cell is editing the overflow is allowed so that components with dropdown like Select can be displayed correctly.

I will re-run the screenshot tests for these changes, too.

src/table/body-cell/styles.scss
pan-kot: Update: the negative space is still needed at least for the compact mode.

src/table/body-cell/styles.scss
pan-kot: Update: the overflow visible is still needed for editable content that includes dropdowns. I will contribute a new screenshot test to capture that.

YueyingLu (approved):

georgylobko (approved):

PR #3063, opened 2024-11-26
fix: Fix accessibility findings with app layout toolbar

avinashbot:
### Description

Made the following app layout changes:
- Provided a label to the navigation region around the side navigation trigger button
- Removed `undefined` classname and testid findings
- Provided matching id to bottom split panel to match aria-controls on the split panel trigger
- Miscellaneous dev page labelling fixes

#### Unlabeled side navigation trigger

The issue we're facing is one of an unlabeled `nav` element on the page. We can't just assign the `ariaLabels.navigation` value to it, because we're already assigning to the open side navigation panel, and we can't have two navigation landmarks with the same labels. The "floating button" design doesn't have this issue because it goes away as soon as the panel is opened. We can do the same thing here, by just removing the navigation role around the button container when the panel is open.

I didn't change the `nav` to a `div` out of pure caution - didn't want to give this change the chance to break any consumer tests.

Related links, issue #, if available: AWSUI-60014

### How has this been tested?

Passing existing tests, verified locally with VoiceOver. Will definitely need running through the dev pipeline (AwsUi-v3-dwaraa) and a dry-run (7134801478), just to be extra sure.

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/app-layout/visual-refresh-toolbar/toolbar/index.tsx
```diff
@@ -180,7 +180,11 @@ export function AppLayoutToolbarImplementation({
     >
       <div className={styles['toolbar-container']}>
         {hasNavigation && (
-          <nav className={clsx(styles['universal-toolbar-nav'])}>
+          <nav
+            className={clsx(styles['universal-toolbar-nav'])}
+            role={navigationOpen ? 'presentation' : 'navigation'}
+            aria-label={navigationOpen ? undefined : ariaLabels?.navigation}
```
just-boris, line 186: To make this code more error-proof, it could be a separate object

```js
const navLandmarkAttributes = navigationOpen ? {role: 'presentation'} : {role: 'navigation', 'aria-label': ariaLabels?.navigation};
```

just-boris (approved):

just-boris (approved):

PR #2794, opened 2024-10-01
fix: Fixes split panel bouncing on the edge between side/bottom when scrollbar appears

pan-kot:
### Description

Rel: AWSUI-54993

The issue:

https://github.com/user-attachments/assets/7c0dbd99-382d-49a2-acd8-91d71b42d048

### How has this been tested?

* New integration test
* Manual testing with Classic, VR, Toolbar

<details>
   <summary>Review checklist</summary>

_The following items are to be evaluated by the author(s) and the reviewer(s)._

#### Correctness

- _Changes include appropriate documentation updates._
- _Changes are backward-compatible if not indicated, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#public-apis)._
- _Changes do not include unsupported browser features, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#browsers-support)._
- _Changes were manually tested for accessibility, see [accessibility guidelines](https://cloudscape.design/foundation/core-principles/accessibility/)._

#### Security

- _If the code handles URLs: all URLs are validated through [the `checkSafeUrl` function](https://github.com/cloudscape-design/components/blob/main/src/internal/utils/check-safe-url.ts)._

#### Testing

- _Changes are covered with new/existing unit tests?_
- _Changes are covered with new/existing integration tests?_
</details>

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #3065, opened 2024-11-27
fix: Prevent infinite loop when split panel changes position

jperals:
Issue: `AWSUI-59954`

### Description

After the split panel changes its position from bottom to side or vice versa, there is a chance that a scrollbar appears or disappears from the page, which changes the available space again, potentially causing an infinite loop.

To prevent that, give the split panel some "resistance" to change from bottom to side, requiring some extra space on top of the "nominal" 280px.

### Testing

- Added integration tests
- Manually tested on Chrome, Firefox and Safari
- Ran through my pipeline

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/app-layout/split-panel/split-panel-utils.ts
```diff
@@ -10,17 +12,23 @@ import { SPLIT_PANEL_MIN_WIDTH, SPLIT_PANEL_SCROLLBAR_MARGIN } from './constants
...
  */
-export function checkSplitPanelForcedPosition({
+export function useSplitPanelForcedPosition({
   isMobile,
   splitPanelMaxWidth,
 }: {
   isMobile: boolean;
   splitPanelMaxWidth: number;
 }) {
+  const isForcedRef = useRef<boolean>(false);
```
georgylobko, line 22: Would it be easier to use closures instead of turning this function into react hook?

jperals: It was not difficult, I only had to rename a couple files :). I felt like doing that and resort to refs would be a more conventional and therefore less risky option than introducing my own solutions.

src/app-layout/split-panel/use-split-panel-forced-position.ts
```diff
@@ -0,0 +1,39 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { useRef } from 'react';
+
+import { SPLIT_PANEL_MIN_WIDTH } from './constants';
+
+// Magnitude which should be equal or larger than the browser scrollbar thickness
+const scrollbarThreshold = 20;
```
georgylobko, line 9: Let's calculate the actual scroll width and use this value as a fallback

src/app-layout/split-panel/use-split-panel-forced-position.ts
```diff
@@ -0,0 +1,39 @@
...
+  const isForcedRef = useRef<boolean>(false);
+  if (isMobile) {
+    return true;
+  }
+  // After the split panel changes its position from bottom to side or vice versa,
+  // there is a chance that a scrollbar appears or disappears from the page,
+  // which changes the available space again, potentially causing an infinite loop.
+  // To prevent that, we give the split panel some "resistance" to change from bottom to side,
+  // requiring some extra space on top of the "nominal" 280px.
+  const isAtBottom = isForcedRef.current;
```
georgylobko, line 34: Since `isForcedRef.current` is initially `false`, `isAtBottom` doesn't actually reflect what does this variable reflect. Let's use `isForcedRef.current` directly instead.

src/app-layout/split-panel/use-split-panel-forced-position.ts
jperals: Done in latest revision, thank you for the suggestion. Ran through the Lighthouse tests in my pipeline to make sure there was no performance regression.

The existing `browserScrollbarSize` function already takes care of memoizing the value by using the `calculateOnce` function.

src/app-layout/split-panel/use-split-panel-forced-position.ts
jperals: This should not be an issue anymore in the latest revision.

src/app-layout/__integ__/app-layout-split-panel-forced-position.test.ts
```diff
@@ -0,0 +1,129 @@
...
+      (_description, { scrollbarWhenOnSide, scrollbarWhenOnBottom, scrollbarDisplayCondition }) => {
+        test(
+          'transitions from side to bottom',
+          setupTest({ scrollbarDisplayCondition, theme, initialViewportWidth: wide }, async page => {
+            // Open split panel and ensure it is on the side.
+            await expect(page.verifySplitPanelPosition('side')).resolves.toBe(true);
+            await expect(page.hasPageScrollbar()).resolves.toBe(scrollbarWhenOnSide);
+
+            // Split panel transitions to the bottom when not enough horizontal space.
+            for (let newWidth = wide; newWidth >= narrow; newWidth--) {
```
georgylobko, line 103: Why not just `setWindowSize` immediately to `narrow` width? What is the purpose of this loop?

src/app-layout/__integ__/app-layout-split-panel-forced-position.test.ts
```diff
@@ -0,0 +1,129 @@
...
+        test(
+          'transitions from bottom to side',
+          setupTest({ scrollbarDisplayCondition, theme, initialViewportWidth: narrow }, async page => {
+            // Open split panel and ensure it is forced to the bottom.
+            await expect(page.verifySplitPanelPosition('bottom')).resolves.toBe(true);
+            await expect(page.hasPageScrollbar()).resolves.toBe(scrollbarWhenOnBottom);
+
+            // Split panel transitions to the side when enough horizontal space.
+            for (let newWidth = narrow; newWidth <= wide; newWidth++) {
+              await page.setWindowSize({ ...viewports.desktop, width: newWidth });
```
georgylobko, line 120: Same question here

georgylobko (commented):
Good job, thank you. I left a couple of questions regarding integ tests implementation.

Also, I’m curious - what happens with the new integration tests if we remove the logic that handles this behavior? Does it result in an infinite loop?

src/app-layout/__integ__/app-layout-split-panel-forced-position.test.ts
jperals: The purpose of these loops is to catch eventual crashes due to infinite render loops, as reported in `AWSUI-59954` (page-level scrollbar appears only when split panel is at the bottom), whenever they might happen around the "nominal" breakpoint. If changing the viewport width just once directly from e.g, 270px to 290px, we might not catch the bug, which might happen somewhere in between. And if figuring out what the exact point of failure is, this value might change in the future, opening the door to the bug again.

Unfortunately though, I cannot find a way to catch the other form of the issue (`AWSUI-54993`, page-level scrollbar appears only when split panel is on the side), which manifests itself with flickering instead of a crash. So to your question,

> what happens with the new integration tests if we remove the logic that handles this behavior?

these tests pass, so technically they don't cover the changes in this PR, but they cover the crash case which was present before reverting https://github.com/cloudscape-design/components/pull/2794. And in fact the flickering itself was never really covered, I just verified that the tests https://github.com/cloudscape-design/components/pull/2794 also passed when the other changes in that PR are reverted.

georgylobko (approved):

PR #3081, opened 2024-11-30
chore: Cleanup unused code

just-boris:
### Description

Run `npx knip` (static analyzer to detect unused code) and then manually a fix some false matches

### How has this been tested?

* PR build
* `npm start` locally
* dry-run 7136287246

<details>

#### Correctness

#### Security

#### Testing

</details>

---


package.json
```diff
@@ -75,7 +75,6 @@
     "@types/node": "^18.18.5",
     "@types/react": "^16.14.20",
     "@types/react-dom": "^16.9.14",
-    "@types/react-resizable": "^1.7.4",
```
just-boris, line 78: The actual dependency was removed long ago, but the types stayed: https://github.com/cloudscape-design/components/pull/616

package.json
```diff
@@ -123,7 +122,6 @@
     "react-router-dom": "^5.3.0",
     "react-test-renderer": "^16.14.0",
     "react-virtual": "^2.8.2",
-    "replace-in-files": "^3.0.0",
```
just-boris, line 126: Not used

src/__integ__/page-objects/async-dropdown-page.ts
```diff
@@ -25,7 +25,7 @@ declare const window: ExtendedWindow;
 export const DEBOUNCE_FILTERING_DELAY = 300;
 export const RESPONSE_PAGE_SIZE = 20;
 
-export class AsyncDropdownPage extends BasePageObject {
```
just-boris, line 28: Here and in many places below – if an export is not used anywhere in our code and not a part of our public API, remove the export.

After that, ESLint started detecting unused code which could be removed

src/internal/keycode.ts
```diff
@@ -17,8 +17,4 @@ export enum KeyCode {
   escape = 27,
   enter = 13,
   tab = 9,
-  shift = 16,
-  control = 17,
-  alt = 18,
-  meta = 91,
```
just-boris, line 23: Never used enum members. Actually, we should cease the entire `KeyCode` enum and adopt `event.key` instead.

--
P.S. Enum members do not get minified, so it makes sense to remove unused code here just for this reason

src/mixed-line-bar-chart/interfaces.ts
```diff
@@ -6,9 +6,7 @@ import { CartesianChartProps } from '../internal/components/cartesian-chart/inte
 
 export type ChartDataTypes = number | string | Date;
 
-export type SeriesType = 'line' | 'bar';
```
just-boris, line 9: Never used. The actually used type for this use case is declared at `IDataSeries['type']`

pages/property-filter/custom-forms.tsx
```diff
@@ -3,14 +3,15 @@
 
 import React, { useEffect, useState } from 'react';
 
+import { PropertyFilterOperatorFormProps } from '@cloudscape-design/collection-hooks';
+
 import { DatePicker, FormField, RadioGroup, TimeInput, TimeInputProps } from '~components';
 import Calendar, { CalendarProps } from '~components/calendar';
 import DateInput from '~components/date-input';
-import { ExtendedOperatorFormProps } from '~components/property-filter/interfaces';
```
just-boris, line 9: These two are the same type. Replaced private type with its public equivalent

avinashbot (commented):
Need to look at it more thoroughly, but I'm not sure how I feel about un-exporting the props interfaces objects for basically all our internal components. I like having that convention of exporting `ComponentName` and `ComponentNameProps` from component files.

just-boris:
I explained it in my pre-review comment: https://github.com/cloudscape-design/components/pull/3081#discussion_r1864659773

There were some unused types which were hanging in the code only because they were exported. This is more useful than keeping an export "just in case". You can always restore the export of `ComponentNameProps` type if you actually have a use for it

avinashbot (dismissed):

avinashbot (approved):

PR #3086, opened 2024-12-02
feat: Make split panel visible when collapsed at the bottom in new toolbar layout

jperals:
### Description

Issue: `AWSUI-60031`

Now the split panel in the new toolbar layout behaves the same as in the old layout when it is displayed at the bottom:
- When collapsed:
  - it does not completely disappear but instead the header and the open button keep being visible overlaying the bottom part of the content
  - the focus moves to the split panel's own open button instead of the toolbar's split panel trigger
- The `findSplitPanelOpenButton` test util will return the open button in the split panel and not the toolbar split panel trigger

### How has this been tested?

- Manually on dev pages
- Adapted existing tests
- Ran through my pipeline, got an expected diff
- Dry run (`7138307483`)

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/__integ__/toolbar-tooltips/app-layout-toolbar-split-panel-trigger-tooltip.test.ts
+++ b/src/app-layout/__integ__/toolbar-tooltips/app-layout-toolbar-split-panel-trigger-tooltip.test.ts
@@ -1,8 +1,9 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
-import testutilStyles from '../../../../lib/components/app-layout/test-classes/styles.selectors';
 import createWrapper from '../../../../lib/components/test-utils/selectors';
 import { AppLayoutDrawersPage, setupTest } from '../utils';
+
+import testutilStyles from '../../../../lib/components/app-layout/test-classes/styles.selectors.js';
 
 const wrapper = createWrapper().findAppLayout();
 
--- a/src/app-layout/__integ__/toolbar-tooltips/app-layout-toolbar-split-panel-trigger-tooltip.test.ts
+++ b/src/app-layout/__integ__/toolbar-tooltips/app-layout-toolbar-split-panel-trigger-tooltip.test.ts
@@ -60,7 +60,7 @@ describe('refresh-toolbar', () => {
         'Shows tooltip correctly for split panel trigger for keyboard (tab) interactions',
         setupTest({ theme, size, splitPanelPosition }, async (page: AppLayoutDrawersPage) => {
           await expect(page.isExisting(tooltipSelector)).resolves.toBe(false);
-          await page.click(createWrapper().findAppLayout().findToolbar().toSelector());
+          await page.click(wrapper.findToolbar().toSelector());
           await page.keys(['Tab', 'Tab']);
           await expect(page.isFocused(splitPanelTriggerSelector)).resolves.toBe(true);
           await expect(page.getText(tooltipSelector)).resolves.toBe(expectedTooltipText);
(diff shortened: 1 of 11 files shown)


just-boris:
This change broke split panel slide in animation. 

It works fine in the mainline and does not on your PR preview build: https://d21d5uik3ws71m.cloudfront.net/components/783ac3514ee5eb19adb067e5c023a114f456691e/dev-pages/index.html#/light/app-layout/with-split-panel?appLayoutToolbar=true

jperals:
> This change broke split panel slide in animation.

Thanks for noticing! This was addressed in the latest commit ([c074481](https://github.com/cloudscape-design/components/pull/3086/commits/c074481806f99ea930299e4f3b0341d94d097884)).

src/app-layout/__integ__/toolbar-tooltips/app-layout-toolbar-split-panel-trigger-tooltip.test.ts
```diff
@@ -12,10 +14,21 @@ describe('refresh-toolbar', () => {
...
+    const splitPanelTriggerSelector = wrapper.findByClassName(testutilStyles['split-panel-trigger']).toSelector();
     const tooltipSelector = wrapper.findDrawerTriggerTooltip().toSelector();
     const expectedTooltipText = 'Open panel';
 
+    async function focusSplitPanelTriggerWithKeyboard(page: AppLayoutDrawersPage) {
+      await expect(page.isExisting(tooltipSelector)).resolves.toBe(false);
+      // Click on the last breadcrumb (which is not a link)
+      await page.click(wrapper.findBreadcrumbs().findBreadcrumbGroup().findBreadcrumbLink(2).toSelector());
+      await page.keys(['Tab', 'Tab']); // Focus the first drawer trigger past the split panel trigger
+      await page.keys(['Shift', 'Tab']); // Tab back to the split panel trigger
```
just-boris, line 26: Do we need to do this back-and-forward focus in every test? Is it really worth reusing this code? For example, you could only reuse the assertion part.

src/app-layout/visual-refresh-toolbar/toolbar/drawer-triggers.tsx
```diff
@@ -135,12 +136,13 @@ export function DrawerTriggers({
               className={clsx(
                 styles['drawers-trigger'],
                 testutilStyles['drawers-trigger'],
-                splitPanelTestUtilStyles['open-button']
+                testutilStyles['split-panel-trigger'],
+                splitPanelResolvedPosition === 'side' && splitPanelTestUtilStyles['open-button']
```
just-boris, line 140: Let's keep this test util class assigned in both cases. 

The collapsed split panel will not use this class. All customers' tests will open split panel using this button in every position

src/split-panel/bottom.tsx
```diff
@@ -45,7 +46,14 @@ export function SplitPanelContentBottom({
   const isMobile = useMobile();
 
   const headerRef = useRef<HTMLDivElement>(null);
-  useResizeObserver(headerRef, entry => reportHeaderHeight(entry.borderBoxHeight));
+  const closedPanelBlockSize = useRef<string>();
+
+  useResizeObserver(headerRef, entry => {
+    const { borderBoxHeight } = entry;
+    closedPanelBlockSize.current = `calc(${borderBoxHeight}px + ${tokens.borderPanelTopWidth})`;
```
just-boris, line 53: This line does not look like idiomatic React code. What is happening here?

jperals: I am using `closedPanelBlockSize.current` as `blockSize` in line 86 in order to keep the animated transition. This gives the element the same height that it would already have without applying any value, _but_ a value needs to be explicitly set because you cannot transition from or to undefined; a value has to be defined for the transitioned property in both states.

What do you find most non-idiomatic here? Do you have any suggestion? Or does it look better knowing this information?

just-boris: For idiomatic code, it should be a state, not ref. Also, can we reuse `reportHeaderHeight` somehow which already sets the state?

jperals: We can pass down `reportHeaderHeight` from the app layout, although it is only needed for the toolbar layout, so either we have to make it optional in the split panel context, or we pass it anyway in the three app layouts, whether it is actually used in the split panel or not.

just-boris (approved):

PR #3090, opened 2024-12-02
chore: Fix multi modal funnel support

michaeldowseza:
### Description

Currently the funnel name selector was resolving to the first modal header it could find. If there are multiple modals on the page, even if the modal is not visible its header could still be found in the dom. This change adds a unique id for each modal associated with a funnel for the selector to resolve correctly.

### How has this been tested?

Added unit test

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/modal/index.tsx
+++ b/src/modal/index.tsx
@@ -41,7 +41,7 @@ function ModalWithAnalyticsFunnel({
       funnelType="modal"
       optionalStepNumbers={[]}
       totalFunnelSteps={1}
-      funnelNameSelectors={[`[${DATA_ATTR_MODAL_ID}="${modalId}"] .${styles['header--text']}`]}
+      funnelNameSelectors={[`[${DATA_ATTR_MODAL_ID}="${CSS.escape(modalId)}"] .${styles['header--text']}`]}
     >
       <AnalyticsFunnelStep
         mounted={props.visible}
--- a/src/modal/index.tsx
+++ b/src/modal/index.tsx
@@ -8,8 +8,10 @@ import {
   AnalyticsFunnelSubStep,
 } from '../internal/analytics/components/analytics-funnel';
 import { useFunnel } from '../internal/analytics/hooks/use-funnel';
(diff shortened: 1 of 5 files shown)


src/modal/index.tsx
```diff
@@ -34,7 +41,7 @@ function ModalWithAnalyticsFunnel({
       funnelType="modal"
       optionalStepNumbers={[]}
       totalFunnelSteps={1}
-      funnelNameSelectors={[`.${styles['header--text']}`]}
+      funnelNameSelectors={[`[${DATA_ATTR_MODAL_ID}="${modalId}"] .${styles['header--text']}`]}
```
connorlanigan, line 44: Since we do not have any guarantees about what characters `modalId` can contain, I suggest that we use [CSS.escape](https://developer.mozilla.org/en-US/docs/Web/API/CSS/escape_static) to make it safe to use in this CSS selector

connorlanigan (changes requested):

src/modal/index.tsx
connorlanigan: "safe to pass to attributes" is not the same as "safe to pass unescaped into a CSS selector" though. See e.g. [this MDN paragraph](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id#syntax) describing that an ID (and thus the value generated by React's `useId`) can be essentially anything that does not contain whitespace:
> An ID attribute's value must not contain ASCII whitespace characters. [...] Technically, the value for an ID attribute may contain any other Unicode character. However, when used in CSS selectors, either from JavaScript using APIs like `Document.querySelector()` or in CSS stylesheets, ID attribute values must be valid CSS identifiers. This means that if an ID attribute value is not a valid CSS identifier (for example, `my?id` or `1234`) then it must be escaped before being used in a selector, either using the `CSS.escape()` method or manually.

michaeldowseza: Fair point. Changed.

connorlanigan (approved):

PR #3091, opened 2024-12-02
refactor: Internal dnd container

pan-kot:
### Description

Extracted drag-&-drop utilities from collection preferences to internal utils and added a test page to use them for a custom semantic list, table, and containers list.

### How has this been tested?

* Existing tests including screenshot to ensure no regression for collection preferences
* Manual testing on collection preferences and new dnd test pages

<details>

#### Correctness

#### Security

#### Testing

</details>

---


pages/dnd/reorderable-list.tsx
```diff
@@ -0,0 +1,105 @@
...
+          {renderOption(props)}
+        </li>
+      )}
+      renderActiveOption={props => <Box>{renderOption(props)}</Box>}
+      i18nStrings={i18nStrings}
+    >
+      {content => (
+        <ul className={styles.list} aria-label="re-orderable list" role="list">
+          {staticOptions.map(option => (
+            <React.Fragment key={option.id}>{renderStaticOption?.(option)}</React.Fragment>
```
jperals, line 59: Minor as this is a dev page: when the first 3 items are marked as fixed:
- Their visual height decreases by about 1px
- They become divs immediately nested in an unordered list (whereas technically only `li`, `script` or `template` are allowed: see [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ul#technical_summary))

src/internal/components/dnd-container/index.tsx
```diff
@@ -0,0 +1,164 @@
...
+export function DndContainer<Option>({
+  getId,
+  sortedOptions,
+  onChange,
+  disabled,
+  renderOption,
+  renderActiveOption,
+  i18nStrings,
+  dragOverlayClassName = styles['drag-overlay'],
+  children = content => content,
```
jperals, line 50: Do we need this ability to customize the rendering in this way? So far it is used only in the newly added example dev page and it looks to me like the same could be achieve by rendering the DndContainer inside the `tbody` instead of it wrapping the table.

pages/dnd/reorderable-list.tsx
jperals: Does `renderStaticOption` actually bring much? You can just customize the rendering in `renderOption`

jperals: > Their visual height decreases by about 1px

This seems to be because of [this 1px transparent bottom border](https://github.com/cloudscape-design/components/pull/3091/files#diff-2f72ef25158ce8220ac390da383d63cfc322a7ba26a7efabef5c19a726786259R42) applied when sorting.

pan-kot: Great observation! I will fix that by adding the li with style.option to make the rendering identical.

Yes, the renderOption can be customised but I want the first N items to be fixed in their positions. If I just disable the drag handles while rendering the options inside dnd container the user will be still able to change the order of those by moving other items around.

src/internal/components/dnd-container/index.tsx
pan-kot: I didn't use it in collection preference to not cause accidental regression. However, the reason it is done this way is to prevent DndContext to render the live region inside a semantic container like <ul> or <tbody>. There is already this issue in the collection prefs but it is not captured by AXE because the collection prefs is not open by default in the test page.

pages/dnd/reorderable-list.tsx
jperals: OK, thanks for explaining. Still a very minor, non-blocking issue: while dragging a list item, the list becomes 1px shorter —the last item probably has a 1px transparent bottom border only while not dragging.

jperals (approved):

pages/dnd/reorderable-list.tsx
pan-kot: Thanks, updated!

jperals: Again, not blocking this because it's an internal dev page, but now the borders between list items have rounded corners :)

jperals (approved):

PR #3107, opened 2024-12-06
chore: Enhancing date-range-picker pages

dpitcock:
### Description

This adds more controls to the date-range-picker pages so we can toggle on and off features. It also accepts urlParams to toggle the features too so specific functionality can be shared.

The updated pages will make it easier to add additional changes for when there is a granularity prop added to the date-range-picker component

### How has this been tested?

Integ tests updated to reflect new urls and utilize params accordingly
Updates to internal Integ Test made to match this package in CR-165057689

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/date-range-picker/time-offset.ts
```diff
@@ -39,11 +39,12 @@ export function shiftTimeOffset(
 
   /*
     This regex matches an ISO date-time with
+    - optionally matches date-only format
```
dpitcock, line 42: date-only format is still considered a valid ISO 8601 representation. It represents a calendar date without any time information. This format will be used when only the date is relevant, and the time doesn't need to be specified.

pages/date-range-picker/custom-control.page.tsx
```diff
@@ -1,33 +1,86 @@
...
           onChange={e => setValue(e.detail.value)}
           locale={enLocale.code}
-          i18nStrings={i18nStrings}
+          i18nStrings={generateI18nStrings(dateOnly, monthOnly)}
           relativeOptions={[]}
-          placeholder="Filter by a date and time range"
+          placeholder={generatePlaceholder(dateOnly, monthOnly)}
           isValidRange={isValid}
           rangeSelectorMode="absolute-only"
+          isDateEnabled={(date: Date) => checkIfDisabled(date, disabledDates, monthOnly)}
```
just-boris, line 82: This argument type is not needed because `isDateEnabled` is already typed on  its own

pages/date-range-picker/with-default-date.page.tsx
```diff
@@ -1,43 +0,0 @@
...
-    endDate: '2018-01-19T15:30:00Z',
-  });
-
-  return (
-    <Box padding="s">
-      <h1>Date range picker with selected date</h1>
-      <Link id="focus-dismiss-helper">Focusable element before the date range picker</Link>
-      <br />
-      <br />
-      <ScreenshotArea>
```
just-boris, line 23: If there is a `<ScreenshotArea>`, it is likely used in some screenshot tests. Did you clean them up before merging this?

src/date-range-picker/time-offset.ts
just-boris: This is a functional change, but the PR is called "Enhancing date-range-picker pages" which does not match it.

1. Do you want to extract it in another PR and annotate properly and add relevant tests?
2. Is there any other reason to bundle these two changes together?

pages/date-range-picker/simple.page.tsx
```diff
@@ -13,15 +13,40 @@ import {
...
+  const showRelativeOptions = urlParams.showRelativeOptions ?? dateRangePickerDemoDefaults.showRelativeOptions;
+  const dateOnly = urlParams.dateOnly ?? dateRangePickerDemoDefaults.dateOnly;
+  const disabledDates =
+    (urlParams.disabledDates as DisabledDate) ?? (dateRangePickerDemoDefaults.disabledDates as DisabledDate);
+  const withDisabledReason = urlParams.withDisabledReason ?? dateRangePickerDemoDefaults.withDisabledReason;
+  const invalid = urlParams.invalid ?? dateRangePickerDemoDefaults.invalid;
+  const warning = urlParams.warning ?? dateRangePickerDemoDefaults.warning;
+  const rangeSelectorMode =
+    urlParams.rangeSelectorMode ??
+    (dateRangePickerDemoDefaults.rangeSelectorMode as DateRangePickerProps.RangeSelectorMode);
```
just-boris, line 49: Wouldn't it be easier to make all defaults false/undefined, instead of writing this wall of text for every page?

src/date-range-picker/time-offset.ts
dpitcock: Removed these changes as it is no longer needed after [this change](https://github.com/cloudscape-design/components/blob/main/src/date-range-picker/utils.tsx#L82) was added in a [recent PR](https://github.com/cloudscape-design/components/pull/3111/files?w=1 )

pages/date-range-picker/simple.page.tsx
dpitcock: It's a tradeoff. I understand your approach. While it may seem verbose, using centralized defaults in dateRangePickerDemoDefaults allows us to maintain consistency across multiple pages and easily update default values in one place. This approach reduces the risk of inconsistencies and makes maintenance easier in the long run. It's worth noting that specific initial values can be assigned in place of the defaults if desired for any particular page. Also, not all pages have this 'wall of text,' as many don't allow for changes to as many props, resulting in a more concise implementation on those pages.

pages/date-range-picker/with-default-date.page.tsx
dpitcock: All tests that run against `<...>/index.html#/light/date-range-picker/with-default-date` and `<...>/index.html#/light/date-range-picker/with-default-date-only` will continue to run but this time against the same page`<...>/index.html#/light/date-range-picker/with-value` with params added to load the page with the component as date-only or not`?dateOnly=true` and `?dateOnly=false`. So the same screenshots will exist. However as this PR  is merged, the regression tests will see the former as removed, and those after it as new, even though they are visually the same

pages/date-range-picker/with-value.page.tsx
```diff
@@ -0,0 +1,80 @@
...
+        checked={dateOnly}
+        onChange={({ detail }) => setUrlParams({ dateOnly: detail.checked })}
+      >
+        Date-only
+      </Checkbox>
+      <br />
+      <Link id="focus-dismiss-helper">Focusable element before the date range picker</Link>
+      <br />
+      <br />
+      <ScreenshotArea>
```
dpitcock, line 58: This is the ScreenshotArea that will be used to continue testing for both default-date and default-date-only tests when the url with the right params are passed into this page

pages/date-range-picker/simple.page.tsx
dpitcock: Other integ tests screenshots added in CR-165057689

just-boris (approved):

PR #3111, opened 2024-12-09
fix: allowing dateOnly prop to format value string in date-range-picker

dpitcock:
### Description

Currently the date-range-picker does not format the date when the dateOnly prop is true. It assumes the provided value is dateOnly, but that is not always the case. With this code, we check the initial format and updated it to reflect  as dateOnly when the value is providing the time too (which is not needed)

This fixes an a bug that can be seen on the [date only page](https://d1dnby8l05mrxs.cloudfront.net/polaris/index.html#/light/date-range-picker/with-default-date-only) . It also fortifies the testing of the many of utils and time-offset functions

### How has this been tested?
- tests added

<details>

#### Correctness

#### Security

#### Testing

</details>

---


dpitcock:
Pulled changes into another pull-request

src/date-range-picker/utils.tsx
```diff
@@ -1,15 +1,20 @@
...
 
 export function formatValue(
   value: null | DateRangePickerProps.Value,
   { timeOffset, dateOnly }: { timeOffset: { startDate?: number; endDate?: number }; dateOnly: boolean }
 ): null | DateRangePickerProps.Value {
-  if (!value || value.type === 'relative') {
+  if (
+    !value ||
+    value.type === 'relative' ||
+    (value.endDate === '' && value.endDate === '' && value.type === 'absolute')
```
connorlanigan, line 16: We're checking `endDate` here twice, I assume one of these should be `startDate`

src/date-range-picker/index.tsx
```diff
@@ -141,7 +133,9 @@ const DateRangePicker = React.forwardRef(
     checkControlled('DateRangePicker', 'value', value, 'onChange', onChange);
 
     const normalizedTimeOffset = normalizeTimeOffset(value, getTimeOffset, timeOffset);
-    value = isDateOnly(value) ? value : shiftTimeOffset(value, normalizedTimeOffset);
+    // value = isDateOnly(value) ? value : shiftTimeOffset(value, normalizedTimeOffset);
```
connorlanigan, line 136: This line can then be removed

src/date-range-picker/__tests__/utils.test.ts
```diff
@@ -0,0 +1,295 @@
...
+    });
+
+    it('should return shifted time offset for relative value', () => {
+      const relativeValue: DateRangePickerProps.RelativeValue = { type: 'relative', amount: 7, unit: 'day' };
+      const shiftedValue: DateRangePickerProps.AbsoluteValue = {
+        type: 'absolute',
+        startDate: '2023-01-01',
+        endDate: '2023-01-07',
+      };
+      ShiftTimeOffsetSpy.mockReturnValue(shiftedValue);
```
connorlanigan, line 209: My understanding is that `shiftTimeOffset` will _not_ shift a relative value ([code](https://github.com/cloudscape-design/components/blob/edb6713937169bbd4c25efb540fa6361f849bb2d/src/date-range-picker/time-offset.ts#L36-L38)) - this mocked value seems to be a value that will never be returned in the actual component, it will be the `relativeValue` itself. Can we update this test?

connorlanigan (changes requested):

src/date-range-picker/__tests__/utils.test.ts
dpitcock: Because the shiftTimeOffset was mocked it never went into that branch. I added more tests to the shiftTimeOffset function to make assertions when there is a relative  type and here I have the mock returning the correct value

connorlanigan (approved):

PR #3113, opened 2024-12-09
refactor: Improves dnd container API

pan-kot:
### Description

A follow-up for https://github.com/cloudscape-design/components/pull/3091.

The refactoring improves D&D container API:
* Makes API similar to the Board component's API with items + onItemsChange + renderItem;
* Removes dependencies on dnd-kit types;
* Some properties renaming.

Besides, the update includes a few small a11y fixes such as adding missing ARIA description to the drag handle and making accessibility nodes render into a portal to not violate HTML structure.

### How has this been tested?

* Existing tests and a few new basic tests for D&D container

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/collection-preferences/content-display/__tests__/content-display.test.tsx
```diff
@@ -517,13 +519,13 @@ function testOption({
...
-  expectAriaLabel(wrapper, dragHandle, `Drag handle, Item ${index + 1}`);
+  expectAriaLabel(wrapper, dragHandle, `Drag handle Item ${index + 1}`);
   expectLabelForToggle(option);
 }
 
-async function expectAnnouncement(wrapper: ContentDisplayPreferenceWrapper, announcement: string) {
+async function expectAnnouncement(announcement: string) {
   await new Promise(resolve => setTimeout(resolve, 0));
-  const liveRegion = wrapper.find('[aria-live="assertive"]');
+  const liveRegion = createWrapper().find('[aria-live="assertive"]');
```
pan-kot, line 528: The aria-live was moved to the portal node, hence searching from the root.

src/collection-preferences/content-display/content-display-option.tsx
```diff
@@ -13,28 +13,22 @@ const componentPrefix = 'content-display-option';
 export const getClassName = (suffix?: string) => styles[[componentPrefix, suffix].filter(Boolean).join('-')];
 
 interface ContentDisplayOptionProps {
-  dragHandleAriaLabel?: string;
-  listeners?: DragHandleProps['listeners'];
+  dragHandleAttributes: DragHandleProps['attributes'];
+  dragHandleListeners?: DragHandleProps['listeners'];
   onToggle?: (option: OptionWithVisibility) => void;
   option: OptionWithVisibility;
-  disabled?: boolean;
```
pan-kot, line 20: The disabled is replaced with aria-disabled that comes as part of the dragHandleAttributes.

src/collection-preferences/content-display/content-display-option.tsx
```diff
@@ -13,28 +13,22 @@ const componentPrefix = 'content-display-option';
...
   (
-    { dragHandleAriaLabel, listeners, onToggle, option, disabled }: ContentDisplayOptionProps,
+    { dragHandleAttributes, dragHandleListeners, onToggle, option }: ContentDisplayOptionProps,
     ref: ForwardedRef<HTMLDivElement>
   ) => {
     const idPrefix = useUniqueId(componentPrefix);
     const controlId = `${idPrefix}-control-${option.id}`;
-
-    const dragHandleAttributes = {
-      ['aria-label']: [dragHandleAriaLabel, option.label].join(', '),
```
pan-kot, line 32: The joining of the drag handle label and option label now happens inside the dnd container and the updated ARIA label comes as part of the dragHandleAttributes.

src/collection-preferences/content-display/index.tsx
```diff
@@ -130,37 +130,20 @@ export default function ContentDisplayPreference({
...
-          )}
-          renderActiveOption={props => (
-            <ContentDisplayOption
-              option={props.option}
-              dragHandleAriaLabel={props.dragHandleAriaLabel}
-              listeners={props.listeners}
-              onToggle={onToggle}
-            />
-          )}
+          items={sortedAndFilteredOptions.map(data => ({ id: data.id, label: data.label, data }))}
```
pan-kot, line 133: I changed option + getOptionId to items, where the ID and the label are provided as separate properties. That is similar to what we do in the Board component. An advantage is that item's metadata can be extended as necessary. The label is used to generate the ARIA labels for drag handles which is a combination of the i18n drag handle aria label and option label.

src/internal/components/dnd-container/index.tsx
```diff
@@ -1,61 +1,39 @@
...
+  const { activeItemId, setActiveItemId, collisionDetection, handleKeyDown, sensors } = useDragAndDropReorder({
+    items,
   });
-  const activeOption = activeItem ? sortedOptions.find(option => getId(option) === activeItem) : null;
-  const isDragging = activeItem !== null;
-  const announcements = useLiveAnnouncements({ sortedOptions, getId, isDragging, ...i18nStrings });
+  const activeItem = activeItemId ? items.find(item => item.id === activeItemId) : null;
+  const isDragging = activeItemId !== null;
+  const announcements = useLiveAnnouncements({ items, isDragging, ...i18nStrings });
+  const accessibilityContainer = useMemo(() => document.createElement('div'), []);
```
pan-kot, line 31: By default, the DndContext renders accessibility nodes (for ARIA description and announcements) in-place. That creates an HTML violation when using the context for semantic elements such as lists or tables (this is an existing issue with collection prefs).

To address the issue, the previous implementation used render-prop children. The current implementation uses the accessibility.container prop of the DndContext that specifies location where the a11y props must go. The accessibility.container takes an Element which does not exist during the first render unless provided from the outside. To solve this, we create the element with document.createElement() and append it to the portal node when the portal becomes available.

src/collection-preferences/content-display/index.tsx
```diff
@@ -130,37 +130,20 @@ export default function ContentDisplayPreference({
...
-              )}
-              style={props.style}
-            >
-              <ContentDisplayOption
-                ref={props.ref}
-                listeners={props.listeners}
-                dragHandleAriaLabel={props.dragHandleAriaLabel}
-                onToggle={onToggle}
-                option={props.option}
-                disabled={props.attributes['aria-disabled']}
```
pan-kot, line 152: From all attributes that DndContext provides the option only takes aria-disabled as we don't really need all like the role and tabIndex etc. However, it also does not take the aria-describedby which comes from the i18nStrings.dragHandleAriaDescription. As result, the ARIA description was not used at all.

jperals: > the ARIA description was not used at all (an existing bug of collection prefs)

The ARIA description is passed to dnd-kit's `DndContext.screenReaderInstructions`, which enables screen reader announcements of the dragging instructions when focusing the drag handle. Is it necessary that it is an attribute of the drag handle as well?

pan-kot: I checked in the html output for the ARIA description content and it looks like that it is only added as a hidden div which is not referenced from anywhere in the current main:

```
<div id="DndDescribedBy-2" style="display: none;">
  Use Space or Enter to activate drag for an item, then use the arrow keys to move
  the item's position. To complete the position move, use Space or Enter, or to
  discard the move, use Escape.
</div>
```

jperals: That's how dnd-kit manages the announcement of these instructions, by just adding a visually hidden div and a live region to the doc ([link](https://github.com/clauderic/dnd-kit/blob/e9215e820798459ae036896fce7fd9a6fe855772/packages/core/src/components/Accessibility/Accessibility.tsx#L66-L69)). If we want to do use this string also as the ARIA description for the drag handle as well, won't we have them announced twice?

pan-kot: As seen by the link that you shared, the description is not added as a live region - but as a hidden text. The hidden text is not announced to the screen-readers unless assigned with aria-labelledby or aria-describedby.

pan-kot: Sure, I can dive deeper and do more changes but preferably not in scope of the current PR. I plan a few more upgrades to the dnd-container and can see if some improvements can be incorporated. So far the current behaviour looks ok - the description and the announcements are all there.

jperals (approved):

PR #3116, opened 2024-12-09
fix: setting update offset to handle date-only format

dpitcock:
### Description

Currently the regex check within the setTimeOffset function in the date-range-picker returns false for date only dates despit them being considered a valid ISO 8601 representation. It represents a calendar date without any time information. This format will be used when only the date is relevant, and the time doesn't need to be specified

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/date-range-picker/__tests__/time-offset.test.ts
+++ b/src/date-range-picker/__tests__/time-offset.test.ts
@@ -9,6 +9,30 @@ describe('Date range picker', () => {
     describe('setTimeOffset', () => {
       test.each([
         [
+          'absolute date-only dates with no offset',
+          { type: 'absolute', startDate: '2020-10-12', endDate: '2020-10-12' },
+          { startDate: undefined, endDate: undefined },
+          { type: 'absolute', startDate: '2020-10-12+00:00', endDate: '2020-10-12+00:00' },
+        ],
+        [
+          'absolute date-only dates with positive offset',
+          { type: 'absolute', startDate: '2020-10-12', endDate: '2020-10-12' },
+          { startDate: -240, endDate: -240 },
+          { type: 'absolute', startDate: '2020-10-12-04:00', endDate: '2020-10-12-04:00' },
+        ],
+        [
+          'absolute date-only dates with positive offset',
+          { type: 'absolute', startDate: '2020-10-12', endDate: '2020-10-12' },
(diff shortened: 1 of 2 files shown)


src/date-range-picker/__tests__/time-offset.test.ts
```diff
@@ -8,6 +8,30 @@ describe('Date range picker', () => {
   describe('time offset handling', () => {
     describe('setTimeOffset', () => {
       test.each([
+        [
```
pan-kot, line 11: it looks like the tests are added for setTimeOffset but the change is made to shiftTimeOffset

src/date-range-picker/time-offset.ts
```diff
@@ -39,11 +39,12 @@ export function shiftTimeOffset(
 
   /*
     This regex matches an ISO date-time with
+    - optionally matches date-only format
     - optional seconds;
     - optional milliseconds;
     - optional time offset or 'Z'.
   */
-  const dateTimeRegex = /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}(:\d{2})?(\.\d{1,3})?(((\+|-)\d{2}(:\d{2})?)|Z)?$/;
+  const dateTimeRegex = /^\d{4}-\d{2}-\d{2}(T\d{2}:\d{2}(:\d{2})?(\.\d{1,3})?(((\+|-)\d{2}(:\d{2})?)|Z)?)?$/;
```
pan-kot, line 47: we do we even need the shiftTimeOffset for date-only values?

dpitcock: Removed this changes as it is no longer needed after this change https://github.com/cloudscape-design/components/pull/3111/files?w=1

src/date-range-picker/__tests__/time-offset.test.ts
dpitcock: Removed these changes as it is no longer needed after this change https://github.com/cloudscape-design/components/pull/3111/files?w=1

dpitcock:
Change is no longer necessary as this change ensures when the value is absolute and date-only the  https://github.com/cloudscape-design/components/pull/3111/files?w=1 shiftTimeOffset function is not called

PR #3122, opened 2024-12-10
chore: Add loading prop to Property and Text Filter to handle re-announcement of countText

johannes-weber:
### Description

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/property-filter/count-text.page.tsx
+++ b/pages/property-filter/count-text.page.tsx
@@ -1,0 +1,146 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import React from 'react';
+
+import Button from '~components/button';
+import { I18nProvider } from '~components/i18n';
+import messages from '~components/i18n/messages/all.en';
+import PropertyFilter, { PropertyFilterProps } from '~components/property-filter';
+import SpaceBetween from '~components/space-between';
+
+export default function () {
+  const [query, setQuery] = React.useState<PropertyFilterProps['query']>({
+    tokens: [
+      {
+        operator: '=',
+        propertyKey: 'instanceid',
+        value: 'i-2dc5ce28a0328391',
+      },
+    ],
+    operation: 'and',
+  });
+  const [countText, setCountText] = React.useState('');
+  const [loading, setLoading] = React.useState(false);
+  const propertyFilterRef = React.useRef<PropertyFilterProps.Ref>(null);
+
+  return (
+    <I18nProvider messages={[messages]} locale="en">
+      <h1>Demo page for countText live announcement testing</h1>
+      <SpaceBetween size="l">
+        <PropertyFilter
+          query={query}
+          onChange={({ detail }) => setQuery(detail)}
+          countText={countText}
+          ref={propertyFilterRef}
+          expandToViewport={true}
+          filteringAriaLabel="Find distributions"
+          loading={loading}
+          filteringOptions={[
+            {
+              propertyKey: 'instanceid',
+              value: 'i-2dc5ce28a0328391',
+            },
+            {
+              propertyKey: 'instanceid',
+              value: 'i-d0312e022392efa0',
+            },
+            {
+              propertyKey: 'instanceid',
+              value: 'i-070eef935c1301e6',
+            },
+            {
+              propertyKey: 'instanceid',
+              value: 'i-3b44795b1fea36ac',
+            },
+            { propertyKey: 'state', value: 'Stopped' },
+            { propertyKey: 'state', value: 'Stopping' },
+            { propertyKey: 'state', value: 'Pending' },
+            { propertyKey: 'state', value: 'Running' },
+            {
+              propertyKey: 'instancetype',
+              value: 't3.small',
+            },
+            {
+              propertyKey: 'instancetype',
+              value: 't2.small',
+            },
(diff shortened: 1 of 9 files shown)


src/text-filter/__tests__/use-debounce-search-result-callback.test.tsx
```diff
@@ -0,0 +1,125 @@
...
+      renderTestComponent({ announceCallback, loading: false, searchQuery: '', countText: undefined });
+      expect(announceCallback).not.toHaveBeenCalled();
+    });
+
+    test('should call announceCallback when searchQuery and countText are present', () => {
+      renderTestComponent({ announceCallback, loading: false, searchQuery: 'A', countText: '123 matches' });
+      expect(announceCallback).toHaveBeenCalledTimes(1);
+    });
+
+    test('should call announceCallback twice when searchQuery changes and countText stays the same', () => {
```
orangevolon, line 51: Maybe we can drop **twice** from the test title Because the test purpose is not to make sure it calls it twice, but more like to make sure it is announced when the search term is changing and the count text is unchanged.

src/text-filter/__tests__/text-filter.test.tsx
```diff
@@ -137,6 +140,52 @@ describe('countText', () => {
...
+      jest.runAllTimers();
+      expect(getPoliteRegion()).toBeNull();
+    });
+
+    test('includes the live announcement when loading switches from true to false', () => {
+      const { rerender } = renderTextFilter(<TextFilter filteringText="test" loading={true} countText="10 matches" />);
+      jest.runAllTimers();
+      expect(getPoliteRegion()).toBeNull();
+
+      rerender(<TextFilter filteringText="test" loading={false} countText="123 matches" />);
```
connorlanigan, line 174: In this test, the `countText` also changes from `10 matches` to `123 matches`. I suggest to use the same countText for both renders, so that we can be more certain that the live announcement is caused by the loading and not by the countText changing (we test that in the test below anyway)

src/text-filter/__tests__/text-filter.test.tsx
```diff
@@ -137,6 +140,52 @@ describe('countText', () => {
...
+      jest.runAllTimers();
+      expect(getPoliteRegion()).toHaveTextContent('123 matches');
+    });
```
connorlanigan, line 177: In this test, the `countText` also changes from `10 matches` to `123 matches`. I suggest to use the same countText for both renders, so that we can be more certain that the live announcement is caused by the loading and not by the countText changing (we test that in the test below anyway)

connorlanigan (changes requested):

pages/property-filter/count-text.page.tsx
```diff
@@ -0,0 +1,146 @@
...
+              operators: ['=', '!=', '>', '<', '<=', '>='],
+              propertyLabel: 'Average latency',
+              groupValuesLabel: 'Average latency values',
+            },
+          ]}
+        />
+        <Button onClick={() => setCountText('18 matches')}>Set count text to [18 matches]</Button>
+        <Button onClick={() => setCountText('36 matches')}>Set count text to [36 matches]</Button>
+        <Button onClick={() => setCountText('')}>Remove count text</Button>
+        <Button onClick={() => setLoading(prevIsLoading => !prevIsLoading)}>
```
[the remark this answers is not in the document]
johannes-weber: Addressed - Great suggestion, it makes toggling them more intuitive.

src/text-filter/__tests__/use-debounce-search-result-callback.test.tsx
johannes-weber: Addressed - Right, I've updated the test name + added an additional assertion after the first render to ensure the announcement is called on the initial render + rerender as expected.

src/text-filter/use-debounce-search-result-callback.ts
```diff
@@ -0,0 +1,39 @@
...
+      return;
+    }
+    announceCallback();
+  }, LIVE_REGION_DELAY);
+
+  useEffect(() => {
+    loadingRef.current = loading;
+    debounceLiveAnnouncement();
+  }, [searchQuery, countText, loading, debounceLiveAnnouncement]);
+}
```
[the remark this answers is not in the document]
johannes-weber: I agree. If think if there is a (auto) refresh mechanism for a collection in place, it means this collection *could* change after a refresh. Announcing the `countText` in this case informs users who use assistive technology about, even if there are no changes in the `countText`.

src/text-filter/internal.tsx
```diff
@@ -69,7 +82,11 @@ const InternalTextFilter = React.forwardRef(
           onChange={event => fireNonCancelableEvent(onChange, { filteringText: event.detail.value })}
           __onDelayedInput={event => fireNonCancelableEvent(onDelayedChange, { filteringText: event.detail.value })}
         />
-        {showResults ? <SearchResults id={searchResultsId}>{countText}</SearchResults> : null}
+        {showResults ? (
+          <SearchResults renderLiveRegion={!loading} id={searchResultsId} ref={searchResultsRef}>
```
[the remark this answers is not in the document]
johannes-weber: The announcement is happening only once. The live region element (`aria-live="polite"`) receives the latest value and announces it.

connorlanigan (approved):

PR #3138, opened 2024-12-13
chore: Fix elements scrolling with elementScrollTo in integration tests

Al-Dani:
### Description

* AWSUI-59931
* https://github.com/cloudscape-design/browser-test-tools/pull/118

### How has this been tested?

* used [browser-test-tools/scroll-validation](https://github.com/cloudscape-design/browser-test-tools/tree/scroll-validation) branch to run integration tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/__integ__/app-layout-drawers.test.ts
+++ b/src/app-layout/__integ__/app-layout-drawers.test.ts
@@ -9,6 +9,8 @@ import { viewports } from './constants';
 import { testIf } from './utils';
 
 const wrapper = createWrapper().findAppLayout();
+import vrDrawerStyles from '../../../lib/components/app-layout/visual-refresh/styles.selectors.js';
+import vrToolbarDrawerStyles from '../../../lib/components/app-layout/visual-refresh-toolbar/drawer/styles.selectors.js';
 
 class AppLayoutDrawersPage extends BasePageObject {
   async openFirstDrawer() {
@@ -242,7 +244,14 @@ describe.each(['classic', 'refresh', 'refresh-toolbar'] as const)('%s', theme =>
       setupTest({ theme }, async page => {
         await page.openFirstDrawer();
         const resizeHandleBefore = await page.getResizeHandlePosition();
-        await page.elementScrollTo(wrapper.findActiveDrawer().toSelector(), { top: 100 });
+        const scrollableContainer =
+          theme === 'classic'
+            ? wrapper.findActiveDrawer().toSelector()
+            : theme === 'refresh'
+              ? `.${vrDrawerStyles['drawer-content-container']}`
+              : `.${vrToolbarDrawerStyles['drawer-content-container']}`;
+
+        await page.elementScrollTo(scrollableContainer, { top: 100 });
         const resizeHandleAfter = await page.getResizeHandlePosition();
         await expect(resizeHandleAfter).toEqual(resizeHandleBefore);
       })
--- a/src/app-layout/__integ__/runtime-drawers.test.ts
+++ b/src/app-layout/__integ__/runtime-drawers.test.ts
@@ -7,6 +7,9 @@ import createWrapper, { AppLayoutWrapper } from '../../../lib/components/test-ut
 import { viewports } from './constants';
 import { getUrlParams, Theme } from './utils';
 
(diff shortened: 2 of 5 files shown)


src/cards/__integ__/sticky-header.test.ts
```diff
@@ -44,7 +44,7 @@ describe('Cards Sticky Header', () => {
   const scrollTopToBtn = '#scroll-to-top-btn';
   const toggleStickinessBtn = '#toggle-stickiness-btn';
   const toggleVerticalOffsetBtn = '#toggle-vertical-offset-btn';
-  const overflowParentPageHeight = 300;
```
Al-Dani, line 47: it isn't failing, but also doesn't match with actual height https://github.com/cloudscape-design/components/blob/main/pages/cards/sticky-header.page.tsx#L68

src/select/__integ__/select.test.ts
```diff
@@ -109,8 +109,6 @@ test(
     const { height: actualDropdownHeight } = await page.getBoundingBox(optionsSelector);
     const availableDropdownHeight = smallestContainerHeight - triggerHeight;
     expect(actualDropdownHeight).toBeLessThan(availableDropdownHeight);
-    const { top: containerScrollTop } = await page.getElementScroll('#smallest_container');
-    expect(containerScrollTop).toBe(0);
```
Al-Dani, line 113: that check was here since the beginning (`CR-16802930`, but it's not even when it was introduced, just refactored to `.ts`). 
It tries to scroll an element with `overflow: hidden` and I don't really understand the intention, so I just deleted

src/app-layout/__integ__/runtime-drawers.test.ts
```diff
@@ -117,7 +120,14 @@ describe.each(['classic', 'refresh', 'refresh-toolbar'] as Theme[])('%s', theme
         const getScrollPosition = () => page.getBoundingBox('[data-testid="drawer-sticky-header"]');
         const scrollBefore = await getScrollPosition();
 
-        await page.elementScrollTo(wrapper.findActiveDrawer().toSelector(), { top: 100 });
```
Al-Dani, line 120: the actual scrollable elements are different in Classic vs VR

src/app-layout/__integ__/app-layout-drawers.test.ts
```diff
@@ -242,7 +244,14 @@ describe.each(['classic', 'refresh', 'refresh-toolbar'] as const)('%s', theme =>
       setupTest({ theme }, async page => {
         await page.openFirstDrawer();
         const resizeHandleBefore = await page.getResizeHandlePosition();
-        await page.elementScrollTo(wrapper.findActiveDrawer().toSelector(), { top: 100 });
+        const scrollableContainer =
+          theme === 'classic'
+            ? wrapper.findActiveDrawer().toSelector()
+            : theme === 'refresh'
+              ? `.${vrDrawerStyles['drawer-content-container']}`
+              : `.${vrToolbarDrawerStyles['drawer-content-container']}`;
```
orangevolon, line 252: Do we need new test-utils selectors for these? How customers would scroll the elements in their tests?

Al-Dani: Yup, interesting question, I was thinking about it too.

We don't have any test utils to get scrollable containers except `findOptionsContainer` in select-alike components that says _"Use this element to **scroll** through the list of options"_.

Moreover, currently doing `page.elementScrollTo(wrapper.findActiveDrawer().toSelector(), { top: 100 });` in customer tests  means assuming that the returned element is scrollable. I don't think we support such assumptions.

It seems like a possible, but quite specific test-case where users need that. We could treat it as a feature request, but we haven't heard from customers yet that it's needed

orangevolon: I also don't think that this particular PR is blocked by this comment either, but if something is scrollable, I think it's fair to give the customers a way to scroll safely. The theme I've seen so far is that people make feature requests when they get really frustrated by something. Mostly they just use CSS selectors to get to the element they want to scroll.

orangevolon (approved):

PR #3143, opened 2024-12-16
feat: Tabs scroll snapping

georgylobko:
### Description

Reintroduced tabs scroll snapping reverted in [this commit](https://github.com/cloudscape-design/components/pull/3121).

The main issue with the previous implementation was that the custom scrollIntoView could scroll in a way that pushed the active element out of view.

This PR adjusts the scrollIntoView logic to always center the active element in the container, which should fix the problem.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/tabs/__integ__/tabs.test.ts
```diff
@@ -223,13 +223,25 @@ test(
     await page.setWindowSize({ width: 500, height: 1000 });
     await page.click('#add-tab');
     await page.click(page.paginationButton('right', true));
+    await page.click(page.paginationButton('right', true));
```
jperals, line 226: Why is it necessary to click once more in the right arrow now? Actually, it looks like the component scrolls a bit _further_ now than before, making the seventh tab visible with one single click whereas before it wasn't.

georgylobko: The component actually scrolls slightly less than before, which is why this condition is necessary

gethinwebster (approved):

PR #3146, opened 2024-12-17
refactor: Dnd area with placeholders

pan-kot:
### Description

A follow-up for: https://github.com/cloudscape-design/components/pull/3113

Also see: [a3UOAhaNX7KH]

The PR includes the following changes:
* DndContainer renamed to DndArea (because there is no visible styles that come with it)
* Border-radius style is included with the API (replacing custom class name prop)
* Placeholder is rendered by DndArea via render-item className
* Active item is assigned a default text style so that it displays correctly w/o need to wrap it in a Box
* DndArea is now providing animations for sortable items render-item className
* DragHandle API is better explicit and easier to use together with DndArea

### How has this been tested?

* Manual testing with collection preferences and dnd pages

<details>

#### Correctness

#### Security

#### Testing

</details>

---


pages/dnd/reorderable-list.tsx
```diff
@@ -41,14 +40,14 @@ export function ReorderableList<Option extends { id: string }>({
...
-      <DndContainer
+      <DndArea
         items={sortableOptions.map(option => ({ id: option.id, label: option.id, data: option }))}
         onItemsChange={items => onReorder([...staticOptions, ...items.map(item => item.data)])}
         renderItem={props => {
-          const className = clsx(props.isDragging && styles.placeholder, props.isSorting && styles.sorting);
+          const className = clsx(props.className, styles.option, props.isSorting && styles.sorting);
           const content = renderOption({ ...props, option: props.item.data });
           return props.isActive ? (
-            <Box>{content}</Box>
```
jperals, line 51: Was the previous version not a `li` element because it is the element being dragged and therefore outside of the `ul`?

pages/dnd/styles.scss
```diff
@@ -14,7 +14,13 @@
   margin-block: 0;
   margin-inline: 0;
 }
-
+.option {
```
jperals, line 17: Nitpick: missing spaces between rulesets

pages/dnd/styles.scss
```diff
@@ -35,49 +40,9 @@
...
-  > td:first-child {
-    border-start-start-radius: awsui.$border-radius-item;
-    border-end-start-radius: awsui.$border-radius-item;
-  }
-  > td:last-child {
-    border-start-end-radius: awsui.$border-radius-item;
-    border-end-end-radius: awsui.$border-radius-item;
-  }
+.row {
+  border-block-end: 1px solid transparent;
```
jperals, line 44: Why is this rule needed? Removing it in the browser dev tools doesn't seem to make any difference.

src/internal/components/dnd-area/styles.scss
```diff
@@ -0,0 +1,70 @@
...
+@use '@cloudscape-design/component-toolkit/internal/focus-visible' as focus-visible;
+
+@mixin animated {
+  @include styles.with-motion {
+    transition: transform awsui.$motion-duration-transition-quick awsui.$motion-easing-transition-quick;
+  }
+}
+
+@mixin drag-overlay-with-radius($border-radius) {
+  border-start-start-radius: $border-radius;
```
jperals, line 17: Shouldn't border radius styles be defined in user land rather than by the DndArea component?

pages/dnd/reorderable-list.tsx
pan-kot: That's a good point. I actually used the Box previously so that the font style is correct. The active node is rendered inside portal and does not have the font correctly inherited - I addressed that by adding `@include styles.default-text-style;` to the respective class name.

Ideally, I want the consumers to not depend on isActive and always return:
```
<li className={className} style={props.style}>
  {content}
</li>
```

That is possible for the list and container examples but not table where the column widths need to be set. I will dive deep to see if the semantics violation is a real problem here.

pages/dnd/styles.scss
pan-kot: Without this style the placeholder would overlap the bottom border

Screenshot 2024-12-18 at 15 44 03

src/internal/components/dnd-area/styles.scss
pan-kot: The DndArea uses className that the consumer assigns to sortable items to set the following styles:
1. Box shadow of the active item;
2. Focus outline of the active item;
3. Dragging item placeholder overlay;
4. Sortable items animation.

The box shadow, focus outline, and placeholder overlay styles require border-radius and those cannot inherit the border-radius of the item because are applied as separate absolute-positioned elements.

pages/dnd/reorderable-list.tsx
jperals: Thanks for the explanation.

> I will dive deep to see if the semantics violation is a real problem here.

Were you able to take a look at this point?

pan-kot: Yes, the semantic structure of the active node does not seem to have any effect. The active item is rendered in a portal and is detached from the actual list anyways - the screenreader users only receive the announcements really.

I propose that we use the simpler version to make the code better straightforward. Should we find a corner case with maybe some specific element we can change the integration slightly so that the active node is wrapped correctly - same as the active table row is wrapped with table and tbody now, yet it is done to ensure correct layout in that case.

jperals (approved):

PR #3147, opened 2024-12-17
fix: Hidden breadcrumbs issue

georgylobko:
### Description

This PR addresses an issue with dangling state for breadcrumbs in **nested** AppLayouts, which are not unmounting from the page on navigating away. See the video below to demonstrate the problem.
The root cause is that globally registered breadcrumbs didn't get cleaned up when the nested AppLayout is not visible. 
Now breadcrumbs are cleaned up as soon as the AppLayout component, which they belong to, becomes invisible in the viewport.

https://github.com/user-attachments/assets/b3bb901d-047b-4004-9d3c-f8a1db224bea

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/app-layout/utils/applayout-visibility-context.tsx
```diff
@@ -0,0 +1,5 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import { createContext } from 'react';
+
+export const AppLayoutVisibilityContext = createContext<boolean>(true);
```
just-boris, line 5: For compatibility with widgetization, you need to define context this way: https://github.com/cloudscape-design/components/blob/main/src/app-layout/visual-refresh-toolbar/contexts.ts#L11-L12

Also, you could keep both contexts in a single file for consistency

src/app-layout/visual-refresh-toolbar/index.tsx
```diff
@@ -463,7 +464,11 @@ const AppLayoutVisualRefreshToolbar = React.forwardRef<AppLayoutProps.Ref, AppLa
     return (
       <>
         {/* Rendering a hidden copy of breadcrumbs to trigger their deduplication */}
-        {!hasToolbar && breadcrumbs ? <ScreenreaderOnly>{breadcrumbs}</ScreenreaderOnly> : null}
+        {!hasToolbar && breadcrumbs ? (
+          <AppLayoutVisibilityContext.Provider value={isIntersecting}>
```
just-boris, line 468: How about wrapping the entire content into `AppLayoutVisibilityContext.Provider`, in case if we want to read somewhere else in the future without requiring a merge from live?

PR #3152, opened 2024-12-23
refactor: Reusable internal drag handle

pan-kot:
### Description

Reusing drag handle between collection prefs, split panel, and drawers with potential future extension to board components, see: [a3UOAhaNX7KH].

### How has this been tested?

* New unit tests for drag handle
* Screenshot tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/utils/use-focus-control.ts
+++ b/src/app-layout/utils/use-focus-control.ts
@@ -9,7 +9,8 @@ export interface Focusable {
 export interface FocusControlRefs {
   toggle: RefObject<Focusable>;
   close: RefObject<Focusable>;
-  slider: RefObject<HTMLDivElement>;
+  slider: RefObject<Focusable>;
+  handle: RefObject<HTMLDivElement>;
 }
 
 export interface FocusControlState {
@@ -35,7 +36,8 @@ export function useMultipleFocusControl(
       refs.current[drawerId] = {
         toggle: createRef<Focusable>(),
         close: createRef<Focusable>(),
-        slider: createRef<HTMLDivElement>(),
+        slider: createRef<Focusable>(),
+        handle: createRef<HTMLDivElement>(),
       };
     }
   });
@@ -101,7 +103,8 @@ export function useFocusControl(
   const refs = {
     toggle: useRef<Focusable>(null),
     close: useRef<Focusable>(null),
-    slider: useRef<HTMLDivElement>(null),
+    slider: useRef<Focusable>(null),
+    handle: useRef<HTMLDivElement>(null),
   };
   const previousFocusedElement = useRef<HTMLElement>();
   const shouldFocus = useRef(false);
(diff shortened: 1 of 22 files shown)


src/internal/events/index.ts
```diff
@@ -82,7 +82,13 @@ export function fireCancelableEvent<T>(
   return event.defaultPrevented;
 }
 
-export function fireKeyboardEvent(handler: CancelableEventHandler<BaseKeyDetail>, reactEvent: React.KeyboardEvent) {
+export function fireKeyboardEvent(
+  handler: CancelableEventHandler<BaseKeyDetail> | undefined,
+  reactEvent: React.KeyboardEvent
+) {
+  if (!handler) {
+    return;
```
Al-Dani, line 90: nice!

src/internal/components/drag-handle/index.tsx
```diff
@@ -1,44 +1,83 @@
...
+      ariaValue,
+      disabled,
+      onPointerDown,
+      onKeyDown,
+      className,
+    }: DragHandleProps,
+    ref: React.Ref<DragHandleProps.Ref>
+  ) => {
+    const dragHandleRefObject = useRef<HTMLElement>(null);
+    const dragHandleRef = useMergeRefs(dragHandleRefObject);
```
Al-Dani, line 33: what do we merging here?

src/app-layout/utils/use-focus-control.ts
```diff
@@ -9,7 +9,8 @@ export interface Focusable {
 export interface FocusControlRefs {
   toggle: RefObject<Focusable>;
   close: RefObject<Focusable>;
-  slider: RefObject<HTMLDivElement>;
+  slider: RefObject<Focusable>;
```
Al-Dani, line 12: why do we need to refs now? I see `slider` ref is used to focus on it inside Split panel. Can we instead use `handle` ref?

src/internal/events/index.ts
just-boris: What's nice here, if the same check already happened  on the line 71 anyway?

Al-Dani: true, but there are some occurrences of `handler && fireKeyboardEvent(onKeyDown..)` across the codebase. All other `fire.*Event` have this check built-in, so why not align this method too?

just-boris: Making it optional is nice. Duplicating already existing check is not

src/app-layout/utils/use-focus-control.ts
pan-kot: I prepare the drag handle component to become a public component. Similar to other similar components, it does not give access to the dom node via ref but only allows to cast the focus programmatically - that is what the slider ref is used for.

The handle ref is a custom div wrapper around the drag handle to get the placement information from.

src/internal/components/drag-handle/index.tsx
pan-kot: Good catch! There was the dragHandleRef originally but I had to replace it with the combination of the ref object and ref to use the useForwardFocus. While it is a common strategy in most components, the drag handle is still internal and does not feature the __internalRootRef - I will update it to only keep the ref object for now.

src/internal/events/index.ts
pan-kot: Good catch - I will remove the unneeded if statement 👍

PR #3165, opened 2025-01-02
fix: Make table column announcements less verbose for screen readers

avinashbot:
### Description

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


gethinwebster:
As far as I can see, this change means that the `ariaLabel` only gets applied to the button(s), and therefore is never actually applied to the column itself. It still gets announced when navigating between header cells, but when navigating between body cells you now only get the visible column label. I think that this is _probably_ a breaking change. At the very least, the documentation is now incorrect. But I could see cases where the `ariaLabel` might have been used, perhaps not as we would expect it and without the sorting state, and where changing this behavior could be a regression.

gethinwebster:
In chrome+VO when navigating between header cells I now get announcements of `[visible name] [visible name] [arialabel]`. Previously it was `[arialabel] [arialabel]` from what I can tell. Neither is great, but I'm not sure this is an improvement.

gethinwebster (approved):

PR #3166, opened 2025-01-03
Adjust prompt input height when minRows changes

phuctm97:
### Description

This PR fixes a bug that when `minRows` of `PromptInput` changes, it won't have effect until either `maxRows` or `value` changes. Changing `minRows` should also adjust height.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/prompt-input/internal.tsx
+++ b/src/prompt-input/internal.tsx
@@ -111,7 +111,7 @@ const InternalPromptInput = React.forwardRef(
         const minTextareaHeight = `calc(${LINE_HEIGHT} +  ${tokens.spaceScaledXxs} * 2)`; // the min height of Textarea with 1 row
         textareaRef.current.style.height = `min(max(${scrollHeight}, ${minTextareaHeight}), ${maxRowsHeight})`;
       }
-    }, [maxRows, LINE_HEIGHT, PADDING]);
+    }, [minRows, maxRows, LINE_HEIGHT, PADDING]);
 
     useEffect(() => {
       const handleResize = () => {


src/prompt-input/internal.tsx
```diff
@@ -111,7 +111,7 @@ const InternalPromptInput = React.forwardRef(
         const minTextareaHeight = `calc(${LINE_HEIGHT} +  ${tokens.spaceScaledXxs} * 2)`; // the min height of Textarea with 1 row
         textareaRef.current.style.height = `min(max(${scrollHeight}, ${minTextareaHeight}), ${maxRowsHeight})`;
       }
-    }, [maxRows, LINE_HEIGHT, PADDING]);
+    }, [minRows, maxRows, LINE_HEIGHT, PADDING]);
```
orangevolon, line 114: Since `adjustTextarea` callback doesn't use `minRows`, adding it as its dependency would be confusing. 

For the same reason you should get a linter error like this:

image

We can either mute this eslint error and include a comment that clearly explains how changing the `minRows` affects the `rows` attribute on the underlying `<textarea />` element which ultimately changes the `scrollHeight`

or simply move the `minRows` the the dependency array of `useEffect` at line `130`:

```tsx
useEffect(() => {
  adjustTextareaHeight();
}, [value, adjustTextareaHeight, maxRows, minRows, isCompactMode]
```

orangevolon (changes requested):

just-boris:
Closing for housecleaning. Feel free to ping if you would like to resume this

Issue #3140, opened 2024-12-13
[Feature Request]: Prevent DataPicker and DataRangePicker to resize/jump when clicking through months


apisarev:
### Description

When DatePicker and DateRangePicker popups open above the input, then clicking the left and right buttons resizes the popup content, causing arrow positions to jump up and down. This effect is not noticeable when the popup opens down.

![DataPicker](https://github.com/user-attachments/assets/955ec583-3111-48d3-b33c-991c0f7f7765)

![DataRangePicker](https://github.com/user-attachments/assets/3da96aca-4c92-4b28-a06a-cee37adb9c40)

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/cloudscape-design/components/blob/main/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/cloudscape-design/components/issues) for duplicate requests

apisarev:
To reproduce:
1. Open [DatePicker](https://cloudscape.design/components/date-picker/) or [DateRangePicker](https://cloudscape.design/components/date-range-picker?tabId=playground) playground
2. Add a bit more space above the component by changing some div height from above using the DevTools
3. Enable expandToViewport
4. Scroll the page so that the popup would open up, not down

apisarev:
A possible solution, when the popup opens up, always display 6 rows.

pan-kot:
Hello,

Thanks for reporting the issue and providing so much context for us to understand it and reproduce!

Our team will take a look at ways to mitigate the problem and also consider the solution you suggested. We will communicate the updates in the comments.

pan-kot:
Hello,

The feature was added to the Calendar, Date picker, and Date range picker components. Now, all month calendars render 6 weeks of dates when the first or last week(s) might belong to the previous or next month.

<img width="541" alt="Image" src="https://github.com/user-attachments/assets/efc12f98-4932-44c6-ae6f-d132c38e11e5" />

PR #3178, opened 2025-01-09
fix: Fixes calendar jumping when switching months

pan-kot:
### Description

The PR introduces placeholder dates to Calendar, Date picker, and Date range picker components such that each calendar always has exactly 6 rows of dates which ensures its size stays constant when user switches between months.

Rel: [llayAeObY9Ic], AWSUI-60266, https://github.com/cloudscape-design/components/issues/3140

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/test-utils/dom/date-range-picker/index.ts
```diff
@@ -122,7 +123,7 @@ export class DrpDropdownWrapper extends ComponentWrapper {
   ): CalendarDateWrapper {
     const gridClassName = grid === 'right' ? styles['second-grid'] : styles['first-grid'];
     return this.findComponent(
-      `.${gridClassName} .${gridStyles.week}:nth-child(${row}) .${gridStyles.day}:nth-child(${column})`,
+      `.${gridClassName} .${testStyles['calendar-week']}[data-awsui-weekindex="${row}"] .${testStyles['calendar-date']}:nth-child(${column})`,
```
pan-kot, line 126: Using data- attribute instead of :nth-child() to exclude rows with placeholder dates only. That is necessary for the backward-compatibility, to ensure the indices of the monthly dates do not change.

src/internal/utils/date-time/calendar.ts
```diff
@@ -0,0 +1,165 @@
...
+  isVisible: boolean;
+  isSelected: boolean;
+  isInRange: boolean;
+  isSelectionTop: boolean;
+  isSelectionBottom: boolean;
+  isSelectionLeft: boolean;
+  isSelectionRight: boolean;
+}
+
+export class MonthCalendar {
```
pan-kot, line 38: The calendar abstraction helps calculating indices and styles for calendar grid, which is non-trivial due to presence of placeholder dates that are treated differently.

Ideally, the same abstraction should be reused between Calendar, and Date range picker components. However, those are currently using very different implementations and only one supports the year calendar. We can reuse the code between the components as soon we update those to use the same calendar element implementation.

src/calendar/__tests__/calendar.test.tsx
```diff
@@ -42,13 +44,19 @@ function getDayText(wrapper: CalendarWrapper, row: number, col: number) {
...
-  test('check a11y', async () => {
-    const { container } = renderCalendar();
-    await expect(container).toValidateA11y();
-  });
+test('check a11y', async () => {
+  const { container } = renderCalendar();
+  await expect(container).toValidateA11y();
 });
 
+test.each(range(0, 11).map(month => addMonths(new Date('2025-01-01'), month).toISOString().split('T')[0]))(
```
connorlanigan, line 52: Not critical, but readability could be improved by moving the range to a named variable. For example:
```suggestion
const eachMonthOfTheYear = range(0, 11).map(month => addMonths(new Date('2025-01-01'), month).toISOString().split('T')[0]);
test.each(eachMonthOfTheYear)(
```

connorlanigan (approved):

PR #3181, opened 2025-01-10
fix: Allow text selection in date picker and date range picker

jperals:
### Description

Issue: `AWSUI-60182`

In the case of dropdowns that contain only a list of selectable options (autosuggest, button dropdown, multiselect, select) it makes sense to disable text selection, but not necessarily for anything that the dropdown may contain, and in particular not for date pickers and date range pickers, where this is having undesired side effects. Therefore moving `user-select: none` from the dropdown component down to the options list component.

### How has this been tested?

- Preparing visual regression tests
- Manually tested:
  - Text is selectable in date picker and date ranger picker
  - Double-clicking previous and next buttons in date picker or date range picker does not have unexpected side effects
  - Text is _not_ selectable in autosuggest, button dropdown, multiselect, select

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/date-picker/index.tsx
+++ b/src/date-picker/index.tsx
@@ -168,11 +168,6 @@ const DatePicker = React.forwardRef(
 
     baseProps.className = clsx(baseProps.className, styles.root, styles['date-picker-container']);
 
-    const handleMouseDown = (event: React.MouseEvent) => {
-      // prevent currently focused element from losing it
-      event.preventDefault();
-    };
-
     return (
       <div {...baseProps} ref={mergedRef} onKeyDown={!disabled && !readOnly ? onWrapperKeyDownHandler : undefined}>
         {disabled || readOnly ? (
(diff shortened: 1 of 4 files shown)


src/date-picker/index.tsx
```diff
@@ -168,11 +168,6 @@ const DatePicker = React.forwardRef(
 
     baseProps.className = clsx(baseProps.className, styles.root, styles['date-picker-container']);
 
-    const handleMouseDown = (event: React.MouseEvent) => {
```
jperals, line 171: This was preventing text selection in date picker even after removing `user-select: none` from the dropdown.

It was introduced 3.5 years ago in `CR-49510605` with the justification to prevent an undesired focus ring from appearing when clicking the calendar, but I can't reproduce this today.

src/date-range-picker/styles.scss
```diff
@@ -131,7 +131,6 @@ $calendar-header-color: awsui.$color-text-body-default;
 }
 
 .label {
-  user-select: text;
```
jperals, line 134: These two rules to override the parent's behavior are not necessary anymore

pan-kot (approved):

PR #3182, opened 2025-01-10
fix: Performance bug fix to emit marks for only first tab

guptbhum:
### Description

Bug - {component}Rendered and {component}Updated marks are emitted even if current component is adding to the second tab component. We don't want to publish metrics for the sections which are not visible on page load. 
Fix - Updated the usePerformanceMarks logic to emit markers only if the primary button or table is added to the first tab. 

### How has this been tested?

Locally tested the change
TODO - Updated unit and integration tests

<details>

#### Correctness

#### Security

#### Testing

- _Changes are covered with new/existing unit tests?_ existing tests
- _Changes are covered with new/existing integration tests?_ existing tests
</details>

---

### DIFF
--- a/src/internal/context/tab-context.ts
+++ b/src/internal/context/tab-context.ts
@@ -1,0 +1,12 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import { createContext, useContext } from 'react';
+export interface TabContextProps {
+  isInFirstTab: boolean | null;
+}
+export const TabContext = createContext<TabContextProps>({
+  isInFirstTab: null,
+});
+export const useTabContext = () => {
+  return useContext(TabContext);
+};
--- a/src/internal/hooks/use-performance-marks/index.ts
+++ b/src/internal/hooks/use-performance-marks/index.ts
@@ -4,6 +4,7 @@
 import { useEffect, useState } from 'react';
 
 import { useModalContext } from '../../context/modal-context';
+import { useTabContext } from '../../context/tab-context';
 import { useDOMAttribute } from '../use-dom-attribute';
 import { useEffectOnUpdate } from '../use-effect-on-update';
 import { useRandomId } from '../use-unique-id';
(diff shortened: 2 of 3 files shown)


connorlanigan:
The first enabled tab is not necessarily the tab that is initially open, because developers can control which tab they want to show initially using the `activeTabId` prop. I recommend to store the initial value of the `activeTabId` variable, and use that to determine whether the Tabs component is still showing the initial tab, until the user switches tabs for the first time.

For example, something roughly like:

```js

const [activeTabId, setActiveTabId] = useControllable(...);

const initialTabId = useState(activeTabId)[0];

const isOnInitialTab = useRef(true);

if(isOnInitialTab.current) {
   isOnInitialTab.current = activeTabId === initialTabId;
}

...


<TabContext.Provider
     value={{
       isInInitialTab: isOnInitialTab.current,
     }}
   >

```

src/internal/hooks/use-performance-marks/index.ts
```diff
@@ -64,6 +66,7 @@ export function usePerformanceMarks(
     }
 
     const renderedMarkName = `${name}Rendered`;
+    console.log('mark');
```
orangevolon, line 69: I think this is a left-over.

src/internal/context/tab-context.ts
```diff
@@ -0,0 +1,12 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import { createContext, useContext } from 'react';
+export interface TabContextProps {
+  isInFirstTab: boolean | null;
+}
+export const TabContext = createContext<TabContextProps>({
+  isInFirstTab: null,
```
orangevolon, line 8: I'd suggest to change this to `isInDefaultTab` or `hasTabChanged`. First tab means the first one from the left and the default tab might not be that one.

src/internal/hooks/use-performance-marks/index.ts
```diff
@@ -48,10 +49,11 @@ export function usePerformanceMarks(
 ) {
   const id = useRandomId();
   const { isInModal } = useModalContext();
+  const { isInFirstTab } = useTabContext();
   const attributes = useDOMAttribute(elementRef, 'data-analytics-performance-mark', id);
   const evaluateComponentVisibility = useEvaluateComponentVisibility();
   useEffect(() => {
-    if (!enabled() || !elementRef.current || isInModal) {
+    if (!enabled || !elementRef.current || isInModal || isInFirstTab === false) {
```
orangevolon, line 56: We have changed `!enabled()` to `enabled`. Is this on purpose? 

The type of `enabled` is still function.

https://github.com/cloudscape-design/components/blob/e1c488b01396999a4609573bece9dcf830311632/src/internal/hooks/use-performance-marks/index.ts#L44

orangevolon (changes requested):

just-boris:
Closing for housecleaning. Feel free to ping if you still need this

PR #3199, opened 2025-01-16
revert: Reusable internal drag handle (#3152)

pan-kot:
This reverts commit cf9b42c79e0fdbde2adfbc6744949c2892cdfa01.

### Description

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/visual-refresh-toolbar/drawer/global-drawer.tsx
+++ b/src/app-layout/visual-refresh-toolbar/drawer/global-drawer.tsx
@@ -58,7 +58,7 @@ function AppLayoutGlobalDrawerImplementation({
     minWidth: minDrawerSize,
     maxWidth: maxDrawerSize,
     panelRef: drawerRef,
-    handleRef: refs?.handle,
+    handleRef: refs?.slider,
     onResize: size => onActiveDrawerResize({ id: activeDrawerId!, size }),
   });
   const size = getLimitedValue(minDrawerSize, activeDrawerSize, maxDrawerSize);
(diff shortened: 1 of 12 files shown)


just-boris (approved):

src/app-layout/visual-refresh-toolbar/drawer/global-drawer.tsx
```diff
@@ -111,17 +111,15 @@ function AppLayoutGlobalDrawerImplementation({
...
-                    position="side"
-                    className={testutilStyles['drawers-slider']}
-                    ariaLabel={activeGlobalDrawer?.ariaLabels?.resizeHandle}
-                    ariaValuenow={resizeProps.relativeSize}
-                    onKeyDown={resizeProps.onKeyDown}
-                    onPointerDown={resizeProps.onPointerDown}
-                  />
-                </div>
+                <PanelResizeHandle
+                  ref={refs?.slider}
```
georgylobko, line 115: We need to be very careful with this change, because the global drawer component is shipped via widget API, which means that we might have 4 possible situations:

1. old code - old widget
2. old code - new widget
3. new code - old widget
4. new code - new widget

In case of 2 this code will break, because this component expects `handle`, but will get the `slider` instead and vice versa for case 3. So these situation need special treatment

georgylobko: oh I just realized that this PR reverts the change!

pan-kot: Apparently, that is the reasons of the revert! :)

PR #3202, opened 2025-01-16
fix: Prevent dropdown misplacement in iOS

jperals:
### Description

Issue: `AWSUI-59635`

### How has this been tested?

- Manually, on real iOS device and with Chrome's mobile simulator
- Ran the changes on my pipeline
- Added unit tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/components/dropdown/__tests__/dropdown-position.test.ts
+++ b/src/internal/components/dropdown/__tests__/dropdown-position.test.ts
@@ -21,7 +21,7 @@ describe('applyDropdownPositionRelativeToViewport', () => {
     dropInlineStart: false,
   };
 
-  test("anchored to the trigger's block start", () => {
+  test("sets block end when the dropdown is anchored to the trigger's block start (expands up)", () => {
     const dropdownElement = document.createElement('div');
     applyDropdownPositionRelativeToViewport({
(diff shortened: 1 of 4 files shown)


src/internal/components/dropdown/__tests__/dropdown-position.test.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  test("sets block end when the dropdown is anchored to the trigger's block start (expands up)", () => {
+    const dropdownElement = document.createElement('div');
+    applyDropdownPositionRelativeToViewport({
+      dropdownElement,
+      triggerRect,
+      position: { ...baseDropdownPosition, dropBlockStart: true },
+      isMobile: false,
+    });
+    expect(dropdownElement.style.insetBlockEnd).toBeTruthy();
```
jperals, line 32: In a couple cases (here and 3 tests further down) I'm testing just that a value is set (truthy), instead of the actual value. This is because the value is a CSS expression using `calc()`, and I don't think it makes sense to be so specific about the implementation. We already have visual regression tests for the expected user-facing end result, which is what matters.

michaeldowseza (approved):

PR #3203, opened 2025-01-16
fix: Fixes collection preferences active item style

pan-kot:
### Description

Fixing a small visual regression introduced in: https://github.com/cloudscape-design/components/pull/3195

The `<li>` styles include borders that overlay the focus outline added when the item is active.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/collection-preferences/content-display/index.tsx
+++ b/src/collection-preferences/content-display/index.tsx
@@ -134,11 +134,14 @@ export default function ContentDisplayPreference({
           itemDefinition={{ id: item => item.id, label: item => item.label }}
           onItemsChange={({ detail }) => onChange(detail.items)}
           disableReorder={columnFilteringText.trim().length > 0}
-          renderItem={({ ref, item, style, className, dragHandleProps }) => {
+          renderItem={({ ref, item, style, className, dragHandleProps, isDragGhost }) => {
             className = clsx(className, getOptionClassName());
             const content = (
               <ContentDisplayOption ref={ref} option={item} onToggle={onToggle} dragHandleProps={dragHandleProps} />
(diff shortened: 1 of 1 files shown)


src/collection-preferences/content-display/index.tsx
```diff
@@ -134,11 +134,14 @@ export default function ContentDisplayPreference({
...
           disableReorder={columnFilteringText.trim().length > 0}
-          renderItem={({ ref, item, style, className, dragHandleProps }) => {
+          renderItem={({ ref, item, style, className, dragHandleProps, isDragGhost }) => {
             className = clsx(className, getOptionClassName());
             const content = (
               <ContentDisplayOption ref={ref} option={item} onToggle={onToggle} dragHandleProps={dragHandleProps} />
             );
+            if (isDragGhost) {
+              return content;
+            }
```
just-boris, line 144: Approved because it is a hotfix, but this begs the question – should there be any `aria-hidden` or other attributes set for this content?

This exacerbates when distributing `SortableArea` as a public component. Every team is supposed to not forget about this use-case, making it prone to issues

just-boris (approved):

src/collection-preferences/content-display/index.tsx
pan-kot: This code is a small revert to our current approved implementation of the collection preferences content display, so I expect it to be alright. The issue will styles should be solvable in a way that depending on the isDragGhost is not needed.

just-boris: Ok

PR #3205, opened 2025-01-17
fix: Allows table cells be expandable and editable at the same time

pan-kot:
### Description

The PR makes it possible to have cell editing in the first column of tables with expandable rows.

Rel: [tNAkA40KWyvj], AWSUI-43194.

### How has this been tested?

* New unit and integration test
* Screenshot tests
* Dry-run to live

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/table/editable.page.tsx
+++ b/pages/table/editable.page.tsx
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 import React, { ForwardedRef, forwardRef, useContext, useEffect, useRef, useState } from 'react';
 
-import { Box, Button, Checkbox, Link, Modal, SpaceBetween } from '~components';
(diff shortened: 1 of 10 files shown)


gethinwebster (dismissed):

pan-kot:
@gethinwebster, I added a small refactoring on top, removing the onMouseEnter/onMouseExit handlers from the cell. Apologies for not creating a separate commit for that!

The reason the handlers are removed is the following:
1. After the original change, the handlers were used to set the hover state used as a condition to render the success icon;
2. The success icon renders if either hover or focus is present;
3. The focus is always present after the field is submitted, so the hover dependency is not actually needed.

gethinwebster (approved):

PR #3206, opened 2025-01-20
chore: Replace removeEventListener with abort controller

georgylobko:
### Description

Replaced `removeEventListener` with `AbortController.abort` where it brought code clarify, specifically when multiple event listeners should be removed at once.

Dry-run: 7201850483

The `AbortController` is supported in `jest` from version 27. For those who use older versions (if any), should either update their `jest` or use a workaround, for instance, this one: https://github.com/facebook/create-react-app/issues/11626#issuecomment-1785565456

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/animate.ts
+++ b/src/internal/animate.ts
@@ -93,7 +93,7 @@ export function animate({
           if (event.target === element) {
             // Clean up remaining inline styles
             element.style.transitionProperty = '';
-            element.removeEventListener('transitionstart', onTransitionEnd);
+            element.removeEventListener('transitionend', onTransitionEnd);
             if (onTransitionsEnd) {
               ongoingAnimations.delete(id);
               if (ongoingAnimations.size === 0) {
--- a/src/table/__tests__/use-focus-navigation.test.tsx
+++ b/src/table/__tests__/use-focus-navigation.test.tsx
@@ -9,7 +9,7 @@ import { renderHook } from '../../__tests__/render-hook';
 import useTableFocusNavigation from '../use-table-focus-navigation';
 
 const focusFn = jest.fn();
-const abortFn = jest.fn();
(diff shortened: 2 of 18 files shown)


src/internal/animate.ts
```diff
@@ -93,7 +93,7 @@ export function animate({
           if (event.target === element) {
             // Clean up remaining inline styles
             element.style.transitionProperty = '';
-            element.removeEventListener('transitionstart', onTransitionEnd);
+            element.removeEventListener('transitionend', onTransitionEnd);
```
georgylobko, line 96: This change does not have to do anything with `AbortController`, I just noticed this typo

src/table/__tests__/use-focus-navigation.test.tsx
```diff
@@ -48,7 +48,7 @@ const TestComponent = () => {
 
   const handleKeyup = (evt: React.KeyboardEvent) => {
     if (evt.key === 'BIGUP') {
-      jest.spyOn(tableRef.current!, 'removeEventListener').mockImplementation(rootRemoveEventListener as any);
+      abortFn();
```
cansuaa, line 51: I don't see AbortController being used in `table/use-table-focus-navigation.ts` so the test shouldn't be changed. And it's very normal that the related test is passing because basically here you are directly calling the function and asserting that it is called below in the test. So even abort was used in the file, I don't think these two lines are really testing the same thing.

georgylobko: Good catch. I initially introduced `AbortController` in the `table/use-table-focus-navigation.ts`, but then decided to get rid of it, because it didn't bring any benefit there. Addressed

cansuaa (approved):

PR #3219, opened 2025-01-26
chore: Attaches tooltip classes to popover container

pan-kot:
### Description

The tooltip test classes were attached to the div that contains a popover container. However, when the container becomes hidden due to scroll logic, the outer div stays and gives a false positive when checked with isDisplayed if using webdriverio v9. In the PR the classes are moved to the popover container so that the tests produce expected results.

### How has this been tested?

See a failed dry-run: https://github.com/cloudscape-design/browser-test-tools/actions/runs/12945599198/job/36179765443?pr=109

The tests succeed with the fix applied: https://github.com/cloudscape-design/components/pull/3218 

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/components/tooltip/index.tsx
+++ b/src/internal/components/tooltip/index.tsx
@@ -1,7 +1,6 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 import React from 'react';
-import clsx from 'clsx';
 
 import PopoverArrow from '../../../popover/arrow';
 import PopoverBody from '../../../popover/body';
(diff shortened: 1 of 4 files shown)


src/internal/components/tooltip/index.tsx
```diff
@@ -39,7 +38,7 @@ export default function Tooltip({
 
   return (
     <Portal>
-      <div className={clsx(styles.root, className)} {...contentAttributes} data-testid={trackKey}>
+      <div className={styles.root} {...contentAttributes} data-testid={trackKey}>
```
just-boris, line 41: Why is `styles.root` staying here? it is also a test-util only style, what is the use-case of selecting it, if it is incorrectly placed in the DOM?

Why does this DOM node needed in the first place?

--

There is an avatar test util pointing on this DOM node too: https://github.com/cloudscape-design/chat-components/blob/81f675f568a82891661839f382262b80ea276289/src/test-utils/dom/avatar/index.ts#L14

Do you plan updating it in a separate PR?

pan-kot: That's a good call out. I will create follow-up PRs that will 1) Remove the dependency on the tooltip wrapper from Avatar and 2) Remove the root style from this node.

I will also check if there is an opportunity to remove the node altogether. I think the "contentAttributes" and the testid can be passed down to the internal container.

PR #3233, opened 2025-01-30
feat: Allow tab content to be retained in memory

gethinwebster:
### Description

Allow inactive tab content to be rendered into the DOM, for use-cases such as:
- Eager rendering (and/or avoiding re-rendering) of heavy tab content
- Retaining content state in memory

Related links, issue #, if available: AWSUI-25984

### How has this been tested?

New unit and integ tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/tabs/in-memory-integ.page.tsx
+++ b/pages/tabs/in-memory-integ.page.tsx
@@ -1,0 +1,74 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import React, { useEffect, useState } from 'react';
+
+import { FormField, Input } from '~components';
+import SpaceBetween from '~components/space-between';
+import Tabs, { TabsProps } from '~components/tabs';
+
+const TabWithState = () => {
+  const [value, setValue] = useState('');
+  return (
+    <FormField label="Input">
+      <Input value={value} onChange={e => setValue(e.detail.value)} />
+    </FormField>
+  );
+};
+const TabWithLoading = () => {
+  const [loaded, setLoaded] = useState(false);
+  useEffect(() => {
+    setTimeout(() => setLoaded(true), 1000);
+  }, []);
+  return <div>{loaded ? 'Loaded' : 'Loading...'}</div>;
+};
+
+export default function TabsDemoPage() {
+  const tabs: Array<TabsProps.Tab> = [
+    {
+      label: 'Tab with state',
+      id: 'state',
+      content: <TabWithState />,
+      contentRenderStrategy: 'lazy',
+    },
+    {
+      label: 'Tab with state (not retained)',
+      id: 'state2',
+      content: <TabWithState />,
+      contentRenderStrategy: 'active',
+    },
+    {
+      label: 'Lazy loading',
+      id: 'lazy',
+      content: <TabWithLoading />,
+      contentRenderStrategy: 'lazy',
+    },
+    {
+      label: 'Eager loading',
+      id: 'eager',
+      content: <TabWithLoading />,
+      contentRenderStrategy: 'eager',
+    },
+  ];
+  const [selectedTab, setSelectedTab] = useState(tabs[0].id);
+  const [loaded, setLoaded] = useState(new Set([tabs[0].id]));
+  return (
+    <>
+      <h1>Tabs</h1>
+
+      <SpaceBetween size="xs">
+        <div>
+          <h2>Controlled component</h2>
+          <Tabs
+            tabs={tabs}
+            activeTabId={selectedTab}
+            onChange={event => {
+              setSelectedTab(event.detail.activeTabId);
+              setLoaded(new Set([...loaded.values(), event.detail.activeTabId]));
+            }}
+            i18nStrings={{ scrollLeftAriaLabel: 'Scroll left', scrollRightAriaLabel: 'Scroll right' }}
+          />
+        </div>
+      </SpaceBetween>
+    </>
+  );
+}
--- a/src/tabs/__integ__/in-memory.test.ts
+++ b/src/tabs/__integ__/in-memory.test.ts
@@ -1,0 +1,59 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import { BasePageObject } from '@cloudscape-design/browser-test-tools/page-objects';
+import useBrowser from '@cloudscape-design/browser-test-tools/use-browser';
+
+import createWrapper from '../../../lib/components/test-utils/selectors';
+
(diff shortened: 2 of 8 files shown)


src/tabs/interfaces.ts
```diff
@@ -78,6 +78,11 @@ export interface TabsProps extends BaseComponentProps {
    * If the tab content is too short, it will stretch. If the tab content is too long, a vertical scrollbar will be shown.
    */
   fitHeight?: boolean;
+  /**
+   * Determines whether the content for a specific tab should be rendered. This can be used to
+   * retain non-active tabs in memory. By default, only the active tab content is rendered.
+   */
+  renderTabContent?: (tab: TabsProps.Tab) => boolean | undefined;
```
pan-kot, line 85: Should we call this something like `shouldRenderTabContent`? Otherwise it can be confused with a renderer function that determines the actual content.

gethinwebster: good point, I think it _was_ that in my initial revision, not sure why I changed it: will update!

src/tabs/interfaces.ts
```diff
@@ -134,6 +138,13 @@ export namespace TabsProps {
      * using the `activeTabHref` property of the `change` event's detail.
      */
     href?: string;
+    /**
+     * Determines whether tab content is rendered (but hidden) when the tab is inactive:
```
pan-kot, line 142: I would remove the "(but hidden)" or rephrase, because technically the content is not hidden if the tab is active. Can maybe say "(but hidden when the tab is not active)".

src/tabs/index.tsx
```diff
@@ -60,6 +72,18 @@ export default function Tabs({
     changeHandler: 'onChange',
   });
 
+  const [viewedTabs, setViewedTabs] = useState(new Set(activeTabId));
+
+  useEffect(() => {
```
pan-kot, line 77: This effect can be removed in favour of decorating the setActiveTabId method, e.g.

```
const setActiveTabIdDecorated = (newActiveTabId) => {
  setActiveTabId(newActiveTabId);
  setViewedTabs(prev => new Set(prev).add(newActiveTabId));
}
```

This way the related state is set within one render cycle so should be better for performance and is better clear, too.

src/tabs/__tests__/tabs.test.tsx
```diff
@@ -3,6 +3,7 @@
 import React from 'react';
 import { fireEvent, render } from '@testing-library/react';
 
+import { warnOnce } from '@cloudscape-design/component-toolkit/internal';
```
pan-kot, line 6: There are no tests that involve warnOnce in this file.

src/tabs/__integ__/in-memory.test.ts
pan-kot, line 1: nit: should the file be called "tab-rendering-strategy" to better align with the feature name?

pan-kot: The same is relevant for the test page name.

pages/tabs/in-memory-integ.page.tsx
```diff
@@ -0,0 +1,74 @@
...
+    },
+    {
+      label: 'Eager loading',
+      id: 'eager',
+      content: <TabWithLoading />,
+      contentRenderStrategy: 'eager',
+    },
+  ];
+  const [selectedTab, setSelectedTab] = useState(tabs[0].id);
+  const [loaded, setLoaded] = useState(new Set([tabs[0].id]));
```
pan-kot, line 53: what this state is used for?

pages/tabs/in-memory-integ.page.tsx
```diff
@@ -0,0 +1,74 @@
...
+  return (
+    <>
+      <h1>Tabs</h1>
+
+      <SpaceBetween size="xs">
+        <div>
+          <h2>Controlled component</h2>
```
pan-kot, line 60: Does component have to be controlled? Also, this h2 title can be removed and the h1 updated to reflect the test page name.

src/tabs/__integ__/in-memory.test.ts
```diff
@@ -0,0 +1,59 @@
...
+    await browser.url('#/light/tabs/in-memory-integ');
+    await page.waitForVisible(wrapper.findTabs().findTabContent().toSelector());
+    await testFn(page);
+  });
+};
+
+test(
+  'eager-loaded tab content should already be available',
+  setupTest(async page => {
+    await page.pause(2000);
```
pan-kot, line 22: The dependency on page timeout can be easily removed here, I think. For instance, when the timeout inside the component is resolved you can set some flag to the global object or add a DOM node to the bottom of the page, e.g. `<div id="loaded">Tab loaded</div>`. Then you can query it in the test using waitForAssertion() helper.

src/tabs/__integ__/in-memory.test.ts
```diff
@@ -0,0 +1,59 @@
...
+    await page.click(wrapper.findTabs().findTabLinkByIndex(3).toSelector());
+    const activeTabContent = wrapper.findTabs().findTabContent().toSelector();
+    await expect(page.getText(activeTabContent)).resolves.toBe('Loading...');
+    await page.waitForAssertion(() => expect(page.getText(activeTabContent)).resolves.toBe('Loaded'));
+  })
+);
+test(
+  'tab state is retained when switching away and back',
```
pan-kot, line 38: nit: I would rename the test slightly to be "eager-loaded tab state is retained...", similar to the first two. Likewise, the last one can be "active-loaded tab state is not retained...".

src/tabs/index.tsx
gethinwebster: Given that this is a controllable component, I think we need to use an effect. Although it's unlikely, the active tab could change without `setActiveTabId` being called

src/tabs/interfaces.ts
gethinwebster: updated

pages/tabs/in-memory-integ.page.tsx
gethinwebster: no, that's a legacy of the old approach, removed

src/tabs/index.tsx
pan-kot: Oh, I see. Can the visited tabs be a ref then, e.g.:

```
const visitedTabsRef = useRef(new Set<string>());
visitedTabsRef.current.add(activeTabId);
```

Since changing the tab always causes a new render no matter if done in controlled or uncontrolled way, I think this approach should work, too.

gethinwebster: good point: there's no explicit re-render required here, and the updated value isn't needed until later (switching away from the tab) anyway

pan-kot (dismissed):

pan-kot (approved):

pan-kot (approved):

PR #3239, opened 2025-01-31
fix: Rigorizing the isStuck logic in container use-sticky-header

dpitcock:
### Description

The previous isStuck logic was causing some unintended visual issues. When the isStuck condition was falsey set to true, it was removing the border radius on an internal element. This caused that element to overlay on top of its parent container, hiding the curved borders on the corners. This problem became more noticeable when the screen was viewed at different zoom levels.

The isStuck was being falsely set to true because why the top pixel values from sequentially calling `.getBoundingClientRect()` on a parent and nested child element might not return the same values, even if they are in the same position. Here are some reasons as to why:

- Timing issues: The `.getBoundingClientRect()` method returns the size and position of the element relative to the viewport, and this can change slightly between the two calls if the element is being scrolled or resized during that time. This is especially true if the function is being called on scroll, as the element's position may have shifted slightly between the two calls.
- Pixel differences and rounding: The values returned by `.getBoundingClientRect() `are in pixels, but there can be small differences in the decimal places due to browser rounding and the way floating-point numbers are represented in JavaScript. This can result in slight discrepancies, even if the elements are in the same position.
- Subpixel rendering: Modern browsers use subpixel rendering to improve the appearance of text and graphics on high-resolution displays. This can result in slightly different bounding rectangle values, even for elements that appear to be in the same position.

To mitigate the issue, I wrap the returned values in a `Math.round()` function as I noticed the differences were often in the .00001's of pixels. This seems to resolve the issue.

broken-corner

Related links, issue #, if available: AWSUI-60267

### How has this been tested?

- unit test cases have been added for more conditionals
- integ passes test
- visual regression tests pass in dev pipeline

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/container/__tests__/sticky-header.test.tsx
+++ b/src/container/__tests__/sticky-header.test.tsx
@@ -25,7 +25,64 @@ beforeEach(() => {
 describe.each(['embedded', 'full-page', 'cards', 'default', 'stacked'] as Array<
   ContainerProps['variant'] | 'embedded' | 'full-page' | 'cards'
 >)('useStickyHeader with variant %s', variant => {
-  test('should set isStuck to true when rootTop is less than headerTop', () => {
+  test('should set isStuck to false when rootTop headerTop are equal and headerTop is not 0', () => {
+    const rootRef = {
+      current: document.createElement('div'),
+    };
+    rootRef.current.getBoundingClientRect = jest.fn().mockReturnValue({ top: 100 });
+
+    const headerRef = {
+      current: document.createElement('div'),
+    };
+    headerRef.current.getBoundingClientRect = jest.fn().mockReturnValue({ top: 100 });
+
+    const { result } = renderHook(() => useStickyHeader(rootRef, headerRef, variant, true, 0, 0, false));
+    act(() => {
+      window.dispatchEvent(new Event('scroll'));
+    });
+
+    expect(result.current.isStuck).toBe(false);
+  });
+
+  test('should set isStuck to true when rootTop less than a headerTop at 0', () => {
+    const rootRef = {
+      current: document.createElement('div'),
+    };
+    rootRef.current.getBoundingClientRect = jest.fn().mockReturnValue({ top: -100 });
+
+    const headerRef = {
+      current: document.createElement('div'),
+    };
+    headerRef.current.getBoundingClientRect = jest.fn().mockReturnValue({ top: 0 });
+
+    const { result } = renderHook(() => useStickyHeader(rootRef, headerRef, variant, true, 0, 0, false));
+    act(() => {
+      window.dispatchEvent(new Event('scroll'));
+    });
+
+    expect(result.current.isStuck).toBe(true);
+  });
+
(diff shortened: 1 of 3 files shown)


src/container/__tests__/sticky-header.test.tsx
```diff
@@ -21,97 +22,215 @@ beforeEach(() => {
   jest.resetAllMocks();
 });
 
-test('should set isStuck to true when rootTop is less than headerTop', () => {
-  const rootRef = {
-    current: document.createElement('div'),
-  };
-  rootRef.current.getBoundingClientRect = jest.fn().mockReturnValue({ top: 100 });
+describe.each(['embedded', 'full-page', 'cards', 'default', 'stacked', 'borderless'] as Array<
```
dpitcock, line 25: all possible options of variant are being tested here, yet the code-coverage does not pick this up

src/container/use-sticky-header.ts
```diff
@@ -95,18 +102,20 @@ export const useStickyHeader = (
...
       if (rootRef.current && headerRef.current) {
         const rootTopBorderWidth = parseFloat(getComputedStyle(rootRef.current).borderTopWidth) || 0;
-        const rootTop = rootRef.current.getBoundingClientRect().top + rootTopBorderWidth;
-        const headerTop = headerRef.current.getBoundingClientRect().top;
+        //using Math.round to adjust for rounding errors in floating-point arithmetic and timing issues
+        const rootTop = Math.round((rootRef.current.getBoundingClientRect().top + rootTopBorderWidth) * 10000) / 10000;
+        const headerTop = Math.round(headerRef.current.getBoundingClientRect().top * 10000) / 10000;
 
-        if (rootTop < headerTop) {
+        if (rootTop < headerTop && (variant === 'full-page' || headerTop === 0)) {
```
dpitcock, line 109: this does not apply isStuck correctly when it is a table with StickyHeader inside content layout using AppLayout

src/container/use-sticky-header.ts
```diff
@@ -95,9 +97,9 @@ export const useStickyHeader = (
       }
       if (rootRef.current && headerRef.current) {
         const rootTopBorderWidth = parseFloat(getComputedStyle(rootRef.current).borderTopWidth) || 0;
-        const rootTop = rootRef.current.getBoundingClientRect().top + rootTopBorderWidth;
-        const headerTop = headerRef.current.getBoundingClientRect().top;
-
+        //using Math.round to adjust for rounding errors in floating-point arithmetic and timing issues
+        const rootTop = Math.round((rootRef.current.getBoundingClientRect().top + rootTopBorderWidth) * 10000) / 10000;
```
avinashbot, line 101: I don't understand how this works - why not just call `Math.round()` directly if the differences are within a single decimal point?

dpitcock: I wanted to account for false positives when screens are zoomed in, but I see your point, the zoom will never be that much where a partial pixel would make a difference

avinashbot (approved):

PR #3244, opened 2025-02-03
feat: Add ariaLabel property to icon component

avinashbot:
### Description

```diff
+   /**
+    * Specifies alternate text for the icon. We recommend that you provide this for accessibility.
+    */
+   ariaLabel?: string;

    /**
-    * Specifies alternate text for a custom icon (using the `url` attribute). We recommend that you provide this for accessibility.
+    * Specifies alternate text for a custom icon (using the `url` attribute).
     * This property is ignored if you use a predefined icon or if you set your custom icon using the `svg` slot.
     *
+    * @deprecated Use `ariaLabel` instead.
     */
    alt?: string;

```

### Motivation

If you want to have an accessible label for your icon (assuming you don't have visible text next to it), you currently have to do this:

```tsx
<span role="img" aria-label="Success">
  <Icon name="status-positive" />
</span>
```

Boo! Annoying! The current approach feels manual and seriously lacks discoverability - people can't find an `ariaLabel` property, and sometimes end up (understandably) misusing the `alt` property. We could really just let people do this instead:

 ```tsx
<Icon name="status-positive" ariaLabel="Success" />
```

This alone doesn't fix the fact that `alt` is still confusing, so I also deprecated `alt` (`ariaLabel` will be used if provided, but we'll still fall back to "alt" for `` only).

### How has this been tested?

Unit tests. Ignore all the force pushes, I was too lazy to run unit tests on my laptop.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/icon/__tests__/icon.test.tsx
+++ b/src/icon/__tests__/icon.test.tsx
@@ -85,6 +85,28 @@ describe('Icon Component', () => {
       expect(img).toHaveAttribute('alt', 'custom icon');
     });
 
+    test('should render a custom icon with alternate text when a url and ariaLabel are provided', () => {
+      const { container } = render(<Icon url={url} ariaLabel="custom icon" />);
+      const wrapper = createWrapper(container);
+      expect(wrapper.findIcon()!.getElement()).not.toHaveAttribute('aria-label');
+      expect(container.querySelector('img')).toHaveAttribute('alt', 'custom icon');
+    });
+
+    test('should prefer ariaLabel when alt is also provided', () => {
+      const { container } = render(<Icon url={url} alt="icon alt" ariaLabel="custom icon" />);
+      const wrapper = createWrapper(container);
+      expect(wrapper.findIcon()!.getElement()).not.toHaveAttribute('aria-label');
+      expect(container.querySelector('img')).toHaveAttribute('alt', 'custom icon');
+    });
+
+    test('should not set aria-hidden="true" for custom svg icons if an ariaLabel is provided', () => {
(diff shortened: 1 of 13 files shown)

### CODE REVIEW

File: src/icon/__tests__/icon.test.tsx [hunk 1]
```diff
@@ -85,6 +85,28 @@ describe('Icon Component', () => {
...
     });
 
+    test('should render a custom icon with alternate text when a url and ariaLabel are provided', () => {
+      const { container } = render(<Icon url={url} ariaLabel="custom icon" />);
+      const wrapper = createWrapper(container);
+      expect(wrapper.findIcon()!.getElement()).not.toHaveAttribute('aria-label');
+      expect(container.querySelector('img')).toHaveAttribute('alt', 'custom icon');
+    });
+
+    test('should prefer ariaLabel when alt is also provided', () => {
```
georgylobko, line 95: the description says that `ariaLabel` should take precedence over `alt`, but the assertions below do the opposite

avinashbot: Ah, yes you're right. I wrote the description before I realized using the new property is better than letting the old property (`alt`) override it. The tests are correct; updated the description.

georgylobko: I don't see any changes here. Did you commit them?

avinashbot: The code and tests are correct. It's the PR description that's wrong.

georgylobko (approved):

PR #3253, opened 2025-02-05
fix: Increase input clear button target size to 24×24 pixels

avinashbot:
### Description

Not sold on this yet, but it does work. We don't have to update all "inline-icon" buttons, just the ones that are within 24px of another interactive control. In this case, the clear button is on top of the input element, so we're literally 0px away from the nearest interactive control. 

Related links, issue #, if available: AWSUI-60212

### How has this been tested?

Should be visually identical. Just pushed to my dev branch, and everything looks good.

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/button/constants.scss
```diff
@@ -78,6 +78,20 @@ $variants: (
...
+    'hover-background': transparent,
+    'hover-border-color': transparent,
+    'hover-color': awsui.$color-text-button-inline-icon-hover,
+    'active-background': transparent,
+    'active-border-color': transparent,
+    'active-color': awsui.$color-text-button-inline-icon-default,
+    'disabled-background': transparent,
+    'disabled-border-color': transparent,
+    'disabled-color': awsui.$color-text-button-inline-icon-disabled,
+  ),
```
orangevolon, line 94: I think introducing a new variant to the button would make it a bit confusing for the customers especially with a name like `pointer-target` which makes sense to us, but not necessarily to the user.

Why are we even using inline variant of the button inside the input? It's not inline with text at all. Alternatively we could introduce something like `block-icon` which is 24x24 and is a block. Since the `icon` variant also `24x20`.

avinashbot: This variant is internal button only. So it isn't exposed to users, just to the input component.

orangevolon (approved):

src/button/constants.scss
avinashbot: (Just to answer your second question, there is a more standard inline-block icon variant, but the visuals are completely different and the paddings+borders are much heavier than just 24x24.)

PR #3258, opened 2025-02-06
chore: Add proper textual name, role and state to breadcrumb last item

taheramr:
### Description

Adding proper textual name, role and state for the last item in the breadcrumb group. 

**Why?**
At narrow viewports, the truncated final breadcrumb item is focusable, but it has no role and disabled state. (Note: The breadcrumb items are not long enough to trigger truncation in the demo. However, truncation occurs frequently in practice.)

AWSUI-60244

### How has this been tested?
Using inspector and inspecting the A11y tree before and after

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/breadcrumb-group/item/item.tsx
```diff
@@ -70,7 +70,7 @@ type ItemProps = React.HTMLAttributes<HTMLElement> & {
 const Item = React.forwardRef<HTMLElement, ItemProps>(
   ({ anchorAttributes, children, isLast, ...itemAttributes }, ref) =>
     isLast ? (
-      <span ref={ref} className={styles.anchor} {...itemAttributes}>
+      <span role="link" aria-disabled={true} ref={ref} className={styles.anchor} {...itemAttributes}>
```
avinashbot, line 73: Hmm, I wonder if we should only do this if the last item is focusable in the first place rather than applying it to every last item? Because the problem is only there if we make the element focusable, which is only the case if it's truncated.

I don't have a good answer; something to check with our accessibility specialists later today?

avinashbot: Had a chat, here's the outcome:
- It would nice for *all* terminal breadcrumb group items (not just truncated ones) to be `role="link"`, `aria-disabled="true"`, and also tabbable. This makes for a consistent navigation experience through the entire breadcrumb list.
- And if we're going to do that, we should also add `aria-current="page"` to the final breadcrumb item.

src/breadcrumb-group/__integ__/breadcrumb-group.test.ts
```diff
@@ -163,29 +163,31 @@ describe('BreadcrumbGroup', () => {
...
-        await expect(page.isTooltipDisplayed()).resolves.toBe(false);
-        await expect(page.getActiveElementId()).resolves.toBe('focus-target-short-text');
-      },
-      { width: 950, height: 800 }
-    )
+  test.each([
+    [{ width: 770, height: 800 }, 'when truncated'],
+    [{ width: 1100, height: 800 }, 'default state'],
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  ])('Last item should be focusable %s', (viewportSize, scenario) =>
```
avinashbot, line 177: Couldn't you just remove the second variable?

src/breadcrumb-group/__tests__/breadcrumb-item.test.tsx
```diff
@@ -114,10 +114,16 @@ describe('BreadcrumbGroup Item', () => {
 
     test('should not be a link', () => {
       expect(lastLink.getElement()).not.toHaveAttribute('href');
-      expect(lastLink.getElement()).not.toHaveAttribute('aria-current');
       expect(lastLink.getElement().tagName).toEqual('SPAN');
     });
 
+    test('should have proper aria-labels and focusable', () => {
+      expect(lastLink.getElement()).toHaveAttribute('aria-current');
```
avinashbot, line 121: `aria-current="page"` specifically.

```suggestion
      expect(lastLink.getElement()).toHaveAttribute('aria-current', 'page');
```

src/breadcrumb-group/__tests__/breadcrumb-item.test.tsx
```diff
@@ -114,10 +114,16 @@ describe('BreadcrumbGroup Item', () => {
 
     test('should not be a link', () => {
       expect(lastLink.getElement()).not.toHaveAttribute('href');
-      expect(lastLink.getElement()).not.toHaveAttribute('aria-current');
       expect(lastLink.getElement().tagName).toEqual('SPAN');
     });
```
avinashbot, line 119: Looks like maybe this whole test shouldn't exist anymore?

```suggestion
```

avinashbot (changes requested):

avinashbot (approved):

PR #3260, opened 2025-02-06
chore: Adds pseudo element around the annotation trigger

orangevolon:
### Description

Re-opens #3250 with an addition of right margin for the inline hotspot element.

### Why?

After merging #3246, it turned out that this is causing the `text-wrap: ellipsis` to kick in, in the Firefox browser. This was happening because of the difference in width calculation between the browsers. The issue was visible in **examples/hands-on-tutorial** demo.

This PR adds an equal right margin so the truncation doesn't happen.

### How has this been tested?

- Dev pipeline: Added Firefox examples suite to dev pipeline.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/hotspot/styles.scss
+++ b/src/hotspot/styles.scss
@@ -35,5 +35,5 @@
 }
 
 .inlineWrapper {
-  margin-inline-start: awsui.$space-xxs;
+  margin-inline: awsui.$space-xxs;
 }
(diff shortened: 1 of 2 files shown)


src/hotspot/styles.scss
```diff
@@ -35,5 +35,5 @@
 }
 
 .inlineWrapper {
-  margin-inline-start: awsui.$space-xxs;
+  margin-inline: awsui.$space-xxs;
```
orangevolon, line 38: This is the only diff with #3246

cansuaa (approved):

PR #3208, opened 2025-01-21
feat: Public layout component

georgylobko:
### Description

doc: tb1MA6m8ZZEt

Related links, issue #, if available: n/a

### How has this been tested?

<details>
   <summary>Review checklist</summary>

_The following items are to be evaluated by the author(s) and the reviewer(s)._

#### Correctness

- _Changes include appropriate documentation updates._
- _Changes are backward-compatible if not indicated, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#public-apis)._
- _Changes do not include unsupported browser features, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#browsers-support)._
- _Changes were manually tested for accessibility, see [accessibility guidelines](https://cloudscape.design/foundation/core-principles/accessibility/)._

#### Security

- _If the code handles URLs: all URLs are validated through [the `checkSafeUrl` function](https://github.com/cloudscape-design/components/blob/main/src/internal/utils/check-safe-url.ts)._

#### Testing

- _Changes are covered with new/existing unit tests?_
- _Changes are covered with new/existing integration tests?_
</details>

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #3268, opened 2025-02-10
chore: Allow toolbar in AppLayout component to be hidden

georgylobko:
### Description

Introduce the navigationTriggerHide prop for the AppLayout toolbar (currently a private prop) and update the toolbar rendering condition. If drawers, navigation, and the split panel have no triggers, the toolbar will not be displayed.

This PR is part of the public layout component feature: [cloudscape-design/components#3208](https://github.com/cloudscape-design/components/pull/3208).

The goal of this PR is to release the private changes first to prevent potential merge conflicts.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/app-layout/__tests__/__snapshots__/widget-contract.test.tsx.snap
```diff
@@ -1593,24 +1593,12 @@ Map {
...
       "onNavigationToggle": [Function],
-      "onSplitPanelToggle": [Function],
-      "splitPanelFocusRef": {
-        "current": null,
-      },
-      "splitPanelToggleProps": {
-        "active": false,
-        "ariaLabel": undefined,
-        "controlId": "split-panel6",
-        "displayed": false,
```
georgylobko, line 1611: wow, if displayed is set to false for the split panel, all related props are excluded from the toolbar props

georgylobko (commented):
All errors from the codecov report have been covered in the ongoing PR: https://github.com/cloudscape-design/components/pull/3208

They are not covered within the scope of this PR, because they logically belong to the new public component

src/app-layout/visual-refresh-toolbar/multi-layout.ts
```diff
@@ -63,15 +68,24 @@ export function mergeProps(
...
+      !props.navigationTriggerHide &&
+      !checkAlreadyExists(!!toolbar.hasNavigation, 'navigation')
+    ) {
       toolbar.hasNavigation = true;
       toolbar.navigationOpen = props.navigationOpen;
       toolbar.navigationFocusRef = props.navigationFocusRef;
       toolbar.onNavigationToggle = props.onNavigationToggle;
     }
-    if (props.splitPanel && !checkAlreadyExists(!!toolbar.hasSplitPanel, 'splitPanel')) {
+    if (props.navigationTriggerHide) {
```
just-boris, line 81: Cross posted from #3208:

Why is this check needed? What will happen in this  combination?

```js
<AppLayout navigation="top" content={<AppLayout navigation="inner" navigationTriggerHide={true}>}>
```

georgylobko: This check is necessary because hasNavigation is set to true when the navigation prop is provided:

https://github.com/cloudscape-design/components/blob/main/src/app-layout/visual-refresh-toolbar/toolbar/index.tsx#L69

src/app-layout/__tests__/__snapshots__/widget-contract.test.tsx.snap
georgylobko: I'd say it is not worth it in this scenario because split panel visibility determines by triggering `setSplitPanelToggle` https://github.com/cloudscape-design/components/blob/main/src/split-panel/implementation.tsx#L67, which means that we need to render all widget components in the widget-contract test here: https://github.com/cloudscape-design/components/blob/main/src/app-layout/__tests__/widget-contract.test.tsx#L15, which would lead to rendering all widgets props, including refs with classNames which are different in every build. From my perspective this scenario is out of this widget contract scope. 

WDYT?

just-boris: ok, got it. This is still unfortunate that with this change all splitPanelToggle related props are fully wiped from the snapshot. We need to get these covered somehow.

How about adding an extra test, where `createWidgetizedComponentMock` instead of `return null` it will `return <Implementation {...props}>`? This will allow to capture the real split panel props

georgylobko: I thought about that and even implemented it, but ended up with a potentially flaky test where classNames are regenerated on every (?) build, so I'm afraid that at some point, once we get used to have this snapshot rewritten on every build, we might miss a real issue.

This PR introduces a new scenario when the toolbar is not rendered when one of the condition is hidden split panel trigger (closeBehavior=hidden). Therefore, I would add a scenario to check this case.

src/app-layout/visual-refresh-toolbar/multi-layout.ts
georgylobko: Yes, I can, but I don't really understand what would be the benefit of doing this

src/app-layout/__tests__/__snapshots__/widget-contract.test.tsx.snap
just-boris: The goal is to have at least one snapshot featuring `splitPanelToggleProps` there. Currently there are none

> ended up with a potentially flaky test where classNames are regenerated on every (?) build

This is solvable, if we stop including JSX into the snapshot

```js
function sanitizeProps(props) {
   return Object.fromEntries(Object.entries(props).map(([key, value]) => {
      return [key, ReactIs.isElement(value) ? '__JSX__' : value];
   }))
}


function createWidgetizedComponentMock(Implementation: React.ComponentType) {
  return () => {
    return function Widgetized(props: any) {
      renderedProps.set(Implementation, sanitizeProps(props));
      return <Implementation {...props}>;
    };
  };
}
```

P.S. this suggestion is to add an extra test with the new mock. The current tests and mock should also stay

src/app-layout/visual-refresh-toolbar/multi-layout.ts
georgylobko: Just double checked: that's not that simple to move it to a level above, because navigation trigger visibility is determined based on both `hasNavigation` coming from toolbarProps and `nagivation` coming from internal props. So `hasNavigation` should be explicitly set to `false` in this case: https://github.com/cloudscape-design/components/blob/main/src/app-layout/visual-refresh-toolbar/toolbar/index.tsx#L69

just-boris: > I don't really understand what would be the benefit of doing this

This is another cross-version data exchange. The less we do that – the better. That's why I still recommend to refactor it more future-proof

src/app-layout/__tests__/__snapshots__/widget-contract.test.tsx.snap
georgylobko: Done

src/app-layout/visual-refresh-toolbar/multi-layout.ts
georgylobko: Feel free to suggest a better solution. I went through the codebase and this one seems to be the easiest one

just-boris: Here is the one: https://github.com/cloudscape-design/components/pull/3278

We do not need to keep legacy support for 6 month old commit, can remove it now

src/app-layout/__tests__/__snapshots__/widget-contract-split-panel.test.tsx.snap
```diff
@@ -0,0 +1,1945 @@
...
+              aria-haspopup="true"
+              class="awsui_trigger_lpshu_13vrw_145 awsui_trigger-button-styles_lpshu_13vrw_145 awsui_selected_lpshu_13vrw_212 awsui_navigation-toggle_1fj9k_z5zo8_7"
+              data-shift-focus="awsui-layout-drawer-trigger"
+              type="button"
+            >
+              <span
+                class=""
+              >
+                <span
+                  class="awsui_icon_h11ix_tsgp4_189 awsui_size-normal-mapped-height_h11ix_tsgp4_248 awsui_size-normal_h11ix_tsgp4_244 awsui_variant-normal_h11ix_tsgp4_320"
```
just-boris, line 96: Yes, these classes is the problem which we just discussed: https://github.com/cloudscape-design/components/pull/3268#discussion_r1950614871

Did you try `sanitizeProps` function, why did it not work?

src/app-layout/__tests__/widget-contract-split-panel.test.tsx
```diff
@@ -0,0 +1,96 @@
...
+import AppLayout from '../../../lib/components/app-layout';
+import SplitPanel from '../../../lib/components/split-panel';
+import createWrapper from '../../../lib/components/test-utils/selectors';
+
+const isObject = (value: any) => Object.prototype.toString.call(value) === '[object Object]';
+const isRef = (value: any) => {
+  if (!isObject(value)) {
+    return false;
+  }
+  if ('current' in value && Object.keys(value).length === 1) {
```
georgylobko, line 17: Checking if `current` is a single key in the object as there is not recommended way to determine react refs. `React.isValidElement(props.current)` returns `false` as ref contains react node (react fiber to be specific), not react element. `ReactIs.isElement `also returns false.

src/app-layout/visual-refresh-toolbar/multi-layout.ts
georgylobko: done

src/app-layout/__tests__/__snapshots__/widget-contract-split-panel.test.tsx.snap
georgylobko: done

src/app-layout/__tests__/widget-contract-split-panel.test.tsx
just-boris: We could check `value instanceof Element` because what we really want is not to skip refs, but to prevent DOM nodes from serialization

--

Optional, for your consideration

just-boris (dismissed):

src/app-layout/__tests__/widget-contract-split-panel.test.tsx
georgylobko: Refactored this part a bit

just-boris (approved):

PR #3285, opened 2025-02-12
chore: Add linter for no 'only' tests

gethinwebster:
### Description

Self-explanatory: let's lint this rather than leaving it for humans to catch

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,6 +94,7 @@
     "eslint": "^8.28.0",
     "eslint-config-prettier": "^9.1.0",
     "eslint-plugin-header": "^3.1.1",
+    "eslint-plugin-no-only-tests": "^3.3.0",
     "eslint-plugin-no-unsanitized": "^4.0.2",
     "eslint-plugin-prettier": "^5.1.3",
     "eslint-plugin-react": "^7.31.11",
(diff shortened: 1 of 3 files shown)


georgylobko (approved):
That's a good one! Should the same be added to other repos, like board components?

gethinwebster:
> That's a good one! Should the same be added to other repos, like board components?

yeah makes sense, will open a few more PRs

package.json
```diff
@@ -94,6 +94,7 @@
     "eslint": "^8.28.0",
     "eslint-config-prettier": "^9.1.0",
     "eslint-plugin-header": "^3.1.1",
+    "eslint-plugin-no-only-tests": "^3.3.0",
```
just-boris, line 97: Late to the party, but why using this plugin instead of the Jest one. it offers this rule but also other useful for us too: https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/no-focused-tests.md

gethinwebster: good suggestion, following up here https://github.com/cloudscape-design/components/pull/3301

PR #3286, opened 2025-02-12
chore: Improve performance for repeated i18n format calls

avinashbot:
### Description

After a bit of a deep dive, I can reduce the time taken by the format function by 10x (in my contrived dev page example).

Related links, issue #, if available: AWSUI-60323

### How has this been tested?

Tested manually using the included dev page. Used a table rather than some other component because I wanted to reproduce their use-case closely rather than just rendering an i18n provider supported component over and over again. No automated tests, but we do have perf tests in the pipeline, so a regression here would presumably be visible there.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/i18n/performance.page.tsx
+++ b/pages/i18n/performance.page.tsx
@@ -1,0 +1,62 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import React, { useMemo, useState } from 'react';
+
+import Button from '~components/button';
+import Header from '~components/header';
+import I18nProvider from '~components/i18n';
+import enMessages from '~components/i18n/messages/all.en';
+import Table from '~components/table';
+
+export default function WrappedI18nPerformancePage() {
+  return (
+    <I18nProvider locale="en" messages={[enMessages]}>
+      <I18nPerformancePage />
+    </I18nProvider>
+  );
+}
+
+function I18nPerformancePage() {
+  const [itemCount, setItemCount] = useState(10_000);
+
+  const items = useMemo(
+    () =>
+      new Array(itemCount).fill(null).map((_value, i) => ({
+        id: `${i}`,
+        value: `Item ${i + 1}`,
+        children: [{ id: `${i}-0`, value: 'Child' }],
+      })),
+    [itemCount]
+  );
+
+  return (
+    <Table<{ id: string; value: string; children: { id: string; value: string }[] }>
+      header={
+        <Header
+          variant="h1"
+          counter={`(${itemCount})`}
+          description="Reproduction page for AWSUI-60323"
+          actions={<Button onClick={() => setItemCount(itemCount => itemCount + 1000)}>More items</Button>}
+        >
+          Performance test
+        </Header>
+      }
+      items={items}
+      trackBy="id"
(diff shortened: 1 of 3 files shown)


src/i18n/provider.tsx
```diff
@@ -66,38 +66,55 @@ export function I18nProvider({ messages: messagesArray, locale: providedLocale,
...
+
+    const cachedFormatter = localeFormatterCache.get(cacheKey);
+    if (cachedFormatter) {
+      // If an IntlMessageFormat instance was cached for this locale, just use that.
+      intlMessageFormat = cachedFormatter;
+    } else {
+      // Widen the locale string (e.g. en-GB -> en) until we find a locale
+      // that contains the message we need.
+      let message: string | MessageFormatElement[] | undefined;
+      const matchableLocales = getMatchableLocales(locale);
```
orangevolon, line 100: I was thinking if we can come up with a unit test case for this by just simply asserting against `getMatchableLocales` not being called.

src/i18n/provider.tsx
```diff
@@ -66,38 +66,55 @@ export function I18nProvider({ messages: messagesArray, locale: providedLocale,
     locale = 'en';
   }
 
+  // Create a per-render cache of messages and IntlMessageFormat instances.
+  // Not memoizing it allows us to reset the cache when the component rerenders
```
orangevolon, line 70: We are we resetting this cache?

src/i18n/provider.tsx
```diff
@@ -66,38 +66,55 @@ export function I18nProvider({ messages: messagesArray, locale: providedLocale,
...
+  // with potentially different locale or messages. We expect this component to
+  // be placed above AppLayout and therefore rerender very infrequently.
+  const localeFormatterCache = new Map<string, IntlMessageFormat>();
```
orangevolon, line 73: Micro-FE related question: Are we expecting this cache to be reset when the console page is changed?

pages/i18n/performance.page.tsx
```diff
@@ -0,0 +1,62 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
orangevolon, line 1: Are we measuring the performance somewhere for this page?

src/i18n/provider.tsx
avinashbot: Depends on how the page is structured - whether the navigation is a full page one or a router one, or whether the provider is placed outside the router or inside of it. But the main problem we're solving is the "thousands of calls in a single render" one, and so we're fine with this falling out of scope when big page changes happen.

pages/i18n/performance.page.tsx
avinashbot: No - I don't think we have a precedent for performance testing in our unit tests, and I'm a little worried about flakiness or resource usage here. I wanted to add this just so we have a "test bench" for i18n provider performance in the future.

src/i18n/provider.tsx
avinashbot: Or just directly mocking the IntlMessageFormat constructor and checking for that, since it's the more direct cause of perf issues. But a unit test checking the number of calls is a good call, will add that.

orangevolon (approved):

PR #3290, opened 2025-02-12
fix: Prevent filtering highlighting from crashing when match is >10000 chars

avinashbot:
### Description

Reported by a user on Slack. This is just #1885 but the limit is set even lower. Based on empirical testing, the regex limit seems to be set at 32768 characters now.

We initially set it to 100,000 characters last time to avoid breaking any legitimate workflows and it worked in previous Chrome versions. But in retrospect, autosuggest matches with over 10,000 characters are probably well past most reasonable uses of the component anyway, so this should be fine.

### How has this been tested?

Updated the tests.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/components/option/highlight-match.tsx
+++ b/src/internal/components/option/highlight-match.tsx
@@ -7,7 +7,7 @@ import styles from './styles.css.js';
 const splitOnFiltering = (str: string, highlightText: string) => {
   // We match by creating a regex using user-provided strings, so we skip
   // highlighting if the generated regex would be too memory intensive.
-  if (highlightText.length > 100000) {
+  if (highlightText.length > 10_000) {
     return { noMatches: [str], matches: null };
   }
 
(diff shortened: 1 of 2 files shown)


just-boris (approved):

src/internal/components/option/highlight-match.tsx
```diff
@@ -7,7 +7,7 @@ import styles from './styles.css.js';
 const splitOnFiltering = (str: string, highlightText: string) => {
   // We match by creating a regex using user-provided strings, so we skip
   // highlighting if the generated regex would be too memory intensive.
-  if (highlightText.length > 100000) {
+  if (highlightText.length > 10_000) {
```
jperals, line 10: I think it would be good to show some error/feedback in the UI, but I guess this is a net improvement if the component was previously crashing.

jperals (approved):

PR #3291, opened 2025-02-12
fix: fixing broken corners of table with sticky header bottom when scrolled

dpitcock:
### Fix for Sticky Header breaking bottom border at corners

The current implementation of the sticky header component can lead to a visual issue where the bottom corners of the sticky header jut out past the rounded borders of the table, causing a break in the visible border.

This pull request addresses this issue by making an enhancement to the useStickyHeader hook used by the Sticky Header and its parent component.

The key changes are:

Adding a new property called `isStuckAtBottom` to the object returned by the useStickyHeader hook. The isStuckAtBottom state is introduced to track whether the sticky header is aligned with the bottom edge of the table content. This is determined by comparing the bottom positions of the table container and the sticky header, and setting isStuckAtBottom accordingly. The internal table component then uses this state to apply an additional CSS class when the sticky header is in use, is stuck, and is stuck at the bottom edge, enabling the fix for the visual issue with the bottom border corners.

This fix should resolve the issue described in AWSUI-60364, providing a seamless visual experience for users when working with sticky headers in tables.

### How has this been tested?

- unit tests added and updated
- integ tests uneffected
- visual regression tests have only one expected change with the corrected borders in dev-v3-denpitco pipeline

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/container/styles.scss
+++ b/src/container/styles.scss
@@ -60,11 +60,9 @@
     }
   }
 
-  &.with-stuck-sticky-header {
-    &.with-stuck-sticky-header-at-bottom {
-      border-end-end-radius: 0;
-      border-end-start-radius: 0;
-    }
+  &.with-stuck-sticky-header-at-bottom {
+    border-end-end-radius: 0;
+    border-end-start-radius: 0;
   }
 }
 
--- a/src/container/styles.scss
+++ b/src/container/styles.scss
@@ -59,6 +59,13 @@
       inset-block-start: calc(-1 * #{awsui.$border-divider-section-width});
     }
   }
+
+  &.with-stuck-sticky-header {
+    &.with-stuck-sticky-header-at-bottom {
+      border-end-end-radius: 0;
+      border-end-start-radius: 0;
+    }
+  }
 }
 
 .with-side-media {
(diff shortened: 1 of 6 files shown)


src/table/sticky-header.tsx
```diff
@@ -78,6 +78,7 @@ function StickyHeader(
       tabIndex={-1}
       ref={secondaryWrapperRef}
       onScroll={onScroll}
+      data-stuck-at-bottom={isStuckAtBottom}
```
pan-kot, line 81: It looks like this data-attribute is not being used in the code.

src/container/styles.scss
```diff
@@ -59,6 +59,13 @@
       inset-block-start: calc(-1 * #{awsui.$border-divider-section-width});
     }
   }
+
+  &.with-stuck-sticky-header {
```
pan-kot, line 63: Why the with-stuck-sticky-header class needed?

src/table/sticky-header.tsx
dpitcock: Good catch, that was from earlier experimentation

src/container/styles.scss
dpitcock: I see, it is redundant because I use both bot conditionals to render the class

PR #3303, opened 2025-02-20
chore: hide scrollbars in dev pages when motion is disabled on Safari

jperals:
### Description

When updating Safari from 15 to 16 in our visual regression Safari tests, scrollbars appear and they happen to be flaky —with variable opacity, and sometimes not appearing at all, causing unexpected screenshot diffs. Therefore bringing back scrollbars hiding for Safari in our dev pages.

After this change, scrollbars are not visible when using Safari and motion is disabled.

_Note:_ I noticed that for the setting to have effect, motion needs to be disabled when the page is loaded. So in order to switch scrollbars on or off manually, it is necessary not only to toggle the motion disabled option (either via the checkbox or via the URL parameters) but also to reload the page. This seems to be a browser limitation and does not affect our tests.

This reverts commit f9aeff86a555776e50b7e55eb40e5b59fe708ead "in essence", but with two differences:
- Hides page-level scrollbars as well, not only those inside the body
- Hides scrollbars only in Safari, does not affect Chrome in MacOs (which is what most devs use to manually test)

Other options considered:
- Disabling scrollbars via feature flags. No such option found for Safari / Browserstack
- Adding a delay in test initialization. This would affect the tests for all browsers and for the sake of covering scrollbars in tests, they area already covered in Chrome tests.

### How has this been tested?

Ran in my pipeline. Scrollbars are consistently hidden in Safari components tests.

<details>

#### Correctness

#### Security

#### Testing

</details>

---


pages/app/styles.scss
```diff
@@ -20,6 +20,15 @@ body {
     caret-color: transparent !important;
   }
   // stylelint-enable @cloudscape-design/no-implicit-descendant, selector-max-type
+
+  html:has(&.safari) {
```
orangevolon, line 24: Two questions:

1. Where `&` points to? `<html>` or `<body>` tag?
2. Would it be possible to set `::-webkit-scrollbar` on `body` rather than `html` so it would make the selector simpler?
```css
body.safari::-webkit-scrollbar {
  inline-size: 0;
  block-size: 0;
}
```

jperals: 1. `&` expands to the parent selector. So in this case the selector becomes `html:has(:global(.awsui-motion-disabled).safari)`
2. This is on purpose, because there are scrollbars at the page level, and we want to hide them too. Just double-checked on Safari that targeting the `html` element instead of `body` is needed for this

orangevolon: Makes sense. I think we can add this conversation as a comment on top of the selector. Just to have a bit more context around it.

jperals: OK, added a comment

orangevolon (approved):

PR #3277, opened 2025-02-11
chore: Fix mocks application in app layout tests

just-boris:
### Description

Remove a lot of `jest.mock` calls

Related links, issue #, if available: n/a

### How has this been tested?

Test-only changes

<details>
   <summary>Review checklist</summary>

_The following items are to be evaluated by the author(s) and the reviewer(s)._

#### Correctness

- _Changes include appropriate documentation updates._
- _Changes are backward-compatible if not indicated, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#public-apis)._
- _Changes do not include unsupported browser features, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#browsers-support)._
- _Changes were manually tested for accessibility, see [accessibility guidelines](https://cloudscape.design/foundation/core-principles/accessibility/)._

#### Security

- _If the code handles URLs: all URLs are validated through [the `checkSafeUrl` function](https://github.com/cloudscape-design/components/blob/main/src/internal/utils/check-safe-url.ts)._

#### Testing

- _Changes are covered with new/existing unit tests?_
- _Changes are covered with new/existing integration tests?_
</details>

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #3318, opened 2025-02-27
feat: Table cell align for all cells

pan-kot:
### Description

This PR introduces a new API property for the table component, the: `cellVerticalAlign?: "middle" | "top"`. This property will complement the existing `columnDefinitions.verticalAlign?: "middle" | "top"` property, but is applicable for the entire table, including the selection column. When both properties are set, the `columnDefinitions.verticalAlign` takes precedence.

On the screenshot you can see how selection checkboxes are aligned to the cell's top when `cellVerticalAlign = "top"`. The last column alignment is overridden by setting the `columnDefinitions.verticalAlign = "middle"`.

Screenshot 2025-02-27 at 17 46 38

Rel: AWSUI-59979

### How has this been tested?

* Existing screenshot tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/table/__tests__/table-feature-metrics.test.tsx
+++ b/src/table/__tests__/table-feature-metrics.test.tsx
@@ -1,0 +1,50 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import * as React from 'react';
+import { render } from '@testing-library/react';
+
+import * as baseComponentHooks from '../../../lib/components/internal/hooks/use-base-component';
+import Table from '../../../lib/components/table';
+
+const useBaseComponentSpy = jest.spyOn(baseComponentHooks, 'default');
+
+jest.mock('../../../lib/components/internal/hooks/use-mobile', () => ({
+  useMobile: jest.fn(),
+}));
+
+test('reports cellVerticalAlign and columnDefinitionsVerticalAlign correctly', () => {
+  const def = (id: string, verticalAlign: 'middle' | 'top') => ({ id, header: '', cell: () => null, verticalAlign });
+
+  render(<Table columnDefinitions={[def('1', 'middle')]} items={[]} />);
+
+  expect(useBaseComponentSpy).toHaveBeenCalledWith(
+    'Table',
+    {
+      props: expect.objectContaining({ cellVerticalAlign: 'middle' }),
+      metadata: expect.objectContaining({ usesColumnDefinitionsVerticalAlign: false }),
+    },
+    expect.anything()
+  );
+
+  render(<Table columnDefinitions={[def('1', 'middle')]} items={[]} cellVerticalAlign="top" />);
+
+  expect(useBaseComponentSpy).toHaveBeenCalledWith(
+    'Table',
+    {
+      props: expect.objectContaining({ cellVerticalAlign: 'top' }),
+      metadata: expect.objectContaining({ usesColumnDefinitionsVerticalAlign: true }),
+    },
+    expect.anything()
+  );
+
+  render(<Table columnDefinitions={[def('1', 'top'), def('2', 'top')]} items={[]} cellVerticalAlign="top" />);
+
+  expect(useBaseComponentSpy).toHaveBeenCalledWith(
+    'Table',
+    {
+      props: expect.objectContaining({ cellVerticalAlign: 'top' }),
+      metadata: expect.objectContaining({ usesColumnDefinitionsVerticalAlign: false }),
+    },
+    expect.anything()
+  );
+});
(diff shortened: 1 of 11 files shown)


src/table/__tests__/table.test.tsx
```diff
@@ -14,6 +15,8 @@ import bodyCellStyles from '../../../lib/components/table/body-cell/styles.css.j
 import headerCellStyles from '../../../lib/components/table/header-cell/styles.css.js';
 import styles from '../../../lib/components/table/styles.css.js';
 
+const useBaseComponentSpy = jest.spyOn(baseComponentHooks, 'default');
```
orangevolon, line 18: Would it make sense to mock the dependencies (e.g. `useComponentMetrics` and `useComponentMetadata`) rather than an internal hook? The reason for this is that if we break something inside the internal hook this test will not be catching the regression.

src/table/index.tsx
```diff
@@ -61,6 +63,9 @@ const Table = React.forwardRef(
           hasInstanceIdentifier: Boolean(analyticsMetadata?.instanceIdentifier),
           usesVisibleColumns: !!props.visibleColumns,
           usesColumnDisplay: !!props.columnDisplay,
+          usesColumnDefinitionsVerticalAlign: props.columnDefinitions.some(
+            def => def.verticalAlign !== cellVerticalAlign
```
orangevolon, line 67: If they use and the value is the same as the `cellVerticalAlign` then the result would be `false`.
Shouldn't we test for `def.verticalAlign !== undefined`?

src/table/internal.tsx
```diff
@@ -595,6 +599,7 @@ const InternalTable = React.forwardRef(
                                       rowIndex,
                                       itemKey: `${getTableItemKey(row.item)}`,
                                     }}
+                                    verticalAlign={cellVerticalAlign}
```
orangevolon, line 602: Q: Why don't we override the cell's `verticalAlign` setting for this one?

src/table/__tests__/table.test.tsx
pan-kot: This test is not intended to capture possible issues inside the useBaseComponent, but only the props that we pass to it from the table. If we mock the inner hooks instead, that would be a violation of the hook's encapsulation. Instead, we treat it as a black box.

src/table/index.tsx
pan-kot: That is by design: if one sets both cellVerticalAlign and verticalAlign to the same value - that won't make any real difference. The purpose of the metric is to find instances where the verticalAlign is used purposefully.

src/table/interfaces.tsx
```diff
@@ -117,6 +117,14 @@ export interface TableProps<T = any> extends BaseComponentProps {
    * * `verticalAlign` ('middle' | 'top') - Determines the alignment of the content in the table cell.
    */
   columnDefinitions: ReadonlyArray<TableProps.ColumnDefinition<T>>;
+
+  /**
+   * Determines the alignment of the content inside table cells.
+   * This property affects all cells, including the ones in the selection column.
+   * To target individual cells use `columnDefinitions.verticalAlign`.
+   */
```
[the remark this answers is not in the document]
pan-kot: I think the `columnDefinitions.verticalAlign` taking precedence is the only reasonable assumption because otherwise it won't work at all if cellVerticalAlign is set.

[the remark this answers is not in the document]
orangevolon: I think that's reasonable to assume but making it explicit and clear would be even better in my opinion.

orangevolon (approved):

src/table/__tests__/table.test.tsx
just-boris: This mock does not only apply to your new tests, but for all tests in this file, which may lead to coverage issues.

If you want to test metrics reporting specifically, extract this into a separate `table-metrics.test.tsx`, do not do this into the main test file

src/table/__tests__/table-feature-metrics.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { render } from '@testing-library/react';
+
+import * as baseComponentHooks from '../../../lib/components/internal/hooks/use-base-component';
+import Table from '../../../lib/components/table';
+
+const useBaseComponentSpy = jest.spyOn(baseComponentHooks, 'default');
+
+jest.mock('../../../lib/components/internal/hooks/use-mobile', () => ({
+  useMobile: jest.fn(),
+}));
```
just-boris, line 13: Why is this mock needed?

Evergreen rule – less mocks, better tests

See this PR for an example why mocks are bad: https://github.com/cloudscape-design/components/pull/3277

pan-kot: That was a copy-paste artifact from the table tests file, it is not needed for the metrics tests - removed.

just-boris (approved):

PR #3322, opened 2025-02-28
fix: Add missing animations to discreet split panel

jperals:
### Description

The split panel is completely hidden when closed when `closeBehavior="hide"` (this is called the "discreet split panel"). When using this feature, the split panel has no enter or leave animations, which was not intended. This PR gives it the same treatment as the "non-discreet" split panel (when `closeBehavior="collapse"`) when it comes to animations. Note that in this sense, the following existing caveats in the "non-discreet" split panel also apply to the discreet one:

- Animations are only applied starting from the second interaction (related: https://github.com/cloudscape-design/components/pull/2970)
- In Toolbar VR, animations are only applied for viewport widths larger than 688px (related: https://github.com/cloudscape-design/components/blob/7556903a7a7dabebb7d6b43ee30e7353b4fae954/src/app-layout/resize/styles.scss#L10)

Ticket: `AWSUI-60450`

### How has this been tested?

- Manually using the `app-layout/dashboard-content-type` dev page
- Added motion test for that page
- Added unit test coverage for split panel with `closeBehavior="hide"` (which was missing)
- Dry-run to live

<details>

#### Correctness

#### Security

#### Testing

</details>

---


pages/app-layout/dashboard-content-type.page.tsx
```diff
@@ -23,12 +24,14 @@ export default function () {
         breadcrumbs={<Breadcrumbs />}
         navigation={<Navigation />}
         tools={<Tools>{toolsContent.long}</Tools>}
-        contentHeader={
```
jperals, line 26: Refactored to stop using the deprecated `contentHeader` property. This makes it easier to find this header in the newly added motion test.

src/__integ__/utils.ts
```diff
@@ -2,6 +2,16 @@
 // SPDX-License-Identifier: Apache-2.0
 import * as glob from 'glob';
 
+export type Theme = 'classic' | 'refresh' | 'refresh-toolbar';
```
jperals, line 5: De-duplicated Theme and Size from a few existing integration tests

src/__tests__/utils.tsx
```diff
@@ -62,16 +62,21 @@ function wrap({ default: Component }: { default: React.ComponentType }, Wrapper:
 
 export function requireComponent(componentName: string): any {
   if (componentName === 'split-panel') {
-    // eslint-disable-next-line @typescript-eslint/no-var-requires
+    // eslint-disable-next-line @typescript-eslint/no-var-requires, @typescript-eslint/no-require-imports
```
jperals, line 65: Had to add eslint disable comments for the `require` statements as I touched this file. I have no quick-and-easy alternatives for this, as sync `import` does not support dynamic (i.e, non hard-coded) paths. Async imports could be a solution, but this requires changes in all tests that consume these functions.

src/__tests__/utils.tsx
```diff
@@ -62,16 +62,21 @@ function wrap({ default: Component }: { default: React.ComponentType }, Wrapper:
...
+  // eslint-disable-next-line @typescript-eslint/no-require-imports
   return require(path.join(definitionsDir, componentName));
 }
 
 export function requireDesignTokensFile(fileName: string): any {
+  // eslint-disable-next-line @typescript-eslint/no-require-imports
   return require(path.join(designTokensDir, fileName));
 }
+
+export const testIf = (condition: boolean) => (condition ? test : test.skip);
```
jperals, line 82: De-duplicated `testIf` from existing unit tests

src/split-panel/__tests__/split-panel.test.tsx
```diff
@@ -71,221 +72,230 @@ afterEach(() => {
...
-    expect(wrapper.findOpenPanelSide()).not.toBeNull();
-  });
-
-  describe.each(['bottom', 'side'] as const)('Position %s', position => {
-    test('displays header, content and aria-label', () => {
-      const { wrapper } = renderSplitPanel({ contextProps: { position } });
-      expect(wrapper.findHeader().getElement()).toHaveTextContent('Split panel header');
-      expect(wrapper.getElement()).toHaveTextContent('Split panel content');
-      expect(wrapper.findSlider()!.getElement()).toHaveAttribute('aria-label', 'resizeHandleAriaLabel');
+  describe.each<SplitPanelProps['closeBehavior']>(['collapse', 'hide'])('closeBehavior=%s', closeBehavior => {
```
jperals, line 75: Wrapped all existing tests in this suite in a `describe.each` to test both values of `closeBehavior`.

Using `testIf` for the cases involving the open split panel button, because it doesn't exist when `closeBehavior="hide"`

src/split-panel/implementation.tsx
```diff
@@ -179,7 +180,7 @@ export function SplitPanelImplementation({
 
   const mergedRef = useMergeRefs(splitPanelRefObject, __internalRootRef);
 
-  if (closeBehavior === 'hide' && !isOpen) {
+  if (closeBehavior === 'hide' && !isOpen && animationDisabled) {
```
jperals, line 183: This is the key change: unless animations are still disabled render the split panel even if it is hidden (at least up to the relevant wrappers which contain the animated size changes), so that we can animate on enter and leave. Then the implementation of the side and bottom split panel respectively take care of not rendering internal content or using ARIA-hidden when hidden.

abdhalees (approved):

PR #3333, opened 2025-03-06
chore: Exposes internal dependencies for chart components

pan-kot:
### Description

Fixes small issues and exposes internal components to be consumed by new chart components (https://github.com/cloudscape-design/chart-components/pull/2).

### How has this been tested?

* New unit tests
* Screenshot tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/expandable-section/internal.tsx
```diff
@@ -109,7 +109,7 @@ export default function InternalExpandableSection({
         <ExpandableSectionHeader
           id={triggerControlId}
           descriptionId={descriptionId}
-          className={clsx(styles.header, styles[`header-${baseVariant}`])}
+          className={clsx(styles.header, styles[`header-${baseVariant}`], 'awsui-screenreader-text')}
```
pan-kot, line 112: The `awsui-screenreader-text` class name is used to locate the part of the expandable section to get extracted for screenreader announcement. It it used in the series-detail-text util. The same approach is to be reused in the chart-components, where we can't use the hashed internal class name.

src/internal/components/chart-popover/index.tsx
```diff
@@ -14,13 +14,15 @@ import { nodeBelongs } from '../../utils/node-belongs';
...
 
-interface ChartPopoverProps extends PopoverProps {
+export interface ChartPopoverProps extends PopoverProps {
   /** Title of the popover */
   title?: React.ReactNode;
 
   /** References the element the container is positioned against. */
-  trackRef: React.RefObject<HTMLElement | SVGElement>;
+  trackRef?: React.RefObject<HTMLElement | SVGElement>;
+  getTrack?: () => null | HTMLElement | SVGElement;
```
pan-kot, line 25: getTrack is an alternative for trackRef. The difference is that it is called every time the track element is requested, whilst the trackRef can get stale.

src/internal/components/chart-popover/index.tsx
```diff
@@ -128,19 +139,20 @@ function ChartPopover(
         keepPosition={true}
         allowVerticalOverflow={true}
         allowScrollToFit={isPinned}
+        hoverArea={true}
```
pan-kot, line 142: the hover-area style was moved to popover container and is controllable with a flag.

src/internal/components/chart-popover/index.tsx
```diff
@@ -48,6 +51,9 @@ interface ChartPopoverProps extends PopoverProps {
 
   /** Popover content */
   children?: React.ReactNode;
+
+  /** Popover footer */
+  footer?: React.ReactNode;
```
pan-kot, line 56: The footer is now a property, so no need to duplicate or expose the separate chart-popover-footer component.

build-tools/tasks/package-json.js
```diff
@@ -23,9 +23,15 @@ function getComponentsExports() {
     './contexts/form-field': './contexts/form-field.js',
     // Public internal components
     './internal/tooltip-do-not-use': './internal/tooltip-do-not-use/index.js',
```
pan-kot, line 25: I think the more scalable way to expose internal exports is to introduce a consistent namespace: `/internal/do-not-use/*`. I re-exported the tooltip from this namespace. Once the change is consumed by chat-components, the old one can be removed.

src/expandable-section/internal.tsx
jperals: Can we solve this alternatively by extracting a selector with this class to a scss file to be added to the `internal/do-not-use` directory?

This is an approach we are already using, and we would avoid customers using this class name directly.

src/internal/do-not-use/chart-tooltip.ts
```diff
@@ -0,0 +1,6 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import InternalChartTooltip, { ChartPopoverProps } from '../components/chart-popover/index.js';
+export type InternalChartTooltipProps = ChartPopoverProps;
+export { InternalChartTooltip };
```
jperals, line 6: Why not use default exports for these components?

src/internal/components/chart-popover/index.tsx
jperals: Any particular reason for this? Now `hover-area` and its associated styles are added to the generic popover component, although there is no other use case AFAICS.

src/popover/utils/positions.ts
```diff
@@ -265,11 +267,17 @@ export function calculatePosition({
       ? getIntersection([rect, viewport])
       : getIntersection([rect, viewport, container]);
 
-    const fitsWithoutOverflow =
-      visibleArea && visibleArea.inlineSize === body.inlineSize && visibleArea.blockSize === body.blockSize;
+    // When min height is set, the popover is considered fitting the container if the available space
+    // is the same or larger than min height, even if it means the scrollbar is needed.
```
jperals, line 271: Why? This looks to me like `minHeight` is used as a max rather than a min.

build-tools/tasks/package-json.js
```diff
@@ -23,9 +23,15 @@ function getComponentsExports() {
...
+    './internal/do-not-use/chart-tooltip': './internal/do-not-use/chart-tooltip.js',
+    './internal/do-not-use/expandable-section': './internal/do-not-use/expandable-section.js',
+    './internal/do-not-use/i18n': './internal/do-not-use/i18n.js',
+    './internal/do-not-use/tooltip': './internal/do-not-use/tooltip.js',
     './internal/widget-exports': './internal/widget-exports.js',
     './test-utils/dom/internal/tooltip': './test-utils/dom/internal/tooltip.js',
     './test-utils/selectors/internal/tooltip': './test-utils/selectors/internal/tooltip.js',
+    './test-utils/dom/internal/chart-tooltip': './test-utils/dom/internal/chart-tooltip.js',
```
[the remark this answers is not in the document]
pan-kot: Yes, we don't expect these to be used directly, plus the path still involves `internal`, and the risk of someone using internal test selectors is smaller.

src/internal/do-not-use/chart-tooltip.ts
pan-kot: That is to reinforce the component name, which includes the word "internal".

src/popover/utils/positions.ts
pan-kot: The idea was to force the popover into its position when its height is at least minHeight. For example, if the popover content is ~500px, the minHeight is set to 200px, and the available space is 300px - the popover will be rendered to the preferred position because its actual height is more or equal to min height. If the available space is less than minHeight - the popover will be rendered to the next suitable position instead.

jperals: I still don't understand. In this example, the actual popover content height (500) is more than the minHeight (200) so it should not be affected/constrained by it, right (minHeight is a min, not a max)? So how should the minHeight of 200 here affect the positioning, if the actual popover height will be the same regardless of minHeight?

jperals: Is `minHeight` something like "minimum vertical space that needs to be available in order let the positioning logic override the preferred position"?

pan-kot: Yes, exactly. Would you recommend a better name for the prop?

jperals: Yes :)

jperals: Maybe `minVisibleBlockSize` or `minVisibleHeight` is marginally better? The whole positioning logic concept is not easy to convey in the name but at least the "visible" part hopefully gives some hint to it.

My main problem with just `minHeight` is, it would suggest that it is for setting the actual popover height to a minimum value, e.g, adding white space below the inner content if necessary.

jperals: If you change it, please make sure to change it also wherever it is referenced elsewhere like comments and test descriptions.

pan-kot: Thanks, I will change it to minVisibleBlockSize 👍

jperals (approved):

src/internal/components/chart-popover/index.tsx
```diff
@@ -31,6 +33,7 @@ interface ChartPopoverProps extends PopoverProps {
     </>)
   */
   trackKey?: string | number;
+  minHeight?: number;
```
jperals, line 36: Do we want to change this name here accordingly as well now that we have the chance? Doing so later once chart-components is already using this might be harder

jperals (approved):

jperals (approved):

PR #3338, opened 2025-03-07
fix: Wrap alert and flash actions more responsively

gethinwebster:
### Description

Wrap alert and flash actions more responsively, based on actual content rather than breakpoints. This ensures they make the best use of the available space (not wrapping until smaller widths when content is small, and wrapping already at larger widths when content is longer).

Related links, issue #, if available: AWSUI-55633, AWSUI-60122

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/alert/internal.tsx
```diff
@@ -127,57 +138,59 @@ const InternalAlert = React.forwardRef(
...
-                <div className={clsx(styles.message, styles.text)}>
-                  <div
-                    className={clsx(
-                      header && styles.header,
-                      headerReplacementType !== 'original' ? styles.hidden : analyticsSelectors.header
-                    )}
-                    ref={headerRef}
-                  >
-                    {header}
+              <div className={styles['alert-wrapper']}>
```
gethinwebster, line 144: looking into this to see if it might cause problems with test utils

gethinwebster: (looks to be fine, no use of child combinators)

src/alert/styles.scss
```diff
@@ -76,6 +68,9 @@
 }
 
 .alert-focus-wrapper {
+  flex: 1;
+  min-inline-size: 70%; // fallback for browsers that don't support calc-size
+  min-inline-size: calc-size(max-content, min(size, 70%));
```
just-boris, line 73: Looks like a non-standard feature yet: https://developer.mozilla.org/en-US/docs/Web/CSS/calc-size

> Experimental. Expect behavior to change in the future.


Leaving this up to your consideration if there could be breaking changes in this feature, making it work differently in newer browser releases

src/alert/actions-wrapper/index.tsx
```diff
@@ -1,15 +1,35 @@
...
 
+import { getIsRtl } from '@cloudscape-design/component-toolkit/internal';
 import { getAnalyticsMetadataAttribute } from '@cloudscape-design/component-toolkit/internal/analytics-metadata';
 
 import InternalButton, { InternalButtonProps } from '../../button/internal';
 import { GeneratedAnalyticsMetadataAlertButtonClick } from '../analytics-metadata/interfaces';
 
 import styles from './styles.css.js';
 
+export const useActionsWrappingDetection = (containerWidth: number, style: string) => {
```
just-boris, line 14: Should we call it `className` instead of `style`?

src/alert/internal.tsx
```diff
@@ -85,9 +82,13 @@ const InternalAlert = React.forwardRef(
       replacementContentRef,
     } = useDiscoveredContent({ type, header, children });
 
+    const [containerWidth, containerMeasureRef] = useContainerWidth();
+    const containerRef = useMergeRefs(containerMeasureRef, __internalRootRef);
     const headerRef = useMergeRefs(headerRefAction, headerRefContent);
     const contentRef = useMergeRefs(contentRefAction, contentRefContent);
 
+    const actionsRef = useActionsWrappingDetection(containerWidth, styles['action-wrapped']);
```
just-boris, line 90: Instead of solving this here, could we propagate `containerWidth` into the `ActionsWrapper` and contain offset application logic inside of it?

Bonus effect of this, we can replace `actionsRef.current.classList.add/remove` (not the best React practice) with an `isWrapped` state and conditional class in JSX

src/alert/styles.scss
just-boris: I tried to understand CSS standardization process, but could not really figure how close is this spec to finalization 🤷

just-boris (dismissed):
Looks good, suggestions optional

src/alert/styles.scss
gethinwebster: agree it's not entirely clear, but I think the relatively limited usage of it here makes the risk acceptable

src/alert/internal.tsx
gethinwebster: good idea, I think that tidies it up nicely

just-boris (dismissed):

just-boris (approved):

PR #3354, opened 2025-03-17
fix: Fixes table sticky cells clicks

pan-kot:
### Description

Rel: AWSUI-60466

Before:

https://github.com/user-attachments/assets/4ec25600-9815-4e09-8365-c7dd59bf74ea

Now:

https://github.com/user-attachments/assets/c0641138-7cfb-45ae-a77d-a50069465c1d

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/table/__integ__/table-sticky-columns.test.ts
+++ b/src/table/__integ__/table-sticky-columns.test.ts
@@ -1,0 +1,28 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { BasePageObject } from '@cloudscape-design/browser-test-tools/page-objects';
+import useBrowser from '@cloudscape-design/browser-test-tools/use-browser';
+
+import createWrapper from '../../../lib/components/test-utils/selectors';
+
+import styles from '../../../lib/components/table/styles.selectors.js';
+
+test(
+  'does not cause table wrapper to scroll when clicking on an interactive element inside a sticky column',
+  useBrowser(async browser => {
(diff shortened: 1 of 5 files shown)


Al-Dani:
Did you check lighthouse tests?

Also, codecov is unhappy about the test coverage

YueyingLu:
The code LGTM. 
With manual check, keyboard focusing in Chrome still behaves a bit weird (for example, when focus change from checkbox to the first column, the rest non-sticky columns move), but it is not a regression.
It would be nice to have test cover the hook. Everything else looks to me fine :)

pan-kot:
> The code LGTM. With manual check, keyboard focusing in Chrome still behaves a bit weird (for example, when focus change from checkbox to the first column, the rest non-sticky columns move), but it is not a regression. It would be nice to have test cover the hook. Everything else looks to me fine :)

Thanks! The fix was only done for pointer interactions - the keyboard interactions are not supported due to the difference in the events order. However, the keyboard interaction should be a lesser concern, especially when enableKeyboardNavigation=true, in which case the transition from one cell to another is happening incrementally.

I think that particular interaction does not really require a test, especially considering the problem only occurs in Chrome. I will look at increasing the test coverage for the hook though!

pan-kot:
> Did you check lighthouse tests?
> 
> Also, codecov is unhappy about the test coverage

The lighthouse tests pass. I added unit test coverage and an integ test. I plan to add behaviour tests once merged to cover Safari and Firefox, too.

Al-Dani (approved):

PR #3355, opened 2025-03-18
chore: Pass theme through to funnel metrics

michaeldowseza:
### Description

- Pass the given environment theme down to the funnel metrics.
- Extract common visual theme code into helper function

### How has this been tested?
- Updated unit test
- Internal pipeline testing

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/analytics/components/analytics-funnel.tsx
+++ b/src/internal/analytics/components/analytics-funnel.tsx
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 import React, { useContext, useEffect, useRef, useState } from 'react';
 
-import { PACKAGE_VERSION } from '../../environment';
+import { PACKAGE_VERSION, THEME } from '../../environment';
 import { useDebounceCallback } from '../../hooks/use-debounce-callback';
 import { useUniqueId } from '../../hooks/use-unique-id';
 import { useVisualRefresh } from '../../hooks/use-visual-mode';
@@ -157,6 +157,7 @@ const InnerAnalyticsFunnel = ({ mounted = true, children, stepConfiguration, ...
         },
       ];
 
+      const componentTheme = THEME === 'polaris' && isVisualRefresh ? 'vr' : THEME;
       funnelInteractionId = FunnelMetrics.funnelStart({
         funnelName,
         funnelIdentifier: props.funnelIdentifier,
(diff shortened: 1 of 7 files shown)


src/internal/analytics/components/analytics-funnel.tsx
```diff
@@ -157,6 +157,7 @@ const InnerAnalyticsFunnel = ({ mounted = true, children, stepConfiguration, ...
         },
       ];
 
+      const componentTheme = THEME === 'polaris' && isVisualRefresh ? 'vr' : THEME;
```
just-boris, line 160: We have a similar piece of logic here: https://github.com/cloudscape-design/components/blob/7ff86dc1b6363ac4d16139f07590bd944799e019/src/internal/hooks/use-telemetry/index.ts#L9

Can we somehow deduplicate these, so we will not have conflicting data in different metrics?

michaeldowseza: Some more fun places: https://github.com/cloudscape-design/board-components/blob/a364adaceaca5f794ce4500c2b792c834f649971/src/internal/base-component/use-telemetry.ts#L10

michaeldowseza: Maybe a `useTheme(theme:string)` function in component-toolkit is needed

src/internal/analytics/components/analytics-funnel.tsx
```diff
@@ -157,6 +157,12 @@ const InnerAnalyticsFunnel = ({ mounted = true, children, stepConfiguration, ...
         },
       ];
 
+      let componentTheme = THEME;
+      if (THEME === 'polaris') {
+        // This is the only place we specify the theme as classic so we cannot reuse the getVisualTheme function :(
```
just-boris, line 162: Can we talk to the consumers of this code and ask them to stop relying on this?

Also how much is there real data with that field? According to my data, it is a very small number

just-boris (approved):

src/internal/analytics/components/analytics-funnel.tsx
michaeldowseza: Yes this was my plan after this. I need to dive into where they got used but for now just maintaining the existing functionality.

PR #3359, opened 2025-03-19
fix: Require less cursor precision when moving from tooltip target to tooltip content

avinashbot:
### Description

For accessibility reasons, you should be allowed to move your cursor from the tooltip trigger to the tooltip itself. And you can't do that in the slider. And you can also do that in other components, but you have to gently maneuver your cursor from the trigger through the arrow and into the tooltip, but that feels needlessly precise. So we create an extension to the tooltip's bounding box that extends the full height/width of the tooltip but covers the empty space next to the arrow. 

Slider needed a little more attention because it the mouseenter/mouseleave handlers were placed on an element too far away from where the tooltip was actually being rendered. So I just moved those handlers out to a parent div (new div because I guess the intention is to not make the logic cover the step labels).

Related links, issue #, if available: AWSUI-60211, AWSUI-60222

### How has this been tested?

Manually. There should be no visual or behavioral changes, only the sizes of the mouseover/mouseout triggers.

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/slider/__tests__/slider.test.tsx
```diff
@@ -325,7 +339,8 @@ describe('Slider i18n', () => {
 });
 
 describe('Slider a11y', () => {
-  test('Valides a11y', () => {
+  // FIXME: This test was never functional. This will be fixed in a subsequent PR.
```
avinashbot, line 342: See #3376

src/popover/container.scss
```diff
@@ -6,12 +6,60 @@
...
+      block-size: $arrow-height;
+    }
+  }
+
+  &:has(&-arrow-position-right-top),
+  &:has(&-arrow-position-right-bottom) {
+    &::before {
+      inset-block: 0;
+      inset-inline-start: -$arrow-height;
+      block-size: $arrow-height;
```
jperals, line 51: Looks like `block-size` should be `inline-size` in this and the next rule set.

I've added some styles with the dev tools to debug this pseudoelement, and it only has the expected position and dimensions when the popover/tooltip is anchored to the top or bottom, not left or right.

Screenshot 2025-03-26 at 08 24 16
Screenshot 2025-03-26 at 08 24 57

avinashbot: Yeah, whoops, good find - I messed up the logical properties. Fixed and manually tested more cases, so new approval request coming soon.

src/breadcrumb-group/item/item.tsx
```diff
@@ -28,40 +27,29 @@ const BreadcrumbItemWithPopover = <T extends BreadcrumbGroupProps.Item>({
...
 }: BreadcrumbItemWithPopoverProps<T>) => {
-  const [showPopover, setShowPopover] = useState(false);
-  const textRef = useRef<HTMLElement>(null);
-  const popoverContent = (
-    <Tooltip trackRef={textRef} value={item.text} size="medium" onDismiss={() => setShowPopover(false)} />
-  );
-
-  useEffect(() => {
-    if (showPopover) {
-      return registerTooltip(() => {
```
jperals, line 39: Do we need to remove this registry logic, and in the scope of this PR? _I think_ the purpose of this was to make sure two tooltips are never seen at the same time. Is there no chance we are not introducing regressions?

jperals: Or is this just handling hiding the tooltip on pressing Escape which is already handled by the tooltip [here](https://github.com/cloudscape-design/components/blob/0e10faf3e2a2463ebbacf77a09eabb3de5003781/src/internal/components/tooltip/index.tsx#L41-L63)?

src/slider/internal.tsx
```diff
@@ -127,120 +127,122 @@ export default function InternalSlider({
...
-        />
-      )}
       <div
-        ref={handleRef}
-        className={clsx(styles['tooltip-thumb'], {
-          [styles.readonly]: readOnly,
-        })}
-        style={{
-          [customCssProps.sliderTooltipPosition]: `calc(${percent}% - ${thumbSize}px)`,
+        onMouseEnter={() => {
```
jperals, line 131: There is now a sort of intermediate state in the slider where the tooltip is already open but the handle is not shown in hover state. This did not occur before —tooltip appearing and handle in hover state would always happen at the same time.

https://github.com/user-attachments/assets/efe05d8c-e0a2-425d-9e23-0229d30b0b9a

avinashbot: That intermediate state is what this PR explicitly enables - the whole idea is to allow the cursor to move off the slider handle and then into the tooltip much easier. I don't think it should show hover styling when the mouse is no longer on the handle, so this means that intermediate state has to exist.

jperals (approved):

PR #3367, opened 2025-03-21
chore: Add support for sticky options in internal options list component

jperals:
### Description

Needed as part of #3336. No public-facing changes.

### How has this been tested?

- Created dev page `dropdown/list-with-sticky-items` for manual testing
- Added unit test coverage
- Ran in my pipeline. No visual diffs
- Integration tests will be implemented in #3336

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/internal/components/option/interfaces.ts
```diff
@@ -38,6 +38,7 @@ export interface DropdownOption {
   disabled?: boolean;
   disabledReason?: string;
   option: OptionDefinition | OptionGroup;
+  afterHeader?: boolean;
```
jperals, line 41: Added this for the sticky option to be aware of whether there is content in the header slot of the list, because in this case it shouldn't have top border:

Screenshot 2025-03-21 at 12 16 44

The motivation for this is the Multiselect filter in combination with the "Select all" feature:

Screenshot 2025-03-21 at 12 20 33

src/internal/components/selectable-item/styles.scss
```diff
@@ -34,7 +34,8 @@
     border-block-end-color: transparent;
   }
 
-  &:not(:first-child) {
+  &:not(:first-child),
+  &.virtual {
```
jperals, line 38: Since we moved the "strut" in the virtual list implementation below the options (see changes in `select/parts/virtual-list.tsx`, this selector was not matching the first option anymore, therefore added `&.virtual` for the virtual list case.

src/internal/components/selectable-item/styles.scss
```diff
@@ -124,6 +125,52 @@
...
+    z-index: 4;
+
+    &:not(.highlighted):not(.selected) {
+      // Prevent covering the list border despite the higher z-index
+      border-inline-start-width: awsui.$border-item-width;
+      border-inline-start-color: awsui.$color-border-dropdown-container;
+      border-inline-end-color: awsui.$color-border-dropdown-container;
+      padding-inline: styles.$control-padding-horizontal;
+
+      &:not(.with-scrollbar) {
```
jperals, line 145: When a scrollbar is present in the list menu, we need to remove the right border and the top-right border radius in order to prevent this:

Screenshot 2025-03-19 at 16 38 12

pages/dropdown/list-with-sticky-item.page.tsx
```diff
@@ -0,0 +1,106 @@
...
+                  </Box>
+                ) : null
+              }
+            >
+              <ListComponent
+                menuProps={{ statusType: 'finished', ref, open }}
+                getOptionProps={(option, index) => ({ option: { ...option }, key: index, open })}
+                filteredOptions={options}
+                filteringValue={''}
+                firstOptionSticky={true}
```
jperals, line 96: Used the term `firstOptionSticky` in components which use the term "option", and `firstItemSticky` in components which use the term "item".

src/select/parts/plain-list.tsx
```diff
@@ -40,10 +44,19 @@ const PlainList = (
     listBottom,
     useInteractiveGroups,
     screenReaderContent,
+    firstOptionSticky,
   }: SelectListProps,
   ref: React.Ref<SelectListProps.SelectListRef>
 ) => {
+  const [width, menuMeasureRef] = useContainerQuery(
```
jperals, line 51: In VirtualList there was already a container query that I could reuse for this, but here in PlainList I had to add it.

src/select/parts/virtual-list.tsx
```diff
@@ -68,12 +77,19 @@ const VirtualListOpen = forwardRef(
...
       ariaSetsize: filteredOptions.length,
+      withScrollbar: hasScrollbar,
+      firstOptionSticky,
     });
 
     return (
       <OptionsList {...menuProps} ref={menuRef}>
-        <div aria-hidden="true" key="total-size" className={styles['layout-strut']} style={{ height: totalSize }} />
         {finalOptions}
+        <div
```
jperals, line 87: Moved the strut after the options so that sticky options stick "naturally" with `position: sticky`, otherwise they would get pushed to the bottom.

This is similar to the fact that the bottom option `listBottom` is render after the strut. So now the strut is in between.

src/internal/hooks/use-virtual/index.ts
```diff
@@ -34,8 +36,15 @@ export function useVirtual<Item extends object>({
...
   estimateSize,
+  firstItemSticky,
 }: UseVirtualProps<Item>): RowVirtualizer {
-  const rowVirtualizer = useVirtualDefault({ size: items.length, parentRef, estimateSize, overscan: 5 });
+  const rowVirtualizer = useVirtualDefault({
+    size: items.length,
+    parentRef,
+    estimateSize,
+    overscan: 5,
+    rangeExtractor: firstItemSticky ? stickyRangeExtractor : undefined,
```
jperals, line 46: React-Virtual docs for `rangeExtractor`: https://tanstack.com/virtual/latest/docs/api/virtualizer#rangeextractor

jperals: This is necessary to make sure that the sticky item (with index 0) is always present in the DOM, despite scrolling down with the virtual list.

georgylobko (commented):
I would suggest to introduce an integ test to make sure the sticky elements preserves its position during scroll.

jperals:
> I would suggest to introduce an integ test to make sure the sticky elements preserves its position during scroll.

Added here as part of #3336: https://github.com/cloudscape-design/components/pull/3336/files#diff-c7daac428eeda2da008dd7de44bfc8f94e3b101e07c6f8f7c790325b642aa004R33-R46

There will also be visual regression testing covering a scrolled state.

PR #3369, opened 2025-03-24
fix: Fixes table editable success state placement when items order change

pan-kot:
### Description

Once a table cell is edited, the successful state it added to a cell by col- and row index. If column definitions change, it causes the focus to be lost and the successful state resets. However, if rows order change, e.g. because the new value is sorted differently, the successful state then stays attached to the wrong cell. The PR fixes that for tables with trackBy property used.

See before:

https://github.com/user-attachments/assets/cee720b3-62ad-45e4-8f59-0f96eb1d8041

See after:

https://github.com/user-attachments/assets/36630e92-2cdf-4f0e-bf9c-91e459306aee

Rel: AWSUI-60542

### How has this been tested?

* New unit tests
* Manual tests in the integ test page

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/table/__integ__/inline-editing.test.ts
+++ b/src/table/__integ__/inline-editing.test.ts
@@ -165,7 +165,7 @@ describe.each([false, true])('enableKeyboardNavigation=%s', enableKeyboardNaviga
         await page.keys(['Tab', 'Tab']);
         await page.keys(['ArrowDown', 'ArrowRight']);
       } else {
-        await page.keys(range(11).map(() => 'Tab'));
+        await page.keys(range(12).map(() => 'Tab'));
       }
       const targetRow = (expandableRows ? distributionIdRow1 : domainNameRow1) as [number, number];
       await expect(page.isFocused(cellEditButton$(...targetRow))).resolves.toBe(true);
--- a/src/table/__tests__/table-editable.test.tsx
+++ b/src/table/__tests__/table-editable.test.tsx
@@ -1,0 +1,109 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import * as React from 'react';
+import { act, fireEvent, render, waitFor } from '@testing-library/react';
+
+import Table, { TableProps } from '../../../lib/components/table';
+import createWrapper from '../../../lib/components/test-utils/dom';
+
+import bodyCellStyles from '../../../lib/components/table/body-cell/styles.css.js';
+
+jest.mock('../../../lib/components/internal/hooks/use-mobile', () => ({
+  useMobile: jest.fn(),
+}));
+
+interface Item {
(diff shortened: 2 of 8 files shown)


src/table/__integ__/inline-editing.test.ts
```diff
@@ -165,7 +165,7 @@ describe.each([false, true])('enableKeyboardNavigation=%s', enableKeyboardNaviga
         await page.keys(['Tab', 'Tab']);
         await page.keys(['ArrowDown', 'ArrowRight']);
       } else {
-        await page.keys(range(11).map(() => 'Tab'));
+        await page.keys(range(12).map(() => 'Tab'));
```
jperals, line 168: Why this change?

src/table/__tests__/table-editable.test.tsx
```diff
@@ -0,0 +1,109 @@
...
+        await new Promise((resolve, reject) => setTimeout(reject, 1000));
+      }}
+      ref={ref}
+    />
+  );
+
+  const button = wrapper.findEditCellButton(2, 2)!;
+
+  fireEvent.click(button.getElement());
+  act(() => {
```
jperals, line 79: Is the `act` needed?

src/table/__integ__/inline-editing.test.ts
pan-kot: Because I made 2nd column sortable on the page - it made the header focusable, so an extra tab stop for when keyboard nav is not enabled.

src/table/__tests__/table-editable.test.tsx
pan-kot: I copied this from the other file, unchanged. I will check if the act can be removed, it probably can.

pan-kot: It works, but prints an annoying warning - I'd rather keep it as is.

src/table/internal.tsx
```diff
@@ -597,15 +598,17 @@ const InternalTable = React.forwardRef(
...
+                                      itemKey: rowId,
                                     }}
                                     verticalAlign={cellVerticalAlign}
                                   />
                                 )}
 
                                 {visibleColumnDefinitions.map((column, colIndex) => {
-                                  const isEditing = cellEditing.checkEditing({ rowIndex, colIndex });
-                                  const successfulEdit = cellEditing.checkLastSuccessfulEdit({ rowIndex, colIndex });
+                                  const colId = `${getColumnKey(column, colIndex)}`;
```
jperals, line 608: Unnecessary string interpolation, this could be just
```
const colId = getColumnKey(column, colIndex);
```

pan-kot: It is needed because getColumnKey() returns string or number. However, I made the cellId.col and cellId.row to only accept strings for better predictability.

jperals (approved):

PR #3370, opened 2025-03-24
chore: Optimise permutations pages

pan-kot:
### Description

Several permutations pages were above the max height limit. In the PR I optimised permutations to avoid duplicates or split them into separate pages. Additionally, the screenshot-area was updated to report the exceeded height as a console error, which breaks our integration tests conveniently.

Rel: https://github.com/cloudscape-design/components/pull/3363, AWSUI-60498

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


pages/date-range-picker/month-calendar-permutations.page.tsx
```diff
@@ -26,18 +26,34 @@ const intervals = [
...
     locale: ['en-GB'],
     startOfWeek: [1],
-    isDateEnabled: [() => true, () => false, (date: Date) => date.getDate() % 2 !== 0],
     onChange: [() => {}],
     timeInputFormat: ['hh:mm:ss'] as const,
     i18nStrings: [i18nStrings],
-    dateOnly: [false, true],
     customAbsoluteRangeControl: [undefined],
   })),
+  // Disabled dates
```
jperals, line 40: Non-blocking suggestion: Add h2 headers in between each permutation block to better map the parts to the code

pages/date-range-picker/month-calendar-permutations.page.tsx
```diff
@@ -26,18 +26,34 @@ const intervals = [
...
+  {
+    value: [{ start: { date: '2021-08-30', time: '' }, end: { date: '2021-09-03', time: '' } }],
+    setValue: [() => {}],
+    i18nStrings: [i18nStrings],
+    isDateEnabled: [() => false, (date: Date) => date.getDate() % 2 !== 0],
+    customAbsoluteRangeControl: [undefined],
```
jperals, line 46: Nitpick: likely unnecessary prop value. Same in the permutation block below

pages/date-range-picker/month-calendar-permutations.page.tsx
```diff
@@ -26,18 +26,34 @@ const intervals = [
 ];
 
 const permutations = createPermutations<DateRangePickerCalendarProps>([
+  // Selection range
```
jperals, line 29: Personally I'd create new pages instead of modifying new ones, to make the transition smoother. But I'll leave this up to you

pages/date-range-picker/month-calendar-permutations.page.tsx
pan-kot: I did it for the slider permutations, but can't do here as easily due to the way those dropdowns are rendered. If I add the labels - it messes up with the elements placement.

pages/date-range-picker/month-calendar-permutations.page.tsx
```diff
@@ -52,7 +68,7 @@ export default function DateRangePickerCalendarPage() {
     <Box padding="s">
       <h1>Date-range-picker month calendar page for screenshot tests</h1>
       <ScreenshotArea>
-        <div style={{ blockSize: `${intervals.length * 400}px` }}>
+        <div style={{ blockSize: `${(1 + permutations.length) * 400}px` }}>
```
jperals, line 71: Why not just apply some block-end margin instead of hard-coding the block size?

pages/date-range-picker/year-calendar-permutations.page.tsx
```diff
@@ -48,13 +56,13 @@ export default function DateRangePickerCalendarPage() {
     <Box padding="s">
       <h1>Date-range-picker year calendar page for screenshot tests</h1>
       <ScreenshotArea>
-        <div style={{ blockSize: `${intervals.length * 400}px` }}>
+        <div style={{ blockSize: `${(1 + permutations.length) * 300}px` }}>
```
jperals, line 59: Same question about block-end margin vs block-size

pages/date-range-picker/month-calendar-permutations.page.tsx
jperals: I see, the individual permutation items don't have a block size because they display dropdowns. But applying a block size to each of them instead of a combination of inset-block-size for each item + blockSize for the parent would be a little bit cleaner IMO.

pages/slider/permutations.page.tsx
```diff
@@ -96,7 +177,12 @@ export default function SliderPermutations() {
       <ScreenshotArea disableAnimations={true}>
         <PermutationsView
           permutations={permutations}
-          render={permutation => <Slider ariaLabel="Slider" i18nStrings={i18nStrings} {...permutation} />}
+          render={(permutation, index) => (
+            <Box>
```
jperals, line 181: Is it OK to include the box and the label in the permutation screenshot?

pages/slider/permutations.page.tsx
```diff
@@ -2,44 +2,146 @@
...
   {
+    label: ['Value'],
     value: [-10, 0, 10],
     min: [-10],
     max: [10],
-    hideFillLine: [false, true],
-    disabled: [false, true],
-    tickMarks: [false, true],
-    invalid: [false, true],
-    step: [0.5, 1, 5],
```
jperals, line 20: I think it would be nice to keep covering 0.5 as a value for `step `

pages/utils/screenshot-area.tsx
```diff
@@ -28,6 +28,19 @@ export default function ScreenshotArea({
...
   const [documentHeight, ref] = useContainerQuery(() => document.documentElement.scrollHeight);
+
+  const isExceedingHeightWarning =
+    documentHeight && documentHeight > MAX_PAGE_HEIGHT
+      ? `Warning: this page has height ${documentHeight}px, which is above the limit of ${MAX_PAGE_HEIGHT}px.
+      Taking a screenshot of this page may not include the full content. Remove ScreenshotArea if using this page for
+      integration tests only, or distribute test samples into separate pages to make screenshot tests work correctly.`
+      : '';
+
+  // Reporting the excessive height as a console error makes integration and screenshot tests for this page fail.
```
jperals, line 39: 1. Will this break any existing test?
2. Why not make the page completely error out instead of showing a warning, to minimize chances of devs ignoring the warning? We currently do this when the number of permutations exceeds a certain threshold: https://github.com/cloudscape-design/components/blob/main/pages/utils/permutations-view.tsx#L28

jperals (commented):
Do we have a chance to lower this number? https://github.com/cloudscape-design/components/blob/3668b0b7f2a9158568b7261762fb57d823f22453/pages/utils/permutations-view.tsx#L24

pages/date-range-picker/month-calendar-permutations.page.tsx
pan-kot: I think there is a benefit of having those on the same page as it makes it easier to compare the output and ensure the set is complete. I don't have a strong preference either.

pages/slider/permutations.page.tsx
pan-kot: That does look ok to me, as these components don't interfere with the slider. One disadvantage is that a change to the box styles would cause these permutations to fail, but that is true for the most of our screenshot tests anyways.

pages/slider/permutations.page.tsx
pan-kot: updated

pages/utils/screenshot-area.tsx
pan-kot: I originally intended to fix the DRP permutations only, but then more pages failed due to introducing this warning - so I fixed them all.

I think there is no strong reason to throw an error because any page with this warning will fail in the github actions because we run a11y tests on all pages. I don't mind the approach though

jperals: OK, I thought they would fail in the internal pipeline. If they fail in the PR checks, I still think failing during local development is even better, but I won't block the PR over this.

pages/slider/permutations.page.tsx
pan-kot: Yes, just did it

pan-kot:
> Do we have a chance to lower this number?
> 
> https://github.com/cloudscape-design/components/blob/3668b0b7f2a9158568b7261762fb57d823f22453/pages/utils/permutations-view.tsx#L24

I tried to lower it to 200 than 250 - but there are still pages that use more than that.

jperals:
> > Do we have a chance to lower this number?
> > https://github.com/cloudscape-design/components/blob/3668b0b7f2a9158568b7261762fb57d823f22453/pages/utils/permutations-view.tsx#L24
> 
> I tried to lower it to 200 than 250 - but there are still pages that use more than that.

All good, thanks for checking :)

jperals (approved):

PR #3371, opened 2025-03-24
fix: Implement TopNavigation breakpoints in CSS

### LINKED ISSUE #3337: [Bug]: TopNavigation size changes after SSR hydration
### Browser

_No response_

### Package version

v3.0.913

### React version

v18.3.1

### Description

On my page that uses SSR, I noticed that the top navigation immediately resized after pageload, going from a "narrow" mode to a larger style:

Top navigation changing size after pageload

I also noticed that the class `awsui_narrow_k5dlb_2ayr9_187` was attached to the top navigation in the page source, but removed after the page's JS executed.

I believe [this code is responsible for the issue](https://github.com/cloudscape-design/components/blob/main/src/top-navigation/internal.tsx#L98-L103).

Deferring the breakpoint logic until the code runs on the client wouldn't solve the issue, since there could still be a mismatch between the appearance of the component as rendered in SSR and its appearance on the client. I believe the best solution would be to use CSS media queries instead of JavaScript to implement the breakpoint-based appearance modifiers.

### Source code

_No response_

### Reproduction

Render the top navigation with SSR and open in a desktop browser. When I noticed the issue, my browser viewport was 1024x wide.

### Code of Conduct

TrevorBurnham:
### Description

This PR switches from JavaScript-based breakpoints to CSS container queries for TopNavigation, making its appearance consistent when rendered with SSR. I've added Sass mixins for this purpose that can be reused across other components to replace the `useBreakpoints` hook.

I've opted to use container queries rather than media queries for backward compatibility's sake, since `useBreakpoints` checks container size rather than viewport size. However, this introduces a complication: To use container queries, we need to add `container-type: inline-size` to the TopNavigation element's styles.

I noticed that adding `container-type: inline-size` causes overflowing content to get clipped in Safari due to a longstanding bug in that browser. [**Update 4/5:** After updating to Safari 18.4 on my Mac, I no longer see the bug, so it looks like Apple just fixed this.] To address that, I've enabled `expandToViewport` to render utility dropdowns in a portal. I've also added `position: relative` and `z-index: 1` to the `TopNavigation` styles to give it a stacking context, which appears to help with the issue. However, there is still a possibility that users who are rendering their own dropdowns in the `TopNavigation` (e.g. by using an `Autosuggest` searchbox) without `expandToViewport` may see issues in older versions of Safari where the dropdown is rendered behind other content with a higher `z-index`. I would suggest calling this out in the release notes and recommending the use of `expandToViewport` for dropdowns in `TopNavigation`.

Once this change is merged, I'd love to start applying this same change to other components that use `useBreakpoints`. Container queries are [supported in all modern browsers](https://caniuse.com/css-container-queries).

Fixes #3337 

### How has this been tested?

I've run this change locally, eyeballed it at different breakpoints, and clicked various controls to ensure that they're visible. I'm open to ideas for testing the breakpoint behavior more thoroughly.

<details>

#### Correctness

#### Security

#### Testing

</details>

---


pan-kot:
Hello @TrevorBurnham,

Thanks for contributing the fix!

Our team is taking a look and running internal screenshot tests and manual tests.

Could you please fix the failing unit tests? P.S. the "Measure bundle size" action is expected to fail, please ignore it.

TrevorBurnham:
@pan-kot Sure thing. I've pushed a fix for the unit tests: [51e8d78](https://github.com/cloudscape-design/components/pull/3371/commits/51e8d7828897b623d90c717fbd0e0b391e454a39)

TrevorBurnham:
Thanks! Can you give me instructions for viewing the component before/after? What I've been doing is running `npm run build` and then running a project against the built components directory, but some of the styles seem to be lost in that process (possibly because the "visual refresh" flag doesn't work).

pan-kot:
> Thanks! Can you give me instructions for viewing the component before/after? What I've been doing is running `npm run build` and then running a project against the built components directory, but some of the styles seem to be lost in that process (possibly because the "visual refresh" flag doesn't work).

The npm run build should be fine: you can build the mainline and then your commit, and compare the differences. The `./pages/lib` folder should contain the static assets for the dev pages.

TrevorBurnham:
I've updated the PR branch to fix the visual inconsistencies and improve code coverage.

One call-out: Previously `TopNavigation` passed a prop called `offsetRight` to the `ButtonTrigger` for utility menu dropdowns, then added styles based on the `.offset-right-*` class. Since the `offsetRight` value depended on the breakpoint in JavaScript, that approach doesn't work; there needs to be a CSS selector that can be used in the `TopNavigation` styles to reference the `ButtonTrigger` element. So I added a prop called `internalClass` to allow a classname to be passed through. Let me know if there's a better way to do this.

src/internal/components/menu-dropdown/index.tsx
```diff
@@ -20,6 +20,7 @@ export const ButtonTrigger = React.forwardRef(
   (
     {
       testUtilsClass,
+      internalClass,
```
pan-kot, line 23: We normally call this `className`, when it is applied to the top element. I would also add a comment, explicitly documenting the use case.

src/test-utils/dom/top-navigation/index.ts
```diff
@@ -101,10 +101,10 @@ export class TopNavigationMenuDropdownWrapper extends ButtonDropdownWrapper {
   }
 
   findTitle(): ElementWrapper | null {
-    return this.findByClassName(buttonDropdownStyles.title);
+    return createWrapper().findComponent(`.${buttonDropdownStyles.title}`, ElementWrapper);
```
pan-kot, line 104: What is the reason of this change?

src/test-utils/dom/button-dropdown/index.ts
```diff
@@ -40,20 +40,30 @@ export default class ButtonDropdownWrapper extends ComponentWrapper {
    * Finds an item in the open dropdown by item id. Returns null if there is no open dropdown.
    *
    * This utility does not open the dropdown. To find dropdown items, call `openDropdown()` first.
+   *
```
pan-kot, line 43: There are many button dropdown tests that use expandToViewport (see example: https://github.com/cloudscape-design/components/blob/main/src/button-dropdown/__tests__/button-dropdown-highlighted.test.tsx#L24), but the expandToViewport option wasn't needed. I think that is because the `findOpenDropdown()` method searches from the root, not component root - assuming there is at most one open dropdown. Why do we need the expandToViewport option now?

src/internal/styles/foundation/breakpoints.scss
```diff
@@ -39,3 +51,27 @@ $_largest_breakpoint: $breakpoint-x-large;
     @content;
   }
 }
+
+// Container query for widths greater than the given breakpoint.
```
pan-kot, line 55: Nice 👍

src/test-utils/dom/top-navigation/index.ts
TrevorBurnham: Without this change, the test

```
TopNavigation Utility part › Menu dropdown › has a title in the dropdown
```

fails to find the title element, due to the menu dropdown having `expandToViewport`.

src/test-utils/dom/button-dropdown/index.ts
TrevorBurnham: Good call-out. This change wasn't needed.

src/internal/components/menu-dropdown/index.tsx
TrevorBurnham: Done, thanks!

src/top-navigation/styles.scss
```diff
@@ -22,18 +24,52 @@
...
+        .utility-menu-dropdown-button {
+          margin-inline-end: awsui.$space-s;
+        }
+      }
+    }
   }
 
-  &.narrow > .padding-box {
-    block-size: awsui.$space-xxxl;
+  // Narrow container (width < 465px)
```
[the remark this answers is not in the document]
TrevorBurnham: Done, thanks.

TrevorBurnham:
I've rebased this branch and made the following changes:

1. Added `position: relative` and `z-index: 1` to the `TopNavigation` to work around a Safari bug where overflow content from containers with `container-type: inline-size` can be incorrectly rendered behind other elements, causing clipping. I've updated the PR description with more details about this bug, which I think is the most serious potential issue with this change.
2. Renamed `internalClassName` to just `className` and added an explanatory comment re: feedback.
3. Reduced unnecessary test changes and comments re: feedback.

TrevorBurnham:
@pan-kot I've rebased and updated this branch.

src/internal/components/menu-dropdown/interfaces.ts
```diff
@@ -19,12 +20,14 @@ export interface ButtonTriggerProps {
   expanded?: boolean;
 }
 
-export interface MenuDropdownProps extends InternalButtonDropdownProps {
+export interface MenuDropdownProps extends Omit<InternalButtonDropdownProps, 'items'> {
+  items: InternalButtonDropdownProps['items'];
```
pan-kot: why this change?

src/test-utils/dom/top-navigation/index.ts
pan-kot: There is an existing method in the dropdown wrapper, let's use it:
```
return this.findOpenDropdown()?.findByClassName(buttonDropdownStyles.title) ?? null;
```

src/internal/components/menu-dropdown/interfaces.ts
TrevorBurnham: Good call-out! I've removed this change.

pan-kot:
Discussed offline: before merging this one we need to merge ssr test server util, which is in turn blocked by migrating test pages to vite. I am closing the PR for now - but it can be reopened once the blockers are resolved.

PR #3379, opened 2025-03-26
fix: Fixes popover placement when rendered inside inline-size container

pan-kot:
### Description

Fixing a bug with incorrect popover placement when rendered inside a container with `containerType: inline-size`.

Before:

Screenshot 2025-03-26 at 09 21 03

After:

Screenshot 2025-03-26 at 09 21 22

### How has this been tested?

* Added new test page and integ test

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/utils/dom.ts
+++ b/src/internal/utils/dom.ts
@@ -10,7 +10,9 @@ export function isContainingBlock(element: HTMLElement): boolean {
   return (
     (!!computedStyle.transform && computedStyle.transform !== 'none') ||
     (!!computedStyle.perspective && computedStyle.perspective !== 'none') ||
-    (!!computedStyle.containerType && computedStyle.containerType !== 'normal') ||
+    (!!computedStyle.containerType &&
+      computedStyle.containerType !== 'normal' &&
+      computedStyle.containerType !== 'inline-size') ||
     computedStyle.contain?.split(' ').some(s => ['layout', 'paint', 'strict', 'content'].includes(s))
   );
 }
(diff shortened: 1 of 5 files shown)


avinashbot (changes requested):
Thanks for jumping on the fix! But this breaks in Safari 😓 

Screenshot 2025-03-26 at 09 31 23

My original logic was correct; but for an older version of the container-query spec, which Safari still implements. As I mentioned in the related bug thread, I think the easiest way to handle this is some sort of UA check specifically for Safari.

src/internal/utils/dom.ts
```diff
@@ -10,11 +10,17 @@ export function isContainingBlock(element: HTMLElement): boolean {
   return (
     (!!computedStyle.transform && computedStyle.transform !== 'none') ||
     (!!computedStyle.perspective && computedStyle.perspective !== 'none') ||
-    (!!computedStyle.containerType && computedStyle.containerType !== 'normal') ||
+    ![undefined, 'normal', 'inline-size'].includes(computedStyle.containerType) ||
+    // In Safari, container-type: "inline-size" creates a containing block
+    (isSafari() && ![undefined, 'normal'].includes(computedStyle.containerType)) ||
```
just-boris, line 15: What if this behavior changes in future versions? Are we going to get paged and make an emergency fix?

pan-kot: Yes, this can happen. I will contribute screenshot tests for the new test page so that we will see if the behaviour changes. There is also a workaround that can be applied on the consumer side: using the `contain` with value other than "none" forces a containing block and makes behaviour consistent across browsers.

pan-kot: I am drafting the alternative approach here: https://github.com/cloudscape-design/components/pull/3387

PR #3400, opened 2025-04-02
feat: CopyToClipboard - Allow separate content from text to copy for inline variant

ernst-dev:
### Description

**Original customer ask:**
> As of now, when `CopyToClipboard` uses `inline` variant, it will ignore `copyButtonText` and use `textToCopy` for both display and copy value.
Make `inline` variant to have different `copyButtonText` and `textToCopy` values like `button` variant.

**Customer usecase:**
> When you want to display a date with `CopyToClipboard`, you may want to display the date in localized, readable format but allow copying in language agnostic, timezone agnostic way, for example, UTC format.
It's a problem because as shown in the screenshot, different variants have different icon color and hurts visual uniformity without good reasons.

**Main problem:**
With any custom implementation, i.e. combining `SpaceBetween` using the `variant="icon"` and a string, it will result in a grey icon, this is particularly a problem when used in conjunction with regular inline `CopyToClipboard` which has a blue icon like inside a `SplitPanel` with `KeyValuePairs`.

**Solution:**
Adds a new property: `textToDisplay` which allows for specifying a separate display value from the copy value for `variant="inline"`.

This is achieved by assigning a `string` to `textToDisplay` and a separate `string` to `textToCopy`.
- e.g. a Human readable date/time to display but a timestamp to copy

If `content` has no value then `textToCopy` will be displayed instead.

`AWSUI-45434`
`65TZAG3kNuqE`

### How has this been tested?
- Manual
- Unit

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/__tests__/snapshot-tests/__snapshots__/documenter.test.ts.snap
+++ b/src/__tests__/snapshot-tests/__snapshots__/documenter.test.ts.snap
@@ -111,8 +111,8 @@ If the label is assigned via the \`i18nStrings\` property, this label will be ig
         "name": "AlertProps.Type",
         "type": "union",
         "values": [
-          "success",
           "error",
+          "success",
           "warning",
           "info",
         ],
@@ -165,9 +165,9 @@ exports[`Documenter definition for anchor-navigation matches the snapshot: ancho
     {
       "cancelable": false,
       "description": "Fired when an active anchor link changes.
+
 Note: This event is triggered both by the component's internal scroll-spy logic,
-or when the \`activeHref\` prop is manually updated.
-",
+or when the \`activeHref\` prop is manually updated.",
       "detailInlineType": {
         "name": "AnchorNavigationProps.Anchor",
         "properties": [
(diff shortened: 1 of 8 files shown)


src/copy-to-clipboard/__tests__/copy-to-clipboard.test.tsx
```diff
@@ -70,6 +70,29 @@ describe('CopyToClipboard', () => {
...
+
+    const { container } = render(<CopyToClipboard {...defaultProps} variant="inline" content="Copy test content" />);
+    const wrapper = createWrapper(container).findCopyToClipboard()!;
+
+    expect(wrapper.findCopyButton().getElement().textContent).toBe('');
+    expect(wrapper.findCopyButton().getElement()).toHaveAccessibleName('Copy');
+    expect(wrapper.findTextToCopy()!.getElement().textContent).toBe('Copy test content');
+
+    // Assert content written to the clipboard
+    act(() => wrapper.findCopyButton().click());
```
georgylobko, line 90: the `.click` that came from out test utils should already be wrapped with `act`, so you can remove it here

src/copy-to-clipboard/__tests__/copy-to-clipboard.test.tsx
```diff
@@ -70,6 +70,29 @@ describe('CopyToClipboard', () => {
...
+    expect(mockedWriteText).toHaveBeenCalledWith('Text to copy');
+
+    jest.resetAllMocks();
```
georgylobko, line 93: this is unnecessary since the mock for this test is already reset via `afterEach` on top of this file

src/__tests__/snapshot-tests/__snapshots__/documenter.test.ts.snap
ernst-dev, line 1: I believe this is such a large diff because of this change: https://github.com/cloudscape-design/components/pull/3413

pan-kot (approved):

PR #3409, opened 2025-04-08
feat: Support React 19

avinashbot:
### Description

Adds support for React 19. Includes two changes:
- Change version requirement to `">=16.8.0"` to make future react version support automatic rather than opt-in (if bugs come up, we'll fix them as bugs). Will do this for other packages too, to reduce the manual effort needed for every package when a new version comes out.
- Remove `react-dom` from `peerDependencies`. A little controversial, but having a peer dependency on react-dom broke a number of consumer package builds because they themselves didn't have a peer dependency on react-dom, and our peer dependency was pulling in react@19. So we decided to treat this one as an implicit peer dependency and hope for the best.

Related links, issue #, if available: AWSUI-60489

### How has this been tested?

Passed dry-run build. Lost the link to it, but started another one here: 7313558320.

<details>

#### Correctness

#### Security

#### Testing

</details>

---


package.json
```diff
@@ -165,7 +164,8 @@
     {
       "path": "lib/components/internal/widget-exports.js",
       "brotli": false,
-      "limit": "751 kB"
+      "limit": "751 kB",
+      "ignore": "react-dom"
```
orangevolon, line 168: Just out of curiosity, wouldn't this `ignore` also affect the `751 kb` size? I mean does it measure now without the `react-dom` and makes sure that it is less than 751KB? 

Also why do we need to ignore this at all?

avinashbot: It was automatically ignored before because it was part of `peerDependencies`. Now that it's not part of peer dependencies, it's included in the bundle calculation (it rises up to 8-something kB). I think that if we're ignoring react, it also makes sense to continue ignoring react-dom.

orangevolon (approved):

orangevolon (approved):

PR #3427, opened 2025-04-15
fix: Popover is positioned incorrectly when the layout changes

connorlanigan:
### Description

When an Annotation Popover is open, and the user interacts with the page so that the layout changes (e.g. toggling the side navigation panel), the Popover needs to update its position. With the existing heuristic, the Popover recalculates whenever the user clicks on the page (or interacts with a button using the keyboard), but the current logic assumes that any layout update then happens in a single frame. In the case of the navigation panel opening, it slides in with an animation, so the Popover only adapts to the first frame.

With this change, the Popover continuously updates its position for one second after the user interacts with the page. This should give enough time for any layout-related animations to finish, and it smoothly updates the Popover's position while its trigger moves. Since we're using `requestIdleCallback`, the browser will only execute these updates if the CPU is otherwise idle.

- AWSUI-60630

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/popover/container.tsx
+++ b/src/popover/container.tsx
@@ -99,11 +99,11 @@ export default function PopoverContainer({
     /*
     This is a heuristic. Some layout changes are caused by user clicks (e.g. toggling the tools panel, submitting a form),
     and by tracking the click event we can adapt the popover's position to the new layout.
-
-    TODO: extend this to Enter and Spacebar?
     */
 
-    const onClick = (event: UIEvent | KeyboardEvent) => {
+    const controller = new AbortController();
+
+    const onClick = async (event: UIEvent | KeyboardEvent) => {
       if (
         // Do not update position if keepPosition is true.
         keepPosition ||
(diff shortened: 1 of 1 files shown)


src/popover/container.tsx
```diff
@@ -114,14 +114,22 @@ export default function PopoverContainer({
...
+      const targetTime = performance.now() + 1_000;
+
+      while (performance.now() < targetTime) {
+        if (controller.signal.aborted) {
+          break;
+        }
+
         updatePositionHandler();
-      });
+        await new Promise(r => requestIdleCallback(r));
```
orangevolon, line 127: Not sure if having an extra micro-task here is necessary. It took me some time to understand what's happening. 

The way I read this is that: we create a micro-task, wait for a free cycle and then resolve the promise. This means the calculation doesn't happen in the cpu's free time, we just wait for a free CPU time to run the next cycle.

A simpler approach IMO would be simply having animation pattern:

```tsx
function updatePosition() {
  if (controller.signal.aborted) return;
  if (performance.now() >= targetTime) return;
  
  updatePositionHandler();
  requestIdleCallback(updatePosition);
}

requestIdleCallback(updatePosition);
```

orangevolon (commented):
Wondering if we could find a way to have a unit test for the `1000ms` part of the logic.

src/popover/container.tsx
```diff
@@ -114,14 +114,29 @@ export default function PopoverContainer({
...
+          break;
+        }
+
         updatePositionHandler();
-      });
+
+        if (typeof requestIdleCallback !== 'undefined') {
+          await new Promise(r => requestIdleCallback(r));
+        } else {
+          await new Promise(r => setTimeout(r, 50));
```
orangevolon, line 133: wouldn't it be better to fallback to `requestAnimationFrame` rather than `setTimeout`?

orangevolon (approved):

src/popover/container.tsx
connorlanigan: The position update does not necessarily need to be a smooth animation at the full refresh rate that the browser can do, it just shouldn't lag behind too much. So by using `setTimeout`, we get a reasonably good-looking behaviour without demanding too much performance from a potentially non-idle CPU.

PR #3429, opened 2025-04-16
Do not bail on calculating position of popover if its trigger is inside Shadow DOM

n1313:
### Description

This PR fixes a problem that prevents Popovers from working from inside Shadow DOM.

Currently, Popover bails on calculating dimensions of its body if its trigger does not belong to the document. This check is done using `nodeContains` utility. This is an optimization to prevent calculations from running in virtual environments where Popover wouldn't work anyway. However, this also prevents Popover from working from inside Shadow DOM, as the `nodeContains` doesn't work across the Shadow DOM boundary and the check fails.

This PR fixes this by checking if Popover is rendered inside Shadow DOM, and if so, checking that the trigger is contained inside it. 

With this change customers will be able to use Popover in applications that use Shadow DOM.

Related links, issue #, if available: related to AWSUI-60527.

### How has this been tested?

1. unit tests
2. I have built a small playground page with a webcomponent wrapped in Shadow DOM that contains a Popover. I have verified that the Popover does not work in the playground before the fix, and does work after

   | before | after |
   |--|--|
   |before|after|

   I can include the playground code into this PR, but the code is a bit hacky.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/popover/__tests__/popover.test.tsx
+++ b/src/popover/__tests__/popover.test.tsx
@@ -42,6 +42,12 @@ class PopoverInternalWrapper extends PopoverWrapper {
     }
     return this.findByClassName(styles.body);
   }
+  findContainer({ renderWithPortal } = { renderWithPortal: false }): ElementWrapper | null {
+    if (renderWithPortal) {
+      return createWrapper().findByClassName(styles.container);
+    }
+    return this.findByClassName(styles.container);
+  }
 }
 
 function renderPopover(props: PopoverProps & { ref?: React.Ref<PopoverProps.Ref> }) {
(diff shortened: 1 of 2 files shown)


src/popover/use-popover-position.ts
```diff
@@ -63,10 +63,14 @@ export default function usePopoverPosition({
       const document = popover.ownerDocument;
       const track = trackRef.current;
 
+      const rootNode = popover.getRootNode();
+      const isInShadowDom = rootNode !== document && rootNode.nodeType === Node.DOCUMENT_FRAGMENT_NODE;
```
pan-kot, line 67: I think this check `rootNode !== document`  might result in false positives when popover is rendered within an iframe, in which case the getRootNode() will return the iframe's document, not top document.

n1313: I will have to double-check this, but my first thought is that the iframe's document wouldn't be a `Node.DOCUMENT_FRAGMENT_NODE`.

pan-kot: You're right. I haven't realised that is not the global document but `popover.ownerDocument`. I think it would be nice to add a new util to utils/dom, something like:

```
function isInShadowDom(element: Element) {
  const rootNode = element.getRootNode();
  return rootNode !== element.ownerDocument && rootNode.nodeType === Node.DOCUMENT_FRAGMENT_NODE;
}
```

src/popover/__tests__/popover.test.tsx
```diff
@@ -415,3 +421,90 @@ test('does not add portal to the body unless visible', () => {
...
+    // this assertion fails if Popover bails early on calculating its position
+    expect(wrapper.findContainer()!.getElement()).toHaveAttribute(
+      'style',
+      'inset-block-start: 0; inset-inline-start: 13px;'
+    );
+  });
+
+  it('renders the popover header correctly with portal', () => {
+    const wrapper = renderPopoverInShadowDom({ children: 'Trigger', header: 'Memory error', renderWithPortal: true });
+    wrapper.findTrigger().click();
```
pan-kot, line 502: Considering the complications of supporting the click() event, how about we scope the tests to the popover container (internal) component instead? You can add a new test file `popover-container.test.tsx` and assert the expected placement, but without the need to make it visible first.

PR #3443, opened 2025-04-23
fix: Fix off-center popover flash when clicking CopyToClipboard

TrevorBurnham:
### Description

Because the clipboard API is async, the `CopyToClipboard` popover briefly renders in the "pending" state before rendering in the "success" or "error" state, which is perceived as a visual glitch.

The fix here is to wait to open the popover until we're out of the "pending" state. To achieve that, I've modified the `InternalPopover` API so that it doesn't render a popover if the given `content` is `null`, regardless of its internal `visible` state.

### How has this been tested?

You can replicate the bug and verify this fix in the local demo: Run `npm run start`, open the local URL, then go to any of the `CopyToClipboard` examples and click around.

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/copy-to-clipboard/internal.tsx
+++ b/src/copy-to-clipboard/internal.tsx
@@ -70,7 +70,9 @@ export default function InternalCopyToClipboard({
       triggerType="custom"
       dismissButton={false}
       renderWithPortal={popoverRenderWithPortal}
-      content={<InternalStatusIndicator type={status}>{statusText}</InternalStatusIndicator>}
+      content={
+        status === 'pending' ? null : <InternalStatusIndicator type={status}>{statusText}</InternalStatusIndicator>
+      }
       __onOpen={onClick}
     >
       <InternalButton
--- a/src/popover/__tests__/popover.test.tsx
+++ b/src/popover/__tests__/popover.test.tsx
@@ -153,6 +153,45 @@ describe('Dismiss button', () => {
       });
       expect(wrapper.findBody({ renderWithPortal })).toBeTruthy();
     });
+
+    it('does not render the popover body if content is null when a click is fired on the trigger', () => {
+      const wrapper = renderPopover({ children: 'Trigger', content: null, renderWithPortal });
+      wrapper.findTrigger().click();
+      act(() => {
+        wrapper
+          .findTrigger()
+          .getElement()
+          .dispatchEvent(new MouseEvent('mousedown', { bubbles: true }));
+      });
+      expect(wrapper.findBody({ renderWithPortal })).toBeNull();
+    });
+
+    it('renders the popover body if content becomes non-null after a click is fired on the trigger', () => {
+      const { container, rerender } = render(
+        <Popover content={null} renderWithPortal={renderWithPortal}>
+          Trigger
+        </Popover>
+      );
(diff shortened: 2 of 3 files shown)


src/copy-to-clipboard/internal.tsx
```diff
@@ -70,7 +70,9 @@ export default function InternalCopyToClipboard({
       triggerType="custom"
       dismissButton={false}
       renderWithPortal={popoverRenderWithPortal}
-      content={<InternalStatusIndicator type={status}>{statusText}</InternalStatusIndicator>}
+      content={
+        status === 'pending' ? null : <InternalStatusIndicator type={status}>{statusText}</InternalStatusIndicator>
+      }
```
cansuaa, line 75: When the content is passed as null, the screen reader stops working, at least VoiceOver doesn't announce the popover text anymore. You can try [here](https://d21d5uik3ws71m.cloudfront.net/components/0624f337ccdabf118390527ee956ab4a34b5fff5/dev-pages/index.html#/light/copy-to-clipboard/scenario-split-panel). 
It only announced the popover text when I commented this line, which could be because the content changes from null to something else very fast. I will reach out to someone from a11y in our team to clarify why this could be happening.

TrevorBurnham: That's very interesting! I asked Claude about this, and it said:

> Screen readers need time to register the live region before it can announce changes. A small delay (50-100ms) between adding the element and updating its content helps ensure reliable announcements.

So the old copy behavior was (probably accidentally?) allowing the live region to be registered. This seems like a more general issue with popovers that don't have a dismiss button (since those that do have a dismiss button don't use `aria-live`).

I found that using `LiveRegion` instead of `aria-live` makes the announcement work, likely because `LiveRegion` uses timers to work around this issue. I've pushed a proof of concept here: https://github.com/TrevorBurnham/cloudscape-components/commit/664ac49d5475e4894b52ab7a0e275254c95e6b1f Let me know if you'd like me to open up a separate PR for that change.

TrevorBurnham: By the way, this issue is addressed in the new [Popover API](https://developer.mozilla.org/en-US/docs/Web/API/Popover_API): Anything with the [`popover` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/popover) is announced to screen readers as soon as it's displayed.

Using that attribute would require a deeper rethinking of the popover component, and browser support isn't quite where it needs to be for Cloudscape to adopt it yet (it requires Safari 17), but I thought I'd note it in case using the popover API is on the roadmap.

At any rate, I think using `LiveRegion` is the best workaround for now, so let me know how you'd like me to proceed: Should I open a separate PR to use `LiveRegion` in popovers, then revisit this PR once that's merged?

cansuaa: Hi again, I discussed this with another colleague and they pointed out that this approach taken in the PR basically enables controlling popover's open state by passing `null` value for `content` which is not something we want. We already have an internal ticket for this problem, which I will mention this PR there. 
Thank you for contributing but we cannot accept this PR for the given reason.

just-boris:
Closing in favor of the follow up PR: https://github.com/cloudscape-design/components/pull/3855

PR #3465, opened 2025-04-29
chore: Handle undefined values when validating props

just-boris:
### Description

We have a new function to validate in runtime if there is a property from an unsupported system.

Two fixes in this PR

1. Do not throw in this use-case `<Button variant={undefined}>`
2. Add `isDevelopment` condition to remove this code from production builds

### How has this been tested?

Updated unit test

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/base-component/index.ts
+++ b/src/internal/base-component/index.ts
@@ -5,6 +5,7 @@ import { initAwsUiVersions } from '@cloudscape-design/component-toolkit/internal
 
 import { AnalyticsMetadata } from '../analytics/interfaces';
 import { PACKAGE_SOURCE, PACKAGE_VERSION, THEME } from '../environment';
+import { isDevelopment } from '../is-development';
 
 // these styles needed to be imported for every public component
 import './styles.css.js';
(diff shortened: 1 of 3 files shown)


src/internal/base-component/index.ts
```diff
@@ -45,11 +46,14 @@ export function validateProps(
...
 ) {
+  if (!isDevelopment) {
+    return;
+  }
   for (const [prop, value] of Object.entries(props)) {
     if (excludedProps.includes(prop)) {
       throw new Error(`${componentName} does not support "${prop}" property when used in ${THEME} theme`);
     }
-    if (allowedEnums[prop] && !allowedEnums[prop].includes(value)) {
+    if (value && allowedEnums[prop] && !allowedEnums[prop].includes(value)) {
```
jperals, line 56: This will also not throw if `value` is `false`. Is this intended?

just-boris: This function is currently used only for string enums, I do not expect a boolean there.

If something changes in the future, we can update the function

jperals (approved):

PR #3477, opened 2025-05-06
feat: Treeview

cansuaa:
### Description

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/components/expand-toggle-button/index.tsx
+++ b/src/internal/components/expand-toggle-button/index.tsx
@@ -1,0 +1,42 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import React, { useRef } from 'react';
+import clsx from 'clsx';
+
+import InternalIcon from '../../../icon/internal';
+import { useSingleTabStopNavigation } from '../../../internal/context/single-tab-stop-navigation-context';
+
+import styles from './styles.css.js';
+
+export function ExpandToggleButton({
+  isExpanded,
+  onExpandableItemToggle,
+  expandButtonLabel,
+  collapseButtonLabel,
+}: {
+  isExpanded?: boolean;
+  onExpandableItemToggle?: () => void;
+  expandButtonLabel?: string;
+  collapseButtonLabel?: string;
+}) {
+  const buttonRef = useRef<HTMLButtonElement>(null);
+  const { tabIndex } = useSingleTabStopNavigation(buttonRef);
+  return (
+    <button
+      type="button"
+      ref={buttonRef}
+      tabIndex={tabIndex}
+      aria-label={isExpanded ? collapseButtonLabel : expandButtonLabel}
+      aria-expanded={isExpanded}
+      className={styles['expand-toggle']}
+      onClick={onExpandableItemToggle}
+    >
+      <InternalIcon
+        size="small"
+        name="caret-down-filled"
+        className={clsx(styles['expand-toggle-icon'], isExpanded && styles['expand-toggle-icon-expanded'])}
+      />
+    </button>
+  );
+}
--- a/build-tools/utils/pluralize.js
+++ b/build-tools/utils/pluralize.js
@@ -80,6 +80,7 @@ const pluralizationMap = {
   TokenGroup: 'TokenGroups',
   TopNavigation: 'TopNavigations',
   TreeView: 'TreeViews',
+  Treeview: 'Treeviews',
   TutorialPanel: 'TutorialPanels',
   Wizard: 'Wizards',
 };
--- a/pages/treeview/basic.page.tsx
+++ b/pages/treeview/basic.page.tsx
@@ -1,0 +1,279 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import React, { useState } from 'react';
+
+import Badge from '~components/badge';
+import Box from '~components/box';
+import ButtonGroup from '~components/button-group';
+import Container from '~components/container';
+import Header from '~components/header';
+import Icon from '~components/icon';
+import SpaceBetween from '~components/space-between';
+import StatusIndicator from '~components/status-indicator';
+import Treeview, { TreeviewProps } from '~components/treeview';
+
+const progressiveStepContent = (
+  <div style={{ display: 'flex' }}>
+    <StatusIndicator type="warning">Checked 5 nodes</StatusIndicator>
+    <div style={{ borderLeft: '1px solid grey', marginLeft: '8px', marginRight: '8px' }} />
+    <SpaceBetween direction="horizontal" size="s">
+      <StatusIndicator type="warning">1</StatusIndicator>
+      <StatusIndicator type="in-progress">1</StatusIndicator>
+      <StatusIndicator type="success">2</StatusIndicator>
+    </SpaceBetween>
+  </div>
+);
+
+const progressiveStepItemsContent = [
+  <StatusIndicator key={'success-child'} type="success">
+    node-17 (eksclu-node-12345)
+  </StatusIndicator>,
+  <StatusIndicator key={'loading-child'} type="loading">
+    node-18 (eksclu-node-09876)
+  </StatusIndicator>,
+  <StatusIndicator key={'success-child-2'} type="success">
+    node-19 (eksclu-node-ab123)
+  </StatusIndicator>,
+];
+
+const items: TreeviewProps.TreeItem[] = [
+  {
+    id: '1',
+    content: 'Item 1',
+    items: [
+      {
+        id: '1.1',
+        content: 'Item 1.1',
+        items: [
+          {
+            id: '1.1.1',
+            content: 'Item 1.1.1',
+          },
+          {
+            id: '1.1.2',
+            content: 'Item 1.1.2',
+          },
+        ],
+      },
+      {
+        id: '1.2',
+        content: 'Item 1.2',
+      },
+      {
+        id: '1.3',
+        content: 'Item 1.3',
+        details: <Box color="text-status-inactive">us-east-1</Box>,
+        actions: <Actions />,
+        items: [
+          {
+            id: '1.3.1',
+            content: 'Item 1.3.1',
+          },
+          {
+            id: '1.3.2',
+            content: 'Item 1.3.2',
+          },
+        ],
+      },
+    ],
+  },
+  {
+    id: '2',
+    content: 'Item 2',
+  },
+  {
+    id: '3',
+    content: 'Item 3',
+    items: [
+      {
+        id: '3.1',
+        content: 'Item 3.1',
+      },
+    ],
+  },
+  {
+    id: '4',
+    content: <RdsAccessRoleTreeItemContent />,
+    items: [
+      {
+        id: '4.1',
+        content: 'Item 4.1',
+        items: [
+          {
+            id: '4.1.1',
+            content: 'Item 4.1.1',
+          },
+          {
+            id: '4.1.2',
+            content: 'Item 4.1.2',
+            details: <Box color="text-status-inactive">us-east-1</Box>,
+            items: [
+              {
+                id: '4.1.2.1',
+                content: 'Item 4.1.2.1',
+              },
+            ],
+          },
+          {
+            id: '4.1.3',
+            content: 'Item 4.1.3',
+            items: [
+              {
+                id: '4.1.3.1',
+                content: 'Item 4.1.3.1',
+              },
+              {
+                id: '4.1.3.2',
+                content: 'Item 4.1.3.2',
+              },
+              {
+                id: '4.1.3.3',
+                content: 'Item 4.1.3.3',
+              },
+            ],
+          },
+        ],
+      },
+      {
+        id: '4.2',
+        content: 'Item 4.2',
+      },
+    ],
+  },
+  {
+    id: '5',
+    content: 'Item 5',
+  },
+  {
+    id: '6',
+    content: progressiveStepContent,
+    items: [
+      {
+        id: '6.1',
+        content: progressiveStepItemsContent[0],
+      },
+      {
+        id: '6.2',
+        content: progressiveStepItemsContent[1],
+      },
+      {
+        id: '6.3',
+        content: progressiveStepItemsContent[2],
+      },
+    ],
+  },
+  {
+    id: '7',
+    content: 'Item 7',
+  },
+];
+
+function Actions() {
+  const [pressed, setPressed] = useState(false);
+
+  return (
+    <ButtonGroup
+      variant="icon"
+      items={[
+        {
+          id: 'settings',
+          iconName: 'settings',
+          type: 'icon-button',
+          text: 'Settings',
+        },
+        {
+          id: 'menu',
+          type: 'menu-dropdown',
+          text: 'Menu',
+          items: [
+            { id: 'start', text: 'Start' },
+            { id: 'stop', text: 'Stop', disabled: true },
+            {
+              id: 'hibernate',
+              text: 'Hibernate',
+              disabled: true,
+            },
+            { id: 'reboot', text: 'Reboot', disabled: true },
+            { id: 'terminate', text: 'Terminate' },
+          ],
+        },
+        {
+          type: 'icon-toggle-button',
+          id: 'favorite',
+          text: 'Favorite',
+          pressed: pressed,
+          iconName: 'star',
+          pressedIconName: 'star-filled',
+        },
+      ]}
+      onItemClick={({ detail }) => {
+        if (detail.id === 'favorite') {
+          setPressed(!pressed);
+        }
+      }}
+    />
+  );
+}
+
+function RdsAccessRoleTreeItemContent() {
+  return (
+    <div>
+      <div style={{ display: 'flex', flexDirection: 'row', justifyContent: 'space-between', alignItems: 'baseline' }}>
+        <div style={{ display: 'flex', flexDirection: 'row', gap: '4px' }}>
+          <Icon name="user-profile-active" />
+          <span>RdsAccessRole</span>
+          <Badge color="red">1</Badge>
+          <Badge color="blue">1</Badge>
+          <Badge color="green">3</Badge>
+        </div>
+
+        <div>
+          <Actions />
+        </div>
+      </div>
+
+      <SpaceBetween size="s" direction="horizontal">
+        <Box color="text-status-inactive">
+          <SpaceBetween size="xxs" direction="horizontal">
+            <Icon name="globe" />
+            us-east-1
+          </SpaceBetween>
+        </Box>
+        <Box color="text-status-inactive">
+          <SpaceBetween size="xxs" direction="horizontal">
+            <Icon name="ticket" />
+            prod-eng-xyz-zip-f
+          </SpaceBetween>
+        </Box>
+      </SpaceBetween>
+    </div>
+  );
+}
+
+export default function BasicTreeview() {
+  const [expandedItems, setExpandedItems] = useState<Array<string>>(['1', '4.1']);
+
+  return (
+    <Box padding="xl">
+      <Header variant="h2">Basic treeview</Header>
+
+      <div style={{ width: '60%' }}>
+        <Container>
+          <Treeview
+            items={items}
+            onExpandableItemToggle={({ detail }: any) => {
+              if (detail.expanded) {
+                return setExpandedItems(prev => [...prev, detail.id]);
+              } else {
+                return setExpandedItems(prev => prev.filter(id => id !== detail.id));
+              }
+            }}
+            expandedItems={expandedItems}
+          />
+        </Container>
+      </div>
+
+      <div style={{ marginTop: '10px' }}>Expanded items: {expandedItems.map(id => `Item ${id}`).join(', ')}</div>
+    </Box>
+  );
+}
--- a/src/__tests__/snapshot-tests/__snapshots__/documenter.test.ts.snap
+++ b/src/__tests__/snapshot-tests/__snapshots__/documenter.test.ts.snap
@@ -20399,6 +20399,96 @@ exports[`Documenter definition for tree-view matches the snapshot: tree-view 1`]
 }
 `;
 
+exports[`Documenter definition for treeview matches the snapshot: treeview 1`] = `
+{
+  "dashCaseName": "treeview",
+  "events": [
+    {
+      "cancelable": false,
+      "description": "Called when an item's expand toggle is clicked.",
+      "detailInlineType": {
+        "name": "TreeviewProps.ExpandableItemToggleDetail",
+        "properties": [
+          {
+            "name": "expanded",
+            "optional": false,
+            "type": "boolean",
+          },
+          {
+            "name": "id",
+            "optional": false,
+            "type": "string",
+          },
+        ],
+        "type": "object",
+      },
+      "detailType": "TreeviewProps.ExpandableItemToggleDetail",
+      "name": "onExpandableItemToggle",
+    },
+  ],
+  "functions": [],
+  "name": "Treeview",
+  "properties": [
+    {
+      "description": "Provides an \`aria-label\` to the treeview that screen readers can read (for accessibility).
+If there's a visible label element that you can reference, use this instead of \`ariaLabel\`.
+Don't use \`ariaLabel\` and \`ariaLabelledby\` at the same time.",
+      "name": "ariaLabel",
+      "optional": true,
+      "type": "string",
+    },
+    {
+      "description": "Sets the \`aria-labelledby\` property on the treeview.",
+      "name": "ariaLabelledby",
+      "optional": true,
+      "type": "string",
+    },
+    {
+      "deprecatedTag": "Custom CSS is not supported. For testing and other use cases, use [data attributes](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes).",
+      "description": "Adds the specified classes to the root element of the component.",
+      "name": "className",
+      "optional": true,
+      "type": "string",
+    },
+    {
+      "description": "An array of expanded item IDs. Each item ID must match the \`id\` property of an item in the \`items\` array.",
+      "name": "expandedItems",
+      "optional": true,
+      "type": "ReadonlyArray<string>",
+    },
+    {
+      "deprecatedTag": "The usage of the \`id\` attribute is reserved for internal use cases. For testing and other use cases,
+use [data attributes](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes). If you must
+use the \`id\` attribute, consider setting it on a parent element instead.",
+      "description": "Adds the specified ID to the root element of the component.",
+      "name": "id",
+      "optional": true,
+      "type": "string",
+    },
+    {
+      "description": "An array of treeview items.
+Each item has the following properties:
+* \`id\` (string) - The unique identifier of the item.
+* \`content\` (React.ReactNode) - The content of the item.
+* \`details\` (optional, React.ReactNode) - The details of the item, displayed below the content.
+* \`actions\` (optional, React.ReactNode) - Actions related to item. We recommend using button group.
+* \`items\` (optional, TreeItem[]) - The nested items of the item.",
+      "name": "items",
+      "optional": false,
+      "type": "ReadonlyArray<TreeviewProps.TreeItem>",
+    },
+    {
+      "description": "If \`true\`, adds guide lines connecting child items to the expanded parent item.",
+      "name": "showGuideLines",
+      "optional": true,
+      "type": "boolean",
+    },
+  ],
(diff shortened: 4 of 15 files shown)


src/internal/components/expand-toggle-button/index.tsx
cansuaa, line 1: Previously this component was inside table folder as it was only being used in expandable rows feature. Moved the component to internal folder to use it in both table and tree-view.

src/internal/components/structured-item/__tests__/structured-item.test.tsx
```diff
@@ -0,0 +1,13 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
cansuaa: Per PR description, structured item changes can be skipped. They'll be added to the repo by the upcoming List component. I just added it here to use it inside the tree-view.

src/tree-view/interfaces.ts
cansuaa, line 1: The descriptions will go through content review.

src/tree-view/internal.tsx
```diff
@@ -0,0 +1,89 @@
...
+        setInternalExpandedItems([...internalExpandedItems, id]);
+      } else {
+        setInternalExpandedItems(internalExpandedItems.filter(expandedId => expandedId !== id));
+      }
+    }
+
+    if (onItemToggle) {
+      fireNonCancelableEvent(onItemToggle, { id, item, expanded });
+    }
+  };
```
cansuaa: If customer doesn't pass expandedItems, the component will control its expand/collapse state.

src/tree-view/tree-item/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+    expandedItems,
+    renderItem,
+    getItemId,
+    getItemChildren,
+  });
+  const nextLevel = level + 1;
+  let customIcon = null;
+
+  if (renderItemToggleIcon) {
+    customIcon = renderItemToggleIcon(isExpanded);
```
cansuaa: I'll add a integ/screenshot test for this so don't mind codecov.

pages/tree-view/basic.page.tsx
```diff
@@ -0,0 +1,411 @@
...
+          <Badge color="green">3</Badge>
+        </div>
+
+        <div>
+          <Actions actionType="button-group" />
+        </div>
+      </div>
+
+      <div style={{ display: 'flex', gap: '10px' }}>
+        <Box color="text-status-inactive">
```
pan-kot: nit: there is no margin between the icon and text in this component

pages/tree-view/basic.page.tsx
```diff
@@ -0,0 +1,411 @@
...
+        </Box>
+      </div>
+    </div>
+  );
+}
+
+export default function BasicTreeView() {
+  const [expandedItems, setExpandedItems] = useState<Array<string>>(['1', '4.1']);
+  const [useDifferentIcon, setUseDifferentIcon] = useState(false);
+  const [useDifferentIconWithAnimation, setUseDifferentIconWithAnimation] = useState(false);
```
pan-kot: What is the reason of not using animation? We add animation unconditionally to other components, and it should be controlled globally with reduced-motion settings.

pages/tree-view/basic.page.tsx
```diff
@@ -0,0 +1,411 @@
...
+          prod-eng-xyz-zip-f
+        </Box>
+      </div>
+    </div>
+  );
+}
+
+export default function BasicTreeView() {
+  const [expandedItems, setExpandedItems] = useState<Array<string>>(['1', '4.1']);
+  const [useDifferentIcon, setUseDifferentIcon] = useState(false);
```
pan-kot: nit: I would call this "useCustomIcon" to emphasise the alternative is the component's default.

pages/tree-view/basic.page.tsx
```diff
@@ -0,0 +1,411 @@
...
+      <h1>Basic tree view</h1>
+
+      <Checkbox checked={useDifferentIcon} onChange={({ detail }) => setUseDifferentIcon(detail.checked)}>
+        Use different CDS icon
+      </Checkbox>
+      <Checkbox
+        checked={useDifferentIconWithAnimation}
+        onChange={({ detail }) => setUseDifferentIconWithAnimation(detail.checked)}
+      >
+        Use icon with animation
```
pan-kot: nit: the state seems to affect the custom icon only, but that is not reflected in the label

pages/tree-view/basic.page.tsx
```diff
@@ -0,0 +1,411 @@
...
+      </Checkbox>
+
+      <div style={{ width: '60%' }}>
```
pan-kot: Why do we use width 60%? This does not work well in a small screen, and can affect the screenshot tests for this page. I suggest to use responsive styles, or e.g. our Grid component for that instead.

pan-kot:
I wonder if actions shouldn't go to the next line in a constrained layout - that does not look optimal to me. I would rather expect actions to take a smaller portion of the item size maybe, but stay on the right.

Screenshot 2025-06-12 at 16 45 45

pages/tree-view/common.tsx
```diff
@@ -0,0 +1,61 @@
...
+        <StatusIndicator type="warning">{item.warningCount}</StatusIndicator>
+      )}
+      {item.successCount && item.successCount > 0 && (
+        <StatusIndicator type="success">{item.successCount}</StatusIndicator>
+      )}
+    </div>
+  );
+}
+
+export function Actions() {
```
pan-kot: nit: there is Actions component in the base tree view page, too. Would it make sense to move all helper component definitions from base page to here?

pages/tree-view/dynamic-items.page.tsx
```diff
@@ -0,0 +1,81 @@
...
+import { allItems, items } from './generate-data';
+
+const allExpandableItemIds = allItems.filter(item => item.children && item.children.length > 0).map(item => item.id);
+
+export default function TestPage() {
+  const [expandedItems, setExpandedItems] = useState<Array<string>>([]);
+
+  return (
+    <>
+      <h1>Test performance page</h1>
```
pan-kot: nit: the file name for this page is "dynamic-items". Let's use a consistent file and header name for pages

pages/tree-view/dynamic-items.page.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+      <SpaceBetween size="s">
+        <Button
+          onClick={() => {
+            setExpandedItems(allExpandableItemIds);
+            console.time('expand-all');
+            requestAnimationFrame(() => console.timeEnd('expand-all'));
```
pan-kot, line 27: What the page does not log - is the time it takes to expand/collapse a single item. When all tree is expanded - a simple operation like that takes the same amount of time as expanding the whole tree.

src/internal/components/expand-toggle-button/index.tsx
```diff
@@ -14,11 +14,15 @@ export function ExpandToggleButton({
   onExpandableItemToggle,
   expandButtonLabel,
   collapseButtonLabel,
+  customIcon,
+  hasLargeFocusOffset,
```
pan-kot: Why do we need this setting? - can we enlarge the offset for all toggle buttons instead, unconditionally?

src/test-utils/dom/tree-view/index.ts
```diff
@@ -0,0 +1,106 @@
...
+  findItemToggle(): ElementWrapper | null {
+    return this.findByClassName(expandToggleStyles['expand-toggle']);
+  }
+
+  /**
+   * Finds all visible child items of the tree item.
+   * @param options
+   * * expanded (boolean) - Flag to find the expanded/collapsed items
+   */
+  findChildren(options: { expanded?: boolean } = {}): Array<TreeItemWrapper> {
```
pan-kot: nit: would it be better clear if we call the methods: `findChildItems` and `findChildItemById`? These will be better relatable to the `findItems` and `findItemById`.

src/tree-view/__tests__/tree-item.test.tsx
```diff
@@ -0,0 +1,115 @@
...
+const defaultProps: TreeViewProps<Item> = {
+  items: defaultItems,
+  getItemId: item => item.id,
+  getItemChildren: item => item.items,
+  renderItem: item => ({
+    content: item.title,
+  }),
+};
+
+function renderTreeView(props: Partial<TreeViewProps<Item>> = {}) {
```
pan-kot, line 10: nit: there are common functions and default props across tree view test files - would it make sense to move those to some common.ts?

src/tree-view/__tests__/tree-view.test.tsx
```diff
@@ -0,0 +1,203 @@
...
+        expect(item.findSecondaryContent()!.getElement()).toHaveTextContent(`Description ${index + 1}`);
+      }
+    }
+    if (data.hasActions) {
+      expect(item.findActions()!.findButtonDropdown()!.getElement()).toBeVisible();
+    }
+  });
+});
+
+test('should render with other CDS components inside', () => {
```
pan-kot: nit: I would remove the "CDS" from the test description. First of all, it might be unclear. Second, it is not really needed.

src/tree-view/__tests__/tree-view.test.tsx
```diff
@@ -0,0 +1,203 @@
...
+
+  const contentPopover = item.findContent()!.findPopover()!;
+  const descriptionLink = item.findSecondaryContent()!.findLink()!;
+  const actionsButtonDropdown = item.findActions()!.findButtonDropdown()!;
+
+  expect(contentPopover.getElement()).toBeVisible();
+  expect(descriptionLink.getElement()).toBeVisible();
+  expect(actionsButtonDropdown.getElement()).toBeVisible();
+
+  fireEvent.click(contentPopover.findTrigger().getElement());
```
pan-kot: Why do we use fireEvent? Does `contentPopover.findTrigger().click()` also work here?

src/tree-view/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+import InternalTreeView from './internal';
+
+export { TreeViewProps };
+
+const TreeView = <T,>({ items, ...rest }: TreeViewProps<T>) => {
+  const baseProps = getBaseProps(rest);
+  const baseComponentProps = useBaseComponent('TreeView', {
+    props: {},
+    metadata: {
+      itemsCount: items?.length,
```
pan-kot: nit: the items array is always defined as per TS declaration, so no need to use optional chaining here. Also, what do we need the items count for? This will likely change during the component lifetime, so can be zero at first, and then change to non-zero, after data is loaded.

src/tree-view/interfaces.ts
```diff
@@ -0,0 +1,94 @@
...
+export interface TreeViewProps<T = any> extends BaseComponentProps {
+  /**
+   * An array of tree view items.
+   */
+  items: ReadonlyArray<T>;
+
+  /**
+   * Use it to map your data to render the item.
+   * For each item the below properties must be returned:
+   * * `content` (React.ReactNode) - The content of the item.
```
pan-kot: nit: in other similar API docs we use ReactNode instead of React.ReactNode.

src/tree-view/interfaces.ts
```diff
@@ -0,0 +1,94 @@
...
+   * Sets the `aria-labelledby` property on the tree view.
+   * If there's a visible label element that you can reference, use this instead of `ariaLabel`.
+   * Don't use `ariaLabel` and `ariaLabelledby` at the same time.
+   */
+  ariaLabelledby?: string;
+
+  /**
+   * Sets the `aria-describedby` property on the tree view.
+   */
+  ariaDescribedby?: string;
```
pan-kot, line 55: Should we add ariaDescription, too? We add it to the newer components, e.g. Slider.

src/tree-view/interfaces.ts
```diff
@@ -0,0 +1,94 @@
...
+   * @i18n
+   */
+  i18nStrings?: TreeViewProps.I18nStrings<T>;
+
+  /**
+   * @awsuiSystem core
+   * Overrides the default expand toggle.
+   * Use it to have different icons/animations for toggle.
+   */
+  renderItemToggleIcon?: (isExpanded: boolean) => React.ReactNode;
```
pan-kot: I would make the arguments an object, even though as of today we only support the single isExpanded arg. We can later extend this with e.g. the level, or selection state etc.

src/tree-view/internal.tsx
```diff
@@ -0,0 +1,89 @@
...
+import InternalTreeItem from './tree-item';
+
+import styles from './styles.css.js';
+import testUtilStyles from './test-classes/styles.css.js';
+
+type InternalTreeViewProps<T> = TreeViewProps<T> & InternalBaseComponentProps;
+
+const InternalTreeView = <T,>({
+  expandedItems,
+  items = [],
```
pan-kot: We should not assign items to an empty array here because items prop is required. If it fails in tests - we should modify the default component props setting instead.

Also, all property defaults if we need some must be set in the index file in order to be correctly parsed by the documenter.

src/tree-view/internal.tsx
```diff
@@ -0,0 +1,89 @@
...
+  i18nStrings,
+  __internalRootRef,
+  ...rest
+}: InternalTreeViewProps<T>) => {
+  const baseProps = getBaseProps(rest);
+  const isExpandStateControlled = !!expandedItems;
+  const [internalExpandedItems, setInternalExpandedItems] = useState(expandedItems || []);
+
+  if (isExpandStateControlled && !onItemToggle) {
+    warnOnce(
```
pan-kot: Why do we use a custom warning instead of using the `useControllable` util?

src/tree-view/styles.scss
```diff
@@ -0,0 +1,49 @@
...
+
+.tree,
+.treeitem-group,
+.treeitem {
+  @include styles.styles-reset-ul;
+}
+
+.treeitem {
+  display: grid;
+  grid-template-columns: 28px 1fr;
```
pan-kot: nit: let's move the values like 28px or 32px to the named vars in the file's top, for readability and better maintainability, should we use those in multiple places, which we might as well need once we start adding support for connector lines.

src/tree-view/tree-item/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+  const i18n = useInternalI18n('tree-view');
+  const { id, isExpandable, isExpanded, children, icon, content, secondaryContent, actions } = transformTreeItemProps({
+    item,
+    index,
+    expandedItems,
+    renderItem,
+    getItemId,
+    getItemChildren,
+  });
+  const nextLevel = level + 1;
```
pan-kot, line 45: do we have a max level?

src/tree-view/tree-item/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+    <li
+      id={id}
+      className={clsx(
+        styles.treeitem,
+        testUtilStyles.treeitem,
+        isExpandable && [testUtilStyles.expandable],
+        isExpanded && [testUtilStyles.expanded]
+      )}
+      aria-expanded={isExpandable ? isExpanded : undefined}
+      aria-level={level > 0 ? level : undefined}
```
pan-kot: I would add a comment, explaining why aria-level is not added to the root level.

src/tree-view/tree-item/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+      data-testid={`treeitem-${id}`}
```
pan-kot: Why do we prefix the testid as such? In other components I see that we either don't use a prefix, or have some custom prefix that starts from "awsui-".

src/tree-view/tree-item/utils.ts
```diff
@@ -0,0 +1,33 @@
...
+
+import { TreeViewProps } from '../interfaces';
+
+interface TransformTreeItemPropsParams
+  extends Pick<TreeViewProps, 'expandedItems' | 'renderItem' | 'getItemId' | 'getItemChildren'> {
+  item: any;
+  index: number;
+}
+
+export function transformTreeItemProps({
```
pan-kot: nit: this function is only used once, inside the tree item - wouldn't it be easier to have the code inline?

before:
```
const { id, isExpandable, isExpanded, children, icon, content, secondaryContent, actions } = transformTreeItemProps({
    item,
    index,
    expandedItems,
    renderItem,
    getItemId,
    getItemChildren,
  });
```

after:
```
const { icon, content, secondaryContent, actions } = renderItem(item, index);
const id = getItemId(item, index);
const children = getItemChildren(item, index) || [];
const isExpandable = children.length > 0;
const isExpanded = isExpandable && expandedItems.includes(itemId);
```

pages/tree-view/basic.page.tsx
cansuaa: This would be up to customer's choice of whether they want to have animation on the toggle or not. In the default toggle icon, there is animation.

pages/tree-view/dynamic-items.page.tsx
cansuaa: Oh I forgot to update the header, the file was named test as well before, will update!

src/internal/components/expand-toggle-button/index.tsx
cansuaa: In the table as well? Then why was the original offset design token was chosen over the one I am using for treeview?

src/tree-view/__tests__/tree-view.test.tsx
cansuaa: Yes indeed I wrote it like this initially, changed in other places but must have forgotten this one.

src/tree-view/interfaces.ts
cansuaa: I am a bit confused about that, prompt input is also new but that one doesn't have ariaDescription and most components don't. The implementation part was also manual

src/tree-view/internal.tsx
cansuaa: Yes you are right I'll remove it

src/tree-view/internal.tsx
cansuaa: Oh I wasn't aware of this util thanks for bringing it, let me take a look and update

src/tree-view/tree-item/index.tsx
cansuaa: Not at the moment

src/tree-view/tree-item/index.tsx
cansuaa: I thought it was more readable this way

src/tree-view/tree-item/utils.ts
cansuaa: Yes it'd be easier for me but again I thought it was more readable this way. Unless you are interested in how those values are obtained, you don't need to read through the logic

pages/tree-view/basic.page.tsx
```diff
@@ -0,0 +1,324 @@
...
+        </Box>
+      </SpaceBetween>
+    </div>
+  );
+}
+
+export default function BasicTreeView() {
+  const [expandedItems, setExpandedItems] = useState<Array<string>>(['1', '4.1']);
+  const [useCustomIcon, setUseCustomIcon] = useState(false);
+  const [useCaretIconWithSlowerAnimation, setUseCaretIconWithSlowerAnimation] = useState(false);
```
pan-kot: nit: The slower animation is only available for custom icons. Would it make sense to represent this differently, to avoid confusion? For example, instead of using two boolean flags, you can use a select with states: "default-icon", "custom-icon", "custom-icon-with-slow-animation".

pages/tree-view/common.tsx
```diff
@@ -0,0 +1,120 @@
...
+    </div>
+  );
+}
+
+export function Actions(
+  { actionType }: { actionType?: 'button-group' | 'button-dropdown' | 'inline-button-dropdown' } = {
+    actionType: 'button-dropdown',
+  }
+) {
+  const [pressed, setPressed] = useState(false);
```
pan-kot: nit: the name "pressed" does not reflect the change in state that well (because there are multiple actions), would it make sense to rename to e.g. "markedAsFavourite"?

pages/tree-view/basic.page.tsx
pan-kot: That will better represent possible states. E.g. now setting both flags is the same as setting the first flag only.

pages/tree-view/basic.page.tsx
```diff
@@ -0,0 +1,324 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
pan-kot, line 1: I would add sample content with lots of text so that we can see how the responsiveness works, including content and description slots. It looks like there are some issues with responsiveness now: the actions sometimes go to the next line, and then render in-between content and description, which is strange. In item with badges, the actions don't go to the next line, but instead become narrow.

Screenshot 2025-06-23 at 16 52 46

src/tree-view/internal.tsx
```diff
@@ -0,0 +1,80 @@
...
+      setExpandedItems([...(expandedItems || []), id]);
+    } else {
+      setExpandedItems((expandedItems || []).filter(expandedId => expandedId !== id));
+    }
+    fireNonCancelableEvent(onItemToggle, { id, item, expanded });
+  };
+
+  return (
+    <div {...baseProps} ref={__internalRootRef} className={clsx(baseProps.className, styles.root, testUtilStyles.root)}>
+      <ul
```
pan-kot, line 53: Should we use role "tree"?

Screenshot 2025-06-23 at 16 54 47


https://www.w3.org/WAI/ARIA/apg/patterns/treeview/examples/treeview-1b/:
Screenshot 2025-06-23 at 16 55 04

pages/tree-view/basic.page.tsx
cansuaa: I will add long texts, that is a good point. The item with the badge is a custom implementation and only uses the content slot so it may act differently. I am changing its title to indicate it is a custom one.

The wrapping behaviour where actions go to the next line between content and description is intended, that's how the design is.

src/tree-view/internal.tsx
cansuaa: The "tree" and "treeitem" roles will not be added in the initial release, they will be added as a follow up together with focus management. Adding a comment to the code.

src/internal/components/structured-item/interfaces.ts
```diff
@@ -9,4 +9,5 @@ export interface StructuredItemProps {
   actions?: ReactNode;
   secondaryContent?: ReactNode;
   disablePaddings?: boolean;
+  disableActionsWrapping?: boolean;
```
pan-kot: nit: should we call it `wrapActions` - similar to table's `wrapLines` (and also similar flex-wrap)? It can be true or false by default.

src/tree-view/__tests__/a11y.test.tsx
```diff
@@ -0,0 +1,107 @@
...
+test('i18nStrings adds custom label to expand/collapse toggle', () => {
+  const { wrapper } = renderTreeViewWithI18NProvider({
+    i18nStrings: {
+      expandButtonLabel: item => `Expand ${item.title}`,
+      collapseButtonLabel: item => `Collapse ${item.title}`,
+    },
+  });
+
+  const collapsedItem = wrapper.findItemById('1', { expanded: false })!;
+  expect(collapsedItem.findItemToggle()!.getElement()).toHaveAccessibleName('Expand Item 1 Item 1'); // i18n string + item content
```
pan-kot: Why does this include content twice? I thought the label is rendered by i18n - do we also append content?

src/tree-view/interfaces.ts
```diff
@@ -0,0 +1,97 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import React from 'react';
+
+import { BaseComponentProps } from '../internal/base-component';
+import { NonCancelableEventHandler } from '../internal/events';
+
+export interface TreeViewProps<T = any> extends BaseComponentProps {
+  /**
+   * Specifies the items to display in the tree view.
```
pan-kot: Should we say "top-level items", and explain how to define children?

src/tree-view/interfaces.ts
pan-kot: Oh, I see - let's ignore it then.

src/internal/components/expand-toggle-button/index.tsx
pan-kot: I am not sure if it was a deliberate decision. I think if a larger outline offset does look ok in the table, too - we don't need to create an extra prop. It also makes sense that both toggle components look similar.

src/tree-view/tree-item/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+      <div className={styles['expand-toggle-wrapper']}>
+        {isExpandable && (
+          <div className={styles.toggle}>
+            <ExpandToggleButton
+              isExpanded={isExpanded}
+              customIcon={customIcon}
+              hasLargeFocusOffset={true}
+              expandButtonLabel={joinStrings(
+                i18n('i18nStrings.expandButtonLabel', i18nStrings?.expandButtonLabel?.(item)),
+                announcementLabel ?? (content as string)
```
pan-kot: Why do we pass content if announcement label is not provided? I see two problems with that:
1. If content is not a string - this can be stringified as [Object object]
2. This conflicts with the button label in case it includes the content name as well: e.g. if you want it to say "Expand {name} element" - we don't need the content to be appended to that.

src/tree-view/tree-item/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+        )}
+      </div>
+
+      <div className={styles['structured-item-wrapper']}>
+        <InternalStructuredItem
+          icon={icon}
+          content={content}
+          secondaryContent={secondaryContent}
+          actions={actions}
+          disableActionsWrapping={true}
```
pan-kot: Can we do so for the List as well?

src/internal/components/structured-item/interfaces.ts
cansuaa: I also thought about that but then I also saw `disableActionsWrapping` used in other places. When I used the wrapActions I setted it to true by default but it is kinda weird that there is this API and it's already enabled by default, no?

src/tree-view/__tests__/a11y.test.tsx
cansuaa: The i18n value of expandButtonLabel is `Expand` and we append the announcementLabel or content to it so it announces what item it will expand. In this test I override the i18n value to also include the item title so that is why the item title shows twice in the label.

src/tree-view/interfaces.ts
cansuaa: I can update it but I don't think we need to mention children explicitly here when `getItemChildren` API is defined below

src/tree-view/tree-item/index.tsx
cansuaa: In the API of announcementLabel we say "Make sure to provide the `announcementLabel` if `content` is not a string."

src/tree-view/tree-item/index.tsx
cansuaa: That should be a separate discussion and PR. If it's decided to not wrap there as well we can simply remove this API, structured item is not a public component anyway.

src/internal/components/structured-item/interfaces.ts
cansuaa: Updated to `wrapActions`

src/tree-view/__tests__/a11y.test.tsx
pan-kot: But that only works when the item content is a string, isn't it? Say, the content is appended and the component is accessible. Then, the consumer changes the content from `"Item 1"` to `<>Item 1</>`, which is legal as per API - now, the component is no longer accessible.

cansuaa: Exactly, that is why `announcementLabel` API is added. It is meant to be used when content is/can be something other than a string. The same logic is also applied in the List component.

src/tree-view/interfaces.ts
pan-kot: We do sometimes mention related properties in the API docs, e.g. we can say "Specifies items to display as the top-level in the tree view. Use `getItemChildren` to provide the nested items."

Otherwise, one might think the `items` refers to all items.

src/tree-view/tree-item/index.tsx
pan-kot: Should we then then do: `typeof content === 'string' ? content : ''`?

src/tree-view/interfaces.ts
cansuaa: Hmm, good point. Though `getItemChildren` is required so they must provide it, mentioning it in the `items` description could help understanding how the API works, will update.

src/tree-view/tree-item/index.tsx
cansuaa: Sure, updated and added an extra test.

pan-kot (approved):

PR #3482, opened 2025-05-07
chore: Finalize drawer animation and visual design

at-susie:
### Description

This PR: 
- applies the latest visual design decision including gap background color and new icons
- fixes width animation issues when a global drawer enters a view

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/app-layout/visual-refresh-toolbar/drawer/styles.scss
```diff
@@ -31,7 +31,6 @@ $global-drawer-expanded-mode: #{awsui.$motion-duration-refresh-only-slow} #{awsu
   inline-size: var(#{custom-props.$drawerSize});
 
   block-size: 100%;
-  overflow: hidden;
```
georgylobko, line 34: What is the purpose of this change? Did you try to run screenshot tests on your dev pipeline?

at-susie: To address the motion issue that we discussed in the review meeting on 5/7, the solution I found is to set `min-inline-size` to a child element. I did it on [this code ](https://github.com/cloudscape-design/components/pull/3482/files#diff-4eb6e4407d23a20b1eb481d5fc1d1a654fc6029ed23179b1fcc0294f77427324R109). But then I found that this `overflow: hidden` drives an unwanted behavior. So I removed it and added `overflow: clip` to the `.global-tools`.

at-susie: I have not run screenshot test. But that's a good call.

at-susie: Ok, as we chatted, I brought the `overflow: hidden;` back, and made changes that are only applied global-drawer.

src/app-layout/visual-refresh-toolbar/drawer/global-drawer.tsx
```diff
@@ -117,53 +117,57 @@ function AppLayoutGlobalDrawerImplementation({
...
-                />
-              </div>
-            )}
-            <div
-              className={clsx(styles['drawer-content-container'], sharedStyles['with-motion-horizontal'])}
-              data-testid={`awsui-app-layout-drawer-content-${activeDrawerId}`}
-            >
-              {!isMobile && activeGlobalDrawer?.isExpandable && (
-                <div className={styles['drawer-expanded-mode-button']}>
+            <div className={clsx(styles['global-drawer-wrapper'])}>
```
georgylobko, line 120: Can we achieve the same visual representation without this wrapper?

src/app-layout/visual-refresh-toolbar/drawer/styles.scss
```diff
@@ -140,4 +98,125 @@ $global-drawer-expanded-mode: #{awsui.$motion-duration-refresh-only-slow} #{awsu
     display: flex;
     align-items: center;
   }
+
+  &.drawer-global {
+    display: block;
+
+    > .global-drawer-wrapper {
+      display: grid;
+      min-inline-size: var(#{custom-props.$drawerSize});
```
georgylobko, line 107: `min-inline-size` and `inline-size` have already been set for the parent element. Why do we need to set them twice?

src/app-layout/visual-refresh-toolbar/drawer/styles.scss
```diff
@@ -140,4 +98,125 @@ $global-drawer-expanded-mode: #{awsui.$motion-duration-refresh-only-slow} #{awsu
...
+      inline-size: var(#{custom-props.$drawerSize});
+      grid-template-columns: $global-drawer-gap-size 16px 1fr;
```
georgylobko, line 109: How did you get this 16px? I'd suggest to create a variable to avoid having "magic numbers"

src/app-layout/visual-refresh-toolbar/drawer/styles.scss
```diff
@@ -55,58 +56,14 @@ $global-drawer-expanded-mode: #{awsui.$motion-duration-refresh-only-slow} #{awsu
     }
   }
 
-  &.drawer-global {
```
georgylobko, line 58: I'd suggest to bring back this hierarchy. The less changes - the easier to review them.

src/app-layout/visual-refresh-toolbar/drawer/global-drawer.tsx
at-susie: I explored, but it seems it's pretty tough to achieve it while having expected motion

src/app-layout/visual-refresh-toolbar/drawer/styles.scss
at-susie: I understand, but we cannot achieve the expected visual and motion without changing the hierarchy. 
What are the issues?

georgylobko: Why we cannon achieve this? The `global-drawer` belongs to the root element of this component, so it should stay as before

georgylobko: To avoid misunderstandings: previously the `.drawer-global` selector began at line 58, but now it begins on line 102, which makes the git history cumbersome

src/app-layout/visual-refresh-toolbar/drawer/styles.scss
at-susie: Screenshot 2025-05-13 at 12 01 07
The `inline-size` seems unnecessary, but the `min-inline-size` is necessary, otherwise it has an unwanted width changing behavior when it's entering like the attached image.

src/app-layout/visual-refresh-toolbar/drawer/global-drawer.tsx
at-susie: As per our conversation, it's fine to have this

src/app-layout/visual-refresh-toolbar/drawer/styles.scss
```diff
@@ -140,4 +98,124 @@ $global-drawer-expanded-mode: #{awsui.$motion-duration-refresh-only-slow} #{awsu
     display: flex;
     align-items: center;
   }
+
+  &.drawer-global {
```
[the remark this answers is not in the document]
georgylobko: That should not be a case. These style were there before, eslint should be happy with them

src/app-layout/visual-refresh-toolbar/drawer/styles.scss
```diff
@@ -87,14 +82,35 @@ $global-drawer-expanded-mode: #{awsui.$motion-duration-refresh-only-slow} #{awsu
...
+            background: awsui.$color-gap-global-drawer;
+            border-inline-start: awsui.$border-divider-section-width solid awsui.$color-border-layout;
+            box-sizing: border-box;
+          }
+        }
+      }
+
+      > .global-drawer-wrapper {
+        inline-size: 100%;
+        grid-template-columns: 8px 1fr;
```
georgylobko, line 101: Can we replace this magic number with a token?

src/app-layout/visual-refresh-toolbar/drawer/styles.scss
```diff
@@ -140,4 +157,65 @@ $global-drawer-expanded-mode: #{awsui.$motion-duration-refresh-only-slow} #{awsu
     display: flex;
     align-items: center;
   }
+
+  > .global-drawer-wrapper {
+    display: grid;
+    min-inline-size: var(#{custom-props.$drawerSize});
```
georgylobko, line 163: The question still stands: do we really need to set the inline size twice? Based on our earlier discussion, we agreed that it’s not necessary

src/app-layout/visual-refresh-toolbar/drawer/styles.scss
at-susie: Thank you for the callout. Updated and pushed the fix

src/app-layout/visual-refresh-toolbar/drawer/styles.scss
at-susie: While `inline-size` is not necessary, which I removed, the `min-inline-size` is still necessary otherwise the unwanted width change happens.

Uploading Screen Recording 2025-05-13 at 15.54.49.mov…

georgylobko: Ok, should we remove `min-inline-size` from the parent selector then?

georgylobko (approved):

PR #3491, opened 2025-05-12
chore: Updating Granularity type to that of CalendarProps

dpitcock:
### Description

Currently the DatePicker imports its Granularity from CalendarProps, so this will ensure the types are the same between DatePicker and DateRangePacker and the one source of truth for the Granularity type lives in the CalendarProps

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/__tests__/snapshot-tests/__snapshots__/documenter.test.ts.snap
+++ b/src/__tests__/snapshot-tests/__snapshots__/documenter.test.ts.snap
@@ -7979,10 +7979,10 @@ Default: the user's current time offset as provided by the browser.",
     },
     {
       "defaultValue": "'day'",
-      "description": "Specifies the granularity at which users will be able to select a date range.
+      "description": "Specifies the granularity at which users will be able to select a date.
 Defaults to \`day\`.",
       "inlineType": {
-        "name": "DateRangePickerProps.Granularity",
+        "name": "CalendarProps.Granularity",
         "type": "union",
         "values": [
           "day",
--- a/src/date-range-picker/calendar/grids/interfaces.ts
+++ b/src/date-range-picker/calendar/grids/interfaces.ts
@@ -1,5 +1,6 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
+import { CalendarProps } from '../../../calendar/interfaces';
 import { DateRangePickerProps, DayIndex } from '../../interfaces';
 
 export interface GridBaseProps {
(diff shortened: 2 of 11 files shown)


src/date-range-picker/calendar/grids/interfaces.ts
```diff
@@ -33,10 +34,10 @@ export interface GridProps extends GridBaseProps {
   currentMonthAriaLabel?: string;
   startOfWeek?: DayIndex;
   todayAriaLabel?: string;
-  granularity: DateRangePickerProps.Granularity;
+  granularity: CalendarProps.Granularity;
```
dpitcock, line 37: Added here rather this way because the granularity is required compared to optional in CalendarProps.

src/__tests__/snapshot-tests/__snapshots__/documenter.test.ts.snap
```diff
@@ -7979,10 +7979,10 @@ Default: the user's current time offset as provided by the browser.",
     },
     {
       "defaultValue": "'day'",
-      "description": "Specifies the granularity at which users will be able to select a date range.
+      "description": "Specifies the granularity at which users will be able to select a date.
```
dpitcock, line 7982: Note to self, need to confirm with design this line is ok.

src/date-range-picker/calendar/header/header-button.tsx
```diff
@@ -4,14 +4,13 @@ import React from 'react';
 import clsx from 'clsx';
 
 import { InternalButton } from '../../../button/internal';
-import { DateRangePickerProps } from '../../interfaces';
+import { CalendarProps } from '../../../calendar/interfaces';
 
 import testutilStyles from '../../test-classes/styles.css.js';
 
-interface HeaderButtonProps {
+interface HeaderButtonProps extends Pick<CalendarProps, 'granularity'> {
```
jperals, line 11: Actually `granularity` does not seem to be used anywhere for this interface?

src/date-range-picker/calendar/grids/interfaces.ts
jperals: I think you could do
```
export interface GridBaseProps extends Required<Pick<CalendarProps, 'granularity'>>
```

src/date-range-picker/calendar/header/header-button.tsx
dpitcock: good callout. It was used before to apply a classname that had either next-month, or next-year but it was only for the test util, so it was redundant, and next-page covers it. Removing

jperals (approved):

PR #3501, opened 2025-05-13
fix: Do not generate test utils for non-public components

connorlanigan:
### Description

Before this change, our build process would generate a `find` function for every folder/component in the `src/test-utils/dom` folder, even if that corresponding component is not public (such as the Annotation component). This can lead customers to use a test util function that was not expected to be exposed directly (i.e. `createWrapper().findAnnotation()`).

With this change, we only generate a `find` function if the component is public.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/build-tools/tasks/test-utils.js
+++ b/build-tools/tasks/test-utils.js
@@ -22,16 +22,22 @@ function compileTypescript(theme) {
 
 function generateTestUtilsForComponents(signalCompletion) {
   const componentNamesKebabCase = listPublicItems(path.join(testUtilsSrcDir, 'dom'));
+  const publicComponents = listPublicItems('src');
 
-  const components = componentNamesKebabCase.map(testUtilsFolderName => {
-    const name = pascalCase(testUtilsFolderName);
-    const pluralName = pluralizeComponentName(name);
-    return {
-      name,
-      pluralName,
-      testUtilsFolderName,
-    };
-  });
+  const isPublicComponent = name => publicComponents.includes(name);
+
+  const components = componentNamesKebabCase
+    .filter(component => isPublicComponent(component) || component === 'annotation')
+    // The test utils for the internal Annotation component have already been exposed. We need to keep them for backwards compatibility.
+    .map(testUtilsFolderName => {
+      const name = pascalCase(testUtilsFolderName);
+      const pluralName = pluralizeComponentName(name);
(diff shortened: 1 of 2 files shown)


build-tools/tasks/test-utils.js
```diff
@@ -22,16 +22,22 @@ function compileTypescript(theme) {
...
-    return {
-      name,
-      pluralName,
-      testUtilsFolderName,
-    };
-  });
+  const isPublicComponent = name => publicComponents.includes(name);
+
+  const components = componentNamesKebabCase
+    .filter(component => isPublicComponent(component) || component === 'annotation')
```
just-boris, line 30: How about silently skipping also throw an error that there is a folder in an unexpected location?

Because there are also some teams importing these directories directly assuming "if they are not in `internal`, it must be public"

build-tools/tasks/test-utils.js
```diff
@@ -22,16 +22,22 @@ function compileTypescript(theme) {
...
-  const components = componentNamesKebabCase.map(testUtilsFolderName => {
-    const name = pascalCase(testUtilsFolderName);
-    const pluralName = pluralizeComponentName(name);
-    return {
-      name,
-      pluralName,
-      testUtilsFolderName,
-    };
-  });
+  const isPublicComponent = name => publicComponents.includes(name);
```
pan-kot, line 27: Can we fix that by moving all internal components to under test-utils/internal.

We already have should have snapshot tests for the test utils, so if more components are added to the root folder in the future - the test will show that.

connorlanigan: Currently, there are no internal components in the test-utils folder except the Annotation. This PR thus does not change the test utils that are currently generated, but guards against exposing internal components similarly in the future.

build-tools/tasks/test-utils.js
just-boris: Okay, I found we have a snapshot test: https://github.com/cloudscape-design/components/blob/main/src/__tests__/snapshot-tests/__snapshots__/test-utils-wrappers.test.tsx.snap

So this is already validated if anything unexpected is exposed

just-boris: (feel free to resolve the thread)

PR #3510, opened 2025-05-15
chore: Fix animation issue

at-susie:
### Description

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/app-layout/visual-refresh-toolbar/skeleton/styles.scss
```diff
@@ -158,10 +158,14 @@
   z-index: 0;
 
   &.navigation,
-  &.tools,
-  &.global-tools {
```
georgylobko, line 162: We cannot make such changes as they were previously made on purpose, so this is necessary to keep

src/app-layout/visual-refresh-toolbar/drawer/global-drawer.tsx
```diff
@@ -71,7 +71,7 @@ function AppLayoutGlobalDrawerImplementation({
   const isExpanded = activeGlobalDrawer?.isExpandable && expandedDrawerId === activeDrawerId;
 
   return (
-    <Transition nodeRef={drawerRef} in={show || isExpanded} appear={show || isExpanded} timeout={0}>
+    <Transition nodeRef={drawerRef} in={show || isExpanded} appear={show || isExpanded} timeout={100}>
```
georgylobko, line 74: Why 100?

at-susie: If we set this 0, we cannot make [this exiting transition](https://github.com/cloudscape-design/components/pull/3510/files#diff-4eb6e4407d23a20b1eb481d5fc1d1a654fc6029ed23179b1fcc0294f77427324R67) smooth. 

It's tricky to explain why 100, it's 100ms duration from the moment of getting from `entered` to `exiting` state in the Transition component It is not the duration for closing motion, which I can see a similar behavior to when a `.split-panel-side` is closed.

src/app-layout/visual-refresh-toolbar/skeleton/styles.scss
at-susie: I updated the scope like [this](https://github.com/cloudscape-design/components/pull/3510/files#diff-5bc01b1addbc8b77ec4465eb32ec4b96d5a47c73d76b7e582b06de57a1a38254R167).

Given this will be a separate PR to the main branch and not to yours, is this an acceptable change?

georgylobko: It doesn’t, because this change affects the behavior of all global drawers and could introduce unexpected bugs. The first thing that comes to mind is how it might impact drawers that use preserveInactiveContent.

If you want to animate the drawer closing, I’d recommend looking into how you can achieve that using the Transition component props here: https://reactcommunity.org/react-transition-group/transition

at-susie: Ok thanks, I'll try out Transition component.

at-susie: Ok, Transition component works. I've brought the original styles in the skeleton back. Now there is no change on the `src/app-layout/visual-refresh-toolbar/skeleton/styles.scss`.

src/app-layout/visual-refresh-toolbar/drawer/global-drawer.tsx
```diff
@@ -75,7 +75,7 @@ function AppLayoutGlobalDrawerImplementation({
     (isExpandedPrevious && !isExpanded);
 
   return (
-    <Transition nodeRef={drawerRef} in={show || isExpanded} appear={show || isExpanded} timeout={0}>
+    <Transition nodeRef={drawerRef} in={show || isExpanded} appear={show || isExpanded} timeout={60}>
```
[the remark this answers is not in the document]
at-susie: The `unmountOnExit` doesn't seem to help. As long as we keep `timeout={0}`, the Transition component instantly becomes `exited` state where we cannot have smooth exiting motion.

PR #3511, opened 2025-05-15
feat: Add showButton property to InternalDragHandle

johannes-weber:
Use-case: This will be consumed in board components where we need to render the arrow icons under some conditions on the initial render.

### Description

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/components/drag-handle-wrapper/__tests__/drag-handle-wrapper.test.tsx
+++ b/src/internal/components/drag-handle-wrapper/__tests__/drag-handle-wrapper.test.tsx
@@ -29,11 +29,28 @@ function getDirectionButton(direction: Direction) {
   );
 }
 
-function expectDirectionButtonHidden(direction: Direction) {
-  // Direction buttons get hidden via transition which doesn't end in JSDOM, so we just listen
-  // for the exiting classname instead.
-  const motionExitingClass = styles['direction-button-wrapper-motion-exiting'];
-  expect(getDirectionButton(direction)?.parentElement).toHaveClass(motionExitingClass);
+// Direction buttons get hidden via transition which doesn't end in JSDOM, so we use
+// the "exiting", "exited" or "hidden" classname instead to verify it's hidden.
+const DIRECTION_BUTTON_HIDDEN_CLASSES = [
+  styles['direction-button-wrapper-motion-exiting'],
+  styles['direction-button-wrapper-motion-exited'],
+  styles['direction-button-wrapper-hidden'],
+];
+
+function expectDirectionButtonToBeHidden(direction: Direction) {
+  expect(
+    DIRECTION_BUTTON_HIDDEN_CLASSES.some(className =>
+      getDirectionButton(direction)!.parentElement!.classList.contains(className)
+    )
+  ).toBe(true);
+}
+
+function expectDirectionButtonToBeVisible(direction: Direction) {
+  expect(
+    !DIRECTION_BUTTON_HIDDEN_CLASSES.some(className =>
+      getDirectionButton(direction)!.parentElement!.classList.contains(className)
+    )
+  ).toBe(true);
 }
 
 function renderDragHandle(props: Omit<DragHandleWrapperProps, 'children'>) {
@@ -139,7 +156,7 @@ test("doesn't show direction buttons by default", () => {
     directions: { 'block-start': 'active' },
   });
 
-  expect(getDirectionButton('block-start')).not.toBeInTheDocument();
+  expectDirectionButtonToBeHidden('block-start');
   expect(getDirectionButton('block-end')).not.toBeInTheDocument();
 });
 
@@ -169,8 +186,8 @@ describe('triggerMode = focus (default)', () => {
     expect(getDirectionButton('block-end')).toBeInTheDocument();
 
     fireEvent.blur(dragHandle);
-    expectDirectionButtonHidden('block-start');
-    expectDirectionButtonHidden('block-end');
+    expectDirectionButtonToBeHidden('block-start');
+    expectDirectionButtonToBeHidden('block-end');
   });
 });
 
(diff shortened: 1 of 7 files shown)


src/internal/components/drag-handle-wrapper/__tests__/drag-handle-wrapper.test.tsx
```diff
@@ -29,11 +29,28 @@ function getDirectionButton(direction: Direction) {
...
-  expect(getDirectionButton(direction)?.parentElement).toHaveClass(motionExitingClass);
+// Direction buttons get hidden via transition which doesn't end in JSDOM, so we use
+// the "exiting", "exited" or "hidden" classname instead to verify it's hidden.
+const DIRECTION_BUTTON_HIDDEN_CLASSES = [
+  styles['direction-button-wrapper-motion-exiting'],
+  styles['direction-button-wrapper-motion-exited'],
+  styles['direction-button-wrapper-hidden'],
+];
+
+function expectDirectionButtonToBeHidden(direction: Direction) {
```
johannes-weber, line 40: In case a direction button is active but not visible, it is rendered in the DOM to be ready for the transition - but set to invisible.

I also included the `*-motion-exiting` and `*-motion-exited` classes in that utility for more general test usage.

src/internal/components/drag-handle-wrapper/portal-overlay.tsx
```diff
@@ -16,14 +16,14 @@ export default function PortalOverlay({
   isDisabled,
   children,
 }: {
-  track: HTMLElement | null;
+  track: React.RefObject<HTMLElement | null>;
```
johannes-weber, line 19: We need to pass the `RefObject` instead of the `HTMLElement`.

Why? - During testing the `showButtons=true` flag, I discovered that `track` was all the time `null` on when rendering, no arrow buttons have been shown.
This was because `dragHandleRef` on the parent component was null on the initial render (passed null as `track` property). Once DOM element get assigned to the `dragHandleRef` in the parent component's, it won't trigger a re-render of `PortalOverlay` and it stayed null.

src/internal/components/drag-handle-wrapper/portal-overlay.tsx
```diff
@@ -36,19 +36,23 @@ export default function PortalOverlay({
       // It could be that the portal hasn't been attached to the DOM yet - ensure the ref exists and is attached DOM tree.
       if (ref.current && document.body.contains(ref.current)) {
         const isRtl = getIsRtl(ref.current);
-        const { insetInlineStart, insetBlockStart, inlineSize, blockSize } = getLogicalBoundingClientRect(track);
+        const { insetInlineStart, insetBlockStart, inlineSize, blockSize } = getLogicalBoundingClientRect(
+          track.current!
```
johannes-weber, line 40: `!` becaues we know that `track.current` won't be null because we checked it in line 26.
Not sure why TypeScript doesn't consider that check above.

avinashbot: It's because it's in a function and the current could be unset at any point. Perhaps you could add a check for `track.current` inside the if condition on line 37.

src/internal/components/drag-handle-wrapper/interfaces.ts
```diff
@@ -11,4 +11,5 @@ export interface DragHandleWrapperProps {
   tooltipText?: string;
   children: React.ReactNode;
   triggerMode?: TriggerMode;
+  showButtons?: boolean;
```
avinashbot, line 14: So this describes the starting state rather than the active value? Perhaps `initialShowButtons` or `initialOpen` would be better?

johannes-weber: That's right. I like the suggestion and think the "initial" in it makes it clear! Updating that.

src/internal/components/drag-handle-wrapper/portal-overlay.tsx
```diff
@@ -36,19 +36,23 @@ export default function PortalOverlay({
...
+        );
         // For simplicity, we just make all our calculations independent of
         // the browser's scrolling edge. When it comes to applying the changes,
         // translate is independent of writing direction, so we need to invert
         // the X coordinate ourselves just before applying the values.
         const newX = (insetInlineStart + getScrollInlineStart(document.documentElement)) * (isRtl ? -1 : 1);
         const newY = insetBlockStart + document.documentElement.scrollTop;
-        if (lastX !== newX || lastY !== newY) {
+        // Check if the current ref has style updates already applied (in case the ref changes)
```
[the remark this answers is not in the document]
johannes-weber: I cross checked the issue which I had before - I discovered an edge case when I used the `PortalOverlay` in a test component (where I switched the `track` for testing in runtime). Since this is not a use-case, I reverted the changes.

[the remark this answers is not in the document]
avinashbot: But the current code should still be fine if track changes, just not if ref changes, right?

src/internal/components/drag-handle-wrapper/index.tsx
```diff
@@ -156,7 +157,7 @@ export default function DragHandleWrapper({
       setShowButtons(false);
     } else if (triggerMode === 'keyboard-activate' && (event.key === 'Enter' || event.key === ' ')) {
       // toggle buttons when Enter or space is pressed in 'keyboard-activate' triggerMode
-      setShowButtons(prevShowButtons => !prevShowButtons);
+      setShowButtons(prevshowButtons => !prevshowButtons);
```
avinashbot, line 160: Super minor: the previous one seems more correct...

avinashbot (approved):

src/internal/components/drag-handle-wrapper/index.tsx
johannes-weber: I'll update that in a separate clean up PR. I'll go ahead and merge this one first.

PR #3516, opened 2025-05-16
chore: Improve Table task flow capture

michaeldowseza:
### Description

- Add support for pagination capture
- Capture on debounce to include client side
- Extended table demo to be more extension of a full table scenario

### How has this been tested?

- Updated unit tests
- Added new integ test for pagination

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/internal/hooks/use-table-interaction-metrics/__tests__/use-table-interaction-metrics.test.tsx
```diff
@@ -142,6 +150,7 @@ describe('useTableInteractionMetrics', () => {
         })
       );
 
+      jest.runAllTimers();
```
michaeldowseza, line 153: Since the event is emitted on a debounce and we are using fake timers here we need to run them before we can assert

src/internal/hooks/use-table-interaction-metrics/index.ts
```diff
@@ -16,25 +17,27 @@ to be the cause of the current loading state.
...
   itemCount: number;
   getComponentIdentifier: () => string | undefined;
   getComponentConfiguration: () => JSONObject;
   interactionMetadata: () => string;
 }
 
-export function useTableInteractionMetrics({
+export function useTableInteractionMetrics<T>({
   elementRef,
+  items,
```
avinashbot, line 33: Why is `items` needed at all? Just to be placed in the dependency array so that the effect fires when it changes? (If so, a comment here to prevent it from being accidentally removed in a refactor would be nice)

michaeldowseza: Yes, it's needed as a dependency since everything else is mostly coming from refs which don't trigger any changes. Added a comment.

avinashbot (approved):

PR #3519, opened 2025-05-17
chore: Improve Table task flow filtering capture

michaeldowseza:
### Description

Improves capture of table filtering analytics and extends capture to property filter.

### How has this been tested?

- Added additional dev page for a more extensive table example with property filter
- Added a new integration test

#### Correctness

#### Security

#### Testing

</details>

---


pages/funnel-analytics/with-table-property-filter.page.tsx
```diff
@@ -0,0 +1,174 @@
...
+  Button,
+  CollectionPreferences,
+  CollectionPreferencesProps,
+  Header,
+  Pagination,
+  PropertyFilter,
+  PropertyFilterProps,
+  Table,
+  TableProps,
+} from '~components';
```
orangevolon, line 17: I think we need to stick to our recommendation to import components separately.

pages/funnel-analytics/with-table-property-filter.page.tsx
```diff
@@ -0,0 +1,174 @@
...
+  contentDisplayPreference,
+  defaultPreferences,
+  EmptyState,
+  getMatchesCountText,
+  pageSizeOptions,
+  paginationLabels,
+} from '../table/shared-configs';
+
+const componentMetricsLog: any[] = [];
+(window as any).__awsuiComponentlMetrics__ = componentMetricsLog;
```
orangevolon, line 33: Non blocking but alternatively we could use:
```tsx
declare global {
  interface Window {
    __awsuiComponentMetrics: MetricsType
  }
}
```

pages/funnel-analytics/with-table-property-filter.page.tsx
```diff
@@ -0,0 +1,174 @@
...
+          />
+        ),
+      },
+      pagination: { pageSize: preferences.pageSize },
+      sorting: {},
+    }
+  );
+
+  return (
+    <>
```
orangevolon, line 111: Since we only have `<Table />` and no other child, does the fragment serve any purpose here?

src/property-filter/internal.tsx
```diff
@@ -185,6 +187,27 @@ const PropertyFilterInternal = React.forwardRef(
...
     })();
 
+    const countValue = useMemo(() => {
+      if (!countText || typeof countText !== 'string') {
+        return undefined;
+      }
+
+      const m = countText.match(/\d+/);
+      return m ? parseInt(m[0]) : 0;
+    }, [countText]);
```
orangevolon, line 197: I find this logic a bit confusing.

| `countText` | `countValue` |
| --- | --- |
| `undefined` | `undefined` |
| `""` | `undefined` |
| `"10"` | `10` |
| "invalid number" | `0` |

Should the fallback value be `undefined` or `0`?

src/property-filter/internal.tsx
```diff
@@ -185,6 +187,27 @@ const PropertyFilterInternal = React.forwardRef(
...
+
+    const tableComponentContext = useTableComponentsContext();
+    if (tableComponentContext?.filterRef?.current) {
+      const groupedTokens = tokenGroupToTokens<InternalToken>(internalQuery.tokens);
```
orangevolon, line 201: Wondering if it would be a good idea to wrap this into a `useEffect` instead of calling `tokenGroupToTokens`, `map` and `filter` on every re-render.

src/text-filter/internal.tsx
```diff
@@ -46,11 +46,22 @@ const InternalTextFilter = React.forwardRef(
...
     useForwardFocus(ref, inputRef);
 
+    const countValue = useMemo(() => {
+      if (!countText || typeof countText !== 'string') {
+        return undefined;
+      }
+
+      const m = countText.match(/\d+/);
+      return m ? parseInt(m[0]) : 0;
+    }, [countText]);
```
orangevolon, line 56: The same comment about the logic applies here too.

pages/funnel-analytics/with-table-property-filter.page.tsx
michaeldowseza: This is not standardised across our dev pages 🤷🏻‍♂️ What would be the benefit for us to do so?

src/internal/context/__tests__/table-component-context.test.tsx
```diff
@@ -12,6 +12,8 @@ describe('Verify TableComponentsContext', () => {
       return (
         <div>
           <div data-testid="filterText">{tableComponentsContext?.filterRef?.current?.filterText}</div>
+          <div data-testid="filterCount">{tableComponentsContext?.filterRef?.current?.filterCount}</div>
+          <div data-testid="filtered">{`${tableComponentsContext?.filterRef?.current?.filtered}`}</div>
```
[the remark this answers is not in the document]
michaeldowseza: React does not render the Boolean value here so string interpolation is needed.

```
    Expected element to have text content:
      true
    Received:
```

src/property-filter/internal.tsx
michaeldowseza: Maybe falling back to `undefined` make more sense here then. The case this would be present is a countText is provided in the text-filter/property-filter but it does not represent the number of matches.

src/table/__integ__/component-metrics.test.ts
```diff
@@ -112,3 +136,91 @@ describe('pagination', () => {
...
+      expect(componentsLog.length).toBe(2);
+      expect(componentsLog[1].name).toBe('componentUpdated');
+      expect(componentsLog[1].detail).toEqual({
+        taskInteractionId: expect.any(String),
+        componentName: 'table',
+        actionType: 'filter',
+        componentConfiguration: {
+          ...baseComponentConfiguration,
+          filtered: true,
+          totalNumberOfResources: 92,
```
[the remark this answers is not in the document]
michaeldowseza: Yes, from the filtered mock data.

pages/funnel-analytics/with-table-property-filter.page.tsx
michaeldowseza: I have a different future refactor in mind that'll add this :)

src/text-filter/internal.tsx
michaeldowseza: Addressed.

pages/funnel-analytics/with-table-property-filter.page.tsx
orangevolon: I think if we are [recommending the usage](https://cloudscape.aws.dev/get-started/integration/bundling/#import-components) to the customers, we might wanna apply it to our code too.
For us it won't have a benefit as we don't need to optimize bundling for the pages. For the customers who use our codebase as an example though it might be beneficial.

src/internal/context/__tests__/table-component-context.test.tsx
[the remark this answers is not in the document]
orangevolon: Got it 👍

orangevolon (approved):

PR #3520, opened 2025-05-18
chore: Improve Table task flow preferences capture

michaeldowseza:
### Description

- Capture table preference changes in existing componentUpdated event

For additional context you can look at this draft [PR](https://github.com/cloudscape-design/components/pull/3495) that includes all the relevant changes and shows the final state.

### How has this been tested?

- Added additional integ and unit test

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/collection-preferences/index.tsx
+++ b/src/collection-preferences/index.tsx
@@ -12,6 +12,7 @@ import { InternalButton } from '../button/internal';
 import { useInternalI18n } from '../i18n/context';
 import { getBaseProps } from '../internal/base-component';
 import { CollectionPreferencesMetadata } from '../internal/context/collection-preferences-metadata-context';
+import { useTableComponentsContext } from '../internal/context/table-component-context';
 import { fireNonCancelableEvent } from '../internal/events';
 import checkControlled from '../internal/hooks/check-controlled';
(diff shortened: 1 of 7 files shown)


src/collection-preferences/index.tsx
```diff
@@ -126,6 +127,19 @@ export default function CollectionPreferences({
   }
 
   const referrerId = useUniqueId();
+  const tableComponentContext = useTableComponentsContext();
+  if (tableComponentContext?.preferencesRef?.current) {
+    tableComponentContext.preferencesRef.current.pageSize = preferences?.pageSize;
```
just-boris, line 132: 1. Do you not need to clean up this if child component decides to unmount?
2. React does not guarantee that render function return value will be necessary flushed to the DOM. In [suspense](https://react.dev/reference/react/Suspense) use-case there could be fake renders which get interrupted before reaching the DOM. Only effect hooks promise reliable timing

michaeldowseza: Changed to `useEffect` hook. I can't think of any scenario where it might be unmounted separately from the table but I don't think there's any harm in adding the clean-up.

just-boris (approved):

PR #3529, opened 2025-05-21
chore: Refine exit motion for global drawer

at-susie:
### Description
This PR fixes an animation issue for global-drawer.
We would like to have a smooth transition when a drawer is being closed.

Before this change, the style drawer-hidden was added immediately when a close button is clicked, which was one of the two reasons for the unwanted sudden closing behavior. To address this, an additional condition has been added to make sure only after the Transition's component state becomes exited.:
`[styles['drawer-hidden']]: !show && state === 'exited'`
And I updated relevant styles accordingly.

Additionally, I've added Transition component to the `global-tools` in the skeleton to ensure only after it gets the `panel-hidden` style.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/visual-refresh-toolbar/drawer/global-drawer.tsx
+++ b/src/app-layout/visual-refresh-toolbar/drawer/global-drawer.tsx
@@ -75,7 +75,7 @@ function AppLayoutGlobalDrawerImplementation({
     (wasExpanded && !isExpanded);
 
   return (
-    <Transition nodeRef={drawerRef} in={show || isExpanded} appear={show || isExpanded} timeout={0}>
+    <Transition nodeRef={drawerRef} in={show || isExpanded} appear={show || isExpanded} timeout={60}>
       {state => {
         return (
           <aside
@@ -89,7 +89,7 @@ function AppLayoutGlobalDrawerImplementation({
               !animationDisabled && sharedStyles['with-motion-horizontal'],
               !animationDisabled && isExpanded && styles['with-expanded-motion'],
               {
-                [styles['drawer-hidden']]: !show,
+                [styles['drawer-hidden']]: !show && state === 'exited',
                 [styles['last-opened']]: lastOpenedDrawerId === activeDrawerId || isExpanded,
                 [testutilStyles['active-drawer']]: show,
                 [styles['drawer-expanded']]: isExpanded,
--- a/src/app-layout/visual-refresh-toolbar/drawer/styles.scss
+++ b/src/app-layout/visual-refresh-toolbar/drawer/styles.scss
@@ -64,6 +64,15 @@ $drawer-resize-handle-size: awsui.$space-m;
 
     @include desktop-only {
       inline-size: var(#{custom-props.$drawerSize});
+      &.exiting {
+        display: block;
+        border-inline-start: none;
+        > .global-drawer-wrapper {
+          opacity: 0;
+          // When multiple drawers are open and one of them is being closed, an unwanted width-changing transition appears.
+          // To avoid this, we set opacity to 0 to make the unwanted transition invisible.
+        }
+      }
     }
 
     @include mobile-only {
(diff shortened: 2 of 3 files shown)


src/app-layout/visual-refresh-toolbar/drawer/global-drawer.tsx
```diff
@@ -75,7 +75,7 @@ function AppLayoutGlobalDrawerImplementation({
     (wasExpanded && !isExpanded);
 
   return (
-    <Transition nodeRef={drawerRef} in={show || isExpanded} appear={show || isExpanded} timeout={0}>
+    <Transition nodeRef={drawerRef} in={show || isExpanded} appear={show || isExpanded} timeout={60}>
```
at-susie, line 78: @georgylobko suggested having [unmountOnExit](https://reactcommunity.org/react-transition-group/transition#Transition-prop-unmountOnExit) while keeping the `timeout={0}`. I explored that, but it doesn't seem to work. In my understanding, as long as we keep the `timeout={0}`, the `Transition` component instantly becomes `exited` state, which prevents us from having _exiting_ motion.

at-susie:
Replaced the 3rd party Transition component with the internal one. This change provides an expected motion behavior that we wanted to achieve with minimum change on the global drawer style.

One note is that the `appear` props is not available in the internal Transition component. If necessary we can introduce i, but before that I want to discuss if it's necessary.

src/app-layout/visual-refresh-toolbar/drawer/global-drawer.tsx
```diff
@@ -75,8 +75,9 @@ function AppLayoutGlobalDrawerImplementation({
     (wasExpanded && !isExpanded);
 
   return (
-    <Transition nodeRef={drawerRef} in={show || isExpanded} appear={show || isExpanded} timeout={0}>
-      {state => {
+    <Transition in={show} exit={!show}>
```
at-susie, line 78: Note: Removed `isExpanded` condition because when 'isExpanded' it's always `show`.

at-susie: Note: The appear props is not available in the internal Transition component. If necessary we can introduce it, but before that I want to discuss if it's necessary.

src/app-layout/visual-refresh-toolbar/drawer/global-drawer.tsx
```diff
@@ -75,8 +75,9 @@ function AppLayoutGlobalDrawerImplementation({
...
+      {(state, transitionRef) => {
+        console.log('Transition state:', state);
```
just-boris, line 80: Need to remove this console log

```suggestion
```

at-susie: Removed

georgylobko (changes requested):
> Fixed by removing the defaultActive condition from animationDisabled

this code was introduced on purpose and should be there, it makes sure that animation is not applied on initial load in consoles. you cannot just remove existing conditions

> Fixed by using hasOpened state to control Transition's in prop

why? you already have a prop `show`. what is the purpose of this addition?

at-susie:
@georgylobko Thanks for your feedback!

>Fixed by removing the defaultActive condition from animationDisabled

Got it, I've brought it back. 


>Fixed by using hasOpened state to control Transition's in prop

When a drawer mounts with `show=true`, the Transition component's initial state is 'entered' (from [its useState logic](https://github.com/cloudscape-design/components/pull/3529/files#diff-729e8465a999c31d5ca4256c933336e4a0632cf67f15aeb2265fba3b5871bea8R44)). This skips the animation sequence because it starts in the final state, preventing it from having the first-open motion. This ensures we get:

- Initial mount: 'exited' state
- After useEffect: 'enter' -> 'entering' -> 'entered'

Without `hasOpened`, we'd get:

- Initial mount: 'entered' state (no animation)
- Subsequent opens: 'enter' -> 'entering' -> 'entered'

The state is necessary to force the proper animation sequence on first mount, while still respecting the `animationDisabled` logic.

just-boris:
Closing for housecleaning. Feel free to ping if you would like to resume this work

PR #3530, opened 2025-05-21
chore: Updating tests to the masking to ensure date formats are covered

dpitcock:
### Updating tests to the masking utils

This pull request adds more comprehensive tests for the useMask hook in the masked-input component. The new tests cover the following scenarios:

Date Masks
- Handling of different date formats (slashed, ISO)
- Validation and auto-completion for day, month, and year segments
- Handling of separators in date strings

Time Masks
- Validation and auto-completion different segments
- Handling of different separators (:, -, /)
- Handling of partial time inputs

General Improvements
- Refactored the test setup to use a more parameterized approach, making it easier to add new mask types in the future
- Added tests to ensure the correct behavior when the maximum value for a segment is reached
- Improved the handling of invalid input values, ensuring the hook returns a valid value even when the initial input is invalid

These changes ensure that the useMask hook can handle a wider range of date and time formats, and that the masking functionality is thoroughly tested across various use cases.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/internal/components/masked-input/__tests__/utils/mask-format.test.ts
```diff
@@ -1,117 +1,509 @@
...
+        });
+      });
 
-    test('should not change a complete value', () => {
-      expect(maskFormat.autoComplete('01:00')).toBe('01:00');
-      expect(maskFormat.autoComplete('04:20')).toBe('04:20');
-    });
-  });
+      describe('getValidValue', () => {
+        mask.segments.forEach((segment, segmentIndex) => {
```
dpitcock, line 223: This is ugly, but I couldn't think of a better way that encompasses a test case for each segment. It looks ugly here, but in the test breakdown it all makes sense

dpitcock: Thanks for the confirmation. Yea we might even need tests for the tests for the tests...  I will do more like I've done below.

pan-kot: I kept all tests but made setup declarative, removing all iterators. @gethinwebster please take another look

src/internal/components/masked-input/__tests__/utils/mask-format.test.ts
```diff
@@ -1,117 +1,177 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
+
 import MaskFormat from '../../../../../../lib/components/internal/components/masked-input/utils/mask-format';
 
+const hourMask = { min: 0, max: 23, length: 2 };
+const minuteMask = { min: 0, max: 59, length: 2 };
+const secondMask = { min: 0, max: 59, length: 2 };
+function createTimeMask(separator: ':' | '-', segmentsCount?: 2 | 3) {
+  segmentsCount = segmentsCount ?? (Math.random() > 0.5 ? 2 : 3);
```
gethinwebster, line 10: why is there randomness here?

src/internal/components/masked-input/utils/mask-format.ts
```diff
@@ -60,33 +62,33 @@ class MaskFormat {
     return inputSegments.every((segmentValue, i) => {
       const segment = this.segments[i];
 
-      // disallow empty segments
+      // Disallow empty segments
```
gethinwebster, line 65: suggest removing these writing changes, I'm not sure they serve any real purpose and just end up making the diff/history more difficult to follow

src/internal/components/masked-input/__tests__/utils/mask-format.test.ts
pan-kot: That is because for tests that assert the first 1-2 segments, it should not matter whether the mask supports 2 or 3 segments.

src/internal/components/masked-input/utils/mask-format.ts
pan-kot: I would keep these as the file also includes a couple of real changes - adding a new comment, and removing an unnecessary one. But I noted your feedback for the future!

PR #3540, opened 2025-05-27
feat: Implementation of the Button component Style API.

DaemonCahill:
This PR implements the core Style API for the Button component.


just-boris (approved):

src/button/styles.scss
```diff
@@ -63,6 +65,8 @@
   display: inline-block; // for <a> as a button
   text-decoration: none;
   cursor: pointer;
+  #{custom-props.$styleFocusRingBoxShadow}: 0 0 0 var(#{custom-props.$styleFocusRingBorderWidth}, 2px)
```
gethinwebster, line 68: I think it'd be good to create a variable for this `2px` e.g. in internal/styles/foundation/shadows.scss so that we're not ending up with the same magic number in multiple places

Al-Dani (dismissed):

PR #3549, opened 2025-06-03
chore: Track table counter text from title as fallback

michaeldowseza:
### Description

Extracts count text from Table header as a fallback to the Filter count when a custom filter is used.

### How has this been tested?

- Added integ tests for the Table integration
- Added unit tests for the added utility method that does the parsing

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/analytics/utils/parse-count-text.ts
+++ b/src/internal/analytics/utils/parse-count-text.ts
@@ -1,0 +1,21 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * Extracts the count value from table header/filter text.
+ *
+ * Parses various counter string formats and extracts the relevant numeric value:
+ * - "Items (100)" - Extracts 100 (first number found)
+ * - "1/100" - Extracts 100 (denominator of fraction, representing total count)
+ * - "100+" - Extracts 100 (first number found)
+ * - "1,234 items" - Extracts 1234 (handles comma-separated numbers)
+ */
+export const parseCountValue = (countText: string | undefined): number | undefined => {
+  if (!countText || typeof countText !== 'string') {
+    return undefined;
+  }
+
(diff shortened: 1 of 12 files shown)


src/internal/analytics/utils/parse-count-text.ts
```diff
@@ -0,0 +1,21 @@
...
+ * - "1/100" - Extracts 100 (denominator of fraction, representing total count)
+ * - "100+" - Extracts 100 (first number found)
+ * - "1,234 items" - Extracts 1234 (handles comma-separated numbers)
+ */
+export const parseCountValue = (countText: string | undefined): number | undefined => {
+  if (!countText || typeof countText !== 'string') {
+    return undefined;
+  }
+
+  const target = countText.includes('/') ? countText.split('/')[1] : countText;
```
jperals, line 17: The number found at right part of the slash could at the same time contain a comma, right?

src/internal/analytics/utils/parse-count-text.ts
```diff
@@ -0,0 +1,21 @@
...
+  const match = target.match(/\d[\d,]*/);
+  return match ? parseInt(match[0].replace(/,/g, ''), 10) : undefined;
```
jperals: The comma is language-specific, some languages use dot instead, and even space and "half-space" is being proposed as standards: https://en.wikipedia.org/wiki/Decimal_separator#Digit_grouping

But before getting deeper let's first make sure this is actually worth handling —have we seen customers actually use separators in the counters?

src/internal/analytics/utils/parse-count-text.ts
michaeldowseza: As discussed. Removed this use case until we get more data.

src/internal/analytics/utils/parse-count-text.ts
michaeldowseza: Same as above.

jperals (approved):

PR #3553, opened 2025-06-03
feat: Add children-based API for button group

gethinwebster:
- **feat: Allow button native element attributes to be set**
- **feat: Add children-based API for button group**

### Description

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/button-group/interfaces.ts
```diff
@@ -90,7 +92,20 @@ export interface ButtonGroupProps extends BaseComponentProps {
...
+   *
+   * Each item should contain a single focusable element which must have a unique `data-itemid` attribute.
+   *
+   * For example:
+   * - `<Button data-itemid="button-one">One</Button>`
+   * - `<ButtonDropdown nativeButtonAttributes={{'data-itemid': 'button-two'}} items={...} />`
+   *
+   * @awsuiSystem core
+   */
+  children?: ReactNode;
```
pan-kot, line 108: I think this denies the purpose of the button group component, the main complexity of which lies in the way we deal with focusing, keyboard nav, and tooltips - neither of these features can be easily replicated from the outside as we don't share the related components/tools.

Should we instead think of some advanced custom item type, e.g. `{ type: 'custom', button: (props) => <MyButton {...props} />, tooltip: (props) => <MyTooltipContent {...props} /> }`?

gethinwebster: It still deals with focus & keyboard navigation, only tooltips are not included. But the tooltips are only really relevant for icon-only buttons, so I don't think there's so much lost by not supporting those for custom buttons which by definition are likely to be different from what's supported via `items`

pages/button-group/composition.page.tsx
```diff
@@ -0,0 +1,31 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import React from 'react';
+
+import { Box, Button, ButtonDropdown, ButtonGroup, Header } from '~components';
+
+export default function ButtonGroupPage() {
```
pan-kot, line 8: Let's also update the permutations page for screenshot test cov.

src/button-group/interfaces.ts
pan-kot: Is it possible to combine the children and items API? Otherwise, the icon-buttons added as children won't have tooltips, which is a downgrade to UX and potentially an a11y issue.

pan-kot: Should we use `before` and `after` instead of children, maybe? So then one can use `items` but also append or prepend some custom actions. Alternatively, we can introduce a new item type "custom" - so one can pass it like `items: [{ id: 'custom-button', type: 'custom', content: <Button /> }]`.

src/button-group/interfaces.ts
```diff
@@ -90,7 +92,20 @@ export interface ButtonGroupProps extends BaseComponentProps {
    * * `text` (string) - The name of the group rendered as ARIA label for this group.
    * * `items` ((ButtonGroupProps.IconButton | ButtonGroupProps.MenuDropdown)[]) - The array of items that belong to this group.
    */
-  items: ReadonlyArray<ButtonGroupProps.ItemOrGroup>;
+  items?: ReadonlyArray<ButtonGroupProps.ItemOrGroup>;
+  /**
+   * Instead of passing `items`, group items can be passed as children. When using this,
+   * you will need to manage click handlers and state of buttons.
+   *
+   * Each item should contain a single focusable element which must have a unique `data-itemid` attribute.
```
pan-kot, line 100: This sentence is a bit confusing, until one sees the example. Can we say it more explicitly, e.g. "Pass interactive elements directly to the children prop as a React elements array. Every item in the array ..."?

src/button-group/interfaces.ts
```diff
@@ -106,7 +121,12 @@ export interface InternalButtonGroupProps
     InternalBaseComponentProps {}
 
 export namespace ButtonGroupProps {
-  export type Variant = 'icon';
+  export type Variant =
+    | 'icon'
+    /**
+     * @awsuiSystem core
+     */
+    | 'children';
```
pan-kot, line 129: I guess we do this as otherwise the following is confusing: `<ButtonGroup variant="icon" children={children} />`. However, this can be fixed if we allow both items and children to co-exist.

src/button-group/internal.tsx
```diff
@@ -95,7 +92,10 @@ const InternalButtonGroup = forwardRef(
         return;
       }
       // Ignore navigation when the focused element is not an item.
-      if (document.activeElement && !document.activeElement.matches(`.${testUtilStyles.item}`)) {
+      if (
+        document.activeElement &&
+        !('dataset' in document.activeElement && (document.activeElement as HTMLElement).dataset.itemid)
```
pan-kot, line 97: Should we add a dev warning if finding a focusable item w/o itemid? That should only be possible if the consumers misuse the API, forgetting the item IDs.

src/button-group/internal.tsx
```diff
@@ -148,43 +148,44 @@ const InternalButtonGroup = forwardRef(
...
-                  <div key={index} role="group" aria-label={itemOrGroup.text} className={styles.group}>
-                    {itemOrGroup.items.map((item, subIndex) => itemContent(item, `${index + 1},${subIndex + 1}`))}
-                  </div>
-                ) : (
-                  itemContent(itemOrGroup, `${index + 1}`)
-                )}
-              </React.Fragment>
-            );
-          })}
+          {variant === 'children'
```
pan-kot, line 151: I think that we need to allow both items and custom items simultaneously. If decided otherwise, I recommend adding a dev warning if both items and children props are used (or, rather if using "icon" variant with no items or "children" variant with no childen).

src/button-group/__tests__/button-group-children.test.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+import React from 'react';
+import { render } from '@testing-library/react';
+
+import Button from '../../../lib/components/button';
+import ButtonGroup from '../../../lib/components/button-group';
+import createWrapper from '../../../lib/components/test-utils/dom';
+
+describe('ButtonGroup children property', () => {
+  test('renders children', () => {
```
pan-kot, line 12: I would add a couple more tests:
* Rendering a button w/o testid (assert warning, make sure it won't crash);
* Re-rendering the button group with an extra button (make sure it is included to the keyboard nav);
* Re-rendering the button group with focused button gone (make sure re-focusing occurs).

gethinwebster:
Closing for #3785

PR #3554, opened 2025-06-04
feat: Allow button-dropdown native attributes to be set

gethinwebster:
### Description

Allow button native attributes to be set.

Related links, issue #, if available: AWSUI-60864

### How has this been tested?

New unit tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/button-dropdown/interfaces.ts
+++ b/src/button-dropdown/interfaces.ts
@@ -127,10 +127,10 @@ export interface ButtonDropdownProps extends BaseComponentProps, ExpandToViewpor
    *
    * @awsuiSystem core
    */
-  nativeButtonAttributes?: ButtonProps['nativeAttributes'];
+  nativeTriggerAttributes?: ButtonProps['nativeAttributes'];
 
   /**
-   * Attributes to add to the native  `button` element of the `mainAction`.
+   * Attributes to add to the native `button` element of the `mainAction`.
    *
    * It is not supported to use this attribute to apply custom styling.
    *
--- a/src/button-dropdown/interfaces.ts
+++ b/src/button-dropdown/interfaces.ts
@@ -119,6 +119,24 @@ export interface ButtonDropdownProps extends BaseComponentProps, ExpandToViewpor
    * Sets the button width to be 100% of the parent container width. Button content is centered.
    */
   fullWidth?: boolean;
+
+  /**
+   * Attributes to add to the native `button` element.
+   *
+   * It is not supported to use this attribute to apply custom styling.
+   *
+   * @awsuiSystem core
+   */
+  nativeButtonAttributes?: ButtonProps['nativeAttributes'];
+
+  /**
+   * Attributes to add to the native  `button` element of the `mainAction`.
+   *
+   * It is not supported to use this attribute to apply custom styling.
+   *
+   * @awsuiSystem core
+   */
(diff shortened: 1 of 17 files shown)


src/button/index.tsx
```diff
@@ -40,6 +40,7 @@ const Button = React.forwardRef(
       fullWidth,
       form,
       i18nStrings,
+      nativeAttributes,
```
pan-kot, line 43: Should we add them to metrics metadata? E.g. `nativeAttributesUsed: boolean`

gethinwebster: This is only exposed for core, and core shouldn't be used anywhere where metrics metadata is actually gathered

src/button/interfaces.ts
```diff
@@ -100,6 +100,17 @@ export interface BaseButtonProps {
    * @i18n
    */
   i18nStrings?: ButtonProps.I18nStrings;
+
+  /**
+   * Attributes to add to the native `button` element.
+   *
+   * It is not supported to use this attribute to apply custom styling.
```
pan-kot, line 107: What does "not supported" actually mean? Technically, style property can still be used.

src/button-dropdown/interfaces.ts
```diff
@@ -119,6 +119,24 @@ export interface ButtonDropdownProps extends BaseComponentProps, ExpandToViewpor
...
   fullWidth?: boolean;
+
+  /**
+   * Attributes to add to the native `button` element.
+   *
+   * It is not supported to use this attribute to apply custom styling.
+   *
+   * @awsuiSystem core
+   */
+  nativeButtonAttributes?: ButtonProps['nativeAttributes'];
```
pan-kot: Should we call this `nativeTriggerAttributes`? Otherwise, one might think that these attributes are applied to each dropdown button, or maybe both trigger and main buttons, while native main button attributes can extend/override this.

src/button/interfaces.ts
gethinwebster: it means we don't provide any promises that it will not break in the future if you use it for styling.

src/button-dropdown/interfaces.ts
```diff
@@ -119,6 +119,24 @@ export interface ButtonDropdownProps extends BaseComponentProps, ExpandToViewpor
...
+
+  /**
+   * Attributes to add to the native  `button` element of the `mainAction`.
```
pan-kot: nit: there is an extra space after "native"

src/button-dropdown/interfaces.ts
gethinwebster: good point, will update

src/button/interfaces.ts
pan-kot: Should we say something like: "Do not use this property to provide custom inline styles to the component"?

src/button/internal.tsx
```diff
@@ -326,6 +322,7 @@ export const InternalButton = React.forwardRef(
           disabled={disabled && !__focusable && !isDisabledWithReason}
           aria-disabled={hasAriaDisabled ? true : undefined}
           {...disabledReasonProps}
+          {...nativeAttributes}
```
pan-kot, line 325: This can technically override the button class names, and there is no ability to access the button's assigned classnames from the outside. Should we either exclude `className` from the native attributes type, or make sure the class name that comes from native attributes is merged with the one we define using clsx?

PR #3557, opened 2025-06-05
feat: Implementation of the Flashbar component Style API.

DaemonCahill:
This PR implements the core Style API for the Flashbar component.


pages/app/themes/style-api.tsx
```diff
@@ -0,0 +1,52 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+export const fontFamily =
+  'InterVariable, "Inter var", Inter, -apple-system, "system-ui", "Helvetica Neue", "Segoe UI", Oxygen, Ubuntu, Cantarell, "Open Sans", sans-serif';
```
just-boris: This list of fonts is not deterministic it depends on OS and which fonts are installed.

This may cause problems with screenshot tests. Can we pick some cross-platform options here?

pages/flashbar/style-permutations.page.tsx
```diff
@@ -0,0 +1,92 @@
...
+      [sampleNotifications.success],
+      [sampleNotifications.error],
+      [sampleNotifications.warning],
+      [sampleNotifications.info],
+      [sampleNotifications['in-progress']],
+    ],
+  },
+]);
+
+export default function FlashbarStylePermutations() {
```
just-boris: Why do we need both pages, `style-custom.page.tsx` and `style-permutations.page.tsx`? Don't they show the same use-case

src/__tests__/snapshot-tests/__snapshots__/documenter.test.ts.snap
```diff
@@ -20062,6 +20081,25 @@ It prevents users from clicking the button, but it can still be focused.",
...
       "type": "string",
     },
+    {
+      "inlineType": {
+        "name": "ButtonProps.Style",
+        "properties": [
+          {
+            "name": "root",
+            "optional": true,
+            "type": "{ background?: { active?: string | undefined; default?: string | undefined; disabled?: string | undefined; hover?: string | undefined; } | undefined; borderColor?: { active?: string | undefined; default?: string | undefined; disabled?: string | undefined; hover?: string | undefined; } | undefined; ... 5 more ...; pa...",
```
just-boris: oof, this `... 5 more ...; pa...`  will be printed as is on the website.

We need to do something about fixing the documentation printout eventually.

src/flashbar/collapsible-flashbar.tsx
```diff
@@ -265,14 +265,29 @@ export default function CollapsibleFlashbar({ items, ...restProps }: FlashbarPro
...
                 }}
-                style={
-                  !isFlashbarStackExpanded || transitioning
-                    ? {
-                        [customCssProps.flashbarStackIndex]:
-                          (item as StackableItem).collapsedIndex ?? (item as StackableItem).expandedIndex ?? index,
-                      }
-                    : undefined
-                }
+                style={{
```
just-boris, line 302: Call it a personal preference, but I think with that much inline conditions, this `style` object build would be more readable after extracting into a separate function

just-boris (commented):
Minor comments only

src/flashbar/interfaces.ts
```diff
@@ -40,6 +40,69 @@ export namespace FlashbarProps {
     warningIconAriaLabel?: string;
   }
 
+  export interface Style {
+    item?: {
+      root?: {
```
gethinwebster, line 45: this also needs `focusRing`
Screenshot 2025-06-12 at 11 53 44
, so that a focused flash item can have an accessible focus ring color set

src/flashbar/interfaces.ts
```diff
@@ -40,6 +40,69 @@ export namespace FlashbarProps {
...
+        borderColor?: {
+          error?: string;
+          info?: string;
+          inProgress?: string;
+          success?: string;
+          warning?: string;
+        };
+        borderRadius?: string;
+        borderWidth?: string;
+        color?: {
```
gethinwebster, line 62: What's the plan for dealing with visual contexts? Currently, e.g. a link inside a flashbar gets the incorrect color 
Screenshot 2025-06-12 at 11 53 44, applies for various other tokens too: https://github.com/cloudscape-design/components/blob/scottob-flashbar-style-api/style-dictionary/visual-refresh/contexts/flashbar.ts#L9

gethinwebster (changes requested):

pages/app/themes/style-api.tsx
DaemonCahill: Good callout. I'll remove this from the Button PR.

pages/flashbar/style-permutations.page.tsx
DaemonCahill: Good callout. I'll merge them into a single test page.

src/__tests__/snapshot-tests/__snapshots__/documenter.test.ts.snap
DaemonCahill: Okay I will work with Amr on this one.

src/flashbar/collapsible-flashbar.tsx
DaemonCahill: Cleaned it up using functions.

src/flashbar/interfaces.ts
DaemonCahill: The initial support for the Style API will not handle visual contexts. Nested components in slots like Links, Buttons, etc will also need to be styled.

src/flashbar/interfaces.ts
DaemonCahill: Ah! I didn't know that was focusable. Adding...

src/flashbar/__integ__/pages/base.ts
```diff
@@ -22,9 +22,26 @@ export class FlashbarBasePage extends BasePageObject {
...
+    return createWrapper().findFlashbar().findByClassName(selectors['dismiss-button']).toSelector();
+  }
+
   isFlashFocused(index: number) {
     return this.isFocused(
       flashbar.findItems().get(index).findByClassName(selectors['flash-focus-container']).toSelector()
     );
   }
+
+  getComputedStyles(selector: string, property: string, pseudoElement: string | null = null) {
```
just-boris: 1. Did you try using the built-in method: https://webdriver.io/docs/api/element/getCSSProperty
1. If you tried and you are sure we need a custom one, let's move it to https://github.com/cloudscape-design/browser-test-tools/blob/main/src/page-objects/base.ts

DaemonCahill: Native works!

src/flashbar/flash.tsx
```diff
@@ -170,6 +188,38 @@ export const Flash = React.forwardRef(
       analyticsAttributes[DATA_ATTR_ANALYTICS_SUPPRESS_FLOW_EVENTS] = 'true';
     }
 
+    const getFlashStyles = () => {
```
just-boris: Is this a duplication with `getItemStyles` code from `src/flashbar/collapsible-flashbar.tsx`?

Can we reuse this code somehow?

just-boris (dismissed):
Approved, but it would be nice to address that comment

just-boris (approved):

PR #3585, opened 2025-06-17
chore: Use CSS-native :focus-visible to track visible focus state

avinashbot:
### Description

See https://caniuse.com/css-focus-visible. We were waiting on Safari until last year, but I think we can do without the JS fallback now. As long as nobody managed to pull a sneaky runtime dependency on `data-awsui-focus-visible`, we should be good to switch.

Actually, I was the one who pulled a sneaky runtime dependency on it while implementing the UAP buttons, and I had to fiddle with the tests slightly (see the `document.body.matches` mocks in some of the test files).

Related links, issue #, if available: AWSUI-60845

### How has this been tested?

Will test in my dev pipeline. A manual bug bash may also be in order, but it doesn't have to be a big thing.

The dry-run failure for chat components is fine. It'll go green again when https://github.com/cloudscape-design/chat-components/pull/75 is merged, and shouldn't stop you from reviewing the PR.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/common/popover-drilldown-sample-data.ts
+++ b/pages/common/popover-drilldown-sample-data.ts
@@ -31,6 +31,35 @@ export default [
     ],
   },
   {
+    title: 'AWS-MiscellaneousOverflowingStringNameService',
+    data: [
+      {
+        x: '2023-04-01',
+        y: 0,
+      },
+      {
+        x: '2023-05-01',
+        y: 12,
+      },
+      {
+        x: '2023-06-01',
+        y: 31,
+      },
+      {
+        x: '2023-07-01',
+        y: 52,
+      },
+      {
+        x: '2023-08-01',
+        y: 24,
+      },
+      {
+        x: '2023-09-01',
+        y: 4,
+      },
+    ],
+  },
+  {
     title: 'Amazon Relational Database Service',
     data: [
       {
--- a/pages/flashbar/style-custom.page.tsx
+++ b/pages/flashbar/style-custom.page.tsx
@@ -1,0 +1,162 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import React from 'react';
+
+import { Flashbar, FlashbarProps, SpaceBetween } from '~components';
+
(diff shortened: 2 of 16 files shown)


src/app-layout/__tests__/split-panel.test.tsx
```diff
@@ -45,7 +45,13 @@ beforeEach(() => {
   Object.defineProperty(document.documentElement, 'clientHeight', { value: 800, configurable: true });
   originalGetComputedStyle = window.getComputedStyle;
   window.getComputedStyle = fakeComputedStyle;
+
+  const originalMatches = document.body.matches;
+  jest.spyOn(document.body, 'matches').mockImplementationOnce(selectors => {
+    return selectors === ':has(:focus-visible)' ? true : originalMatches.call(document.body, selectors);
+  });
```
just-boris, line 52: Why is this needed?

src/internal/components/drag-handle-wrapper/__tests__/drag-handle-wrapper.test.tsx
```diff
@@ -18,8 +18,14 @@ beforeAll(() => {
   (window as any).PointerEvent ??= PointerEventMock;
 });
 
+beforeEach(() => {
+  const originalMatches = document.body.matches;
+  jest.spyOn(document.body, 'matches').mockImplementation(selectors => {
+    return selectors === ':has(:focus-visible)' ? true : originalMatches.call(document.body, selectors);
+  });
+});
```
just-boris, line 26: 1. Same question, why this needed?
2. If it repeats twice, is it time to extract a helper like `simulateFocusForUseCaseNNN`?
3. Currently the mock applies to all tests. Do you need a test when focus-visible state is not active?

avinashbot: We were simulating `data-awsui-focus-visible` in some tests, and I replaced them with `document.matches(':has(:focus-visible))')`. But it makes more sense to put this logic and the mocking function responsible for it in one central place. Will do when I open the "chunk PR" that includes this.

PR #3591, opened 2025-06-19
chore: Detect and report missing css styles

just-boris:
### Description

Added check for this old issue: AWSUI-55709

Since releasing VR only artifacts, a few more teams reported the same. Let's add a check to monitor this.

### How has this been tested?

* Added a unit test
* Checked the dev pages locally to ensure no false alarms
* Deployed to my dev pipeline, reproduced missing styles via Devtools' request blocking

<details>

#### Correctness

#### Security

#### Testing

</details>

---


build-tools/tasks/clean.js
```diff
@@ -14,6 +14,7 @@ module.exports = task('clean', () => {
       `${workspace.targetPath}/**`,
       `${workspace.staticSitePath}/**`,
       `${workspace.generatedTestUtils}/**`,
+      `${workspace.generatedPath}/custom-css-properties/**`,
```
just-boris, line 17: `workspace.generatedPath`  contains other non-generated files: https://github.com/cloudscape-design/components/tree/main/src/internal/generated

Ideally we should clean up them all (or move to a different path), but not in this PR

src/internal/base-component/styles.scss
```diff
@@ -9,3 +9,8 @@
 */
 @use 'awsui:globals';
 @use '../styles/global.scss';
+@use '../generated/custom-css-properties' as custom-styles;
+
+:root {
+  --awsui-version-info-#{custom-styles.$awsui-commit-hash}: true;
```
just-boris, line 15: Core of this check. We check presence of this CSS var in Javascript. If not found – the styles are missing

src/internal/metrics.ts
```diff
@@ -2,6 +2,6 @@
 // SPDX-License-Identifier: Apache-2.0
 import { Metrics } from '@cloudscape-design/component-toolkit/internal';
 
-import { PACKAGE_SOURCE, PACKAGE_VERSION } from './environment';
+import { PACKAGE_SOURCE, PACKAGE_VERSION, THEME } from './environment';
 
-export const metrics = new Metrics(PACKAGE_SOURCE, PACKAGE_VERSION);
+export const metrics = new Metrics({ packageSource: PACKAGE_SOURCE, packageVersion: PACKAGE_VERSION, theme: THEME });
```
just-boris, line 7: Addressing this follow-up https://github.com/cloudscape-design/component-toolkit/pull/136#discussion_r2153239411

Yes I know this `THEME` value does not support `vr`, but it is fine for this use-case

src/internal/hooks/use-base-component/styles-check.ts
```diff
@@ -0,0 +1,54 @@
...
+  const result = getComputedStyle(document.body).getPropertyValue(`--awsui-version-info-${GIT_SHA}`);
+  if (!result) {
+    console.error(
+      `Missing Cloudscape CSS for theme "${THEME}", version "${PACKAGE_VERSION}", and git sha "${GIT_SHA}".`
+    );
+    metrics.sendOpsMetricObject('awsui-missing-css-asset', {});
+  }
+}
+
+export function idleWithDelay(cb: () => void) {
```
just-boris, line 16: Exported `checkMissingStyles` and `idleWithDelay` as separate functions to allow write simpler unit tests on each part of the functionality

src/__integ__/themes.test.ts
```diff
@@ -23,7 +23,7 @@ class CustomPropertyPageObject extends BasePageObject {
         for (const [prop, value] of (element as any).computedStyleMap()) {
           // Custom Property
           if (prop.startsWith('--')) {
-            const valueWithoutPostfix = prop.substring(2, prop.length - 7);
+            const valueWithoutPostfix = prop.replace(/^--/, '').replace(/-[\d\w]+$/, '');
```
just-boris, line 26: This code assumed the hash is always 7 characters, which is not true

src/internal/base-component/styles.scss
abdhalees: so if an app doesn't bundle and server styles but another app on the same page has a similar git commit. it wouldn't be detected. 

but I guess this is acceptable, and better than what we have now.

src/internal/metrics.ts
abdhalees: can you explain a bit more?

abdhalees (approved):

src/internal/base-component/styles.scss
just-boris: Yes, there is a chance of missing some cases, if they happen to have the same commit

src/internal/metrics.ts
just-boris: Before it was like this

```js
const metrics = new Metrics(PACKAGE_SOURCE, PACKAGE_VERSION);
metrics.initMetrics(theme);
```

This is not safe, because it `metrics.initMetrics(theme)` sets theme as a global variable inside `@cloudscape-design/component-toolkit` package

Multiple packages may call the same toolkit instance and the same `metrics.initMetrics(theme)` will overwrite each other theme.

So, now we finally got the theme properly scoped to each bundle

PR #3610, opened 2025-06-25
chore: Upgrade dependencies

just-boris:
### Description

Run `npm upgrade` and fix all conflicts

### How has this been tested?

* Dry-run
* Deployment to my dev pipeline

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3627,21 +3627,21 @@
       "license": "MIT"
     },
     "node_modules/@types/react": {
-      "version": "16.14.65",
+      "version": "16.14.34",
       "dev": true,
       "license": "MIT",
       "dependencies": {
         "@types/prop-types": "*",
-        "@types/scheduler": "^0.16",
+        "@types/scheduler": "*",
         "csstype": "^3.0.2"
       }
     },
     "node_modules/@types/react-dom": {
-      "version": "16.9.25",
+      "version": "16.9.17",
       "dev": true,
       "license": "MIT",
-      "peerDependencies": {
-        "@types/react": "^16.0.0"
+      "dependencies": {
+        "@types/react": "^16"
       }
     },
     "node_modules/@types/react-router": {
@@ -3672,10 +3672,10 @@
       }
     },
     "node_modules/@types/react-transition-group": {
-      "version": "4.4.12",
+      "version": "4.4.5",
       "dev": true,
       "license": "MIT",
-      "peerDependencies": {
+      "dependencies": {
         "@types/react": "*"
       }
     },


package.json
```diff
@@ -68,15 +68,14 @@
     "@types/balanced-match": "^1.0.1",
     "@types/d3-scale": "^2.2.6",
     "@types/d3-shape": "^1.3.8",
-    "@types/glob": "^7.2.0",
```
just-boris, line 71: Not needed, because latest Glob has types built-in

package.json
```diff
@@ -68,15 +68,14 @@
...
     "@types/jest": "^29.5.13",
     "@types/lodash": "^4.14.176",
     "@types/node": "^20.17.14",
     "@types/react": "^16.14.20",
     "@types/react-dom": "^16.9.14",
     "@types/react-router": "^5.1.18",
     "@types/react-router-dom": "^5.3.2",
-    "@types/react-test-renderer": "^18.3.0",
+    "@types/react-test-renderer": "^16.9.12",
```
just-boris, line 78: This version must be aligned with the rest of React typings

package.json
```diff
@@ -138,12 +137,15 @@
...
-    "typescript": "^4.6.4",
+    "typescript": "^4.9.5",
     "wait-on": "^8.0.2",
     "webpack": "^5.94.0",
     "webpack-cli": "^5.1.4",
     "webpack-dev-server": "^5.2.1"
   },
+  "overrides": {
+    "typescript": "^4.9.5"
+  },
```
just-boris, line 148: Mirroring the override from other packages we own: https://github.com/cloudscape-design/board-components/blob/af05dce2fd7907230b48740ff88eb4d20aef2dbb/package.json#L114-L116

Without this line, the main build and the documenter use different typescript versions which leads to issues

tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "lib": ["ES2020", "DOM"],
+    "lib": ["ES2021", "DOM"],
```
just-boris, line 3: Needed by `@formatjs/ecma402-abstract` dependency we have for i18n.

src/property-filter/__tests__/property-filter-token-editor.test.tsx
```diff
@@ -361,7 +361,9 @@ describe.each([false, true] as const)('token editor labels, isMobile = %s', isMo
       expect(editor.propertyField().findControl()!.find('button')!.getElement()).toHaveAccessibleName(
         'Property string'
       );
-      expect(editor.operatorField().findControl()!.find('button')!.getElement()).toHaveAccessibleName('Operator =');
+      expect(editor.operatorField().findControl()!.find('button')!.getElement()).toHaveAccessibleName(
+        enableTokenGroups ? 'Operator =' : 'Operator = Equals'
+      );
```
just-boris, line 366: Previous version was not correct.

This is how `enableTokenGroups=true` look like

image

This is `enableTokenGroups=false`

image

The "Equals" string is indeed there and the test should assert that.

The issue has been exacerbated by our `i18nStrings` mock not providing the `operatorEqualsText` like it should

src/__a11y__/a11y-page-object.ts
```diff
@@ -57,7 +57,7 @@ export default class A11yPageObject extends BasePageObject {
 
       axe.configure(spec);
 
-      return axe.run(options, runOptions).then(
+      axe.run(options, runOptions).then(
```
just-boris, line 60: `this.browser.executeAsync` expects a callback, not a promise. So we should return nothing to follow the typings

src/table/__integ__/component-metrics.test.ts
```diff
@@ -12,7 +12,9 @@ declare const window: ExtendedWindow;
 
 class TableWithAnalyticsPageObject extends BasePageObject {
   async getComponentMetricsLog() {
-    const componentsLog = await this.browser.execute(() => window.__awsuiComponentlMetrics__);
+    const componentsLog: ExtendedWindow['__awsuiComponentlMetrics__'] = await this.browser.execute(
+      () => window.__awsuiComponentlMetrics__
+    );
```
just-boris, line 17: `this.browser.execute` is a polymorph function now: https://github.com/webdriverio/webdriverio/commit/bc5a7438cc8a402099ed9797adcc8ca4549012d9

Which does not play well with `__awsuiComponentlMetrics__` which is an `Array<any>`

src/property-filter/__tests__/property-filter-token-editor.test.tsx
jperals: Why was the `i18nStrings` mock not providing the strings correctly? Does this PR fix this? If so, how? If not, why was the assertion work before? Does `toHaveAccessibleName` match partial strings although the value passed here is not a regular expression?

src/property-filter/__tests__/common.tsx
```diff
@@ -74,6 +74,7 @@ export const providedI18nStrings = {
     'i18nStrings.editTokenHeader': 'Edit filter',
     'i18nStrings.propertyText': 'Property',
     'i18nStrings.operatorText': 'Operator',
+    'i18nStrings.operatorEqualsText': 'Equals',
```
just-boris, line 77: @jperals this is the i18n mock and it was missing the string

src/property-filter/__tests__/property-filter-token-editor.test.tsx
jperals: Oh, because of the missing string which was added in `src/property-filter/__tests__/common.tsx` 👍

src/__tests__/snapshot-tests/__snapshots__/documenter.test.ts.snap
```diff
@@ -5982,10 +5982,51 @@ For more info on custom languages, see the [Code editor API](/components/code-ed
           ""d"",
           ""r"",
           ""properties"",
-          ""slim"",
-          ""abap"",
           ""abc"",
           ""actionscript"",
+          ""c_cpp"",
```
jperals, line 5987: Why did these items become unsorted?

jperals: Although some elements at the start were already unsorted. So I guess there is already no strict criteria here, so not a blocker

just-boris: We do not sort union types at all. This is serialization result of ace-builds type definition. They changed something on their end – we got it reflected in our doc printout

src/property-filter/__tests__/property-filter-token-editor.test.tsx
just-boris: If you are still wondering why dependency upgrade changed this test, it is `dom-accessibility-api` package which jest-dom uses under the hood: https://github.com/testing-library/jest-dom/blob/918b6fbcde10d4409ee8f05c6e4eecbe96a72b7a/src/to-have-accessible-name.js#L1

I checked, different versions of  their `computeAccessibleName` returns different results for the same HTML of our components.

Since the new behavior actually better matches the real state, I take this as an improvement

PR #3611, opened 2025-06-25
fix: AnnotationContext popover position not recalculating onClick

ernst-dev:
### Description

Simple fix using window [capture](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#capture) to pick up on click events regardless of `stopPropagation` being set on buttons. This allows the [tool toggle button](https://github.com/cloudscape-design/components/blob/dev-v3-ernstka-annotation-context-fix/src/app-layout/visual-refresh-toolbar/toolbar/trigger-button/index.tsx#L98) to maintain its current behaviour as `stopPropagation` was added to it for its own reasons, but enables the popover to recalculate its position when a click event happens.

`AWSUI-60630`

### How has this been tested?

Manually verified

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/popover/container.tsx
+++ b/src/popover/container.tsx
@@ -162,7 +162,7 @@ export default function PopoverContainer({
     const updatePositionOnResize = () => requestAnimationFrame(() => updatePositionHandler());
     const refreshPosition = () => requestAnimationFrame(() => positionHandlerRef.current());
 
-    window.addEventListener('click', onClick, { signal: controller.signal });
+    window.addEventListener('click', onClick, { capture: true, signal: controller.signal });
     window.addEventListener('resize', updatePositionOnResize, { signal: controller.signal });
     window.addEventListener('scroll', refreshPosition, { capture: true, signal: controller.signal });
 


avinashbot (approved):

src/popover/container.tsx
```diff
@@ -111,6 +112,11 @@ export default function PopoverContainer({
     updatePositionHandler(true);
   });
 
+  // Create a debounced version of the position handler to avoid performance issues with the mutation observer
+  const debouncedPositionHandler = useDebounceCallback(() => {
+    requestAnimationFrame(() => positionHandlerRef.current());
```
just-boris: There is no corresponding `cancelAnimationFrame`, so there is a chance this handler gets called after component is already unmounted

src/popover/container.tsx
```diff
@@ -111,6 +112,26 @@ export default function PopoverContainer({
...
+  }, 50);
+
+  useEffect(() => {
+    // React to layout changes at the DOM level, such as flashbars being added and shifting content down
+    let observer: MutationObserver;
+
+    if (contentRef.current) {
+      observer = new MutationObserver(debouncedPositionHandler);
```
just-boris: Mutation observer has debouncing built-in, you do not need to do this

src/popover/container.tsx
```diff
@@ -111,6 +112,26 @@ export default function PopoverContainer({
...
+      observer.observe(contentRef.current.getRootNode(), {
```
just-boris: Does it really catch the issue? The tools button is very far outside the `contentRef` content

src/popover/container.tsx
```diff
@@ -111,58 +111,18 @@ export default function PopoverContainer({
...
 
-  // Recalculate position on DOM events.
-  useLayoutEffect(() => {
-    /*
-    This is a heuristic. Some layout changes are caused by user clicks (e.g. toggling the tools panel, submitting a form),
-    and by tracking the click event we can adapt the popover's position to the new layout.
-    */
+  // Recalculate position when the DOM changes.
+  usePopoverObserver(popoverRef, () => {
+    updatePositionHandler();
```
ernst-dev: Open question, should this be wrapped with a conditional to not update the position when `keepPosition` is `true`?

src/popover/container.tsx
```diff
@@ -111,58 +111,29 @@ export default function PopoverContainer({
...
+    const popoverOffset = popoverRef.current && getLogicalBoundingClientRect(popoverRef.current);
+    // istanbul ignore if - tested via integration tests
+    if (
+      !keepPosition ||
+      !popoverOffset ||
+      popoverOffset.insetBlockStart < 0 ||
+      popoverOffset.insetBlockEnd > window.innerHeight
+    ) {
+      return;
+    }
```
ernst-dev: Moved from old onClick function

src/popover/use-position-observer.ts
```diff
@@ -0,0 +1,31 @@
...
+
+interface Coords {
+  x?: number;
+  y?: number;
+}
+
+export default function usePopoverObserver(popoverRef: React.RefObject<HTMLElement>, callback: () => void) {
+  const lastPosition = useRef<Coords>();
+
+  useEffect(() => {
```
avinashbot, line 21: 1. Rather than a `let` and a `?.` operator in the cleanup function, you could have just simplified this by bailing early.

```ts
useEffect(() => {
  if (!popoverRef.current) {
    return;
  }

  const observer = new MutationObserver(() => {
```

2. `callback` is in the dependencies tree but it's not memoized in the caller, which means it'll disconnect and recreate a MutationObserver on every popover rerender. And since repositioning is a rerender (the code is a little complex now, but I'm assuming based on the fact that positioning the popover happens via a `useState`), this might lead to a silly infinite loop. I'd be okay with either wrapping the caller in a `useCallback` or using our slightly hacky `useStableCallback` util. Also, this way, since `popoverRef` is basically stable forever, you can just store state in a variable inside `useEffect` and skip the `useRef`.

3. Why measure `popoverRef.current.parentElement` rather than `popoverRef.current` directly?

avinashbot (changes requested):

src/popover/container.tsx
```diff
@@ -3,12 +3,12 @@
 import React, { useLayoutEffect, useRef } from 'react';
 import clsx from 'clsx';
 
-import { nodeContains } from '@cloudscape-design/component-toolkit/dom';
 import { getLogicalBoundingClientRect, useResizeObserver } from '@cloudscape-design/component-toolkit/internal';
 
 import { useVisualRefresh } from '../internal/hooks/use-visual-mode';
 import { InternalPosition, PopoverProps } from './interfaces';
 import usePopoverPosition from './use-popover-position.js';
+import usePopoverObserver from './use-position-observer';
```
just-boris: File name and hook name do not match

src/popover/container.tsx
```diff
@@ -111,58 +111,30 @@ export default function PopoverContainer({
...
-
-        if (typeof requestIdleCallback !== 'undefined') {
-          await new Promise(r => requestIdleCallback(r));
-        } else {
-          await new Promise(r => setTimeout(r, 50));
-        }
-      }
-    };
+  // Recalculate position on resize or scroll events.
+  useLayoutEffect(() => {
```
avinashbot, line 133: Just noticed it, and it's super optional, but we could probably change this from a `useLayoutEffect` to a regular `useEffect` now, right?

just-boris (commented):
Should there be an extra test?

avinashbot (approved):
Left a minor comment. Shame that `updatePositionHandler` wasn't as stable as it originally seemed and we have to resort to `useStableCallback`.

src/popover/__tests__/use-position-observer.test.tsx
```diff
@@ -0,0 +1,190 @@
...
+    });
+
+    // Mock triggerRef
+    mockBoundingClientRect = { x: 100, y: 100 };
+    mockTriggerRef = {
+      current: {
+        getBoundingClientRect: jest.fn().mockReturnValue(mockBoundingClientRect),
+        ownerDocument: document,
+      } as unknown as HTMLElement,
+    };
```
just-boris: With that much mocking happening in this setup – do these tests really reproduce the issue?

When I mentioned "this needs some tests" in my past comment, I assumed it would be an integration test

ernst-dev: True yes, I thought a set of unit tests might be nice for all the conditions, but it ended up bloating and needing a lot of mocks, happy to drop these in favour of more verbose integration tests, especially considering I had to add istanbul ignores for the callback anyway even though I added tests for container.

Will do a shift for this

just-boris (dismissed):

PR #3613, opened 2025-06-26
chore: Increase font size of h3 header description text

timogasda:
### Description
We are increasing the font size of the description text in Header to be consistent across all header levels.
Before, the description text for h3 was smaller, but now it's the same size for h1, h2, and h3.

Related links, issue #, if available: AWSUI-60895

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/header/styles.scss
+++ b/src/header/styles.scss
@@ -204,8 +204,9 @@
     }
   }
   &-variant-h3 {
-    @include styles.font-body-s;
+    @include styles.font-body-m;
     &:not(.refresh) {
+      @include styles.font-body-s;
       padding-block-end: awsui.$space-scaled-xxs;
     }
   }


src/header/styles.scss
```diff
@@ -204,8 +204,9 @@
     }
   }
   &-variant-h3 {
-    @include styles.font-body-s;
+    @include styles.font-body-m;
```
cansuaa, line 207: I see in variant-h2 we don't use this mixin but instead have a different design token, should that also use the mixin?

timogasda: The `fontHeaderH2DescriptionSize` token is used here because it's `font-body-s` in classic but `font-body-m` in refresh. So we still need this token to not break classic.

cansuaa: Isn't what you did for variant-h3 the same?

Nevertheless this is a minor comment, not a blocker so approving the PR.

cansuaa (approved):

src/header/styles.scss
timogasda: True, but in the h2 variant there's also the `$font-header-h2-description-line-height` token, so it's a bit more complicated. So I'd rather not touch that if I don't want any changes there.

PR #3619, opened 2025-06-26
fix: Remove Style API from Toggle Button

DaemonCahill:
This PR removes the Style API from the Toggle Button interface.

### DIFF
--- a/src/__tests__/snapshot-tests/__snapshots__/documenter.test.ts.snap
+++ b/src/__tests__/snapshot-tests/__snapshots__/documenter.test.ts.snap
@@ -20134,26 +20134,6 @@ It prevents users from clicking the button, but it can still be focused.",
       "type": "string",
     },
     {
-      "description": "Specifies an object of selectors and properties that are used to apply custom styles.
-
-- \`root.background\` {active, default, disabled, hover} - (Optional) Background for component states.
-- \`root.borderColor\` {active, default, disabled, hover} - (Optional) Border color for component states.
-- \`root.borderRadius\` (string) - (Optional) Border radius style.
-- \`root.borderWidth\` (string) - (Optional) Border width style.
-- \`root.color\` {active, default, disabled, hover} - (Optional) Color for component states.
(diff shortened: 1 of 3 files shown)


src/toggle-button/interfaces.ts
```diff
@@ -7,7 +7,7 @@ import { IconProps } from '../icon/interfaces';
 import { BaseComponentProps } from '../internal/base-component';
 import { NonCancelableEventHandler } from '../internal/events';
 
-export interface ToggleButtonProps extends BaseComponentProps, BaseButtonProps {
+export interface ToggleButtonProps extends BaseComponentProps, Omit<BaseButtonProps, 'style'> {
```
just-boris: How about instead of adding and then omitting, move the `style` prop declaration from the `BaseButtonProps` to `ButtonProps`

DaemonCahill: Makes sense. Updating.

src/__tests__/snapshot-tests/__snapshots__/documenter.test.ts.snap
```diff
@@ -5982,51 +5982,10 @@ For more info on custom languages, see the [Code editor API](/components/code-ed
           ""d"",
           ""r"",
           ""properties"",
-          ""abc"",
```
DaemonCahill: @just-boris Any idea what this stuff is? The documenter kicked it out after I moved the style prop to ButtonProps.

just-boris: I upgraded dependencies today: https://github.com/cloudscape-design/components/pull/3610

You need to run `npm install && npm run build` and then update the snapshots. They will be back to expected

DaemonCahill: Okay!

just-boris (dismissed):

YueyingLu (approved):

PR #3623, opened 2025-06-27
feat: make drawer content take up full height

pranshuteotia:
### Description

Currently the drawer content does not expand to the full height of its container. Because of this if the drawer content has an embeded iFrame then it results in a broken experience. This PR fixes that.

### How has this been tested?

dry-run

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/__tests__/snapshot-tests/__snapshots__/documenter.test.ts.snap
+++ b/src/__tests__/snapshot-tests/__snapshots__/documenter.test.ts.snap
@@ -3576,6 +3576,28 @@ use the \`id\` attribute, consider setting it on a parent element instead.",
       "type": "string",
     },
     {
+      "description": "Attributes to add to the native element.
+Some attributes will be automatically combined with internal attribute values:
+- \`className\` will be appended.
+- Event handlers will be chained, unless the default is prevented.
+
+We do not support using this attribute to apply custom styling.",
+      "inlineType": {
+        "name": "Omit<React.HTMLAttributes<HTMLElement>, "children"> & Record<\`data-\${string}\`, string>",
+        "type": "union",
+        "values": [
+          "Omit<React.HTMLAttributes<HTMLElement>, "children">",
+          "Record<\`data-\${string}\`, string>",
+        ],
+      },
+      "name": "nativeAttributes",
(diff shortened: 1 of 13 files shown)


src/drawer/styles.scss
```diff
@@ -75,4 +78,5 @@
 
 .test-utils-drawer-content {
   /* used in test-utils */
+  flex: 1;
```
georgylobko, line 81: The purpose of this class is to access the dom node by this selector, so no style rules should be used here

src/drawer/styles.scss
```diff
@@ -16,6 +16,9 @@
 .drawer {
   @include styles.styles-reset;
   word-wrap: break-word;
+  height: 100%;
+  display: flex;
```
georgylobko, line 20: What is the purpose of this additional flex styles? I tried to apply `height: 100%` without them on my local machine - works fine

pranshuteotia: We not only need the drawer to be full height. We also need the drawer content to take up the rest of the space left after the drawer header. So display flex allows use to add `flex-grow: 1` to the content which will make the drawer content take up the rest of the space.
Screenshot 2025-06-30 at 3 50 39 PM

Now after adding the the flex styles the content is rendering properly.
Screenshot 2025-06-30 at 3 53 54 PM

src/drawer/styles.scss
pranshuteotia: Ack. I'll move this into another class

georgylobko: btw, this style rule can only be applied when the parent element is also `display: flex`, which is not a case for the runtime drawers

pranshuteotia: I am not sure if I understand? I do have display flex added to the parent which is the drawer component itself

georgylobko: got it. then this behavior is specific for your use-case, not for this general component. in this case it's better to implement it on your parent level

just-boris:
Closing for housecleaning. Feel free to ask for reopen, if the issue is still relevant

PR #3630, opened 2025-06-30
fix: Added Flashbar style support for focus ring and dismiss button variants

DaemonCahill:
This PR adds support for the Flashbar style API to modify focus rings and dismiss button styles based on the Flashbar item variant.

### DIFF
--- a/src/__tests__/snapshot-tests/__snapshots__/documenter.test.ts.snap
+++ b/src/__tests__/snapshot-tests/__snapshots__/documenter.test.ts.snap
@@ -9485,18 +9485,23 @@ If the \`action\` property is set, this property is ignored. **Deprecated**, rep
 - \`item.root.borderRadius\` (string) - (Optional) Item border radius.
 - \`item.root.borderWidth\` (string) - (Optional) Item border width.
 - \`item.root.color\` {error, info, inProgress, success, warning} - (Optional) Color for item types.
-- \`item.root.focusRing.borderColor\` (string) - (Optional) Item focus ring border color.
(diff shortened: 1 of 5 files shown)


src/flashbar/style.tsx
```diff
@@ -0,0 +1,144 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import customCssProps from '../internal/generated/custom-css-properties';
+import { FlashbarProps } from './interfaces';
+
+export function getCollapsibleFlashStyles(style: FlashbarProps.Style, type: string = 'info') {
+  const background =
+    style?.item?.root?.background &&
+    (type === 'in-progress'
+      ? style?.item?.root?.background.inProgress
```
gethinwebster, line 10: Seeing this, I'm wondering about having a rule that style API key names must exactly match associated states/variants. It means having to quote some key values when using the API, but I think that's going to be more understandable overall, and reduce a whole load of conditional logic in our implementation.

Alternatively, we could have a rule that style API key names are always camelCase, and a helper function to convert. But having individual conditions scattered all over our code like this isn't scalable or maintainable, it needs to be standardized and/or centralized.

DaemonCahill: 100% agree this doesn't scale. I'm going to do some refactoring and abstracting helpers as we continue development. But I definitely think we should keep the entire API camelCase.

gethinwebster (approved):

PR #3635, opened 2025-07-08
feat: Show resizer for last table column

Al-Dani:
### Description

We've received a report that hovering over last column's resizer causes layout shifts in Firefox & Safari.
While we could fix that, the question is: why do we have a non-visible interactive element in the first place? It's there, it receives focus and shows focus ring, but you need to know it exists to interact with it.

So, let's make it simple 

Related links, issue #, if available: AWSUI-60946

Before (with focus ring):
image

After:
image

### How has this been tested?

- ✅ RUN through my dev pipeline to check screenshots 
- checked with visual design

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/table/resizer/index.tsx
+++ b/src/table/resizer/index.tsx
@@ -231,7 +231,7 @@ export function Resizer({
         data-focus-id={focusId}
       />
       <span
-        className={clsx(styles.divider, isDragging && styles['divider-active'])}
+        className={clsx(styles['divider-interactive'], isDragging && styles['divider-active'])}
         data-awsui-table-suppress-navigation={true}
         ref={resizerSeparatorRef}
         id={separatorId}
--- a/src/table/resizer/styles.scss
+++ b/src/table/resizer/styles.scss
@@ -16,7 +16,8 @@
 $handle-width: awsui.$space-xl;
 $active-separator-width: 2px;
 
-th:not(:last-child) > .divider {
+th:not(:last-child) > .divider,
+.divider-interactive {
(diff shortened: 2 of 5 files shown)


src/table/resizer/styles.scss
```diff
@@ -16,7 +16,8 @@
 $handle-width: awsui.$space-xl;
 $active-separator-width: 2px;
 
-th:not(:last-child) > .divider {
+th:not(:last-child) > .divider,
```
mariksav, line 19: Wouldn't it be sufficient to remove the `th:not(:last-child)` and keep the rest as it was, because the rule didn't match because it was the icon in the last column? And then the next line becomes obsolete as well.

src/table/resizer/index.tsx
```diff
@@ -231,7 +231,7 @@ export function Resizer({
         data-focus-id={focusId}
       />
       <span
-        className={clsx(styles.divider, isDragging && styles['divider-active'])}
+        className={clsx(styles['divider-interactive'], isDragging && styles['divider-active'])}
```
mariksav, line 234: This line can stay as it initially was if we apply the changes from the comment beneath.

src/table/resizer/styles.scss
Al-Dani: No, because then the divider would be present for non-resizable columns too, which looks awkward.

mariksav (approved):
Ok good to know, everything else looks good :)

PR #3636, opened 2025-07-09
chore: Make disabled selected multiselect tokens readonly

gethinwebster:
### Description

Changes disabled multiselect items so that they get rendered as `readOnly` tokens for accessibility purposes.

Related links, issue #, if available: AWSUI-60953

### How has this been tested?

New unit test

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/multiselect/internal.tsx
+++ b/src/multiselect/internal.tsx
@@ -31,6 +31,8 @@ type InternalMultiselectProps = SomeRequired<
 > &
   InternalBaseComponentProps;
 
+type ExtendedToken = TokenGroupProps.Item & { _readOnly: boolean };
+
 const InternalMultiselect = React.forwardRef(
   (
     {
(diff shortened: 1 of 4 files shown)


src/multiselect/internal.tsx
```diff
@@ -200,6 +203,7 @@ const InternalMultiselect = React.forwardRef(
             limitShowFewerAriaLabel={tokenLimitShowFewerAriaLabel}
             disableOuterPadding={true}
             readOnly={readOnly}
+            isItemReadOnly={item => (item as ExtendedToken)._readOnly}
```
Al-Dani, line 206: why not having `readOnly` a per-item property for a token group instead? Similar to disabled. I understand it would be a private property for now, but it feels like something that's useful for customers, so we could expose that officially later

gethinwebster: because it's hard to make those actually private without doing lots of sanitization/mapping.

Alternatively, I guess we could just expose it publicly...

Al-Dani: Ah, riiiight
To me, it makes sense to make them public. Current global readonly is less useful IMO. Not sure how it aligns with the new `Token`, but it should be a low-effort addition anyway.

No hard pushback on the current approach, though :)

Al-Dani (approved):

PR #3644, opened 2025-07-11
chore: Disable error overlay

just-boris:
### Description

Hide overlay, because it prints not real errors

image

### How has this been tested?

The error does not get printed anymore

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/webpack.config.base.cjs
+++ b/pages/webpack.config.base.cjs
@@ -136,6 +136,13 @@ module.exports = ({
       ...(moduleReplacements || []).map(({ from, to }) => replaceModule(from, to)),
     ],
     devServer: {
+      client: {
+        overlay: {
+          errors: true,
+          warnings: true,
+          runtimeErrors: false,
+        },
+      },
       devMiddleware: {
         publicPath: '/',
       },


pages/webpack.config.base.cjs
```diff
@@ -136,6 +136,9 @@ module.exports = ({
       ...(moduleReplacements || []).map(({ from, to }) => replaceModule(from, to)),
     ],
     devServer: {
+      client: {
```
georgylobko, line 139: here we can disable only runtime errors, but leave the remaining ones as is

```
client: {
        overlay: {
          errors: true,
          warnings: true,
          runtimeErrors: false,
        },
      },
```

just-boris: I would like to remove all of them, because these are duplicates of the warnings I already see in the terminal

Do you have a strong opinion to keep them here?

georgylobko: From my perspective we should not suppress errors until we have a strong reason to do so, like this one.

just-boris: Okay, changed

teodoranemes (approved):

PR #3646, opened 2025-07-14
chore: Introduce focus behavior when flash items gets dismissed

taheramr:
### Description

This introduces focus logic whenever a flash item gets dismissed, it works in both the collapsible and non-collapsible versions of the flashbar.

The idea works as the following:
- for each flash item we attach a ref
- when a flash item gets dismissed, we call a onDissmied callback that figures out the next flash item to focus and focues it
- if no flash items are remianing, we focus the main element (or an element with role="main")
- we prioritise focusing the dismiss button in case a flash is dismissed, otherwise we focus the focus invisible container of the flash item

AWSUI-60902

### How has this been tested?
Added integration tests to it

Tested it locally as well in the flashbar interactive page

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/flashbar/collapsible-flashbar.tsx
+++ b/src/flashbar/collapsible-flashbar.tsx
@@ -42,6 +42,16 @@ export default function CollapsibleFlashbar({ items, style, ...restProps }: Flas
   const [exitingItems, setExitingItems] = useState<ReadonlyArray<FlashbarProps.MessageDefinition>>([]);
   const [isFlashbarStackExpanded, setIsFlashbarStackExpanded] = useState(false);
 
+  const getElementsToAnimate = useCallback(() => {
+    const flashElements = isFlashbarStackExpanded ? expandedItemRefs.current : collapsedItemRefs.current;
+    return { ...flashElements, notificationBar: notificationBarRef.current };
+  }, [isFlashbarStackExpanded]);
+
+  const prepareAnimations = useCallback(() => {
+    const rects = getDOMRects(getElementsToAnimate());
+    setInitialAnimationState(rects);
+  }, [getElementsToAnimate]);
+
   const collapsedItemRefs = useRef<Record<string, HTMLElement | null>>({});
   const expandedItemRefs = useRef<Record<string, HTMLElement | null>>({});
   const [initialAnimationState, setInitialAnimationState] = useState<Record<string, DOMRect> | null>(null);
@@ -50,9 +60,6 @@ export default function CollapsibleFlashbar({ items, style, ...restProps }: Flas
   const [transitioning, setTransitioning] = useState(false);
   const flashbarElementId = useUniqueId('flashbar');
   const itemCountElementId = useUniqueId('item-count');
-
-  // Create a ref to hold the prepareAnimations function
-  const prepareAnimationsRef = useRef<(() => void) | null>(null);
 
   const { baseProps, isReducedMotion, isVisualRefresh, mergedRef, ref, flashRefs, handleFlashDismissed } = useFlashbar({
     items,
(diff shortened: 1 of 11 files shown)


src/flashbar/common.tsx
```diff
@@ -68,12 +71,61 @@ export function useFlashbar({
     }
   }, [nextFocusId, ref]);
 
+  const handleFlashDismissed = (dismissedId?: string) => {
```
taheramr, line 127: initially I created this function inside each version (collapsible and non-collapsible). But it seems like having it centralised here is best since also they both (technicaly) share the same logic

src/flashbar/common.tsx
```diff
@@ -68,12 +71,61 @@ export function useFlashbar({
...
+      if (!nextFlashElement) {
+        // If the next flash element is not available, it might be because the flashbar is collapsed
+        // In that case, try to focus on the notification bar button or the main element
+        const notificationBarButton = ref.current?.querySelector(`.${styles.button}`);
+        if (notificationBarButton) {
+          (notificationBarButton as HTMLElement).focus();
+          return;
+        }
+
+        const mainElement = document.querySelector('main') ?? document.querySelector('[role="main"]');
```
gethinwebster: This 1) seems quite a big assumption make (I'm not sure we have similar logic in any other components?) and 2) is duplicated in two places. At the very least we should only have this logic once, but I'm also not sure about having it at all

src/flashbar/collapsible-flashbar.tsx
```diff
@@ -284,10 +290,26 @@ export default function CollapsibleFlashbar({ items, style, ...restProps }: Flas
                       isVisualRefresh && styles['flash-refresh']
                     )}
                     key={getItemId(item)}
-                    ref={shouldUseStandardAnimation(item, index) ? transitionRootElement : undefined}
+                    ref={(el: HTMLDivElement | null) => {
```
gethinwebster, line 286: can you use `useMergeRefs` here instead of doing it manually? https://github.com/cloudscape-design/component-toolkit/blob/837fb68b01bfd6cf6f1b0c8264bd93f0eb189369/src/internal/use-merge-refs/index.tsx#L13

src/flashbar/collapsible-flashbar.tsx
```diff
@@ -42,17 +42,19 @@ export default function CollapsibleFlashbar({ items, style, ...restProps }: Flas
...
+  const notificationBarRef = useRef<HTMLDivElement | null>(null);
+  const [transitioning, setTransitioning] = useState(false);
+  const flashbarElementId = useUniqueId('flashbar');
+  const itemCountElementId = useUniqueId('item-count');
 
-  const prepareAnimations = useCallback(() => {
-    const rects = getDOMRects(getElementsToAnimate());
-    setInitialAnimationState(rects);
-  }, [getElementsToAnimate]);
+  // Create a ref to hold the prepareAnimations function
```
gethinwebster: why did this need to change?

src/flashbar/collapsible-flashbar.tsx
taheramr: I tried doing this a bit but this part is wrapped around a function called `renderList()`, so if I used the `useMergeRefs` rules of react eslint will get us an error casue I'm calling a hook inside a callback.

I tried moving around the code but it introduces more changes than it should, let me know wdyt or if you have suggestions that I miss here

src/flashbar/collapsible-flashbar.tsx
taheramr: I think I was overcomplicating this to make sure the next item in the collapsible flashbar is focused (opened or closed).

I removed the logic as it was already taken care of in the common.tsx with the setTimeout and I tested it.

src/flashbar/common.tsx
taheramr: The reasoning behind this is to try to focus the next most important item that has content which is usually the `main` html element. Another data point, if using the AppLayout it by defaults wraps the content section with a main element, so it makes sense to focus it.

YueyingLu: Agree on Gethin's comment. Why not focus last focus element before focusing on flashbar, if there is a11y guidance on it? What about components with similar dismiss behavior like Alert, should be consistent?

src/flashbar/common.tsx
```diff
@@ -68,12 +71,61 @@ export function useFlashbar({
...
+
+    const nextItemId = items[nextItemIndex].id ?? nextItemIndex;
+
+    // Try to focus on the next item, but with a small delay to ensure the DOM is updated
+    // This is especially important for collapsible flashbars where the next item might become visible
+    const attemptFocus = () => {
+      const nextFlashElement = flashRefs.current[nextItemId];
+      if (!nextFlashElement) {
+        // If the next flash element is not available, it might be because the flashbar is collapsed
+        // In that case, try to focus on the notification bar button or the main element
```
YueyingLu: Is the notification bar button not the "next flashbar item" button?

taheramr: yes if it exits using the
```
      const nextFlashElement = flashRefs.current[nextItemId];
```

YueyingLu (approved):

PR #3655, opened 2025-07-16
feat: Allow custom primary actions in prompt input

gethinwebster:
### Description

Allow custom primary actions. For example, to replace the default single action with multiple action buttons.

Related links, issue #, if available: AWSUI-60922

### How has this been tested?

Updated unit tests and permutations

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/prompt-input/permutations.page.tsx
+++ b/pages/prompt-input/permutations.page.tsx
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 import React from 'react';
 
+import { Button } from '~components';
 import PromptInput, { PromptInputProps } from '~components/prompt-input';
 
 import img from '../icon/custom-icon.png';
@@ -76,7 +77,10 @@ const permutations = createPermutations<PromptInputProps>([
   {
     value: ['Short value for custom primary actions'],
     actionButtonIconName: [undefined, 'send'],
-    customPrimaryAction: [undefined, 'Custom primary'],
+    customPrimaryAction: [
+      undefined,
+      <Button variant="icon" iconName="add-plus" ariaLabel="Custom action" key="button" />,
+    ],
     secondaryActions: [undefined, 'secondary actions'],
     disableSecondaryActionsPaddings: [false, true],
   },
--- a/pages/prompt-input/permutations.page.tsx
+++ b/pages/prompt-input/permutations.page.tsx
@@ -76,7 +76,7 @@ const permutations = createPermutations<PromptInputProps>([
   {
     value: ['Short value for custom primary actions'],
     actionButtonIconName: [undefined, 'send'],
-    primaryAction: [undefined, 'Custom primary'],
+    customPrimaryAction: [undefined, 'Custom primary'],
     secondaryActions: [undefined, 'secondary actions'],
     disableSecondaryActionsPaddings: [false, true],
   },
--- a/pages/prompt-input/permutations.page.tsx
+++ b/pages/prompt-input/permutations.page.tsx
@@ -73,6 +73,13 @@ const permutations = createPermutations<PromptInputProps>([
     disableSecondaryActionsPaddings: [false, true],
     disableSecondaryContentPaddings: [false, true],
   },
+  {
+    value: ['Short value for custom primary actions'],
+    actionButtonIconName: [undefined, 'send'],
+    primaryAction: [undefined, 'Custom primary'],
+    secondaryActions: [undefined, 'secondary actions'],
+    disableSecondaryActionsPaddings: [false, true],
+  },
 ]);
 
 export default function PromptInputPermutations() {
--- a/pages/prompt-input/simple.page.tsx
+++ b/pages/prompt-input/simple.page.tsx
@@ -205,27 +205,25 @@ export default function PromptInputPage() {
                   disableSecondaryActionsPaddings={true}
                   customPrimaryAction={
                     hasPrimaryActions ? (
-                      <Box padding={{ right: 'xxs', top: 'xs' }}>
-                        <ButtonGroup
-                          variant="icon"
-                          items={[
-                            {
-                              type: 'icon-button',
-                              id: 'record',
-                              text: 'Record',
-                              iconName: 'microphone',
-                              disabled: isDisabled || isReadOnly,
-                            },
-                            {
-                              type: 'icon-button',
-                              id: 'submit',
-                              text: 'Submit',
-                              iconName: 'send',
-                              disabled: isDisabled || isReadOnly,
-                            },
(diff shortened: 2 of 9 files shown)


src/test-utils/dom/prompt-input/index.ts
```diff
@@ -13,16 +13,27 @@ export default class PromptInputWrapper extends ComponentWrapper {
...
 
+  /**
+   * Finds the primary action button. Note that if you provide `primaryAction`
+   * then this will return null, you should use `findPrimaryActionSlot` instead.
+   */
   findActionButton(): ElementWrapper<HTMLButtonElement> {
     return this.findByClassName<HTMLButtonElement>(testutilStyles['action-button'])!;
   }
 
+  /**
```
gethinwebster, line 23: unfortunately we got a lot of these typings wrong on the first release :( and teams are already using them so we can't easily add `| null`

src/test-utils/dom/prompt-input/index.ts
```diff
@@ -13,16 +13,27 @@ export default class PromptInputWrapper extends ComponentWrapper {
...
+   * Finds the secondary actions slot. Note that, despite its typings, this may return null.
+   */
   findSecondaryActions(): ElementWrapper<HTMLDivElement> {
     return this.findByClassName<HTMLDivElement>(testutilStyles['secondary-actions'])!;
   }
 
-  findSecondaryContent(): ElementWrapper<HTMLDivElement> {
-    return this.findByClassName<HTMLDivElement>(testutilStyles['secondary-content'])!;
+  findSecondaryContent(): ElementWrapper<HTMLDivElement> | null {
```
gethinwebster, line 30: couldn't find any teams using this specific selector, and checked with dry run, so it should be safe to make this change

pages/prompt-input/permutations.page.tsx
```diff
@@ -73,6 +73,13 @@ const permutations = createPermutations<PromptInputProps>([
     disableSecondaryActionsPaddings: [false, true],
     disableSecondaryContentPaddings: [false, true],
   },
+  {
+    value: ['Short value for custom primary actions'],
+    actionButtonIconName: [undefined, 'send'],
+    customPrimaryAction: [undefined, 'Custom primary'],
```
pan-kot: Nit: I would render an actual button here, not text.

Image

pages/prompt-input/simple.page.tsx
```diff
@@ -191,6 +203,31 @@ export default function PromptInputPage() {
                   warning={hasWarning}
                   ref={ref}
                   disableSecondaryActionsPaddings={true}
+                  customPrimaryAction={
+                    hasPrimaryActions ? (
+                      <Box padding={{ right: 'xxs', top: 'xs' }}>
+                        <ButtonGroup
```
pan-kot: The focus outline for custom primary actions looks different compared to the default one. It is in a close proximity to the area's border. Is that expected?

Image

pan-kot: nit: the custom actions do not have any associated effect, while the original action does trigger an alert

src/prompt-input/internal.tsx
```diff
@@ -173,19 +174,21 @@ const InternalPromptInput = React.forwardRef(
...
-          onClick={() => fireNonCancelableEvent(onAction, { value })}
-          variant="icon"
-        />
+      <div className={clsx(styles['primary-action'], testutilStyles['primary-action'])}>
+        {customPrimaryAction ?? (
+          <InternalButton
+            className={clsx(styles['action-button'], testutilStyles['action-button'])}
+            ariaLabel={actionButtonAriaLabel}
+            disabled={disabled || readOnly || disableActionButton}
+            __focusable={readOnly}
```
pan-kot, line 183: What is the purpose of __focusable - can this be supported for custom actions?

src/test-utils/dom/prompt-input/index.ts
```diff
@@ -13,16 +13,26 @@ export default class PromptInputWrapper extends ComponentWrapper {
...
+    return this.findByClassName<HTMLDivElement>(testutilStyles['secondary-content']);
+  }
+
+  findPrimaryActionSlot(): ElementWrapper<HTMLDivElement> | null {
```
pan-kot: Should we instead use `findCustomPrimaryAction(): ElementWrapper`? This way it should be easier to comprehend the purpose of the method. Otherwise, I would recommend adding api-docs to describe it.

pages/prompt-input/simple.page.tsx
gethinwebster: this is expected: the styling of the custom actions matches that of secondary actions

src/test-utils/dom/prompt-input/index.ts
gethinwebster: updated

pages/prompt-input/simple.page.tsx
gethinwebster: (however, updated styling to make it simpler and somewhat more consistent)

pages/prompt-input/permutations.page.tsx
gethinwebster: done

src/prompt-input/internal.tsx
gethinwebster: this is for focusable-disabled buttons (when input is read-only). This could be achieved using nativeAttributes for custom actions.

src/test-utils/dom/prompt-input/index.ts
```diff
@@ -13,16 +13,26 @@ export default class PromptInputWrapper extends ComponentWrapper {
...
     return this.findByClassName<HTMLDivElement>(testutilStyles['secondary-actions'])!;
   }
 
-  findSecondaryContent(): ElementWrapper<HTMLDivElement> {
-    return this.findByClassName<HTMLDivElement>(testutilStyles['secondary-content'])!;
+  findSecondaryContent(): ElementWrapper<HTMLDivElement> | null {
+    return this.findByClassName<HTMLDivElement>(testutilStyles['secondary-content']);
+  }
+
+  findCustomPrimaryAction(): ElementWrapper<HTMLDivElement> | null {
```
YueyingLu, line 34: `customPrimaryAction` is core only, should this util core only too?

YueyingLu: Or should `findActionButton` return either primary action or the custom action?

gethinwebster: that would require changing the typing of `findActionButton` (currently it has `HTMLButtonElement`) which would be a breaking change

PR #3667, opened 2025-07-18
fix: Add backward compatibility for expandable table test util

Al-Dani:
### Description

in #3477 we moved styles file that was used in test utils, which led to regenerated hashes, which led to customers' integration test failing because of mismatched versions.

We discussed that using both old and new selector could be a solution if we absolutely have to change the selector.
Because there is no "old" selector anymore (we moved the file), I just copied the exact class name from the snapshot. Messy, but..

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/test-utils/dom/table/index.ts
+++ b/src/test-utils/dom/table/index.ts
@@ -178,7 +178,7 @@ export default class TableWrapper extends ComponentWrapper {
    */
   findExpandToggle(rowIndex: number): ElementWrapper | null {
     // the file was moved, which changed the hash, so we cannot use old test class anymore.
-    const oldSelector = `tbody tr:nth-child(${rowIndex}) .awsui_expand-toggle_1ss49`;
+    const oldSelector = `tbody tr:nth-child(${rowIndex}) .awsui_expand-toggle_1ss49_1w02f_153`;
     const newSelector = `tbody tr:nth-child(${rowIndex}) .${expandToggleStyles['expand-toggle']}`;
     return this.findNativeTable().findAny(oldSelector, newSelector);
   }
--- a/src/test-utils/dom/table/index.ts
+++ b/src/test-utils/dom/table/index.ts
@@ -177,7 +177,10 @@ export default class TableWrapper extends ComponentWrapper {
    * @param rowIndex 1-based index of the row.
    */
   findExpandToggle(rowIndex: number): ElementWrapper | null {
-    return this.findNativeTable().find(`tbody tr:nth-child(${rowIndex}) .${expandToggleStyles['expand-toggle']}`);
+    // the file was moved, which changed the hash, so we cannot use old test class anymore.
+    const oldSelector = `tbody tr:nth-child(${rowIndex}) .awsui_expand-toggle_1ss49`;
+    const newSelector = `tbody tr:nth-child(${rowIndex}) .${expandToggleStyles['expand-toggle']}`;
+    return this.findNativeTable().findAny(oldSelector, newSelector);
   }
 
   /**


src/test-utils/dom/table/index.ts
```diff
@@ -177,7 +177,10 @@ export default class TableWrapper extends ComponentWrapper {
    * @param rowIndex 1-based index of the row.
    */
   findExpandToggle(rowIndex: number): ElementWrapper | null {
-    return this.findNativeTable().find(`tbody tr:nth-child(${rowIndex}) .${expandToggleStyles['expand-toggle']}`);
+    // the file was moved, which changed the hash, so we cannot use old test class anymore.
+    const oldSelector = `tbody tr:nth-child(${rowIndex}) .awsui_expand-toggle_1ss49`;
+    const newSelector = `tbody tr:nth-child(${rowIndex}) .${expandToggleStyles['expand-toggle']}`;
+    return this.findNativeTable().find(`:is(${oldSelector},${newSelector})`);
```
just-boris: there is `this.findAny` test util for this use-case: https://github.com/cloudscape-design/test-utils/blob/f29b4a82c71b27d39eb8c7bd4059de5e49c86ccb/src/core/selectors.ts#L38-L40

Same `:is()` under the hood, but cleaner to read

michaeldowseza (dismissed):

src/test-utils/dom/table/index.ts
```diff
@@ -177,7 +177,10 @@ export default class TableWrapper extends ComponentWrapper {
    * @param rowIndex 1-based index of the row.
    */
   findExpandToggle(rowIndex: number): ElementWrapper | null {
-    return this.findNativeTable().find(`tbody tr:nth-child(${rowIndex}) .${expandToggleStyles['expand-toggle']}`);
+    // the file was moved, which changed the hash, so we cannot use old test class anymore.
+    const oldSelector = `tbody tr:nth-child(${rowIndex}) .awsui_expand-toggle_1ss49`;
```
just-boris: Sorry for not capturing it earlier, but this old class name does not work. The real class name has more hashes, for example: `.awsui_expand-toggle_1xe88_1yndx_153`

Al-Dani: thanks, updated the class with the full hash (copied class from the built for the commit right before #3477)

Al-Dani:
these 30 changed files are coming from merge from main, my bad.

the meaningful commit is: https://github.com/cloudscape-design/components/pull/3667/commits/e54c0653870a7efac5fd5da17e5eadad7da9a6fd

just-boris:
> these 30 changed files are coming from merge from main, my bad.

Made a rebase in Github UI, it renders the diff correctly now

src/test-utils/dom/table/index.ts
just-boris: Still I recommend using this form `[class*=awsui_expand-toggle_1xe88]` because it allows compatibility with all previous versions, not just the very previous change in this file

Al-Dani: my understanding is that `findAny` [util](https://github.com/cloudscape-design/test-utils/blob/main/src/core/selectors.ts#L38) would call `getUnscopedClassName` that'll replace the classname in the selector to `[class*=xxxxxx]` thanks to [`trimContentHash`](https://github.com/cloudscape-design/test-utils/blob/main/src/core/utils.ts#L57) and the fact that the class has all hash parts now

just-boris: You are right, I forgot how my own code works 🤦 

approved already

PR #3674, opened 2025-07-22
fix: Harmonize drawer headers' letter spacing with other headers

jperals:
### Description

The letter spacing of headers of drawers and split panels are not consistent with our header styles, since they are styled like an h3 header except for the letter spacing.

(And we also save a few pixels).

As an example, the screenshot below shows a split panel header with a Box with `variant="h3"` and `margin="n"` in the split panel content.

Screenshot 2025-07-22 at 08 38 24

This change is confirmed with Visual Design.

### How has this been tested?

Tested in my pipeline, visual diffs are as expected.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/top-navigation/styles.scss
+++ b/src/top-navigation/styles.scss
@@ -230,23 +230,24 @@
   border-block-end: awsui.$border-divider-section-width solid awsui.$color-border-divider-default;
 
   &-text {
-    @include styles.font-panel-header;
     color: awsui.$color-text-heading-default;
     flex: 1;
     margin-block: 0;
     margin-inline: 0;
     text-align: center;
 
+    &--title {
+      /* used in test-utils */
+      @include styles.font-panel-header;
+    }
+
     &--secondary {
       font-size: awsui.$font-header-h2-description-size;
       line-height: awsui.$font-header-h2-description-line-height;
       font-weight: styles.$font-weight-normal;
+      @include styles.font-smoothing;
     }
   }
-}
-
-.overflow-menu-header-text--title {
-  /* used in test-utils */
 }
 
 .overflow-menu-back-button {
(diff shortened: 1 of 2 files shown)


src/top-navigation/styles.scss
```diff
@@ -230,25 +230,26 @@
...
+    &--title {
+      /* used in test-utils */
+      @include styles.font-panel-header;
+    }
+
     &--secondary {
       font-size: awsui.$font-header-h2-description-size;
       line-height: awsui.$font-header-h2-description-line-height;
       font-weight: styles.$font-weight-normal;
+      @include styles.font-smoothing;
```
jperals, line 248: This is the only rule that `.overflow-menu-header-text--secondary` really uses from `styles.font-panel-header`, so we add it here explicitly instead of including `styles.font-panel-header` in the parent selector and overriding all its other rules in this child selector.

src/top-navigation/styles.scss
```diff
@@ -230,25 +230,26 @@
...
   &-text {
-    @include styles.font-panel-header;
     color: awsui.$color-text-heading-default;
     flex: 1;
     margin-block: 0;
     margin-inline: 0;
     text-align: center;
 
+    &--title {
+      /* used in test-utils */
```
just-boris, line 240: Originally we put these comments to prevent SASS removing empty rules during compilation. 

In this case, it is not really needed, but you can keep it if it helps with readaibility

just-boris (approved):

src/top-navigation/styles.scss
jperals: Yes, I decided to keep here because it seems like we don't have good coverage for the test util that uses this selector, so I want to make sure it is not removed in future refactors.

PR #3680, opened 2025-07-23
feat: resizeDrawer method for plugin api

georgylobko:
### Description

This PR introduces a new method `resizeDrawer(drawerId: string, size: number)` for Cloudscape plugin api to programmatically resize drawers declared within `AppLayout` component.

**Usage example:**

```javascript
awsuiPlugins.appLayout.resizeDrawer('circle-global', 400)
```

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/__integ__/runtime-drawers.test.ts
+++ b/src/app-layout/__integ__/runtime-drawers.test.ts
@@ -148,6 +148,12 @@ describe('Visual refresh toolbar only', () => {
     hasHorizontalScroll() {
       return this.browser.execute(() => document.body.scrollWidth - document.body.clientWidth > 0);
     }
+
+    async getDrawerWidth(drawerId: string) {
+      const { width } = await this.getBoundingBox(findDrawerById(wrapper, drawerId)!.toSelector());
+
+      return width;
+    }
   }
   function setupTest(testFn: (page: PageObject) => Promise<void>) {
     return useBrowser(async browser => {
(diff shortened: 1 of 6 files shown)


src/app-layout/__integ__/runtime-drawers.test.ts
```diff
@@ -362,4 +368,26 @@ describe('Visual refresh toolbar only', () => {
...
+    'should programmatically resize drawers',
+    setupTest(async page => {
+      await page.setWindowSize(viewports.desktopWide);
+      const drawerId1 = 'circle-global';
+      const drawerId2 = 'circle3-global';
+      await page.click(wrapper.findDrawerTriggerById(drawerId1).toSelector());
+      await page.click(wrapper.findDrawerTriggerById(drawerId2).toSelector());
+
+      await expect(page.getDrawerWidth(drawerId1)).resolves.toBe(351);
+      await expect(page.getDrawerWidth(drawerId2)).resolves.toBe(321);
```
just-boris: See the warnings about flaky tests above in this PR.

It seems like this type of assertion is unreliable.

We should use `page.waitForAssertion()`

(this one could be a follow up PR)

src/internal/plugins/controllers/drawers.ts
```diff
@@ -185,6 +189,25 @@ export class DrawersController {
...
+
+    this.drawerResizeListener = listener;
+
+    return () => {
+      this.drawerResizeListener = null;
+    };
+  };
+
+  resizeDrawer = (drawerId: string, size: number) => {
+    this.drawerResizeListener?.(drawerId, size);
```
just-boris, line 208: I recall some feedback from consumers of our past API that they would like to know if this call found an app layout or not.

What do you think about returning a `{ succeeded: true | false }` object to help them?

(this one we need to decide before merging to avoid two variants of this API)

georgylobko: What is the use case for this parameter? Making sure that the version of the app layout has this method handler and it has been triggered? If so, why we didn't return this object remaining methods?

just-boris: Yes, making sure there is a compatible version 

> why we didn't return this object remaining methods

because we did not know it is useful. Might add it to all of them as well

src/app-layout/__integ__/runtime-drawers.test.ts
georgylobko: addressed

src/internal/plugins/controllers/drawers.ts
georgylobko: do we have such a request from sidecar team? I assume they have their own way to determine if the method they're about to call exists in this api, such as 
```javascript
if (awsuiPlugins.appLayout.openDrawer) {
    awsuiPlugins.appLayout.openDrawer('drawer-id');
}
```

on top of that, this api will be available before the tela drawer, so I have doubts that it's necessary

just-boris: With this code they can change if the API is available, but they do not know if there is a _compatible app layout available_

You can know this by checking the private property `this.drawerResizeListener`, but we do not want that

just-boris (approved):

PR #3684, opened 2025-07-24
feat: icons-for-expandable-buttons

mariksav:
### Description

Customers want to assign icons to expandable ButtonDropdowns, which wasn't possible so far. In order to enable that functionality, the rendering for icons for expandable Dropdown Buttons was added and test cases implemented. Furthermore, the style was changed to keep the look of the page consistent.

Issue: `AWSUI-60667`

### How has this been tested?

Tested locally with unit and integration tests. 

---

### DIFF
--- a/pages/button-dropdown/icon-expandable.page.tsx
+++ b/pages/button-dropdown/icon-expandable.page.tsx
@@ -22,7 +22,7 @@ export const items: ButtonDropdownProps['items'] = [
     id: 'category2',
     text: 'category2',
     iconName: 'settings',
-    items: [...Array(5)].map((_, index) => ({
+    items: [...Array(3)].map((_, index) => ({
       id: 'category2Subitem' + index,
       text: 'Cat 2 Sub item ' + index,
     })),
@@ -35,21 +35,21 @@ export const items: ButtonDropdownProps['items'] = [
     id: 'category3',
     text: 'category3',
     iconName: 'folder',
-    items: [...Array(3)].map((_, index) => ({
+    items: [...Array(2)].map((_, index) => ({
       id: 'category3Subitem' + index,
       text: 'Sub item ' + index,
     })),
   },
 ];
 
-export default function SimpleButtonDropdownScenario() {
+export default function IconExpandableButtonDropdown() {
   const [expandToViewport, setExpandToViewport] = useState(false);
   const [expandableGroups, setExpandableGroups] = useState(true);
 
   return (
     <div className={styles.container}>
       <article>
-        <h1>Simple Expandable Dropdown</h1>
+        <h1>Icon Expandable Dropdown</h1>
         <SpaceBetween size="m" direction="horizontal">
           <label>
             <input
@@ -71,23 +71,13 @@ export default function SimpleButtonDropdownScenario() {
           </label>
         </SpaceBetween>
 
-        <div style={{ padding: '100px', display: 'inline-block' }} id="dropdown-container">
+        <div style={{ padding: '10px', display: 'inline-block', marginTop: '10px' }} id="dropdown-container">
           <ButtonDropdown
             id="simpleDropdown"
             expandToViewport={expandToViewport}
             expandableGroups={expandableGroups}
             items={items}
-          >
-            Dropdown with Icons
-          </ButtonDropdown>
-        </div>
-
-        <div style={{ margin: '20px 0' }}>
-          <ButtonDropdown
-            id="simpleDropdown"
-            expandToViewport={expandToViewport}
-            expandableGroups={expandableGroups}
-            items={items}
+            data-testid="icon-dropdown"
           >
             Dropdown with Icons
           </ButtonDropdown>
--- a/pages/button-dropdown/icon-expandable.page.tsx
+++ b/pages/button-dropdown/icon-expandable.page.tsx
@@ -1,10 +1,11 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
-import React, { useState } from 'react';
+import React, { useContext } from 'react';
 
 import ButtonDropdown, { ButtonDropdownProps } from '~components/button-dropdown';
 import SpaceBetween from '~components/space-between';
 
+import AppContext, { AppContextType } from '../app/app-context';
 import ScreenshotArea from '../utils/screenshot-area';
 
 import styles from './styles.scss';
@@ -17,14 +18,13 @@ export const items: ButtonDropdownProps['items'] = [
     items: [...Array(2)].map((_, index) => ({
       id: 'category1Subitem' + index,
       text: 'Sub item ' + index,
-      iconName: 'star',
     })),
   },
   {
     id: 'category2',
     text: 'category2',
-    iconName: 'settings',
-    items: [...Array(3)].map((_, index) => ({
+    iconUrl: 'data:image/png;base64,aaaa',
+    items: [...Array(2)].map((_, index) => ({
       id: 'category2Subitem' + index,
       text: 'Cat 2 Sub item ' + index,
     })),
@@ -32,11 +32,19 @@ export const items: ButtonDropdownProps['items'] = [
   {
     id: 'item1',
     text: 'Item 1',
+    iconName: 'settings',
   },
   {
     id: 'category3',
     text: 'category3',
-    iconName: 'folder',
+    iconSvg: (
+      <svg width="16" height="16" viewBox="0 0 16 16" fill="none" focusable="false">
+        <path
+          d="M2 4V12C2 12.5523 2.44772 13 3 13H13C13.5523 13 14 12.5523 14 12V6C14 5.44772 13.5523 5 13 5H8L6 3H3C2.44772 3 2 3.44772 2 4Z"
+          fill="currentColor"
+        />
+      </svg>
+    ),
     items: [...Array(2)].map((_, index) => ({
       id: 'category3Subitem' + index,
       text: 'Sub item ' + index,
@@ -44,9 +52,17 @@ export const items: ButtonDropdownProps['items'] = [
   },
 ];
 
+type DemoContext = React.Context<
+  AppContextType<{
+    expandableGroups: boolean;
+  }>
+>;
+
 export default function IconExpandableButtonDropdown() {
-  const [expandToViewport, setExpandToViewport] = useState(false);
-  const [expandableGroups, setExpandableGroups] = useState(true);
+  const {
+    urlParams: { expandableGroups = true },
+    setUrlParams,
+  } = useContext(AppContext as DemoContext);
 
   return (
     <div className={styles.container}>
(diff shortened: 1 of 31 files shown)


pages/button-dropdown/icon-expandable.page.tsx
```diff
@@ -0,0 +1,92 @@
...
+    iconName: 'folder',
+    items: [...Array(2)].map((_, index) => ({
+      id: 'category3Subitem' + index,
+      text: 'Sub item ' + index,
+    })),
+  },
+];
+
+export default function IconExpandableButtonDropdown() {
+  const [expandToViewport, setExpandToViewport] = useState(false);
```
jperals: Is it useful to test this feature with `expandToViewport` set to `true` and `false`? If not, I would remove this and the corresponding input and not set this prop in the component.

pages/button-dropdown/icon-expandable.page.tsx
```diff
@@ -0,0 +1,92 @@
...
+  const [expandableGroups, setExpandableGroups] = useState(true);
```
jperals: In the case of `expandableGroups`, I suggest we keep it because we don't have enough visual regression coverage of icons in button dropdown items also for non-expandable items, but I would suggest to:
- Hook it to the URL parameters instead of the React state. Check how other dev pages use `urlParams` and `setUrlParams`. Then you can easily access the page in the given state by setting the parameters in the URL. This is especially useful in integration tests (either functional or visual regression) —and you can also send direct links to a given state
- Cover the case with `expandableGroups=false` in the visual regression tests (once you have hooked the state into the URL parameters, it is very easy to visit that state without having to interact with the page)

pages/button-dropdown/icon-expandable.page.tsx
```diff
@@ -0,0 +1,92 @@
...
+              type="checkbox"
+              checked={expandableGroups}
+              onChange={e => setExpandableGroups(!!e.target.checked)}
+            />{' '}
+            expandableGroups
+          </label>
+        </SpaceBetween>
+
+        <ScreenshotArea style={{ paddingBlockEnd: 300, paddingInlineEnd: 100 }}>
+          <div style={{ padding: '10px', display: 'inline-block', marginTop: '10px' }} id="dropdown-container">
```
jperals: Is this div necessary? If any of its styles are needed they can probably be merged into the parent Screenshot area.

pages/button-dropdown/icon-expandable.page.tsx
```diff
@@ -0,0 +1,92 @@
...
+            <ButtonDropdown
+              id="simpleDropdown"
```
jperals: I think in your current visual regression tests implementation you can get rid of the `id` and `data-testid` here.

pages/button-dropdown/icon-expandable.page.tsx
```diff
@@ -0,0 +1,92 @@
...
+import React, { useState } from 'react';
+
+import ButtonDropdown, { ButtonDropdownProps } from '~components/button-dropdown';
+import SpaceBetween from '~components/space-between';
+
+import ScreenshotArea from '../utils/screenshot-area';
+
+import styles from './styles.scss';
+
+export const items: ButtonDropdownProps['items'] = [
```
jperals, line 13: It would be nice to also cover `iconUrl` and `iconSvg` in the visual regression tests. Maybe you can have one of each (i.e, `iconName`, `iconUrl` and `iconSvg`)?

src/button-dropdown/__tests__/button-dropdown-items.test.tsx
```diff
@@ -284,6 +288,28 @@ const items: ButtonDropdownProps.Items = [
       expect(wrapper.findExpandableCategoryById('category1')!.getElement()).toHaveTextContent('category1');
     });
 
+    test('should render icons in expandable category headers', () => {
```
jperals: This test suite has a `describe` block called `should render icons`. Would it make more sense to add this test in there and let its description just be `in expandable category headers`?

src/button-dropdown/__tests__/button-dropdown-items.test.tsx
```diff
@@ -284,6 +288,28 @@ const items: ButtonDropdownProps.Items = [
...
+        {
+          id: 'category1',
+          text: 'category1',
+          iconName: 'folder',
+          items: [
+            { id: 'i1', text: 'item1' },
+            { id: 'i2', text: 'item2' },
+          ],
+        },
+        {
```
jperals: This second item doesn't seem to do anything, I'd propose to remove it.

src/button-dropdown/__tests__/button-dropdown-items.test.tsx
```diff
@@ -284,6 +288,28 @@ const items: ButtonDropdownProps.Items = [
...
+          id: 'i3',
+          text: 'item3',
+        },
+      ];
+      const wrapper = renderButtonDropdown({ ...props, expandableGroups: true, items: groupedCategories });
+      wrapper.openDropdown();
+      const categoryElement = wrapper.findExpandableCategoryById('category1')!;
+      expect(categoryElement.findIcon()).toBeTruthy();
```
jperals: The existing tests check the 3 types of icon (`iconName`, `iconUrl` and `iconSvg`), so it would be good to have the same coverage in this new test as well.

src/button-dropdown/category-elements/expandable-category-element.tsx
```diff
@@ -90,7 +90,12 @@ const ExpandableCategoryElement = ({
               },
             } as GeneratedAnalyticsMetadataButtonDropdownExpand | GeneratedAnalyticsMetadataButtonDropdownCollapse)
       )}
+
+      //Add Rendering of icon image in expandable button
```
jperals: Is this comment useful, or was it only meant to stay during development?

src/button-dropdown/category-elements/mobile-expandable-category-element.tsx
```diff
@@ -82,7 +82,11 @@ const MobileExpandableCategoryElement = ({
               },
             } as GeneratedAnalyticsMetadataButtonDropdownExpand)
       )}
+      //Add Rendering of icon image in expandable button
```
jperals: Same comments as in the file above

src/button-dropdown/category-elements/styles.scss
```diff
@@ -113,7 +113,7 @@
 
   &-right {
     transform: rotate(-90deg);
-
+    margin-inline-start: auto;
```
jperals, line 116: What is this line for?

src/button-dropdown/category-elements/expandable-category-element.tsx
```diff
@@ -90,7 +90,12 @@ const ExpandableCategoryElement = ({
...
     >
+      {(item.iconName || item.iconUrl || item.iconSvg) && (
+        <InternalIcon name={item.iconName} url={item.iconUrl} svg={item.iconSvg} alt={item.iconAlt} />
```
jperals, line 95: Remember that there is a missing space between this icon and the text. Same in the mobile version below.

jperals (dismissed):

src/button-dropdown/category-elements/styles.scss
mariksav: Moves the text to the icon and removes the gap between them

jperals: Why? Note that this selector matches the caret icon that expands and collapses the dropdown, not the new icon you added.

Screenshot 2025-07-24 at 17 14 27

mariksav: sry my mistake, without it the carrot isnt it the right position, so its still necessary

pages/button-dropdown/icon-expandable.page.tsx
```diff
@@ -0,0 +1,102 @@
...
+  } = useContext(AppContext as DemoContext);
+
+  return (
+    <div className={styles.container}>
+      <article>
+        <h1>Icon Expandable Dropdown</h1>
+        <SpaceBetween size="m" direction="horizontal">
+          <label>
+            <input id="expandToViewport" type="checkbox" /> expandToViewport
+          </label>
```
jperals: This labeled input can be removed now that they don't do anything.

The SpaceBetween around it can also be removed since it will only have one child —or better, you can use this SpaceBetween to manage the separation with the ScreenshotArea below: Remove `direction="horizontal"`, move the ScreenshotArea into it, and then you can remove the `marginTop` style property from the ScreenshotArea.

pages/button-dropdown/icon-expandable.page.tsx
```diff
@@ -0,0 +1,102 @@
...
+            />{' '}
+            expandableGroups
+          </label>
+        </SpaceBetween>
+
+        <ScreenshotArea
+          style={{
+            paddingBlockEnd: 300,
+            paddingInlineEnd: 100,
+            padding: '10px',
```
jperals: I would not use `paddingBlockEnd` and `paddingInlineEnd` together with the shorthand `padding` property, which overrides all paddings, so it is not clear "which one wins".

I would propose

```
paddingBlockStart: 10,
paddingBlockEnd: 300,
paddingInlineStart: 10,
paddingInlineEnd: 100,
```

src/button-dropdown/__tests__/button-dropdown-items.test.tsx
```diff
@@ -432,6 +436,52 @@ const items: ButtonDropdownProps.Items = [
...
+          },
+        ];
+        const wrapper = renderButtonDropdown({ ...props, expandableGroups: true, items: groupedCategories });
+        wrapper.openDropdown();
+        expect(wrapper.findExpandableCategoryById('category1')!.findIcon()).toBeTruthy();
+        expect(wrapper.findExpandableCategoryById('category2')!.findIcon()).toBeTruthy();
+        expect(wrapper.findExpandableCategoryById('category3')!.findIcon()).toBeTruthy();
+      });
+
+      test('in category headers when expandableGroups is false', () => {
```
jperals: What about reusing the same test above with `test.foreach([true, false])`, and then you also have coverage for all icon types when `expandableGroups` is `false`?

src/button-dropdown/category-elements/styles.scss
```diff
@@ -17,7 +17,8 @@
   border-inline-width: 0;
   font-weight: bold;
   display: flex;
-  justify-content: space-between;
+  align-items: center;
```
jperals: This keeps the icon at the vertical center when the text wraps:

Screenshot 2025-07-25 at 18 09 13

whereas in non-group items it currently stays aligned to the top:

Screenshot 2025-07-25 at 18 14 58

src/button-dropdown/category-elements/styles.scss
```diff
@@ -17,7 +17,8 @@
...
+  gap: awsui.$space-xxxs;
```
jperals: The gap in non-group items is `$space-xxs`: https://github.com/cloudscape-design/components/blob/f6bc21bcb3cae1ff0f23160b4ee8462eb77ec35f/src/button-dropdown/item-element/styles.scss#L95

I would not make it smaller here, for consistency and because with bold text if feels even more condensed.

taheramr (approved):

PR #3686, opened 2025-07-24
feat: Adds support for custom formats in date input

pan-kot:
### Description

The PR partially adopts changes from https://github.com/cloudscape-design/components/pull/3555, by only adding the feature to the date input.

Rel: [DvXYAOkww6To]

### How has this been tested?

* New unit tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---


pages/date-input/common.tsx
```diff
@@ -0,0 +1,39 @@
...
+  granularity: DateInputProps.Granularity;
+  inputFormat: undefined | DateInputProps.InputFormat;
+  format: undefined | DateInputProps.Format;
+}) => {
+  const isIso = inputFormat === 'iso' || (inputFormat === undefined && format === 'iso');
+  const separator = isIso ? '-' : '/';
+  return `YYYY${separator}${granularity === 'month' ? 'MM' : 'MM' + separator + 'DD'}`;
+};
+
+export const locales = [
```
jperals, line 20: Minor: this can be de-duplicated from `pages/date-range-picker/absolute-format.localization.page.tsx`.

pan-kot: Done

pages/date-input/permutations-formats.page.tsx
```diff
@@ -0,0 +1,53 @@
...
+    granularity: ['day', 'month'],
+    locale: locales,
+  },
+]);
+
+export default function DateInputPermutations() {
+  return (
+    <Box padding="l">
+      <h1>Date Input permutations - formats</h1>
+      <ScreenshotArea disableAnimations={true}>
```
jperals: Is there any animation that we need to disable?

pan-kot: No, I copied that from the existing permutations page, which is also static. I will remove this from both pages.

pan-kot: Done

src/date-input/__tests__/date-input.test.tsx
```diff
@@ -320,4 +322,92 @@ describe('Date Input component', () => {
     wrapper.setInputValue('2019/01/31');
     expect(onChangeSpy).toHaveBeenCalledWith(expect.objectContaining({ detail: { value: '2019-01-31' } }));
   });
+
+  test.each([{ inputFormat: undefined }, { inputFormat: 'iso' }, { inputFormat: 'slashed' }] as const)(
+    'accepts value in slashed format, inputFormat=$inputFormat',
+    ({ inputFormat }) => {
+      const getValue = (wrapper: DateInputWrapper) => wrapper.findNativeInput()!.getElement().value;
```
jperals: I guess you added the `inputValue` property to the return value of `renderDateInput` after adding this line, but you can use it here too instead.

pan-kot: Good finding!

src/date-input/interfaces.ts
```diff
@@ -4,14 +4,54 @@
...
+   **/
+  granularity?: DateInputProps.Granularity;
+
+  /**
+   * The format as it is displayed. It can take the following values:
+   * * `iso`: ISO 8601 format without time, e.g.: 2024-01-30 (or 2024-01).
+   * * `long-localized`: a more human-readable, localized format, e.g.: January 30, 2024 (or January, 2024).
+   * * `slashed`: similar to ISO 8601 but with '/' in place of '-'. e.g.: 2024/01/30 (or 2024/01).
+   *
+   * @default 'slashed'
```
jperals: Why do we use this decoration here and in the next prop but in the previous one we tell the default in "prose" instead?

Also, I cannot find any other usages of this `@default` decorator in the codebase, but I can find `@defaultValue` in the Dropdown component.

src/date-input/interfaces.ts
```diff
@@ -4,14 +4,54 @@
...
+   */
+  format?: DateInputProps.Format;
+
+  /**
+   * Specifies the date format to use when the format is 'long-localized' and the user needs to edit the date.
+   * The format of the input as it is being interacted with. It can take the following values:
+   * * `iso`: ISO 8601 format without time, e.g.: 2024-01-30 (or 2024-01).
+   * * `slashed`: similar to ISO 8601 but with '/' in place of '-'. e.g.: 2024/01/30 (or 2024/01).
+   *
+   * @default 'slashed'.
```
jperals: Also, the sentence here ends with a dot and in the previous prop it doesn't

src/date-input/interfaces.ts
```diff
@@ -4,14 +4,54 @@
...
+   * * `iso`: ISO 8601 format without time, e.g.: 2024-01-30 (or 2024-01).
+   * * `long-localized`: a more human-readable, localized format, e.g.: January 30, 2024 (or January, 2024).
+   * * `slashed`: similar to ISO 8601 but with '/' in place of '-'. e.g.: 2024/01/30 (or 2024/01).
+   *
+   * @default 'slashed'
+   */
+  format?: DateInputProps.Format;
+
+  /**
+   * Specifies the date format to use when the format is 'long-localized' and the user needs to edit the date.
```
jperals: What about "is editing the date" instead of "needs to edit the date"?

src/date-input/index.tsx
```diff
@@ -11,7 +11,7 @@ import InternalDateInput from './internal';
 
 export { DateInputProps };
 
-const DateInput = React.forwardRef((props: DateInputProps, ref: Ref<HTMLInputElement>) => {
+const DateInput = React.forwardRef((props: DateInputProps, ref: Ref<DateInputProps.Ref>) => {
```
jperals, line 14: Can this break any customers?

src/date-input/__tests__/utils.test.ts
```diff
@@ -54,3 +55,36 @@ describe('generateMaskArgs', () => {
...
+    expect(normalizeIsoDateString('201', 'month')).toBe('');
+    expect(normalizeIsoDateString('2012', 'month')).toBe('2012');
+    expect(normalizeIsoDateString('2012-', 'month')).toBe('2012');
+    expect(normalizeIsoDateString('2012-0', 'month')).toBe('2012');
+    expect(normalizeIsoDateString('2012-01', 'month')).toBe('2012-01');
+    expect(normalizeIsoDateString('2012-01-', 'month')).toBe('2012-01');
+    expect(normalizeIsoDateString('2012-01-0', 'month')).toBe('2012-01');
+    expect(normalizeIsoDateString('2012-01-01', 'month')).toBe('2012-01');
+    expect(normalizeIsoDateString('2012-1', 'month')).toBe('2012');
+    expect(normalizeIsoDateString('2012-12', 'month')).toBe('2012-12');
```
jperals, line 89: What about covering here also the case with a non-zero 1-digit day, like in the day granularity cases above, e.g, `2012-12-1`?

src/date-input/__tests__/date-input.test.tsx
```diff
@@ -320,4 +322,91 @@ describe('Date Input component', () => {
...
+    expect(wrapper.findNativeInput().getElement().value).toBe('3 February 2001');
+  });
+
+  test('does not call onChange on focus/blur events when format is long-localized', () => {
+    const { wrapper, onChangeSpy } = renderDateInput({
+      value: '2001-02-03',
+      format: 'long-localized',
+      locale: 'en-GB',
+    });
+    wrapper.focus();
```
jperals, line 397: When using `inputFormat`, there are other cases  the value of the input changes but the underlying value doesn't:

- Changing the locale
- Changing `format`

Should we also cover that `onChange` does not get called in these cases?

src/internal/components/masked-input/index.tsx
```diff
@@ -51,6 +52,10 @@ const MaskedInput = React.forwardRef(
     });
 
     const inputProps = { ...rest, ...baseProps, ...formFieldContext, ...maskProps };
+    if (showUnmaskedValue) {
```
jperals: Nitpick: this can be done without mutations:

```
const inputProps = {
  ...rest,
  ...baseProps,
  ...formFieldContext,
  ...maskProps,
  value: showUnmaskedValue ? value : maskProps.value
}
```
Also, not sure we want to pass `...maskProps` at all when `showUnmaskedValue` is true. It includes event listeners which are meant for when the input is focused.

src/date-input/index.tsx
pan-kot: Yes, I will make a dry-run to minimise the risks

src/date-input/interfaces.ts
pan-kot: Good point - I will remove that. It was a copy from https://github.com/cloudscape-design/components/pull/3555

src/date-input/__tests__/utils.test.ts
pan-kot: Good point, I will add these!

src/date-input/__tests__/date-input.test.tsx
pan-kot: Good point - let me add these, too!

src/internal/components/masked-input/index.tsx
pan-kot: I removed the maskProps at first, but then it caused a warning in the input due to the lack of the onChange handler. I ended up keeping both maskProps and the overridden value.

src/date-input/index.tsx
pan-kot: Build ref: 7458334964

jperals (approved):

PR #3688, opened 2025-07-24
feat: Implementation of the Alert component Style API

teodoranemes:
### Description

This PR implements the core Style API for the Alert component.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/alert/internal.tsx
```diff
@@ -132,6 +135,7 @@ const InternalAlert = React.forwardRef(
                 hasAction && styles['with-action'],
                 dismissible && styles['with-dismiss']
               )}
+              style={getAlertStyles(style)}
             >
               <div className={styles['alert-wrapper']}>
                 <div className={styles['alert-focus-wrapper']} tabIndex={-1} ref={focusRef}>
```
gethinwebster, line 141: the style API for this component most likely needs root focus ring properties as well, for this `alert-focus-wrapper` style

src/alert/interfaces.ts
```diff
@@ -47,6 +47,33 @@ export namespace AlertProps {
   export interface AnalyticsMetadata {
     errorContext?: ErrorContext;
   }
+
+  export interface Style {
```
YueyingLu, line 51: Why style is applying to all types but not to specific type(info, success, error, warning)?

teodoranemes: Because all the alert types share the same properties and don't have dynamic states that change at runtime, like the button, where the style was applied to the specific type.

src/alert/internal.tsx
teodoranemes: Added it now.

gethinwebster (commented):
Just noticed in `styles.scss` that there is a (existing) usage of the border-width variable:

```
  // To account for vertical misalignment due to button borders
  padding-block: awsui.$border-width-alert;
```

It's not entirely clear what the actual purpose of this is (the line and the comment don't seem to really match up), but it would be good to either:
1. Confirm that this padding should indeed match the alert border width, and then make changes to the style API implementation to take this into account
2. Or, if they shouldn't actually be directly related, find a different variable to use for this line instead

(I suspect 2. is more likely the case, but it'd be good to clarify)

PR #3691, opened 2025-07-25
chore: Applayout manages the global scroll-padding-top

orangevolon:
### Description

Adds `useGlobalScrollPadding` hook used in both `<AppLayout />` and `<AppLayoutToolbar />` components.
This sets a `scroll-padding-block-start` to the scrolling element (`html` or `body`) so that the elements don't stay hidden under the sticky header, when they are focused.

Related links, issue #, if available: AWSUI-61007

### How has this been tested?

Added a new page and integration test for it.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/__integ__/awsui-applayout.test.ts
+++ b/src/app-layout/__integ__/awsui-applayout.test.ts
@@ -199,6 +199,38 @@ describe.each(['classic', 'refresh', 'refresh-toolbar'] as Theme[])('%s', theme 
     })
   );
 
+  (theme !== 'classic' ? test : test.skip)(
+    'element should not be hidden under the sticky header when focused',
+    setupTest({ pageName: 'global-scroll-padding' }, async page => {
+      // Getting the header offset depending on the theme
+      let headerOffset = (await page.getBoundingBox('#h')).height;
+      if (theme === 'refresh-toolbar') {
+        headerOffset += (await page.getBoundingBox(wrapper.findToolbar().toSelector())).height;
+      }
+
+      // Set the focus to the second button
+      await page.click(wrapper.findContentRegion().findHeader().toSelector());
+      await page.keys('Tab');
+      await page.keys('Tab');
+      const secondButtonSelector = wrapper.findContentRegion().findButton('[data-testid="button-2"]').toSelector();
+      await expect(page.isFocused(secondButtonSelector)).resolves.toBe(true);
+
+      // Scroll to a point where half of the first button is visible
+      const firstButtonSelector = wrapper.findContentRegion().findButton('[data-testid="button-1"]').toSelector();
+      const firstButtonBoundingBox = await page.getBoundingBox(firstButtonSelector);
+      const firstButtonTopOffset = firstButtonBoundingBox.top - headerOffset;
+      const firstButtonHalfOffset = firstButtonTopOffset + firstButtonBoundingBox.height / 2;
+      await page.windowScrollTo({ top: firstButtonHalfOffset });
+
+      // Set the focus back to the first button
+      await page.keys(['Shift', 'Tab']);
+      await expect(page.isFocused(firstButtonSelector)).resolves.toBe(true);
+
+      // Assert that whole of the first button is visible and not only the half of it
+      await expect(page.getWindowScroll()).resolves.toEqual({ top: firstButtonTopOffset, left: 0 });
+    })
+  );
+
   testIf(theme === 'refresh-toolbar')(
     'should keep header visible and in position while scrolling',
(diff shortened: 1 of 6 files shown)


src/app-layout/__tests__/main.test.tsx
```diff
@@ -117,6 +117,17 @@ describeEachAppLayout({ themes: ['classic', 'refresh-toolbar'], sizes: ['desktop
...
 
+    (theme !== 'classic' ? test : test.skip)('should set the header height to the scrolling element', () => {
+      Object.defineProperty(document, 'scrollingElement', { value: document.body });
+      renderComponent(
+        <div id="b">
+          <div style={{ height: 40 }} id="h" />
+          <AppLayout />
+        </div>
+      );
+      expect(document.scrollingElement).toHaveStyle('scroll-padding-block-start: 40px');
```
orangevolon, line 132: For some reason the toolbar height in the `refresh-toolbar` is `0`. This case is covered properly in integ anyways.

src/app-layout/__integ__/awsui-applayout.test.ts
```diff
@@ -199,6 +199,38 @@ describe.each(['classic', 'refresh', 'refresh-toolbar'] as Theme[])('%s', theme
     })
   );
 
+  (theme !== 'classic' ? test : test.skip)(
+    'element should not be hidden under the sticky header when focused',
```
georgylobko, line 203: Can we make this test simpler? 

1. land on the page
2. scroll to the point when one of the buttons isn't visible
3. set focus to the button
4. make sure it's within the viewport

src/app-layout/visual-refresh-toolbar/index.tsx
```diff
@@ -313,6 +314,8 @@ const AppLayoutVisualRefreshToolbar = React.forwardRef<AppLayoutProps.Ref, AppLa
       stickyNotifications: resolvedStickyNotifications,
     });
 
+    useGlobalScrollPadding(verticalOffsets.header);
```
georgylobko, line 317: We can apply this hook in the `src/app-layout/index` component level, regardless of the theme (classic, vr etc.) here: https://github.com/cloudscape-design/components/blob/main/src/app-layout/index.tsx#L105

Because the value of header is calculated in the root component level, so we can cover all themes at once. In this case we can even get rid of the hook above and put this code as is.

src/app-layout/visual-refresh-toolbar/toolbar/index.tsx
```diff
@@ -87,7 +87,9 @@ export function AppLayoutToolbarImplementation({
   } = toolbarProps;
   const drawerExpandedMode = !!expandedDrawerId;
   const ref = useRef<HTMLElement>(null);
-  useResizeObserver(ref, entry => setToolbarHeight(entry.borderBoxHeight));
+  useResizeObserver(ref, entry => {
```
georgylobko: Did your local linter change that?

src/app-layout/__integ__/awsui-applayout.test.ts
orangevolon: This is exactly what this test is doing. How do you suggest to make it simpler?

src/app-layout/visual-refresh-toolbar/index.tsx
orangevolon: The reason to have it in a hook is that: The header height calculation for the toolbar variant is different from the refresh variant. This value is calculated at `computeVerticalLayout` line `309` of this file. To avoid repeating the height calculation logic in `index.ts` I created this hook which can be reused inside each app-layout rather than in the top level `index.ts`.

src/app-layout/visual-refresh-toolbar/toolbar/index.tsx
orangevolon: It might've. Otherwise there should be no change here.

src/app-layout/__integ__/awsui-applayout.test.ts
georgylobko: discussed in DM

src/app-layout/visual-refresh-toolbar/index.tsx
georgylobko: discussed in DM

src/app-layout/__tests__/main.test.tsx
orangevolon: Discussed; Added a comment

src/app-layout/visual-refresh-toolbar/toolbar/index.tsx
orangevolon: Removed the change

PR #3692, opened 2025-07-27
fix: Add LiveRegion to non-dismissible Popover

johannes-weber:
### Description

Previously, `aria-live` and `aria-atomic` attributes on the popover content weren't effectively announcing content to screen readers.

Implements LiveRegion component to properly announce non-dismissible popover content to screen readers.

Note: This change only affects non-dismissible popovers since:
- Dismissible popovers receive focus and are naturally navigable by screen reader users
- Non-dismissible popovers lack focus management, requiring content to be announced via live regions

Related links, issue #, if available: `AWSUI-60746`

### How has this been tested?

- update / add existing unit tests
- green dry run

How can reviewers test these changes efficiently?
- Open the popup examples with a screen reader turned on
- Navigate via keyboard to the popover trigger and press enter
- The popover content is live announced.
- Close the popover and re-open it. Live announcement is happening again.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/popover/__tests__/popover.test.tsx
+++ b/src/popover/__tests__/popover.test.tsx
@@ -279,9 +279,12 @@ describe('ARIA labels', () => {
   });
 
   it('uses aria-live if dismissButton is false', () => {
-    const wrapper = renderPopover({ children: 'Trigger', content: 'Popover', dismissButton: false });
+    const wrapper = renderPopover({ children: 'Trigger', content: 'Popover content', dismissButton: false });
     wrapper.findTrigger().click();
-    expect(wrapper.find('[aria-live="polite"]')).toBeTruthy();
+
+    const liveRegion = createWrapper().findLiveRegion()!;
+    expect(liveRegion).toBeDefined();
+    expect(liveRegion.getElement()).toHaveTextContent('Popover content');
   });
 
   it('does not pass the dismissAriaLabel to the dismiss button if not defined', () => {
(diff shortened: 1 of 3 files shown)


src/popover/__tests__/popover.test.tsx
```diff
@@ -308,15 +311,39 @@ describe('ARIA labels', () => {
...
+        children: 'Trigger',
+        content: 'Popover',
+        dismissButton: true,
+        renderWithPortal: true,
+      });
+      wrapper.findTrigger().click();
+      expect(createWrapper().findLiveRegion()).toBeNull();
     });
-    wrapper.findTrigger().click();
-    await expect(document.body).toValidateA11y();
```
connorlanigan, line 319: Would it be useful to still call the `toValidateA11y` line to ensure other a11y checks?

johannes-weber: I've removed it by purpose: When adding the `LiveRegion`, it returns a false positive about "Ensures all page content is contained by `landmarks`".

Why does this happen?
The `LiveRegion` component appends an empty `div` with the aria-live property to the `body` ([link to source](https://github.com/cloudscape-design/components/blob/main/src/live-region/controller.ts#L37)). The div serves as central announcer. After a specific timeout the content to announce get's added to this div ([link to source](https://github.com/cloudscape-design/components/blob/main/src/live-region/controller.ts#L65)).

Because that div is appended to the body, and won't be part of any `landmark` elements, the a11y check is failing. 

Why is this a false positive?
While the landmark warning is technically true, it doesn't affect the functional a11y of the live region or the non dismissible popover.

connorlanigan (approved):

PR #3694, opened 2025-07-28
feat: Add support for download links in ButtonDropdown

TrevorBurnham:
This PR adds the ability to specify `download: true` for an item in a `ButtonDropdown`, in conjunction with an `href`, to get a download link instead of a normal link. This matches the behavior of the `download` prop on the `Button` component.

### How has this been tested?

I've added unit tests as part of this PR.

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/button-dropdown/__tests__/button-dropdown-items.test.tsx
+++ b/src/button-dropdown/__tests__/button-dropdown-items.test.tsx
@@ -499,3 +499,227 @@ describe('Internal ButtonDropdown badge property', () => {
     expect(wrapper.findAllByClassName(iconStyles.badge)?.map(item => item.getElement())).toHaveLength(2);
   });
 });
+
+describe('ButtonDropdown download property', () => {
+  const testProps = { expandToViewport: false };
+
+  it('should render download attribute with boolean true value', () => {
+    const items: ButtonDropdownProps.Items = [
+      {
+        id: 'download-item',
+        text: 'Download file',
+        href: 'https://example.com/file.pdf',
+        download: true,
+      },
+    ];
+    const wrapper = renderButtonDropdown({ ...testProps, items });
+    wrapper.openDropdown();
+
+    const anchor = wrapper.findItemById('download-item')!.find('a')!.getElement();
+    expect(anchor).toHaveAttribute('download', '');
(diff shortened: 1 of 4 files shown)


src/button-dropdown/__tests__/button-dropdown-items.test.tsx
```diff
@@ -457,3 +457,227 @@ describe('Internal ButtonDropdown badge property', () => {
...
+    ];
+    const wrapper = renderButtonDropdown({ ...testProps, items });
+    wrapper.openDropdown();
+
+    const item = wrapper.findItemById('no-href-item')!;
+    expect(item.find('a')).toBe(null);
+    expect(item.getElement()).toHaveTextContent('Button item');
+  });
+
+  it('should not render download attribute when item is disabled', () => {
```
orangevolon, line 589: The link is already not clickable when disabled. Why do we want to remove the `download` attribute?

src/button-dropdown/interfaces.ts
```diff
@@ -165,7 +167,7 @@ export namespace ButtonDropdownProps {
   }
 
   export interface CheckboxItem
-    extends Omit<ButtonDropdownProps.Item, 'href' | 'external' | 'externalIconAriaLabel' | 'itemType'> {
+    extends Omit<ButtonDropdownProps.Item, 'href' | 'download' | 'external' | 'externalIconAriaLabel' | 'itemType'> {
```
orangevolon, line 204: Wondering if we need a visual clue for download similar to external.

TrevorBurnham: In practice, I think we'd want to use `iconName="download"` whenever we use the `download` attribute. But I don't think that should be enforced; someone might want to use a different (custom?) icon to differentiate some downloads.

Having `download` and `iconName` as orthogonal options matches the API of the `Button` component.

src/button-dropdown/__tests__/button-dropdown-items.test.tsx
TrevorBurnham: Honestly, Claude added this. I agree that it doesn't really matter. The `Button` component keeps the `download` attribute when it's disabled.

PR #3695, opened 2025-07-28
feat: Implementation of the Container component Style API 

DaemonCahill:
This PR adds Style API support for the `Container` component. A screenshot of the demo page is included below.

Screenshot 2025-07-29 at 3 17 45 PM

### DIFF
--- a/src/container/interfaces.ts
+++ b/src/container/interfaces.ts
@@ -81,6 +81,25 @@ export interface ContainerProps extends BaseComponentProps {
    * @visualrefresh `stacked` variant
    */
   variant?: 'default' | 'stacked';
+
+  /**
+   * Specifies an object of selectors and properties that are used to apply custom styles.
+   *
+   * - `root.background` (string) - (Optional) Background color of the container.
+   * - `root.borderColor` (string) - (Optional) Border color of the container.
+   * - `root.borderRadius` (string) - (Optional) Border radius of the container.
+   * - `root.borderWidth` (string) - (Optional) Border width of the container.
+   * - `content.paddingBlock` (string) - (Optional) Block dimension padding of the content slot.
+   * - `content.paddingInline` (string) - (Optional) Inline dimension padding of the content slot.
+   * - `header.paddingBlock` (string) - (Optional) Block dimension padding of the header slot.
+   * - `header.paddingInline` (string) - (Optional) Inline dimension padding of the header slot.
+   * - `footer.borderColor` (string) - (Optional) Border color of the footer divider.
+   * - `footer.borderWidth` (string) - (Optional) Border width of the footer divider.
+   * - `footer.paddingBlock` (string) - (Optional) Block dimension padding of the footer slot.
+   * - `footer.paddingInline` (string) - (Optional) Inline dimension padding of the footer slot.
+   * @awsuiSystem core
+   */
+  style?: ContainerProps.Style;
 }
 
 export namespace ContainerProps {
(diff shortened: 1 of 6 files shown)


src/container/interfaces.ts
```diff
@@ -81,6 +81,25 @@ export interface ContainerProps extends BaseComponentProps {
...
+   *
+   * - `root.background` (string) - (Optional) Background color of the container.
+   * - `root.borderColor` (string) - (Optional) Border color of the container.
+   * - `root.borderRadius` (string) - (Optional) Border radius of the container.
+   * - `root.borderWidth` (string) - (Optional) Border width of the container.
+   * - `content.paddingBlock` (string) - (Optional) Block dimension padding of the content slot.
+   * - `content.paddingInline` (string) - (Optional) Inline dimension padding of the content slot.
+   * - `header.paddingBlock` (string) - (Optional) Block dimension padding of the header slot.
+   * - `header.paddingInline` (string) - (Optional) Inline dimension padding of the header slot.
+   * - `footer.borderColor` (string) - (Optional) Border color of the footer divider.
```
gethinwebster: as this is applied as a divider, i.e. only to one side, it should be named differently from a standard border property to make that clear. Suggest `dividerColor` and `dividerWidth`

src/container/interfaces.ts
```diff
@@ -81,6 +81,25 @@ export interface ContainerProps extends BaseComponentProps {
    * @visualrefresh `stacked` variant
    */
   variant?: 'default' | 'stacked';
```
gethinwebster, line 83: does this interact well with the `stacked` variant?

src/container/internal.tsx
```diff
@@ -65,6 +66,7 @@ export default function InternalContainer({
   disableFooterPaddings = false,
   fitHeight,
   media,
```
gethinwebster, line 68: it looks like media styling (border radii) are based on `awsui.$border-radius-container`, does that need to be updated to take account for custom border radius via style API?

src/container/interfaces.ts
```diff
@@ -81,6 +81,25 @@ export interface ContainerProps extends BaseComponentProps {
...
+
+  /**
+   * Specifies an object of selectors and properties that are used to apply custom styles.
+   *
+   * - `root.background` (string) - (Optional) Background color of the container.
```
gethinwebster: I wonder if we want to word this (and similarly in other components too) "Background of the container", as that's technically how it's applied, not purely as a background color. Might make it clearer that you can also do things like gradients with this.

src/container/interfaces.ts
DaemonCahill: We only use idiomatic CSS properties so we can't change that, but we can add a nested selector for the footer divider. I'll update the interface accordingly.

src/container/interfaces.ts
DaemonCahill: Agreed! Updated.

src/container/interfaces.ts
DaemonCahill: When styling `stacked` variants we will not override custom styles to form a contiguous border because we can't make assumptions over what each independent `Container` has for values in their respective Style APIs, and subsequently which value should win if they are in conflict with one another. Another constraint is that `Container` styles will not migrate to stacked variants of adjacent`Tabs` or `ExpandableSections`. This is by design and we don't want to open that box of nightmares. _tldr;_ `stacked` variants will apply custom styles unmodified, and changing certain values for these variants (i.e. `borderRadius`) may have undesirable visual outcomes. I added a sample to the demo page.

Screenshot 2025-07-29 at 2 54 43 PM

src/container/internal.tsx
DaemonCahill: Accounted for in the style logic and added to the demo page.

Screenshot 2025-07-29 at 3 16 39 PM

PR #3699, opened 2025-07-29
feat: Implementation of the Link component Style API

teodoranemes:
### Description

This PR implements the core Style API for the Link component.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/link/interfaces.ts
```diff
@@ -106,6 +106,17 @@ export interface LinkProps extends BaseComponentProps {
...
+  /**
+   * Specifies an object of selectors and properties that are used to apply custom styles.
+   *
+   * - `root.color` {active, default, hover} (string) - (Optional) Text color for link.
+   * - `root.focusRing.borderColor` (string) - (Optional) Focus ring border color.
+   * - `root.focusRing.borderRadius` (string) - (Optional) Focus ring border radius.
+   * - `root.focusRing.borderWidth` (string) - (Optional) Focus ring border width.
+   * @awsuiSystem core
+   */
+  style?: LinkProps.Style;
```
gethinwebster, line 119: I can imagine teams wanting to control link decoration/underline styles, but I think that can probably be a follow-up if/when they request it

src/link/styles.scss
```diff
@@ -45,7 +52,11 @@
   }
 
   @include focus-visible.when-visible {
-    @include styles.link-focus;
+    @include styles.focus-highlight(
```
gethinwebster: does this cause a regression for links in classic? (because of missing these lines: https://github.com/cloudscape-design/components/blob/main/src/internal/styles/forms/mixins.scss#L16-L19)

can you run through a dev pipeline to check?

src/link/__tests__/index.test.tsx
```diff
@@ -377,3 +377,40 @@ describe('"onClick" event', () => {
...
+          },
+        }}
+      >
+        Link
+      </Link>
+    );
+
+    const link = createWrapper(container).findLink()!.getElement();
+
+    expect(getComputedStyle(link).getPropertyValue('--awsui-style-color-active-kcc2gu')).toBe('rgb(163, 15, 15)');
```
gethinwebster: these names are dynamic, so need to be imported rather than hardcoded, like in https://github.com/cloudscape-design/components/blob/main/src/button/__tests__/button.test.tsx#L876

src/link/styles.scss
teodoranemes: I have added back the link-focus to be sure that there is no regression for the classic mode.

src/link/interfaces.ts
teodoranemes: Yes, I agree with your assessment, but we will start with the basics for the Style API and add more customization options later if customers request them.

src/link/__tests__/index.test.tsx
teodoranemes: Imported the names for the dynamic properties.

gethinwebster (dismissed):

gethinwebster (approved):

YueyingLu (approved):

src/internal/styles/forms/mixins.scss
```diff
@@ -12,33 +12,22 @@
 @use './constants' as constants;
 @use '../../generated/custom-css-properties/index.scss' as custom-props;
```
gethinwebster: this import can be removed now I think

teodoranemes: Solved.

src/link/styles.scss
```diff
@@ -8,11 +8,18 @@
...
+@use '../internal/generated/custom-css-properties/index.scss' as custom-props;
+@use '../internal/styles/foundation' as foundation;
 
 .link {
   @include styles.styles-reset;
   display: inline;
   white-space: inherit;
+
+  #{custom-props.$styleFocusRingBoxShadow}: 0 0 0
+    var(#{custom-props.$styleFocusRingBorderRadius}, awsui.$border-link-focus-ring-shadow-spread)
```
gethinwebster: this should be width not radius I think

teodoranemes: Looking at the name of the used token that defines the spread radius for the focus ring shadow, I was thinking that it is related to the radius, but I checked the documentation and replaced it with the width.

gethinwebster (approved):

YueyingLu (approved):

PR #3705, opened 2025-07-30
fix: Add group role to element referenced by the alert focus ref

jorycunningham:
### Description

This PR add `role="group"` to the element that gets focused when focus is called on the component (the element that has the 'focusRef' ref.

Element that receive focus should not be generic elements, adding `role="group"` resolves this.

### How has this been tested?

Tested manually in browser and added a corresponding test to `src/alert/__tests__/alert.test.tsx`

<details>

#### Correctness

- X Changes include appropriate documentation updates._
- X Changes are backward-compatible if not indicated, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#public-apis)._
- X Changes do not include unsupported browser features, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#browsers-support)._
- X Changes were manually tested for accessibility, see [accessibility guidelines](https://cloudscape.design/foundation/core-principles/accessibility/)._

#### Security

N/A

#### Testing

-  X Changes are covered with new/existing unit tests?_
</details>

---

### DIFF
--- a/src/alert/internal.tsx
+++ b/src/alert/internal.tsx
@@ -134,7 +134,7 @@ const InternalAlert = React.forwardRef(
               )}
             >
               <div className={styles['alert-wrapper']}>
-                <div className={styles['alert-focus-wrapper']} tabIndex={-1} ref={focusRef}>
+                <div className={styles['alert-focus-wrapper']} tabIndex={-1} ref={focusRef} role="group">
                   <div className={clsx(styles.icon, styles.text)}>
                     <InternalIcon name={typeToIcon[type]} size={size} ariaLabel={statusIconAriaLabel} />
                   </div>
--- a/src/alert/__tests__/alert.test.tsx
+++ b/src/alert/__tests__/alert.test.tsx
@@ -160,6 +160,15 @@ describe('Alert Component', () => {
     await expect(container).toValidateA11y();
   });
 
+  describe('a11y', () => {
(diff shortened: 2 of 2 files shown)


src/alert/internal.tsx
```diff
@@ -134,7 +134,7 @@ const InternalAlert = React.forwardRef(
               )}
             >
               <div className={styles['alert-wrapper']}>
-                <div className={styles['alert-focus-wrapper']} tabIndex={-1} ref={focusRef}>
+                <div className={styles['alert-focus-wrapper']} tabIndex={-1} ref={focusRef} role="group">
```
georgylobko, line 137: Should this role always be set, even if `focusRef` is not provided?

jorycunningham: Yes, I don't see any harm in that.

avinashbot: Is there any harm here of having an explicitly unlabeled group (instead relying on the implicit label based on the text content)?

jorycunningham: Great question. In this case I'm intentionally falling back to "Name from contents" for the accessible name of the group so that the user will hear the alert contents when the group element receives focus. I think that's that's the most user-friendly thing to do.

PR #3710, opened 2025-08-01
fix: Support icons for button dropdown groups

mariksav:
### Description

Fix for https://github.com/cloudscape-design/components/pull/3684

Customers want to assign icons to expandable ButtonDropdowns, which wasn't possible so far. In order to enable that functionality, the rendering for icons for expandable Dropdown Buttons was added and test cases implemented. Furthermore, the style was changed to keep the look of the page consistent.

Issue: AWSUI-60667

How has this been tested?
Tested locally with unit and integration tests.
Ran through own pipeline.
---

### DIFF
--- a/src/button-dropdown/category-elements/category-element.tsx
+++ b/src/button-dropdown/category-elements/category-element.tsx
@@ -32,14 +32,12 @@ const CategoryElement = ({
     >
       {item.text && (
         <p className={clsx(styles.header, { [styles.disabled]: disabled })} aria-hidden="true">
-          <span style={{ display: 'flex', alignItems: 'center' }}>
-            {(item.iconName || item.iconUrl || item.iconSvg) && (
-              <span className={styles['icon-wrapper']}>
-                <InternalIcon name={item.iconName} url={item.iconUrl} svg={item.iconSvg} alt={item.iconAlt} />
-              </span>
-            )}
-            {item.text}
-          </span>
+          {(item.iconName || item.iconUrl || item.iconSvg) && (
+            <span className={styles['icon-wrapper']}>
+              <InternalIcon name={item.iconName} url={item.iconUrl} svg={item.iconSvg} alt={item.iconAlt} />
+            </span>
+          )}
+          {item.text}
         </p>
       )}
       <ul className={styles['items-list-container']} role="group" aria-label={item.text} aria-disabled={disabled}>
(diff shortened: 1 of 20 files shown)


src/button-dropdown/category-elements/category-element.tsx
```diff
@@ -31,7 +32,14 @@ const CategoryElement = ({
     >
       {item.text && (
         <p className={clsx(styles.header, { [styles.disabled]: disabled })} aria-hidden="true">
-          {item.text}
+          <span style={{ display: 'flex', alignItems: 'center' }}>
```
avinashbot: Inline styles here - best to avoid them unless necessary for things like dynamic calculations. Inline styles don't match our [Content-Security-Policy](https://cloudscape.design/get-started/dev-guides/csp/) recommendations.

avinashbot: Also, out of curiosity, since you have `align-items: center` on the `.header` anyway, why have another wrapper span? (Maybe there are good reasons, just curious)

src/button-dropdown/category-elements/styles.scss
```diff
@@ -132,3 +134,7 @@
 .in-dropdown {
   margin-block-end: -1px;
 }
+
+.icon-wrapper {
```
avinashbot, line 138: If the parent is going to be a `display: flex`, I think you could just put `gap: awsui.$space-xxs` on the parent and not have to deal with a wrapper.

mariksav: that was the same idea that I had as well, but it caused the caret to be misplaced in button-dropdown/disabled-reason. So I circumvented that issue with the wrapper

src/button-dropdown/category-elements/category-element.tsx
mariksav: Good point! I used it during the troubleshooting after the rollback when `align-items: center` wasn't in the header yet, but it might not be necessary any more. I'm going to test it and remove it if possible.

mariksav:
I removed the span wrapper and everything works fine.

avinashbot (approved):

mariksav:
I retested the change and saw that the span wrapper was actually necessary, because otherwise the website formats in an unpleasant way in a few cases. So I will need to readd it, but I removed the inline styling. Please let me know what you think

PR #3734, opened 2025-08-07
feat: Native form submit for wizard component

ktdotgit:
### Description

jGwYAejXwIQb
AWSUI-60156

### How has this been tested?
- Manual testing, unit testing, and integration testing

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/wizard/__integ__/wizard.test.ts
+++ b/src/wizard/__integ__/wizard.test.ts
@@ -20,11 +20,34 @@ class WizardPageObject extends BasePageObject {
   toggleScrollableContainer() {
     return this.click(createWrapper().findToggle().findNativeInput().toSelector());
   }
+  getInput(selector?: string) {
+    return wizardWrapper.findContent().findInput(selector);
+  }
+  getInputSelector(selector?: string) {
+    return wizardWrapper.findContent().findInput(selector).findNativeInput().toSelector();
+  }
+  getFormFieldSelector(selector?: string) {
+    return wizardWrapper.findContent().findFormField(selector).toSelector();
+  }
+  getFormFieldErrorSelector(selector?: string) {
+    return wizardWrapper.findContent().findFormField(selector).findError().toSelector();
+  }
+  async type(text: string) {
+    // `await this.keys(text);` doesn't work as it key presses too quickly and doesn't
+    // allow the separator to be appended so the cursor position gets messed up.
+    for (let k = 0; k < text.length; k++) {
+      await this.keys(text[k]);
+    }
+  }
+  // gh pr create
+  // takes u to interactive interface
+  // will ask u if u want to be in draft mode
+  // name as "feat: <name of feature>"
 }
 
-function setupTest(testFn: (page: WizardPageObject) => Promise<void>) {
+function setupTest(testFn: (page: WizardPageObject) => Promise<void>, url?: string) {
   return useBrowser(async browser => {
-    await browser.url('/#/light/wizard/simple?visualRefresh=false');
+    await browser.url(url ?? '/#/light/wizard/simple?visualRefresh=false');
     const page = new WizardPageObject(browser);
     await page.waitForVisible(wizardWrapper.findPrimaryButton().toSelector());
     await testFn(page);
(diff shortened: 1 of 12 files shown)


src/wizard/__tests__/wizard.test.tsx
```diff
@@ -556,3 +557,54 @@ describe('i18n', () => {
...
+  ];
+
+  test('invokes onNavigate function on non-last step', () => {
+    const onNavigate = jest.fn();
+    const onSubmit = jest.fn();
+
+    const [wrapper] = renderDefaultWizard({ activeStepIndex: 0, steps, allowSkipTo: false, onNavigate, onSubmit });
+    const inputStep1 = wrapper.findContent()?.findInput()?.getElement() as HTMLElement;
+
+    fireEvent.submit(inputStep1);
```
ktdotgit: I'm not completely happy having to use `fireEvent.submit` here. It's too strong of an invocation. I'd rather use something like `fireEvent.keyDown` and pass in the `Enter` KeyCode but I could not get this to work. I tried basically every combination I could think of. I tried both `findInput` and `findNativeInput`. I tried focusing on the element before `fireEvent.keyDown`.

ktdotgit: The other variations I tried... I could not get this to work. I looked at other examples which employed typing/pressing 'Enter' in an Input element.

```
const [wrapper] = renderDefaultWizard({ activeStepIndex: 2, steps, allowSkipTo: false, onNavigate, onSubmit });
const inputStep3 = wrapper.findContent()?.findInput()?.findNativeInput() as ElementWrapper<HTMLInputElement>;

inputStep3.keydown({ key: 'Enter', keyCode: KeyCode.enter });
```

```
const [wrapper] = renderDefaultWizard({ activeStepIndex: 0, steps, allowSkipTo: false, onNavigate, onSubmit });
const inputStep1 = wrapper.findContent()?.findInput()?.getElement() as HTMLElement;

fireEvent.keyDown(inputStep1, { key: 'Enter' });
```

I also tried calling `focus` on the `input` element before `keydown` but it's still not working. I also tracked/spied on `onKeyDown` and it's being called as expected but perhaps I am incorrectly assuming that `keyDown enter === input submit`.

src/wizard/__integ__/wizard.test.ts
```diff
@@ -20,18 +20,95 @@ class WizardPageObject extends BasePageObject {
...
+    return wizardWrapper.findContent().findFormField(selector).findError().toSelector();
+  }
+  async type(text: string) {
+    // `await this.keys(text);` doesn't work as it key presses too quickly and doesn't
+    // allow the separator to be appended so the cursor position gets messed up.
+    for (let k = 0; k < text.length; k++) {
+      await this.keys(text[k]);
+    }
+  }
+  // gh pr create
```
ktdotgit: Whoops forgot to delete these notes.

src/wizard/__integ__/wizard.test.ts
```diff
@@ -20,18 +20,95 @@ class WizardPageObject extends BasePageObject {
...
+  getFormFieldErrorSelector(selector?: string) {
+    return wizardWrapper.findContent().findFormField(selector).findError().toSelector();
+  }
+  async type(text: string) {
+    // `await this.keys(text);` doesn't work as it key presses too quickly and doesn't
+    // allow the separator to be appended so the cursor position gets messed up.
+    for (let k = 0; k < text.length; k++) {
+      await this.keys(text[k]);
+    }
+  }
```
ernst-dev: This can be achieved by using the following inside the test itself instead of a helper function on the `WizardPageObject` class: `await page.keys(['input text'])` like [here](https://github.com/cloudscape-design/components/blob/main/src/autosuggest/__integ__/autosuggest.test.ts#L40)

src/wizard/__tests__/wizard.test.tsx
ernst-dev: You could try something like [this](https://github.com/cloudscape-design/components/blob/main/src/input/__tests__/input.test.tsx#L458-L475), otherwise I think integration tests will suffice and are better for these kinds of actions. If the example linked doesn't lead to a working unit test I'd just make sure that all the cases are covered in integ

ktdotgit: Yes, for both integ test and manual testing, the "right" and "expected" things are being done.

src/wizard/__integ__/wizard.test.ts
ktdotgit: Out of curiosity, how does it differentiate between a user wanting to type in the literal string "Enter" and wanting to hit return ("Enter")?

ernst-dev: This is leaning WebDriverIO's `keys` functions which will treat a string either as a chain of text to type or a keycode such as 'Enter' or 'ArrowLeft': https://webdriver.io/docs/api/browser/keys/

src/__tests__/functional-tests/outer-form-submit.test.tsx
```diff
@@ -19,7 +19,7 @@ afterEach(() => {
   clearVisualRefreshState();
 });
 
-const skippedComponents = ['button'];
+const skippedComponents = ['button', 'wizard'];
```
ktdotgit: FYI, before doing this... I tried setting the hidden `submit` button to `disabled` and setting `onClick` to `preventDefault()` but this test was still triggering it.

PR #3736, opened 2025-08-08
poc

pan-kot:
### Description

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/app-layout/visual-refresh-toolbar/drawer/local-drawer.tsx
```diff
@@ -145,7 +146,7 @@ export function AppLayoutDrawerImplementation({
               {toolsContent}
             </div>
             {activeDrawerId !== TOOLS_DRAWER_ID && (
-              <div className={styles['drawer-content']} style={{ blockSize: drawerHeight }}>
+              <div key={activeDrawerId} className={styles['drawer-content']} style={{ blockSize: drawerHeight }}>
```
pan-kot, line 149: Passing React key forces component re-mount when new drawer is rendered. This clears the Drawer's internal state including the error state of the built-in error boundary. Without this, if an error boundary is triggered in one drawer - it will be shown in all drawers.

src/container/internal.tsx
```diff
@@ -143,62 +153,71 @@ export default function InternalContainer({
...
-                    [styles['header-sticky-disabled']]: __stickyHeader && !isSticky,
-                    [styles['header-sticky-enabled']]: isSticky,
-                    [styles['header-dynamic-height']]: hasDynamicHeight,
-                    [styles['header-stuck']]: isStuck,
-                    [styles['with-paddings']]: !disableHeaderPaddings,
-                    [styles['with-hidden-content']]: !children || __hiddenContent,
-                    [styles['header-with-media']]: hasMedia,
-                    [styles['header-full-page']]: __fullPage && isRefresh,
-                  }
+        <BuiltInErrorBoundary wrapper={content => <InternalBox padding="m">{content}</InternalBox>}>
```
pan-kot, line 160: The container paddings are implemented separately in its header, body, and footer. When we replace all these slots with the error boundary fallback, these paddings are lost - that's why we wrap the fallback in a padded box.

src/container/internal.tsx
```diff
@@ -143,62 +153,71 @@ export default function InternalContainer({
...
+            <div
+              className={clsx(styles['content-inner'], testStyles['content-inner'], {
+                [styles['with-paddings']]: !disableContentPaddings,
+                [styles['with-header']]: !!header,
+              })}
+              style={getContentStyles(style)}
+            >
+              <BuiltInErrorBoundary
+                wrapper={content => (
+                  <InternalBox padding={disableContentPaddings ? 'm' : undefined}>{content}</InternalBox>
```
pan-kot, line 206: When `disableContentPaddings` is used, the container has no internal paddings around content, expecting the paddings to come from the content. However, the internal content paddings are lost in case there is an error in the content - that's why we add them back with a padded box around the fallback message.

src/error-boundary/internal.tsx
```diff
@@ -0,0 +1,125 @@
...
+    super(props);
+    this.state = { hasError: false };
+  }
+
+  static getDerivedStateFromError() {
+    return { hasError: true };
+  }
+
+  componentDidCatch(error: Error, errorInfo: React.ErrorInfo): void {
+    this.props.onError?.({ error, errorInfo, errorBoundaryId: this.props.errorBoundaryId });
```
jperals, line 115: Since the `onError` prop is mandatory, is the `?` necessary?

pages/error-boundary/error-boundary-fallbacks.page.tsx
```diff
@@ -0,0 +1,142 @@
...
+      <SpaceBetween size="m">{children}</SpaceBetween>
+    </SpaceBetween>
+  );
+}
+
+function Example({ code, ...props }: Omit<ErrorBoundaryProps, 'children' | 'onError'> & { code: string }) {
+  return (
+    <SpaceBetween size="xxs">
+      <Box variant="awsui-inline-code">{code}</Box>
+      <ErrorBoundary {...props} onError={() => {}}>
```
jperals, line 125: Would it be possible to call `error.preventDefault` here instead of adding an event listener to the window?

pages/error-boundary/error-boundary-fallbacks.page.tsx
```diff
@@ -0,0 +1,142 @@
...
+  SpaceBetween,
+  StatusIndicator,
+} from '~components';
+import { colorBackgroundStatusWarning } from '~design-tokens';
+
+import { SimplePage } from '../app/templates';
+
+// Mute console errors on this page to not trigger unexpected console error checks in e2e tests.
+window.addEventListener('error', e => e.preventDefault(), true);
+console.error = () => {};
```
jperals, line 23: Would it be possible to ignore errors at the test level instead?

src/container/internal.tsx
jperals: Thank you for these nice explanations. Why not incorporate them as code comments (maybe in a slightly more succinct form)?

src/error-boundary/fallback.tsx
```diff
@@ -0,0 +1,101 @@
...
+    : { hasFeedback: false, Feedback: () => <></> };
+
+  // This ensures that the description string provided via i18nStrings also supports the <Feedback> injection,
+  // because the i18n() helper propagates the second argument as is, without applying intl-format to it.
+  // We wrap the format with try-catch to avoid intl errors caused by incorrectly referenced components.
+  function safeFormat(descriptionText?: string) {
+    try {
+      return descriptionText ? new IntlMessageFormat(descriptionText).format(formatArgs) : undefined;
+    } catch {
+      return descriptionText;
```
jperals, line 80: Do we want to issue a warning in this case?

src/error-boundary/interfaces.ts
```diff
@@ -0,0 +1,82 @@
...
+   * * `descriptionText` (string): The text of the fallback message description. You can inject inline feedback action
+   * here by providing `components.Feedback`, and using <Feedback>custom text</Feedback> pseudo-tag in the text.
+   * * `refreshActionText` (string): The text of the default refresh action button.
+   * * `components.Feedback` (React.ComponentType<I18nFeedbackProps>): The inline action to be injected into the description text.
+   * @i18n
+   */
+  i18nStrings?: ErrorBoundaryProps.I18nStrings;
+
+  /**
+   * Content, rendered when no error is captured.
```
jperals: Minor: I think that comma after "Content" does not belong there.

src/error-boundary/internal.tsx
```diff
@@ -0,0 +1,125 @@
...
+  return (
+    <div ref={mergedRef} className={styles['error-boundary']} data-awsui-boundary-id={props.errorBoundaryId}>
+      {!thisSuppressed ? (
+        <ErrorBoundaryImpl {...props} forcedError={forcedError}>
+          <ErrorBoundariesContext.Provider value={{ ...props, suppressed: nextSuppressed }}>
+            {children}
+          </ErrorBoundariesContext.Provider>
+        </ErrorBoundaryImpl>
+      ) : (
+        <>{children}</>
```
jperals, line 68: Nitpick: could this

```
(
  <>{children}</>
)
```

be reduced to
```
children
```
?

src/error-boundary/internal.tsx
```diff
@@ -0,0 +1,125 @@
...
+  const thisSuppressed = context.suppressed === true || context.suppressed === RootSuppressed;
+  const nextSuppressed = suppressNested || thisSuppressed;
+  return !thisSuppressed ? (
+    <ErrorBoundaryImpl {...context} wrapper={wrapper}>
+      <ErrorBoundariesContext.Provider value={{ ...context, suppressed: nextSuppressed }}>
+        {children}
+      </ErrorBoundariesContext.Provider>
+    </ErrorBoundaryImpl>
+  ) : (
+    <>{children}</>
```
jperals, line 91: Same nitpick as above

src/i18n/messages/all.en.json
```diff
@@ -172,6 +172,11 @@
   "drawer": {
     "i18nStrings.loadingText": "Loading content"
   },
+  "error-boundary": {
```
jperals, line 178: What about the other locales?

src/container/internal.tsx
pan-kot: Good point - I will add the code comments 👍

src/error-boundary/internal.tsx
pan-kot: Great catch - I will update that.

pages/error-boundary/error-boundary-fallbacks.page.tsx
pan-kot: I didn't know about `error.preventDefault` - will research and refactor this code if it applies.

pages/error-boundary/error-boundary-fallbacks.page.tsx
pan-kot: Good point - I will check if we have a good way. Currently we do the logs checking inside the browser-test-tools. We can technically catch and filter the errors by content in the tests, but that can be dirty. There is no dedicated integ tests for this page, but it fails in the a11y tests that apply to all pages.

src/error-boundary/fallback.tsx
pan-kot: I don't think that it will be helpful: the warning will only appear if the error boundary fallback is triggered, which likely means that the component is being tested - at which point the problem should be apparent by looking at the rendered description component.

src/error-boundary/interfaces.ts
pan-kot: Thanks - will update and re-run the documenter.

src/error-boundary/internal.tsx
pan-kot: Thanks, I will check TS is happy with that and update if possible.

src/i18n/messages/all.en.json
pan-kot: Those were not ready until recently - I will contribute them as a separate PR and remove from here.

jperals:
Note: you might want to update the PR title

src/i18n/messages/all.en.json
pan-kot: The messages were already added to main, plus I created an additional PR to change the messages format so to match the current impl: https://github.com/cloudscape-design/components/pull/4088

pages/error-boundary/error-boundary-fallbacks.page.tsx
pan-kot: So apparently the preventDefault() cannot be used locally because there is no event to work with. The onError only takes the error obj, and it seems that the browser's "error" event can be only captured on the window or iframe level.

pages/error-boundary/error-boundary-fallbacks.page.tsx
pan-kot: I haven't found a sensible way - the logic is defined in the browser-test-tools and it takes no arguments currently to suspend that. The check is done by the useBrowser and after the test function - that still makes it possible to use the try-catch around it, but then one needs to ensure that the actual test failures are not suppressed and are reported in a readable way.

I suggest to keep it as is for now, but should we have more cases like this - we can modify the browser-test-tools to disabled the error check as needed.

src/error-boundary/internal.tsx
pan-kot: When this is removed it causes the TS error (in the consumer component):
```
  Its return type 'ReactNode' is not a valid JSX element.
    Type 'undefined' is not assignable to type 'Element | null'.

284           <BuiltInErrorBoundary>{children}</BuiltInErrorBoundary>
```

src/error-boundary/internal.tsx
```diff
@@ -0,0 +1,125 @@
...
+    super(props);
+    this.state = { hasError: false };
+  }
+
+  static getDerivedStateFromError() {
+    return { hasError: true };
+  }
+
+  componentDidCatch(error: Error, errorInfo: React.ErrorInfo): void {
+    this.props.onError({ error, errorInfo, errorBoundaryId: this.props.errorBoundaryId });
```
jperals: [This change](https://github.com/cloudscape-design/components/pull/3736/commits/dbaf0f5610702254c2ba69bd56574262e58315f2#diff-c98418fe14ea739d043491def8b153fb6de2a73e64da34c44c45c2dde6fd2df5R115) breaks the unit tests below, which assert we are permissive and don't crash if not providing the `onError` prop (even if it is required)

PR #3737, opened 2025-08-08
feat: Add Inline label text to multiselect

mariksav:
### Description

Customers requested the `inlinLabelText` property, which was already available in select, to be also in multiselect. Implemented it + new dev page + new unit tests.
Related links, issue #, if available: AWSUI-60534

### How has this been tested?

New unit test and integration tests.

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/multiselect/__tests__/multiselect.test.tsx
```diff
@@ -155,6 +155,21 @@ test('does not render tokens when no tokens are present', () => {
   expect(wrapper.findTokens()).toHaveLength(0);
 });
 
+test('renders inline label text when provided', () => {
+  const { wrapper } = renderMultiselect(
+    <Multiselect selectedOptions={[]} options={defaultOptions} {...({ inlineLabelText: 'Select items:' } as any)} />
+  );
+  const inlineLabel = wrapper.findByClassName(selectPartsStyles['inline-label']);
```
jperals: Why are we not introducing a new `findInlineLabel` test util as in Select?

src/multiselect/__tests__/multiselect.test.tsx
```diff
@@ -155,6 +155,21 @@ test('does not render tokens when no tokens are present', () => {
   expect(wrapper.findTokens()).toHaveLength(0);
 });
 
+test('renders inline label text when provided', () => {
+  const { wrapper } = renderMultiselect(
+    <Multiselect selectedOptions={[]} options={defaultOptions} {...({ inlineLabelText: 'Select items:' } as any)} />
```
jperals: Why the destructuring and the `as any`? Since `inlineLabelText` is part of the component interface, I think these shouldn't be necessary.

src/multiselect/__tests__/multiselect.test.tsx
```diff
@@ -155,6 +155,21 @@ test('does not render tokens when no tokens are present', () => {
...
+  );
+  const inlineLabel = wrapper.findByClassName(selectPartsStyles['inline-label']);
+  expect(inlineLabel).not.toBeNull();
+  expect(inlineLabel!.getElement()).toHaveTextContent('Select items:');
```
jperals, line 164: In the Select component we also test the `for` attribute of the label: https://github.com/cloudscape-design/components/blob/8767da0142e31201ac776297aaf9ff937ebbb8dd/src/select/__tests__/select.test.tsx#L425

I think it would be good to have this here as well.

src/select/parts/styles.scss
```diff
@@ -147,7 +147,7 @@ $inlineLabel-border-radius: 2px;
   position: relative;
   inset-inline-start: calc(awsui.$border-width-field + awsui.$space-field-horizontal - awsui.$space-scaled-xxs);
   margin-block-start: awsui.$space-scaled-xs;
-  padding-block-end: 2px;
```
jperals: Why is this necessary? Isn't it affecting the style of the inline label in existing places (e.g, Select)?

src/select/parts/styles.scss
```diff
@@ -153,6 +153,11 @@ $inlineLabel-border-radius: 2px;
   z-index: 1;
 }
 
+.inline-label-multiselect {
```
jperals: To make the reason of this custom style more obvious I would suggest either changing the selector to something like `inline-label-inline-tokens` (because it is specific to using multiselect's inline tokens, not just multiselect) or adding an explanatory comment —or both.

src/select/parts/styles.scss
```diff
@@ -153,6 +153,11 @@ $inlineLabel-border-radius: 2px;
...
+  padding-block-end: 0;
```
jperals, line 158: Not sure we want to go in so much detail, but note that in compact mode, the inline label still overlaps the inline tokens by 1px:
Screenshot 2025-08-12 at 11 19 13

Actually there are also cases where the inline label overlaps content in Select (even in comfortable mode: [link](https://d21d5uik3ws71m.cloudfront.net/components/d9414731186cc161d81c9fb1a3ac4e74fe082056/dev-pages/index.html#/light/select/trigger.permutations)):
Screenshot 2025-08-12 at 11 26 23

So I wonder if this padding could be removed or at least reduced for all cases, not only this one. We might get into situations where the lower edge of the character (for example, a "g" or a "q") sticks out of the label background, but IMO that is preferable to the label background overlapping option content in the select or multiselect.

mariksav: Thats what I mentioned to Timo as well on Monday in the meeting and that it would make sense to apply the change  to select as well. But the large icon variant is supposed to be removed from select anyways so this problem will be removed implicitly. For multiselect i can increase the gap between inlineLabelText and token, but at some point the inlineLabelText is just going to be above the button and not inside any more. Thats why I chose this position.

jperals: OK, makes sense.

If you want to try, it should be possible to clip the background without moving the text with something like:
```
background-repeat: no-repeat;
background-size: 100% calc(100% - 2px);
```

(in this case, the background would be clipped at the bottom by 2px).

I would be OK with the changes as they are though.

mariksav: I increased the distance a little bit, while keeping the text on the line.

PR #3752, opened 2025-08-13
chore: Button style api refactor

teodoranemes:
### Description

Added `style` file for the Button component to maintain consistency across all Style API Components.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/button/style.tsx
```diff
@@ -0,0 +1,44 @@
...
+import customCssProps from '../internal/generated/custom-css-properties';
+import { ButtonProps } from './interfaces';
+
+export function getButtonStyles(style: ButtonProps['style']) {
+  if (SYSTEM !== 'core' || !style?.root) {
+    return undefined;
+  }
+
+  return {
+    borderRadius: style?.root?.borderRadius,
```
avinashbot, line 13: If you removed the possibility for `style` or `style.root` to be nullable in the condition earlier, TypeScript will let you remove all the optional chaining operators (`?.`) here, which would be much cleaner.

src/button/style.tsx
```diff
@@ -0,0 +1,44 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import { SYSTEM } from '../internal/environment';
+import customCssProps from '../internal/generated/custom-css-properties';
+import { ButtonProps } from './interfaces';
+
+export function getButtonStyles(style: ButtonProps['style']) {
+  if (SYSTEM !== 'core' || !style?.root) {
```
avinashbot, line 8: Minor: just out of curiosity, I know we have this for bundlers/minifiers to remove, but I don't know how complex this is allowed to be. For example, are you allowed to combine it with other conditionals like this?

~~Do you know if there's a way to test this?~~ Got curious, did some research:

- Webpack uses Terser, which removes replaced conditionals
  - https://webpack.js.org/guides/production/
  - https://terser.org/docs/miscellaneous/#conditional-compilation
- Vite uses esbuild, which does not
  - https://vite.dev/config/build-options.html#build-minify
  - https://esbuild.github.io/api/#minify-considerations:~:text=The%20minification%20algorithm%20in%20esbuild%20does%20not%20yet%20do%20advanced%20code%20optimizations

Oh well, probably need to come up with something smarter eventually. But that's for later, this isn't a valid reason to block this PR.

avinashbot (approved):
Just a minor style comment about the usages of `?.` (approving anyway since it's just a style thing). Other comment is just for me.

src/button/style.tsx
teodoranemes: Thank you for the suggestion!

PR #3758, opened 2025-08-14
fix: Fix focus management for flashbar when motion is enabled

avinashbot:
### Description

I'll be honest here, I didn't feel like diving into the weird transition/animation logic going on with the flashbar components. So I just changed the throttle logic to do a debounce instead (so it always waits for the animation to complete while still preventing repeated flashbar insertions from annoying users). Sue me.

Actually, if new flash items go from bottom to top now (as they do with collapsible flashbars), debounce is actually better. So yay me, actually!

Related links, issue #, if available: AWSUI-61128

### How has this been tested?

Added unit tests. But also tested locally in a dev page. Try it out yourself — go to the `flashbar/interactive` page, enable stack items, and add an error flash. The focus move to the newly added flash message should work (it didn't before).

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/flashbar/__integ__/pages/base.ts
+++ b/src/flashbar/__integ__/pages/base.ts
@@ -12,7 +12,6 @@ export class FlashbarBasePage extends BasePageObject {
   async toggleCollapsedState() {
     const selector = this.getNotificationBar();
     await this.click(selector);
-    await this.pause(500);
   }
 
   countFlashes() {
--- a/src/flashbar/__integ__/pages/base.ts
+++ b/src/flashbar/__integ__/pages/base.ts
@@ -12,6 +12,7 @@ export class FlashbarBasePage extends BasePageObject {
   async toggleCollapsedState() {
     const selector = this.getNotificationBar();
     await this.click(selector);
+    await this.pause(500);
   }
 
   countFlashes() {
--- a/src/flashbar/utils.ts
+++ b/src/flashbar/utils.ts
@@ -3,7 +3,7 @@
 import { IconProps } from '../icon/interfaces';
 import { FlashbarProps } from './interfaces';
 
-export const FOCUS_THROTTLE_DELAY = 2000;
+export const FOCUS_DEBOUNCE_DELAY = 500;
 
 // Since the position of a notification changes when the Flashbar is collapsed,
 // it is useful on some situations (e.g, for animating) to know the original position of the item
--- a/src/flashbar/__integ__/collapsible.test.ts
+++ b/src/flashbar/__integ__/collapsible.test.ts
@@ -119,6 +119,7 @@ describe('Collapsible Flashbar', () => {
           await page.toggleStackingFeature();
           await page.addErrorFlash();
           await page.toggleCollapsedState();
+          await page.pause(FOCUS_DEBOUNCE_DELAY);
           await expect(page.isFlashFocused(1)).resolves.toBe(true);
         })
       );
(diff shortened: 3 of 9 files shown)


src/flashbar/__integ__/pages/base.ts
```diff
@@ -12,6 +12,7 @@ export class FlashbarBasePage extends BasePageObject {
   async toggleCollapsedState() {
     const selector = this.getNotificationBar();
     await this.click(selector);
+    await this.pause(500);
```
jperals: Is this pause actually needed? The component should work correctly even without the user waiting before interacting.

src/flashbar/utils.ts
```diff
@@ -3,7 +3,7 @@
 import { IconProps } from '../icon/interfaces';
 import { FlashbarProps } from './interfaces';
 
-export const FOCUS_THROTTLE_DELAY = 2000;
```
jperals, line 6: Does this mean that previously flash announcements would be throttled if they were 2 seconds apart or less, and now they will be debounced only if they are half a second apart or less? Can this be detrimental to user experience?

src/flashbar/__integ__/pages/base.ts
avinashbot: The component does work fine without waiting; it's just that the focus states will wait for animations to complete. The pause is only for whenever we test focus states.

I could disable animations, but they're complex javascript-driven animations so I want them to be covered.

src/flashbar/utils.ts
avinashbot: Yes, but I think this is necessary The way throttle works is that leading calls are not delayed, but any calls during that throttle period are deferred.

```
0---x----x----x----1---------x--------2 seconds
^                  ^                  ^
          "accepted" calls
```

This means that the first flash message is immediately focused, and only repeated calls within 2 seconds are skipped. But with debounce, even the first call is delayed.

```
0---x----x----x----1------------------2 seconds
                             ^
          "accepted" calls
```

So waiting 2 seconds from the flash addition until focus is moved is a bit too long IMO.

src/flashbar/__integ__/pages/base.ts
jperals: If the pause is only for when we test focus states, wouldn't it be better to keep it in the specific test where it was, rather than in the method?

Right now, there is only one place where this method is used, but we don't know if there will be future usages it will be preferable to not wait (and it might not even be obvious that the pause is there).

src/flashbar/utils.ts
jperals: OK, so the point is, that we don't know in advance if the first item to arrive will be the one we want to focus on, so we need to "wait" and see if more will arrive in short time. And for this first pause 2 seconds is too long.

On the other hand, it is still true that for the subsequent ones, we will "jump" to the next even if they are 500ms apart, potentially without enough time for the announcement to read it all. Is this a concern?

src/flashbar/__integ__/pages/base.ts
avinashbot: This just seemed cleaner because I saw myself copy pasting this over and over again in a dozen different tests while making this change. No strong feeling either way; I can move them back to the tests.

src/flashbar/utils.ts
avinashbot: This _is_ a concern; but I think 500ms is the best middle ground. Anything over 1s is too much. Anything under 200ms is pointless (IMO) Unless we create an unholy abomination that's a combination of debouncing and throttling.

src/flashbar/__integ__/pages/base.ts
avinashbot: Spoke in Slack, I misunderstood — thought this comment was referring to the whole approach, not this specific wait. Undid my changes in response to this and just removed this wait.

jperals (approved):

PR #3760, opened 2025-08-15
chore: Add util function to handle all inProgress to ‘in-progress’ use cases

teodoranemes:
### Description

Added util function to handle all inProgress to ‘in-progress’  use cases for the Flashbar component.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/flashbar/style.tsx
```diff
@@ -71,29 +70,27 @@ export function getDismissButtonStyles(style: FlashbarProps['style'], type: stri
   }
   const activeColor =
```
patrixd, line 66: Here we are using `getStylePropertyKey(type as FlashbarProps.Type)` 4 times in the same function. I suggest to move it to a variable before this line so the code is clearer to read.

src/flashbar/style.tsx
```diff
@@ -3,22 +3,23 @@
 import { SYSTEM } from '../internal/environment';
 import customCssProps from '../internal/generated/custom-css-properties';
 import { FlashbarProps } from './interfaces';
+import { getStylePropertyKey } from './utils';
 
 export function getCollapsibleFlashStyles(style: FlashbarProps['style'], type: string = 'info') {
   if (SYSTEM !== 'core' || !style) {
     return undefined;
   }
   const background =
```
patrixd, line 13: Here we are using `getStylePropertyKey(type as FlashbarProps.Type)` 3 times in the same function. I suggest to move it to a variable before this line so the code is clearer to read.

src/flashbar/utils.ts
```diff
@@ -114,6 +114,10 @@ type LabelName =
   | 'infoIconAriaLabel'
   | 'inProgressIconAriaLabel';
 
+export function getStylePropertyKey(type: FlashbarProps.Type): string {
```
patrixd, line 117: I wonder if we should create a function to get the style so there is cleaner code

Example:
```
getStyle(style?.item?.dismissButton?.focusRing?.borderColor, type);

export function getStyle(...){
    return property &&
    (type === 'in-progress'
      ? property['inProgress']
      : property[type]);
```

patrixd (dismissed):
Approved with suggestions of improvement to make the code more readable.

src/flashbar/style.tsx
teodoranemes: Added a variable to make the code more readable, thank you for the suggestion!

src/flashbar/style.tsx
teodoranemes: Added it also for this function.

patrixd (approved):

PR #3761, opened 2025-08-15
refactor: Refactored test pages for date range picker

pan-kot:
### Description

The PR makes a big refactoring of the date range picker test pages. Main changes:
1. Introduced page templates (those are helpful for all components, so created them under ./app)
2. Removed unneeded simple page (the with-value page is literally the same)
3. Simplified localized page by making use of the built-in i18n provider
4. Rewrote all custom component settings in favour of a standartized settings, which can be reused on every page
5. Removed irrelevant props in favour of component defaults or default settings

UI changes:

https://github.com/user-attachments/assets/07525e65-2db4-4226-b0d0-f16fd104046d

To reviewers: I suggest to ignore the diffs to the most part as all changes that can affect automated tests are verified anyways by the tests. It is better to focus on the new code instead, see if the updated pages and the setup are viable.

### How has this been tested?

* Existing integration tests still work
* Screenshot tests are validated in the dev pipeline

<details>

#### Correctness

#### Security

#### Testing

</details>

---


pages/date-range-picker/absolute-format-all-locales.page.tsx
```diff
@@ -0,0 +1,33 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
pan-kot, line 1: This file is renamed and refactored `pages/date-range-picker/absolute-format.localization.page.tsx`

pages/date-range-picker/absolute-format-all-locales.page.tsx
```diff
@@ -0,0 +1,33 @@
...
+  const { props, settings, setSettings } = useDateRangePickerSettings({ rangeSelectorMode: 'absolute-only' });
+  return (
+    <SimplePage
+      title="Date range picker: all locales for absolute format"
+      settings={<Settings settings={settings} setSettings={setSettings} />}
+    >
+      {locales.map(locale => (
+        <div key={`pickers-${locale}`} dir={rtlLocales.has(locale) ? 'rtl' : 'ltr'}>
+          <Grid gridDefinition={[{ colspan: 1 }, { colspan: 11 }]}>
+            <div style={{ textAlign: 'right' }}>{locale}</div>
```
mariksav, line 24: I see you used inline styling at multiple places and that inline styling was already in the code without your changes. I remember Avinash once told me inline styling should be avoided if necessary. Is that a concern here or is it possible to be do it differently? Or is it necessary?

pages/date-range-picker/localised.page.tsx
```diff
@@ -76,78 +21,26 @@ const localizedErrors = {
...
-        Date-only
-      </Checkbox>
-      <Checkbox checked={monthOnly} onChange={({ detail }) => setUrlParams({ monthOnly: detail.checked })}>
-        Month-only
-      </Checkbox>
-      <br />
+    <SimplePage
+      title="Date range picker: localized"
+      settings={<Settings settings={settings} setSettings={setSettings} />}
+      i18n={{ messages: [messages], locale: 'de-DE' }}
```
mariksav, line 29: Very nitpicky and probably not part of your task (as it was apparently already convention), but would it not make sense to format the date in the local (german) dev page in the german date format (Day/ Month/ Year) instead of Year/ Month /Day. Just a general question and obviously not blocking.

pages/date-range-picker/month-calendar-permutations.page.tsx
```diff
@@ -34,66 +34,57 @@ const permutations = createPermutations<DateRangePickerCalendarProps>([
...
-        </div>
-      </ScreenshotArea>
-    </Box>
+    <PermutationsPage title="Date range picker permutations: month calendar" i18n={{}}>
+      <PermutationsView
+        permutations={permutations}
+        render={permutation => {
+          i++;
+          return (
+            <div style={{ insetBlockStart: `${i * 400}px`, position: 'relative' }}>
```
mariksav, line 70: Wouldn’t it be nicer to have this a const for this (400)?

pages/date-range-picker/year-calendar-permutations.page.tsx
```diff
@@ -30,58 +30,52 @@ const permutations = createPermutations<DateRangePickerCalendarProps>([
...
-                  </Dropdown>
-                </div>
-              );
-            }}
-          />
-        </div>
-      </ScreenshotArea>
-    </Box>
+    <PermutationsPage title="Date range picker permutations: year calendar" i18n={{}}>
+      <div style={{ blockSize: `${(1 + permutations.length) * 300}px` }}>
```
mariksav, line 54: Wouldn’t it be nicer to have this a const for this (300)?

pages/date-range-picker/absolute-format-all-locales.page.tsx
pan-kot: I think it is fine to keep inline styles in the test pages, but we should avoid that in components, unless the styles need to be computed.

pages/date-range-picker/localised.page.tsx
pan-kot: That part is explicitly out of scope for now, see here: dp7CACkt4mzI.

pages/date-range-picker/month-calendar-permutations.page.tsx
pan-kot: I would say it is not necessary here as it is a test page, and the 400px represents a workaround we use to ensure the dropdown size can fit into the screenshot. If we add more content to the dropdown in the future - we'd go and update this to a bigger value.

pages/date-range-picker/year-calendar-permutations.page.tsx
pan-kot: Same as above. While we can introduce a const - that won't change much, as the value is empirical anyways. You can also see how it is different between pages, so I reusable one won't work either.

PR #3773, opened 2025-08-20
chore: Removes timeout wait from property filter unit tests

pan-kot:
### Description

Refactored tests to avoid timeouts. It is also possible to use fake timers for that, but the custom signal solutions looks cleaner imo.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/property-filter/__tests__/property-filter-stories-token-create-async.test.tsx
```diff
@@ -37,15 +37,15 @@ const defaultProps: Partial<PropertyFilterProps> = {
...
+  test('waits and selects property, then waits and selects option', () => {
     const { wrapper } = render({ asyncProperties: true });
 
     wrapper.input.focus();
     expect(wrapper.input.dropdown()).toBe(true);
     expect(wrapper.input.options()).toEqual([]);
     expect(wrapper.input.status()).toBe('Loading status');
 
-    await new Promise(resolve => setTimeout(resolve, 1));
+    window.loadingComplete();
```
avinashbot: A couple of things:

1. This spits out `act(() => ...)` warnings because this callback updates a component.
2. Couldn't you do something like `await waitFor(() => { ...assertions... })` instead? Would make the tests a little less custom.

I'm good with either, just curious why you went this route.

pan-kot: Good point. The route I've taken is the same as we do in the integ tests for several components. However, I do like the waitFor option in this case. As these are the unit tests, there should be no flakiness from that caused by too small delay. I will try it out.

src/property-filter/__tests__/stories-components.tsx
```diff
@@ -72,7 +72,7 @@ function StatefulAsyncPropertyFilter(props: PropertyFilterProps) {
       setFilteringStatusType('finished');
       loadProperties(detail.filteringText);
       loadOptions(detail.filteringProperty, detail.filteringText);
-    }, 1);
+    }, 0);
```
avinashbot, line 75: Technically, you can just remove this number and it should just run on the next microtask :)

avinashbot (approved):
Great, thanks for addressing the comment!

src/property-filter/__tests__/stories-components.tsx
pan-kot: Good point!

PR #3779, opened 2025-08-20
chore: Remove all legacy markdown from Style API interfaces

DaemonCahill:
This PR removes all of the legacy markdown for the Style API that was replaced by documenter automation.


teodoranemes (commented):
Approved, the only markdown left is in the chat components package for the Avatar Style API.

teodoranemes (approved):

PR #3781, opened 2025-08-21
chore: Nested STSN providers support

pan-kot:
### Description

The PR makes it possible to nest STSN providers. The use cases that we have are:
1. Rendering a button group inside table cells;
2. Rendering a button group inside tree view items (we are adding keyboard nave in there, too).

The idea of the change is rather simple: at any given point in time only one provider can be active. The active provider is the outer-most provider with navigationActive=true. If the outer provider changes state from true to false - the nested provider becomes active instead. This allows a dynamic switch of navigation contexts (e.g. when a tree toggle is focused - the tree navigation is engaged; when the focus is inside a tree item - the button group which can be rendered inside receives control). In some cases, it is not even necessary to do the switching. Thus, the table navigation controller is already capable of handling the nested button group navigation on its own, the only difference is that the button group focus is no longer looped.

https://github.com/user-attachments/assets/8fb839f5-a5d4-4fe9-b254-649cd9a1bec6

Rel: AWSUI-61152, [AaSEAuz8NXla]

### How has this been tested?

* Added new unit tests for nested STSN providers
* Enhanced test page for table navigation

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/table-fragments/grid-navigation-custom.page.tsx
+++ b/pages/table-fragments/grid-navigation-custom.page.tsx
@@ -7,6 +7,7 @@ import {
   AppLayout,
   Button,
   ButtonDropdown,
+  ButtonGroup,
   Checkbox,
   CollectionPreferences,
   ColumnLayout,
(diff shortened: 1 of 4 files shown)


gethinwebster:
Would this be a good time to also centralize these two implementations? https://github.com/cloudscape-design/component-toolkit/tree/main/src/internal/single-tab-stop

pan-kot:
> Would this be a good time to also centralize these two implementations? https://github.com/cloudscape-design/component-toolkit/tree/main/src/internal/single-tab-stop

Let's merge this one first as a dependency for the tree view impl. I will follow it up by opening a pr to the toolkit and then later once toolkit changes are released - replacing the components based implementation with the toolkit imports.

Edit: created PR: https://github.com/cloudscape-design/component-toolkit/pull/155

gethinwebster (approved):

PR #3785, opened 2025-08-22
chore: Add navigable-group

gethinwebster:
### Description

Add navigable group component, as a more abstract version of button group.

Related links, issue #, if available: AaSEAuz8NXla

### How has this been tested?

New unit and integ tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/navigable-group/interfaces.ts
```diff
@@ -0,0 +1,42 @@
...
+export interface NavigableGroupProps extends BaseComponentProps {
+  /**
+   * The children that will be rendered inside the navigation group.
+   *
+   * The following Cloudscape components are supported, but do not
+   * need to be direct children of the group:
+   * - Button
+   * - Button dropdown
+   * - Toggle button
+   * - Link
```
johannes-weber, line 16: The example page do contain a Checkbox component as well - I think Checkbox is missing here, isn't it?

src/navigable-group/__tests__/navigable-group.test.tsx
```diff
@@ -0,0 +1,310 @@
...
+      button3.focus();
+      button3.keydown(KeyCode.right);
+      expect(button1.getElement()).toHaveFocus();
+      button1.keydown(KeyCode.left);
+      expect(button3.getElement()).toHaveFocus();
+    });
+
+    test('works with custom getItemKey function', () => {
+      const { container } = render(
+        <NavigableGroup getItemKey={element => element.getAttribute('data-custom-id') || element.id}>
```
johannes-weber: Nitpick: I think the `|| element.id` default could be removed because it's not used.

src/navigable-group/internal.tsx
```diff
@@ -0,0 +1,168 @@
...
+        if (target) {
+          target.focus();
+        }
+      },
+    }));
+
+    function getNextFocusTarget(): null | HTMLElement {
+      if (containerObjectRef.current) {
+        const focusables = getFocusablesFrom(containerObjectRef.current);
+        return focusables.find(button => getItemKey(button) === focusedIdRef.current) ?? focusables[0] ?? null;
```
johannes-weber:
```suggestion
        return focusables.find(el => getItemKey(el) === focusedIdRef.current) ?? focusables[0] ?? null;
```

Nitpick: I think `button` is misleading because it could be any other component as well.

src/navigable-group/internal.tsx
```diff
@@ -0,0 +1,168 @@
...
+      if (!containerObjectRef.current || !focusTarget) {
+        return;
+      }
+      // Ignore navigation when the focused element is not a registered focusable.
+      if (document.activeElement && !navigationAPI.current?.isRegistered(document.activeElement)) {
+        return;
+      }
+      event.preventDefault();
+
+      const focusables = getFocusablesFrom(containerObjectRef.current);
```
johannes-weber, line 108: Do we know that `focusables` will be always an non-empty array? I'm wondering if there should add an additional early return in case of it's empty. E.g. I've elements in the navigable group got disabled while navigating.

src/navigable-group/internal.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+    function getNextFocusTarget(): null | HTMLElement {
+      if (containerObjectRef.current) {
+        const focusables = getFocusablesFrom(containerObjectRef.current);
+        return focusables.find(button => getItemKey(button) === focusedIdRef.current) ?? focusables[0] ?? null;
+      }
+      return null;
+    }
+
+    function onUnregisterActive(focusableElement: HTMLElement) {
```
johannes-weber, line 57: Missing unit test: I don't see a unit test for the use-case when an item gets removed from the component.

johannes-weber (changes requested):

src/navigable-group/interfaces.ts
gethinwebster: checkbox _can_ be used, but it's not currently quite so well supported (as it's hard to set data- attributes on the focusable element) so I left it out of the official docs for now

src/navigable-group/internal.tsx
gethinwebster: https://github.com/cloudscape-design/components/pull/3785/files#diff-547bd0732a7f5dde74c85be4380ee3f4df6b5f26367733d70cffc134d5f4c3a6R106

src/navigable-group/__tests__/navigable-group.test.tsx
gethinwebster: done

src/navigable-group/internal.tsx
gethinwebster: updated

src/navigable-group/internal.tsx
gethinwebster: yes: there is a check earlier whether there is an active focus target, which by definition is one of `focusables`. So, if `focusables` is an empty array, this handler will already have exited earlier.

src/test-utils/dom/navigable-group/index.ts
```diff
@@ -0,0 +1,9 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import { ElementWrapper } from '@cloudscape-design/test-utils-core/dom';
+
+import styles from '../../../navigable-group/test-classes/styles.selectors.js';
+
+export default class NavigableGroupWrapper extends ElementWrapper {
```
just-boris:
```suggestion
export default class NavigableGroupWrapper extends ComponentWrapper {
```

johannes-weber (approved):

PR #3788, opened 2025-08-24
perf: Remove outdated array ponyfills

TrevorBurnham:
All target browsers have long supported the array functions `findIndex`, `find`, and `includes`. This PR removes "ponyfills" that redefine those functions.

MDN:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex#browser_compatibility

### How has this been tested?

✅ All unit tests pass.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/hooks/forward-focus/radio-group.ts
+++ b/src/internal/hooks/forward-focus/radio-group.ts
@@ -16,7 +16,7 @@ export default function useRadioGroupForwardFocus(
   value: string | null
 ): [React.Ref<HTMLInputElement>, number] {
   const itemRef = useRef<HTMLInputElement | null>(null);
-  const itemIndex = items && findIndex(items, item => item.value === value);
+  const itemIndex = items && items.findIndex(item => item.value === value);
   useImperativeHandle(forwardedRef, () => ({
     focus() {
       itemRef.current?.focus();
@@ -24,13 +24,4 @@ export default function useRadioGroupForwardFocus(
   }));
 
   return [itemRef, itemIndex !== undefined && itemIndex !== -1 ? itemIndex : 0];
-}
-
-function findIndex<T>(items: ReadonlyArray<T>, predicate: (t: T) => any): number {
-  for (let i = 0; i < items.length; i++) {
-    if (predicate(items[i])) {
-      return i;
-    }
-  }
-  return -1;
 }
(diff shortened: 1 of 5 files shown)


src/tag-editor/__tests__/tag-editor.test.tsx
```diff
@@ -713,5 +713,26 @@ describe('Tag Editor component', () => {
...
+    test('should call focus method on ref with error tags', () => {
+      const TestComponent = () => {
+        const ref = React.useRef<TagEditorProps.Ref>(null);
+        return (
+          <>
+            <TagEditor
+              ref={ref}
+              i18nStrings={i18nStrings}
+              tags={[{ key: 'aws:invalid', value: '', existing: false }]}
+              onChange={() => {}}
```
Harsh-Anand-Singh, line 726: The test includes `onChange={() => {}}` which creates a non-interactive controlled component since the handler doesn't update state.

TrevorBurnham: We don't care about whether the component is interactive or not in this test. We're only interested in whether the focus method gets called.

TrevorBurnham:
I've rebased against `main` and added more tests to satisfy Codecov.

TrevorBurnham:
Update: I've added another test case that should fill in the missing coverage for real this time.

Harsh-Anand-Singh (dismissed):

src/tag-editor/__tests__/tag-editor.test.tsx
```diff
@@ -713,5 +713,118 @@ describe('Tag Editor component', () => {
...
+            />
+            <button onClick={() => ref.current?.focus()}>Focus</button>
+          </>
+        );
+      };
+
+      const { container } = render(<TestComponent />);
+      const wrapper = createWrapper(container).findTagEditor()!;
+      const button = container.querySelector('button')!;
```
YueyingLu: This is querying the first button in the container, in this case is the X button in the input. Better to add an id to the button or move it to above the tageditor @TrevorBurnham 

Same as the test below.

YueyingLu:
Thanks for making this refactoring! @TrevorBurnham

src/tag-editor/__tests__/tag-editor.test.tsx
TrevorBurnham: Done: d6aee14

PR #3792, opened 2025-08-24
feat: Replace use of :focus-visible shim

TrevorBurnham:
All targeted browsers now support `:focus-visible`: https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible#browser_compatibility

This commit removes the use of [the shim defined in @cloudscape-design/component-toolkit](https://github.com/cloudscape-design/component-toolkit/tree/c338b2df31e59bc08578c8f9bf2ff182dc8d70c8/src/internal/focus-visible). Once this change is merged, the shim can be removed.

## Benefits

This change would reduce CSS and JS bundle sizes by ~1KB. It would also eliminate the global `mousedown`/`keydown` event listeners required by the shim, which should improve runtime performance.

Additionally, following the browser's native `:focus-visible` behavior is arguably an a11y improvement.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/components/abstract-switch/styles.scss
+++ b/src/internal/components/abstract-switch/styles.scss
@@ -5,7 +5,6 @@
 
 @use '../../styles' as styles;
 @use '../../styles/tokens' as awsui;
-@use '@cloudscape-design/component-toolkit/internal/focus-visible' as focus-visible;
 
 .content,
 .description,
@@ -25,7 +24,7 @@
 }
 
 .native-input {
-  @include focus-visible.when-visible {
+  &:focus-visible {
     & + .outline {
       display: block;
     }
@@ -70,6 +69,7 @@
   /* stylelint-disable selector-max-type */
   & > input,
   & > svg,
+  /* stylelint-disable-next-line no-descending-specificity */
   & > .outline {
     position: absolute;
     inline-size: 100%;
--- a/src/alert/styles.scss
+++ b/src/alert/styles.scss
@@ -6,7 +6,6 @@
 @use 'sass:map';
 @use '../internal/styles/tokens' as awsui;
 @use '../internal/styles' as styles;
-@use '@cloudscape-design/component-toolkit/internal/focus-visible' as focus-visible;
 @use '../internal/generated/custom-css-properties/index.scss' as custom-props;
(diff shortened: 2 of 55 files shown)


src/internal/components/abstract-switch/styles.scss
```diff
@@ -65,25 +79,10 @@
...
-    z-index: 1;
-    user-select: none;
-    cursor: default;
-    margin-block: 0;
-    margin-inline: 0;
+.native-input {
+  &:focus-visible {
+    & + .outline {
+      display: block;
+    }
```
TrevorBurnham: The reordering here is to satisfy stylelint's `no-descending-specificity` rule. Without the order change, stylelint complains:

```
Expected selector "& > .outline" to come before 
     selector ".native-input:focus-visible + .outline"  
     no-descending-specificity
```

TrevorBurnham: On second thought, I've restored the original ordering here to prevent possible changes in behavior.

TrevorBurnham:
I've updated this branch to take a more careful approach. Notably, the shim had greater specificity (two class selectors instead of one).

In one case I've used a double-specifier (`:focus-visible:focus-visible`) to ensure that the styles take precedence as intended. In other cases I _think_ it should be fine to just use `:focus-visible`, but unexpected style interactions are always possible. If you really want to err on the side of caution, you could use `:focus-visible:focus-visible` consistently to preserve the existing specificity.

TrevorBurnham:
@avinashbot Ah, sorry that I missed those earlier PRs!

The behavior in this PR would be the same: `:focus-visible` by its nature delegates to the browser's own heuristics. That's good from an a11y standpoint, but I agree that it can be jarring if it means that an outline appears because of a programmatic focus change triggered by a mouse event.

Can you tell me exactly what behavior you're seeing in Safari? Like, does the close button on a modal get visible focus after that modal was opened via mouse click?

TrevorBurnham:
I've updated the branch and added a list of benefits to the PR description. Performance-wise, the benefits are more modest than I'd have expected. I think it really comes down to whether you want the shim's behavior or the native `:focus-visible` behavior.

PR #3793, opened 2025-08-25
fix: Fix focus management for flashbar when motion is enabled

avinashbot:
Reverts cloudscape-design/components#3782.


johannes-weber:
From the revert: 
> Adding a surprise focus move where it was previously broken turns out to perhaps be a bad idea. Will revisit tomorrow.

Why did we re-add it? What did you find out about "perhaps be a bad idea" from the revert?

johannes-weber (approved):

PR #3796, opened 2025-08-25
chore: consistently use direct imports for Lodash functions

TrevorBurnham:
This commit replaces barrel imports like

```
import { range } from 'lodash'
```

with direct imports like

```
import range from 'lodash/range'
```

All of these imports are in test code (Lodash is a devDependency only), so this won't affect the build. But it will make code style more consistent, and it should slightly improve test performance.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/__a11y__/to-validate-a11y.ts
+++ b/src/__a11y__/to-validate-a11y.ts
@@ -3,7 +3,8 @@
 
 import Axe from 'axe-core';
 import { HtmlValidate } from 'html-validate';
-import { compact, uniq } from 'lodash';
+import compact from 'lodash/compact';
+import uniq from 'lodash/uniq';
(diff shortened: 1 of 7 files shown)


src/__a11y__/to-validate-a11y.ts
```diff
@@ -3,7 +3,8 @@
 
 import Axe from 'axe-core';
 import { HtmlValidate } from 'html-validate';
-import { compact, uniq } from 'lodash';
+import compact from 'lodash/compact';
```
pan-kot, line 6: Should we add an ESLint rule for pages to reinforce this?

pan-kot (approved):

PR #3804, opened 2025-08-27
feat: Standalone Token

ernst-dev:
### Description

Adds a new component, Token. This is an atomic component split-out from TokenGroup for compact data representation. It is implemented in TokenGroup itself as well as Multiselect using both the `normal` and `inline` variants.

`zydLArebErmk`
`akrbA0gbwxcl`

### How has this been tested?

Added unit and integration tests, a simple dev page as well as a permutations page.

This can be tested on the dev pages for Token, TokenGroup and Multiselect.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/test-utils/dom/token-group/token.ts
+++ b/src/test-utils/dom/token-group/token.ts
@@ -20,6 +20,6 @@ export class TokenGroupItemWrapper extends ComponentWrapper {
     const selector = selectors['dismiss-button'];
     const legacySelector = legacySelectors['dismiss-button'];
 
-    return this.findAny(`:is(.${legacySelector}, .${selector})`)!;
+    return this.find(`:is(.${legacySelector}, .${selector})`)!;
   }
 }


ernst-dev:
Successful dry-run `7539627246`

src/__tests__/snapshot-tests/__snapshots__/test-utils-selectors.test.tsx.snap
```diff
@@ -652,10 +652,12 @@ exports[`test-utils selectors 1`] = `
...
     "awsui_root_4yi2u",
   ],
+  "token": [
+    "awsui_dismiss-button_1i2wg",
+    "awsui_root_1i2wg",
+  ],
   "token-group": [
-    "awsui_dismiss-button_dm8gx",
     "awsui_root_dm8gx",
-    "awsui_token_dm8gx",
```
avinashbot, line 658: Concerned about backwards compatibility here; you may have to manually "anchor" the old test-util classes to make sure there's no disruption. I'm especially concerned because these are dependended upon by the multiselect test utils, which are much more commonly used.

src/internal/components/option/index.tsx
```diff
@@ -68,18 +73,65 @@ const Option = ({
     />
   );
 
+  // If labelContent is defined it is expected to be JSX, this is not safe to nest in a span so this condition nests it in a div instead.
```
avinashbot: Is this just duplicated only so that you can use `div` instead of `span`? I think you can change it to always be `div` (assuming the classnames always set the `display` to be consistent).

src/internal/components/option/option-parts.tsx
```diff
@@ -11,19 +11,42 @@ import analyticsSelectors from './analytics-metadata/styles.css.js';
...
+  labelContainerRef,
+  labelRef,
+  labelId,
+  label,
+  prefix,
+  highlightText,
+  triggerVariant,
+}: LabelProps) => {
+  // If label is not JSX, use the normal label structure with Highlight matching, else render the JSX
+  if (!React.isValidElement(label)) {
```
avinashbot: What is and isn't JSX is _very_ open-ended (`0` and `true` are valid, `undefined` is only valid from 18+). `isValidElement` is also a deprecated API and we shouldn't be using it in new code.

I assume we can only do highlight matching for strings, so maybe you can check if label is a string or undefined to decide whether or not to render the normal structure? Or maybe some kind of internal flag that our dropdowns could use?

src/popover/styles.scss
```diff
@@ -31,6 +32,19 @@ $trigger-underline-offset: 0.25em;
   max-inline-size: 100%;
   color: inherit;
   text-align: inherit;
+
+  // stylelint-disable-next-line selector-combinator-disallowed-list
```
avinashbot: There's a reason we disallow these; we're trying to avoid these CSS-based inter-component dependencies as much as humanly possible (we used to have them in previous versions). I think we need a better solution here; I assumed we were going for an explicit variant?

Since you mentioned they aren't required for this, it would be nice not to merge these types of declarations for now.

src/__tests__/snapshot-tests/__snapshots__/test-utils-selectors.test.tsx.snap
avinashbot: Reference: https://github.com/cloudscape-design/components/pull/3868

src/test-utils/dom/token/index.ts
```diff
@@ -0,0 +1,50 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import { ComponentWrapper, ElementWrapper } from '@cloudscape-design/test-utils-core/dom';
+
+import OptionWrapper from '../internal/option';
+
+import selectors from '../../../token/styles.selectors.js';
```
avinashbot: Could we make a separate stable style file for test utils? Similar to [test-classes/styles.scss](https://github.com/cloudscape-design/components/blob/main/src/button/test-classes/styles.scss) used in a lot of components?

src/token/dismiss-button.tsx
```diff
@@ -32,19 +35,19 @@ function DismissButton(
...
         }
 
-        onDismiss();
+        fireNonCancelableEvent(onDismiss);
       }}
       aria-label={dismissLabel}
       {...(disabled || readOnly ? {} : getAnalyticsMetadataAttribute(analyticsMetadata))}
     >
-      <InternalIcon name="close" />
+      <InternalIcon className={clsx(inline && styles['dismiss-icon-inline'])} name="close" />
```
avinashbot: Don't like that we're doing CSS overrides inside our own components. Could we do this through a wrapper around the icon or by styling the parent button?

src/token/internal.tsx
```diff
@@ -0,0 +1,184 @@
...
+  });
+
+  const buildOptionDefinition = () => {
+    const isLabelAnElement = React.isValidElement(label);
+    const labelObject = !isLabelAnElement ? { label: String(label) } : { labelContent: label };
+
+    if (isInline) {
+      if (isLabelAnElement) {
+        throw new Error(
+          'Invariant violation: only plain text (strings or numbers) are supported when variant="inline".'
```
avinashbot: If it's based on the input values, maybe a dev warning (`warnOnce`) would be better than crashing the app? I remember we had a bit of an incident with runtime throws only about a month ago.

src/token/styles.scss
```diff
@@ -0,0 +1,128 @@
...
+    padding-inline: 0;
+    display: flex;
+    align-items: center;
+    align-self: center;
+  }
+}
+
+.dismiss-icon-inline {
+  // !important is essential here to remove padding from the icon
+  padding-block: 0 !important;
```
avinashbot: As much as I hate it, I would rather address this through negative margins than with an `!important`. We want to avoid `!important` as much as humanly possible.

src/token/styles.scss
```diff
@@ -0,0 +1,128 @@
...
+  max-inline-size: 100%;
+
+  @include focus-visible.when-visible {
+    @include styles.focus-highlight(0px);
+  }
+}
+
+.token-option-inline {
+  // !important is essential here to center nested content inside the option
+  align-items: center !important;
```
avinashbot: Same comment about `!important` here, though maybe in this case, since it's an internal component, maybe we handle it with a property.

src/token/internal.tsx
```diff
@@ -0,0 +1,184 @@
...
+        /* istanbul ignore next: Tested with integration tests */
+        setShowTooltip(true);
+      }}
+      onMouseLeave={() => {
+        /* istanbul ignore next: Tested with integration tests */
+        setShowTooltip(false);
+      }}
+      tabIndex={!!tooltipContent && isInline && isEllipsisActive ? 0 : undefined}
+      // The below data attribute is to tell a potentially nested Popover to have less spacing between the text and the underline
+      data-token-inline={isInline || undefined}
```
avinashbot: Commented in the popover styles, but I strongly feel it's a bad idea creating this sort of silent leaky contract between components like this.

src/token/internal.tsx
```diff
@@ -0,0 +1,184 @@
...
+        styles.root,
+        !isInline ? styles['token-normal'] : styles['token-inline'],
+        analyticsSelectors.token,
+        baseProps.className
+      )}
+      aria-label={ariaLabel}
+      aria-labelledby={!ariaLabel ? ariaLabelledbyId : undefined}
+      aria-disabled={!!disabled}
+      role={role ?? 'group'}
+      onFocus={() => {
```
avinashbot, line 114: Is there a reason these couldn't be tested with unit tests? focus/blur/mouseenter/mouseleave should work fine here (though only through synthetic events, not keyboard tabbing behavior). Regarding our recent discussion about integ testing and flakiness, perhaps it's better to get unit test coverage (or both unit and integration test coverage) over just integration test coverage.

src/token/internal.tsx
```diff
@@ -0,0 +1,184 @@
...
+        <Tooltip
+          trackRef={labelContainerRef}
+          value={
+            <LiveRegion>
+              <span data-testid="tooltip-live-region-content">{tooltipContent}</span>
+            </LiveRegion>
+          }
+          size="medium"
+          // This is a non-existant class for testing purposes
+          className="token-tooltip"
```
avinashbot: Looks like you forgot to delete this?

src/token/styles.scss
```diff
@@ -0,0 +1,128 @@
...
+    @include styles.focus-highlight(0px);
+  }
+
+  &:focus {
+    outline: none;
+    text-decoration: none;
+  }
+
+  &:hover {
+    cursor: pointer;
```
avinashbot: Minor: this line doesn't need to be hover-only.

avinashbot (changes requested):

src/popover/styles.scss
ernst-dev: Removing these for this PR, this was actually a suggestion from Andrei to avoid adding a new variant

src/internal/components/option/index.tsx
```diff
@@ -30,13 +31,18 @@ const Option = ({
...
+  labelRef,
+  labelId,
   ...restProps
 }: OptionProps) => {
   if (!option) {
     return null;
   }
   const { disabled } = option;
   const baseProps = getBaseProps(restProps);
+  const SpanOrDivTag = option.labelContent ? 'div' : 'span';
```
ernst-dev, line 45: Keeping this as Select uses Option inside a button, when a div is used an A11Y validation fails because you can't nest a div inside a button

src/token/interfaces.ts
```diff
@@ -0,0 +1,76 @@
...
+   * Make sure that you add a listener to this event to update your application state.
+   */
+  onDismiss?: NonCancelableEventHandler;
+
+  /**
+   * Content to display in the tooltip when `variant="inline"`. The tooltip appears when the token label is truncated due to insufficient space.
+   *
+   * Only applies to plain text labels.
+   */
+  tooltipContent?: string;
```
YueyingLu, line 71: In the akrbA0gbwxcl doc it was not mentioned, just want to confirm it was approved by P&C and a11y?

src/token/internal.tsx
```diff
@@ -0,0 +1,188 @@
...
+  };
+
+  useResizeObserver(labelContainerRef, () => {
+    if (isInline) {
+      setIsEllipsisActive(isLabelOverflowing() ?? false);
+    }
+  });
+
+  const buildOptionDefinition = () => {
+    const isLabelAnElement = isReactElement(label);
```
YueyingLu: Why not check if the label is string or number? It is more straightforward and comply with "Only plain text (strings or numbers) are supported when variant="inline"."

src/__tests__/snapshot-tests/__snapshots__/test-utils-selectors.test.tsx.snap
YueyingLu: Agree on the concern, the above reference is for when root selector is updated: use `legacyRootSelector`.
For non-root selector, like ones used in `findDismiss()` and others, you need to do something like this  https://github.com/cloudscape-design/components/commit/8fa9df2b9fdbc19b20dd8afc46e5eba3035d51fe.

src/token/interfaces.ts
ernst-dev: This was a suggestion from Timo and was approved in the Release SO

src/test-utils/dom/token-group/token.ts
```diff
@@ -17,6 +17,9 @@ export default class TokenWrapper extends ComponentWrapper {
   }
 
   findDismiss(): ElementWrapper {
-    return this.findByClassName(selectors['dismiss-button'])!;
+    const selector = selectors['dismiss-button'];
+    const legacySelector = legacySelectors['dismiss-button'];
+
+    return this.findAny(`:is(.${legacySelector}, .${selector})`)!;
```
YueyingLu: If you use findAny already, just need to pass the array of selector https://github.com/cloudscape-design/test-utils/blob/c7200f200913812a551c8286bb4d1dccd5087565/src/core/selectors.ts#L39

ernst-dev: Changed it to a `.find` instead, `.findAny` with standard usage does not return a dismiss button

PR #3806, opened 2025-08-27
feat: TreeView keyboard navigation

cansuaa:
### Description
This PR adds keyboard navigation to tree-view.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


pages/tree-view/items/permutations-items.tsx
```diff
@@ -21,27 +21,57 @@ export const textItems: Item[] = [
...
-    children: [{ id: '1.1', content: 'Item 1.1', secondaryContent: 'Description 1.1', actions: 'Actions slot' }],
+    children: [
+      {
+        id: '1.1',
+        content: 'Item 1.1',
+        secondaryContent: 'Description 1.1',
+        actions: 'Actions slot',
+        children: [
+          {
+            id: '1.2',
```
pan-kot: nit: Having item 1.2 nested under 1.1 is confusing. I it better to name it 1.1.1 instead.

pages/tree-view/keyboard-navigation.page.tsx
```diff
@@ -0,0 +1,104 @@
...
+            }}
+            connectorLines="vertical"
+          />
+        </Container>
+      </Grid>
+
+      <Button>Focus here</Button>
+
+      <Grid gridDefinition={[{ colspan: { m: 7, xs: 12 } }]}>
+        <Container header={<h2>Focusable non-CDS elements inside tree-item</h2>}>
```
pan-kot: nit: I would rather say "registered" vs "non-registered" elements. There are cloudscape components like inputs or segmented controls that are not registered to single tab stop nav

pages/tree-view/permutations.page.tsx
```diff
@@ -59,14 +60,34 @@ export default function TreeViewPermuations() {
         Expand all
       </Checkbox>
 
+      <Checkbox
+        checked={urlParams.showConnectorLines ?? false}
+        onChange={event => {
+          setUrlParams({ showConnectorLines: event.detail.checked });
+          window.location.reload();
```
pan-kot: why do we need a reload here?

src/test-utils/dom/tree-view/index.ts
```diff
@@ -39,7 +39,7 @@ class TreeViewItemWrapper extends ComponentWrapper {
    * Finds the expand toggle of the tree view item.
    */
   findItemToggle(): ElementWrapper | null {
-    return this.findByClassName(expandToggleStyles['expand-toggle']);
+    return this.find(`.${expandToggleStyles['expand-toggle']}:not(.${expandToggleStyles.invisible})`);
```
pan-kot: why do we use expand toggle styles and not test utils styles here?

src/internal/components/expand-toggle-button/index.tsx
```diff
@@ -16,24 +16,45 @@ export function ExpandToggleButton({
...
   const buttonRef = useRef<HTMLButtonElement>(null);
   const { tabIndex } = useSingleTabStopNavigation(buttonRef);
+
+  if (invisible) {
+    return (
+      <button
+        type="button"
+        ref={buttonRef}
+        tabIndex={tabIndex}
+        aria-label={isExpanded ? collapseButtonLabel : expandButtonLabel}
```
pan-kot: I believe this should be aria-hidden instead, as the sole purpose of this button is to capture focus for keyboard nav.

src/internal/components/expand-toggle-button/styles.scss
```diff
@@ -40,8 +40,16 @@
...
-    @include styles.focus-highlight(awsui.$space-button-focus-outline-gutter);
+  // focus ring for tree-view is different and handled in tree-item/styles
+  &:not([data-awsui-tree-view-toggle-button]) {
+    @include focus-visible.when-visible {
+      @include styles.focus-highlight(awsui.$space-button-focus-outline-gutter);
+    }
+  }
+
+  &.invisible {
+    color: transparent;
```
pan-kot: Why do we set `color: transparent`? This should only affect text content of the invisible buttons, but there is no content.

src/internal/components/structured-item/index.tsx
```diff
@@ -19,7 +19,10 @@ export default function InternalStructuredItem({
   wrapActions = true,
 }: StructuredItemProps) {
   return (
-    <div className={clsx(styles.root, testClasses.root, disablePaddings && styles['disable-paddings'])}>
+    <div
+      data-awsui-structured-item={true}
```
pan-kot: Can we use `styles.root` for the element lookup instead?

```
return element.closest(`.${styles.root}`)
```

src/tree-view/keyboard-navigation/index.tsx
```diff
@@ -0,0 +1,302 @@
...
+  getClosestTreeItem,
+  getClosestTreeItemContent,
+  getToggleButtonOfTreeItem,
+  isElementDisabled,
+} from './utils';
+
+export function KeyboardNavigationProvider({
+  getTreeView,
+  children,
+  pageUpDownSize = 10,
```
pan-kot: We used the same approach in the table so that page size can be parametrised. However, there does not seem to be the need. I recommend removing the argument in favour of a constant defined in this file.

src/tree-view/keyboard-navigation/index.tsx
```diff
@@ -0,0 +1,302 @@
...
+    // so that when tree-view is focused again, the last focused tree-item can be re-focused
+    if (target === null) {
+      this.setNavigationActive(true);
+      return;
+    }
+
+    // Disable navigation if the new focus target is inside the tree-item content
+    // so that the focusables inside can be navigated with tab
+    const closestTreeItemContent = getClosestTreeItemContent(target);
+    if (closestTreeItemContent) {
```
pan-kot: nit: this can be simplified to `this.setNavigationActive(!closestTreeItemContent)`

src/tree-view/keyboard-navigation/utils.ts
```diff
@@ -0,0 +1,69 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+export function focusElement(element: null | HTMLElement) {
+  if (element) {
+    element.focus();
```
pan-kot: nit: the util can be replaced with `element?.focus()` 1-liner

src/tree-view/internal.tsx
```diff
@@ -37,6 +40,9 @@ const InternalTreeView = <T,>({
     controlledProp: 'expandedItems',
     changeHandler: 'onItemToggle',
   });
+  const treeViewRefObject = useRef(null);
+
+  const allVisibleItemsIndeces = getAllVisibleItemsIndeces({ items, expandedItems, getItemId, getItemChildren });
```
pan-kot: typo: Indeces -> Indices

src/tree-view/interfaces.ts
```diff
@@ -90,6 +95,8 @@ export namespace TreeViewProps {
     expanded: boolean;
   }
 
+  export type ConnectorLinesVariant = 'vertical';
```
pan-kot: Should we use 'vertical' or `'vertical' | 'none'`? If the latter - we should also assign the default in index.tsx. Also, let's add this prop to the metrics.

src/internal/components/expand-toggle-button/index.tsx
cansuaa: Focusable elements shouldn't have aria-hidden

> aria-hidden="true" should not be used on elements that can receive focus

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-hidden

src/test-utils/dom/tree-view/index.ts
cansuaa: I'm a little hesitant because the last time I moved the toggle file, the hash used in the test util of table was changed and caused issues on the customer end so I'm not sure if this would cause a similar problem.

src/internal/components/structured-item/index.tsx
cansuaa: What's wrong with using a data attribute?

pages/tree-view/removable-elements.page.tsx
```diff
@@ -0,0 +1,183 @@
...
+import TreeView from '~components/tree-view';
+
+import { Item, stsnRegisteredItems } from './items/keyboard-navigation-items';
+
+interface FlattenedItem extends Item {
+  parentId: string;
+  noActionButton?: boolean;
+}
+
+const flattenItems = (items: Item[]) => {
```
pan-kot, line 19: nit: this function is defined twice in the tree-view pages, I would move it to some common utils instead

src/internal/components/expand-toggle-button/index.tsx
pan-kot: Ok, so apparently no matter what we do the focused element is going to be announced. Even replacing a button with a div does not help. There is, however, a problem: the element is announced e.g. as "Item 2, button" - which is not really telling. It is not clear that the button is fake and has no associated interaction. I think we should add a somewhat special aria label or aria role-description, or maybe it can be enough to use aria-disabled=true.

pan-kot:
```suggestion
        aria-disabled={true}
```

pan-kot: The invisible toggle represents non-expandable items, therefore the `isExpanded` check is never needed. Besides, using the `collapseButtonLabel` is semantically incorrect either. It works because of the code in the parent component:

```
collapseButtonLabel={
  isExpandable
    ? joinStrings(
      i18n('i18nStrings.collapseButtonLabel', i18nStrings?.collapseButtonLabel?.(item)),
      itemLabelToAnnounce
    )
    : itemLabelToAnnounce
}
``` 

I recommend to make it better clear by declaring 3 labels instead: expandButtonLabel, collapseButtonLabel, and invisibleButtonLabel (can work on naming here)

pan-kot: Actually, instead of using aria-disabled I think it worths considering using a non-button element. E.g. a div w/o a role is announced as "group" - so it is clear that the element is not interactive.

src/internal/components/expand-toggle-button/styles.scss
```diff
@@ -40,8 +40,15 @@
...
-    @include styles.focus-highlight(awsui.$space-button-focus-outline-gutter);
+  // focus ring for tree-view is different and handled in tree-item/styles
+  &:not([data-awsui-tree-view-toggle-button]) {
+    @include focus-visible.when-visible {
+      @include styles.focus-highlight(awsui.$space-button-focus-outline-gutter);
+    }
+  }
+
+  &.invisible {
+    pointer-events: none;
```
pan-kot: why is this actually needed?

src/test-utils/dom/tree-view/index.ts
pan-kot: As a rule we use test classes for any new components and features. We can evaluate whether it is safe to replace the existing tree view selectors, but let's at least use the test classes for the newly added invisible style.

src/tree-view/__integ__/keyboard-navigation.test.ts
```diff
@@ -0,0 +1,75 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import { BasePageObject } from '@cloudscape-design/browser-test-tools/page-objects';
+import useBrowser from '@cloudscape-design/browser-test-tools/use-browser';
+import { createWrapper } from '@cloudscape-design/test-utils-core/selectors';
+
+import '../../../lib/components/test-utils/selectors';
```
pan-kot: why do we import it as such instead of `import createWrapper from '../../../lib/components/test-utils/selectors'`?

src/tree-view/__tests__/items.tsx
```diff
@@ -0,0 +1,175 @@
...
+export const defaultProps: TreeViewProps<Item> = {
+  items: items,
+  getItemId: item => item.id,
+  getItemChildren: item => item.items,
+  renderItem: item => ({
+    content: item.title,
+  }),
+};
+
+export interface KeyboardNavigationItem extends TreeViewProps.TreeItem {
```
pan-kot: nit: every tree view is keyboard-navigable as the feature is not conditional. I would rename the var to something like itemsWithActions or similar so that the name describes the items shape and not the test case it is designed for. Otherwise, you can remove these items from the common file and define them in the keyboard nav tests instead.

src/tree-view/tree-item/index.tsx
```diff
@@ -55,55 +59,67 @@ const InternalTreeItem = <T,>({
       ? (content as string)
       : '';
 
-  //  Role `treeitem` isn't used in the initial release per discussion with A11Y team. It requires focus management to be implemented so they will be added as a follow up together.
   return (
     <li
+      role="treeitem"
```
pan-kot, line 65: I have a question to the next line. Why do we assign the list item ID as id attribute? We already have the prefixes data-testid we can use for selectors.

src/tree-view/__tests__/keyboard-navigation.test.tsx
```diff
@@ -0,0 +1,383 @@
...
+      announcementLabel: item.announcementLabel,
+    })}
+    {...props}
+  />
+);
+
+function renderTreeView(props: Partial<TreeViewProps<Item>> = {}) {
+  const { container, rerender } = render(<TestTreeView {...props} />);
+  const wrapper = createWrapper(container).findTreeView()!;
+  return { container, treeView: container.querySelector('[role="tree"]')!, wrapper, rerender };
```
pan-kot: Why do we query the role="tree" element separately? Unless we expect customers do the same we should expose a test utils method for that.

src/tree-view/__tests__/keyboard-navigation.test.tsx
```diff
@@ -0,0 +1,383 @@
...
+  const children = firstRootLevelItem.findChildItems()!;
+  expect(children[0].getElement().getAttribute(treeItemIndexAttribute)).toBe('1');
+  expect(children[1].getElement().getAttribute(treeItemIndexAttribute)).toBe('2');
+
+  // second root level item's index should be updated
+  expect(secondRootLevelItem.getElement().getAttribute(treeItemIndexAttribute)).toBe('3');
+
+  const allItems = wrapper.findItems();
+  const lastRootLevelItem = allItems[allItems.length - 1];
+  expect(lastRootLevelItem.getElement().getAttribute(treeItemIndexAttribute)).toBe(`${allItems.length - 1}`);
```
pan-kot: nit: for this test, wouldn't it be easier and better explicit to query all <li> elements and extract their indices? You can assert that the id's are e.g. "0, 1, 2" before any item is expanded, and then "0, 1, 2, 3, 4" - after items are expanded. For the test to be better precise you can extract not only indices but also text content, e.g. to have lists like ["Item 1 - 0", "Item 2 - 1", "Item 3 - 2"] and ["Item 1 - 0", "Item 1.1 - 1", ...].

src/tree-view/__tests__/keyboard-navigation.test.tsx
```diff
@@ -0,0 +1,383 @@
...
+
+    wrapper.findItemById('2-button-group')!.findItemToggle()!.getElement().focus();
+    fireEvent.keyDown(treeView, { keyCode: KeyCode.pageDown });
+    expect(readActiveElement()).toBe('BUTTON[Item 5.3]');
+
+    fireEvent.keyDown(treeView, { keyCode: KeyCode.pageUp });
+    expect(readActiveElement()).toBe('BUTTON[Item 2]');
+  });
+
+  test('page up and down keys - focuses on the top or bottom tree-items if there are less than 10 items', () => {
```
pan-kot: nit: you can also modify the above test to assert the same, but when there is actually more than 10 items: pressing pageDown twice on a list of 15 must move focus to the last item. Pressing it the 3rd time should have no effect

src/tree-view/__tests__/keyboard-navigation.test.tsx
```diff
@@ -0,0 +1,383 @@
...
+  });
+
+  test('ignores invalid key', () => {
+    const { treeView, wrapper } = renderTreeView();
+
+    const firstToggle = wrapper.findItemById('1-button-actions')!.findItemToggle()!.getElement();
+    firstToggle.focus();
+    expect(firstToggle).toHaveFocus();
+
+    fireEvent.keyDown(treeView, { keyCode: KeyCode.backspace });
```
pan-kot, line 285: nit: we can also check that the default is not prevented in this case

src/tree-view/keyboard-navigation/utils.ts
```diff
@@ -0,0 +1,66 @@
...
+export function focusElement(element: null | HTMLElement) {
+  element?.focus();
+}
+
+export function getClosestTreeItem(element: Element) {
+  return element.closest('li[data-awsui-tree-item-index]') as null | HTMLLIElement;
+}
+
+export function getToggleButtonOfTreeItem(treeItem: HTMLLIElement) {
+  const toggleElement = treeItem.querySelector('[data-awsui-tree-view-toggle-button=true]');
```
pan-kot: can we use a class name selector here instead?

src/tree-view/keyboard-navigation/utils.ts
```diff
@@ -0,0 +1,66 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+export function focusElement(element: null | HTMLElement) {
```
pan-kot: is this function really needed?

src/tree-view/tree-item/index.tsx
```diff
@@ -55,55 +59,67 @@ const InternalTreeItem = <T,>({
...
+              collapseButtonLabel={
+                isExpandable
+                  ? joinStrings(
+                      i18n('i18nStrings.collapseButtonLabel', i18nStrings?.collapseButtonLabel?.(item)),
+                      itemLabelToAnnounce
+                    )
+                  : itemLabelToAnnounce
+              }
               onExpandableItemToggle={() => onItemToggle({ id, item, expanded: !isExpanded })}
+              invisible={!isExpandable}
```
pan-kot: nit: should we pass isExpandable instead of introducing the prop `invisible`?

src/tree-view/tree-item/index.tsx
```diff
@@ -55,55 +59,67 @@ const InternalTreeItem = <T,>({
...
+            secondaryContent={secondaryContent}
+            actions={actions}
+            wrapActions={false}
+          />
+        </div>
       </div>
 
       {isExpanded && children.length && (
-        <ul className={styles['treeitem-group']}>
+        <ul role="group" className={styles['treeitem-group']}>
```
pan-kot, line 117: why do we use role="group" here?

src/tree-view/__tests__/keyboard-navigation.test.tsx
cansuaa: The root test util returns the wrapping div but the keyboard actions need to happen on the ul element. Would customers need to test the keyboard navigation for some reason?

src/tree-view/__tests__/keyboard-navigation.test.tsx
cansuaa: You mean having them in a loop and assert each item's index?

pages/tree-view/keyboard-navigation.page.tsx
```diff
@@ -0,0 +1,105 @@
...
+export default function TreeViewKeyboardNavigation() {
+  const [pressedItems, setPressedItems] = useState<ReadonlyArray<string>>([]);
+
+  return (
+    <SpaceBetween size="m">
+      <h1>TreeView keyboard navigation</h1>
+
+      <Button>Focus here</Button>
+
+      <Grid gridDefinition={[{ colspan: { m: 7, xs: 12 } }]}>
```
pan-kot: nit: these sections use the same grid definition config, is there an opportunity to use a single grid for all three examples?

src/tree-view/__tests__/keyboard-navigation-provider.test.tsx
```diff
@@ -0,0 +1,159 @@
...
+      current: {
+        updateFocusTarget: () => {},
+        getFocusTarget: () => null,
+        isRegistered: () => false,
+      },
+    });
+    expect(() => navigation.getNextFocusTarget()).not.toThrow();
+    expect(() => navigation.onUnregisterActive()).not.toThrow();
+    expect(() => navigation.cleanup()).not.toThrow();
+    await waitFor(() => {
```
pan-kot: Why is this waitFor needed? If there is an error in the timeout, it won't be captured by this code.

src/tree-view/__tests__/keyboard-navigation-provider.test.tsx
```diff
@@ -0,0 +1,159 @@
...
+    }
+    expect(() => render(<TestTreeView />)).not.toThrow();
+  });
+
+  test('does not throw when key is pressed while there is no focused tree-item', async () => {
+    const { container } = render(<TestTreeView />);
+    const treeView = container.querySelector('ul')!;
+
+    await new Promise(resolve => setTimeout(resolve, 0)); // wait for refresh to run
+    expect(() => fireEvent.keyDown(treeView, { keyCode: KeyCode.down })).not.toThrow();
```
pan-kot: The fireEvent will now throw an error, so this assertion makes no difference. You can imitate the error in the keydown handler and see what options can be used to capture it. Maybe we can check that the component is still rendered.

src/tree-view/__tests__/keyboard-navigation.test.tsx
pan-kot: Not necessarily, you can extract indices like `wrapper.findItems().map(i => i.getElement().getAttribute(treeItemIndexAttribute))`, and then assets against your expected set of indices.

src/tree-view/keyboard-navigation/utils.ts
```diff
@@ -0,0 +1,52 @@
...
+export function findTreeItemContentById(treeView: HTMLUListElement, treeItemId: string) {
+  const treeItemContent = treeView.querySelector(
+    `li[data-awsui-tree-item-index][id="${treeItemId}"] .${treeItemStyles['tree-item-structured-item']}`
+  );
+  return treeItemContent as null | HTMLLIElement;
+}
+
+export function isTreeItemToggle(element: Element) {
+  return (
+    (element.tagName === 'BUTTON' || element.tagName === 'DIV') &&
```
pan-kot: nit: it is really necessary to validate the tag name here?

src/tree-view/__tests__/keyboard-navigation-provider.test.tsx
cansuaa: Oh I don't need it anymore, I was just trying out stuff with codecov

src/tree-view/__tests__/keyboard-navigation-provider.test.tsx
cansuaa: Yeah the test moved away from its original purpose, I'll rename it and assert treeView is in the document.

pan-kot (approved):

PR #3814, opened 2025-08-29
feat: add tag and filtering tag to property filter

BaldMonster33:
### Description

This is to include tags and filteringTags to be supported in property filter.

Will use internally review for this

### How has this been tested?

Unit tests added.

Demo tag page added

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/property-filter/tags-demo.page.tsx
+++ b/pages/property-filter/tags-demo.page.tsx
@@ -1,0 +1,258 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import React, { useState } from 'react';
+
+import { Box, Container, Header, PropertyFilter, SpaceBetween } from '~components';
+import { PropertyFilterProps } from '~components/property-filter';
+
+import { i18nStrings, labels } from './common-props';
+
+// Sample filtering properties for the demo
+const filteringProperties: PropertyFilterProps.FilteringProperty[] = [
+  {
+    key: 'service',
+    propertyLabel: 'Service',
+    groupValuesLabel: 'Service values',
+    operators: ['=', '!=', ':', '!:'],
+  },
+  {
+    key: 'environment',
+    propertyLabel: 'Environment',
+    groupValuesLabel: 'Environment values',
+    operators: ['=', '!=', ':', '!:'],
+  },
+  {
+    key: 'logGroup',
+    propertyLabel: 'Log group',
+    groupValuesLabel: 'All log groups', // Use "All log groups" as the group header
+    operators: [
+      // Set token type for equals and not equals operators as enum so that
+      // the default multi-choice form and formatter are used.
+      { operator: '=', tokenType: 'enum' },
+      { operator: '!=', tokenType: 'enum' },
+      // Keep token type for contains and not contains operators as is.
+      ':',
+      '!:',
+    ],
+    defaultOperator: '=',
+  },
+];
+
+// Sample filtering options with various tag configurations
+const filteringOptions: PropertyFilterProps.FilteringOption[] = [
+  // Simple test options with tags
+  {
+    propertyKey: 'service',
+    value: 'ec2',
+    label: 'Amazon EC2',
+    tags: ['compute', 'virtual-machines'],
+  },
+  {
+    propertyKey: 'service',
+    value: 's3',
+    label: 'Amazon S3',
+    tags: ['storage', 'object-storage'],
+    filteringTags: ['bucket', 'simple-storage'],
+  },
+  {
+    propertyKey: 'service',
+    value: 'lambda',
+    label: 'AWS Lambda',
+    tags: ['serverless', 'compute', 'functions'],
+    filteringTags: ['faas', 'function-as-a-service'],
+  },
+  {
+    propertyKey: 'service',
+    value: 'rds',
+    label: 'Amazon RDS',
+    tags: ['database', 'relational', 'managed'],
+    filteringTags: ['mysql', 'postgresql', 'oracle', 'sql-server'],
+  },
+  {
+    propertyKey: 'environment',
+    value: 'production',
+    label: 'Production',
+    tags: ['live', 'critical'],
+  },
+  {
+    propertyKey: 'environment',
+    value: 'staging',
+    label: 'Staging',
+    tags: ['pre-production', 'testing', 'qa'],
+    filteringTags: ['stage', 'pre-prod'],
+  },
+  {
+    propertyKey: 'environment',
+    value: 'development',
+    label: 'Development',
+    tags: ['dev', 'testing', 'experimental'],
+    filteringTags: ['sandbox', 'playground'],
+  },
+  // Option without tags for comparison
+  {
+    propertyKey: 'service',
+    value: 'cloudfront',
+    label: 'Amazon CloudFront',
+  },
+
+  // Log group examples - individual options only
+  // Group header "All log groups" serves as the bulk selection mechanism
+  {
+    propertyKey: 'logGroup',
+    value: 'Log-group-01',
+    label: 'Log-group-01',
+    tags: ['Standard', 'AccountID', 'Tag'],
+  },
+  {
+    propertyKey: 'logGroup',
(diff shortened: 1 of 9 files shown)


src/property-filter/__tests__/property-filter-tags.test.tsx
```diff
@@ -0,0 +1,256 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { OptionDefinition } from '../../internal/components/option/interfaces';
+import { filterOptions } from '../filter-options';
+
+describe('Property Filter filterOptions with tags', () => {
```
pan-kot: should we have a test suite that secures filtering on all props, not just the newly introduced ones?

src/property-filter/__tests__/property-filter-tags.test.tsx
```diff
@@ -0,0 +1,256 @@
...
+  const createOption = (overrides: Partial<OptionDefinition> = {}): OptionDefinition => ({
+    value: 'test-value',
+    label: 'Test Label',
+    ...overrides,
+  });
+
+  describe('matchSingleOption tag functionality', () => {
```
pan-kot: nit: I would make the tests structure more flat, I think a single "describe" should be enough

pages/property-filter/tags-demo.page.tsx
```diff
@@ -0,0 +1,258 @@
...
+  {
+    key: 'environment',
+    propertyLabel: 'Environment',
+    groupValuesLabel: 'Environment values',
+    operators: ['=', '!=', ':', '!:'],
+  },
+  {
+    key: 'logGroup',
+    propertyLabel: 'Log group',
+    groupValuesLabel: 'All log groups', // Use "All log groups" as the group header
```
pan-kot: nit: I think the code comment here is unnecessary

pages/property-filter/tags-demo.page.tsx
```diff
@@ -0,0 +1,258 @@
...
+      { operator: '!=', tokenType: 'enum' },
+      // Keep token type for contains and not contains operators as is.
+      ':',
+      '!:',
+    ],
+    defaultOperator: '=',
+  },
+];
+
+// Sample filtering options with various tag configurations
```
pan-kot: nit: the comments annotating demo vars also seem unnecessary

pages/property-filter/tags-demo.page.tsx
```diff
@@ -0,0 +1,258 @@
...
+    label: 'Log-group-10',
+    tags: ['Standard', 'AccountID', 'Tag'],
+  },
+  // Simulate pagination boundary
+  {
+    propertyKey: 'logGroup',
+    value: 'LOAD_MORE',
+    label: '... Load more (showing 10 of 10,000+)',
+    tags: ['Pagination', 'API-Call-Required'],
+    // description: 'Load additional log groups from API',
```
pan-kot: why are these lines commented out?

src/property-filter/interfaces.ts
```diff
@@ -386,6 +386,8 @@ export interface InternalFilteringOption {
   property: null | InternalFilteringProperty;
   value: string;
   label: string;
+  tags?: ReadonlyArray<string>;
```
pan-kot, line 389: There are no unit tests that actually verify that the tags and filtering tags are added to the options. I believe we also need new test utils for these.

pages/property-filter/tags-demo.page.tsx
```diff
@@ -0,0 +1,258 @@
...
+      header={
+        <Header variant="h1" description="Demo page showcasing Property Filter with tags support">
+          Property Filter Tags Demo
+        </Header>
+      }
+    >
+      <SpaceBetween size="l">
+        <Box>
+          <Header variant="h2">Interactive Demo</Header>
+          <p>This demo showcases the Property Filter component with tags support. Try searching for:</p>
```
pan-kot: nit: This description is not very necessary either: the test pages are used mainly for automated or manual testing and prototyping, not as legitimate demos. The developers working with this page in the future will be aware of its purpose from the code.

pages/property-filter/tags-demo.page.tsx
```diff
@@ -0,0 +1,258 @@
...
+          filteringProperties={filteringProperties}
+          filteringOptions={filteringOptions}
+          query={query}
+          onChange={({ detail }) => setQuery(detail)}
+          countText={`${query.tokens.length} ${query.tokens.length === 1 ? 'filter' : 'filters'} applied`}
+        />
+
+        <Box>
+          <Header variant="h3">Current Query</Header>
+          <pre style={{ background: '#f5f5f5', padding: '10px', borderRadius: '4px' }}>
```
pan-kot: This will not work nicely with the dark mode, and can cause our automated a11y tests to fail. Instead, you can use `<Box variant="code">` or `<Box variant="pre">`.

pages/property-filter/tags-demo.page.tsx
BaldMonster33: Good point. Will get this removed.

src/property-filter/__tests__/property-filter-tags.test.tsx
BaldMonster33: Fair point. Looks like we are not actually testing other props in the tests. I can add it to the main property-filter.test.tsx file

BaldMonster33: Oh wait. Looks like we do have some in property-filter.filtering-input.test.tsx. I think I should add those there.

BaldMonster33: Okay added one that tests tag and filteringTag to the main test file. Not sure if we will still need the file that I created...

pages/property-filter/tags-demo.page.tsx
BaldMonster33: Removed.

pages/property-filter/tags-demo.page.tsx
BaldMonster33: Removed

pages/property-filter/tags-demo.page.tsx
BaldMonster33: I think I meant to delete that. Will remove it.

pages/property-filter/tags-demo.page.tsx
BaldMonster33: Updated

src/property-filter/interfaces.ts
BaldMonster33: Hmm couldn't find if we have existing tests for interface options like this. Feel like they are more implicitly tested by other functions. Would having propertyFilter render and provide those options in the filteringOptions enough for this one?

pages/property-filter/property-filter-tags.page.tsx
```diff
@@ -0,0 +1,198 @@
...
+  },
+  {
+    propertyKey: 'logGroup',
+    value: 'Log-group-10',
+    label: 'Log-group-10',
+    tags: ['Standard', 'AccountID', 'Tag'],
+  },
+  {
+    propertyKey: 'logGroup',
+    value: 'LOAD_MORE',
```
pan-kot: What is this option for? The property filter supports async loading with scroll, there is no need to create a fake option to trigger loading.

src/property-filter/__tests__/property-filter.filtering-input.test.tsx
```diff
@@ -201,6 +216,27 @@ describe('filtering input', () => {
       ).toEqual(['string : value1', 'string : value2']);
     });
 
+    test('filters suggestions by tags and filteringTags in dropdown', () => {
```
pan-kot, line 219: good 👍

pages/property-filter/property-filter-tags.page.tsx
BaldMonster33: Removed

pan-kot:
@BaldMonster33 the PR looks good now, but there is now a linting error that causes the build to fail. Please address

pan-kot (approved):

PR #3818, opened 2025-08-29
chore: Enable more SVGO optimizations

TrevorBurnham:
### Description

This PR upgrades `svgo` to the latest version (2.8.0 -> 4.0.0) and enables more optimizations. This brings down the total size of all icons from 35KB to 33KB, while preserving all accessibility attributes. 🚀 

### How has this been tested?

I've verified that this doesn't introduce any visual changes by doing a pixel diff on the "Icon variant normal" demo page:

Icon variant normal

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/build-tools/tasks/generate-icons.js
+++ b/build-tools/tasks/generate-icons.js
@@ -31,6 +31,21 @@ function getIcon(iconName, content) {
     plugins: [
       'preset-default',
       {
+        name: 'removeViewBox',
+        active: false,
+      },
+      {
+        name: 'convertPathData',
+        params: {
+          removeUseless: true,
+          straightCurves: true,
+          lineShorthands: true,
+          curveSmoothShorthands: true,
+        },
+      },
+      'removeUselessDefs',
+      'removeEmptyContainers',
+      {
         name: 'awsuiValidateAttributes',
         type: 'visitor',
         fn: () => ({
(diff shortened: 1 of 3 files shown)


georgylobko (commented):
In overall looks good to me. But there are many diffs in screenshot tests for `settings` icon. I would encourage you to investigate the reason behind that. Screenshot attached.

image

TrevorBurnham:
Ah, thanks for catching that! Which browser produced the screenshot with the diff? Are the changes perceptible? I don't see the screenshot test you're referencing; does it provide separate before and after images?

georgylobko:
> Ah, thanks for catching that! Which browser produced the screenshot with the diff? 

Chrome and Firefox, maybe other browsers as well

> Are the changes perceptible? 

From what I can see, no, they aren't. But if this PR is causing changes to the SVGs, it could potentially affect custom icons in ways that users might notice. I'd want to understand why this is happening before we
merge this PR

> I don't see the screenshot test you're referencing; does it provide separate before and after images?

I ran them in my dev pipeline and cannot share the link here.

TrevorBurnham:
Hmm… I took a closer look and found that the byte savings here are even smaller than I'd thought: 35069 bytes -> 34954 bytes. (The overly rosy 33KB claim in the PR description came from an earlier attempt that used precision `2`, which introduced visible discrepancies.) Since it's less than 100 bytes of savings and there are still failing visual regression tests, the juice isn't worth the squeeze.

PR #3824, opened 2025-09-02
feat: AI drawer header actions

georgylobko:
### Description

This PR adds a new API for declaring header actions in global panels registered through the plugin API. The goal is to make it easier to add actions to panel headers. Currently, teams have to create custom buttons
with absolute positioning, which causes accessibility issues. This PR fixes that by using the existing `ButtonGroup` component instead.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/app-layout/utils/external-global-left-panel-widget.tsx
+++ b/pages/app-layout/utils/external-global-left-panel-widget.tsx
@@ -3,7 +3,7 @@
 import React from 'react';
 import ReactDOM, { unmountComponentAtNode } from 'react-dom';
 
-import { Box, StatusIndicator } from '~components';
+import { Box } from '~components';
 import { registerLeftDrawer } from '~components/internal/plugins/widget';
 
 import styles from '../styles.scss';
@@ -94,7 +94,6 @@ registerLeftDrawer({
       id: 'add',
       iconName: 'add-plus',
       text: 'Add',
-      popoverFeedback: <StatusIndicator type="success">Message copied</StatusIndicator>,
     },
   ],
 
--- a/pages/app-layout/utils/external-widget.tsx
+++ b/pages/app-layout/utils/external-widget.tsx
@@ -3,7 +3,6 @@
 import React, { useEffect, useImperativeHandle, useRef, useState } from 'react';
 import ReactDOM, { unmountComponentAtNode } from 'react-dom';
 
-import { StatusIndicator } from '~components';
 import Box from '~components/box';
 import Drawer from '~components/drawer';
 import awsuiPlugins from '~components/internal/plugins';
@@ -73,7 +72,6 @@ awsuiPlugins.appLayout.registerDrawer({
       id: 'add',
       iconName: 'add-plus',
       text: 'Add',
-      popoverFeedback: <StatusIndicator type="success">Message copied</StatusIndicator>,
     },
   ],
   onHeaderActionClick: ({ detail }) => {
@@ -190,7 +188,6 @@ awsuiPlugins.appLayout.registerDrawer({
       id: 'add',
       iconName: 'add-plus',
       text: 'Add',
-      popoverFeedback: <StatusIndicator type="success">Message copied</StatusIndicator>,
     },
   ],
   onHeaderActionClick: ({ detail }) => {


src/app-layout/__tests__/runtime-drawers.test.tsx
```diff
@@ -1183,9 +1198,34 @@ describe('toolbar mode only features', () => {
 
         findDrawerTriggerById('global-drawer', renderProps)!.click();
         expect(onToggle).toHaveBeenCalledWith({ isOpen: true, initiatedByUserAction: true });
-        renderProps.globalDrawersWrapper.findCloseButtonByActiveDrawerId('global-drawer')!.click();
+        findCloseButtonByActiveDrawerId('global-drawer', renderProps)!.click();
         expect(onToggle).toHaveBeenCalledWith({ isOpen: false, initiatedByUserAction: true });
       });
+
+      (type === 'global-ai' ? test : test.skip)(
```
just-boris: I started a new file for this new API: https://github.com/cloudscape-design/components/blob/main/src/app-layout/__tests__/runtime-drawers-widgetized.test.tsx

It is better to add new exclusive features there

georgylobko: this test is no longer specific for the ai drawer - it now applies for to both global and global-ai

src/app-layout/visual-refresh-toolbar/drawer/global-drawer.tsx
```diff
@@ -146,34 +179,30 @@ function AppLayoutGlobalDrawerImplementation({
...
+                          activeGlobalDrawer?.onHeaderActionClick?.(event);
+                      }
+                    }}
+                    ariaLabel="Global panel actions"
+                    items={drawerActions}
+                    __internalRootRef={(root: HTMLElement) => {
+                      if (!root) {
+                        return;
+                      }
+                      refs.close = { current: root.querySelector('[data-itemid="close"]') as unknown as Focusable };
```
georgylobko, line 203: I'm open to any suggestion to make it better

pages/app-layout/utils/external-global-left-panel-widget.tsx
```diff
@@ -69,20 +67,38 @@ registerLeftDrawer({
...
+          text: 'Remove',
+        },
+      ],
+    },
+    {
+      type: 'icon-button',
+      id: 'add',
+      iconName: 'add-plus',
+      text: 'Add',
+      popoverFeedback: <StatusIndicator type="success">Message copied</StatusIndicator>,
```
just-boris: ⛳  This is not going to work in the real world when the React instance used in the widget and version used in app layout are different

Perhaps we should omit all slot properties from this API (or print a warning they are not supported in this mode)

src/app-layout/visual-refresh-toolbar/drawer/global-drawer.tsx
```diff
@@ -74,6 +76,37 @@ function AppLayoutGlobalDrawerImplementation({
   const animationDisabled =
     (activeGlobalDrawer?.defaultActive && !drawersOpenQueue.includes(activeGlobalDrawer.id)) ||
     (wasExpanded && !isExpanded);
+  let drawerActions: ReadonlyArray<ButtonGroupProps.InternalItemOrGroup> = [
```
just-boris: ⛳  Is it the same code as above? 

This needs to be deduplicated

src/app-layout/visual-refresh-toolbar/drawer/global-drawer.tsx
just-boris: Ok to merge as a temporary solution.

But let's ask in the P&C channel for help. Most likely we should expose a new package-private API to assign refs on button group items

src/button-group/interfaces.ts
```diff
@@ -132,6 +135,10 @@ export namespace ButtonGroupProps {
     popoverFeedback?: React.ReactNode;
   }
 
+  export interface InternalIconButton extends IconButton {
```
just-boris: ⛳ This lives inside `ButtonGroupProps` namespace which is public. 

You need to move all private additions outside this namespace to avoid exporting to public

georgylobko: addressed

pages/app-layout/utils/external-global-left-panel-widget.tsx
georgylobko: Printed a warning for unsupported props

src/app-layout/visual-refresh-toolbar/drawer/global-drawer.tsx
georgylobko: no, it might look similar, but it's not the same. global drawer and left global drawer have slightly different set of props

pages/app-layout/utils/external-global-left-panel-widget.tsx
just-boris: We need to remove these instances, because having them with a warning may confuse if this use-case is supported or not (it is not supported)

just-boris (approved):

PR #3833, opened 2025-09-03
fix: Improve accessibility for grouped options in dropdowns

avinashbot:
### Description

Previously, we kept a flat structure in our listboxes. Even though options can be visually nested, they're not programatically nested, though we included plenty of tricks to make it appear as if they are. This kept confusing accessibility vendors (and for good reason).

```html
<ul role="listbox">
  <li aria-hidden="true">Group heading 1</li>
  <li role="option">
    <span aria-hidden="true">Option 1</span>
    <span class="visually-hidden"><!-- "Group 1 Option 1" or "Option 1" depending on entry direction --></span>
  </li>
  <li role="option">
    <span aria-hidden="true">Option 2</span>
    <span class="visually-hidden"><!-- "Group 1 Option 2" or "Option 2" depending on entry direction --></span>
  </li>
  <!-- Outside the group now -->
  <li role="option">
    <span aria-hidden="true">Option 3</span>
    <span class="visually-hidden">Option 3</span>
  </li>
  <!-- ... -->
</ul>
```

The ideal situation would obviously be not to use visually hidden spans to hint at group labels on first entry, but to actually use groups. We avoided doing this before for ambiguous accessibility reasons, but I don't have any context on that. So the ideal situation would be something like:

```html
<div role="listbox">
  <div role="group" aria-labelledby="group-1-heading">
    <div role="presentation" id="group-1-heading">Group heading 1</div>
    <div role="option">Option 1</div>
    <div role="option">Option 2</div>
  </div>
  <!-- Outside the group now -->
  <div role="option">Option 3</div>
  <!-- ... -->
</div>
```

This PR achieves the grouping/labelling part, but doesn't address the visually hidden part yet. I personally know messing with the latter causes some teams' tests to break, so it's safer to do this in chunks. This is the more meaningful change. To avoid creating a weird nesting hell of `ul`/`li` elements, and since their roles are being overridden anyway, I switched them all to be divs to allow for easier nesting of groups inside listboxes. Otherwise, we'd have a nightmare of `ul` > `li` > `ul` > `li` nesting for no accessibility benefit.

I achieved this as pragmatically as possible. I started trying to rewrite the core options logic, and I realized this was way more work and way more risk than I wanted out of an accessibility fix. So I wrote a little utility to "re-nest" the items after all the internal filtering/selection logic is taken care of, but just before they're rendered.

Related links, issue #, if available: AWSUI-61101

### How has this been tested?

Updated our tests. Have done a dry run at multiple stages.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/select/__tests__/use-select.test.ts
+++ b/src/select/__tests__/use-select.test.ts
@@ -137,6 +137,7 @@ describe('useSelect', () => {
         option: { type: 'child', option: { label: 'Child 1', value: 'child1' } },
         selected: false,
         isNextSelected: false,
+        isPreviousSelected: false,
         indeterminate: false,
         id: getOptionId(menuId!, 1),
       });
@@ -156,6 +157,7 @@ describe('useSelect', () => {
       option: { type: 'parent', option: { label: 'Group 3', options: [{ value: 'child31', label: 'Child 3-1' }] } },
       selected: false,
       isNextSelected: true,
+      isPreviousSelected: false,
       indeterminate: false,
       id: getOptionId(menuId!, 7),
     });
@@ -168,6 +170,7 @@ describe('useSelect', () => {
       option: { type: 'child', option: { value: 'child31', label: 'Child 3-1' } },
       selected: true,
       isNextSelected: false,
+      isPreviousSelected: false,
       indeterminate: false,
       id: getOptionId(menuId!, 8),
     });


src/internal/components/option/option-parts.tsx
```diff
@@ -117,11 +117,7 @@ export const FilteringTags = ({
         const match = filteringTag.toLowerCase().indexOf(searchElement) !== -1;
         if (match) {
           return (
-            <span
-              className={clsx(styles.tag, triggerVariant && styles['trigger-variant'])}
-              key={key}
-              aria-disabled={true}
```
avinashbot, line 123: This line actually does nothing. This wasn't raised before because the whole thing was wrapped inside an `aria-hidden={true}`, which meant all of this was skipped.

src/internal/components/option/index.tsx
```diff
@@ -69,18 +68,14 @@ const Option = ({
     />
   );
 
-  // The option is conditionally assigned 'title' and 'aria-disabled' attributes to ensure it is viewed as a (generic) group  by assistive technology only when necessary.
-  // Omitting the props might be necessary if they are provided on the parent element to avoid nested groups.
-  // See https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/generic_role
-  const genericGroupProps = isGenericGroup
-    ? {
-        title: option.label ?? option.value,
-        'aria-disabled': disabled,
```
avinashbot, line 78: This also failed accessibility checks because it was previously marked as `aria-hidden`. Putting `aria-disabled` on a generic span does (well, it's supposed to) do nothing.

src/select/utils/render-options.tsx
```diff
@@ -71,5 +79,26 @@ export const renderOptions = ({
...
+  };
+
+  const unflattenedOptions = unflattenOptions(options);
+  return unflattenedOptions.map(nestedDropdownOption => {
+    const index = nestedDropdownOption.index;
+    const props = getNestedItemProps(nestedDropdownOption);
+
+    if (nestedDropdownOption.type === 'parent') {
+      const { children } = nestedDropdownOption;
+      const optionId = props.id ?? `${idPrefix}-option-${index}`;
```
avinashbot, line 92: Previously, we called `getItemProps()` inside the `renderListItem()` function, but it makes sense to pull this logic "up" to this level because it's helpful to check some of those props and apply it to the group itself, like `id` or `aria-disabled`.

YueyingLu (dismissed):

YueyingLu (approved):

PR #3835, opened 2025-09-03
chore: Small fix for wizard focus effect to handle React 18 strict mode

pan-kot:
### Description

The PR changes Wizard's implementation of the header auto-focus slightly, so that it does not break tests in React 18 + strict-mode. That is not a real bug: React 18 in strict-mode fires useEffectOnUpdate callback even when the dependencies never change. That is not expected to happen in production environment under any circumstances.

### How has this been tested?

* I tested it with React 18 tests setup: https://github.com/cloudscape-design/components/pull/3829

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/wizard/wizard-form.tsx
```diff
@@ -52,11 +53,12 @@ export default function WizardFormWithAnalytics(props: WizardFormProps) {
   const __internalRootRef = useComponentMetadata('WizardForm', PACKAGE_VERSION, analyticsMetadata as AnalyticsMetadata);
   const stepHeaderRef = useRef<HTMLDivElement | null>(null);
 
-  useEffectOnUpdate(() => {
-    if (stepHeaderRef && stepHeaderRef.current) {
+  const previousActiveStepIndex = usePrevious(props.activeStepIndex);
+  useEffect(() => {
```
just-boris: React documentation specifically calls such "fix" an anti-pattern: https://react.dev/learn/synchronizing-with-effects#how-to-handle-the-effect-firing-twice-in-development

Thanks for finding React 18 compatibility issues, but we need to fix them properly.

pan-kot: I read the React guidelines, too - but I couldn't find a good answer for this particular problem. We don't need any cleanup here, but we do need to ensure the focus() only fires after step change but never on the first render.

src/internal/hooks/use-effect-on-update.ts
```diff
@@ -1,18 +1,34 @@
...
-import { DependencyList, EffectCallback, useEffect, useRef } from 'react';
+
+import { DependencyList, EffectCallback, useEffect } from 'react';
+
+import { usePrevious } from './use-previous';
 
 // useEffect, which skips the initial render
 export function useEffectOnUpdate(callback: EffectCallback, deps: DependencyList) {
-  const isFirstRender = useRef<boolean>(true);
+  const previousDeps = usePrevious(deps);
```
just-boris: It is better to track previous deps with a local ref without `usePrevious` hook for two reasons

1. `usePrevious` contains an extra `useEffect` which you can save
2. Having all vanilla React hooks here will help with static analysis, linting, etc.

src/internal/hooks/use-effect-on-update.ts
```diff
@@ -1,18 +1,34 @@
...
 
   useEffect(() => {
-    if (isFirstRender.current) {
-      isFirstRender.current = false;
-    } else {
+    // We double-check if the dependencies are actually different for the hook to work
+    // expectedly with React 18 in strict-mode, which fires useEffect callback twice.
+    if (previousDeps !== undefined && !isDepsEqual(previousDeps, deps)) {
```
just-boris: When we have such tricks, it would be useful to check them with a unit test running across all supported React versions, what do you think?

just-boris (dismissed):
Approved with nice to have comments

src/wizard/wizard-form.tsx
just-boris: Thanks, the second version looks better

src/internal/hooks/use-effect-on-update.ts
pan-kot: I added unit tests, but for now only running them with React 16. I will add React 18 support for unit tests as a follow-up, as it covers all tests, not just this one.

PR #3839, opened 2025-09-04
chore: Fix visual diffs caused by #3833

avinashbot:
### Description

Some of the last few pixel changes I made got missed in my dev pipeline because they were getting compared with the larger select icon changes. Mostly me undoing whatever CSS I added/removed because I thought it wasn't important.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/components/selectable-item/styles.scss
+++ b/src/internal/components/selectable-item/styles.scss
@@ -105,6 +105,7 @@
     }
   }
 
+  &.selected + &.selected,
   &.selected.previous-item-selected {
     border-start-start-radius: 0;
     border-start-end-radius: 0;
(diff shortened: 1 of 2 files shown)


src/select/parts/styles.scss
```diff
@@ -30,8 +30,7 @@ $inlineLabel-border-radius: 2px;
 
 .option-group {
   // Mirrors the negative margin styling of the selectable item.
-  &:not(:first-child),
-  &.virtual {
```
avinashbot, line 34: `.virtual` items already have a negative block-start margin, regardless of whether or not they're in a group or not, so this was unnecessary.

src/internal/components/selectable-item/styles.scss
```diff
@@ -105,6 +105,7 @@
     }
   }
 
+  &.selected + &.selected,
```
avinashbot, line 108: I should have just added a rule here instead of replacing; this was still doing things for cases like the select-all checkbox, which is really hard to target at this level without major JS refactoring.

YueyingLu (approved):

PR #3842, opened 2025-09-05
chore: Fixes virtual dropdown tests for React 18

pan-kot:
### Description

This fixes failing autosuggest tests with React 18, reproducible here: https://github.com/cloudscape-design/components/pull/3829

Rel: AWSUI-61204

### How has this been tested?

* I run the related integ tests for autosuggest and select a few dozen times locally and got no failures

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/__integ__/page-objects/async-dropdown-page.ts
+++ b/src/__integ__/page-objects/async-dropdown-page.ts
@@ -69,6 +69,7 @@ class AsyncDropdownPage extends BasePageObject {
       items,
       hasNextPage
     );
+    await this.pause(10);
   }
 
   async reject() {


src/__integ__/page-objects/async-dropdown-page.ts
```diff
@@ -69,6 +69,7 @@ class AsyncDropdownPage extends BasePageObject {
       items,
       hasNextPage
     );
+    await this.pause(10);
```
just-boris: Why is this happening?

pan-kot: Here is what I observed. I this test:
```
test(
  'the bottom of the dropdown lines up with the highlighted option, when going down the list',
  setup({ virtualScrolling }, async page => {
    await page.respondWith(0, true);
    await page.keys(['ArrowDown', 'ArrowDown', 'ArrowDown']);
    const { bottom: optionBottom } = await page.getHighlightedPosition();
    const { bottom: dropdownBottom } = await page.getDropdownPosition();
    // a small give for borders and negative margins
    expect(Math.abs(optionBottom - dropdownBottom)).toBeLessThan(3);
  })
);
```
we issue 3 keydown events. However, only two keydown events are received (sometimes, it is flaky).

Screenshot 2025-09-10 at 14 41 50

I think there is a chance that the key-down handler is called before the options are ready, in which case the 1st keydown is just ignored. Adding a small delay fixes the problem.

just-boris: We have a reusable utility for this case: https://github.com/cloudscape-design/browser-test-tools/blob/2f3ae51ccd4f4a9869d09b4701e02b2c4322c0ff/src/page-objects/base.ts#L162-L164

Let's use that one

PR #3844, opened 2025-09-05
feat: Support secondaryText and labelTag for button-dropdown

kuahuu:
### Description

Added support for secondaryText and labelTag properties to ButtonDropdown items, enabling richer item content similar to other Cloudscape components like Select and Multiselect.

This PR also added deprecated annotation for `description` prop for button dropdown item.

New features:
**secondaryText**: Displays additional descriptive text below the main item text
**labelTag**: Adds a visual tag/label to items for categorization or status indication

Implementation details:
* Extended ButtonDropdownProps.Item interface to include the new properties
* Updated ItemElement component to render secondary text and label tags
* Added corresponding styles for proper layout and visual hierarchy
* Added comprehensive test coverage for the new functionality
* Add `deprecated` annotation into the `description` prop
* Introduce `findSecondaryText` and `findLabelTag` test utils

#### Unit Tests:
* Added tests in button-dropdown-items.test.tsx for secondaryText rendering
* Added tests in button-dropdown-keyboard.test.tsx for keyboard navigation with secondaryText
* All existing ButtonDropdown tests continue to pass

#### Manual Testing:
* Verified visual appearance and layout of items with secondaryText and labelTag
* Tested keyboard navigation and accessibility
* Confirmed backward compatibility with existing ButtonDropdown usage

#### Reviewers can test by running:
```bash
npm test -- --testPathPattern="button-dropdown"
npm run test:integ -- --testPathPattern="button-dropdown"
```

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/button-dropdown/disabled-reason.page.tsx
+++ b/pages/button-dropdown/disabled-reason.page.tsx
@@ -126,7 +126,7 @@ export const selectableGroupItems: ButtonDropdownProps.Items = [
         itemType: 'checkbox',
         checked: true,
         disabled: true,
-        disabledReason: '',
+        disabledReason: 'disabled reason',
       },
     ],
   },


src/__tests__/snapshot-tests/__snapshots__/documenter.test.ts.snap
cansuaa, line 1: There are different changes in the file, I only expect to see your changes though. Could you please merge from main again and update the snapshot? I believe there was a change regarding generating the documenter and that is why this has different changes.

kuahuu: Merged the changes from main but still see the gaps in documenter snap. 

It seems some description/text has been reorders, could be related to the generation logic change you mentioned

src/button-dropdown/interfaces.ts
```diff
@@ -24,11 +24,13 @@ export interface ButtonDropdownProps extends BaseComponentProps, ExpandToViewpor
...
    * - `text` (string) - description shown in the menu for this item. Mandatory for individual items, optional for categories.
+   * - `secondaryText` (string) - (Optional) Secondary text displayed in the menu for this item.
    * - `lang` (string) - (Optional) The language of the item, provided as a BCP 47 language tag.
    * - `disabled` (boolean) - whether the item is disabled. Disabled items are not clickable, but they can be highlighted with the keyboard to make them accessible.
    * - `disabledReason` (string) - (Optional) Displays text near the `text` property when item is disabled. Use to provide additional context.
-   * - `description` (string) - additional data that will be passed to a `data-description` attribute.
+   * - `description` (string) - additional data that will be passed to a `data-description` attribute. **Deprecated**, has no effect.
    * - `ariaLabel` (string) - (Optional) - ARIA label of the item element.
+   * - `labelTag` (string) - (Optional) - Label for the item. Displayed near the item text for visually impaired users.
```
cansuaa: The new APIs shouldn't be added here because this list is all the APIs available in all types which is not the case for the new APIs, they are only available in action and checkbox so instead of here, they should be down

src/button-dropdown/item-element/styles.scss
```diff
@@ -106,3 +106,21 @@
...
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  gap: awsui.$space-xxs;
+}
+
+.text-with-icon {
+  display: flex;
+  align-items: center;
+}
```
cansuaa: This is not used anywhere?

src/__tests__/snapshot-tests/__snapshots__/documenter.test.ts.snap
kuahuu: Pull the latest main and rebuild the documenter.test.ts.snap file.

src/button-dropdown/interfaces.ts
kuahuu: Moved these 2 API description below to applicable types.

src/button-dropdown/item-element/styles.scss
kuahuu: Removed in latest PR

kuahuu:
Rebased from the latest origin main and resolve the conflict.

src/test-utils/dom/button-dropdown/index.ts
```diff
@@ -100,6 +100,20 @@ export default class ButtonDropdownWrapper extends ComponentWrapper {
...
+   * Finds the secondary text element for a dropdown item. Returns null if the element has no secondary text.
+   */
+  findSecondaryText(): ElementWrapper | null {
+    return createWrapper().find(`[data-testid="button-dropdown-secondary-text"]`);
+  }
+
+  /**
+   * Finds the label tag element for a dropdown item. Returns null if the element has no label tag.
+   */
+  findLabelTags(): ElementWrapper | null {
```
cansuaa: `findLabelTag` is more correct

src/button-dropdown/item-element/index.tsx
```diff
@@ -175,8 +176,24 @@ const MenuItemContent = ({ item, disabled }: { item: InternalItem | InternalChec
...
+      <div className={styles['content-wrapper']}>
+        <div className={styles['main-row']}>
+          <div>
+            {item.text}
+            {hasExternal && <ExternalIcon disabled={disabled} ariaLabel={item.externalIconAriaLabel} />}
+          </div>
+          {item.labelTag && (
+            <InternalBox fontWeight="light" data-testid={'button-dropdown-label-tag'}>
+              {item.labelTag}
+            </InternalBox>
```
cansuaa: Instead of having data-testid, I suggest wrapping `{item.labelTag}` with a span and a className. For className to work, you'll need to have the className defined in the styles files with a comment like this:
```
.label-tag {
  /* used in test-utils */
} 
```

Additionally, something we started doing recently is having a different styles file where we add the classNames used for test-util so here you can follow a similar approach where you create a folder called `test-classes` under button-dropdown folder and create a new `styles.scss` where you have this class defined. The same thing applies to the secondaryText below. You don't need to define any other class in that - the old ones should stay as is.

src/button-dropdown/interfaces.ts
```diff
@@ -39,6 +39,8 @@ export interface ButtonDropdownProps extends BaseComponentProps, ExpandToViewpor
    * - `iconAlt` (string) - (Optional) Specifies alternate text for the icon when using `iconUrl`.
    * - `iconUrl` (string) - (Optional) Specifies the URL of a custom icon.
    * - `iconSvg` (ReactNode) - (Optional) Custom SVG icon. Equivalent to the `svg` slot of the [icon component](/components/icon/).
+   * - `secondaryText` (string) - (Optional) Secondary text displayed in the menu for this item.
+   * - `labelTag` (string) - (Optional) - Label for the item. Displayed near the item text for visually impaired users.
```
cansuaa: In the documenter snapshot these still appear in their old place, could you update the snapshot again?

pages/button-dropdown/events.page.tsx
cansuaa, line 1: The integ tests are failing since these pages are updated but the tests not. Could you run them locally and update as expected?

src/button-dropdown/item-element/index.tsx
cansuaa: Or actually, instead of using InternalBox, use a div with a classname and apply the styles to the div. For color you can use this design token `color-text-dropdown-item-secondary`, this way the users who wish to change the colors will be able to do so. With InternalBox a different design token would be used which is an implementation detail and users wouldn't expect that design token to be used here.

cansuaa: And one more thing, the label tag shouldn't have the same color with description. In select component, it is that way and it's better to keep it consistent.

src/test-utils/dom/button-dropdown/index.ts
kuahuu: Updated to `findLabelTag`

src/button-dropdown/item-element/index.tsx
kuahuu: Replaced the InternalBox with div + css classes defined in the styles.scss file.

But the description and label tag seem use the same color in `Select` component
Screenshot 2025-10-10 at 1 39 41 AM

pages/button-dropdown/events.page.tsx
kuahuu: Can you suggest how to resolve the following issue when run `npm run test:integ`?

```
FatalError: Jest: Got error running globalSetup - /Volumes/workplace/cloudscape-components/components/build-tools/integ/global-setup.js, reason: Cannot find local chromedriver. Did you run 'npm i -g chromedriver'?
```

src/button-dropdown/interfaces.ts
kuahuu: Pull and build again and now new props are showing on the right place

package.json
```diff
@@ -177,7 +177,7 @@
     {
       "path": "lib/components/internal/widget-exports.js",
       "brotli": false,
-      "limit": "962 kB",
+      "limit": "963 kB",
```
timogasda: are we sure this is needed?

pages/button-dropdown/permutations-main-action.page.tsx
```diff
@@ -17,13 +17,15 @@ const launchInstanceItem = {
...
   iconName: 'file',
+  secondaryText: 'Use an existing template',
 } as const;
 
 const viewInstancesItem = {
   text: 'View instances',
   href: 'https://instances.com',
   external: true,
   externalIconAriaLabel: '(opens in new tab)',
+  labelTag: 'Ctrl + Click',
```
timogasda: Something broke here. If you go to this permutations page in the browser, you'll see the following error message: 

> hook.js:608 Warning: React does not recognize the `labelTag` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `labeltag` instead. If you accidentally passed it from a parent component, remove it from the DOM element.

That's also why the a11y tests are failing

src/button-dropdown/__integ__/button-dropdown-expandable.test.ts
timogasda, line 1: There's a few integ tests that are failing. This is because you've added the new descriptions in a lot of places, which is now breaking the element text assertions.

Please 1) make sure to update test assertions where needed, and 2) review if we really need to add the new descriptions to all 9 button-dropdown pages that you're currently touching. I suggest we keep it simpler and only add it to the permutation pages and maybe 1-2 integ test pags.

pages/button-dropdown/events.page.tsx
timogasda: You need to follow these instructions: https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#run-integration-tests.

Basically, you globally install `chromedriver` and then run the integ tests. You can specifically run the button dropdown tests only using this command:

```bash
npx jest -c jest.integ.config.js src/button-dropdown/__integ__
```

Make sure that you have the dev server (`npm start`) running somewhere else in the background

src/button-dropdown/item-element/index.tsx
timogasda: As Cansu said, the label tag should be darker, just like in select. See the "128 Gb" here:
Screenshot 2025-10-10 at 10 29 17

kuahuu: ahh seems there are both label tag and tags on the selector, and I was referring to tags. 

Updated to `color: awsui.$color-text-body-default;` as reference

Screenshot 2025-10-23 at 2 38 53 PM

pages/button-dropdown/events.page.tsx
kuahuu: Thanks for the guidance. I'm now able to run the integration test and resolve all failure.

package.json
kuahuu: Yes, this was needed otherwise the bundle size was 962.1kb before.

pages/button-dropdown/permutations-main-action.page.tsx
kuahuu: Found the issue was these props are passing to another internal component and wasn't included in the destructuring assignment, causing it to be spread onto a DOM element. Remove the labelTag in this page as we do have labelTag in other permutation pages.

src/button-dropdown/__integ__/button-dropdown-expandable.test.ts
kuahuu: Thanks for the callout. I've removed some description out of the example pages, and make the integ tests all passed now.

package.json
cansuaa: It needs to be bumped up again actually.

src/button-dropdown/__tests__/button-dropdown-items.test.tsx
```diff
@@ -498,6 +498,250 @@ describe('Internal ButtonDropdown badge property', () => {
...
+    const items = [{ id: 'i1', text: 'Option 1' }];
+    const wrapper = renderButtonDropdown({ items });
+    wrapper.openDropdown();
+
+    expect(wrapper.findSecondaryText()).toBeNull();
+    expect(wrapper.findLabelTag()).toBeNull();
+  });
+});
+
+describe('ButtonDropdown download property', () => {
```
cansuaa: This testing block doesn't belong to your changes and seems to be duplicated, I see the same test block below. Could you remove?

kuahuu: Resolved in latest PR.

package.json
kuahuu: Bumping again and passed `npm run build` on local

pages/button-dropdown/disabled-reason.page.tsx
```diff
@@ -130,7 +142,7 @@ export default function DescriptionPage() {
   const [isRightAligned, setIsRightAligned] = useState(false);
   return (
     <>
-      <h1>Descriptions in ButtonDropdown</h1>
+      <h1>Failure Reason in ButtonDropdown</h1>
```
cansuaa: Why did we change the title of the page?

pages/button-dropdown/simple.page.tsx
```diff
@@ -89,29 +112,39 @@ const withExpandedGroups: ButtonDropdownProps['items'] = [
       {
         id: 'start',
         text: 'Start',
+        secondaryText: 'This is the Start option',
+        labelTag: 'Ctrl + S',
```
cansuaa: Minor: Among these items let's have a few items without secondaryText or labelTag

src/button-dropdown/item-element/index.tsx
```diff
@@ -176,8 +176,16 @@ const MenuItemContent = ({ item, disabled }: { item: InternalItem | InternalChec
...
-      {hasExternal && <ExternalIcon disabled={disabled} ariaLabel={item.externalIconAriaLabel} />}
+      <div className={styles['content-wrapper']}>
+        <div className={styles['main-row']}>
+          <div>
+            {item.text}
+            {hasExternal && <ExternalIcon disabled={disabled} ariaLabel={item.externalIconAriaLabel} />}
+          </div>
+          {item.labelTag && <div className={styles['label-tag']}>{item.labelTag}</div>}
+        </div>
+        {item.secondaryText && <div className={styles['secondary-text']}>{item.secondaryText}</div>}
```
cansuaa: There is a color use case we've missed. When the option is highlighted/hovered, the color of the secondaryText and the background color of hovered option don't have enough contrast hence fails at accessibility. Checking select, you can see a different color is applied when the item is hovered, let's follow the same logic. 

You can refer here and use the same color used here: https://github.com/cloudscape-design/components/blob/main/src/internal/components/option/styles.scss#L77-L79

kuahuu: Following the same pattern to add colors for secondaryText and labelTag when highlighted.

Right now, I use the same color for both text when highlighted, as I wasn't able to find any other example for highlighted color. Please let me know if they should use separate colors here.

pages/button-dropdown/disabled-reason.page.tsx
kuahuu: Not so sure why this is changed. May be caused by rebased. Updated to the original text.

pages/button-dropdown/simple.page.tsx
kuahuu: Update first 3 options without secondaryText or labelTag

src/button-dropdown/item-element/index.tsx
cansuaa: Only secondaryText needs a different color when highlighted. The labelTag already meets the contrast.

kuahuu: Reverted highlight color for labelTag.

src/button-dropdown/item-element/styles.scss
```diff
@@ -106,3 +106,30 @@
...
+
+.secondary-text {
+  @include styles.font(body-s);
+  font-weight: styles.$font-weight-lighter;
+  color: awsui.$color-text-dropdown-item-secondary;
+
+  &.highlighted {
+    color: awsui.$color-text-dropdown-item-secondary-hover;
+  }
```
cansuaa, line 137: This doesn't do anything because `highlighted` class is added to the parent `li` element. You can pass the `highlighted` value to MenuItemContent and add the class directly to secondaryText div

package.json
```diff
@@ -176,7 +176,7 @@
     {
       "path": "lib/components/internal/widget-exports.js",
       "brotli": false,
-      "limit": "977 kB",
+      "limit": "979 kB",
```
cansuaa: I've discussed this situation with the team and we bumped it to 1000kB, so merging from main will have a conflict here but you won't need to deal with this limit after that.

src/button-dropdown/item-element/styles.scss
kuahuu: Updated to inherit this prop from MenuItemContent and define the class when secondary text is highlighted

package.json
kuahuu: Merged and resolved in latest PR

package.json
cansuaa, line 1: You shouldn't need to touch package.json or package-lock.json

src/button-dropdown/item-element/index.tsx
```diff
@@ -176,8 +184,18 @@ const MenuItemContent = ({ item, disabled }: { item: InternalItem | InternalChec
...
+      <div className={styles['content-wrapper']}>
+        <div className={styles['main-row']}>
+          <div>
+            {item.text}
+            {hasExternal && <ExternalIcon disabled={disabled} ariaLabel={item.externalIconAriaLabel} />}
+          </div>
+          {item.labelTag && <div className={styles['label-tag']}>{item.labelTag}</div>}
+        </div>
+        {item.secondaryText && (
+          <div className={clsx(styles['secondary-text'], highlighted && styles.highlighted)}>{item.secondaryText}</div>
```
cansuaa: You still need to define the color in the styles. The existing css rule will apply it to `li` but since you are overriding the color for the secondary text, the existing `highlighted` will not work. You should be able to run the accessibility tests locally to ensure the test passes or you should be able to see the color change in one of the dev pages in dark mode. 

You can also see it [here](https://d21d5uik3ws71m.cloudfront.net/components/ff788182857864981b863d790cb670b463867b9c/dev-pages-react16/index.html#/dark/button-dropdown/simple/) in dark mode, the secondary text doesn't change color

kuahuu: Synced offline and added back hightlight css styling definition.

package.json
kuahuu: Reverted change of package.json and package-lock.json

src/button-dropdown/interfaces.ts
```diff
@@ -40,6 +40,8 @@ export interface ButtonDropdownProps extends BaseComponentProps, ExpandToViewpor
...
```
cansuaa: One last thing, let's make the API descriptions more explanatory;
- secondaryText: `Further information about the action that appears below the label.`
- labelTag: `A label tag that provides additional guidance, shown next to the label.`

kuahuu: Updated the API description as suggested.

src/button-dropdown/item-element/styles.scss
```diff
@@ -106,3 +106,30 @@
...
+  flex: 1;
+  display: flex;
+  flex-direction: column;
+}
+
+.main-row {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  gap: awsui.$space-xxs;
```
cansuaa: Let's increase this to be the same with select, this design token should be used `awsui.$space-s`

src/button-dropdown/item-element/styles.scss
```diff
@@ -106,3 +106,30 @@
...
+}
+
+.label-tag {
+  color: awsui.$color-text-body-default;
+}
+
+.secondary-text {
+  @include styles.font(body-s);
+  font-weight: styles.$font-weight-lighter;
```
cansuaa: The font-weight isn't needed

src/button-dropdown/item-element/styles.scss
kuahuu: Updated

src/button-dropdown/item-element/styles.scss
kuahuu: Updated

src/button-dropdown/interfaces.ts
```diff
@@ -50,6 +52,8 @@ export interface ButtonDropdownProps extends BaseComponentProps, ExpandToViewpor
...
```
cansuaa: The descriptions above are correct, could you update these two as well?

kuahuu: Updated

PR #3851, opened 2025-09-09
chore: Fixes mixed charts popover behaviour for React 18

pan-kot:
### Description

The issue was observed when we run integ tests with React 18: https://github.com/cloudscape-design/components/pull/3829

With React 18 (strict mode or not) the mixed charts popover behaves differently: when the cursor moves from popover content to a bar series it was open for, the popover does not reopen: the user has to move the cursor outside the bar, and then back inside.

That happens because the popover state is controlled with a useLayoutEffect, and it depends on a series of state changes:
1. When the cursor leaves the popover, we set the highlighted group to null (therefore the popover should be hidden).
2. When the cursor enters the next bar, we set the highlighted group to that bar index (which is the same index that was before we hide it).
3. Because both state changes happen in quick succession, React 18 ignores the first one for that layout effect, and because the next state is the same that was before - the callback does not fire, and the popover does not reopen.

See how the popover re-opens when the cursor is moved slowly, but does not reopen when it is moved fast:

https://github.com/user-attachments/assets/edfb4c3b-8194-4a38-9735-d209aec9bc83

### How has this been tested?

* Existing integration tests in React 16 and 18
* Screenshot tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/mixed-line-bar-chart/__tests__/keyboard-navigation.test.tsx
+++ b/src/mixed-line-bar-chart/__tests__/keyboard-navigation.test.tsx
@@ -4,14 +4,33 @@
 import React from 'react';
 import { render } from '@testing-library/react';
 
+import { createWrapper } from '@cloudscape-design/test-utils-core/dom';
+
 import { KeyCode } from '../../../lib/components/internal/keycode';
 import MixedLineBarChart from '../../../lib/components/mixed-line-bar-chart';
-import { MixedLineBarChartWrapper } from '../../../lib/components/test-utils/dom';
 import { barSeries, lineSeries3, thresholdSeries } from './common';
 
 describe('Keyboard navigation', () => {
-  test('opens popover for each series', () => {
-    const { container } = render(
+  function getChart() {
+    return createWrapper().findMixedLineBarChart()!;
+  }
+  function expectValues(a: Array<number>) {
+    for (let i = 0; i < a.length; i++) {
+      const value = a[i];
+      expect(getChart().findDetailPopover()!.findSeries()![i].findValue().getElement()).toHaveTextContent(
+        value.toString()
+      );
+    }
+  }
+  function focusApplication() {
+    getChart().findApplication()!.focus();
+  }
+  function goToNextDataPoint() {
+    getChart().findApplication()!.keydown(KeyCode.right);
+  }
+
+  test('opens popover for each series (mixed)', () => {
+    render(
       <MixedLineBarChart
         height={250}
         xDomain={['Potatoes', 'Chocolate', 'Apples', 'Oranges']}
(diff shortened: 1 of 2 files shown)


src/mixed-line-bar-chart/__integ__/common.ts
```diff
@@ -29,6 +29,8 @@ export class MixedChartPage extends BasePageObject {
       // If a popover was pinned, we need to unpin it before pinning another one.
       if (this.currentIndex) {
         await this.click(this.wrapper.findDetailPopover().findDismissButton().toSelector());
+        // Wait for popover dismiss reopen delay.
+        await this.pause(50);
```
pan-kot, line 33: The new implementation prevents the popover to be shown immediately after dismiss using a small timeout, which needs to be considered also in tests.

src/mixed-line-bar-chart/hooks/use-popover.ts
```diff
@@ -1,16 +1,31 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
-import { useCallback, useState } from 'react';
+
+import { useCallback, useRef, useState } from 'react';
+
+// The delay prevents re-opening popover immediately upon dismissing,
+// so that the popover actually closes. It can be reopened with the next
+// hover or focus event that occurs after the delay.
+const REOPEN_DELAY_MS = 50;
```
pan-kot, line 9: The popover should not be reopen immediately upon dismissal - that is an existing requirement. That was ensured by checking the highlighted series index/id: if same, the popover will not reopen. The new implementation achieves the same with a small timeout, yet now it is also possible to re-open the popover shortly after it is dismissed by hovering over the same series the popover was open for before.

src/mixed-line-bar-chart/chart-container.tsx
```diff
@@ -255,9 +255,10 @@ export default function ChartContainer<T extends ChartDataTypes>({
           scaledX: point.x,
           label: point.datum?.x ?? null,
         });
+        showPopover();
```
pan-kot, line 255: We now show the popover the moment a point or series is highlighted, without relying on the dependencies change.

src/mixed-line-bar-chart/chart-container.tsx
```diff
@@ -283,14 +285,15 @@ export default function ChartContainer<T extends ChartDataTypes>({
...
     },
-    [highlightSeries, setHighlightedPoint, setHighlightedGroupIndex]
+    [highlightSeries, setHighlightedPoint, setHighlightedGroupIndex, showPopover]
   );
 
   const clearHighlightedSeries = useCallback(() => {
     clearAllHighlights();
-    dismissPopover();
-  }, [dismissPopover, clearAllHighlights]);
+    hidePopover();
```
pan-kot, line 292: Hiding the popover (e.g. when the cursor leaves a series) is not the same as dismissing it - we should not prevent its reopen in that case.

src/mixed-line-bar-chart/chart-container.tsx
```diff
@@ -349,12 +352,6 @@ export default function ChartContainer<T extends ChartDataTypes>({
     return () => document.removeEventListener('keydown', onKeyDown);
   }, [dismissPopover]);
 
-  useLayoutEffect(() => {
-    if (highlightedX !== null || highlightedPoint !== null) {
-      showPopover();
-    }
-  }, [highlightedX, highlightedPoint, showPopover]);
```
pan-kot, line 356: If highlightedX was `1` (showing the popover for the 1st group) and then it changes as `1` -> `null` -> `1` in a quick succession, that intermediate transition to `null` is ignored by React 18. As result, the effect callback is not fired, thus not showing the popover.

src/mixed-line-bar-chart/__tests__/keyboard-navigation.test.tsx
```diff
@@ -45,4 +48,24 @@ describe('Keyboard navigation', () => {
     goToNextDataPoint();
     expectValues([47, 7, 8]);
   });
+
+  test('opens popover for each series (line)', () => {
```
pan-kot, line 52: This test is needed to increase test coverage of the mixed chart container

PR #3852, opened 2025-09-09
funnel tests experiment

pan-kot:
### Description

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/analytics/components/analytics-funnel.tsx
+++ b/src/internal/analytics/components/analytics-funnel.tsx
@@ -182,12 +182,16 @@ const InnerAnalyticsFunnel = ({ mounted = true, children, stepConfiguration, ...
       setFunnelInteractionId(funnelInteractionId);
     }, 1);
 
-    /*
-      A funnel counts as "successful" if it is unmounted after being "complete".
-    */
+    // A funnel counts as "successful" if it is unmounted after being "complete".
     /* eslint-disable react-hooks/exhaustive-deps */
     return () => {
       clearTimeout(handle);
+
+      // There is no need in cleanup if the funnel was not started.
+      if (!funnelInteractionId) {
+        return;
+      }
+
       if (props.funnelType === 'single-page' && wizardCount.current > 0) {
         return;
       }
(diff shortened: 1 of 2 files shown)


src/internal/analytics/components/analytics-funnel.tsx
```diff
@@ -182,12 +182,16 @@ const InnerAnalyticsFunnel = ({ mounted = true, children, stepConfiguration, ...
...
-    /*
-      A funnel counts as "successful" if it is unmounted after being "complete".
-    */
+    // A funnel counts as "successful" if it is unmounted after being "complete".
     /* eslint-disable react-hooks/exhaustive-deps */
     return () => {
       clearTimeout(handle);
+
+      // There is no need in cleanup if the funnel was not started.
+      if (!funnelInteractionId) {
```
pan-kot, line 191: The funnel is started with a 1ms timeout in the code above. If the useEffect is called in a quick succession, which is the case in React 18 strict mode, the first cleanup runs before the setTimeout callback. In that case we issue the "cancelled" event, even though the funnel was not actually started. The updated code prevents just that.

src/internal/components/focus-lock/index.tsx
```diff
@@ -44,21 +44,25 @@ function FocusLock(
   // Captures focus when `autoFocus` is set, and the component is mounted or
   // `disabled` changes from true to false.
   useEffect(() => {
+    const assignRestoreFocusTarget = () => {
+      if (document.activeElement && !containerRef.current?.contains(document.activeElement as Node)) {
```
pan-kot, line 48: The funnel complete event is fired when the focus leaves the elements of the funnel. In case the funnel includes a modal, and that modal is opened and then dismissed - the focus is expected to come back to the trigger button, so it stays within the funnel. In React 18 strict mode, the focus does not return to the trigger (and goes to the body), causing the complete event and also worsening the UX and a11y of the modals.

This happens because the code in this useEffect also fires after the modal is opened and the dismiss button is auto-focused. As result, the document.activeElement points not to the trigger, but to the dismiss button instead, which is unexpected. We prevent that by ensuring the active element is outside of the focus-locked container.

PR #3860, opened 2025-09-15
feat: Add radio button component to Core

jperals:
### Description

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/internal/components/radio-button/interfaces.ts
```diff
@@ -71,9 +71,6 @@ export interface RadioButtonProps extends BaseComponentProps {
    */
   readOnly?: boolean;
 
-  /**
```
pan-kot, line 74: Why is that removed?

jperals: Because the component itself is only provided to Core.

But this is also imported from Radio Group, which means non-Core customers will be able to arrive here by inspecting the Radio Group interfaces, so I am adding this back so that both interfaces (Radio Group and internal Radio Button) are kept consistent.

src/radio-button/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+import InternalRadioButton from '../internal/components/radio-button';
+import useBaseComponent from '../internal/hooks/use-base-component';
+import { applyDisplayName } from '../internal/utils/apply-display-name';
+import { RadioButtonProps } from './interfaces';
+
+import styles from './styles.css.js';
+
+export { RadioButtonProps };
+
+const RadioButton = React.forwardRef((props: RadioButtonProps, ref: React.Ref<HTMLInputElement>) => {
```
pan-kot, line 16: should we use `{ disabled = false, readOnly = false, ...props }` here so that there props default value is correctly assigned with documenter?

src/radio-button/__tests__/radio-button.test.tsx
```diff
@@ -0,0 +1,225 @@
...
+
+    rerender(<RadioButton name="group" checked={true} onSelect={onSelect} />);
+    expect(radioButton.findNativeInput().getElement()).toBeChecked();
+
+    rerender(<RadioButton name="group" checked={false} onSelect={onSelect} />);
+    expect(onSelect).not.toHaveBeenCalled();
+  });
+});
+
+test('style API', () => {
```
pan-kot: nit: I would move this test to a separate file e.g. radio-button.style.test.tsx

src/radio-button/__tests__/radio-button.test.tsx
```diff
@@ -0,0 +1,225 @@
...
+      <RadioButton name="group" checked={true} nativeInputAttributes={{ className: 'additional-class' }} />
+    );
+    const input = container.querySelector('input');
+    expect(input).toHaveClass(abstractSwitchStyles['native-input']);
+    expect(input).toHaveClass('additional-class');
+  });
+});
+
+describe('events', () => {
+  test('fires a single onSelect event on input click', () => {
```
pan-kot, line 34: I think there is a missing test for not firing onSelect when checked={true}

src/radio-button/__tests__/radio-button.test.tsx
```diff
@@ -0,0 +1,225 @@
...
+    expect(radioButton.findNativeInput()!.getElement().getAttribute('name')).toBe('my-radio-group-name');
+  });
+
+  test('propagates the value of the `value` prop to the `value` attribute of the native element', () => {
+    const radioButton = renderRadioButton(<RadioButton value="my-radio-button-value" name="group" checked={false} />);
+    expect(radioButton.findNativeInput()!.getElement().getAttribute('value')).toBe('my-radio-button-value');
+  });
+});
+
+describe('native attributes API', () => {
```
pan-kot: This can also probably go to a separate file. Maybe we can group core-related tests into radio-button-core.test.tsx? What do you think?

pages/radio-button/focus-test.page.tsx
```diff
@@ -0,0 +1,27 @@
...
+import React, { useState } from 'react';
+
+import RadioButton from '~components/radio-button';
+
+import ScreenshotArea from '../utils/screenshot-area';
+
+export default function RadioButtonScenario() {
+  const [checked, setChecked] = useState(false);
+  return (
+    <article>
```
pan-kot, line 13: nit: There is now a SimplePage template that can be used here instead.

src/radio-button/__tests__/radio-button.test.tsx
jperals: Moved into `radio-button.core.test.tsx` as per your suggestion further down

src/radio-button/__tests__/radio-button.test.tsx
jperals: Added, and actually also fixed the implementation because it was being fired also when already checked :)

src/radio-button/__tests__/radio-button.test.tsx
jperals: Done, refactored the tests into three separate files.

pages/radio-button/focus-test.page.tsx
jperals: Thanks, I was not aware. For this file, I kept this slightly more verbose implementation because SimplePage isn't meant for having the focus target before the screenshot area (not that it can't be done, but there was some nitpicky issues with that). On the other hand, I refactored the focus target to use the existing FocusTarget component and adopted the SimplePage in the other two dev pages.

src/radio-button/index.tsx
jperals: Discussed offline, the documentation website will show the default values even if these are not in the Documenter snapshot.

src/radio-group/interfaces.ts
```diff
@@ -88,7 +88,12 @@ export namespace RadioGroupProps {
     value: string;
   }
 
-  export type Ref = RadioButtonProps.Ref;
+  export interface Ref {
```
jperals, line 91: Duplicating this instead of importing from RadioButton because for some reason, the Typescript generation of type declarations would import the top-level RadioButtonProps import instead of the internal one, causing the Console build to break.

pages/radio-button/permutations.page.tsx
```diff
@@ -0,0 +1,18 @@
...
+// SPDX-License-Identifier: Apache-2.0
+import React from 'react';
+
+import { SimplePage } from '../app/templates';
+import PermutationsView from '../utils/permutations-view';
+import { permutations, RadioButtonPermutation } from './common';
+
+export default function RadioButtonPermutations() {
+  return (
+    <SimplePage title="RadioButton permutations">
```
pan-kot: For permutation pages we need screenshot area. The simple page template does not add it by default, but it can be added as such: `<SimplePage screenshotArea={{}}>`

jperals: OK, thanks, added

PR #3862, opened 2025-09-15
fix: Fixes mixed chart popover reopen in tests

pan-kot:
### Description

The charts popover should not reopen immediately after it was unpinned and dismissed. The prev implementation did not checked for popover's prev state, which means that any call to dismissPopover() would result in the timeout reset. That affected our tests where we issue the Escape command and then navigate the popover immediately after.

### How has this been tested?

* New unit test for Escape listener that secures regression

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/mixed-line-bar-chart/__tests__/mixed-chart.test.tsx
+++ b/src/mixed-line-bar-chart/__tests__/mixed-chart.test.tsx
@@ -981,6 +981,32 @@ describe('Details popover', () => {
     expect(wrapper.findByClassName(styles.exiting)).not.toBeNull();
   });
 
+  test('can be hidden with Escape and then reopened', () => {
+    const { wrapper } = renderMixedChart(<MixedLineBarChart {...mixedChartProps} />);
+    document.body.classList.add('awsui-motion-disabled');
+    const isPopoverVisible = () => !wrapper.findByClassName(styles.exiting);
+    const isPopoverPinned = () => !!wrapper.findDetailPopover()!.findDismissButton();
+
+    wrapper.findApplication()!.focus();
+    expect(isPopoverVisible()).toBe(true);
+
+    document.dispatchEvent(new KeyboardEvent('keydown', { key: 'Escape', bubbles: true }));
+    expect(isPopoverVisible()).toBe(false);
+
+    wrapper.findApplication()!.keydown(KeyCode.right);
+    expect(isPopoverVisible()).toBe(true);
+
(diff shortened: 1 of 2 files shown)


src/mixed-line-bar-chart/__tests__/mixed-chart.test.tsx
```diff
@@ -981,6 +981,32 @@ describe('Details popover', () => {
...
+    const isPopoverPinned = () => !!wrapper.findDetailPopover()!.findDismissButton();
+
+    wrapper.findApplication()!.focus();
+    expect(isPopoverVisible()).toBe(true);
+
+    document.dispatchEvent(new KeyboardEvent('keydown', { key: 'Escape', bubbles: true }));
+    expect(isPopoverVisible()).toBe(false);
+
+    wrapper.findApplication()!.keydown(KeyCode.right);
+    expect(isPopoverVisible()).toBe(true);
```
pan-kot, line 996: Without the fix, this unit test fails here. That is because we try to show the popover immediately after it was dismissed with Escape, before the 50ms delay is over. However, the delay is only relevant when the popover was pinned before dismissing.

src/mixed-line-bar-chart/__tests__/mixed-chart.test.tsx
```diff
@@ -981,6 +981,40 @@ describe('Details popover', () => {
     expect(wrapper.findByClassName(styles.exiting)).not.toBeNull();
   });
 
+  test('can be hidden with Escape and then reopened', async () => {
+    const { wrapper } = renderMixedChart(<MixedLineBarChart {...mixedChartProps} />);
+    const isPopoverVisible = () => !wrapper.findByClassName(styles.exiting);
```
jperals, line 986: Why check for the `styles.exiting` classname to assert whether the popover is visible? Why not use our own `findDetailPopover()` test util instead?

pan-kot: That is because of animations. Apparently, the findDetailPopover() can return true even after we trigger it to hide. We use the same checks in the existing tests

jperals: If animations interfere with unit tests, wouldn't it be better to disable them? We recommend this to customers and do it ourselves in flashbar tests: https://github.com/cloudscape-design/components/blob/2a71e86ec9468a5e3cb22c95c88f7523b55b6fe9/src/flashbar/__tests__/flashbar.test.tsx#L440-L443

PR #3863, opened 2025-09-15
feat: Add a new whenAppLayoutReady to the plugin api

georgylobko:
### Description

Added a new `whenAppLayoutReady` to the plugin api to keep track of `AppLayout` state loading. It's useful when left drawer needs to be updated - in this case making sure that `AppLayout` has already loaded is necessary.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/internal/plugins/widget/internal.ts
```diff
@@ -49,6 +52,13 @@ export function isAppLayoutReady() {
   return !!getAppLayoutMessageHandler();
 }
 
+/**
+ * Returns a promise that resolves once the app layout has loaded
+ */
+export function whenAppLayoutReady() {
+  return appLayoutReadyDefer.promise;
```
just-boris: 1. This code does not work, because the remove widget and the app layout widget are separate bundles, so they will refer to a different `appLayoutReadyDefer` instance

2. What is going to happen if app layout renders and then unmounts? This code resolves the promise instantly, but there is no app layout on the page

3. What happens if app layout never renders? The promise will hang forever in this case. We should mention to the consuming team that they might want to add some timeout on their end, for example: `Promise.race([whenAppLayoutReady(), timeout(3000)])`

georgylobko: 1. Addressed. Now this promise is stored in `window` object, so any instance will have access to it
2. The purpose of this promise is to be resolved as soon as the app layout is available, it's not designed to handle other use cases such as unmounting. Do you have any suggestions on that?
3. Point taken

just-boris: 2. Reverse the control. 

```js
export function whenAppLayoutReady() {
   if(isAppLayoutReady()) {
        return Promise.resolve();
   }
   return new Promise((resolve) => appLayoutReadyCallbacks.push(resolve));
}

function registerAppLayoutHandler () {
   appLayoutReadyCallbacks.forEach(() => cb());
   appLayoutReadyCallbacks = [];

    // other code...
}
```

just-boris (commented):
See this thread https://github.com/cloudscape-design/components/pull/3863#discussion_r2351792033

src/internal/plugins/widget/internal.ts
```diff
@@ -49,6 +54,20 @@ export function isAppLayoutReady() {
...
+/**
+ * Returns a promise that resolves once the app layout has loaded
+ */
+export function whenAppLayoutReady() {
+  if (isAppLayoutReady()) {
+    return Promise.resolve();
+  }
+  const win = getWindow();
+  win[storageKeyReadyDeferCallbacks] = win[storageKeyReadyDeferCallbacks] ?? [];
+  const defer = new Defer();
```
just-boris: You do not need to introduce the defer helper any more. The code from my previous comment works without it

just-boris (approved):

PR #3816, opened 2025-08-29
fix: Ensure toggle button test utils only find toggle buttons

gethinwebster:
### Description

Previously, `findToggleButton` would also find button instances. This fixes that. (And vice-versa: `findButton` now finds no toggle buttons).

Related links, issue #, if available: n/a

### How has this been tested?

<details>
   <summary>Review checklist</summary>

_The following items are to be evaluated by the author(s) and the reviewer(s)._

#### Correctness

- _Changes include appropriate documentation updates._
- _Changes are backward-compatible if not indicated, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#public-apis)._
- _Changes do not include unsupported browser features, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#browsers-support)._
- _Changes were manually tested for accessibility, see [accessibility guidelines](https://cloudscape.design/foundation/core-principles/accessibility/)._

#### Security

- _If the code handles URLs: all URLs are validated through [the `checkSafeUrl` function](https://github.com/cloudscape-design/components/blob/main/src/internal/utils/check-safe-url.ts)._

#### Testing

- _Changes are covered with new/existing unit tests?_
- _Changes are covered with new/existing integration tests?_
</details>

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #3868, opened 2025-09-16
fix: Ensure toggle button test utils only find toggle buttons

gethinwebster:
### Description

Re-run of #3816 with adjusted implementation for better cross-version compatibility

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/button/legacy-button.page.tsx
+++ b/pages/button/legacy-button.page.tsx
@@ -9,6 +9,7 @@ import ScreenshotArea from '../utils/screenshot-area';
 export default function ButtonsScenario() {
   return (
     <article>
+      <h1>Legacy buttons</h1>
       <ScreenshotArea>
         <button className="awsui_button_vjswe_t8nlg_157">Legacy button</button>
         <button className="awsui_button_vjswe_1qz6m_157">Older legacy button</button>


georgylobko (dismissed):

src/test-utils/dom/button/index.ts
```diff
@@ -7,7 +7,8 @@ import buttonTestUtilsStyles from '../../../button/test-classes/styles.selectors
 import spinnerStyles from '../../../spinner/styles.selectors.js';
 
 export default class ButtonWrapper extends ComponentWrapper<HTMLButtonElement> {
-  static rootSelector: string = styles.button;
+  static rootSelector: string = buttonTestUtilsStyles.root;
+  static legacyRootSelector: string = 'awsui_button_vjswe';
```
georgylobko: I would suggest writing a unit test for this scenario - just copy the HTML markup for the legacy selector and verify it still works correctly. Yes, this scenario is _technically_ covered in this repository: https://github.com/cloudscape-design/test-utils/blob/c7200f200913812a551c8286bb4d1dccd5087565/src/core/test/selectors.test.ts#L116

The difference is that this selector skips the last hash part, which is why I still recommend doing it this way

src/test-utils/dom/toggle-button/index.ts
```diff
@@ -4,14 +4,14 @@ import { usesDom } from '@cloudscape-design/test-utils-core/dom';
...
 import ButtonWrapper from '../button';
 
-import buttonStyles from '../../../button/styles.selectors.js';
-import styles from '../../../toggle-button/styles.selectors.js';
+import testStyles from '../../../toggle-button/test-classes/styles.selectors.js';
 
 export default class ToggleButtonWrapper extends ButtonWrapper {
-  static rootSelector: string = buttonStyles.button;
+  static rootSelector: string = testStyles.root;
+  static legacyRootSelector: string = '';
```
just-boris, line 11: Is this empty selector needed?

gethinwebster: yes, because `extends ButtonWrapper`

just-boris: Understood. Great that you handle it, but also important learning to check all descendants when using this property on other wrappers

No more questions from me on this PR, removing myself

src/test-utils/dom/button/index.ts
gethinwebster: good call, turns out what I had didn't actually work! updated

src/button/__integ__/button.test.ts
```diff
@@ -69,3 +69,15 @@ describe('Button Style API', () => {
     })
   );
 });
+
+describe('legacy selectors', () => {
+  test(
+    'finds legacy and new buttons',
```
georgylobko, line 75: [non-blocking] I'd suggest to turn this test and its page to a unit test since it's faster and less resource-intensive than running it in the browser

src/test-utils/dom/button/index.ts
```diff
@@ -7,7 +7,8 @@ import buttonTestUtilsStyles from '../../../button/test-classes/styles.selectors
 import spinnerStyles from '../../../spinner/styles.selectors.js';
 
 export default class ButtonWrapper extends ComponentWrapper<HTMLButtonElement> {
-  static rootSelector: string = styles.button;
+  static rootSelector: string = buttonTestUtilsStyles.root;
+  static legacyRootSelector: string = 'awsui_button_vjswe_t8nlg_157';
```
georgylobko, line 11: [non-blocking just curious] doesn't the hash at the end get regenerated with every build? I checked both the cloudWatch and lambda consoles though, and it looks like they both use this selector for their buttons

gethinwebster: Yes the last hash segment changes regularly. That's why I added the test as an integ test: the `selectors` utils trim these "full" class selectors and turn them into `[class*=...]` instead, so that they are insensitive to the 2nd part of the hash. `dom` utils use the class as-is, as typically the dom test utils aren't used in multi-version environments.

src/button/__integ__/button.test.ts
gethinwebster: see below, unit and integ tests handle selectors a bit differently, so this is deliberate

georgylobko (dismissed):

georgylobko (approved):

PR #3892, opened 2025-09-24
feat: Add UAP buttons to resizable table column headers

avinashbot:
### Description

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/components/drag-handle-wrapper/styles.scss
+++ b/src/internal/components/drag-handle-wrapper/styles.scss
@@ -11,9 +11,8 @@ $direction-button-wrapper-size: calc(#{awsui.$space-static-xl} + 2 * #{awsui.$sp
 $direction-button-size: awsui.$space-static-xl;
 $direction-button-offset: awsui.$space-static-xxs;
 
-.drag-handle-wrapper {
-  position: relative;
-  display: inline-block;
+.contents {
+  display: contents;
 }
 
 .portal-overlay {
@@ -34,8 +33,7 @@ $direction-button-offset: awsui.$space-static-xxs;
 }
 
 .drag-handle {
-  position: relative;
-  display: flex;
+  display: inline-flex;
 }
 
 .direction-button-wrapper {
--- a/src/table/resizer/index.tsx
+++ b/src/table/resizer/index.tsx
@@ -1,12 +1,13 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
-import React, { useEffect, useRef, useState } from 'react';
+import React, { useCallback, useEffect, useRef, useState } from 'react';
 import clsx from 'clsx';
 
 import { useStableCallback, useUniqueId } from '@cloudscape-design/component-toolkit/internal';
 import { getIsRtl, getLogicalBoundingClientRect, getLogicalPageX } from '@cloudscape-design/component-toolkit/internal';
 import { useSingleTabStopNavigation } from '@cloudscape-design/component-toolkit/internal';
 
+import DragHandleWrapper from '../../internal/components/drag-handle-wrapper';
 import { useVisualRefresh } from '../../internal/hooks/use-visual-mode';
 import { KeyCode } from '../../internal/keycode';
 import handleKey, { isEventLike } from '../../internal/utils/handle-key';
@@ -53,7 +54,9 @@ export function Resizer({
   const resizerToggleRef = useRef<HTMLButtonElement>(null);
   const resizerSeparatorRef = useRef<HTMLSpanElement>(null);
 
+  const [isPointerDown, setIsPointerDown] = useState(false);
   const [isDragging, setIsDragging] = useState(false);
+  const [showUapButtons, setShowUapButtons] = useState(false);
   const [isKeyboardDragging, setIsKeyboardDragging] = useState(false);
   const autoGrowTimeout = useRef<ReturnType<typeof setTimeout> | undefined>();
   const [resizerHasFocus, setResizerHasFocus] = useState(false);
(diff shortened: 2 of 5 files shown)


src/table/resizer/index.tsx
```diff
@@ -53,7 +54,9 @@ export function Resizer({
   const resizerToggleRef = useRef<HTMLButtonElement>(null);
   const resizerSeparatorRef = useRef<HTMLSpanElement>(null);
 
+  const [isPointerDown, setIsPointerDown] = useState(false);
```
avinashbot, line 61: Making a distinction between `isPointerDown` and `isDragging` because we want to support a "click" interaction. Currently, a mousedown starts a drag and a mouseup ends it, which means, if you click the divider slightly off center, it jumps before showing the buttons, which looks really janky. So with this, the column widths are only adjusted when the mouse starts moving.

src/table/resizer/index.tsx
```diff
@@ -183,84 +223,127 @@ export function Resizer({
...
+    isKeyboardDragging,
+    isPointerDown,
+    resizerHasFocus,
+    onWidthUpdateCommit,
+    resizeColumn,
+    updateColumnWidth,
+    updateTrackerPosition,
+  ]);
+
+  useEffect(() => {
```
avinashbot, line 312: We added both `isDragging` and `isPointerDown` to our dependencies in the useEffect above, which means this timer get destroyed and created twice. Not a problem for real users, but this really screws with tests (since jest's time-skipping timer methods won't work if the timer is recreated).

src/internal/components/drag-handle-wrapper/styles.scss
```diff
@@ -11,9 +11,8 @@ $direction-button-wrapper-size: calc(#{awsui.$space-static-xl} + 2 * #{awsui.$sp
 $direction-button-size: awsui.$space-static-xl;
 $direction-button-offset: awsui.$space-static-xxs;
 
-.drag-handle-wrapper {
-  position: relative;
-  display: inline-block;
+.contents {
+  display: contents;
```
avinashbot, line 15: Previously, the drag handle wrapper had three nested divs I had to contend with styling them all. Since two of those divs only exist for event handling/capture reasons, I can safely just set them to `display: contents` and only worry about styling one div.

src/table/resizer/index.tsx
```diff
@@ -183,84 +223,127 @@ export function Resizer({
...
-        className={clsx(
-          styles.resizer,
-          (resizerHasFocus || showFocusRing || isKeyboardDragging) && styles['has-focus'],
-          isVisualRefresh && styles['is-visual-refresh']
-        )}
-        onMouseDown={event => {
-          if (event.button !== 0) {
-            return;
-          }
-          event.preventDefault();
```
avinashbot, line 208: Not sure why this behavior was there before, but clicking on the drag handle doesn't move focus to it. But now it does, like any regular button. Helps with simplifying the UAP logic, and is also nicer because you can switch between drag and keyboard.

src/table/__tests__/resizable-columns.test.tsx
```diff
@@ -219,43 +224,34 @@ test('takes width as min width if it is less than 120px and min width is not set
...
+  firePointermove(250);
   expect(wrapper.findColumnHeaders()[0].getElement()).toHaveStyle({ width: '250px' });
-  fireMouseMove(200);
+  firePointermove(200);
   expect(wrapper.findColumnHeaders()[0].getElement()).toHaveStyle({ width: '200px' });
-  fireMouseup(200);
-  expect(wrapper.findColumnHeaders()[0].getElement()).toHaveStyle({ width: '200px' });
-});
-
-test('should allow column resize with missing mousemove event', () => {
```
gethinwebster, line 243: do you have any ideas why this test existed? seems a strange one, but slightly worrying...

src/table/__tests__/resizable-columns.test.tsx
```diff
@@ -500,29 +496,157 @@ describe('resize in rtl', () => {
...
+    wrapper.findColumnResizer(1)!.focus();
+    expect(findDragHandle().findVisibleDirectionButtonInlineStart()).toBeTruthy();
+    expect(findDragHandle().findVisibleDirectionButtonInlineEnd()).toBeTruthy();
+
+    wrapper.findColumnResizer(1)!.keydown(KeyCode.escape);
+    expect(findDragHandle().findVisibleDirectionButtonInlineStart()).toBeNull();
+    expect(findDragHandle().findVisibleDirectionButtonInlineEnd()).toBeNull();
+  });
+
+  test('hides UAP buttons when arrow keys are pressed before the resizer button is activated', () => {
```
gethinwebster, line 619: This is different to the behavior in other components, e.g. app layout panels, sortable list. Is there a strong reason why it needs to be different here?

src/table/resizer/index.tsx
```diff
@@ -111,7 +202,9 @@ export function Resizer({
       elements.scrollParent.scrollLeft += AUTO_GROW_INCREMENT * (getIsRtl(elements.scrollParent) ? -1 : 1);
     };
 
-    const onMouseMove = (event: MouseEvent) => {
+    const onPointerMove = (event: PointerEvent) => {
+      // TODO: Only set it to true after a certain number of pixels travelled?
```
gethinwebster: suggest removing the TODO if we're happy with this behavior

src/table/__tests__/resizable-columns.test.tsx
johannes-weber: From `zb3AAbRNaDaY`:
> This is intentional (sort of). The drag handle has a button "mode" and a slider "mode". Keyboard users use the slider and mouse users use the button, so switching between modes causes issues.

> The resizer isn't designed to mix mouse/keyboard interactions and trying to make it work this way breaks it. It's existing behavior, so I'll avoid overcomplicating this for now.

src/table/resizer/index.tsx
johannes-weber: Removed. Having no tolerance and start immediately the resize action without is fine. Let's see if we get any customer feedback so we could increase if needed.

src/table/__tests__/resizable-columns.test.tsx
gethinwebster: I feel like that's in some way related to this: https://github.com/cloudscape-design/components/pull/3892/files#r2378371044

If mouse interactions are and remain separate, I'm not sure that changing the focus behavior there makes sense

src/table/__tests__/resizable-columns.test.tsx
johannes-weber: I have no idea and could not find any evidence for when this test was added. I can't think about a use-case where having a `pointerdown` on coordinate X/Y followed by a `pointerup` on X+a/Y+b without a `mousmove` event in between.

avinashbot: I have no idea either, there's no clear context I could find behind it. My best guess is as good as Johannes's. Perhaps some accessibility thing where you can trigger pointerdown somewhere and pointerup somewhere else? Either way, this PR specifically addresses pointer-related accessibility issues, so... I feel there's no need to be overly "defensive" about preserving this behavior.

src/table/__tests__/resizable-columns.test.tsx
avinashbot: The directional buttons serve two purposes:
1. (Main reason) For pointer-only users, to resize the column without dragging
2. (Adjacent) For keyboard users, to give them a visual hint that the focused item can be adjusted, and that they can use arrow keys at this point.

The trouble is that here, once you tap on it, the buttons show up, but you can't resize the columns by pressing the arrow keys. So I added this behavior to hide the buttons when keys are pressed so the buttons don't stick around and confuse you if you're a keyboard user. The two different "modes" of interaction are still separate for now.

PR #3894, opened 2025-09-25
chore: Mock getBoundingClientRect in mixed chart tests

jperals:
### Description

This is necessary for https://github.com/cloudscape-design/component-toolkit/pull/165.

JSDOM does not have access to actual element dimensions and returns 0 for everything. For this reason  in the mixed chart tests we are mocking  `window.getComputedStyle` to return 2px for border width among other things. But we are not mocking `HTMLElement.prototype.getBoundingClientRect`, so these tests still assume that all values are 0 for the values that this function returns. Then the changes in https://github.com/cloudscape-design/component-toolkit/pull/165 causes the calculated dimensions of the chart to be negative, because a non-zero border coming from `window.getComputedStyle` is subtracted from the zero width of the chart coming from `getBoundingClientRect` [here](https://github.com/cloudscape-design/component-toolkit/blob/fbeb3924d21428257f74e47df5fb0751a113595c/src/internal/container-queries/use-resize-observer.ts#L44) (which in reality won't happen, because these calculated dimensions should include the border). And these negative dimension values causes the chart not to render expected inner elements: https://github.com/cloudscape-design/component-toolkit/actions/runs/17982007963/job/51205514110?pr=165

The solution is to add one more mock for component-toolkit's `useContainerQuery`, which is used by the chart to determine its width.

### How has this been tested?

Ran the test with https://github.com/cloudscape-design/component-toolkit/pull/165 applied to the local copy of component-toolkit. The tests only passed after the changes in this PR.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/container-queries.page.tsx
+++ b/pages/container-queries.page.tsx
@@ -21,7 +21,11 @@ function MeasureReporter(props: { id?: string; style?: React.CSSProperties; type
     [props.type]
   );
   if (value === null) {
-    return <div ref={ref}>Loading...</div>;
+    return (
+      <div ref={ref} style={{ ...boxStyles, ...props.style }}>
+        Loading...
+      </div>
+    );
   }
   return (
     <div id={props.id} ref={ref} style={{ ...boxStyles, ...props.style }}>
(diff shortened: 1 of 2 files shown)


pages/container-queries.page.tsx
```diff
@@ -21,7 +21,11 @@ function MeasureReporter(props: { id?: string; style?: React.CSSProperties; type
     [props.type]
   );
   if (value === null) {
-    return <div ref={ref}>Loading...</div>;
+    return (
+      <div ref={ref} style={{ ...boxStyles, ...props.style }}>
```
jperals, line 25: Use the same styles already in the initial render, because https://github.com/cloudscape-design/component-toolkit/pull/165 applies to the initial render only.

src/mixed-line-bar-chart/__tests__/mixed-chart.test.tsx
```diff
@@ -124,14 +125,20 @@ const fakeGetComputedStyle: Window['getComputedStyle'] = (...args) => {
 
 beforeEach(() => {
   window.CSS.supports = () => true;
-  originalGetComputedStyle = window.getComputedStyle;
   window.getComputedStyle = fakeGetComputedStyle;
+  HTMLElement.prototype.getBoundingClientRect = function () {
+    const rect = originalBoundingClientRect.apply(this);
+    rect.width = 900;
```
jperals: Smaller widths would cause less X ticks to appear in some of the tests

PR #3897, opened 2025-09-26
chore: Fixes some integ tests for React 18

pan-kot:
### Description

A follow-up for: https://github.com/cloudscape-design/components/pull/3888

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/anchor-navigation/__integ__/anchor-navigation.test.ts
+++ b/src/anchor-navigation/__integ__/anchor-navigation.test.ts
@@ -20,6 +20,7 @@ describe('AnchorNavigation', () => {
     return useBrowser(async browser => {
       const page = new AnchorNavigationPage(browser);
       await browser.url('#/light/anchor-navigation/basic');
+      await page.waitForVisible(wrapper.toSelector());
       await testFn(page);
     });
   }
@@ -65,11 +66,6 @@ describe('AnchorNavigation', () => {
   test(
     'scrolling to the end of the page makes the last anchor item active',
     setupTest(async page => {
-      // TODO: fix test for React 18
-      if ((await page.getReactVersion()) === '18') {
-        return;
-      }
-
       const lastAnchorLink = wrapper.findAnchorLinkByHref('#section-1-2-1-1');
       await page.windowScrollTo({ top: 99999 }); // Very high value to ensure we are scrolled to the end
       return expect(await page.getElementAttribute(lastAnchorLink.toSelector(), 'aria-current')).toBe('true');
--- a/src/app-layout/__integ__/app-layout-focus-delegation.test.ts
+++ b/src/app-layout/__integ__/app-layout-focus-delegation.test.ts
@@ -48,10 +48,6 @@ describe.each(['classic', 'refresh', 'refresh-toolbar'] as const)('%s', theme =>
         'should not focus panels on page load',
         setupTest(
           async page => {
-            // TODO: fix test for React 18
-            if ((await page.getReactVersion()) === '18') {
-              return;
-            }
             await expect(page.isFocused('body')).resolves.toBe(true);
           },
           { pageName: 'with-split-panel', theme, mobile }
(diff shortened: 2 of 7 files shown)


src/anchor-navigation/__integ__/anchor-navigation.test.ts
```diff
@@ -20,6 +20,7 @@ describe('AnchorNavigation', () => {
     return useBrowser(async browser => {
       const page = new AnchorNavigationPage(browser);
       await browser.url('#/light/anchor-navigation/basic');
+      await page.waitForVisible(wrapper.toSelector());
```
pan-kot, line 23: That is what we do in most tests to ensure the page is fully loaded before performing any action. Without this line of the the integ tests fails consistently when run with React 18.

src/app-layout/__integ__/app-layout-focus-delegation.test.ts
```diff
@@ -48,10 +48,6 @@ describe.each(['classic', 'refresh', 'refresh-toolbar'] as const)('%s', theme =>
         'should not focus panels on page load',
         setupTest(
           async page => {
-            // TODO: fix test for React 18
-            if ((await page.getReactVersion()) === '18') {
-              return;
-            }
             await expect(page.isFocused('body')).resolves.toBe(true);
```
pan-kot, line 51: In all versions of mobile app layout the side navigation hides upon page render so that the user can see the page. In that case the navigation toggle button must not retain focus.

That does not work when run with React 18 in strict mode because the behaviour depends on multiple useEffect behaviours. This is addressed by introducing explicit `autoFocus` flag for the internal use-focus-control utility. The flag is set to `true` by default, which means that the focus should be dispatched to the toggle button on the next render cycle. It is set to `false` for the cases where we close the panel programmatically on the first render (for mobile experiences).

src/app-layout/utils/use-focus-control.ts
```diff
@@ -132,10 +132,12 @@ export function useFocusControl(
     shouldFocus.current = false;
   };
 
-  const setFocus = (force?: boolean) => {
-    shouldFocus.current = true;
+  const setFocus = ({ force, autoFocus = true }: { force: boolean; autoFocus?: boolean } = { force: false }) => {
     if (force && isOpen) {
+      shouldFocus.current = true;
```
pan-kot, line 137: The doFocus() function for classic and visual-refresh app layout requires shouldFocus ref to be set to true in order to function.

src/app-layout/visual-refresh-toolbar/state/use-app-layout.tsx
```diff
@@ -377,7 +377,7 @@ export const useAppLayout = (
   useEffect(() => {
     // Close navigation drawer on mobile so that the main content is visible
     if (isMobile) {
-      onNavigationToggle(false);
+      onNavigationToggle({ isOpen: false, autoFocus: false });
```
pan-kot, line 380: Here and in the effect below we hide the navigation programmatically to optimise UX. In that case the toggle should not get focused. The same is done for visual-refresh and toolbar versions of app layout. The autoFocus is set to true for all other cases (when the panel is opened with a user click or programmatically  from the consumers code).

src/property-filter/__integ__/async-loading.test.ts
pan-kot, line 144: This call is not really supposed to happen, but it does so in React 16 and not in React 18.

src/app-layout/classic.tsx
```diff
@@ -148,14 +148,14 @@ const ClassicAppLayout = React.forwardRef(
         node => node.tagName === 'A' && !!(node as HTMLAnchorElement).href
       );
       if (hasLink) {
-        onNavigationToggle(false);
+        onNavigationToggle({ isOpen: false, autoFocus: true });
```
georgylobko, line 151: [non-blocking] Since the `autoFocus` is already set to `true` be default, do we need to set it explicitly? It seems to be opt-out behavior.

pan-kot: It is, but I wanted it to be required so that the decision is better explicit.

PR #3922, opened 2025-10-09
feat: Bottom drawer (v2)

georgylobko:
### Description

This PR introduces a new panel within existing `AppLayout` component. It does not affect any publicly available api. The only way to register it is using the existing plugin api `registerDrawer` with an additional parameter `position: "bottom"`.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/plugins/widget/index.ts
+++ b/src/internal/plugins/widget/index.ts
@@ -5,7 +5,7 @@ import { getAppLayoutMessageHandler, pushInitialMessage } from './core';
 import { AppLayoutUpdateMessage, DrawerPayload, RegisterDrawerMessage } from './interfaces';
 
 /**
- * Registers a new ai runtime drawer to app layout
+ * Registers a new left runtime drawer to app layout
  * @param drawer
  */
 export function registerLeftDrawer(drawer: DrawerPayload) {


pages/app-layout/utils/external-widget.tsx
```diff
@@ -294,3 +294,67 @@ awsuiPlugins.appLayout.registerDrawer({
   },
   unmountContent: container => unmount(container),
 });
+
+export const registerBottomDrawer = () => {
+  awsuiPlugins.appLayout.registerDrawer({
```
just-boris: Can we from now on use the new `awsuiWidgetsPlugins` for all new runtime API?

src/app-layout/visual-refresh-toolbar/drawer/global-bottom-drawer.tsx
```diff
@@ -0,0 +1,310 @@
...
+                        default:
+                          activeDrawer?.onHeaderActionClick?.(event);
+                      }
+                    }}
+                    ariaLabel="Global panel actions"
+                    items={drawerActions}
+                    __internalRootRef={(root: HTMLElement) => {
+                      if (!root) {
+                        return;
+                      }
```
just-boris, line 284: Do we not need to clean up the refs when unmounting?

src/app-layout/visual-refresh-toolbar/state/props-merger.ts
```diff
@@ -33,9 +33,12 @@ export const mergeProps: MergeProps = (ownProps, additionalProps) => {
     }
     if (props.globalDrawers && !checkAlreadyExists(!!toolbar.globalDrawers, 'globalDrawers')) {
       toolbar.globalDrawersFocusControl = props.globalDrawersFocusControl;
+      toolbar.bottomDrawersFocusRef = props.bottomDrawersFocusRef;
```
just-boris: Why does it go into `props.globalDrawers`  condition instead of its own? What will happen if no global drawers registered?

src/app-layout/visual-refresh-toolbar/interfaces.ts
```diff
@@ -52,6 +53,7 @@ export interface AppLayoutInternals {
   maxGlobalDrawersSizes: Record<string, number>;
   drawers: ReadonlyArray<InternalDrawer>;
   drawersFocusControl: FocusControlState;
+  bottomDrawersFocusControl: FocusControlState;
```
just-boris: `AppLayoutInternals` is the legacy widget interface which we will remove once all consumers update.

All new features should live `AppLayoutWidgetizedState`

src/app-layout/__tests__/__snapshots__/widget-contract-old.test.tsx.snap
```diff
@@ -9,6 +9,8 @@ Map {
       "activeAiDrawerSize": 0,
       "activeDrawer": undefined,
       "activeDrawerSize": 290,
+      "activeGlobalBottomDrawerId": null,
```
just-boris: Why is there a diff in the snapshot? Will it cause any impact in the widget?

src/app-layout/visual-refresh-toolbar/drawer/global-bottom-drawer.tsx
georgylobko: No, because `refs.close` refers to the object link, not its value

pages/app-layout/utils/external-widget.tsx
georgylobko: Moved to the new runtime api

src/app-layout/visual-refresh-toolbar/state/props-merger.ts
georgylobko: Made it go into its own array

src/app-layout/visual-refresh-toolbar/interfaces.ts
georgylobko: addressed

src/app-layout/__tests__/__snapshots__/widget-contract-old.test.tsx.snap
georgylobko: addressed

src/app-layout/visual-refresh-toolbar/state/use-app-layout.tsx
```diff
@@ -344,12 +453,16 @@ export const useAppLayout = (
 
   const splitPanelInternals: SplitPanelProviderProps = {
     bottomOffset: 0,
-    getMaxHeight: useStableCallback(() => {
+    getMaxHeight: useCallback(() => {
```
just-boris, line 456: Why did this change? The idea of `useStableCallback` is to avoid extra re-renders when a dependency changes, because the function is called lazily

src/app-layout/visual-refresh-toolbar/drawer/global-bottom-drawer.tsx
```diff
@@ -0,0 +1,309 @@
...
+import styles from './styles.css.js';
+
+export function AppLayoutBottomDrawerWrapper({ widgetizedState }: { widgetizedState: AppLayoutWidgetizedState }) {
+  const { activeGlobalBottomDrawerId, bottomDrawers } = widgetizedState;
+  const openBottomDrawersHistory = useRef<Set<string>>(new Set());
+  useEffect(() => {
+    if (activeGlobalBottomDrawerId) {
+      openBottomDrawersHistory.current.add(activeGlobalBottomDrawerId);
+    }
+  }, [activeGlobalBottomDrawerId]);
```
just-boris, line 29: Why is it happening inside useEffect? The `openDrawersHistory` with a similar use-case does not have an effect: https://github.com/cloudscape-design/components/blob/f724c4ec701b5e750ad74d19269c98f284019f95/src/app-layout/visual-refresh-toolbar/drawer/global-drawers.tsx#L31

1. If you fixed a bug, how about propagating the fix to the other place?
2. If it is not needed, then let's remove

src/app-layout/visual-refresh-toolbar/state/use-app-layout.tsx
```diff
@@ -159,9 +166,90 @@ export const useAppLayout = (
...
+      aiDrawerMessageHandler(message);
+      return;
+    }
+
+    if (checkBottomDrawerIdExists(id) || message.type === 'registerBottomDrawer') {
+      bottomDrawersMessageHandler(message);
+      return;
+    }
+
+    metrics.sendOpsMetricObject('awsui-widget-drawer-incorrect-id', {
```
just-boris, line 245: Did you decide to have these two metrics (line 245 and line 203) to have the same name?

I still recommend giving them a different name to simplify reading these metrics in the future

src/internal/plugins/widget/interfaces.ts
```diff
@@ -73,6 +75,14 @@ export type AppLayoutUpdateMessage =
...
   | ExitExpandedModeMessage;
 
-export type InitialMessage = RegisterDrawerMessage;
+export interface EmitNotificationPayload {
+  type: FlashbarProps.Type;
+  header: string;
+  content: string;
+}
+
+export type EmitNotificationMessage = Message<'emitNotification', EmitNotificationPayload>;
```
just-boris: How did this end up here?

src/internal/plugins/widget/index.ts
```diff
@@ -5,7 +5,7 @@ import { getAppLayoutMessageHandler, pushInitialMessage } from './core';
 import { AppLayoutUpdateMessage, DrawerPayload, RegisterDrawerMessage } from './interfaces';
 
 /**
- * Registers a new runtime drawer to app layout
+ * Registers a new ai runtime drawer to app layout
```
just-boris: Is this change really needed?

src/app-layout/visual-refresh-toolbar/drawer/local-drawer.tsx
```diff
@@ -19,9 +19,13 @@ import styles from './styles.css.js';
 
 interface AppLayoutDrawerImplementationProps {
   appLayoutInternals: AppLayoutInternals;
+  widgetizedState: AppLayoutWidgetizedState;
```
just-boris: For backward compatibility, you need to use `Partial<AppLayoutWidgetizedState> | undefined` type here

src/app-layout/__tests__/__snapshots__/widget-contract-old.test.tsx.snap
```diff
@@ -347,18 +347,19 @@ Map {
         "toolbar": 0,
       },
     },
-  },
-  "AppLayoutSplitPanelDrawerBottomImplementation" => {
-    "appLayoutInternals": {
+    "widgetizedState": {
```
just-boris: Changes in the snapshot tests means possible issues in production for existing consumers.

You need the state only for this line:

```
activeGlobalBottomDrawerId ? bottomDrawerReportedSize : 0
```

So I recommend propagating this value only, so it will be one-line change in this snapshot, which feels much safer

src/app-layout/visual-refresh-toolbar/state/use-app-layout.tsx
georgylobko: This change is needed because now the split panel's max height depends on the active bottom drawer. If we keep using `useStableCallback`, the split panel get wrong calculations

src/app-layout/visual-refresh-toolbar/drawer/global-bottom-drawer.tsx
georgylobko: Yes, this change fixes a bug with animations. I render all bottom drawers and they are always stay in the DOM tree, but without `activeDrawer` prop. It allows to animate their open/close behavior consistently. And yes, I'll propagate the same bug fix to global drawers in a different PR

src/app-layout/visual-refresh-toolbar/state/use-app-layout.tsx
georgylobko: yes, I did that because the metric gets emitted whenever the is is incorrect. to distinguish the events on lines 203 and 245, you can check the type field for the message type

src/internal/plugins/widget/interfaces.ts
georgylobko: Good finding. Probably I wrongly copied this line when pulling the widgetized plugin refactoring to my branch. Addressed

src/internal/plugins/widget/index.ts
georgylobko: Yes, to distinguish between ai and bottom drawer

src/app-layout/__tests__/__snapshots__/widget-contract-old.test.tsx.snap
georgylobko: addressed

src/app-layout/visual-refresh-toolbar/drawer/global-bottom-drawer.tsx
just-boris: Ok, I will resolve this thread for now, but please make a follow up

src/app-layout/visual-refresh-toolbar/state/use-app-layout.tsx
just-boris: This smells like a beginning of the problem, but up to you if you are comfortable with that risk or not

src/internal/plugins/widget/index.ts
just-boris: Let's call it "left" in the comment too. 

Not for this PR, but in general, we should look for removing "ai" from the naming, because it is not relevant. "Global" drawers may contain "AI" stuff too

src/app-layout/__tests__/__snapshots__/widget-contract-old.test.tsx.snap
just-boris: ⛳️ There is still a diff in the snapshot even after the change is applied. It seems like more cross-boundary changes need to be fixed

georgylobko: addressed

src/internal/plugins/widget/index.ts
georgylobko: addressed

PR #3937, opened 2025-10-15
fix: AI drawer programmatical resize

georgylobko:
### Description

Fixed an issue where the AI panel could grow beyond its maximum width when resized programmatically.

Previously, calling this method would cause the AI panel to exceed its max width and break the page layout:
```javascript
updateDrawer({ type: 'resizeDrawer', payload: { id: 'ai-panel', size: 10000 } });
```

The panel now respects its width constraints.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/visual-refresh-toolbar/state/use-app-layout.tsx
+++ b/src/app-layout/visual-refresh-toolbar/state/use-app-layout.tsx
@@ -153,8 +153,8 @@ export const useAppLayout = (
     activeAiDrawerSize,
     minAiDrawerSize,
     onActiveAiDrawerResize,
-    setMaxAiDrawerSize,
   } = useAiDrawer({
+    getMaxAiDrawerSize: () => maxAiDrawerSize,
     onAiDrawerFocus: () => aiDrawerFocusControl.setFocus(),
     expandedDrawerId,
     setExpandedDrawerId,
@@ -271,10 +271,6 @@ export const useAppLayout = (
     activeGlobalDrawersSizes,
     activeAiDrawerSize,
   });
-
-  useEffect(() => {
-    setMaxAiDrawerSize(maxAiDrawerSize);
-  }, [setMaxAiDrawerSize, maxAiDrawerSize]);
 
   const verticalOffsets = computeVerticalLayout({
     topOffset: placement.insetBlockStart,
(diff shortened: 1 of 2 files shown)


gethinwebster (approved):

src/app-layout/__integ__/ai-drawer.test.ts
```diff
@@ -0,0 +1,60 @@
...
+};
+const findDrawerTriggerById = (wrapper: AppLayoutWrapper, id: string) => {
+  return wrapper.find(`[data-testid="awsui-app-layout-trigger-${id}"]`);
+};
+
+describe('Visual refresh toolbar only', () => {
+  class PageObject extends BasePageObject {
+    getClientWidth() {
+      return this.browser.execute(() => document.body.clientWidth);
+    }
```
just-boris: Can we use the existing `getViewportSize`

https://github.com/cloudscape-design/browser-test-tools/blob/642e5b95d8d8c4817b90e55aeed59af7a92e21a8/src/page-objects/base.ts#L110

pages/app-layout/utils/external-global-left-panel-widget.tsx
```diff
@@ -28,6 +28,14 @@ const AIDrawer = () => {
       >
         exit expanded mode
       </Button>
+      <Button
+        className="resize-to-max-width"
+        onClick={() => {
+          updateDrawer({ type: 'resizeDrawer', payload: { id: 'ai-panel', size: 10000 } });
```
just-boris: To reproduce the issue, we need to have a size larger than the widow width, so `size: window.innerWidth + 300` would be more clear what we try to do here

src/app-layout/visual-refresh-toolbar/state/use-app-layout.tsx
```diff
@@ -157,6 +158,7 @@ export const useAppLayout = (
     onAiDrawerFocus: () => aiDrawerFocusControl.setFocus(),
     expandedDrawerId,
     setExpandedDrawerId,
+    maxAiDrawerSize: maxAiDrawerSizeRef.current,
```
just-boris: This is bad. It does not give you the current actual max size, but a value from the past render. The issue you are trying to solve is still prone to re-ocurring with some tricky conditions.

Not going to block this PR if you want to get this out, but highly recommend to revisit this solution to always read the real max size value without using ref

src/app-layout/__integ__/ai-drawer.test.ts
georgylobko: yes. addressed

pages/app-layout/utils/external-global-left-panel-widget.tsx
georgylobko: addressed

src/app-layout/visual-refresh-toolbar/state/use-app-layout.tsx
georgylobko: Refactored

PR #3947, opened 2025-10-20
chore: Fixes some integ tests for react 18

pan-kot:
### Description

The PR fixes some of the tests that were muted for React 18.

### How has this been tested?

* The flaky tests have been run over 100 times in the dry-run to ensure stability.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/table/expandable-rows-test.page.tsx
+++ b/pages/table/expandable-rows-test.page.tsx
@@ -208,6 +208,7 @@ export default () => {
 const SERVER_DELAY = 1500;
 const ROOT_PAGE_SIZE = 10;
 const NESTED_PAGE_SIZE = 2;
+
 function useTableData() {
   const settings = usePageSettings();
   const delay = settings.useServerMock ? SERVER_DELAY : 0;
@@ -223,24 +224,13 @@ function useTableData() {
     },
     [delay, settings.manualServerMock]
   );
+  const [response, setResponse] = useState<{ loading: boolean; error: boolean; instances: readonly Instance[] }>({
+    loading: settings.useServerMock,
+    error: false,
+    instances: settings.useServerMock ? [] : allInstances,
+  });
 
-  const [loading, setLoading] = useState(false);
-  const [error, setError] = useState(false);
-
-  // Imitate server-side delay when fetching items for the first time.
-  const [readyInstances, setReadyInstances] = useState(settings.useServerMock ? [] : allInstances);
-  useEffect(() => {
-    setLoading(true);
-    setError(false);
-    return getServerResponse(() => {
-      setReadyInstances(allInstances);
-      setLoading(false);
-      setError(settings.emulateServerError);
-    });
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [getServerResponse, setLoading, setError, setReadyInstances]);
-
-  const collectionResult = useCollection(readyInstances, {
+  const collectionResult = useCollection(allInstances, {
     pagination: settings.usePagination ? { pageSize: ROOT_PAGE_SIZE } : undefined,
     sorting: {},
     filtering: {},
@@ -267,22 +257,17 @@ function useTableData() {
       : undefined,
   });
 
-  // Imitate server-side delay when items update.
+  // Imitate server-side delay when fetching items.
   const memoItems = useEqualsMemo(collectionResult.items);
   const [delayedItems, setDelayedItems] = useState(memoItems);
   const readyItems = settings.manualServerMock ? memoItems : delayedItems;
   useEffect(() => {
-    if (!settings.manualServerMock) {
-      setLoading(true);
-      setError(false);
-      return getServerResponse(() => {
-        setLoading(false);
-        setDelayedItems(memoItems);
-        setError(settings.emulateServerError);
-      });
-    }
+    setResponse(prev => ({ ...prev, loading: true, error: false }));
+    return getServerResponse(() => {
+      setResponse(() => ({ loading: false, error: settings.emulateServerError, instances: memoItems }));
+    });
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [getServerResponse, memoItems, setLoading, setError, setDelayedItems]);
+  }, [getServerResponse, memoItems, setDelayedItems]);
 
   // Decorate path options to only show the last node and not the full path.
   collectionResult.propertyFilterProps.filteringOptions = collectionResult.propertyFilterProps.filteringOptions.map(
@@ -361,9 +346,9 @@ function useTableData() {
 
   return {
     ...collectionResult,
-    error: settings.useServerMock ? error : false,
-    loading: settings.useServerMock ? loading : false,
-    items: settings.useServerMock && error ? [] : paginatedItems,
+    error: settings.useServerMock ? response.error : false,
+    loading: settings.useServerMock ? response.loading : false,
+    items: settings.useServerMock && response.error ? [] : paginatedItems,
     actions: {
       clearSelection: () => {
         collectionResult.actions.setSelectedItems([]);


pages/table/expandable-rows-test.page.tsx
```diff
@@ -269,15 +269,18 @@ function useTableData() {
 
   // Imitate server-side delay when items update.
   const memoItems = useEqualsMemo(collectionResult.items);
-  const [readyItems, setReadyItems] = useState(memoItems);
+  const [delayedItems, setReadyItems] = useState(memoItems);
+  const readyItems = settings.manualServerMock ? memoItems : delayedItems;
   useEffect(() => {
```
pan-kot, line 264: This useEffect is firing an extra time during integ tests when run with React 18 + strict-mode, thus enforcing the loading state after the items are already resolved. The updated code avoids that effect as unnecessary as its only purpose is to imitate a server-side delay for manual tests.

src/table/__integ__/expandable-rows.test.ts
```diff
@@ -45,7 +45,7 @@ describe('Expandable rows', () => {
     })
   );
 
-  (process.env.REACT_VERSION !== '18' ? test : test.skip)(
+  test(
```
pan-kot, line 48: This test is now working thanks to the changes to the test page.

src/table/__integ__/performance-marks.test.ts
```diff
@@ -13,15 +13,19 @@ function setupTest(
...
-    }
     const page = new BasePageObject(browser);
     await browser.url(`#/light/table/performance-marks${!inViewport ? '?outsideOfViewport=true' : ''}`);
     const getMarks = async () => {
-      await new Promise(r => setTimeout(r, 200));
-      const marks = await browser.execute(() => performance.getEntriesByType('mark') as PerformanceMark[]);
-      return marks.filter(m => m.detail?.source === 'awsui');
+      let awsuiMarks: PerformanceMark[] = [];
+      await page.waitForAssertion(async () => {
+        await new Promise(r => setTimeout(r, 200));
```
pan-kot: This timeout is likely causing issues. Increasing it three-fold reduces flakiness, but the waitForAssertion seems to do a better job.

src/table/__integ__/resizable-columns.test.ts
```diff
@@ -331,7 +331,7 @@ test(
   })
 );
 
-(process.env.REACT_VERSION !== '18' ? test : test.skip)(
+test(
```
pan-kot, line 334: This test was flaky, but I wasn't able to reproduce the failures when running it 100 times in a loop. I tried running it with visualRefresh=true and visualRefresh=false.

src/table/__integ__/table.test.ts
```diff
@@ -84,6 +84,6 @@ test(
     ariaLabelledby = await page.getElementAttribute(scrollableWrapperSelector, 'aria-labelledby');
     expect(role).toEqual('region');
     expect(ariaLabelledby).not.toBeFalsy();
-    await expect(page.getElementsText(`#${ariaLabelledby}`)).resolves.toEqual([tableHeading]);
+    await expect(page.getElementsText(`#${CSS.escape(ariaLabelledby)}`)).resolves.toEqual([tableHeading]);
```
pan-kot, line 87: When React 18+ is used, we use the React's useId() function to generate unique IDs (used for ARIA labelled-by). However, that function produces ids with colons (":") that cause an invalid selector error unless escaped.

src/table/__integ__/performance-marks.test.ts
jperals: Do we need the timeout at all if we wrap the assertion in `waitForAssertion`?

src/table/__integ__/performance-marks.test.ts
```diff
@@ -13,15 +13,19 @@ function setupTest(
...
+        const marks = await browser.execute(() => performance.getEntriesByType('mark') as PerformanceMark[]);
+        awsuiMarks = marks.filter(m => m.detail?.source === 'awsui');
+        if (awsuiMarks.length === 0) {
```
jperals: Minor: why not use `expect` instead?

src/table/__integ__/table.test.ts
jperals: But if the element with the `aria-labelledby` attribute uses the ID without escaping, do they still match?

pages/table/expandable-rows-test.page.tsx
```diff
@@ -269,15 +269,18 @@ function useTableData() {
 
   // Imitate server-side delay when items update.
   const memoItems = useEqualsMemo(collectionResult.items);
-  const [readyItems, setReadyItems] = useState(memoItems);
+  const [delayedItems, setReadyItems] = useState(memoItems);
```
jperals: Is it intentional that the name of these two variables don't match?

pages/table/expandable-rows-test.page.tsx
jperals: Would it help to merge variables, e.g, `loading` might not be necessary if it can be inferred from the presence of the ready items, and error and items could be merged in one single object?

pages/table/expandable-rows-test.page.tsx
pan-kot: No - will update 👍

pages/table/expandable-rows-test.page.tsx
pan-kot: I din't quite understand that, could you please share a code example?

src/table/__integ__/performance-marks.test.ts
pan-kot: Good point - I will experiment with that

pan-kot: I will move the timeout to the outside of the block, as the initial delay is expected

src/table/__integ__/performance-marks.test.ts
pan-kot: Good point - will updated 👍

src/table/__integ__/table.test.ts
pan-kot: Yes, the matching works - only querying does not

pages/table/expandable-rows-test.page.tsx
jperals: For example:

```
const [response, setResponse] = useState({});
if (!settings.manualServerMock) {
  setResponse({ isLoading: true, isError: false });
  return getServerResponse(() => {
    setResponse({
      isLoading: false,
      isError: settings.emulateServerError,
      items: memoItems
    });
  });
}
```

in case setting three state variables independently was causing some problem.

pages/table/expandable-rows-test.page.tsx
```diff
@@ -269,17 +269,20 @@ function useTableData() {
 
   // Imitate server-side delay when items update.
   const memoItems = useEqualsMemo(collectionResult.items);
-  const [readyItems, setReadyItems] = useState(memoItems);
+  const [delayedItems, setDelayedItems] = useState(memoItems);
+  const readyItems = settings.manualServerMock ? memoItems : delayedItems;
```
jperals, line 263: What is the difference between `settings.manualServerMock` and `settings.useServerMock`?

In the dev page, there only seems to be one related option ("Use server mock")

pan-kot: The server mock means using artificial delays to represent loading state and also random errors. The manual mock is different as instead of delays it waits fro signal to either resolve or reject the next request - this is used in the related integ tests.

pages/table/expandable-rows-test.page.tsx
pan-kot: Thanks - I refactored this now!

jperals (approved):

PR #3952, opened 2025-10-21
chore: Add reference tokens and replace legacy color tokens in style dictionary

jkuelz:
### Description

This change introduces reference tokens for primary, neutral, and status colors into our design token system. This is one of a series of updates to implement better out of the box global theming. We are keeping the legacy tokens for now so that we don't break our internal package builds which still rely on them in the meantime. Because we are keeping both tokens, this update also adds an eslint rule to tell developers to use the new tokens if the old ones are referenced.

Related links, issue #, if available: [quip: ac9YAbqsd7e6], CR-229486129

### How has this been tested?

Ran through dev pipeline: AwsUi-v3-jkuelz

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/style-dictionary/classic/color-palette.ts
+++ b/style-dictionary/classic/color-palette.ts
@@ -77,6 +77,7 @@ const referenceTokens: StyleDictionary.ReferenceDictionary = {
   colorError600: tokens.colorRed600,
   colorError1000: tokens.colorRed1000,
   colorInfo50: tokens.colorBlue50,
+  colorInfo300: tokens.colorBlue300,
   colorInfo400: tokens.colorBlue400,
   colorInfo600: tokens.colorBlue600,
   colorInfo1000: tokens.colorBlue1000,


build-tools/eslint/index.js
```diff
@@ -3,4 +3,5 @@
 module.exports.rules = {
   'ban-files': require('./ban-files'),
   'prefer-live-region': require('./prefer-live-region'),
+  'no-legacy-tokens': require('./no-legacy-tokens'),
```
just-boris, line 6: Good idea to add a linter rule, but could you add it via a PR to another repository: https://github.com/cloudscape-design/build-tools/tree/main/src/eslint

And while adding this, it also needs some tests for positive and negative matches. There are some examples already

build-tools/eslint/no-legacy-tokens.js
```diff
@@ -0,0 +1,61 @@
...
+            matches.forEach(token => {
+              const colorFamily = token.match(/color(Grey|Blue|Green|Red|Yellow)/)?.[1];
+              const suggestion = REFERENCE_TOKEN_SUGGESTIONS[`color${colorFamily}`] || 'semantic reference tokens';
+
+              context.report({
+                node,
+                messageId: 'no-legacy-tokens',
+                data: { token, suggestion },
+              });
+            });
```
just-boris, line 40: This code is repeated two times and can be reused, right?

PR #3958, opened 2025-10-22
fix: Fixes popover position recomputation for charts

pan-kot:
### Description

The popover includes code that should recompute its position when there is a layout shift on the page, e.g. when a split-panel opens, causing the popover target to change. The keepPosition property is only used for charts popover, and for some reason the popovers with keepPosition=true were not updated. However, this causes the charts popover position to be off if the layout shift occurs.

See how legacy charts reacted to a split panel open before the change:

https://github.com/user-attachments/assets/e6da8e9c-1271-4102-b426-5bbfb8525443

With the change (removing keepPosition check), the popover position is correctly adjusted.

https://github.com/user-attachments/assets/aef2d3a4-b2ed-4e6b-8b15-ef49a4a3aaf1

The fix only works well when the layout change does not cause the chart to change size. If that happens, then the popover (even if pinned) is closed in area- and pie charts. In mixed charts (incl. bar- and line charts, too), the popover is not closed, but the track marker is not updated, causing both the marker and the popover to have incorrect position. That is an existing bug, reproducible with both old and new code, and it is not addressed by this change.

https://github.com/user-attachments/assets/1c404a7c-3d47-4cff-914b-dd63fcafa1bd

The fix also works in the new cartesian and pie charts. There, the positions of the tooltip and the marker are correctly recomputed.

https://github.com/user-attachments/assets/4ae1b78e-57f3-4d71-a534-82afe43aa662

Rel: AWSUI-61377

### How has this been tested?

* Manual tests using the new test pages for legacy and new charts
* Removed the now irrelevant integ test

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/popover/container.tsx
+++ b/src/popover/container.tsx
@@ -113,19 +113,12 @@ export default function PopoverContainer({
 
   // Recalculate position when the DOM changes.
   // istanbul ignore next - tested via integration tests
-  usePositionObserver(trackRef, trackKey, () => {
+  usePositionObserver(getTrack.current, trackKey, () => {
     // Do not update position if popover moved offscreen
     const popoverOffset = popoverRef.current && getLogicalBoundingClientRect(popoverRef.current);
-
-    if (
-      keepPosition ||
-      !popoverOffset ||
-      popoverOffset.insetBlockStart < 0 ||
-      popoverOffset.insetBlockEnd > window.innerHeight
-    ) {
+    if (!popoverOffset || popoverOffset.insetBlockStart < 0 || popoverOffset.insetBlockEnd > window.innerHeight) {
       return;
     }
-
     positionHandlerRef.current();
   });
 
--- a/pages/app/index.tsx
+++ b/pages/app/index.tsx
@@ -50,6 +50,7 @@ function isAppLayoutPage(pageId?: string) {
     'prompt-input/simple',
     'funnel-analytics/static-single-page-flow',
     'funnel-analytics/static-multi-page-flow',
+    'charts-with-side-panel',
   ];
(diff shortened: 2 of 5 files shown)


src/popover/container.tsx
```diff
@@ -113,19 +113,12 @@ export default function PopoverContainer({
 
   // Recalculate position when the DOM changes.
   // istanbul ignore next - tested via integration tests
-  usePositionObserver(trackRef, trackKey, () => {
+  usePositionObserver(getTrack.current, trackKey, () => {
```
pan-kot, line 116: getTrack.current represents one of the two supported ways to provide the track element, the second one is used by the charts components.

src/popover/container.tsx
```diff
@@ -113,19 +113,12 @@ export default function PopoverContainer({
...
     // Do not update position if popover moved offscreen
     const popoverOffset = popoverRef.current && getLogicalBoundingClientRect(popoverRef.current);
-
-    if (
-      keepPosition ||
```
pan-kot, line 121: We only use keepPosition=true in charts. The need of preventing the position update in this case is not clear though: when a layout shift occurs, the chart popover can be misplaced, too.

ClaudioGSDB: wouldn't it be better then to remove keepPosition entirely from the popover container from `src/internal/components/chart-popover/index.tsx` then? Since it seems its defaulting to that.
``` javascript
<PopoverContainer
  ...
  keepPosition={true}
  ...
>
```

pan-kot: The keepPosition is also used in the `src/popover/use-popover-position.ts`, so removing it might cause unexpected effects.

PR #3969, opened 2025-10-26
feat: implement style api for input component

mxschll:
### Description

Extends the input component with a style API for customization, allowing developers to override default component styles.

### How has this been tested?

- Unit tests: Style prop validation and CSS class application
- Integration tests: Component rendering with different style combinations
- Visual regression tests: Screenshots for different style combinations

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/input/styles.scss
```diff
@@ -53,14 +61,26 @@
 
   @include placeholder {
     @include styles.form-placeholder;
+    color: var(#{custom-props.$stylePlaceholderColor}, awsui.$color-text-input-placeholder);
```
DaemonCahill: Should these includes be parameterized as opposed to overwritten? Presumably if form-placeholder takes variables then all use cases across the codebase will automatically support styling...

mxschll: Done. I keep passing vars with the fallback value as a parameter, since null checks would be required in the mixins for every parameter otherwise.

src/input/interfaces.ts
```diff
@@ -202,7 +207,49 @@ export namespace InputProps {
...
+        default?: string;
+        disabled?: string;
+        focus?: string;
+        hover?: string;
+      };
+      fontSize?: string;
+      fontWeight?: string;
+      paddingBlock?: string;
+      paddingInline?: string;
+      placeholder?: {
```
gethinwebster: I think `placeholder` should be outside of `root`: it is a different selector to target

src/input/styles.scss
```diff
@@ -72,6 +95,13 @@
     box-shadow: none;
   }
 
+  &:hover:not(:focus):not(:disabled) {
+    border-color: var(#{custom-props.$styleBorderColorHover}, awsui.$color-border-input-default);
```
gethinwebster: should the default here instead be `$styleBorderColorDefault`? Otherwise, if only `styleBorderColorDefault` is supplied and no `styleBorderColorHover`, then on hover the old unstyled `color-border-input-default` would be applied, which I think is unexpected

mxschll: Should it still fallback to `awsui.$color-border-input-default` in case `$styleBorderColorDefault` is not provided? So the styling cascades through three levels: `$styleBorderColorHover` → `$styleBorderColorDefault` → `$color-border-input-default`?

gethinwebster: yeah, it feels to me like that should be the expectation, at least in this sort of case where we're adding styling via the `style` property that doesn't exist in our standard theme (by default, input has no hover state). I'm not sure if there are other components where we've added this additional level of granularity or not @DaemonCahill 

Or I guess the other question is: do we even need this hover state?

src/input/styles.scss
```diff
@@ -72,6 +109,22 @@
     box-shadow: none;
   }
 
+  &:hover:not(:focus):not(:disabled):not(:read-only) {
```
gethinwebster: I think there's a _slight_ risk that we're using a different selector for readonly here (`:read-only` vs `.input-readonly` earlier in the file), having consistency would probably be better

PR #3985, opened 2025-10-29
feat: add i18n support for itemRemovedAriaLive in attribute-editor

Harsh-Anand-Singh:
### Description

This PR provides the `itemRemovedAriaLive` string for AttributeEditor via the i18n provider, reducing friction for customers who previously had to manually provide this label.

The solution merges translated strings from the AWS-UI-Components-I18n package and provides the default announcement when no custom value is specified, while still allowing overrides through the `i18nStrings` prop. Additionally, the `i18nStrings` property documentation has been improved to clearly list all available localized strings with descriptions.

Related links, issue #, if available: AWSUI-61133

### How has this been tested?

I've tested the i18n integration:

- Modified the attribute-editor simple page locally to wrap the component with `I18nProvider` to enable i18n functionality
- Verified that the component without explicit `itemRemovedAriaLive` now receives the default i18n value
- Confirmed that explicit `i18nStrings.itemRemovedAriaLive` values continue to override the default
- Tested with a screen reader to verify the removal announcement is properly announced
- Verified the i18n integration works correctly by adding unit tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/build-tools/utils/custom-css-properties.js
+++ b/build-tools/utils/custom-css-properties.js
@@ -95,6 +95,21 @@ const customCssPropertiesList = [
   'styleBoxShadowDefault',
   'styleBoxShadowDisabled',
   'styleBoxShadowHover',
+  // Readonly state
+  'styleBackgroundReadonly',
+  'styleBorderColorReadonly',
+  'styleBoxShadowReadonly',
+  'styleColorReadonly',
+  // Focus state
+  'styleBackgroundFocus',
+  'styleBorderColorFocus',
+  'styleBoxShadowFocus',
+  'styleColorFocus',
+  // Placeholder style properties
+  'stylePlaceholderColor',
+  'stylePlaceholderFontSize',
+  'stylePlaceholderFontStyle',
+  'stylePlaceholderFontWeight',
   // Alert focus ring properties
   'alertFocusRingBorderColor',
   'alertFocusRingBorderRadius',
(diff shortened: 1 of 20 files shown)


src/attribute-editor/interfaces.ts
```diff
@@ -199,6 +210,12 @@ export interface AttributeEditorProps<T> extends BaseComponentProps {
 
   /**
    * An object containing all the necessary localized strings required by the component.
+   *
+   * * `errorIconAriaLabel` (string) - Provides a text alternative for the error icon in the error message.
```
georgylobko: Do we need to duplicate the strings description? 

The original reference to the table interface refers to the union type `Variant`, that does not have descriptions for each prop. 
https://github.com/cloudscape-design/components/blob/main/src/table/interfaces.tsx#L308-L321

But these strings are already described above. I'd suggest to remove it, because we would need to maintain the same string in both places.

src/attribute-editor/__tests__/attribute-editor.test.tsx
```diff
@@ -620,6 +620,24 @@ describe('Attribute Editor', () => {
       const wrapper = createWrapper(container).findAttributeEditor()!;
       expect(wrapper.findRow(1)!.findRemoveButton()!.getElement()).toHaveTextContent('Custom remove');
     });
+    test('supports providing itemRemovedAriaLive from i18n provider', async () => {
```
georgylobko, line 623: In this case it makes sense to add a test it with i18nStrings without the provider

src/attribute-editor/interfaces.ts
Harsh-Anand-Singh: Sure, did it.

src/attribute-editor/__tests__/attribute-editor.test.tsx
Harsh-Anand-Singh: sure.

PR #3987, opened 2025-10-30
fix: Ensure rapid successive live announcements are not lost

johannes-weber:
### Description

In some cases, when two live regions are used and updated nearly at the same time, one of the announcements sometimes get missed by screen readers (tested with VO on Mac).

With this change, multiple LiveRegion instances are automatically coordinated to prevent
 * screen readers from dropping announcements when they occur simultaneously.
 * Announcements are staggered with a minimum gap to ensure all are heard.

Related links, issue #, if available: `AWSUI-61345`

### How has this been tested?

- created a test page to manually test the changes
- add unit tests for this scenario

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/live-region/controller.ts
+++ b/src/live-region/controller.ts
@@ -10,6 +10,10 @@ import testUtilStyles from './test-classes/styles.css.js';
  * that a message is announced again even if it matches the previous content
  * of the live region.
  *
+ * Multiple LiveRegion instances are automatically coordinated to prevent
+ * screen readers from dropping announcements when they occur simultaneously.
+ * Announcements are staggered with a minimum gap to ensure all are heard.
+ *
  * @see https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions
  */
 export class LiveRegionController {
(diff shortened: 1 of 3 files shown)


src/live-region/controller.ts
```diff
@@ -20,6 +24,22 @@ export class LiveRegionController {
...
 
+  /**
+   * Minimum time gap between announcements from different
+   * LiveRegion instances. This prevents screen readers from dropping
+   * announcements when multiple live regions update simultaneously.
+   *
+   * Screen readers throttle announcements to avoid overwhelming users.
+   * A 500ms gap ensures each announcement is detected and spoken.
+   */
+  private static readonly MIN_ANNOUNCEMENT_GAP_MS = 500;
```
avinashbot, line 35: 500 seems quite long, maybe you can get away with 250?

src/live-region/controller.ts
```diff
@@ -48,22 +68,58 @@ export class LiveRegionController {
...
+    if (effectiveDelay === 0 || forceReannounce) {
       // If the delay is 0, just skip the timeout shenanigans and update the
       // element synchronously. Great for tests.
       return this._updateElement(forceReannounce);
     }
 
-    if (this._timeoutId === undefined) {
-      this._timeoutId = setTimeout(() => this._updateElement(false), this.delay * 1000);
+    // Clear any existing timeout to ensure the latest announcement is used.
+    if (this._timeoutId !== undefined) {
```
avinashbot, line 96: This significantly changes the behavior of the live region. Previously, we throttled announcements — if the live region triggered 50 times in 2 seconds, it would only trigger an update twice (the first one, and another one at the end of the 2s period).

After this update, updates are _debounced_, not throttled (since each call clears the timeout) and (if I can visualize how this works correctly) if this gets called multiple times, each call will just add to the _nextAvailableSlot. So if trigger the live region 50 times in 2 seconds, you'll only hear one announcement 25 seconds later.

The weird paradox here is that we _do_ want to deduplicate and remove announcements if they happen too frequently (which is why this controller exists), but we don't want to do it _between_ different live regions. Maybe you can preserve the existing behavior, give each live region a unique incrementing ID, and only do the slot behavior if the last slot didn't belong to the current live region. Or, do this check inside `_updateElement` so that existing behavior doesn't need to be messed with _unless_ there's the edge case that another live region fired and we have to delay our update a little bit more?

PR #3993, opened 2025-10-30
fix: vertical centering for h2 headers without description in expandable sections

Who-is-PS:
### Description

- Added CSS rule to vertically center h2 headers without descriptions in expandable section
- Ensures consistent vertical alignment between header text and action buttons/icons across all expandable section
- Affected components are expandable section container variant

`AWSUI-61245`
### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/expandable-section/expandable-section-header.tsx
avinashbot, line 205: `.header-text` is set on this element, but `.not-expanded` isn't set here.

src/expandable-section/styles.scss
```diff
@@ -201,6 +206,11 @@ $icon-total-space-medium: calc(#{$icon-width-medium} + #{$icon-margin-left} + #{
     > .icon-container {
       margin-block-start: awsui.$space-expandable-section-icon-offset-top;
     }
+
+    // stylelint-disable-next-line @cloudscape-design/no-implicit-descendant -- header-text is nested within header-container-button
+    .header-text.not-expanded {
```
avinashbot: I don't see `not-expanded` being set on `header-text`. See comment above.

After that, a nicer alternative to disabling this linter error is to move this declaration out of the nesting; there's no need for it to be nested (the classnames are hashed anyway, so you don't have to worry about this `header-text` clashing with any other component's `header-text`).

Who-is-PS: I removed this code section as it was erroneously included. Both this code and the previous code snippet were from earlier commits, but they do not affect the styling functionality. Therefore, I've removed the redundant code to maintain cleanliness.

src/expandable-section/expandable-section-header.tsx
Who-is-PS: Header-test was originally set. I removed the styling for .not-expanded. This was a mistake from my side and it is not longer needed.

PR #3994, opened 2025-10-30
feat: add custom primary actions for wizard component

SpyZzey:
### Description

Adds functionality to overwrite the content of the primary actions for wizards (e.g. Next, Previous, Cancel, ...).
Currently, the Wizard only allows the default primary actions (Next, Previous, Cancel, Finish, Skip to ...). To allow further flexibility, a `customPrimaryActions`-property was introduced to allow custom actions.

### How has this been tested?

**Unit Testing**
- Added unit test to check if custom primary actions are rendered instead of default buttons
- Added unit test to check if custom primary actions work on all steps
- Added unit test to check if custom primary actions overwrite skip-to button
- Added unit test to check if default buttons are used if custom primary actions are null or undefined

**Manual Testing**
- Confirmed custom primary actions are being rendered
- Confirmed custom primary actions can be used to navigate steps
- Confirmed custom primary actions can be customized per step

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/__tests__/snapshot-tests/__snapshots__/documenter.test.ts.snap
+++ b/src/__tests__/snapshot-tests/__snapshots__/documenter.test.ts.snap
@@ -28028,7 +28028,7 @@ Use this if you need to wait for a response from the server before the user can 
   ],
   "regions": [
     {
-      "description": "Specifies right-aligned custom primary actions for the wizard. Overwrites existing buttons (e.g. Cancel, Next, Finish, ...).",
+      "description": "Specifies right-aligned custom primary actions for the wizard. Overwrites existing buttons (e.g. Cancel, Next, Finish).",
       "isDefault": false,
       "name": "customPrimaryActions",
       "systemTags": [


just-boris (commented):
There is a failing build. Let's fix this first and then I will review the rest

pages/wizard/custom-primary-actions.page.tsx
```diff
@@ -0,0 +1,97 @@
...
+    }
+    setActiveStepIndex(activeStepIndex - 1);
+  };
+
+  const onFinish = () => {
+    alert('Finish');
+  };
+
+  const customPrimaryActions = (
+    <SpaceBetween size="m" direction="horizontal">
```
just-boris: We should be using `xs` spacing size between buttons. There is a mention of this in our docs: https://cloudscape.design/components/button/?tabId=api

pages/wizard/custom-primary-actions.page.tsx
```diff
@@ -0,0 +1,97 @@
...
+      )}
+    </SpaceBetween>
+  );
+
+  return (
+    <Wizard
+      id="wizard"
+      steps={steps}
+      i18nStrings={i18nStrings}
+      allowSkipTo={true}
```
just-boris: What does this prop do in this case? if we render a fully custom actions, this property has no effect, right?

src/wizard/__tests__/wizard.test.tsx
```diff
@@ -515,6 +515,83 @@ describe('Custom actions', () => {
...
+    });
+
+    DEFAULT_STEPS.forEach((_, index) => {
+      const customActionButtonWrapper = wrapper.findActions()!.findButton('[data-testid="custom-action"]');
+      expect(customActionButtonWrapper).not.toBeNull();
+      customActionButtonWrapper?.click();
+      expect(onCustomClick).toHaveBeenCalledTimes(index + 1);
+
+      // Navigate to next step for next iteration
+      console.log(wrapper.findContent()?.getElement().innerHTML);
```
just-boris: We do not need console.logs in the mergeable code

src/wizard/__tests__/wizard.test.tsx
```diff
@@ -515,6 +515,83 @@ describe('Custom actions', () => {
...
+
+    const [wrapper, rerender] = renderDefaultWizard({
+      customPrimaryActions: customActions,
+      activeStepIndex: 0,
+    });
+
+    DEFAULT_STEPS.forEach((_, index) => {
+      const customActionButtonWrapper = wrapper.findActions()!.findButton('[data-testid="custom-action"]');
+      expect(customActionButtonWrapper).not.toBeNull();
+      customActionButtonWrapper?.click();
```
just-boris: This should be a strict non-null assertion, `!`. Because otherwise the test silently skips this line

src/wizard/__tests__/wizard.test.tsx
```diff
@@ -515,6 +515,83 @@ describe('Custom actions', () => {
...
+    const [wrapper] = renderDefaultWizard({ customPrimaryActions: customActions });
+
+    expect(wrapper.findPrimaryButton()).toBeNull();
+    expect(wrapper.findCancelButton()).toBeNull();
+    expect(wrapper.findPreviousButton()).toBeNull();
+    expect(wrapper.findActions()!.findButton('[data-testid="custom-cancel"]')).not.toBeNull();
+    expect(wrapper.findActions()!.findButton('[data-testid="custom-next"]')).not.toBeNull();
+  });
+
+  test('custom primary actions work on all steps', () => {
```
just-boris: Is it really needed to check on all steps? I think it would be better to check 3 corner-cases specifically 1) first step 2) any middle step 3) last step

src/wizard/interfaces.ts
```diff
@@ -103,6 +103,13 @@ export interface WizardProps extends BaseComponentProps {
    */
   allowSkipTo?: boolean;
 
+  /**
+   * Specifies right-aligned custom primary actions for the wizard. Overwrites existing buttons (e.g. Cancel, Next, Finish, ...).
```
just-boris: This `...).` punctuation in the end looks strange, you will need to check with a designer if there is a better way

src/wizard/__tests__/wizard.test.tsx
SpyZzey: The DEFAULT_STEPS contains 3 steps, so it'll check the first, middle and last step or would you suggest separating the test into three separate ones?

src/wizard/__tests__/wizard.test.tsx
SpyZzey: The customActionButtonWrapper is asserted the line before, so it would throw an error before. But I'll fix it to be clearer.

src/wizard/__tests__/wizard.test.tsx
SpyZzey: Agree.

pages/wizard/custom-primary-actions.page.tsx
SpyZzey: Correct. The property won't have an effect in this case. It's in the dev demo pages and I used it for manual testing.

src/wizard/__tests__/wizard.test.tsx
just-boris: yes, 3 separate ones would be easier to read and understand why they exist

just-boris (approved):

PR #3997, opened 2025-11-01
chore: Vendor react-keyed-flatten-children dependency

TrevorBurnham:
### Description

We're unable to upgrade to the latest versions of [react-keyed-flatten-children](https://github.com/grrowl/react-keyed-flatten-children) because of how it's packaged: We have Jest customers that are allergic to ESM. So we're stuck on an old version.

From https://github.com/cloudscape-design/components/discussions/3080:

> It is such a small utility that sometimes I wonder if it's worth inlining into the package…

This PR does exactly that: It adds the dep (specifically, `react-keyed-flatten-children@3.2.0` to our `vendor/` dir and adds an appropriate ranged dependency on `react-is`. That ranged dependency should be a significant efficiency improvement for many of our customers: The version of `react-keyed-flatten-children` we were using depended on `react-is@^18.2.0`, which meant that customers with any higher or lower version of `react-is` would wind up with two copies of that dependency in their bundle.

### How has this been tested?

All unit tests pass.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/vendor/react-keyed-flatten-children/index.ts
+++ b/src/internal/vendor/react-keyed-flatten-children/index.ts
@@ -1,0 +1,31 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// Vendored from react-keyed-flatten-children@3.2.0 (MIT License)
+// Original source: https://github.com/grrowl/react-keyed-flatten-children
+
+import { Children, cloneElement, isValidElement, ReactElement, ReactNode } from 'react';
+import { isFragment } from 'react-is';
+
+/* Returns React children into an array, flattening fragments. */
+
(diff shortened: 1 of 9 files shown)


src/internal/vendor/react-keyed-flatten-children/index.ts
```diff
@@ -0,0 +1,31 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// Vendored from react-keyed-flatten-children@3.2.0 (MIT License)
```
pan-kot, line 4: We need to mention the library in 3rd-party licenses, too. Example: https://github.com/cloudscape-design/test-utils/pull/81

TrevorBurnham: Done!

PR #3999, opened 2025-11-03
fix: Multi app layout detection

georgylobko:
### Description

Fixed multiple app layout detection by defaulting `hasToolbar` to `false`.
Currently the issue is when there are nested app layouts on a page and a left drawer is registered, it sneaks into the nested app layout and satisfies this condition: https://github.com/cloudscape-design/components/blob/main/src/app-layout/visual-refresh-toolbar/state/use-app-layout.tsx#L227

which results in two left drawers registered on the page

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


just-boris (commented):
Two asks

1. There is one more `hasToolbar: true` value here: https://github.com/cloudscape-design/components/blob/a59c74fc5ff6e6ef38943a55bee2366c0cc9007a/src/app-layout/visual-refresh-toolbar/state/index.tsx#L34

2. If this change is that important, it needs to have a regression test

PR #4000, opened 2025-11-03
fix: Toolbar with triggerless global drawers

georgylobko:
### Description

Add an additional check to props-merger if global drawers have at least one item with a trigger

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/__tests__/multi-layout-props.test.tsx
+++ b/src/app-layout/__tests__/multi-layout-props.test.tsx
@@ -129,6 +129,50 @@ describe('mergeMultiAppLayoutProps', () => {
     });
   });
 
+  it('should merge ownProps and additionalProps (triggerless global drawers)', () => {
+    const result = mergeProps(ownProps, [
+      {
+        ...additionalPropsBase[0],
+        globalDrawers: [
+          {
+            id: 'drawer-global',
+            ariaLabels: { drawerName: 'Global Drawer' },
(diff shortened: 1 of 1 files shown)


src/app-layout/visual-refresh-toolbar/state/props-merger.ts
```diff
@@ -31,7 +31,11 @@ export const mergeProps: MergeProps = (ownProps, additionalProps) => {
       toolbar.drawersFocusRef = props.drawersFocusRef;
       toolbar.onActiveDrawerChange = props.onActiveDrawerChange;
     }
-    if (props.globalDrawers && !checkAlreadyExists(!!toolbar.globalDrawers, 'globalDrawers')) {
+    if (
```
mxschll, line 34: As far as I understand, drawers without triggers are valid, because they can be opened programmatically via the plugin API. Maybe we could add a comment here as for why these drawers without triggers are filtered out?

georgylobko: Yes, these drawers are valid. But the props-merger function returns properties for _toolbar_ - the horizontal panel on top of the page and contains triggers for drawers. If there is at lest one global drawer with a trigger, we need to provide these properties to the toolbar, otherwise no. It works pretty much the same as `drawers` in this file.

PR #4002, opened 2025-11-03
chore: React 18 compatible implementation of iframe wrapper

just-boris:
### Description

Continuing that story: https://github.com/cloudscape-design/components/pull/3956

The iframe tests still failing. Current attempt – avoid `useEffect` because it has known behavior of being called multiple times: https://stackoverflow.com/questions/72238175/why-useeffect-running-twice-and-how-to-handle-it-well-in-react

They have good reasons for this behavior, but they are incompatible with our use-case

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/utils/iframe-wrapper.tsx
+++ b/pages/utils/iframe-wrapper.tsx
@@ -1,6 +1,6 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
-import React, { useEffect, useRef } from 'react';
+import React, { useCallback, useRef } from 'react';
 
 import { mount, unmount } from '~mount';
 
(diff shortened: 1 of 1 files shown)


avinashbot (approved):
Assuming the root cause is the same as the previous PR, this makes sense to me if we're just concerned with the issue of flaky failures in our build stages.

But if we want to treat this as an indication of a possible edge-case on the customer's end, it's worth knowing that React also started [double-calling callback refs](https://github.com/facebook/react/blob/main/CHANGELOG.md#:~:text=Additionally%2C%20StrictMode%20will%20now%20double%2Dinvoke%20ref%20callback%20functions%20on%20initial%20mount.) in strict mode starting from React 19. So "to avoid double effect issues in React 18+" isn't strictly correct. Just something to consider — maybe the "future-proof" fix is in the styles check, or moving iframe code out of the component lifecycle (some imperative way, unsure how exactly).

But at the same time, it seems like the pressing issue is to deal with the build flakiness introduced in React 18, so I don't want to block this; seems fine for now.

just-boris:
Discussed in DM, the current solution is best we could think of.

Note for the future React 19 compatibility, we can replace `useCallback` with `useDebounceCallback`. This seems to deduplicate render calls as expected

PR #4005, opened 2025-11-04
fix: Fixes chart popover click outside when in iframe

pan-kot:
### Description

Rel: [AWSUI-61421]

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/components/chart-popover/__integ__/chart-popover.test.ts
+++ b/src/internal/components/chart-popover/__integ__/chart-popover.test.ts
@@ -6,25 +6,26 @@ import useBrowser from '@cloudscape-design/browser-test-tools/use-browser';
 
 import createWrapper from '../../../../../lib/components/test-utils/selectors';
 
-test.each([false, true])('can be unpinned by clicking outside, iframe=%s', async iframe => {
-  await useBrowser(async browser => {
-    const page = new BasePageObject(browser);
-    await browser.url(`#/light/charts.test?iframe=${iframe}`);
-    await page.runInsideIframe('#content-iframe', iframe, async () => {
-      const chart = createWrapper().findLineChart();
-      const popover = chart.findDetailPopover();
-      const popoverDismiss = popover.findDismissButton();
+describe.each([false, true])('iframe=%s', iframe => {
+  test(
+    'can be unpinned by clicking outside',
+    useBrowser(async browser => {
+      const page = new BasePageObject(browser);
+      await browser.url(`#/light/charts.test?iframe=${iframe}`);
+      await page.runInsideIframe('#content-iframe', iframe, async () => {
+        const chart = createWrapper().findLineChart();
+        const popover = chart.findDetailPopover();
+        const popoverDismiss = popover.findDismissButton();
 
-      // Show popover with an offset from chart's center.
-      await page.hoverElement(chart.toSelector(), 100);
-      // Pin the popover by clicking on the left from it to ensure the click lands on the SVG, not popover content.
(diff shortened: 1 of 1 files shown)


src/internal/components/chart-popover/index.tsx
```diff
@@ -87,23 +86,27 @@ function ChartPopover(
 ) {
   const baseProps = getBaseProps(restProps);
   const popoverObjectRef = useRef<HTMLDivElement | null>(null);
-
   const popoverRef = useMergeRefs(popoverObjectRef, ref);
 
+  const clickFrameId = useRef<number | null>(null);
```
pan-kot, line 91: We use the same implementation in the internal popover component. The chart popover cannot use it because it imports the popover container instead.

src/internal/components/chart-popover/index.tsx
```diff
@@ -87,23 +86,27 @@ function ChartPopover(
...
+
     const onDocumentClick = (event: MouseEvent) => {
-      if (
-        event.target &&
-        !nodeBelongs(popoverObjectRef.current, event.target as Element) && // click not in popover
-        !nodeContains(container, event.target as Element) // click not in segment
-      ) {
+      // Dismiss popover unless there was a click inside within the last animation frame.
+      // Ignore clicks inside the chart as those are handled separately.
+      if (clickFrameId.current === null && !nodeContains(container, event.target as Element)) {
```
pan-kot: The `!nodeContains(container, event.target as Element)` is a check that is chart-popover specific, and is taken from the prev. implementation - not from the internal popover. Its purpose is to prevent dismiss action in case the click is done within the chart's SVG so that the chart itself can define the appropriate popover behaviour in that case.

src/mixed-line-bar-chart/chart-container.tsx
```diff
@@ -391,11 +392,7 @@ export default function ChartContainer<T extends ChartDataTypes>({
 
   const onApplicationBlur = (event: React.FocusEvent<Element>) => {
     const blurTarget = event.relatedTarget || event.target;
-    if (
-      blurTarget === null ||
-      !(blurTarget instanceof Element) ||
```
pan-kot, line 396: The `blurTarget instanceof Element` fails if the target belongs to a different document. As result, this if condition always resolves to true and thus causes the popover to be dismissed unexpectedly.

See:

https://github.com/user-attachments/assets/af34eeaa-95ff-49c8-b9c7-86b14330db1b

Replacing the check with `!isElement(blurTarget)` fixes the problem.

gethinwebster (approved):

src/internal/components/chart-popover/__integ__/chart-popover.test.ts
```diff
@@ -0,0 +1,31 @@
...
+      const page = new BasePageObject(browser);
+      await browser.url(`#/light/charts.test?iframe=${iframe}`);
+      await page.runInsideIframe('#content-iframe', iframe, async () => {
+        const chart = createWrapper().findLineChart();
+        const popover = chart.findDetailPopover();
+        const popoverDismiss = popover.findDismissButton();
+
+        // Pins popover on the first point.
+        await page.click('h2');
+        await page.keys(['Tab', 'ArrowRight', 'Enter']);
```
pan-kot, line 22: @gethinwebster I changed popover activation from pointer to keyboard because the former approach did not work with React 16 test pages: the page.hoverElement() did not help with offsetting the click. I did test the feature manually on both React versions.

PR #4008, opened 2025-11-04
fix: header focus outline to use CDS styles instead of native outline

Harsh-Anand-Singh:
### Description

This PR fixes the Header component's focus outline to use Cloudscape Design System (CDS) styles instead of the native browser outline.

Related links, issue #, if available: AWSUI-61110

### How has this been tested?

I've tested the focus behavior locally:

**Safari with custom system colors**: 
   - Changed the system appearance accent color from default (System Settings → General → Appearance → Accent color)
   - Confirmed focus outline in Wizard component
   - Confirmed the focus outline now uses CDS blue color and is distinctly different from the custom accent color
   - Verified the outline is no longer affected by system color preferences

**Other focusable headers**:
   - Tested headers in various demo pages (App Layout Toolbar, Wizard, etc.)
   - Confirmed all headers that receive focus they display consistent CDS focus styling

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/header/styles.scss
+++ b/src/header/styles.scss
@@ -5,6 +5,7 @@
 
 @use '../internal/styles/tokens' as awsui;
 @use '../internal/styles' as styles;
+@use '@cloudscape-design/component-toolkit/internal/focus-visible' as focus-visible;
 
 @function space-heading-button-diff($heading-height) {
   @return calc((#{awsui.$size-vertical-input} - #{$heading-height}) / 2);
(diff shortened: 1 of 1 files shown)


src/header/styles.scss
```diff
@@ -220,6 +221,14 @@
   font-size: inherit;
   @include styles.info-link-spacing();
 
+  &:focus {
+    outline: none;
+  }
+
+  @include focus-visible.when-visible {
+    @include styles.link-focus;
```
mxschll, line 229: [non-blocking]: I think link-focus is confusing terminology, because it's not a link. What about focus ring?

Harsh-Anand-Singh: Good point about the naming! The link-focus mixin is actually a generic focus mixin used across many non-link components (SideNavigation, Wizard navigation, Breadcrumbs, etc.). Despite its name, it provides the standard CDS focus outline for any focusable element.

PR #4009, opened 2025-11-04
fix: Handle LiteralUnion types correctly in documenter for CodeEditor language property

Who-is-PS:
### Description

Fixed an issue where `CodeEditorProps.Language` type values were incorrectly formatted in generated component definitions. The `LiteralUnion<BuiltInLanguage, string>` type was not properly recognized as a primitive string type, causing language values to appear as `"\"javascript\""` instead of `"javascript"` in the documentation.

`AWSUI-61340`

### How has this been tested?
- Added comprehensive unit tests in `src/code-editor/__tests__/language-selector.test.tsx`
- Verified built-in languages (e.g., `'javascript'`) display correctly
- Verified custom languages (e.g., `'partiql'`) work with labels
- Confirmed `LiteralUnion` type assignments work without TypeScript errors
- Updated and verified snapshot tests pass with corrected format
- Tested that `lib/components-definitions/components/code-editor.js` now generates clean values

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/patches/@cloudscape-design+documenter+1.0.66.patch
+++ b/patches/@cloudscape-design+documenter+1.0.66.patch
@@ -7,11 +7,15 @@ index d504cb2..277d099 100644
          return 'number';
      }
 +    // Handle LiteralUnion pattern: string literals + generic string fallback
-+    const hasOnlyStringLiteralsAndGenericString = type.types.every(subtype => 
-+        subtype.isStringLiteral() || 
-+        (subtype.flags & typescript_1.default.TypeFlags.String && !subtype.isStringLiteral())
-+    );
-+    if (hasOnlyStringLiteralsAndGenericString) {
++    const hasStringLiteralsAndGenericString = type.types.some(subtype => subtype.isStringLiteral()) &&
++        type.types.some(subtype => 
++            subtype.flags & typescript_1.default.TypeFlags.String && 
++            !subtype.isStringLiteral() &&
++            subtype.symbol?.name?.includes('_')
++        );
++    if (hasStringLiteralsAndGenericString && type.types.every(subtype => 
++        subtype.isStringLiteral() || (subtype.flags & typescript_1.default.TypeFlags.String)
++    )) {
 +        return 'string';
 +    }
      return undefined;


src/__tests__/snapshot-tests/__snapshots__/documenter.test.ts.snap
```diff
@@ -14669,7 +14669,7 @@ including the date, month, week, time, datetime-local, number and range types.",
         "type": "union",
         "values": [
```
gethinwebster, line 14670: Based on this example, I'm not sure the previous behavior was actually incorrect. In this case, the correct type is `number | "any"`, which is no longer preserved correctly.

gethinwebster (commented):
I _think_ that the current behavior is correct/expected. And so this should probably be handled in the website instead of the documenter. Or, maybe the documenter needs to create two different types of unions: string-union and literal-union or something like that. With the current proposed change, we're losing information about the underlying types

src/__tests__/snapshot-tests/__snapshots__/documenter.test.ts.snap
Who-is-PS: I've updated the patch to be more targeted, it now only treats true LiteralUnion 
patterns (those with the { _?: never } marker) as string unions, while preserving mixed unions like number | 'any' with their full type information.

patches/@cloudscape-design+documenter+1.0.66.patch
```diff
@@ -0,0 +1,40 @@
+diff --git a/node_modules/@cloudscape-design/documenter/components/object-definition.js b/node_modules/@cloudscape-design/documenter/components/object-definition.js
```
gethinwebster, line 1: I guess this file shouldn't be here?

src/__tests__/snapshot-tests/__snapshots__/documenter.test.ts.snap
```diff
@@ -4439,16 +4439,16 @@ Sizes are automatically scaled down in compact mode.
...
+          "m",
+          "n",
+          "xxxs",
+          "xxs",
+          "xs",
+          "l",
+          "xl",
+          "xxl",
+          "xxxl",
           "BoxProps.Spacing",
```
gethinwebster, line 4452: this one is still incorrect: "s", "m" etc. are strings, `BoxProps.Spacing` is a type, but all look the same in the new output

PR #4025, opened 2025-11-08
feat: implement style api for toggle button component

mxschll:
### Description

Extends the toggle button component with a style API for customization, allowing developers to override default component styles.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


pages/toggle-button/style-permutations.page.tsx
```diff
@@ -0,0 +1,162 @@
...
+  {
+    variant: ['normal'],
+    children: ['Subscribe'],
+    pressed: [false, true],
+    disabled: [false, true],
+    iconName: ['thumbs-up'],
+    pressedIconName: ['thumbs-up-filled'],
+    onChange: [() => {}],
+    style: [
+      {
```
pan-kot: To improve readability, I recommend extracting styles to some named variables, like "borderlessStyle" or "contrastyStyle" etc. This will improve two things:
1. give the overview of the different styles that we use (maybe we don't need that many?)
2. make it easier to comprehend the diff in between permutations

src/toggle-button/style.tsx
```diff
@@ -0,0 +1,43 @@
...
+    [customCssProps.styleColorDefault]: style?.root?.color?.default,
+    [customCssProps.styleColorDisabled]: style?.root?.color?.disabled,
+    [customCssProps.styleColorHover]: style?.root?.color?.hover,
+    [customCssProps.styleColorPressed]: style?.root?.color?.pressed,
+    [customCssProps.styleFocusRingBorderColor]: style?.root?.focusRing?.borderColor,
+    [customCssProps.styleFocusRingBorderRadius]: style?.root?.focusRing?.borderRadius,
+    [customCssProps.styleFocusRingBorderWidth]: style?.root?.focusRing?.borderWidth,
+  };
+
+  return Object.fromEntries(Object.entries(properties).filter(([, value]) => value));
```
pan-kot: why do we need this?

src/toggle-button/style.tsx
```diff
@@ -0,0 +1,43 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import { SYSTEM } from '../internal/environment';
+import customCssProps from '../internal/generated/custom-css-properties';
+import { ToggleButtonProps } from './interfaces';
+
+export function getToggleButtonStyles(style: ToggleButtonProps['style']) {
+  if (SYSTEM !== 'core') {
```
pan-kot, line 8: In the button utils we do:
```
if (SYSTEM !== 'core' || !style?.root) {
    return undefined;
}
```

is there a convention around it?

src/toggle-button/internal.tsx
```diff
@@ -62,7 +64,10 @@ export const InternalToggleButton = React.forwardRef(
         }}
         {...rest}
         ref={ref}
-        nativeButtonAttributes={nativeButtonAttributes}
+        nativeButtonAttributes={{
+          ...nativeButtonAttributes,
+          style: { ...nativeButtonAttributes?.style, ...getToggleButtonStyles(style) } as React.CSSProperties,
```
pan-kot: The internal button component accepts the `style` prop already - why do we merge it with the native attributes instead?

src/toggle-button/__tests__/styles.test.tsx
```diff
@@ -0,0 +1,260 @@
...
+        },
+        focusRing: {
+          borderColor: '#0073bb',
+          borderRadius: '6px',
+          borderWidth: '2px',
+        },
+      },
+    };
+
+    expect(getToggleButtonStyles(allStyles)).toEqual({
```
pan-kot: If this test only validates the correct props assignment, why not use snapshot tests instead?

pages/toggle-button/style-permutations.page.tsx
mxschll: Good point, will change than!

src/toggle-button/__tests__/styles.test.tsx
mxschll: Is the snapshot test enough for that? I wanted to make sure the function fulfills the task of creating the style object correctly.

src/toggle-button/style.tsx
mxschll: We don't have a convention. As per [this comment](https://github.com/cloudscape-design/components/pull/4026#discussion_r2514295668) I will refactor the existing components with the style api to match this.

src/toggle-button/style.tsx
mxschll: We filter out undefined properties with that, to prevent any unwanted side effects. I am changing it to be more explicit:

```
return Object.fromEntries(Object.entries(properties).filter(([, value]) => value !== undefined));
```

src/toggle-button/internal.tsx
mxschll: The toggle button has a `pressed` state, the button doesn't. We would need to apply the pressed state styling to the native attributes and could pass the other styles to the `style` prop. That adds more complexity than necessary. In general, there is no mechanism yet for nested component styling.

src/toggle-button/__tests__/styles.test.tsx
mxschll: Resolved offline.

src/toggle-button/style.tsx
```diff
@@ -0,0 +1,43 @@
...
+    [customCssProps.styleColorDefault]: style?.root?.color?.default,
+    [customCssProps.styleColorDisabled]: style?.root?.color?.disabled,
+    [customCssProps.styleColorHover]: style?.root?.color?.hover,
+    [customCssProps.styleColorPressed]: style?.root?.color?.pressed,
+    [customCssProps.styleFocusRingBorderColor]: style?.root?.focusRing?.borderColor,
+    [customCssProps.styleFocusRingBorderRadius]: style?.root?.focusRing?.borderRadius,
+    [customCssProps.styleFocusRingBorderWidth]: style?.root?.focusRing?.borderWidth,
+  };
+
+  return Object.fromEntries(Object.entries(properties).filter(([, value]) => value !== undefined));
```
pan-kot: Why do we need to filter out undefined props?

mxschll: If we don't filter them out, we rely on the guarantee that style properties with undefined values don't cause unintended side effects (like resetting styles). By filtering them out, we ensure this behavior remains consistent even if implementation details change in the future.

Also, without filtering, we'd return an object containing all properties, including those with undefined values:

```
{
    borderRadius: '10px',
    borderWidth: undefined,
    paddingBlock: undefined,
    paddingInline: undefined,
    ...
  };
  ```
 
Filtering produces a cleaner, more focused result:
 
 ```
 {
    borderRadius: '10px',
};
  ```

File: src/toggle-button/style.tsx (hunk 2 above)
Comment (pan-kot) [reply src/toggle-button/style.tsx]: What side effects, for example? I noticed that we don't do that e.g. in the button's style utils. If there is a value - let's do it uniformly, and for that a shared utility would be nice to have. If there is no real value - let's not do that.

File: src/toggle-button/internal.tsx (hunk 4 above)
Comment (pan-kot) [reply src/toggle-button/internal.tsx]: Well, I would rather do it as you said: pass part of the style using `style`, and the rest using native attributes. Then add a code comment documenting why do we need that second part. This makes the separation explicit and benefits future refactoring once we have a good approach for nested component styling.

File: src/toggle-button/style.tsx (hunk 6 above)
Comment (pan-kot) [reply src/toggle-button/style.tsx]: I understand the diff, but disagree with the approach. Besides, there is a performance penalty when doing this: not only we need to iterate over all properties, we also need to construct a new object from that.

It is not very big impact, but it grows with the number of properties and components on the page. Imagine a future table with lots of nested styles, propagated to its internal checkboxes, etc. Then imagine that in every row there are a few inline buttons with styles. Currently, all these styles are not handled separately (e.g. in some style provider), but applied in the component directly. This can account for some ~10ms of processing time or even more on lower-performing devices for this transformation alone.

Comment (pan-kot) [reply src/toggle-button/style.tsx]: As this PR intention is to add styles for toggle buttons - I would remove this transformation. We can solve it separately and make sure to include all existing components with style API, which is best to be done with a shared utility.

Comment (pan-kot) [reply src/toggle-button/style.tsx]: Screenshot 2025-11-14 at 08 29 25

Comment (mxschll) [reply src/toggle-button/style.tsx]: I agree, thanks for pointing that out. I'll remove this logic and think of a utility function in the next step that will be used by all styled components when refactoring.

File: src/toggle-button/internal.tsx (hunk 4 above)
Comment (mxschll) [reply src/toggle-button/internal.tsx]: Ok, I will change it to handle that 👍

File: src/toggle-button/style.tsx (hunk 2 above)
Comment (mxschll) [reply src/toggle-button/style.tsx]: Discussion moved here: https://github.com/cloudscape-design/components/pull/4025#discussion_r2523574346

File: src/toggle-button/__tests__/styles.test.tsx [hunk 7]
```diff
@@ -0,0 +1,277 @@
...
+      nativeButtonStyles: {
+        [customCssProps.styleBackgroundPressed]: '#e0e0e0',
+        [customCssProps.styleBorderColorPressed]: '#0073bb',
+        [customCssProps.styleBoxShadowPressed]: '0 0 0 2px #0073bb',
+        [customCssProps.styleColorPressed]: '#0073bb',
+      },
+    });
+
+    // Should NOT contain non-pressed states (these are handled by InternalButton)
+    expect(result.nativeButtonStyles).not.toHaveProperty('borderRadius');
```
pan-kot, line 111: Isn't the check above enough, why do we explicitly check for these props to not be present?

src/toggle-button/__tests__/styles.test.tsx
```diff
@@ -0,0 +1,277 @@
...
+        // and are not extracted to nativeButtonStyles
+        expect(result).toEqual({
+          nativeButtonStyles: {},
+        });
+      });
+    });
+  });
+
+  describe('state properties', () => {
+    STATE_PROPERTIES.forEach(property => {
```
pan-kot, line 152: Isn't that already verified with the "extracts only pressed states to nativeButtonStyles" test?

PR #4026, opened 2025-11-09
feat: implement style api for prompt input component

mxschll:
### Description

Extends the prompt input component with a style API for customization, allowing developers to override default component styles.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/prompt-input/styles.tsx
+++ b/src/prompt-input/styles.tsx
@@ -5,52 +5,42 @@ import customCssProps from '../internal/generated/custom-css-properties';
 import { PromptInputProps } from './interfaces';
 
 export function getPromptInputStyles(style: PromptInputProps['style']) {
-  let properties = {};
-
-  if (style?.root && SYSTEM === 'core') {
-    properties = {
-      borderRadius: style?.root?.borderRadius,
-      borderWidth: style?.root?.borderWidth,
-      fontSize: style?.root?.fontSize,
-      fontWeight: style?.root?.fontWeight,
-      paddingBlock: style?.root?.paddingBlock,
-      paddingInline: style?.root?.paddingInline,
-      ...(style?.root?.backgroundColor && {
-        [customCssProps.promptInputStyleBackgroundDefault]: style.root.backgroundColor?.default,
-        [customCssProps.promptInputStyleBackgroundDisabled]: style.root.backgroundColor?.disabled,
-        [customCssProps.promptInputStyleBackgroundHover]: style.root.backgroundColor?.hover,
-        [customCssProps.promptInputStyleBackgroundFocus]: style.root.backgroundColor?.focus,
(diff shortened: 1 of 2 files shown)


src/prompt-input/internal.tsx
```diff
@@ -206,6 +208,7 @@ const InternalPromptInput = React.forwardRef(
         })}
         ref={__internalRootRef}
         role="region"
+        style={getPromptInputStyles(style) as React.CSSProperties}
```
just-boris: `as React.CSSProperties` should not be needed here. There is probably something wrong with `getPromptInputStyles` function signature and you need to fix the signature instead of type casting (type-castings are unsafe)

src/prompt-input/styles.tsx
```diff
@@ -0,0 +1,56 @@
...
+        [customCssProps.promptInputStyleBoxShadowReadonly]: style.root.boxShadow?.readonly,
+      }),
+      ...(style?.root?.color && {
+        [customCssProps.promptInputStyleColorDefault]: style.root.color?.default,
+        [customCssProps.promptInputStyleColorDisabled]: style.root.color?.disabled,
+        [customCssProps.promptInputStyleColorHover]: style.root.color?.hover,
+        [customCssProps.promptInputStyleColorFocus]: style.root.color?.focus,
+        [customCssProps.promptInputStyleColorReadonly]: style.root.color?.readonly,
+      }),
+      ...(style?.placeholder && {
```
just-boris: Looks like a repetition of AWSUI-61442 bug report

You need to fix it here as well and allow find a future proof solution how to prevent this from happening in all other components

src/prompt-input/internal.tsx
mxschll: You are right. Updating it.

src/prompt-input/styles.tsx
mxschll: Unlike AWSUI-61442, this component doesn't have multiple nested sections that can have different styles. Not sure how it relates to AWSUI-61442.

just-boris: If you do not provide `style.root` the `style.placeholder` will be ignored

mxschll: Sorry, copy-paste error on my side. We can just check `if (SYSTEM === 'core')`. Since we're using optional chaining, any style property that resolves to undefined will be ignored by. To make it cleaner, we could filter out undefined properties before passing the style object (see latest commit).

Alternatively, we could build the style object conditionally from the ground up, only adding properties that are defined. A utility function for this would be helpful since we'll likely need it across multiple components. What do you think?

just-boris: The new revision looks good, approved

But still an ask to pay more attention to these conditions, because they are prone to mistakes

mxschll: Thanks for paying close attention. I'll refactor the existing components that way too.

PR #4027, opened 2025-11-11
feat: implement style api for tabs component

mxschll:
### Description

Extends the tabs component with a style API for customization, allowing developers to override default component styles.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/tabs/style-permutations.page.tsx
+++ b/pages/tabs/style-permutations.page.tsx
@@ -44,7 +44,7 @@ const style1 = {
   underline: {
     color: 'light-dark(#3b82f6, #60a5fa)',
     width: '3px',
-    borderRadius: '3px 3px 0 0',
+    borderRadius: '3px',
   },
   divider: {
     color: 'transparent',
@@ -66,12 +66,12 @@ const style2 = {
       disabled: 'light-dark(#f8fafc, #1e293b)',
     },
     borderColor: {
-      default: 'transparent',
-      hover: 'transparent',
-      active: 'transparent',
-      disabled: 'transparent',
+      default: 'lightblue',
+      hover: 'lightblue',
+      active: 'lightblue',
+      disabled: 'lightblue',
     },
-    borderWidth: '0px',
+    borderWidth: '2px',
     borderRadius: '20px',
     color: {
       default: 'light-dark(#475569, #94a3b8)',
@@ -140,7 +140,7 @@ const style3 = {
   underline: {
     color: 'light-dark(#f59e0b, #fbbf24)',
     width: '4px',
-    borderRadius: '4px 4px 0 0',
+    borderRadius: '4px',
   },
   divider: {
     color: 'light-dark(#fde68a, #78350f)',


src/tabs/tab-header-bar.scss
```diff
@@ -113,11 +115,6 @@ $label-horizontal-spacing: awsui.$space-xs;
   display: flex;
   align-items: stretch;
 
-  &,
-  & > button {
-    background-color: transparent;
```
mxschll, line 118: Removing `transparent` because it overrides custom styles for invalid or disabled. Removal caused no visual regressions. Keeping this property would prevent custom styles from taking effect.

pages/tabs/style-permutations.page.tsx
```diff
@@ -0,0 +1,188 @@
...
+      active: 'transparent',
+      disabled: 'transparent',
+    },
+    borderColor: {
+      default: 'transparent',
+      hover: 'transparent',
+      active: 'transparent',
+      disabled: 'transparent',
+    },
+    borderWidth: '0px',
```
gethinwebster, line 26: having a permutation with non-zero borderWidth is important I think

pages/tabs/style-permutations.page.tsx
```diff
@@ -0,0 +1,188 @@
...
+      default: 'light-dark(#64748b, #94a3b8)',
+      hover: 'light-dark(#1e293b, #e2e8f0)',
+      active: 'light-dark(#0f172a, #ffffff)',
+      disabled: 'light-dark(#cbd5e1, #475569)',
+    },
+    fontSize: '14px',
+    fontWeight: '500',
+    paddingBlock: '12px',
+    paddingInline: '20px',
+    focusRing: {
```
gethinwebster, line 38: I think `focusRing` color might need states, because it is displayed over the background color. So, for example, it might not be possible to find an accessible focus ring color that works over both default and active background colors

pages/tabs/style-permutations.page.tsx
```diff
@@ -0,0 +1,188 @@
...
+      borderColor: 'light-dark(#3b82f6, #60a5fa)',
+      borderRadius: '6px',
+      borderWidth: '2px',
+    },
+  },
+  underline: {
+    color: 'light-dark(#3b82f6, #60a5fa)',
+    width: '3px',
+    borderRadius: '3px 3px 0 0',
```
gethinwebster: this doesn't seem to be actually visible in the page as far as I can see?

src/tabs/interfaces.ts
```diff
@@ -191,4 +196,51 @@ export namespace TabsProps {
      */
     tabsWithActionsAriaRoleDescription?: string;
   }
+
+  export interface Style {
+    tabs?: {
```
gethinwebster, line 201: I don't think the naming here is clear, I would suggest `tabButton`

src/tabs/interfaces.ts
```diff
@@ -191,4 +196,51 @@ export namespace TabsProps {
...
+      fontWeight?: string;
+      focusRing?: {
+        borderColor?: string;
+        borderRadius?: string;
+        borderWidth?: string;
+      };
+      paddingBlock?: string;
+      paddingInline?: string;
+    };
+    underline?: {
```
gethinwebster, line 232: I think this should be part of the `tabButton`

src/tabs/interfaces.ts
```diff
@@ -191,4 +196,51 @@ export namespace TabsProps {
...
+      color?: string;
+      width?: string;
+      borderRadius?: string;
+    };
+    divider?: {
```
gethinwebster, line 237: the structure/naming here isn't clear: what's the difference between `divider` and `headerDivider`?

pages/tabs/style-permutations.page.tsx
mxschll: I propose we add this to the existing components as well (e.g. button).

gethinwebster: For most components it appears outside the component itself so it's less of an issue as the "background" its displayed against isn't state-dependent. I guess the other alternative is to explore whether the ring could be moved outside the background here?

mxschll: Right, it's also displayed outside of the regular button component. That would make the API more streamlined across components, but I'm not sure if moving the ring will break something... I'll try to explore this further.

pages/tabs/style-permutations.page.tsx
mxschll: It's the blue underline here: 
image

But the border radius is unnecessary. Forgot to update that from the previous api iteration.

gethinwebster: cool, yeah it was specifically the border radius I meant :)

PR #4035, opened 2025-11-13
chore: Add Flashbar and Alert dismiss persistence

YueyingLu:
### Description

Add code to support Persistence feature in Flashbar and Alert [qIQZAtdAL9Xj].
The home of core function of Persistence is CR-199098474.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/alert/internal.tsx
+++ b/src/alert/internal.tsx
@@ -135,11 +135,7 @@ const InternalAlert = React.forwardRef(
         retrieveAlertDismiss(persistenceConfig)
           .then(dismissed => {
             if (isMounted) {
-              try {
-                setIsPersistentlyDismissed(!!dismissed);
-              } catch {
-                setIsPersistentlyDismissed(false);
-              }
+              setIsPersistentlyDismissed(!!dismissed);
             }
           })
           .catch(() => {
@@ -152,6 +148,7 @@ const InternalAlert = React.forwardRef(
         isMounted = false;
       };
 
+      // Only track specific properties to avoid re-running when persistenceConfig object reference changes
       // eslint-disable-next-line react-hooks/exhaustive-deps
     }, [persistenceConfig?.uniqueKey, persistenceConfig?.crossServicePersistence]);
 
--- a/src/flashbar/__tests__/persistence-unit.test.tsx
+++ b/src/flashbar/__tests__/persistence-unit.test.tsx
@@ -291,5 +291,23 @@ describe('Flashbar persistence', () => {
         expect(items[1].findContent()!.getElement()).toHaveTextContent('Item 2');
       });
     });
+
+    it('sets all items to visible when Promise.all fails', async () => {
+      // Mock Promise.all to fail
+      const originalPromiseAll = Promise.all;
+      Promise.all = jest.fn().mockRejectedValue(new Error('Promise.all failed'));
+
+      const items = createItems([true, true]);
+      const { container } = render(<Flashbar items={items} />);
+      const wrapper = createWrapper(container).findFlashbar()!;
+
+      await waitFor(() => {
+        expect(wrapper.findItems()).toHaveLength(2);
+        expect(wrapper.findItems()[0].findContent()!.getElement()).toHaveTextContent('Item 0');
+        expect(wrapper.findItems()[1].findContent()!.getElement()).toHaveTextContent('Item 1');
+      });
+
+      Promise.all = originalPromiseAll;
+    });
   });
 });


pages/flashbar/persistence.page.tsx
```diff
@@ -0,0 +1,119 @@
...
+          </Button>
+          <Toggle
+            data-id="stack-items"
+            checked={stackItems}
+            onChange={({ detail }) => {
+              setStackItems(detail.checked);
+              const url = new URL(window.location.href);
+              const params = new URLSearchParams(url.hash.split('?')[1] || '');
+              params.set('stackItems', detail.checked.toString());
+              window.history.replaceState({}, '', `${url.pathname}${url.hash.split('?')[0]}?${params}`);
```
just-boris: Any reason why this code does not use our standard url persistence utility, available via AppContext:

https://github.com/search?q=repo%3Acloudscape-design%2Fcomponents%20AppContext&type=code

src/alert/__tests__/persistence-unit.test.tsx
```diff
@@ -0,0 +1,149 @@
...
+    mockRetrieveAlertDismiss.mockResolvedValue(false);
+  });
+
+  describe('initialization', () => {
+    it('calls retrieveAlertDismiss when persistenceConfig is provided', async () => {
+      const persistenceConfig = { uniqueKey: 'test-alert' };
+      renderAlert({ persistenceConfig, dismissible: true });
+
+      await waitFor(() => {
+        expect(mockRetrieveAlertDismiss).toHaveBeenCalledWith(persistenceConfig);
```
just-boris: It is better to explicitly pass arguments instead of reusing the object. It will be easier to follow what data is being sent in and out

```suggestion
        expect(mockRetrieveAlertDismiss).toHaveBeenCalledWith({ uniqueKey: 'test-alert' });
```

src/alert/internal.tsx
```diff
@@ -124,6 +129,25 @@ const InternalAlert = React.forwardRef(
       i18nStrings?.dismissAriaLabel ?? i18n('dismissAriaLabel', deprecatedDismissAriaLabel)
     );
 
+    useEffect(() => {
+      if (persistenceConfig?.uniqueKey && persistenceConfig.uniqueKey.trim()) {
```
just-boris: Why is there a `trim()`? We should not modify their input in any way

src/alert/internal.tsx
```diff
@@ -124,6 +129,25 @@ const InternalAlert = React.forwardRef(
...
+        retrieveAlertDismiss(persistenceConfig).then(dismissed => {
+          setIsPersistentlyDismissed(!!dismissed);
+        });
+      }
+    }, [persistenceConfig]);
```
just-boris: ⛳  Because `persistenceConfig` is an object, this effect will be called multiple times even if the uniqueKey does not change. 

You need to make this update condition more precise.

src/flashbar/__tests__/persistence-unit.test.tsx
```diff
@@ -0,0 +1,200 @@
...
+      await waitFor(() => {
+        expect(wrapper.findItems()).toHaveLength(1);
+        expect(wrapper.findItems()[0].findContent()!.getElement()).toHaveTextContent('Item 1');
+      });
+    });
+
+    it('handles mixed persistent and non-persistent items', async () => {
+      mockRetrieveFlashbarDismiss
+        .mockResolvedValueOnce(true) // key-0: hidden
+        .mockResolvedValueOnce(false); // key-2: visible
```
just-boris: It would be better to have a more deterministic mock implementation

```js
mockRetrieveFlashbarDismiss = (key) => {
   if(key === 'key-0') {
      return true;
   }
   if(key === 'key-2') {
       return false;
   }
   throw new Error(`Unknown key ${key}`);
}
```

src/flashbar/common.tsx
```diff
@@ -140,3 +144,48 @@ export function useFlashbar({
     handleFlashDismissed,
   };
 }
+
+// Hook for managing flashbar items visibility with persistence
+export function useFlashbarVisibility(items: ReadonlyArray<FlashbarProps.MessageDefinition>) {
+  const [checkedPersistenceKeys, setCheckedPersistenceKeys] = useState<Set<string>>(new Set());
+  const [persistentItemsVisibility, setPersistentItemsVisibility] = useState<Map<string, boolean>>(new Map());
+
+  const visibleItems = useMemo(() => {
```
just-boris: This `useMemo` does nothing, because most teams pass `items` as an inline array, which invalidates memo value on every render

src/flashbar/common.tsx
```diff
@@ -140,3 +144,48 @@ export function useFlashbar({
     handleFlashDismissed,
   };
 }
+
+// Hook for managing flashbar items visibility with persistence
+export function useFlashbarVisibility(items: ReadonlyArray<FlashbarProps.MessageDefinition>) {
+  const [checkedPersistenceKeys, setCheckedPersistenceKeys] = useState<Set<string>>(new Set());
```
just-boris: Need to create the set instance only once instead of every render

```suggestion
  const [checkedPersistenceKeys, setCheckedPersistenceKeys] = useState<Set<string>>(() => new Set());
```

Same for the `new Map` below

src/flashbar/__tests__/persistence-unit.test.tsx
```diff
@@ -0,0 +1,200 @@
...
+        expect(mockRetrieveFlashbarDismiss).toHaveBeenCalledTimes(1);
+      });
+
+      rerender(<Flashbar items={items} />);
+      await waitFor(() => {
+        expect(mockRetrieveFlashbarDismiss).toHaveBeenCalledTimes(1);
+      });
+    });
+
+    it('checks persistence for new items only', async () => {
```
just-boris, line 197: ⛳  This misses one more use-case

1. Start with an empty array
2. Add one item, check that `mockRetrieveFlashbarDismiss` was called with that item id
3. Add one more item, check that `mockRetrieveFlashbarDismiss` was called with only the 2nd item id

My prediction - this test will fail and there is a bug in the code

src/flashbar/common.tsx
```diff
@@ -140,3 +144,48 @@ export function useFlashbar({
...
+
+      await Promise.all(
+        newPersistentItems.map(async item => {
+          const isDismissed = await retrieveFlashbarDismiss(item.persistenceConfig!);
+          newVisibility.set(item.persistenceConfig!.uniqueKey, !isDismissed);
+          newKeys.add(item.persistenceConfig!.uniqueKey);
+        })
+      );
+
+      setPersistentItemsVisibility(newVisibility);
```
just-boris: ⛳  What happens of items list changes multiple times and there are several async `checkNewPersistentItems` calls happening in parallel? They will overwrite each other results.

It would be good to have a unit test on this as well

src/internal/persistence/index.ts
```diff
@@ -0,0 +1,58 @@
...
+): Promise<void> {
+  // No-op
+};
+
+export let retrieveFlashbarDismiss = async function (
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  persistenceConfig: FlashbarProps.PersistenceConfig
+): Promise<boolean> {
+  const result = await new Promise<boolean>(resolve => resolve(false));
+  return result;
```
just-boris: Same code but shorter

```suggestion
  return Promise.resolve(false);
```

src/internal/persistence/index.ts
```diff
@@ -0,0 +1,58 @@
...
+};
+
+export let persistAlertDismiss = async function (
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  persistenceConfig: AlertProps.PersistenceConfig
+): Promise<void> {
+  // No-op
+};
```
just-boris, line 52: Call it a personal preference but I think this is more readable 

```js
export let persistAlertDismiss = () => Promise.resolve()
```

than this

```js
export let persistAlertDismiss = async () => {/* no-op, but do not remove async keyword or everything breaks */} 
```

pages/flashbar/persistence.page.tsx
YueyingLu: Updated to AppContext.

src/alert/__tests__/persistence-unit.test.tsx
YueyingLu: ✅

src/alert/internal.tsx
YueyingLu: My idea was to forbid " ". Removed trim().

src/alert/internal.tsx
YueyingLu: ✅

src/flashbar/__tests__/persistence-unit.test.tsx
YueyingLu: ✅

src/flashbar/common.tsx
YueyingLu: I think it is harmless to keep it, it will benefit for teams use static/stable items.

src/flashbar/common.tsx
YueyingLu: ✅

src/flashbar/__tests__/persistence-unit.test.tsx
YueyingLu: Added the test `checks persistence only for newly added items when start with empty array`

src/flashbar/common.tsx
YueyingLu: Updated the code and added unit test `handles concurrent async calls without race conditions`

src/internal/persistence/index.ts
YueyingLu: ✅

src/internal/persistence/index.ts
YueyingLu: ✅

src/flashbar/__tests__/persistence-unit.test.tsx
```diff
@@ -0,0 +1,282 @@
...
+      // Add third item while first item still loading
+      const updatedItems = [
+        ...initialItems,
+        { ...createItems([true])[0], content: 'Item 2', id: 'item-second', persistenceConfig: { uniqueKey: 'key-2' } },
+      ];
+      rerender(<Flashbar items={updatedItems} />);
+
+      // Resolve second call first (race condition scenario)
+      resolveSecond!(false); // Item 2 should be visible
+      await waitFor(() => {});
```
just-boris: Is this needed for anything?

src/flashbar/common.tsx
just-boris: This benefit is not measurable in any way. This is why typically I recommend to keep the code simple without extra things

just-boris (approved):
All blocking comments resolved, thanks

src/alert/internal.tsx
```diff
@@ -124,6 +129,26 @@ const InternalAlert = React.forwardRef(
       i18nStrings?.dismissAriaLabel ?? i18n('dismissAriaLabel', deprecatedDismissAriaLabel)
     );
 
+    useEffect(() => {
+      if (persistenceConfig?.uniqueKey) {
+        retrieveAlertDismiss(persistenceConfig).then(dismissed => {
+          setIsPersistentlyDismissed(!!dismissed);
+        });
+      }
+      // eslint-disable-next-line react-hooks/exhaustive-deps
```
gethinwebster, line 152: why? what if `persistenceConfig.crossServicePersistence` changes?

src/alert/internal.tsx
```diff
@@ -124,6 +129,26 @@ const InternalAlert = React.forwardRef(
...
+    }, [persistenceConfig?.uniqueKey]);
+
+    const dismiss = () => {
+      fireNonCancelableEvent(onDismiss);
+      if (persistenceConfig && persistenceConfig.uniqueKey) {
```
gethinwebster: [minor] for consistency
```suggestion
      if (persistenceConfig?.uniqueKey) {
```

src/alert/internal.tsx
```diff
@@ -100,6 +102,9 @@ const InternalAlert = React.forwardRef(
     const containerRef = useMergeRefs(containerMeasureRef, __internalRootRef);
     const headerRef = useMergeRefs(headerRefAction, headerRefContent);
     const contentRef = useMergeRefs(contentRefAction, contentRefContent);
+    const [isPersistentlyDismissed, setIsPersistentlyDismissed] = useState(
+      !!(persistenceConfig && persistenceConfig.uniqueKey)
```
gethinwebster, line 106: default-to-dismissed seems dangerous here, especially combined with promise without error handling in the `useEffect` below: if something within `retrieveAlertDismiss` fails then user will never see the alert.

At the very minimum the promise should have error handling, but I would strongly prefer an approach of default-to-visible

src/flashbar/common.tsx
```diff
@@ -140,3 +144,56 @@ export function useFlashbar({
...
+    const newPersistentItems = items.filter(
+      item => item.persistenceConfig?.uniqueKey && !checkedPersistenceKeys.has(item.persistenceConfig.uniqueKey)
+    );
+
+    if (newPersistentItems.length === 0) {
+      return;
+    }
+
+    const checkNewPersistentItems = async () => {
+      const results = await Promise.all(
```
gethinwebster: needs some error handling around `await`s

src/flashbar/common.tsx
```diff
@@ -140,3 +144,56 @@ export function useFlashbar({
...
+  const [persistentItemsVisibility, setPersistentItemsVisibility] = useState<Map<string, boolean>>(() => new Map());
+
+  const visibleItems = useMemo(() => {
+    return items.filter(
+      item =>
+        !item.persistenceConfig?.uniqueKey || persistentItemsVisibility.get(item.persistenceConfig.uniqueKey) === true
+    );
+  }, [items, persistentItemsVisibility]);
+
+  useEffect(() => {
```
gethinwebster, line 158: `useEffect` with async code needs cleanup function

src/alert/internal.tsx
```diff
@@ -124,6 +129,26 @@ const InternalAlert = React.forwardRef(
       i18nStrings?.dismissAriaLabel ?? i18n('dismissAriaLabel', deprecatedDismissAriaLabel)
     );
 
+    useEffect(() => {
```
gethinwebster, line 132: `useEffect` with async code needs cleanup function

src/internal/persistence/index.ts
```diff
@@ -0,0 +1,60 @@
...
+  }
+  if (functions.persistAlertDismiss) {
+    persistAlertDismiss = functions.persistAlertDismiss;
+  }
+  if (functions.retrieveAlertDismiss) {
+    retrieveAlertDismiss = functions.retrieveAlertDismiss;
+  }
+}
+
+// eslint-disable-next-line require-await
```
gethinwebster, line 30: [minor] you could probably avoid these with `return await Promise.resolve();`?

src/internal/persistence/index.ts
```diff
@@ -0,0 +1,60 @@
...
+import { FlashbarProps } from '../../flashbar/interfaces';
+
+interface PersistenceFunction {
+  persistFlashbarDismiss?: (persistenceConfig: FlashbarProps.PersistenceConfig) => Promise<void>;
+  retrieveFlashbarDismiss?: (persistenceConfig: FlashbarProps.PersistenceConfig) => Promise<boolean>;
+  persistAlertDismiss?: (persistenceConfig: AlertProps.PersistenceConfig) => Promise<void>;
+  retrieveAlertDismiss?: (persistenceConfig: AlertProps.PersistenceConfig) => Promise<boolean>;
+}
+
+export function setPersistenceFunctionsForTesting(functions: PersistenceFunction) {
```
gethinwebster, line 15: why not just use mocking?

src/flashbar/common.tsx
```diff
@@ -140,3 +144,56 @@ export function useFlashbar({
     handleFlashDismissed,
   };
 }
+
+// Hook for managing flashbar items visibility with persistence
+export function useFlashbarVisibility(items: ReadonlyArray<FlashbarProps.MessageDefinition>) {
```
gethinwebster, line 149: could we move the complexity of this into the plugin rather than in the component itself? so just have a single call to `retrieveFlashbarDismiss(items)` which then internally handles which ones need to be fetched and which not?

src/flashbar/__tests__/persistence-unit.test.tsx
YueyingLu: Removed

src/alert/internal.tsx
YueyingLu: It is less likely but you are right, I add it.

src/alert/internal.tsx
YueyingLu: `retrieveAlertDismiss` is default to show as visible in the plugin: when the call fail or when timeout. Also added try catch in the component code, to default to visible. 

If the state is visible at the beginning, it will cause flickering -- item show first then disappear after the `retrieveAlertDismiss` finishes.

src/flashbar/common.tsx
YueyingLu: added

src/flashbar/common.tsx
YueyingLu: added

src/alert/internal.tsx
YueyingLu: added

src/internal/persistence/index.ts
YueyingLu: It gets error ` Redundant use of await on a return value `

src/internal/persistence/index.ts
YueyingLu: In the unit test it uses mocking. Here is to test in the browser.

gethinwebster: I don't understand, what is the use-case for overriding these in a browser?

src/flashbar/common.tsx
YueyingLu: 1. Mostly because of testing limitation in the plugin package. Plugin package is used to export utilities, simple input/output functions.  `useFlashbarVisibility` has more complex logic that I want to test in integration test which is not achievable in plugin package. 
2. We may want to support persistence in open source in the future. This code can be useful in component.

src/alert/internal.tsx
```diff
@@ -124,6 +129,43 @@ const InternalAlert = React.forwardRef(
...
+      let isMounted = true;
+      if (persistenceConfig?.uniqueKey) {
+        retrieveAlertDismiss(persistenceConfig)
+          .then(dismissed => {
+            if (isMounted) {
+              try {
```
gethinwebster: Why is this try-catch here? I don't understand why the line below would ever fail. If there's a good reason for it I think it deserves an explanatory comment

src/alert/internal.tsx
gethinwebster: can we do it in a way that don't have to disable the eslint rule? that way we're safer if adjusting this code in the future.

gethinwebster: and if we _do_ disable it, I think a comment is needed to explain why for clarity

src/flashbar/common.tsx
gethinwebster: 1. What exactly are the limitations in terms of testing? It seems like that would be good to resolve regardless.
2. For me, this is even more reason to move it separately: an open source persistence implementation might want a different strategy for loading the persistence status (e.g. maybe a bulk fetch would be supported), so I think by moving it elsewhere we give ourselves much more flexibility in the future.

src/alert/internal.tsx
YueyingLu: Was leftover from development, removed.

src/alert/internal.tsx
YueyingLu: Added comment.

src/internal/persistence/index.ts
YueyingLu: Discussed offline, it is for integ and for manual testing like https://d21d5uik3ws71m.cloudfront.net/components/2eba0b0f43b1f659f0e116743ca2205efbafd685/dev-pages-react16/index.html#/light/flashbar/persistence?mockRetrieveDelay=100

src/flashbar/common.tsx
YueyingLu: Resolved offline.

PR #4038, opened 2025-11-14
fix: select, multiselect option grouping

SpyZzey:
### Description

This PR fixes a bug that groups "root"-level options in a select and multiselect dropdown to the previous group. The issue was non-visual, but leads to issues with accessibility screen readers.

- Added Tests
- Removed a condition in "unflatten-options" that wrongly grouped "non"-child elements to the previous group.

Related Issues: [AWSUI-61453]

### How has this been tested?
- Added unit tests to test the unflattenOptions function.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/components/option/utils/__tests__/unflatten-options.test.ts
+++ b/src/internal/components/option/utils/__tests__/unflatten-options.test.ts
@@ -1,0 +1,87 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import { DropdownOption } from '../../interfaces';
+import { unflattenOptions } from '../unflatten-options';
+
+describe('unflattenOptions', () => {
+  test('should handle parent with children', () => {
+    const options: DropdownOption[] = [
+      { type: 'parent', option: { label: 'Group 1' } },
+      { type: 'child', option: { label: 'Child 1' } },
+      { type: 'child', option: { label: 'Child 2' } },
+    ];
+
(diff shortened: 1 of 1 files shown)


src/internal/components/option/utils/unflatten-options.ts
```diff
@@ -27,7 +27,7 @@ export function unflattenOptions(options: ReadonlyArray<DropdownOption>): Nested
       const wrapped: ParentDropdownOption = { type: 'parent', option, index, children: [] };
       currentParent = wrapped;
       nestedOptions.push(wrapped);
-    } else if (!option.type || option.type === 'child') {
+    } else if (option.type === 'child') {
```
pan-kot, line 30: What is the case for the `option.type` to be undefined?

jperals: Assuming that root-level non-parent nodes have `option.type` set to `undefined`, I think this can be on the right direction, and simpler than [my approach](https://github.com/cloudscape-design/components/pull/4037/files).

Could you add some testing coverage?

SpyZzey: > What is the case for the `option.type` to be undefined?

@pan-kot 
In the flatten-options.ts, the option.type = "child" will be set, if the option is inside the options array of another parent-option (e.g. groups). The parent-option will get the option.type = "parent" set. 

@jperals Exactly. I will add some test cases.

jperals (approved):

PR #4039, opened 2025-11-14
fix: Initialize property filter string value to empty string

jperals:
### Description

Ticket: `AWSUI-61464`
Dry run: `7600116220`

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/property-filter/token-editor.tsx
+++ b/src/property-filter/token-editor.tsx
@@ -105,9 +105,7 @@ export function TokenEditor({
           ? temporaryToken.operator
           : allowedOperators[0];
       const matchedProperty = filteringProperties.find(property => property.propertyKey === newPropertyKey) ?? null;
-      const isCustomType = !!matchedProperty?.externalProperty.operators?.find(
-        o => typeof o !== 'string' && o.operator === operator && !!o.form
-      );
+      const isCustomType = !!matchedProperty?.getValueFormRenderer(operator);
       const value = isCustomType || matchedProperty?.getTokenType(operator) === 'enum' ? null : '';
       setTemporaryToken({ ...temporaryToken, property: matchedProperty, operator, value });
     };


src/property-filter/__tests__/property-filter-token-editor.test.tsx
```diff
@@ -501,7 +531,7 @@ describe('token editor with groups', () => {
         expect.objectContaining({
           detail: {
             operation: 'and',
-            tokenGroups: [{ propertyKey: 'other-string', operator: '=', value: null }],
+            tokenGroups: [{ propertyKey: 'other-string', operator: '=', value: '' }],
```
pan-kot, line 568: This fixes the issue for string properties, but breaks enum and custom properties, see: https://github.com/cloudscape-design/components/pull/4041

pan-kot (changes requested):

src/property-filter/__tests__/property-filter-token-editor.test.tsx
jperals: Thanks. Changed the logic based on the expected behavior of those tests (empty array for enum and null for custom properties). Also refactored into one test for each.

I rely on the presence of a form property in order to infer a custom property (see `token-editor.tsx` below? Is there a better way? Nor the internal `operator`, which is a string, nor the `matchedProperty`, which is of type `InternalFilteringProperty`, have enough information about this, and the  `PropertyFilterTokenType` type returned by `property.getTokenType` only knows about `value` or `enum`.

pan-kot: Yes, relying on the form sounds good 👍

src/property-filter/token-editor.tsx
```diff
@@ -105,7 +105,11 @@ export function TokenEditor({
           ? temporaryToken.operator
           : allowedOperators[0];
       const matchedProperty = filteringProperties.find(property => property.propertyKey === newPropertyKey) ?? null;
-      setTemporaryToken({ ...temporaryToken, property: matchedProperty, operator, value: null });
+      const isCustomType = !!matchedProperty?.externalProperty.operators?.find(
```
pan-kot: The matchedProperty is of internal property type. It supports derived properties for better ergonomics, so there is no need to find the form like this. Instead, you can either check for `getValueFormRenderer(operator)` prop, or even extend the getTokenType() to return "value" | "enum" | "custom".

jperals: OK, thanks. Updated to use `getValueFormRenderer`.

pan-kot (approved):

PR #4042, opened 2025-11-17
chore: Reusing arrow styles between popover and annotation context

pan-kot:
### Description

We plan to reuse the blue, annotation context-style popover in an emerging notifications pattern. To avoid copying the popover arrow styles again, this PR consolidates the arrow styles inside the popover instead. There were a few very minor diff between popover/arrow.scss and annotation-context/annotation/arrow.scss, I will run screenshot tests to ensure we are safe.

### How has this been tested?

* Screenshot tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/popover/arrow.scss
```diff
@@ -22,8 +22,8 @@
     box-shadow: ($box-shadow-size * $x) ($box-shadow-size * $y) 4px -2px awsui.$color-shadow-default;
   }
 
-  $arrow-edge-length: 14px;
-  $arrow-width: 20px;
+  $arrow-edge-length: 1.4 * styles.$base-size;
```
georgylobko, line 25: Can we avoid using magical numbers here?

src/popover/arrow.tsx
georgylobko, line 1: [non-blocking] Does it make sense to turn this into an internal component and move to `src/internal` folder as it's used not only for popover component

src/popover/arrow.scss
pan-kot: I'd keep it as is until we plan to make any changes to this code driven by the design decisions.

This version comes from the existing annotation context implementation: https://github.com/cloudscape-design/components/pull/4042/files#diff-2b21cd8ac3d888e1d5d19e7dcc64ecdac343066dcd107380fe94026378546b3fL25

src/popover/arrow.tsx
pan-kot: Potentially, yes. However, it is a bigger scope as the popover container is also separately used by the chart popover and tooltip internal components. I think we can refactor this one when working on the public tooltip component. Wdyt?

georgylobko: Up to you, this is non-blocking

georgylobko (approved):

PR #4050, opened 2025-11-19
feat: implement style api for text filter component

mxschll:
### Description

Extends the text filter component with a style API for customization, allowing developers to override default component styles.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/text-filter/style-permutations.page.tsx
+++ b/pages/text-filter/style-permutations.page.tsx
@@ -11,41 +11,6 @@ import ScreenshotArea from '../utils/screenshot-area';
 const style1 = {
   root: {
     borderColor: {
-      default: '#ef4444',
-      hover: '#b91c1c',
-      focus: '#dc2626',
-      disabled: '#fca5a5',
-      readonly: '#fca5a5',
-    },
-    borderWidth: '2px',
-    borderRadius: '8px',
-    backgroundColor: {
-      default: '#fecaca',
-      hover: '#fca5a5',
-      focus: '#fca5a5',
-      disabled: '#fca5a5',
-      readonly: '#fee2e2',
-    },
-    boxShadow: {
-      default: '0 1px 2px rgba(0, 0, 0, 0.05)',
-      hover: '0 2px 4px rgba(239, 68, 68, 0.15)',
-      focus: '0 0 0 4px rgba(239, 68, 68, 0.2), 0 2px 4px rgba(239, 68, 68, 0.15)',
-      disabled: 'none',
-      readonly: '0 1px 2px rgba(0, 0, 0, 0.05)',
-    },
-    color: {
-      default: '#7f1d1d',
-      hover: '#7f1d1d',
-      focus: '#7f1d1d',
-      disabled: '#7f1d1d',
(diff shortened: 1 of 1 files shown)


src/text-filter/styles.tsx
```diff
@@ -0,0 +1,46 @@
...
+      [customCssProps.styleColorHover]: style?.root?.color?.hover,
+      [customCssProps.styleColorFocus]: style?.root?.color?.focus,
+      [customCssProps.styleColorReadonly]: style?.root?.color?.readonly,
+      [customCssProps.stylePlaceholderColor]: style?.placeholder?.color,
+      [customCssProps.stylePlaceholderFontSize]: style?.placeholder?.fontSize,
+      [customCssProps.stylePlaceholderFontWeight]: style?.placeholder?.fontWeight,
+      [customCssProps.stylePlaceholderFontStyle]: style?.placeholder?.fontStyle,
+    };
+
+    return properties;
```
cansuaa, line 44: I see slightly different implementations in styles.tsx files. E.g. in prompt input we filter out the undefined values, why aren't we doing the same here?

src/text-filter/internal.tsx
```diff
@@ -79,6 +80,7 @@ const InternalTextFilter = React.forwardRef(
           clearAriaLabel={filteringClearAriaLabel}
           onChange={event => fireNonCancelableEvent(onChange, { filteringText: event.detail.value })}
           __onDelayedInput={event => fireNonCancelableEvent(onDelayedChange, { filteringText: event.detail.value })}
+          style={style}
```
cansuaa, line 83: `InternalInput` has its own style, in its [styles.tsx](https://github.com/cloudscape-design/components/blob/main/src/input/styles.tsx#L10), it checks if style.root exists which in that case if that doesn't exist but placeholder does, the placeholder styles won't be applied. 
Tbh I'm not even sure if InternalInput is supposed to check for this given it also has styles.placeholder

src/text-filter/styles.tsx
mxschll: Yea because of this comment: https://github.com/cloudscape-design/components/pull/4025#discussion_r2523574346

src/text-filter/internal.tsx
mxschll: I agree. I wanted to refactor the `styles.tsx` for existing components to be consistent, but I'll do it for the input in this PR because it's necessary as you pointed out.

mxschll: Adressed

src/text-filter/styles.tsx
cansuaa: I see, then we should remove that logic from prompt input as well

mxschll: Yes, I'll open a subsequent PR for that if that's ok?

cansuaa: Yes, also in the PR linked I don't see matchSnapshot being used but it is used here. Why are there different implementations? Let's have consistency across these similar implementations

PR #4054, opened 2025-11-19
feat: implement style api for autosuggest component

mxschll:
### Description

Extends the autosuggest component with a style API for customization, allowing developers to override default component styles.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/input/styles.scss
+++ b/src/input/styles.scss
@@ -130,7 +130,8 @@
   &.input-invalid {
     @include styles.form-invalid-control(
       $color: var(#{custom-props.$styleColorDefault}, awsui.$color-text-status-error),
-      $border-color: var(#{custom-props.$styleBorderColorDefault}, awsui.$color-text-status-error)
+      $border-color: var(#{custom-props.$styleBorderColorDefault}, awsui.$color-text-status-error),
+      $focus-box-shadow: var(#{custom-props.$styleBoxShadowFocus}, foundation.$box-shadow-focused-light-invalid)
     );
     &.input-has-icon-left {
       padding-inline-start: calc(
@@ -143,7 +144,8 @@
   &.input-warning {
     @include styles.form-warning-control(
       $color: var(#{custom-props.$styleColorDefault}, awsui.$color-text-status-warning),
-      $border-color: var(#{custom-props.$styleBorderColorDefault}, awsui.$color-text-status-warning)
+      $border-color: var(#{custom-props.$styleBorderColorDefault}, awsui.$color-text-status-warning),
+      $focus-box-shadow: var(#{custom-props.$styleBoxShadowFocus}, foundation.$box-shadow-focused-light-invalid)
     );
     &.input-has-icon-left {
       padding-inline-start: calc(
--- a/src/internal/styles/forms/mixins.scss
+++ b/src/internal/styles/forms/mixins.scss
@@ -193,24 +193,32 @@
   @include typography.font-body-s;
 }
 
-@mixin form-invalid-control($color: awsui.$color-text-status-error, $border-color: awsui.$color-text-status-error) {
+@mixin form-invalid-control(
+  $color: awsui.$color-text-status-error,
+  $border-color: awsui.$color-text-status-error,
+  $focus-box-shadow: foundation.$box-shadow-focused-light-invalid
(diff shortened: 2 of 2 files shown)


src/autosuggest/interfaces.ts
```diff
@@ -160,6 +165,51 @@ export namespace AutosuggestProps {
      */
     select(): void;
   }
+
+  export interface Style {
+    root?: {
```
pan-kot, line 170: Should we also support styling for the clear input button?

Image

src/autosuggest/interfaces.ts
```diff
@@ -160,6 +165,51 @@ export namespace AutosuggestProps {
...
+      borderColor?: {
+        default?: string;
+        disabled?: string;
+        focus?: string;
+        hover?: string;
+        readonly?: string;
+      };
+      borderRadius?: string;
+      borderWidth?: string;
+      boxShadow?: {
```
pan-kot, line 187: The box shadow can overlay the focus outline, but that only works for certain designs. Would it make sense to add the capability to style the focus outline directly?

Image

pages/autosuggest/style-permutations.page.tsx
```diff
@@ -0,0 +1,125 @@
...
+};
+
+const enteredTextLabel = (value: string) => `Use: ${value}`;
+
+const permutations = createPermutations<AutosuggestProps>([
+  {
+    value: ['This is a test value'],
+    placeholder: [''],
+    disabled: [false, true],
+    readOnly: [false, true],
```
pan-kot: This creates more combinations than needed: 2x disabled, 2x readOnly, 2x invalid, and 2x style give 16 combinations, out of which 8 feature disabled=true and readOnly=true that are of little interest.

Besides, there are no permutations that show placeholder value or warning state.

I recommend declaring more atomic permutation groups, e.g.:

```
[
  { value: ['value'], disabled: [false, true], invalid: [false, true], warning: [false, true], style: [style] },
  { placeholder: ['Placeholder'], disabled: [false, true], invalid: [false, true], warning: [false, true], style: [style] },
  { value: ['value'], readOnly: [false, true], invalid: [false, true], warning: [false, true], style: [style] },
  { placeholder: ['Placeholder'], readOnly: [false, true], invalid: [false, true], warning: [false, true], style: [style] },
]
```

The same can be done twice if that is really necessary to cover multiple styles. Alternatively, we can be a little creative here and include a style selector component instead, also controlled with an url flag. Thus, opening a page with ?style=style1 will render one style, and using ?style=style2 will show an alternative style (provided there is actual value in having more than one).

src/autosuggest/interfaces.ts
mxschll: Yeah I think so. I also think it should be possible to style the left icon. But the same applies to the regular input + search variant. I'd suggest I write a proposal for that, what do you think?

pages/autosuggest/style-permutations.page.tsx
mxschll: What do you think about this approach at `/textarea/pseudo-selectors`? That way we could interact with it in the screenshot tests (like hover, focus) more easily?

src/autosuggest/interfaces.ts
mxschll: Actually, that's the invalid + focus state box shadow, which is the same discussion as before: how to support intersecting states. I changed it to fall back to the custom focus box shadow, which makes most sense.

pan-kot: Alright, so it is a limitation of the current API then.

mxschll: I am collecting all the cases :) But falling back to the custom focus style allows for switching out the style object.

pages/autosuggest/style-permutations.page.tsx
pan-kot: I think the current approach with a single style is ok for now. The highly dynamic approach is good for manual testing or integration tests, but not so much for permutations. There, we can introduce maybe a single control to select out of a few available styles, while keeping the rest of the states predefined so to ensure good coverage with min configuration.

PR #4064, opened 2025-11-21
fix: prevent list item jump on hover in virtual scrolling (v2)

mxschll:
**Note:** This PR supersedes [PR#3943](https://github.com/cloudscape-design/components/pull/3943) and includes an additional fix. The original implementation used `globalIndex` for the position offset calculation, which caused whitespace issues when scrolling through large lists:

```ts
// ./src/select/utils/render-options.tsx
const adjustedVirtualPosition = virtualItem ? virtualItem.start - globalIndex : undefined;
```

### Description

With virtual scrolling enabled, selectable items would "jump" on hover/selection due to border width changes (AWSUI-61303). Virtual scrolling positions items absolutely, so when border width changes from 1px to 2px on hover, the browser recalculates layout dimensions causing visible position shifts.

Additionally, option groups were missing the separating top border.

|  Before | After | 
|---|---|
| 6ab8eb377aa9c0 | 65b1c41442f210 | 

The solution replaces borders and outlines with box-shadows to simulate borders while maintaining thin actual borders.

In summary: 

- Eliminates jumping behavior with virtual scrolling enabled
- Maintains visual consistency with non-virtual scrolling
- Adds top separator to option groups

### How has this been tested?

Changes ran through screenshot test and have been verified locally.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/components/selectable-item/styles.scss
+++ b/src/internal/components/selectable-item/styles.scss
@@ -265,14 +265,14 @@ $neg-divider-w: calc(-1 * #{$divider-w});
       box-shadow:
         inset 0 $divider-w 0 0 awsui.$color-border-dropdown-item-default,
         inset 0 $neg-divider-w 0 0 awsui.$color-border-dropdown-item-default,
-        inset awsui.$border-item-width 0 0 0 awsui.$color-border-dropdown-container;
+        inset awsui.$border-width-popover 0 0 0 awsui.$color-border-dropdown-container;
 
       &:not(.with-scrollbar) {
         box-shadow:
           inset 0 $divider-w 0 0 awsui.$color-border-dropdown-item-default,
           inset 0 $neg-divider-w 0 0 awsui.$color-border-dropdown-item-default,
-          inset awsui.$border-item-width 0 0 0 awsui.$color-border-dropdown-container,
-          inset calc(-1 * #{awsui.$border-item-width}) 0 0 0 awsui.$color-border-dropdown-container;
+          inset awsui.$border-width-popover 0 0 0 awsui.$color-border-dropdown-container,
+          inset calc(-1 * #{awsui.$border-width-popover}) 0 0 0 awsui.$color-border-dropdown-container;
       }
 
       // Only display top border when not being rendered after the filter
@@ -286,13 +286,13 @@ $neg-divider-w: calc(-1 * #{$divider-w});
         box-shadow:
           inset 0 $divider-w 0 0 awsui.$color-background-dropdown-item-default,
           inset 0 $neg-divider-w 0 0 awsui.$color-border-dropdown-item-default,
-          inset awsui.$border-item-width 0 0 0 awsui.$color-border-dropdown-container;
+          inset awsui.$border-width-popover 0 0 0 awsui.$color-border-dropdown-container;
         &:not(.with-scrollbar) {
           box-shadow:
             inset 0 $divider-w 0 0 awsui.$color-background-dropdown-item-default,
             inset 0 $neg-divider-w 0 0 awsui.$color-border-dropdown-item-default,
-            inset awsui.$border-item-width 0 0 0 awsui.$color-border-dropdown-container,
-            inset calc(-1 * #{awsui.$border-item-width}) 0 0 0 awsui.$color-border-dropdown-container;
+            inset awsui.$border-width-popover 0 0 0 awsui.$color-border-dropdown-container,
+            inset calc(-1 * #{awsui.$border-width-popover}) 0 0 0 awsui.$color-border-dropdown-container;
         }
       }
     }
--- a/src/internal/hooks/use-virtual/index.ts
+++ b/src/internal/hooks/use-virtual/index.ts
@@ -61,19 +61,17 @@ export function useVirtual<Item extends object>({
 
   const virtualItems = useMemo(
     () =>
-      rowVirtualizer.virtualItems.map(virtualItem => {
-        return {
-          ...virtualItem,
-          start: virtualItem.start,
-          measureRef: (node: null | HTMLElement) => {
-            const mountedCount = measuresCache.current.get(items[virtualItem.index]) ?? 0;
-            if (mountedCount < MAX_ITEM_MOUNTS) {
-              virtualItem.measureRef(node);
-              measuresCache.current.set(items[virtualItem.index], mountedCount + 1);
-            }
-          },
-        };
-      }),
+      rowVirtualizer.virtualItems.map(virtualItem => ({
+        ...virtualItem,
+        start: virtualItem.start,
+        measureRef: (node: null | HTMLElement) => {
+          const mountedCount = measuresCache.current.get(items[virtualItem.index]) ?? 0;
+          if (mountedCount < MAX_ITEM_MOUNTS) {
+            virtualItem.measureRef(node);
+            measuresCache.current.set(items[virtualItem.index], mountedCount + 1);
+          }
+        },
+      })),
     [items, rowVirtualizer.virtualItems]
   );
 


mxschll:
> The changes still have two bugs. In the Autosuggest, select-all still shifts when hovering. https://d21d5uik3ws71m.cloudfront.net/components/c8deadcd58165722038f814907d79a25e8edca8f/dev-pages-react18/index.html#/light/multiselect/select-all/?virtualScroll=true&closeAfter=false
> 
> Also, the checkbox in multiselect is shifting slightly when selecting: https://d21d5uik3ws71m.cloudfront.net/components/c8deadcd58165722038f814907d79a25e8edca8f/dev-pages-react18/index.html#/light/multiselect/screenshot

Addressed in new revision.

pages/select/virtual-scroll.page.tsx
```diff
@@ -0,0 +1,47 @@
...
+const options: SelectProps.Options = Array.from({ length: 1000 }, (_, i) => ({
+  value: `${i}`,
+  label: `Option ${i + 1}`,
+}));
+
+export default function () {
+  const [selected, setSelected] = useState<SelectProps['selectedOption']>(null);
+
+  return (
+    <>
```
pan-kot: nit: you can use the SimplePage template helper to create playground pages. The helper does even come with a built-in screenshot-area, but it does not allow to pass custom styles to it, so probably that part can be kept as it.

pages/select/virtual-scroll.page.tsx
```diff
@@ -0,0 +1,47 @@
...
+      <h1>Virtual Scroll</h1>
+
+      <ScreenshotArea
+        style={{
+          height: 500,
+          // Prevents dropdown from expanding outside of the screenshot area
+          overflow: 'auto',
+        }}
+      >
+        <SpaceBetween size="s">
```
pan-kot: This space-between is not needed

pages/select/virtual-scroll.page.tsx
```diff
@@ -0,0 +1,47 @@
...
+          <Select
+            placeholder="Demo with manual filtering"
```
pan-kot: nit: this is technically a different demo, so we might want to change the placeholder text accordingly (I recommend using a better generic one)

src/select/utils/render-options.tsx
```diff
@@ -65,11 +65,14 @@ export const renderOptions = ({
     const ListItem = useInteractiveGroups ? MultiselectItem : Item;
     const isSticky = firstOptionSticky && globalIndex === 0;
 
+    // Adjust virtual position to create 1px overlap between consecutive selected items in multiselect
+    const adjustedVirtualPosition = virtualItem ? virtualItem.start - index : undefined;
```
pan-kot: Should we have a multiselect test page, too?

Potentially, we can create some top-level "virtual-scroll" test page that can feature both select and multiselect, or even additionally include autosuggest and property filter.

pan-kot: nit: I would change the formula to `virtualItem.start - (index * 1)` to emphasise the relation between the index and the number of pixels to adjust the position for (it took me some while to understand why do we subtract indices from position).

src/select/utils/render-options.tsx
```diff
@@ -65,11 +65,14 @@ export const renderOptions = ({
     const ListItem = useInteractiveGroups ? MultiselectItem : Item;
     const isSticky = firstOptionSticky && globalIndex === 0;
 
+    // Adjust virtual position to create 1px overlap between consecutive selected items in multiselect
```
pan-kot: This is true for both select and multiselect - isn't it?

src/internal/components/selectable-item/styles.scss
```diff
@@ -206,6 +249,101 @@
     &:first-of-type:not(.selected, .highlighted) {
       border-block-start-color: awsui.$color-border-dropdown-item-top;
     }
+
+    // When using virtual scrolling, use box-shadows to mimic changing border
+    // width and set the actual border to a fixed width to prevent jumping
+    // behaviour when hovering selectable items
+
+    $virtual-border-offset: calc(#{awsui.$border-item-width} - #{awsui.$border-divider-list-width});
```
pan-kot: should we declare these constants in the top of the file and then reuse for sticky items, too?

pan-kot: The core comment from above can be attributed to these constants, too - so that applicable to both sticky and non-sticky branches.

src/internal/components/selectable-item/styles.scss
```diff
@@ -151,6 +151,49 @@
...
+      // This selector exists just to enable the next rule
+    }
+    &:has(~ .virtual) {
+      $virtual-border-offset: calc(#{awsui.$border-item-width} - #{awsui.$border-divider-list-width});
+
+      &.highlighted:not(.selected),
+      &.selected {
+        border-width: awsui.$border-divider-list-width;
+        padding-block: calc(#{styles.$option-padding-vertical} + #{$virtual-border-offset});
+        padding-inline: calc(#{styles.$control-padding-horizontal} + #{$virtual-border-offset});
```
pan-kot: why do we repeat styles from the non-sticky items?

src/select/parts/virtual-list.tsx
```diff
@@ -107,7 +107,10 @@ const VirtualListOpen = forwardRef(
           aria-hidden="true"
           key="total-size"
           className={styles['layout-strut']}
-          style={{ height: totalSize - stickySize }}
+          style={{
+            // Adjust totalSize to account for 1px overlap per item (matching the position adjustment in renderOptions)
+            height: totalSize - filteredOptions.length - stickySize,
```
pan-kot: Can we compute the correct totalSize inside the `useVirtual()`? That is a util that we own - we can theoretically parametrise it so that the total size and item sizes area adjusted automatically when we need that 1px overlap feature.

src/select/utils/render-options.tsx
mxschll: That's a good idea. We can probably add this generic page to the select or selectable item.

src/select/utils/render-options.tsx
mxschll: Yes, you're right. Applies to both.

src/internal/components/selectable-item/styles.scss
mxschll: We apply these styles to virtual elements, but sticky elements aren't virtual. Perhaps it would be better to merge `.virtual` and `.sticky`, but the linter didn't like my first approach. I'll try a different one…

src/select/parts/virtual-list.tsx
mxschll: Good call, I'll look into that.

mxschll: Changed in the latest revision.

pages/selectable-item/virtual-scroll.page.tsx
```diff
@@ -0,0 +1,98 @@
...
+}));
+
+export default function () {
+  const [selected, setSelected] = useState<SelectProps['selectedOption']>(null);
+  const [selectedMulti, setSelectedMulti] = useState<MultiselectProps.Options>(options.slice(0, 2));
+  const [selectedMultiWithSelectAll, setSelectedMultiWithSelectAll] = useState<MultiselectProps.Options>([]);
+  const [autosuggestValue, setAutosuggestValue] = useState('');
+
+  return (
+    <SimplePage title="Virtual Scroll" i18n={{}}>
```
pan-kot: If you need screenshot tests for this page, you need to also add `screenshotArea={{}}` here

pages/selectable-item/virtual-scroll.page.tsx
```diff
@@ -0,0 +1,98 @@
...
+      <div
+        style={{
+          height: 500,
```
pan-kot, line 45: I believe this height is no longer enough as there are more components now. I believe that for this to actually work as a screenshot testing page, these components probably need to be rendered one at a time, with some configuration (e.g. a segmented control above them)

src/internal/components/selectable-item/styles.scss
pan-kot: would it be possible to create a few simple mixins in this file so to share the common styles between sticky and virtual?

src/internal/hooks/use-virtual/index.ts
```diff
@@ -74,9 +77,25 @@ export function useVirtual<Item extends object>({
...
+    // resulting in a different total adjustment than the standard case
+    return firstItemSticky
+      ? rowVirtualizer.totalSize - items.length - stickySize + 2 // Compensate for the different positioning logic with sticky
+      : rowVirtualizer.totalSize - items.length - stickySize;
+  }, [applyItemOffset, firstItemSticky, virtualItems, rowVirtualizer.totalSize, items.length]);
+
   return {
     virtualItems,
     totalSize: rowVirtualizer.totalSize,
+    adjustedTotalSize,
```
pan-kot: Why do we need the non-adjusted total size?

src/internal/hooks/use-virtual/index.ts
```diff
@@ -74,9 +77,25 @@ export function useVirtual<Item extends object>({
     [items, rowVirtualizer.virtualItems]
   );
 
+  const adjustedTotalSize = useMemo(() => {
```
pan-kot: Why do we need memoization here?

src/internal/hooks/use-virtual/index.ts
```diff
@@ -16,11 +16,13 @@ interface UseVirtualProps<Item> {
   parentRef: React.RefObject<HTMLElement>;
   estimateSize: () => number;
   firstItemSticky?: boolean;
+  applyItemOffset?: boolean;
```
pan-kot: Should we call this item overlap instead of item offset? We can also consider using `itemOverlap: number` so that it is better explicit and better flexible.

src/internal/hooks/use-virtual/index.ts
```diff
@@ -41,6 +43,7 @@ export function useVirtual<Item extends object>({
   parentRef,
   estimateSize,
   firstItemSticky,
+  applyItemOffset = false,
```
pan-kot: is it possible to cover this with unit tests?

src/internal/hooks/use-virtual/index.ts
```diff
@@ -74,9 +77,25 @@ export function useVirtual<Item extends object>({
...
+
+    const stickySize = firstItemSticky && virtualItems.length > 0 ? virtualItems[0].size : 0;
+
+    // Adjust totalSize to account for 1px overlap per item (matching the position adjustment in renderOptions: /select/utils/render-options.tsx)
+    // When firstItemSticky is enabled, the select-all is shifted down by 1 and other items are shifted up by (index + 1),
+    // resulting in a different total adjustment than the standard case
+    return firstItemSticky
+      ? rowVirtualizer.totalSize - items.length - stickySize + 2 // Compensate for the different positioning logic with sticky
+      : rowVirtualizer.totalSize - items.length - stickySize;
+  }, [applyItemOffset, firstItemSticky, virtualItems, rowVirtualizer.totalSize, items.length]);
```
pan-kot: The computations above is rather hard to follow. Would it make sense to streamline those mb as follows:

```
const firstItemSize = virtualItems[0]?.size ?? 0;

let totalSize = rowVirtualizer.totalSize;

// Adjust total size to account for items overlap.
totalSize = totalSize - items.length * itemOverlap; // the itemOverlap defaults to 0

// Adjust total size to account for the sticky first item.
// We add a small offset after the sticky item because ...{explain why}
totalSize = firstItemSticky ? totalSize - firstItemSize + 2 : totalSize;
```

src/internal/hooks/use-virtual/index.ts
mxschll: We don't need it - I'll refactor it to be just total size.

src/internal/hooks/use-virtual/index.ts
mxschll: I thought it might be necessary since it's dependent on `rowVirtualizer` (same as above), but I'll remove

pages/selectable-item/virtual-scroll.page.tsx
```diff
@@ -0,0 +1,118 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
pan-kot: Why is this page nested under `selectable-item`?

pages/selectable-item/virtual-scroll.page.tsx
pan-kot: The segmented control looks nice, but the selection state must use query parameters (see how it is done in other pages). This allows to avoid unnecessary steps in integration tests (instead one can open the page as e.g. `/light/virtual-scroll?component=autosuggest`.

src/internal/components/selectable-item/styles.scss
```diff
@@ -6,6 +6,39 @@
 @use '../../styles' as styles;
 @use '../../styles/tokens' as awsui;
 
+// When using virtual scrolling, use box-shadows to mimic changing border width
+// and set the actual border to a fixed width to prevent jumping behaviour when
+// hovering selectable items
+$virtual-border-offset: calc(#{awsui.$border-item-width} - #{awsui.$border-divider-list-width});
+$virtual-border-offset-double: calc(2 * #{awsui.$border-item-width} - #{awsui.$border-divider-list-width});
+
+@mixin virtual-border-padding {
```
pan-kot: nice 👍

src/internal/hooks/use-virtual/__tests__/use-virtual.test.tsx
```diff
@@ -0,0 +1,133 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
pan-kot: nit: why is this test file called "use-virtual-offset" and not "use-virtual" to match the util name?

src/internal/hooks/use-virtual/__tests__/use-virtual-offset.test.tsx
```diff
@@ -0,0 +1,133 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import React, { useRef } from 'react';
+import { render } from '@testing-library/react';
+
+import { useVirtual } from '../index';
+
+// Mock the vendor react-virtual to avoid ES module issues
```
pan-kot: What does this comment mean - what are ES module issues we need to avoid?

src/internal/hooks/use-virtual/__tests__/use-virtual-offset.test.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+    let capturedResult: ReturnType<typeof useVirtual<TestItem>> | null = null;
+    render(<TestComponent items={items} itemOverlap={1} onResult={result => (capturedResult = result)} />);
+
+    expect(capturedResult).not.toBeNull();
+    // totalSize = 150 - (3 items * 1 overlap) = 147
+    expect(capturedResult!.totalSize).toBe(147);
+  });
+
+  test('adjusts totalSize with no itemOverlap', () => {
```
pan-kot: nit: the test name is not quite correct as no total size adjustment is needed. This can be renamed to "computes totalSize with itemOverlap=0" (and update the above test accordingly).

src/internal/hooks/use-virtual/__tests__/use-virtual.test.tsx
```diff
@@ -0,0 +1,133 @@
...
+        { index: 2, size: 50, start: 100, end: 150, measureRef: mockMeasureRef },
+      ],
+      totalSize: 150,
+      scrollToIndex: mockScrollToIndex,
+    });
+
+    let capturedResult: ReturnType<typeof useVirtual<TestItem>> | null = null;
+    render(<TestComponent items={items} itemOverlap={1} onResult={result => (capturedResult = result)} />);
+
+    expect(capturedResult).not.toBeNull();
```
pan-kot: nit: I don't think that this check makes any sense - the result is never meant to be null, but when it is - the test will fail on the next line anyways

src/internal/hooks/use-virtual/__tests__/use-virtual-offset.test.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+    let capturedResult: ReturnType<typeof useVirtual<TestItem>> | null = null;
+    render(<TestComponent items={items} onResult={result => (capturedResult = result)} />);
+
+    expect(capturedResult).not.toBeNull();
+    // totalSize = 150 - (3 items * 0 overlap) = 150
+    expect(capturedResult!.totalSize).toBe(150);
+  });
+
+  test('adjusts totalSize differently with firstItemSticky', () => {
```
pan-kot: nit: I'd be more precise and say e.g. "subtracts sticky item size from totalSize"

src/internal/hooks/use-virtual/index.ts
```diff
@@ -74,9 +76,15 @@ export function useVirtual<Item extends object>({
     [items, rowVirtualizer.virtualItems]
   );
 
+  // Adjust totalSize to account for 1px overlap per item (matching the position adjustment in renderOptions: /select/utils/render-options.tsx)
```
pan-kot: I would remove the "(matching the position adjustment in renderOptions: /select/utils/render-options.tsx)", which is specific to select and multiselect, while the use-virtual is an abstract utility and is also used in autosuggest.

src/internal/hooks/use-virtual/index.ts
```diff
@@ -74,9 +76,15 @@ export function useVirtual<Item extends object>({
...
+  // When firstItemSticky is enabled, the select-all is shifted down by 1 and other items are shifted down by (index + 2)
```
pan-kot: I don't quite get what "select-all is shifted down by 1" and "other items are shifted down by (index + 2)" means. I think the fact that the sticky item size is subtracted from the total size does not require explanation - what remains unclear is why do we need the 2px margin.

src/select/parts/virtual-list.tsx
```diff
@@ -53,6 +53,7 @@ const VirtualListOpen = forwardRef(
       // eslint-disable-next-line react-hooks/exhaustive-deps
       estimateSize: useCallback(() => fallbackItemHeight, [width?.inner, filteringValue]),
       firstItemSticky: firstOptionSticky,
+      itemOverlap: 1, // 1px overlap
```
pan-kot: nit: I think the comment is not needed as now the property name communicates the intent well

src/select/utils/render-options.tsx
```diff
@@ -65,11 +65,29 @@ export const renderOptions = ({
...
+    // Adjust virtual position to create 1px overlap between consecutive selected items
+    // When firstOptionSticky is enabled (enableSelectAll), the select-all option needs to be shifted down by 1,
+    // and all subsequent items need to be shifted up by (index + 1) instead of just index
+    let adjustedVirtualPosition: number | undefined = undefined;
+
+    if (!virtualItem) {
+      adjustedVirtualPosition = undefined;
+    } else if (!firstOptionSticky) {
+      // Shift every item up by one to create a 1px overlap
+      adjustedVirtualPosition = virtualItem.start - index * 1;
```
pan-kot: Why are we still doing it here? The virtual items are computed by useVirtual, and we already pass the itemOverlap=1 to it. Let's integrate the logic that transforms virtual item start positions with respect to the index and enhance the use-virtual tests to secure that.

src/internal/hooks/use-virtual/__tests__/use-virtual.test.tsx
mxschll: Because it only tests the offset right now and not the pre-existing functionality. I can rename it.

pages/selectable-item/virtual-scroll.page.tsx
mxschll: Ah I see, I would've clicked on the buttons. But that is way more convenient 👍

pages/selectable-item/virtual-scroll.page.tsx
mxschll: We're testing multiple selectables (autosuggest, select, multiselect). They all use selectable item, so it seemed like the best place to put it. What do you think?

src/internal/hooks/use-virtual/__tests__/use-virtual-offset.test.tsx
mxschll: Without mocking, we'll get an exception when running the test: 

```
/Volumes/workplace/github/components/src/internal/vendor/react-virtual.js:4
    export { defaultRangeExtractor, useVirtual } from 'react-virtual';
    ^^^^^^

    SyntaxError: Unexpected token 'export'

      3 | import React, { useEffect, useMemo, useRef } from 'react';
      4 |
    > 5 | import { useVirtual as useVirtualDefault, VirtualItem } from '../../vendor/react-virtual';
        | ^
      6 | import stickyRangeExtractor from './sticky-range-extractor';
```

This get's resolved when using jest.mock(). Maybe I'll just remove the comment.

src/internal/hooks/use-virtual/index.ts
mxschll: Yea per default the sticky item is already overlapping the subsequent item, but when we move it down by 1, it's too much. The 2px offset fixes that.

pages/selectable-item/virtual-scroll.page.tsx
mxschll: Test pages were added in https://github.com/cloudscape-design/components/pull/4097

src/internal/hooks/use-virtual/index.ts
mxschll: Removed 2px offset. It originally addressed another bug where a sticky item overlaps 1px at the top and 1px at the bottom (see screenshot). But I'll address this bug in a separate PR.
image

src/internal/hooks/use-virtual/index.ts
```diff
@@ -61,22 +63,45 @@ export function useVirtual<Item extends object>({
...
-          const mountedCount = measuresCache.current.get(items[virtualItem.index]) ?? 0;
-          if (mountedCount < MAX_ITEM_MOUNTS) {
-            virtualItem.measureRef(node);
-            measuresCache.current.set(items[virtualItem.index], mountedCount + 1);
-          }
-        },
-      })),
-    [items, rowVirtualizer.virtualItems]
+      rowVirtualizer.virtualItems.map((virtualItem, index) => {
+        let adjustedStart: number;
```
pan-kot: I don't quite get this code - could you please add code comments explaining the computations?

src/internal/hooks/use-virtual/__tests__/use-virtual.test.tsx
```diff
@@ -0,0 +1,227 @@
...
+  const parentRef = useRef<HTMLDivElement>(null);
+  const result = useVirtual({
+    items,
+    parentRef,
+    estimateSize: () => 50,
+    firstItemSticky,
+    itemOverlap,
+  });
+
+  onResult(result);
```
pan-kot, line 37: nit: you can probably simplify that by introducing a renderUseVirtual helper:

```
function renderUseVirtual(options) {
  let hookResult;
  const { rerender } = render(<Test onResult={(r) => (hookResult = r)} {...options} />);
  return [
    hookResult!,
    { rerender: (options) =>  rerender(<Test onResult={(r) => (hookResult = r)} {...options} />) },
  ];
}

function Test({ onResult, ...options }) {
   const result = useVirtual(options);
   onResult(result);
   return <...>
}
```

Then, you won't need the capturedResult to be defined in each test:
```
test('computes totalSize with itemOverlap=1', () => {
    // ...
    const [result] = renderUseVirtual({...});
    expect(result.totalSize).toBe(148);
  });
```

src/internal/hooks/use-virtual/__tests__/use-virtual.test.tsx
```diff
@@ -0,0 +1,227 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
pan-kot, line 1: nice tests!

src/internal/hooks/use-virtual/__tests__/use-virtual.test.tsx
```diff
@@ -0,0 +1,227 @@
...
+        onResult={result => (capturedResult = result)}
+      />
+    );
+
+    expect(capturedResult).not.toBeNull();
+    // totalSize = 160 - (maxRenderIndex * itemOverlap) - firstItemSize = 160 - ((3-1) * 1) - 60 = 98
+    expect(capturedResult!.totalSize).toBe(98);
+  });
+
+  test('calculates item positions without itemOverlap=0', () => {
```
pan-kot:
```suggestion
  test('calculates item positions with itemOverlap=0', () => {
```

src/internal/hooks/use-virtual/__tests__/use-virtual.test.tsx
```diff
@@ -0,0 +1,227 @@
...
+    let capturedResult: ReturnType<typeof useVirtual<TestItem>> | null = null;
+    render(<TestComponent items={items} itemOverlap={1} onResult={result => (capturedResult = result)} />);
+
+    expect(capturedResult).not.toBeNull();
+    // totalSize = 150 - (maxRenderIndex * itemOverlap) = 150 - ((3-1) * 1) = 148
+    expect(capturedResult!.totalSize).toBe(148);
+  });
+
+  test('computes totalSize with itemOverlap=0', () => {
+    const items: TestItem[] = [{ id: '1' }, { id: '2' }];
```
pan-kot: Are these input items even needed? This test returns more virtual items than there are items and it is ok. I recommend to remove the props that do not affect the end result (e.g. you can probably pass [] as items).

src/internal/components/selectable-item/styles.scss
```diff
@@ -6,6 +6,43 @@
 @use '../../styles' as styles;
 @use '../../styles/tokens' as awsui;
 
+// When using virtual scrolling, use box-shadows to mimic changing border width
```
pan-kot: can we use box-shadows instead of borders for both virtual and plain lists?

SpyZzey:
Instead of calculating the offset, I think we could reduce the height of the measure-strut (which is used by tanstack virtual to calculate the row offset) by 1 px using css, which escalates down, and each item will be offset by -1px compared to the previous one (similar to what you are doing manually).

mxschll:
> Instead of calculating the offset, I think we could reduce the height of the measure-strut (which is used by tanstack virtual to calculate the row offset) by 1 px using css, which escalates down, and each item will be offset by -1px compared to the previous one (similar to what you are doing manually).

Yes, that’s true. However, there might be cases where we don’t want any overlap, in which case we can simply omit the offset parameter. Defining it explicitly through a parameter feels clearer and avoids potential side effects caused by an implicit -1 px adjustment in the CSS. That might not be immediately obvious to others...

src/internal/hooks/use-virtual/index.ts
mxschll: Added in new revision.

src/internal/hooks/use-virtual/__tests__/use-virtual.test.tsx
mxschll: I am changing it to: `without providing itemOverlap`. That was my initial intend.

src/internal/components/selectable-item/styles.scss
mxschll: Yes, we could, but I’d push back a bit on that. We’re using borders and outlines across other components, so I don’t see a strong reason to deviate from that here, except for this specific need. I’d also like to introduce as few changes to this component as possible to minimize unintended side effects :)

SpyZzey: But for the same component, wouldn't it make more sense to change it for both? I would imagine it could also help removing maintenance later on by just having to update the box-shadow and not both box-shadow and border (if we ever need to change something here again - which probably we will)

src/select/parts/virtual-list.tsx
```diff
@@ -54,6 +54,7 @@ const VirtualListOpen = forwardRef(
       // eslint-disable-next-line react-hooks/exhaustive-deps
       estimateSize: useCallback(() => fallbackItemHeight, [width?.inner, filteringValue]),
       firstItemSticky: firstOptionSticky,
+      itemOverlap: 1,
```
SpyZzey: Thinking about it again, I believe this will break if people change the design token of $border-divider-list-width or $border-item-width. Did you check changing those tokens and see what happens?

(Though I also believe there is a bug in our css currently, which will break too)

mxschll: I believe both tokens are not themeable, meaning people can't change them. But I'll check it again.

SpyZzey: At least we changed them when transitioning from classic to vr. I think that's when the bug was introduced too. At least in classic, it looks fine.

pages/multiselect/screenshot.page.tsx
```diff
@@ -36,11 +36,11 @@ export default function () {
       <h1>Multiselect for screenshot</h1>
 
       <button id="toggle-virtual" onClick={() => setVirtualScroll(vs => !vs)}>
-        Toggle virtual scroll
+        Toggle virtual scroll: {virtualScroll ? 'on' : 'off'}
```
pan-kot, line 39: nit: should we use a Checkbox for that?

```
<Checkbox checked={virtualScroll} onChange={({ detail }) => setVirtualScroll(detail.checked)}>Virtual scroll</Checkbox>
```

src/internal/hooks/use-virtual/index.ts
```diff
@@ -61,22 +61,29 @@ export function useVirtual<Item extends object>({
...
-        measureRef: (node: null | HTMLElement) => {
-          const mountedCount = measuresCache.current.get(items[virtualItem.index]) ?? 0;
-          if (mountedCount < MAX_ITEM_MOUNTS) {
-            virtualItem.measureRef(node);
-            measuresCache.current.set(items[virtualItem.index], mountedCount + 1);
-          }
-        },
-      })),
+      rowVirtualizer.virtualItems.map(virtualItem => {
+        return {
```
pan-kot: nit: as there are no more computations before this return - it can be omitted

src/internal/hooks/use-virtual/index.ts
```diff
@@ -61,22 +61,29 @@ export function useVirtual<Item extends object>({
...
+            }
+          },
+        };
+      }),
     [items, rowVirtualizer.virtualItems]
   );
 
+  // If first item is sticky, substract that item's size from the total size
+  const firstItemSize = virtualItems[0]?.size ?? 0;
+  const adjustedTotalSize = firstItemSticky ? rowVirtualizer.totalSize - firstItemSize : rowVirtualizer.totalSize;
```
pan-kot, line 80: the final solution is very clean!

pan-kot (approved):

pages/multiselect/screenshot.page.tsx
mxschll: Yea, I'll open a separate PR - needed this for my own testing primarily but thought it makes sense to show the state.

src/internal/components/selectable-item/styles.scss
```diff
@@ -179,42 +217,62 @@
...
-      border-inline-start-color: awsui.$color-border-dropdown-container;
-      border-inline-end-color: awsui.$color-border-dropdown-container;
-      padding-inline: 0;
-      > .selectable-item-content {
-        padding-inline: styles.$control-padding-horizontal;
-      }
+      box-shadow:
+        inset 0 $divider-w 0 0 awsui.$color-border-dropdown-item-default,
+        inset 0 $neg-divider-w 0 0 awsui.$color-border-dropdown-item-default,
+        inset awsui.$border-item-width 0 0 0 awsui.$color-border-dropdown-container;
```
SpyZzey: This `awsui.$border-item-width`, should be `awsui.$border-width-popover`

SpyZzey: Otherwise, it looks like this if we change the border-item-width (which is different to the border-width of the popover itself)

Screenshot 2026-01-19 at 15 06 58

src/internal/components/selectable-item/styles.scss
```diff
@@ -179,42 +217,62 @@
...
       &:not(.with-scrollbar) {
-        border-inline-end-width: awsui.$border-item-width;
-      }
-
-      &.with-scrollbar {
-        border-inline-end-width: 0;
+        box-shadow:
+          inset 0 $divider-w 0 0 awsui.$color-border-dropdown-item-default,
+          inset 0 $neg-divider-w 0 0 awsui.$color-border-dropdown-item-default,
+          inset awsui.$border-item-width 0 0 0 awsui.$color-border-dropdown-container,
```
SpyZzey: https://github.com/cloudscape-design/components/pull/4064/files#r2704911910

Probably the same thing here

src/internal/components/selectable-item/styles.scss
```diff
@@ -225,33 +283,43 @@
         }
       }
       &.after-header {
-        border-block-start-color: awsui.$color-background-dropdown-item-default;
+        box-shadow:
+          inset 0 $divider-w 0 0 awsui.$color-background-dropdown-item-default,
+          inset 0 $neg-divider-w 0 0 awsui.$color-border-dropdown-item-default,
+          inset awsui.$border-item-width 0 0 0 awsui.$color-border-dropdown-container;
```
SpyZzey: https://github.com/cloudscape-design/components/pull/4064/files#r2704911910

Probably same thing here

src/internal/components/selectable-item/styles.scss
mxschll: `awsui.$border-item-width` was used before and is not themable. `awsui.$border-width-popover` on the other hand is themable. If we change this var now, this might be a breaking change. Also, the border is applied on the selectable *item* component, so I _think_ `awsui.$border-*item*-width` might be correct.

src/internal/components/selectable-item/styles.scss
mxschll: Moving discussion here: https://github.com/cloudscape-design/components/pull/4064#discussion_r2705302403

src/internal/components/selectable-item/styles.scss
mxschll: Moving discussion here: https://github.com/cloudscape-design/components/pull/4064#discussion_r2705302403

src/internal/components/selectable-item/styles.scss
mxschll: Changing `awsui.$border-item-width` looks like this for me, which seems ok?

image

PR #4068, opened 2025-11-24
feat: implement style api for segmented control component

mxschll:
### Description

Extends the segmented control component with a style API for customization, allowing developers to override default component styles.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/segmented-control/style-permutations.page.tsx
+++ b/pages/segmented-control/style-permutations.page.tsx
@@ -76,7 +76,7 @@ const permutations = createPermutations<SegmentedControlProps>([
 export default function SegmentedControlScenario() {
   return (
     <article>
-      <h1>SegmentedControl permutations</h1>
+      <h1>SegmentedControl Custom Style Permutations</h1>
       <ScreenshotArea>
         <PermutationsView permutations={permutations} render={permutation => <SegmentedControl {...permutation} />} />
       </ScreenshotArea>


pages/segmented-control/style-permutations.page.tsx
```diff
@@ -0,0 +1,85 @@
...
+    },
+    focusRing: {
+      borderColor: 'light-dark(#5f6ac3, #7c3aed)',
+      borderRadius: '12px',
+      borderWidth: '2px',
+    },
+  },
+};
+
+const iconSvg = (
```
jperals, line 41: Nitpick: is the custom SVG relevant to this feature?

src/internal/styles/forms/mixins.scss
```diff
@@ -93,6 +94,9 @@
     border-end-start-radius: $border-radius;
     border-end-end-radius: $border-radius;
     box-shadow: $box-shadow;
+    @if $z-index {
+      z-index: $z-index;
```
jperals, line 98: Why is this necessary?

pages/segmented-control/style-permutations.page.tsx
```diff
@@ -0,0 +1,85 @@
...
+      ],
+    ],
+    label: ['Segmented Control Label'],
+  },
+]);
+
+export default function SegmentedControlScenario() {
+  return (
+    <article>
+      <h1>SegmentedControl permutations</h1>
```
jperals: I suggest to change this heading text to mention its purpose related to custom styling

src/segmented-control/__tests__/__snapshots__/styles.test.tsx.snap
```diff
@@ -0,0 +1,79 @@
...
+
+exports[`getSegmentedControlRootStyles handles all possible style configurations 3`] = `
+{
+  "borderRadius": "8px",
+}
+`;
+
+exports[`getSegmentedControlSegmentStyles handles all possible style configurations 1`] = `
+{
+  "--awsui-style-background-active-d43v8n": undefined,
```
jperals, line 23: What do these hashes depend on? Can they change due to something unrelated?

src/internal/styles/forms/mixins.scss
mxschll: We need to be able to set the z-index of the focus ring here, because otherwise adjacent segments will overlap the focus ring if they have a background color.

pages/segmented-control/style-permutations.page.tsx
mxschll: It's out of scope for now.

src/internal/styles/forms/mixins.scss
mxschll: It could look like this, without `z-index = 1`, but we want the focus ring to stay on top:

image

pages/segmented-control/style-permutations.page.tsx
mxschll: Done.

src/segmented-control/__tests__/__snapshots__/styles.test.tsx.snap
mxschll: The hash is dependent on the list in [custom-css-properties.js](https://github.com/cloudscape-design/components/blob/4e7f8f30e5eb88b42ed0bbefe3a530c4308ab943/build-tools/tasks/generate-custom-css-properties.js#L4) and get generated using this function:

```
const hash = getHashDigest(Buffer.from(JSON.stringify(customCssPropertiesList)), 'md5', 'base36', 6);
```

They will change if we introduce a new custom property, which is rarely the case. I agree that this is not ideal, but I think we can live with that for now until we have a more general solution for testing the style api. What do you think?

pages/segmented-control/style-permutations.page.tsx
jperals: What I mean is whether it makes sense to include this in the dev page for testing. Can using `iconSvg` affect the custom styles (or the other way around) in any way?

I wonder if the dev page can be even further simplified if we only include the prop values that are relevant to test the custom styles feature.

Not a blocker though.

src/segmented-control/__tests__/__snapshots__/styles.test.tsx.snap
jperals: I have doubts whether we need these snapshot tests, provided that we have visual regression tests in the pipeline.

Even if adding new custom properties is rare, when we do it these tests will fail unexpectedly, but the end user experience will not change —in this sense, the visual result is what matters here.

This said, not a blocker.

jperals (approved):
No blocking comments

pages/segmented-control/style-permutations.page.tsx
mxschll: You are right, thanks for pointing that out.

src/segmented-control/__tests__/__snapshots__/styles.test.tsx.snap
mxschll: Yeah, I think we might need to think about a more unified testing strategy for the style api overall. Maybe introducing a small util function would help.

PR #4070, opened 2025-11-24
refactor: Apply padding to option-content-div instead of selectable-item outer-div

SpyZzey:
### Description

Currently, the hierarchical padding is applied to the same div as the border and the padding (border-width padding) when selected/unselected. This PR moves this "business"-padding (e.g. hierarchical padding, general spacing padding) to the inner option-content div, which was previously only used for testing.

This is needed so we can extend the selectable-item components to allow for custom styling in the content by removing this "business"-padding.

### How has this been tested?

- Manual testing.
- Ran through pipeline for screenshot test validation.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/selectable-item/disable-content-styling.permutations.page.tsx
+++ b/pages/selectable-item/disable-content-styling.permutations.page.tsx
@@ -12,10 +12,10 @@ import { permutationsConfigs } from './common';
 // Copy permutations and add selectable-items with no-content-styling.
 const permutations = createPermutations<SelectableItemProps>(permutationsConfigs);
 
-export default function DisableContentStylingSelectableItemPermutations() {
+export default function SelectableItemPermutations() {
   return (
     <>
-      <h1>Selectable item permutations</h1>
+      <h1>Selectable item permutations with disableContentStyling</h1>
       <ScreenshotArea>
         <ul role="listbox" aria-label="list">
           <PermutationsView


pages/selectable-item/permutations.page.tsx
```diff
@@ -66,7 +66,15 @@ const permutations = createPermutations<SelectableItemProps>([
     isChild: [true],
     children: [childOption],
   },
-]);
+];
+
+// Copy permutations and add selectable-items with no-content-styling.
+permutationsConfigs = [
+  ...permutationsConfigs,
+  ...permutationsConfigs.map(config => ({ ...config, disableContentStyling: [true] })),
```
jperals: Note that adding these permutations will cause some operational friction as the visual regression tests will fail. My recommendation is to move the new permutation to a new page. And IMO they don't need to be covered by visual regression, because this is an internal option, not customer-facing.

src/internal/components/selectable-item/styles.scss
```diff
@@ -26,12 +26,30 @@
   border-inline-end-width: 0;
   // to compensate for the loss of padding due to the removal of the left and right borders
   // and differences in default divider + selected border widths (visual refresh)
-  padding-block: styles.$option-padding-with-border-placeholder-vertical;
-  padding-inline: calc(#{styles.$control-padding-horizontal} + #{awsui.$border-item-width});
+  padding-block: calc(#{awsui.$border-item-width} - #{awsui.$border-divider-list-width});
+  padding-inline: awsui.$border-item-width;
+
+  // For custom styles and background on the option-content, the content should not clip through the selectable-item borders..
+  overflow: hidden;
```
jperals, line 33: In which situation can the content overflow? Won't the selectable item just stretch to fit its content?

src/internal/components/selectable-item/index.tsx
```diff
@@ -29,6 +29,7 @@ const SelectableItem = (
     isSelectAll,
     virtualPosition,
     padBottom,
+    disableContentStyling,
```
jperals, line 32: Why do we need both approaches to remove the padding when the content is custom:
- conditionally disable the padding
- move the padding to an internal div
Wouldn't one of them suffice?

avinashbot: +1 - If all the "business" padding is now on this classname now, it seems easier to just remove the `option-content` classname if we want to disable the styling. Having a separate classname just to zero out the padding seems like too much work.

pages/selectable-item/permutations.page.tsx
SpyZzey: It will be customer facing once the arbitrary content for select/multiselect will go live

src/internal/components/selectable-item/styles.scss
SpyZzey: Yes, it will stretch the content. But if we set a background color, it will overlap with the selected/highlighted state borders.

src/internal/components/selectable-item/index.tsx
SpyZzey: That is a good point, will think about it again

SpyZzey: Damn, i don't know why i didn't think about that 😃 
Option-content can't be removed as it was used for testing purposes before. But adding another with the padding and removing the class if disableContentStyling=true would make so much more sense.

pages/selectable-item/permutations.page.tsx
jperals: Still, `disableContentStyling` will keep being internal implementation of an internal component. I think it makes more sense to cover the customer-facing components using the new props.

src/internal/components/selectable-item/styles.scss
jperals: Is it just the rounded borders that would overlap with the inner non-rounded ones? If so, it should be fine. Just making sure that actual content is not potentially clipped.

SpyZzey: Yes, it is just the rounded borders

pages/selectable-item/disable-content-styling.permutations.page.tsx
```diff
@@ -0,0 +1,31 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
jperals, line 1: In order to be rendered as a page, the file name needs to end with `.page.tsx`

pages/selectable-item/common.tsx
```diff
@@ -0,0 +1,72 @@
...
+const optionGroupHeader = 'group header';
+const childOption = 'child option';
+
+export const permutationsConfigs: ComponentPermutations<SelectableItemProps>[] = [
+  {
+    selected: [false],
+    highlighted: [false, true],
+    disabled: [false, true],
+    children: [simpleOption],
+    disableContentStyling: [true],
```
jperals: Shouldn't `disableContentStyling` be `[false]` (for each item)?

pages/selectable-item/disable-content-styling.permutations.page.tsx
```diff
@@ -0,0 +1,32 @@
...
+import ScreenshotArea from '../utils/screenshot-area';
+import { permutationsConfigs } from './common';
+
+// Copy permutations and add selectable-items with no-content-styling.
+const permutations = createPermutations<SelectableItemProps>(permutationsConfigs);
+
+export default function InputPermutations() {
+  return (
+    <>
+      <h1>Selectable item permutations</h1>
```
jperals: Minor: suggest to change the title to reflect the difference between this page and the one that it was copied from.

pages/selectable-item/disable-content-styling.permutations.page.tsx
```diff
@@ -0,0 +1,32 @@
...
+      <ScreenshotArea>
+        <SpaceBetween size="xs">
```
jperals: I don't think this SpaceBetween does anything, since it conatins one single element.

src/internal/components/selectable-item/styles.scss
```diff
@@ -112,35 +147,49 @@
...
-      padding-block: styles.$group-option-padding-with-border-placeholder-vertical;
-      padding-inline: calc(#{styles.$control-padding-horizontal} + #{awsui.$border-item-width});
+      > .selectable-item-content {
+        padding-block: styles.$group-option-padding-vertical;
+        padding-inline: styles.$control-padding-horizontal;
+        font-weight: bold;
+      }
       &.highlighted {
-        color: awsui.$color-text-dropdown-item-highlighted;
+        > .selectable-item-content {
```
jperals: Nitpick: these two nested rulesets could be merged into one

```
&.highlighted > .selectable-item-content { ... }
```

jperals (approved):

PR #4075, opened 2025-11-27
feat: Arbitrary content for select + multiselect

SpyZzey:
### Description
Adds a `renderOption`-property to the Select + Multiselect components to allow for arbitrary content as options.

Related links, issue #, if available: [ZNzTAy9SOccH], [JbVHAstbb3bw]

### How has this been tested?

- Added unit tests for select + multiselect
- Manual Testing

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/multiselect/__tests__/render-option.test.tsx
+++ b/src/multiselect/__tests__/render-option.test.tsx
@@ -29,16 +29,16 @@ describe('Multiselect renderOption', () => {
 
   test('receives correct item properties for item option', () => {
     const renderOption = jest.fn(() => <div>Custom</div>);
-    const childOption = { label: 'Test', value: '1' };
+    const itemOption = { label: 'Test', value: '1' };
     const wrapper = renderMultiselect({
-      options: [childOption],
+      options: [itemOption],
       renderOption,
     });
     wrapper.openDropdown();
     expect(renderOption).toHaveBeenCalledWith(
       expect.objectContaining({
         item: expect.objectContaining({
-          option: expect.objectContaining(childOption),
+          option: expect.objectContaining(itemOption),
           selected: false,
           highlighted: false,
           disabled: false,
--- a/src/select/__tests__/render-option.test.tsx
+++ b/src/select/__tests__/render-option.test.tsx
@@ -29,16 +29,16 @@ describe('Select renderOption', () => {
 
   test('receives correct item properties for item option', () => {
     const renderOption = jest.fn(() => <div>Custom</div>);
-    const childOption = { label: 'Test', value: '1' };
+    const itemOption = { label: 'Test', value: '1' };
     const wrapper = renderSelect({
-      options: [childOption],
+      options: [itemOption],
       renderOption,
     });
     wrapper.openDropdown();
     expect(renderOption).toHaveBeenCalledWith(
       expect.objectContaining({
         item: expect.objectContaining({
-          option: expect.objectContaining(childOption),
+          option: expect.objectContaining(itemOption),
           selected: false,
           highlighted: false,
           disabled: false,


pages/multiselect/custom-render-option.page.tsx
```diff
@@ -0,0 +1,71 @@
...
+      return <div>Group: {item.option.label}</div>;
+    } else if (item.type === 'item') {
+      return <div>Item: {item.option.label}</div>;
+    }
+
+    return null;
+  };
+
+  return (
+    <ScreenshotArea>
```
jperals: Minor: the screenshot area does not need to include the page header.

You can also use the SimplePage component from `pages/app/templates.tsx`

src/internal/components/button-trigger/styles.scss
```diff
@@ -133,6 +133,13 @@ $padding-block-inner-filtering-token: 0px;
     }
   }
 
+  &.custom-option {
+    // Remove default paddings for custom options
+    padding-block: 0;
+    padding-inline-start: 0;
```
jperals, line 139: No `padding-inline-end`?

src/multiselect/interfaces.ts
```diff
@@ -71,12 +74,48 @@ export interface MultiselectProps extends BaseSelectProps {
...
+    option: Option;
+    indeterminate: boolean;
+    highlighted: boolean;
+    selected: boolean;
+  }
+  export type MultiselectItem = MultiselectOptionItem | MultiselectOptionGroupItem | MultiselectSelectAllItem;
+  export type MultiselectOptionItemRenderer = (props: {
+    item: MultiselectItem;
+    filterText?: string;
+  }) => ReactNode | null;
```
jperals, line 118: Is `| null` necessary since the ReactNode type already includes `null`?

pages/multiselect/custom-render-option.page.tsx
jperals: Same in the page for the Select component

src/internal/components/button-trigger/styles.scss
SpyZzey: No, setting the padding-inline-end to 0 causes the select chevron to overlap the trigger content, so I'd just leave it as is.

src/multiselect/interfaces.ts
SpyZzey: No, it is not necessary. I added it to emphasize that the Renderer can also return `null` values to keep the default design instead of not rendering the content.

pages/multiselect/custom-render-option.page.tsx
```diff
@@ -0,0 +1,67 @@
...
+    } else if (item.type === 'item') {
+      return <div>Item: {item.option.label}</div>;
+    }
+
+    return null;
+  };
+
+  return (
+    <SimplePage title="Multiselect with custom item renderer" screenshotArea={{}}>
+      <div style={{ width: '400px' }}>
```
jperals: Nitpick (not blocking because it is not within the components): use `inlineSize` instead to follow our logical properties guidelines

src/__tests__/snapshot-tests/__snapshots__/documenter.test.ts.snap
```diff
@@ -1,4 +1,4 @@
-// Jest Snapshot v1, https://jestjs.io/docs/snapshot-testing
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
jperals: Do we know why this line changed?

src/select/parts/item.tsx
```diff
@@ -33,10 +36,56 @@ export interface ItemProps {
...
+}): SelectProps.SelectOptionItem => {
+  return {
+    type: 'item',
+    index: props.virtualIndex ?? props.index,
+    option: props.option.option as OptionDefinition,
+    selected: props.selected,
+    highlighted: props.highlighted,
+    disabled: props.disabled,
+    parent: props.parentProps
+      ? toSelectOptionGroupItem({
```
jperals, line 75: Since this creates a new object, is there a chance that this creates issues for customers, for example if they try to do determine parent-child relationships by performing identity comparisons?

SpyZzey: It is true that the object identity will be different. But I'd actually discourage people from using identity comparison as with react rerenders, we will create a new object every rerender anyways and therefore, one instance of an object can't be stored without causing unintended issues. Customers should use the data they provided in the option property to identify parents (e.g. with an unique identifier key).

src/__tests__/snapshot-tests/__snapshots__/documenter.test.ts.snap
SpyZzey: It seems like this comment will not be changed for existing files, but I have deleted the file and recreated it from scratch, which will resolve in the goo.gl link. However, goo.gl are deprecated and will face EOL soon.
Jest has resolved this issue in https://github.com/jestjs/jest/pull/15787, Release 30.1.0, however, we are still using 29.7.0. I assume it has been changed manually on main in the past, but deleting and recreating the file will bring back the short link.

SpyZzey: Other files (e.g. the design token snapshots) also use the deprecated goo.gl shortlink. Will update them here too.

jperals (dismissed):

src/multiselect/__tests__/render-option.test.tsx
```diff
@@ -0,0 +1,180 @@
...
+    const renderOption = jest.fn(props => <div>Custom {props.item.index}</div>);
+    const wrapper = renderMultiselect({ options: optionsWithGroups, renderOption });
+    wrapper.openDropdown();
+    expect(renderOption).toHaveBeenCalled();
+    const elementWrapper = wrapper.findDropdown().findOption(1)!.getElement();
+    expect(elementWrapper).not.toBeNull();
+    expect(elementWrapper).toHaveTextContent('Custom 0');
+  });
+
+  test('receives correct item properties for child option', () => {
```
jperals: Is it a "child" if it does not have a parent?

Can we also have a test for the `parent` argument being present for a child with parent?

Same in the Select component tests

PR #4079, opened 2025-11-28
chore: Themeable icon stroke width

at-susie:
### Description

As part of the iconography extension project, this PR introduces themeable border tokens that define stroke widths for each icon size variant. These tokens provide the flexibility needed to customize icon styling.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/visual-refresh-toolbar/toolbar/index.tsx
+++ b/src/app-layout/visual-refresh-toolbar/toolbar/index.tsx
@@ -193,7 +193,6 @@ export function AppLayoutToolbarImplementation({
               selected={!drawerExpandedMode && !!activeAiDrawerId}
               disabled={anyPanelOpenInMobile}
               variant={aiDrawer?.trigger?.customIcon ? 'custom' : 'circle'}
-              hasTooltip={true}
               testId={`awsui-app-layout-trigger-${aiDrawer?.id}`}
               isForPreviousActiveDrawer={true}
             />
--- a/src/select/utils/render-options.tsx
+++ b/src/select/utils/render-options.tsx
@@ -92,7 +92,7 @@ export const renderOptions = ({
       const optionId = props.id ?? `${idPrefix}-option-${index}`;
       return (
         <OptionGroup
-          key={index}
+          key={`group-${index}`}
           virtual={virtualItems?.[index] !== undefined}
           ariaLabelledby={optionId}
           ariaDisabled={props['aria-disabled']}


pages/theming/themed-stroke-width.page.tsx
```diff
@@ -0,0 +1,335 @@
...
+
+import styles from '../icon/icons-list.scss';
+
+export default function () {
+  const [themed, setThemed] = useState<boolean>(false);
+  const [strokeSmall, setStrokeSmall] = useState<string>('1');
+  const [strokeNormal, setStrokeNormal] = useState<string>('1');
+  const [strokeMedium, setStrokeMedium] = useState<string>('1');
+  const [strokeBig, setStrokeBig] = useState<string>('1.5');
+  const [strokeLarge, setStrokeLarge] = useState<string>('2');
```
Who-is-PS: Does values are not aligned with the default values like 2px, 3px and 4px. Is this on purpose?

src/icon/mixins.scss
Who-is-PS, line 1: Just a general question here to the formula at the end. The stroke width is divided by scale factor, which contradicts the token definition, for example for big would the calc: 3px/ 2 = 1.5px but it is defined as 3px in tokens.

pages/theming/themed-stroke-width.page.tsx
```diff
@@ -0,0 +1,335 @@
...
+        </label>
+
+        <div style={{ display: 'grid', gridTemplateColumns: 'repeat(5, 1fr)', gap: '10px', maxWidth: '800px' }}>
+          <FormField label="Small (16px)" description="Default: 2px">
+            <Input
+              type="number"
+              value={strokeSmall}
+              onChange={evt => {
+                const numValue = parseFloat(evt.detail.value);
+                if (!isNaN(numValue) && numValue >= 1) {
```
Who-is-PS: Maybe to much but we could add here min/max constraints

src/icon/mixins.scss
at-susie: Thanks for asking. The Cloudscape icon component provides the `size` property ranging from `small` to `large`. The way of handling the different icon sizes is to scale up and down by changing the CSS `size` property. This allows us to manage only one single icon svg resource and use it for different sizes.

However, when we scale the icon up (for example, 2x for "big" size), the stroke width would also scale proportionally, making it too thick. To maintain consistent visual weight across all icon sizes, we need to compensate by dividing the stroke width by the scale factor to make it render with the expected visual stroke widths, for example "big" size:

- The icon itself is scaled up by 2x (from 16px to 32px)
- The token borderWidthIconBig is defined as 3px
- In the CSS, we apply: stroke-width: calc(3px / 2) = 1.5px
- When the browser renders the scaled-up icon (2x), the effective stroke width becomes: 1.5px × 2 = 3px

pages/theming/themed-stroke-width.page.tsx
at-susie: Good call. I will refine the UIs overall in the example page

pages/theming/themed-stroke-width.page.tsx
```diff
@@ -0,0 +1,335 @@
...
+              type="number"
+              value={strokeSmall}
+              onChange={evt => {
+                const numValue = parseFloat(evt.detail.value);
+                if (!isNaN(numValue) && numValue >= 1 && numValue <= 10) {
+                  setStrokeSmall(evt.detail.value);
+                } else if (evt.detail.value === '') {
+                  setStrokeSmall(evt.detail.value);
+                }
+              }}
```
Who-is-PS: We are using this code in nearly every input. We can refactor this code to function like createStrokeHandler and use it in the onChange.

pages/theming/themed-stroke-width.page.tsx
```diff
@@ -0,0 +1,335 @@
...
+} from '~components';
+import Icon, { IconProps } from '~components/icon';
+import icons from '~components/icon/generated/icons';
+import { applyTheme, Theme } from '~components/theming';
+
+import ScreenshotArea from '../utils/screenshot-area';
+
+import styles from '../icon/icons-list.scss';
+
+export default function () {
```
Who-is-PS: We should give this a name like ThemedStrokeWidthPage

pages/theming/themed-stroke-width.page.tsx
```diff
@@ -0,0 +1,335 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
Who-is-PS, line 1: General note: We are using hard coded validation bounds like 1, 10, and 0,5 
I would suggest to constants like:
const STROKE_WIDTH_CONFIG = {
  MIN: 1,
  MAX: 10,
  STEP: 0.5,
} as const;

pages/theming/themed-stroke-width.page.tsx
at-susie: Added the name.

pages/theming/themed-stroke-width.page.tsx
at-susie: Great callout. I updated it by extracting the logic

pages/theming/themed-stroke-width.page.tsx
at-susie: Updated along with the above comment: https://github.com/cloudscape-design/components/pull/4079#discussion_r2598665318

PR #4084, opened 2025-12-03
chore: Refactor card into separate internal component

jperals:
### Description

Intermediate step for `D347816619`.

Related doc: `LjfAAJG82aYE`

### How has this been tested?

- Run through my pipeline
- Added dev pages

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/card/header-permutations.page.tsx
+++ b/pages/card/header-permutations.page.tsx
@@ -10,12 +10,29 @@ import PermutationsView from '../utils/permutations-view';
 import { actions, CardPage, icon, longContent, longDescription, longHeader, shortHeader } from './common';
 
 const permutations = createPermutations<InternalCardProps & { width?: number }>([
+  // With header
   {
-    header: [shortHeader, longHeader, undefined],
+    header: [shortHeader, longHeader],
     children: [longContent],
     description: [undefined, longDescription],
     actions: [undefined, actions],
     icon: [undefined, icon],
+  },
+  // Without header and without description
+  {
+    header: [undefined],
+    children: [longContent],
+    description: [undefined],
+    actions: [undefined, actions],
+    icon: [undefined, icon],
+  },
+  // Without header but with description
+  {
+    header: [undefined],
+    children: [longContent],
+    description: [longDescription],
+    actions: [undefined, actions],
+    icon: [undefined],
   },
 ]);
 


src/internal/components/card/styles.scss
```diff
@@ -0,0 +1,112 @@
...
+*/
+
+@use 'sass:math';
+
+@use '../../styles' as styles;
+@use '../../styles/tokens' as awsui;
+@use './motion';
+
+@mixin card-style {
+  border-start-start-radius: awsui.$border-radius-container;
```
jperals, line 13: Styles moved over from the cards component.

For now, still using the container design tokens.

I will later introduce card-specific design tokens, which the component will use, falling back to these container tokens for backwards compatibility.

src/test-utils/dom/cards/index.ts
```diff
@@ -21,11 +21,6 @@ class CardSectionWrapper extends ComponentWrapper {
 }
 
 class CardWrapper extends ComponentWrapper {
-  /**
```
jperals, line 24: This was a bug which this PR fixes. I will fix consumer packages that had to use this 2-based indexing.

src/internal/components/card/interfaces.ts
```diff
@@ -0,0 +1,57 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import React from 'react';
+
+import { BaseComponentProps } from '../../base-component';
+
+export interface BaseCardProps extends BaseComponentProps {
```
jperals, line 7: Props that will be public once the component goes public

(see `LjfAAJG82aYE`)

src/internal/components/card/interfaces.ts
```diff
@@ -0,0 +1,57 @@
...
+   */
+  disableHeaderPaddings?: boolean;
+
+  /**
+   * Determines whether the card content has padding. If `true`, removes the default padding from the content area.
+   */
+  disableContentPaddings?: boolean;
+}
+
+export interface InternalCardProps extends BaseCardProps {
```
jperals, line 45: Internal props used by the Cards component

src/internal/components/card/styles.scss
cansuaa: At which point will it fall back to the container tokens exactly?

pages/card/common.tsx
cansuaa, line 1: For some reason I can't see this page in the [here](https://d21d5uik3ws71m.cloudfront.net/components/6e10723790b270758a4ccb3d614852e0a015faee/dev-pages-react16/index.html#/) - I did look at the latest deployed version

pages/card/header-permutations.page.tsx
```diff
@@ -0,0 +1,28 @@
...
+import Card from '~components/internal/components/card';
+import { InternalCardProps } from '~components/internal/components/card/interfaces';
+
+import createPermutations from '../utils/permutations';
+import PermutationsView from '../utils/permutations-view';
+import { actions, CardPage, icon, longContent, longDescription, longHeader, shortHeader } from './common';
+
+const permutations = createPermutations<InternalCardProps & { width?: number }>([
+  {
+    header: [shortHeader, longHeader, undefined],
```
cansuaa: It kinda looks broken when the header is empty but there is an icon, description and content, (the last ones [here](https://d21d5uik3ws71m.cloudfront.net/components/6e10723790b270758a4ccb3d614852e0a015faee/dev-pages-react16/index.html#/card/header-permutations)) though technically it works correctly so just putting this comment for awareness

pages/card/common.tsx
jperals: This is itself not a page (note how it does not end with `page.tsx`) but a file with common code used by the two pages

pages/card/header-permutations.page.tsx
jperals: In the last two cases (in which there are actions) it's because the actions and the descriptions are in different rows, so that the description can occupy the entire second row.

In the first of the three cases, it is because the header is occupying vertical space even if not present —I am now seeing if I can improve this one without affecting the StructuredItem component.

At the same time, this let me see a small issue which I now fixed: when there is no header or no content, the spacing between is removed, but I was not taking into account cases where header was not present but either description, icon or actions were —in these cases, the spacing should be kept.

src/internal/components/card/styles.scss
jperals: Before the component goes public. At this point the critical thing is that there are no regressions for the Cards component.

pages/card/header-permutations.page.tsx
jperals: After some back and fort, I decided to leave out the permutation(s) with icon and description but no header. These are the two things I've tried, which both _work_, but for such an edge case I don't think jsutifies making the logic more complicated:

- In StructuredItem, do not render the main content slot if no content and no actions (https://github.com/cloudscape-design/components/commit/b23c05294e944f7c00fe38de8b2d10017e49a6a6)
- If description is provided but no header, pass the description into the main `content` slot of StructuredItem (where the header should otherwise be) instead of the `secondaryContent`. This opens up the question of what are the exact combinations in which this should be done, e.g, only when a value is passed for `icon` or not necessarily?

cansuaa: Yes I also don't think it's worth to complicate it, I think removing the permutations was the right move since we don't expect customers to use it like that. I just added the comment for you to be aware but I'm glad it made you notice the problem with the spacing

src/internal/components/card/styles.scss
cansuaa: Ah okay, by fallback you mean using them now. I thought somehow it was going to fallback to container design tokens instead of card token

PR #4092, opened 2025-12-08
feat: Tooltip component

Who-is-PS:
### Description

Introduction of new component called Tooltip. Tooltip is used to display helpful contextual information when users hover over or focus on elements. Updated all components which are using the internal tooltip.  

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/tooltip/simple.page.tsx
+++ b/pages/tooltip/simple.page.tsx
@@ -39,7 +39,9 @@ export default function TooltipSimple() {
   };
 
   const handleBlurTooltip = () => {
-    // Only close if not currently interacting with tooltip
+    // Prevents the tooltip from closing when clicking inside it (e.g., to select text for copying).
+    // Without this check, mousedown on the tooltip triggers a blur on the trigger element,
+    // which would close the tooltip mid-interaction.
     if (!isInteractingWithTooltip) {
       setActiveTooltip(null);
     }
@@ -49,6 +51,8 @@ export default function TooltipSimple() {
     setActiveTooltip(id);
   };
 
+  // Track when user is interacting with tooltip content (e.g., selecting text).
+  // Used to prevent premature tooltip closure on blur events.
   const handleTooltipMouseDown = () => {
     setIsInteractingWithTooltip(true);
   };


pages/tooltip/component-integration.page.tsx
```diff
@@ -0,0 +1,592 @@
...
+import StatusIndicator from '~components/status-indicator';
+import Tabs from '~components/tabs';
+import Token from '~components/token';
+import Tooltip from '~components/tooltip';
+
+import ScreenshotArea from '../utils/screenshot-area';
+
+export default function InternalTooltipExamples() {
+  return (
+    <article>
```
pan-kot: nit: there is a `SimplePage` template to create test pages. It is not used very consistently yet, but using it here will help to simplify code:
```
<SimplePage title="Tooltip component integration" subtitle="..." screenshotaArea={{}}>
  ...
</SimplePage>
```

pages/tooltip/component-integration.page.tsx
```diff
@@ -0,0 +1,592 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
pan-kot: nit: when I read the page name for the first time I though it was created for integ tests. I think it is better to call it the same way you called the React component: internal-tooltip-examples.

pages/tooltip/component-integration.page.tsx
```diff
@@ -0,0 +1,592 @@
...
+function TruncatedTextExample() {
+  const ref1 = useRef<HTMLDivElement>(null);
+  const ref2 = useRef<HTMLDivElement>(null);
+  const ref3 = useRef<HTMLDivElement>(null);
+  const [show1, setShow1] = useState(false);
+  const [show2, setShow2] = useState(false);
+  const [show3, setShow3] = useState(false);
+
+  return (
+    <Container header={<Header variant="h2">Truncated Text with Overflow Tooltips</Header>}>
```
pan-kot: Why do we use this examples on the page dedicated to internal usages?

pan-kot: nit: I don't thin that we need all these containers and headers. We can find all details in code instead - I think it is enough to render all internal components featuring a tooltip in a single space-between. Wdyt?

pages/tooltip/component-integration.page.tsx
```diff
@@ -0,0 +1,592 @@
...
+
+  return (
+    <Container header={<Header variant="h2">DateRangePicker - GridCell (Date Disabled Reason)</Header>}>
+      <ScreenshotArea>
+        <DateRangePicker
+          value={value}
+          onChange={({ detail }) => setValue(detail.value)}
+          relativeOptions={[]}
+          isValidRange={() => ({ valid: true })}
+          i18nStrings={{
```
pan-kot: We can use default i18n with `<SimplePage i18n={{}}>`. As this page is dedicated to the tooltips - there is no need to provide much boilerplate for examples.

pages/tooltip/simple.page.tsx
```diff
@@ -0,0 +1,422 @@
...
+import Button from '~components/button';
+import SegmentedControl from '~components/segmented-control';
+import SpaceBetween from '~components/space-between';
+import Tooltip from '~components/tooltip';
+
+import ScreenshotArea from '../utils/screenshot-area';
+
+export default function TooltipSimple() {
+  // Interactive Position Control
+  const [showInteractive, setShowInteractive] = useState(false);
```
pan-kot: Let's use url-params for all page settings as those are much easier to target with integ tests or even when sharing a link to someone. You can check how it is done in other pages with the help of AppContext.

pages/tooltip/simple.page.tsx
```diff
@@ -0,0 +1,422 @@
...
+  const [interactivePosition, setInteractivePosition] = useState<'top' | 'right' | 'bottom' | 'left'>('top');
+  const interactiveRef = useRef<HTMLDivElement>(null);
+
+  const [showTop, setShowTop] = useState(false);
```
pan-kot: nit: this can be replaced with a single `useState<null | "start" | "end" | "top" | "bottom">(null)`

pages/tooltip/simple.page.tsx
```diff
@@ -0,0 +1,422 @@
...
+  const [showCode, setShowCode] = useState(false);
+  const linkRef = useRef<HTMLDivElement>(null);
+  const codeRef = useRef<HTMLDivElement>(null);
+
+  // Password input
+  const [showPassword, setShowPassword] = useState(false);
+  const passwordRef = useRef<HTMLDivElement>(null);
+
+  return (
+    <article>
```
pan-kot: Let's use SimplePage helper here, too

pan-kot: For settings you can use the `settings` slot of the simple page. In the current design - it is hard to understand where are settings control and where are the components with tooltips.

pages/tooltip/simple.page.tsx
```diff
@@ -0,0 +1,422 @@
...
+                  setInteractivePosition(detail.selectedId as 'top' | 'right' | 'bottom' | 'left')
+                }
+                options={[
+                  { id: 'top', text: 'Top' },
+                  { id: 'right', text: 'Right' },
+                  { id: 'bottom', text: 'Bottom' },
+                  { id: 'left', text: 'Left' },
+                ]}
+              />
+              <div
```
pan-kot: Let's create an in-page component like ButtonWithTooltip to not duplicate that much code

pages/tooltip/simple.page.tsx
```diff
@@ -0,0 +1,422 @@
...
+                ref={interactiveRef}
+                onMouseEnter={() => setShowInteractive(true)}
+                onMouseLeave={() => setShowInteractive(false)}
+                style={{ display: 'inline-block' }}
+              >
+                <Button
+                  variant="primary"
```
pan-kot: nit: I'd use standard buttons instead of primary as only one primary is allowed per page - making it a less common use case

pages/tooltip/simple.page.tsx
```diff
@@ -0,0 +1,422 @@
...
+                  { id: 'right', text: 'Right' },
+                  { id: 'bottom', text: 'Bottom' },
+                  { id: 'left', text: 'Left' },
+                ]}
+              />
+              <div
+                ref={interactiveRef}
+                onMouseEnter={() => setShowInteractive(true)}
+                onMouseLeave={() => setShowInteractive(false)}
+                style={{ display: 'inline-block' }}
```
pan-kot: why inline-block display is needed here?

pages/tooltip/simple.page.tsx
```diff
@@ -0,0 +1,422 @@
...
+                )}
+              </div>
+
+              <div
+                ref={codeRef}
+                onMouseEnter={() => setShowCode(true)}
+                onMouseLeave={() => setShowCode(false)}
+                style={{ display: 'inline-block' }}
+              >
+                <button
```
pan-kot: what is different between this one and the one udner truncated content?

pages/tooltip/simple.page.tsx
```diff
@@ -0,0 +1,422 @@
...
+                onBlur={() => setShowPassword(false)}
+              />
+              <span id="password-description" hidden={true}>
+                Password Rules: Minimum of 8 characters, Include at least one lowercase letter, one uppercase letter,
+                one number and one special character, Unique to this website
+              </span>
+              {showPassword && (
+                <Tooltip
+                  content={
+                    <div>
```
pan-kot: nit: there is a weird padding in the top of the content. Also, I think we should be careful with using semantic elements in the tooltip as the list structure is not preserved for the screen-reader users.

src/__tests__/required-props-for-components.ts
```diff
@@ -166,6 +166,10 @@ const defaultProps: Record<string, Record<string, any>> = {
   steps: {
     steps: [],
   },
+  tooltip: {
+    content: 'Tooltip content',
+    getTrack: () => document.createElement('div'),
```
pan-kot: let's use `() => null` as the minimal setup here

pan-kot: and also `null` for the content prop

src/button-group/icon-button-item.tsx
```diff
@@ -62,14 +62,14 @@ const IconButtonItem = forwardRef(
         </InternalButton>
         {(canShowTooltip || canShowFeedback) && (
           <Tooltip
-            trackRef={containerRef}
+            className={clsx(testUtilStyles.tooltip, testUtilStyles['button-group-tooltip'])}
```
pan-kot: why do we actually assign both tooltip and button-group-tooltip styles here?

src/column-layout/styles.scss
```diff
@@ -54,7 +54,10 @@ $column-breakpoint-spans: (
 
 @mixin remove-useless-vertical-borders($col-size, $direction) {
   $items-per-row: math.div($grid-columns, $col-size);
-  $items-modifier: if($direction == 'inline-start', '+1', '');
+  $items-modifier: '';
+  @if $direction == 'inline-start' {
```
pan-kot: why this change?

src/internal/components/tooltip/index.tsx
pan-kot, line 16: why do we keep this one?

src/internal/components/tooltip/index.tsx
```diff
@@ -65,7 +63,7 @@ export default function Tooltip({
 
   return (
     <Portal>
-      <div className={styles.root} {...contentAttributes} data-testid={trackKey}>
+      <div className={styles.root} data-testid={trackKey} role="tooltip">
```
pan-kot: Let's document our use of data-testid in the api-docs for trackKey.

src/test-utils/dom/internal/tooltip.ts
```diff
@@ -1,9 +1,20 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
-import { ElementWrapper } from '@cloudscape-design/test-utils-core/dom';
+import { ComponentWrapper, ElementWrapper } from '@cloudscape-design/test-utils-core/dom';
```
pan-kot: Why do we need the internal tooltip wrapper?

src/test-utils/dom/tooltip/index.ts
```diff
@@ -0,0 +1,31 @@
...
+    return this.findByClassName(popoverStyles.content);
+  }
+
+  /**
+   * Returns the tooltip element by its trackKey (data-testid attribute).
+   * Useful when multiple tooltips are rendered and you need to find a specific one.
+   *
+   * @param trackKey - The trackKey value used when creating the tooltip
+   */
+  static findByTrackKey(trackKey: string): TooltipWrapper | null {
```
pan-kot: Currently, the component wrappers are considered internal API. We do not expect consumers to import the specific wrappers - but the createWrapper() util only. We should discuss if there are better way of introducing this helper or if we should remove it for now.

src/token/__tests__/token-tooltip.test.tsx
```diff
@@ -204,4 +204,34 @@ describe('Token Tooltip behavior', () => {
...
+    const wrapper = createWrapper(container).findToken()!;
+    const tokenElement = wrapper.getElement();
+    const labelElement = wrapper.findLabel().getElement();
+
+    mockEllipsisActive(labelElement, true);
+    triggerResizeObserver(labelElement);
+
+    // Show tooltip
+    fireEvent.mouseEnter(tokenElement);
+    expect(screen.queryByTestId('tooltip-live-region-content')).toBeInTheDocument();
```
pan-kot: Let's use our test-utils to find the tooltip instead. Ideally, it should be discoverable with token's test utils. If we do not offer a method in the token's wrapper yet - than can use the createWrapper().findTooltip() at least.

src/token/internal.tsx
```diff
@@ -113,11 +113,17 @@ function InternalToken({
       aria-labelledby={!ariaLabel ? ariaLabelledbyId : undefined}
       aria-disabled={!!disabled}
       role={role ?? 'group'}
-      onFocus={() => {
-        setShowTooltip(true);
+      onFocus={e => {
+        // Only show tooltip if focus is coming from outside the token
```
pan-kot: why?

src/tooltip/__integ__/tooltip.test.ts
```diff
@@ -0,0 +1,225 @@
...
+
+      // Test hover - tooltip should appear
+      await page.hoverElement('[data-testid="hover-button"]');
+
+      await expect(page.isTooltipVisible()).resolves.toBe(true);
+    })
+  );
+
+  test(
+    'shows tooltip on hover and verifies content',
```
pan-kot, line 28: Why do we need a separate "shows tooltip on hover" test if this one does the same and a little more?

Keep in mind that the integ tests do add a few seconds to the total running time each - so ideally we should focus on testing the behaviours that are not testable in the unit tests and avoid repetition.

src/tooltip/__integ__/tooltip.test.ts
```diff
@@ -0,0 +1,225 @@
...
+      await expect(page.isTooltipVisible()).resolves.toBe(true);
+
+      await browser.keys('Escape');
+      await page.waitForTooltipToDisappear();
+      await expect(page.isTooltipVisible()).resolves.toBe(false);
+    })
+  );
+
+  test(
+    'changes tooltip position when SegmentedControl changes',
```
pan-kot: This does not test the position but the content, which conditionally changes when the button is pressed. We can assert the position by comparing the tooltip and target boxes, but such tests already exist for the popover as it is the popover that actually implements the placement logic.

src/tooltip/__integ__/tooltip.test.ts
```diff
@@ -0,0 +1,225 @@
...
+
+      await page.waitForVisible('[data-testid="truncated-text-button"]');
+
+      // Hover over the truncated text button
+      await page.hoverElement('[data-testid="truncated-text-button"]');
+      await page.waitForTooltip();
+
+      // Verify tooltip shows the complete text
+      const content = await page.getTooltipContent();
+      expect(content).toBe('Lorem ipsum dolor sit amet consectetur adipiscing elit sed do eiusmod tempor incididunt');
```
pan-kot: This is just the text provided to the tooltip in the test page - it has nothing to do with component's behaviour.

pan-kot: I think the only relevant test here is verifying that the tooltip shows on hover with the expected content. By doing this test we also confirm that the integ selectors for the tooltip also work.

src/tooltip/__tests__/tooltip.test.tsx
```diff
@@ -0,0 +1,365 @@
...
+}
+
+describe('Tooltip', () => {
+  it('renders text correctly', () => {
+    const wrapper = renderTooltip({ content: 'Value' });
+
+    expect(wrapper.findContent()!.getElement()).toHaveTextContent('Value');
+  });
+
+  it('renders text content correctly', () => {
```
pan-kot: what is different from the 'renders text correctly' test?

src/tooltip/__tests__/tooltip.test.tsx
```diff
@@ -0,0 +1,365 @@
...
+    const wrapper = renderTooltip({ content: 'Success message' });
+
+    expect(wrapper.findContent()!.getElement()).toHaveTextContent('Success message');
+  });
+
+  it('renders number content correctly', () => {
+    const wrapper = renderTooltip({ content: 42 });
```
pan-kot: I don't think this is much needed. It probably makes sense to render a string, a node, and null/undefined.

src/tooltip/__tests__/tooltip.test.tsx
```diff
@@ -0,0 +1,365 @@
...
+        </div>
+      ),
+    });
+
+    const contentElement = wrapper.findContent()!.getElement();
+    expect(contentElement).toHaveTextContent('Bold text and normal text');
+    expect(contentElement.querySelector('strong')).toHaveTextContent('Bold text');
+  });
+
+  it('renders complex nested content', () => {
```
pan-kot: What is different from 'renders React element content correctly'?

src/tooltip/__tests__/tooltip.test.tsx
```diff
@@ -0,0 +1,365 @@
...
+      ),
+      trackKey: 'complex-content',
+    });
+
+    const contentElement = wrapper.findContent()!.getElement();
+    expect(contentElement.querySelector('p')).toHaveTextContent('Paragraph 1');
+    expect(contentElement.querySelectorAll('li')).toHaveLength(2);
+  });
+
+  it('renders arrow', () => {
```
pan-kot: this part is taken care of by the popover - why do we test it here?

src/tooltip/__tests__/tooltip.test.tsx
```diff
@@ -0,0 +1,365 @@
...
+    const wrapper = renderTooltip({ content: 'Value' });
+
+    expect(wrapper.findArrow()).not.toBeNull();
+  });
+
+  it('does not render a header', () => {
```
pan-kot: is this one really needed?

src/tooltip/__tests__/tooltip.test.tsx
```diff
@@ -0,0 +1,365 @@
...
+    expect(wrapper.findTooltip()?.getElement()).toHaveAttribute('role', 'tooltip');
+  });
+
+  it('trackKey is set correctly for strings', () => {
+    const wrapper = renderTooltip({ content: 'Value' });
+
+    expect(wrapper.findTooltip()?.getElement()).toHaveAttribute('data-testid', 'Value');
+  });
+
+  it('trackKey is auto-generated from number content', () => {
```
pan-kot: nit: it is now "auto-generated" but casted

src/tooltip/__tests__/tooltip.test.tsx
```diff
@@ -0,0 +1,365 @@
...
+    expect(wrapper.findTooltip()?.getElement()).toHaveAttribute('data-testid', trackKey);
+  });
+
+  it('explicit trackKey takes precedence over auto-generated', () => {
+    const wrapper = renderTooltip({ content: 'Auto value', trackKey: 'explicit-key' });
+
+    expect(wrapper.findTooltip()?.getElement()).toHaveAttribute('data-testid', 'explicit-key');
+  });
+
+  it('trackKey is required for complex content', () => {
```
pan-kot: is that much different from 'explicit trackKey takes precedence over auto-generated'?

pan-kot: let's instead add a test with ReactNode content and no explicit trackKey to secure that the component renders correctly and does not crash.

src/tooltip/__tests__/tooltip.test.tsx
```diff
@@ -0,0 +1,365 @@
...
+    expect(() => {
+      act(() => {
+        document.body.dispatchEvent(new KeyboardEvent('keydown', { key: 'Escape', bubbles: true }));
+      });
+    }).not.toThrow();
+
+    expect(wrapper.findTooltip()).not.toBeNull();
+  });
+
+  it('supports top position', () => {
```
pan-kot: The position is controlled by the popover, and these tests only ensure that the tooltip renders at all, which is little useful.

If we want to test that the position, size, or other props are passed to the internal popover - we can do just that, by mocking the popover function.

src/tooltip/__tests__/tooltip.test.tsx
```diff
@@ -0,0 +1,365 @@
...
+
+  it('supports left position', () => {
+    const wrapper = renderTooltip({ content: 'Position left', trackKey: 'left', position: 'left' });
+    expect(wrapper.findTooltip()).not.toBeNull();
+    expect(wrapper.findContent()!.getElement()).toHaveTextContent('Position left');
+  });
+
+  it('uses top as default position when not specified', () => {
+    const wrapper = renderTooltip({ content: 'Default position' });
+    // Position is passed to PopoverContainer, which we can verify by checking the component renders
```
pan-kot: let's verify it then or remove the test

src/tooltip/__tests__/tooltip.test.tsx
```diff
@@ -0,0 +1,365 @@
...
+    const element = document.createElement('div');
+    element.textContent = 'Tracked element';
+    document.body.appendChild(element);
+
+    const getTrack = jest.fn(() => element);
+    renderTooltip({ content: 'Tooltip', getTrack });
+
+    expect(getTrack).toHaveBeenCalled();
+
+    document.body.removeChild(element);
```
pan-kot: why do we call remove?

Who-is-PS: I removed it. DOne

src/tooltip/__tests__/tooltip.test.tsx
Who-is-PS: I removed the test completly.

pages/tooltip/component-integration.page.tsx
Who-is-PS: DONE

pages/tooltip/component-integration.page.tsx
Who-is-PS: DONE

pages/tooltip/component-integration.page.tsx
Who-is-PS: I miss understood the truncated text for the token component. I moved the code to the simple.page of tooltip

Who-is-PS: I kinda like this style, but for consitency to other components, I updated everything and removed the header and containers. DONE

pages/tooltip/component-integration.page.tsx
Who-is-PS: Updated accordingly. DONE

pages/tooltip/simple.page.tsx
Who-is-PS: Updated it. done

pages/tooltip/simple.page.tsx
Who-is-PS: Updated it. done

pages/tooltip/simple.page.tsx
Who-is-PS: Updated it. done

pages/tooltip/simple.page.tsx
Who-is-PS: Created it. done

pages/tooltip/simple.page.tsx
Who-is-PS: Good to know. I updated it

pages/tooltip/simple.page.tsx
Who-is-PS: Yeah it is unnecessary. I removed the inline-block display completely.

src/__tests__/required-props-for-components.ts
Who-is-PS: Good point, i updated it

src/column-layout/styles.scss
Who-is-PS: I once had a liniting issue and run npm run lint and the auto fix to it. I revert this change

src/internal/components/tooltip/index.tsx
Who-is-PS: Do you mean the complete internal tooltip? I actually dont know how to handle it. If we now introduce the public tooltip component, should this be completely removed or deprecated ? What do you think about it?

src/internal/components/tooltip/index.tsx
Who-is-PS: Updated the interface of tooltip.

src/token/internal.tsx
Who-is-PS: Tooltip re-renders when tabbing into a token, then tabbing onto the close button inside the token. Is this a bug or problem ? I removed it from now

src/tooltip/__integ__/tooltip.test.ts
Who-is-PS: I removed the other test and keep the one which covers more.

src/tooltip/__integ__/tooltip.test.ts
Who-is-PS: I removed the test, since it already tested in popover

src/tooltip/__tests__/tooltip.test.tsx
Who-is-PS: it is redundant. i removed it

src/tooltip/__tests__/tooltip.test.tsx
Who-is-PS: I will remove this test here and implement one which tests the render of a string, node, and null/undefined

src/tooltip/__tests__/tooltip.test.tsx
Who-is-PS: Also redundant. i removed it

src/tooltip/__tests__/tooltip.test.tsx
Who-is-PS: Removed it. done

src/tooltip/__tests__/tooltip.test.tsx
Who-is-PS: Unnecessary test. I removed it

pages/tooltip/internal-tooltip-examples.page.tsx
```diff
@@ -0,0 +1,391 @@
...
+import Tabs from '~components/tabs';
+import Token from '~components/token';
+
+import { SimplePage } from '../app/templates';
+
+export default function InternalTooltipExamples() {
+  return (
+    <SimplePage
+      title="Tooltip component integration"
+      subtitle="Examples showing how tooltips enhance various components with contextual help, disabled state explanations, and overflow content. Components demonstrated: Button, Select, Multiselect, Token, FileTokenGroup, SegmentedControl, BreadcrumbGroup, Slider, Calendar, DateRangePicker, and Tabs."
```
pan-kot: nit: I would not list all components because: 1) this list can change so that the description can get out of date and 2) this list is easy to obtain from the code, which is enough.

pages/tooltip/simple.page.tsx
```diff
@@ -0,0 +1,401 @@
...
+          <span id="password-description" hidden={true}>
+            Password Rules: Minimum of 8 characters. Include at least one lowercase letter, one uppercase letter, one
+            number and one special character. Unique to this website.
+          </span>
+          {activeTooltip === 'password' && (
+            <Tooltip
+              content={
+                <div>
+                  <strong>Password Rules:</strong>
+                  <br />• Minimum of 8 characters
```
pan-kot: This is still a list yet weird. I would update this example to not use a list at all in favour of some free-text content. If we want to include some styling - we can e.g. use a status indicator in content.

pan-kot: The problem with this example is that it promotes a use case for the tooltip that is likely problematic from the accessibility perspective, at least in the current implementation.

pages/tooltip/simple.page.tsx
```diff
@@ -0,0 +1,401 @@
...
+  onDeactivate: () => void;
+  testId?: string;
+}
+
+function ButtonWithTooltip({
+  tooltipId,
+  buttonText,
+  tooltipContent,
+  position = 'top',
+  activeTooltip,
```
pan-kot: why not making the tooltip state local to the new ButtonWithTooltip component?

pan-kot: e.g. it is done this way in the TruncatedTextExample component

src/tooltip/interfaces.ts
```diff
@@ -0,0 +1,57 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import { NonCancelableEventHandler } from '../internal/events';
+import { PopoverProps } from '../popover/interfaces';
+
+export interface TooltipProps {
+  /**
+   * Content to display in the tooltip.
+   * Accepts any valid React node including strings, numbers, elements, and fragments.
```
pan-kot: We have many components that accept React nodes, but I don't think we ever add this description - let's remove the "Accepts...." part.

src/tooltip/interfaces.ts
```diff
@@ -0,0 +1,57 @@
...
+   */
+  content: React.ReactNode;
+
+  /**
+   * Function that returns the element to track for positioning the tooltip.
```
pan-kot: I think "track for positioning" can be unclear, let's rephrase. We can say that the returner element is the tooltip anchor maybe, or the element the tooltip points to, etc.

src/tooltip/interfaces.ts
```diff
@@ -0,0 +1,57 @@
...
+   * Can return null if the element is not yet mounted or available.
+   */
+  getTrack: () => null | HTMLElement | SVGElement;
+
+  /**
+   * Unique identifier for the tooltip instance. Changing this value will cause the tooltip
+   * to recalculate its position, similar to how React's key prop works.
+   * If not provided and content is a string or number, it will be used as key.
```
pan-kot: The "it will be used as key" is a bit unclear as one may think the "it" refers to the trackKey. Let's rephrase.

src/tooltip/interfaces.ts
```diff
@@ -0,0 +1,57 @@
...
+   * For complex content (elements, fragments), you must provide an explicit trackKey.
```
pan-kot: This sounds too harsh. When it the track key needed at all? I think it is only for cases when the position needs to be recalculated, e.g. for use cases like Slider. It is not necessary for all use cases.

src/tooltip/interfaces.ts
```diff
@@ -0,0 +1,57 @@
...
+   *
+   * The trackKey is also used as the data-testid attribute on the tooltip element,
+   * enabling test utilities to find specific tooltips via TooltipWrapper.findByTrackKey().
+   */
+  trackKey?: string | number;
+
+  /**
+   * Determines where the tooltip is displayed when opened, relative to the trigger. If the tooltip doesn't have enough space to open in this direction, it automatically chooses a better direction based on available space.
+   * @defaultValue 'top'
```
pan-kot: We do not specify default props like that. Instead, the default values are parsed by Documenter from the index file.

src/tooltip/interfaces.ts
```diff
@@ -0,0 +1,57 @@
...
+export namespace TooltipProps {
+  /**
+   * Position of the tooltip relative to the tracked element.
+   */
+  export type Position = PopoverProps.Position;
+}
+
+/**
+ * Internal tooltip props - includes props not exposed in public API.
+ * @internal
```
pan-kot: What is @internal - why is it needed?

src/tooltip/internal.tsx
```diff
@@ -0,0 +1,98 @@
...
+import styles from './styles.css.js';
+
+type InternalTooltipComponentProps = InternalTooltipProps & InternalBaseComponentProps;
+
+export default function InternalTooltip({
+  content,
+  getTrack,
+  trackKey,
+  className,
+  position = 'top',
```
pan-kot, line 25: We already set this in the index file - let's not duplicate the default assignment. Take a look at SomeRequired type that we use in the internal files for that reason.

src/tooltip/styles.scss
```diff
@@ -0,0 +1,8 @@
+/*
+ Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ SPDX-License-Identifier: Apache-2.0
+*/
+
+.root {
+  /* used in tests */
```
pan-kot: Let's move this style to src/tooltip/test-classes/styles.scss - that is our convention for test selectors to make them better stable.

pages/tooltip/component-integration.page.tsx
pan-kot: Thanks - for test pages it is best to avoid unnecessary elements. We do use more complex structure in some pages, but rarely - e.g. to better explain the complex behaviours of runtime apis, error boundaries, etc.

src/internal/components/tooltip/index.tsx
pan-kot: The new tooltip component should replace the old internal one. See how e.g. InternalButton is used in the code - it is imported from the ./button/internal.

src/token/internal.tsx
pan-kot: If that looks like a bug and is reproducible in the main - let's open a bug report and fix it separately and include tests.

src/tooltip/interfaces.ts
Who-is-PS: what do you think about: "If not provided and content is a string or number, the content will be used as the key." ?

src/tooltip/interfaces.ts
Who-is-PS: How about: "Provide an explicit trackKey when the tooltip needs to recalculate its position dynamically, such as when used with components like Slider."

src/tooltip/internal.tsx
Who-is-PS: After discussion on slack, I set the position to top in index.tsx and internal.tsx file and removed it from the interface as a default prop

src/tooltip/styles.scss
Who-is-PS: I moved it to src/tooltip/test-classes/styles.scss  and updated all the imports

src/tooltip/interfaces.ts
Who-is-PS: After discussion on slack, I set the position to top in index.tsx and internal.tsx file and removed it from the interface as a default prop

src/tooltip/interfaces.ts
Who-is-PS: I removed it. We dont use this tag

src/tooltip/interfaces.ts
Who-is-PS: I removed it, after discussion in SO. trackKey is not needed.

src/tooltip/interfaces.ts
Who-is-PS: I removed it, after discussion in SO. trackKey is not needed.

pages/tooltip/internal-tooltip-examples.page.tsx
Who-is-PS: Yeah, i removed the subtitle

src/tooltip/interfaces.ts
Who-is-PS: True, i removed the Accepts.. part

pages/tooltip/simple.page.tsx
Who-is-PS: I updated it and removed it. DONE

src/tooltip/interfaces.ts
Who-is-PS: Update: Some package were failing, so i moved trackkey to internal props

src/tooltip/interfaces.ts
Who-is-PS: Update: Some package were failing, so i moved trackkey to internal props

src/tooltip/interfaces.ts
Who-is-PS: Updated to: * Function that returns the element the tooltip points to.
   * Can return null if the element is not yet mounted or available.

pages/tooltip/simple.page.tsx
Who-is-PS: I updated the example to have a StatusIndicator with normal text

src/tooltip/__tests__/tooltip.test.tsx
Who-is-PS: Updated the test. Now it renders a react node. Also removed the trackKey test

src/tooltip/__tests__/tooltip.test.tsx
Who-is-PS: Removed the test. We use trackKey as an internal prop.

pages/tooltip/simple.page.tsx
Who-is-PS: I updated the example, we are now using an input field

src/internal/components/tooltip/index.tsx
Who-is-PS: i moved trackKey to internal-prop in interface of tooltip.

src/test-utils/dom/internal/tooltip.ts
Who-is-PS: We dont need it, since we are using our public one. I will delete this one

src/test-utils/dom/tooltip/index.ts
Who-is-PS: Yes, I removed this test. We will only have the test findContent()

src/token/__tests__/token-tooltip.test.tsx
Who-is-PS: Done - Removed. We now only have one tooltip and one test wrapper (src/test-utils/dom/tooltip/index.ts)

src/token/internal.tsx
Who-is-PS: Done - Removed the focus-origin check. The issue with tooltip re-rendering when tabbing to the close button should be filed as a separate bug if reproducible on main.

src/tooltip/__integ__/tooltip.test.ts
Who-is-PS: Done - The integ tests now focus on verifying:

1. Tooltip shows on hover with expected content
2. Tooltip hides on mouse leave
3. Tooltip closes on Escape

src/internal/components/tooltip/index.tsx
Who-is-PS: The /internal/tooltip-do-not-use will have to stay, as it is used by the chat components

src/tooltip/styles.scss
```diff
@@ -0,0 +1,4 @@
+/*
```
pan-kot, line 1: let's remove this empty file

src/tooltip/internal.tsx
```diff
@@ -0,0 +1,99 @@
...
+  onEscape,
+  trackKey,
+  __internalRootRef,
+  ...restProps
+}: InternalTooltipComponentProps) {
+  const baseProps = getBaseProps(restProps);
+
+  // Generate data-testid for internal testing
+  if (!trackKey && (typeof content === 'string' || typeof content === 'number')) {
+    trackKey = content;
```
pan-kot: let's remove track-key and allow data-* attributes instead

src/button-group/icon-button-item.tsx
Who-is-PS: Looking at the code and test-utils, only button-group-tooltip is actually used by ButtonGroupWrapper.findTooltip(). The generic .tooltip class was added in the initial commit but never utilized by any test-utils.

pages/tooltip/simple.page.tsx
Who-is-PS: I updated it

src/test-utils/dom/internal/tooltip.ts
Who-is-PS: restored it to the old version. the changes were unnecessary and the internal version of tooltip should not be changed

pages/tooltip/simple.page.tsx
```diff
@@ -0,0 +1,411 @@
...
+  const passwordRef = useRef<HTMLInputElement>(null);
+  const [isInteractingWithTooltip, setIsInteractingWithTooltip] = useState(false);
+
+  const handleFocusTooltip = (id: typeof activeTooltip) => {
+    setActiveTooltip(id);
+  };
+
+  const handleBlurTooltip = () => {
+    // Only close if not currently interacting with tooltip
+    if (!isInteractingWithTooltip) {
```
pan-kot, line 45: What is this condition for?

pages/tooltip/simple.page.tsx
```diff
@@ -0,0 +1,411 @@
...
+            </button>
+            <span id="code-description" hidden={true}>
+              {`const AWS = require('aws-sdk'); AWS.config.update({ region: 'us-west-2' });`}
+            </span>
+            {activeTooltip === 'code' && (
+              <div onMouseDown={handleTooltipMouseDown} onMouseUp={handleTooltipMouseUp}>
+                <Tooltip
+                  content={
+                    <div>
+                      <code
```
pan-kot: Nit: we can use `<Box variant="awsui-inline-code">` for this instead.

pages/tooltip/simple.page.tsx
```diff
@@ -0,0 +1,411 @@
...
+                />
+              </div>
+            )}
+          </div>
+        </SpaceBetween>
+
+        <div
+          onMouseEnter={() => handleMouseEnterTooltip('password')}
+          onMouseLeave={() => setActiveTooltip(null)}
+          style={{ display: 'inline-block' }}
```
pan-kot: nit: we can use `<span/>` instead of `<div/>` here

pages/tooltip/simple.page.tsx
Who-is-PS: Prevents the tooltip from closing when clicking inside it (e.g., to select text). Without this, mousedown on the tooltip triggers a blur on the trigger element, which would close the tooltip mid-interaction. This was reported as Bug by Gethin

pages/tooltip/simple.page.tsx
Who-is-PS: Done

src/tooltip/styles.scss
Who-is-PS: Done

src/tooltip/internal.tsx
Who-is-PS: I would prefer to remove it after the release, since the CloudEditor package depends on it.
By moving trackKey to internal props, we avoid breaking dependent packages.

What do you think?

pan-kot: let's then add a code comment and remove the usage of it from everywhere but button dropdown

pages/tooltip/simple.page.tsx
pan-kot: Should we include that to the playground examples and document then?

src/button-group/icon-button-item.tsx
pan-kot: let's remove if it is not used then

src/tooltip/__integ__/tooltip.test.ts
```diff
@@ -0,0 +1,136 @@
...
+      await expect(page.isTooltipVisible()).resolves.toBe(true);
+
+      await browser.keys('Escape');
+      await page.waitForTooltipToDisappear();
+      await expect(page.isTooltipVisible()).resolves.toBe(false);
+    })
+  );
+
+  test(
+    'tooltip content can be selected and copied - medium length',
```
pan-kot, line 78: what is the purpose of this test?

src/tooltip/__tests__/tooltip.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+import tooltipStyles from '../../../lib/components/tooltip/test-classes/styles.selectors.js';
+
+class TooltipInternalWrapper extends PopoverWrapper {
+  findTooltip(): ElementWrapper | null {
+    return createWrapper().findByClassName(tooltipStyles.root);
+  }
+  findContent(): ElementWrapper | null {
+    return createWrapper().findByClassName(styles.content);
+  }
+  findArrow(): ElementWrapper | null {
```
pan-kot: this is not used

src/tooltip/__tests__/tooltip.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+    const contentElement = wrapper.findContent()!.getElement();
+    expect(contentElement).toHaveTextContent('Bold text and normal text');
+    expect(contentElement.querySelector('strong')).toHaveTextContent('Bold text');
+  });
+
+  it('has tooltip role attribute', () => {
+    const wrapper = renderTooltip({ content: 'Value' });
+
+    expect(wrapper.findTooltip()?.getElement()).toHaveAttribute('role', 'tooltip');
```
pan-kot: why do we find tooltip by class name instead of doing `createWrapper().findTooltip()` - the official method?

src/tooltip/__tests__/tooltip.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+    />
+  );
+  return new TooltipInternalWrapper(container);
+}
+
+describe('Tooltip', () => {
+  it('renders text content correctly', () => {
+    const wrapper = renderTooltip({ content: 'Success message' });
+
+    expect(wrapper.findContent()!.getElement()).toHaveTextContent('Success message');
```
pan-kot: why do we use findContent() and not findTooltip() here?

src/tooltip/__tests__/tooltip.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+  findTooltip(): ElementWrapper | null {
+    return createWrapper().findByClassName(tooltipStyles.root);
+  }
+  findContent(): ElementWrapper | null {
+    return createWrapper().findByClassName(styles.content);
+  }
+  findArrow(): ElementWrapper | null {
+    return createWrapper().findByClassName(styles.arrow);
+  }
+  findHeader(): ElementWrapper | null {
```
pan-kot: this is not used

src/tooltip/__integ__/tooltip.test.ts
Who-is-PS: This ensures the tooltip persists when hovered, allowing users to select and copy the content, before that tooltip would disappear.

src/tooltip/__tests__/tooltip.test.tsx
Who-is-PS: i removed it

src/tooltip/__tests__/tooltip.test.tsx
Who-is-PS: I removed it

pages/tooltip/simple.page.tsx
Who-is-PS: This is an edge case that developers can implement themselves if needed. Other libraries don't include this functionality, and it adds complexity without clear benefit. So I would leave it out.

pages/tooltip/simple.page.tsx
Who-is-PS: Yes, i updated it

src/tooltip/internal.tsx
Who-is-PS: Added a code comment and removed the function and prop from public tooltip

src/tooltip/__tests__/tooltip.test.tsx
Who-is-PS: I updated it with the official method, it was a mistake from my side.

src/tooltip/__tests__/tooltip.test.tsx
Who-is-PS: I will update the test.

src/tooltip/internal.tsx
Who-is-PS: I will create a CR for the CloudEditor team to update their testing approach. Since TrackKey has been removed, the CR will instruct them to utilize Cloudscape test utilities to locate tooltips, rather than relying on data-testid attributes.

src/button-group/icon-button-item.tsx
Who-is-PS: I removed the unused tooltip class from all components where it was not being referenced in the codebase.

pages/tooltip/simple.page.tsx
pan-kot: Let's then at least explain in the code comments more clearly the purpose of the code (same as you explained in this thread).

src/tooltip/__tests__/tooltip.test.tsx
```diff
@@ -0,0 +1,164 @@
...
+  });
+
+  it('works without onEscape callback', () => {
+    const wrapper = renderTooltip({ content: 'Value' });
+
+    expect(() => {
+      act(() => {
+        document.body.dispatchEvent(new KeyboardEvent('keydown', { key: 'Escape', bubbles: true }));
+      });
+    }).not.toThrow();
```
pan-kot: how can it possibly throw?

src/tooltip/__tests__/tooltip.test.tsx
```diff
@@ -0,0 +1,164 @@
...
+    expect(keydownEvent.stopPropagation).toHaveBeenCalled();
+    expect(onEscape).toHaveBeenCalled();
+  });
+
+  it('does not call onEscape when other keys are pressed', () => {
+    const onEscape = jest.fn();
+
+    renderTooltip({ content: 'Value', onEscape });
+
+    act(() => {
```
pan-kot: this can be replaced with the fireEvent util and w/o act (search in code for examples)

src/tooltip/__tests__/tooltip.test.tsx
Who-is-PS: This was a little bit to defensive. I updated the file

src/tooltip/__tests__/tooltip.test.tsx
Who-is-PS: I replaced it with the fireEvent

src/tooltip/interfaces.ts
Who-is-PS: I submitted a CR to update the cloudeditor test suite, which has been approved.

src/tooltip/internal.tsx
Who-is-PS: I submitted a CR to update the cloudeditor test suite, which has been approved.

pages/tooltip/simple.page.tsx
Who-is-PS: Included comments to explain the code

PR #4097, opened 2025-12-09
chore: add test page for virtual scroll

mxschll:
### Description

Adds a test page for virtual scroll selectables for screenshot tests:

- Select
- Multiselect
- Multuselect with Select All
- Autosuggest
- PropertyFilter

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/multiselect/virtual-scroll.page.tsx
+++ b/pages/multiselect/virtual-scroll.page.tsx
@@ -16,8 +16,8 @@ const options: MultiselectProps.Options = Array.from({ length: 1000 }, (_, i) =>
 }));
 
 export default function () {
-  const { urlParams } = useContext(AppContext as DemoContext);
-  const [selectedType, setSelectedType] = useState(urlParams.type || 'multiselect');
+  const { urlParams, setUrlParams } = useContext(AppContext as DemoContext);
+  const selectedType = urlParams.type || 'multiselect';
 
   const [selectedMulti, setSelectedMulti] = useState<MultiselectProps.Options>(options.slice(0, 2));
   const [selectedMultiWithSelectAll, setSelectedMultiWithSelectAll] = useState<MultiselectProps.Options>([]);
@@ -30,7 +30,9 @@ export default function () {
       settings={
         <SegmentedControl
           selectedId={selectedType}
-          onChange={({ detail }) => setSelectedType(detail.selectedId)}
+          onChange={({ detail }) => {
+            setUrlParams({ type: detail.selectedId });
+          }}
           options={[
             { id: 'multiselect', text: 'Multiselect' },
             { id: 'multiselect-select-all', text: 'Multiselect with Select All' },


pages/selectable-item/virtual-scroll.page.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+  const [selected, setSelected] = useState<SelectProps['selectedOption']>(null);
+  const [selectedMulti, setSelectedMulti] = useState<MultiselectProps.Options>(options.slice(0, 2));
+  const [selectedMultiWithSelectAll, setSelectedMultiWithSelectAll] = useState<MultiselectProps.Options>([]);
+  const [autosuggestValue, setAutosuggestValue] = useState('');
+  const [query, setQuery] = useState<PropertyFilterProps.Query>({ tokens: [], operation: 'and' });
+
+  return (
+    <SimplePage title="Virtual Scroll" i18n={{}} screenshotArea={{}}>
+      <div
```
jperals: This section can instead go in the `settings` param of the SimplePage. This way, it will not be part of the screenshot (unless you split the page for each component as suggested above, in which case you don't need this at all)

pages/selectable-item/virtual-scroll.page.tsx
```diff
@@ -0,0 +1,145 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
jperals: Why not just make 5 different pages, one for each component? This way the tests will be simpler, without having to interact with the page or rely on specific elements in it which are not related to the components being tested.

mxschll: I added query params, so we actually don't need to interact with the segmented control like this: `/?type=select`. I wanted to group these specific components into one page because I want to run a specific test that covers the missing virtual scroll behavior for all of them. It seemed a bit too much to create new pages for each of them instead of having them grouped, but I can do that as well.

pages/selectable-item/virtual-scroll.page.tsx
mxschll: What section exactly? Do you mean the segmented control?

jperals: Yes

pages/selectable-item/virtual-scroll.page.tsx
jperals: OK got it. Personally I would also prefer the pages be in their respective component pages instead of this directory which belongs to an internal component, but not a blocking requirement, I leave this to your judgement

mxschll: Yea you're right, I split it up into separate files.

pages/multiselect/virtual-scroll.page.tsx
```diff
@@ -0,0 +1,86 @@
...
+type DemoContext = React.Context<AppContextType<{ type?: string }>>;
+
+const options: MultiselectProps.Options = Array.from({ length: 1000 }, (_, i) => ({
+  value: `${i}`,
+  label: `Option ${i + 1}`,
+}));
+
+export default function () {
+  const { urlParams } = useContext(AppContext as DemoContext);
+  const [selectedType, setSelectedType] = useState(urlParams.type || 'multiselect');
```
jperals: Nitpicks:
- Usually when we have parameters that can be control via both the URL and the UI, in the UI event callback we set back the URL parameters instead of managing an internal state. This way the URL and the internal state are always in sync
- For simplicity, instead of `multiselect` and `multiselect-select-all`, the param could be a boolean

(Or alternatively, both variations could be rendered separately, for example as a permutations page of two permutations)

mxschll: Good points, I addressed the setting of url parameters in the latest revision.

jperals (approved):

PR #4101, opened 2025-12-10
chore: Adds an integ test to secure uap buttons fix

pan-kot:
### Description

A follow-up test for https://github.com/cloudscape-design/components/pull/4094

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/table/__integ__/resizable-columns.test.ts
+++ b/src/table/__integ__/resizable-columns.test.ts
@@ -118,6 +118,10 @@ class TablePage extends BasePageObject {
       timeoutMsg: `Column at index "${columnIndex}" should have width "${expected}"`,
     });
   }
+
+  getDocumentScrollWidth() {
+    return this.browser.execute(() => document.documentElement.scrollWidth);
+  }
 }
 
 interface PageConfig {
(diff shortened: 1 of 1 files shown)


src/table/__integ__/resizable-columns.test.ts
```diff
@@ -368,3 +372,20 @@ test.each([false, true])(
...
+
+    // Shows UAP buttons
+    await page.click(tableWrapper.findColumnResizer(4).toSelector());
+    // The UAP buttons move outside the screen boundaries. Not ideal - but expected, while the buttons are visible.
+    await page.setWindowSize({ ...defaultScreen, width: defaultScreen.width - 100 });
+    // Hides UAP buttons
+    await page.click(tableWrapper.findHeaderSlot().toSelector());
+
+    const widthAfter = await page.getDocumentScrollWidth();
+    expect(widthAfter).toBe(widthBefore - 100);
```
gethinwebster, line 389: this doesn't seem to fail when running it against reverted changes, I also tried tweaking the `100` but couldn't easily find something that does fail when expected

pan-kot: Yes, you are right!

While it works when checking for the same condition and the UAP buttons are still visible - it does not actually capture the old bug. When we use setVindowSize() with browser-test-tools, it works differently from how it is in reality, so the effect is not there, and when the UAP buttons are shown and then hidden within the table - it does not cause a regression already (once the buttons are hidden).

I will cancel this PR then and will go with no test coverage for the bugfix, as I ran out of ideas.

gethinwebster: possibly a visual regression test that tries to scroll to the side after showing/hiding buttons and resizing screen? but I think it's not 100% crucial to have test coverage for it

PR #4102, opened 2025-12-11
fix: Fixes table resizer overflow causing unwanted scroll

pan-kot:
### Description

This fixes the issue when a table with variant full-page, embedded or borderless would be horizontally scrollable when it should not.

Rel: [AWSUI-61443]

Before:

https://github.com/user-attachments/assets/b68026f2-150f-4de1-aa7a-d8b4e1fb9951

After:

The table is no longer horizontally scrollable when in full-page, embedded or borderless variant.

### How has this been tested?

TBA

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/table/resizer/index.tsx
+++ b/src/table/resizer/index.tsx
@@ -312,7 +312,9 @@ export function Resizer({
   ]);
 
   useEffect(() => {
-    return () => clearTimeout(autoGrowTimeout.current);
+    if (isDragging) {
+      return () => clearTimeout(autoGrowTimeout.current);
+    }
   }, [isDragging]);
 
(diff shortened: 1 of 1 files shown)


src/table/resizer/styles.scss
```diff
@@ -89,10 +94,6 @@ th:not(:last-child) > .divider-disabled {
   }
 }
 
-th:last-child > .resizer:not(.is-visual-refresh) {
```
pan-kot, line 92: this was some old rule that never applied (because the descendant of the th is `.resize-wrapper`, not `.resizer`.

Who-is-PS (approved):

PR #4104, opened 2025-12-11
feat: Add datePlaceholder and timePlaceholder props to date-range-picker component's i18nStrings

Who-is-PS:
### Description

Add configurable `datePlaceholder` and `timePlaceholder` props to the date-range-picker component's i18nStrings, allowing developers to customize placeholder text for date and time inputs in absolute mode.

`AWSUI-61514`

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/date-range-picker/__tests__/date-range-picker-absolute.test.tsx
+++ b/src/date-range-picker/__tests__/date-range-picker-absolute.test.tsx
@@ -1032,13 +1032,14 @@ describe('Date range picker', () => {
 
       describe('date input placeholder', () => {
         test('month picker derives placeholder from date placeholder by removing day part', () => {
+          // Uses German-style placeholder to verify i18n value is actually used (not the default YYYY-MM)
           const { wrapper } = renderDateRangePicker({
             ...defaultProps,
             granularity: 'month',
             dateInputFormat: 'iso',
             i18nStrings: {
               ...i18nStrings,
-              isoDatePlaceholder: 'YYYY-MM-DD',
+              isoDatePlaceholder: 'JJJJ-MM-TT',
             },
           });
 
@@ -1046,11 +1047,11 @@ describe('Date range picker', () => {
 
           expect(wrapper.findDropdown()!.findStartDateInput()!.findNativeInput().getElement()).toHaveAttribute(
             'placeholder',
-            'YYYY-MM'
+            'JJJJ-MM'
           );
           expect(wrapper.findDropdown()!.findEndDateInput()!.findNativeInput().getElement()).toHaveAttribute(
             'placeholder',
-            'YYYY-MM'
+            'JJJJ-MM'
           );
         });
 


src/date-range-picker/interfaces.ts
```diff
@@ -570,6 +570,22 @@ export namespace DateRangePickerProps {
      * @i18n
      */
     previousYearAriaLabel?: string;
+
+    /**
+     * Placeholder text for date inputs in absolute mode.
+     * Should match the expected date format (e.g., "YYYY-MM-DD", "JJJJ-MM-TT" for German).
```
gethinwebster: use "for example" instead of "e.g."

src/i18n/messages/all.ko.json
```diff
@@ -158,6 +158,8 @@
     "i18nStrings.endMonthLabel": "마지막 월",
     "i18nStrings.endDateLabel": "종료 날짜",
     "i18nStrings.endTimeLabel": "종료 시간",
+    "i18nStrings.datePlaceholder": "YYYY-MM-DD",
```
NathanZlion, line 161: Why is this YYYY-MM-DD instead of YYYY/MM/DD like the others and like the `dateTimeConstraintText` i18strings specifies 2 lines below

src/i18n/messages/all.zh-CN.json
```diff
@@ -158,6 +158,8 @@
     "i18nStrings.endMonthLabel": "结束月底",
     "i18nStrings.endDateLabel": "结束日期",
     "i18nStrings.endTimeLabel": "结束时间",
+    "i18nStrings.datePlaceholder": "YYYY-MM-DD",
```
NathanZlion, line 161: same comment as above, I think this should be consistent with the i18nStrings.dateTimeConstraintText

src/i18n/messages/all.zh-TW.json
```diff
@@ -158,6 +158,8 @@
     "i18nStrings.endMonthLabel": "結束月份",
     "i18nStrings.endDateLabel": "結束日期",
     "i18nStrings.endTimeLabel": "結束時間",
+    "i18nStrings.datePlaceholder": "YYYY-MM-DD",
```
NathanZlion, line 161: same comment

src/i18n/messages/all.ar.json
```diff
@@ -158,6 +158,8 @@
     "i18nStrings.endMonthLabel": "نهاية الشهر",
     "i18nStrings.endDateLabel": "تاريخ الانتهاء",
     "i18nStrings.endTimeLabel": "وقت الانتهاء",
+    "i18nStrings.datePlaceholder": "YYYY-MM-DD",
```
NathanZlion, line 161: Why are we using / for some and - for others, IMO it should be consistent with the i18Strings.dateConstraintText

NathanZlion:
I think I understand why now, there's a property `dateInputFormat` which could be `slashed` or `iso` which is not being accounted for here, so the separator is not being considered for the i18strings, I suggest you look into that and incorporate that into the i18Strings,

src/i18n/messages/all.ko.json
Who-is-PS: I just used the translation from AWS-UI-Components-I18n package. I see that there are some differences. @gethinwebster should we make it consistent across the codebase? What do you think?

src/i18n/messages/all.ar.json
gethinwebster: I think there need to be two different translations: `datePlaceholder` and `slashedDatePlaceholder`, similar to how there is `dateTimeConstraintText` and `slashedDateTimeConstraintText`

(although in this specific case of `ar`, it _also_ looks like `dateTimeConstraintText` is incorrect: it should have YYYY-MM-DD)

Who-is-PS: I've created a CR in the i18n package to add two new props:

- `isoDatePlaceholder` (YYYY-MM-DD with dashes)
- `slashedDatePlaceholder` (YYYY/MM/DD with slashes)

This follows the same pattern as `dateConstraintText`/`isoDateConstraintText`/`slashedDateConstraintText`.

The original `datePlaceholder` will remain as-is for backward compatibility. Waiting on translations from Totoro team.

Who-is-PS:
Added `isoDatePlaceholder` (YYYY-MM-DD with dashes) and `slashedDatePlaceholder` (YYYY/MM/DD with slashes) translations to the date-range-picker component.

Totoro translations merged into main via CR-252051089.

src/i18n/messages/all.zh-CN.json
Who-is-PS: Added isoDatePlaceholder (YYYY-MM-DD with dashes) and slashedDatePlaceholder (YYYY/MM/DD with slashes) translations to the date-range-picker component.

src/i18n/messages/all.zh-TW.json
Who-is-PS: Added isoDatePlaceholder (YYYY-MM-DD with dashes) and slashedDatePlaceholder (YYYY/MM/DD with slashes) translations to the date-range-picker component.

Who-is-PS:
> I think I understand why now, there's a property `dateInputFormat` which could be `slashed` or `iso` which is not being accounted for here, so the separator is not being considered for the i18strings, I suggest you look into that and incorporate that into the i18Strings,

I've added format-specific placeholder strings to address this:

- `isoDatePlaceholder` for the ISO format (YYYY-MM-DD with dashes)
- `slashedDatePlaceholder` for the slashed format (YYYY/MM/DD with slashes)

This follows the same pattern already established with `isoDateConstraintText`/`slashedDateConstraintText` and `isoDateTimeConstraintText`/`slashedDateTimeConstraintText`.

Regarding the Totoro translations - I've noticed some inconsistencies across locales. I'll do a thorough audit of all date-related i18n strings and update any mismatches to ensure consistency between the placeholder and constraint text formats.

src/date-range-picker/interfaces.ts
```diff
@@ -570,6 +570,38 @@ export namespace DateRangePickerProps {
...
      */
     previousYearAriaLabel?: string;
+
+    /**
+     * Placeholder text for date inputs in absolute mode.
+     * Should match the expected date format (for example "YYYY-MM-DD", "JJJJ-MM-TT" for German).
+     * Used for both start and end date inputs.
+     * @i18n
+     */
+    datePlaceholder?: string;
```
gethinwebster: hadn't realized we're adding this totally new, in that case I don't think we need `datePlaceholder`, only `iso*` and `slashed*` versions?

src/date-range-picker/calendar/range-inputs.tsx
```diff
@@ -34,7 +34,11 @@ export default function RangeInputs({
   const showTimeInput = !dateOnly && !isMonthPicker;
   const isIso = dateInputFormat === 'iso';
   const separator = isIso ? '-' : '/';
-  const dateInputPlaceholder = `YYYY${separator}MM${isMonthPicker ? '' : `${separator}DD`}`;
+  const defaultDateInputPlaceholder = `YYYY${separator}MM${isMonthPicker ? '' : `${separator}DD`}`;
+  const datePlaceholder = isIso
```
gethinwebster: I think there's a regression here: with `granularity="month"` and using the i18nProvider, previously the placeholder would have been `YYYY/MM` (regardless of language, from the hardcoding) but now it would show `YYYY/MM/DD` (or e.g. for german, `JJJJ/MM/TT`, from the i18nProvider).

src/date-range-picker/interfaces.ts
Who-is-PS: Updated it.

src/date-range-picker/calendar/range-inputs.tsx
Who-is-PS: Fixed by deriving the month placeholder from the date placeholder, when isMonthPicker is true, we split by the separator and take the first two parts (year + month), removing the day portion. So "JJJJ-MM-TT" becomes "JJJJ-MM" for German.

src/date-range-picker/__tests__/date-range-picker-absolute.test.tsx
```diff
@@ -1030,6 +1030,78 @@ describe('Date range picker', () => {
...
 
+      describe('date input placeholder', () => {
+        test('month picker derives placeholder from date placeholder by removing day part', () => {
+          const { wrapper } = renderDateRangePicker({
+            ...defaultProps,
+            granularity: 'month',
+            dateInputFormat: 'iso',
+            i18nStrings: {
+              ...i18nStrings,
+              isoDatePlaceholder: 'YYYY-MM-DD',
```
gethinwebster: it would be good to have a test that uses characters other than YYYY-MM-DD, just to ensure that the default placeholder isn't being used by accident

Who-is-PS: Added German placeholder. Dne

PR #4108, opened 2025-12-15
feat: Custom svg icons for runtime drawers' headerActions

georgylobko:
### Description

Added support for custom svg icons for runtime drawers' `headerActions` 
Screenshot 2025-12-15 at 12 56 21 PM

### How has this been tested?

u tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/runtime-drawer/index.tsx
+++ b/src/app-layout/runtime-drawer/index.tsx
@@ -2,9 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 import React, { useContext, useEffect, useRef } from 'react';
 
-import { warnOnce } from '@cloudscape-design/component-toolkit/internal';
-
-import { ButtonGroupProps, ItemOrGroupRuntime, ItemRuntime } from '../../button-group/interfaces';
+import { ButtonGroupProps, ItemRuntime } from '../../button-group/interfaces';
 import { fireNonCancelableEvent, NonCancelableEventHandler } from '../../internal/events';
 import {
   DrawerConfig as RuntimeDrawerConfig,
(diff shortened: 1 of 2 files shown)


src/app-layout/runtime-drawer/index.tsx
```diff
@@ -83,14 +83,8 @@ function RuntimeDrawerHeader({ mountHeader, unmountHeader }: RuntimeContentHeade
...
-function checkForUnsupportedProps(headerActions: ReadonlyArray<ButtonGroupProps.Item>) {
-  const unsupportedProps = new Set([
-    'iconSvg',
-    'popoverFeedback',
-    'pressedIconSvg',
-    'popoverFeedback',
-    'pressedPopoverFeedback',
-  ]);
+function checkForUnsupportedProps(headerActions: ReadonlyArray<ItemOrGroupRuntime>) {
+  const unsupportedProps = new Set(['popoverFeedback', 'popoverFeedback', 'pressedPopoverFeedback']);
```
NathanZlion: Why do we have popoverFeedback twice in this set, I'm not sure if this is a mistake or no...? looks like it's also part of the previous code

just-boris: What happens if any of these properties will be used? I think it will crash again under the same conditions.

I think instead of a semi-silent warning we want to throw an error here. What do you think?

src/button-group/interfaces.ts
```diff
@@ -112,6 +112,20 @@ export interface InternalIconToggleButton extends ButtonGroupProps.IconToggleBut
   analyticsAction?: string;
 }
 
+export interface IconButtonRuntime extends Omit<ButtonGroupProps.IconButton, 'iconSvg'> {
+  iconSvg?: string;
+}
+export interface IconToggleButtonRuntime extends Omit<ButtonGroupProps.IconToggleButton, 'iconSvg' | 'pressedIconSvg'> {
```
just-boris: Similar to the other comment, how about omitting unsupported properties like `popoverFeedback` from the types?

georgylobko: Sounds good. Done

src/app-layout/runtime-drawer/index.tsx
georgylobko: Omitted these props and removed this check

PR #4114, opened 2025-12-16
feat: Add panel-layout component

gethinwebster:
### Description

Create new "panel layout" component

Related links, issue #, if available: AWSUI-61325

### How has this been tested?

Added unit and integ tests, will add screenshot tests separately

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/panel-layout/internal.tsx
+++ b/src/panel-layout/internal.tsx
@@ -1,9 +1,9 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
-import React from 'react';
+import React, { useEffect } from 'react';
 import clsx from 'clsx';
 
-import { useMergeRefs } from '@cloudscape-design/component-toolkit/internal';
+import { useMergeRefs, useStableCallback } from '@cloudscape-design/component-toolkit/internal';
 
 import { useResize } from '../app-layout/visual-refresh-toolbar/drawer/use-resize';
 import { getBaseProps } from '../internal/base-component';
@@ -76,13 +76,12 @@ const InternalPanelLayout = React.forwardRef<PanelLayoutProps.Ref, InternalPanel
     const actualMinSize = minPanelSize ?? 0;
     const actualPanelSize = Math.max(Math.min(panelSize, actualMaxSize), actualMinSize);
 
-    const notifiedContainerWidth = React.useRef(containerWidth);
-    const notifiedPanelSize = React.useRef(actualPanelSize);
-    if (notifiedContainerWidth.current !== containerWidth || notifiedPanelSize.current !== actualPanelSize) {
-      notifiedContainerWidth.current = containerWidth;
-      notifiedPanelSize.current = actualPanelSize;
-      fireNonCancelableEvent(onLayoutChange, { totalSize: containerWidth, panelSize: actualPanelSize });
-    }
+    const stableLayoutChangeEvent = useStableCallback((details: PanelLayoutProps.PanelResizeDetail) =>
+      fireNonCancelableEvent(onLayoutChange, details)
+    );
+    useEffect(() => {
+      stableLayoutChangeEvent({ totalSize: containerWidth, panelSize: actualPanelSize });
+    }, [containerWidth, actualPanelSize, stableLayoutChangeEvent]);
 
     const resizeHandlePosition = panelPosition === 'side-end' ? 'side' : panelPosition;
     const resizeProps = useResize({
--- a/src/panel-layout/__tests__/panel-layout.test.tsx
+++ b/src/panel-layout/__tests__/panel-layout.test.tsx
@@ -646,7 +646,6 @@ describe('PanelLayout Component', () => {
       mockUseContainerWidth.mockReturnValue([1000, React.createRef()]);
       rerender(<PanelLayout panelContent="Panel" mainContent="Main" onLayoutChange={onLayoutChange} />);
 
-      expect(onLayoutChange).toHaveBeenCalledTimes(1);
       expect(onLayoutChange).toHaveBeenCalledWith(
         expect.objectContaining({
           detail: { totalSize: 1000, panelSize: 200 },
@@ -679,7 +678,6 @@ describe('PanelLayout Component', () => {
         />
       );
 
-      expect(onLayoutChange).toHaveBeenCalledTimes(1);
       expect(onLayoutChange).toHaveBeenCalledWith(
         expect.objectContaining({
           detail: { totalSize: CONTAINER_WIDTH, panelSize: 400 },
@@ -733,7 +731,6 @@ describe('PanelLayout Component', () => {
         />
       );
 
-      expect(onLayoutChange).toHaveBeenCalledTimes(1);
       expect(onLayoutChange).toHaveBeenCalledWith(
         expect.objectContaining({
           detail: { totalSize: CONTAINER_WIDTH, panelSize: 350 },


pages/panel-layout/nested.page.tsx
```diff
@@ -0,0 +1,158 @@
...
+};
+
+export default function NestedPanelLayoutPage() {
+  const {
+    urlParams: { outerPanelPosition = 'side-start', innerPanelPosition = 'side-start', nestedLocation = 'panel' },
+    setUrlParams,
+  } = useContext(AppContext as PageContext);
+
+  return (
+    <I18nProvider messages={[messages]} locale="en">
```
pan-kot: nit: There is now a helper component `SimplePage` for rendering test pages w/o app layout. It includes i18n and screenshot area:

```
<SimplePage title="Nested Panel Layout Demo" i18n={{}} screenshotArea={{}}>
  <>...</>
</SimplePage>
```

src/i18n/messages/all.en.json
```diff
@@ -180,6 +180,10 @@
     "i18nStrings.descriptionText": "{hasFeedback, select, true {Refresh to try again. We are tracking this issue, but you can share <Feedback>more information here</Feedback>.} other {Refresh to try again.}}",
     "i18nStrings.refreshActionText": "Refresh page"
   },
+  "features-notification-drawer": {
```
pan-kot: these are not present in other locales

src/panel-layout/test-classes/styles.scss
```diff
@@ -0,0 +1,11 @@
+/*
+ Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ SPDX-License-Identifier: Apache-2.0
+*/
+
+.root,
+.panel,
+.content,
+.slider {
```
pan-kot: nit: why do we call this `slider` and not `resize-handle`?

src/panel-layout/interfaces.ts
```diff
@@ -0,0 +1,115 @@
...
+   */
+  mainContent: ReactNode;
+
+  /**
+   * Makes the panel content focusable. This should be used if there are no focusable elements
+   * inside the panel, to ensure it is keyboard scrollable.
+   *
+   * Provide either `{ariaLabel: "Panel label"}` or `{ariaLabelledby: "panel-label-id"}`
+   */
+  panelFocusable?: PanelLayoutProps.FocusableConfig;
```
pan-kot, line 60: This prop is not represented in the test pages. I recommend adding it to the existing one, e.g. conditionally when using "Long main content" or "Long panel content" setting (and remove the button from the content).

src/panel-layout/interfaces.ts
```diff
@@ -0,0 +1,115 @@
...
+   *
+   * The actual size may vary, depending on `minPanelSize` and `maxPanelSize`.
+   */
+  defaultPanelSize?: number;
+
+  /**
+   * Size of the panel. If provided, and panel is resizable, the component is controlled,
+   * so you must also provide `onPanelResize`.
+   *
+   * The actual size may vary, depending on `minPanelSize` and `maxPanelSize`.
```
pan-kot: The "may vary" sounds unclear. Does it only mean that the size can be adjusted to either minPanelSize or maxPanelSize, or are there some other rules at play?

src/panel-layout/interfaces.ts
```diff
@@ -0,0 +1,115 @@
...
+   */
+  panelSize?: number;
+
+  /**
+   * The minimum size of the panel.
```
pan-kot, line 30: Maybe we can state it here as: "when provided panelSize is less than minPanelSize the minPanelSize will be used instead" (or alike).

src/panel-layout/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+export { PanelLayoutProps };
+
+const PanelLayout = React.forwardRef<PanelLayoutProps.Ref, PanelLayoutProps>(function PanelLayout(
+  { panelPosition = 'side-start', resizable = false, display = 'all', ...props },
+  ref
+) {
+  const baseComponentProps = useBaseComponent('PanelLayout', {
+    props: { panelPosition, resizable },
+    metadata: {
```
pan-kot, line 20: nit: I would include metadata for panel/main focusable settings

src/panel-layout/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+      hasDefaultSize: props.defaultPanelSize !== undefined,
+      hasSize: props.panelSize !== undefined,
+      hasMinSize: props.minPanelSize !== undefined,
```
pan-kot: I think it is better to call these `hasPanelSize`, `hasPanelMinSize`, and `hasPanelMaxSize`- in case we add some main-related sizes in the future.

Also, the `hasPanelSize` can probably be renamed to `isControllable` to better reflect the idea.

src/panel-layout/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+      hasMaxSize: props.maxPanelSize !== undefined,
+    },
+  });
+  return (
+    <InternalPanelLayout
+      ref={ref}
+      panelPosition={panelPosition}
+      resizable={resizable}
+      display={display}
+      {...props}
```
pan-kot: Let's use `...getExternalProps(props)` here to prevent possible use of internal properties via the public interface.

src/panel-layout/internal.tsx
```diff
@@ -0,0 +1,155 @@
...
+    const sliderRef = React.useRef<HTMLDivElement>(null);
+    const panelRef = React.useRef<HTMLDivElement>(null);
+    const [containerWidth, rootRef] = useContainerWidth();
+
+    React.useImperativeHandle(
+      ref,
+      () => ({
+        focusResizeHandle() {
+          if (resizable && display === 'all' && sliderRef.current) {
+            sliderRef.current.focus();
```
pan-kot: nit: I think we can avoid the checks and do `sliderRef.current?.focus()` as the resize handle should not be rendered when not resizable anyways. This reduces unnecessary logical branching.

src/panel-layout/internal.tsx
```diff
@@ -0,0 +1,155 @@
...
+          }
+        },
+      }),
+      [resizable, display]
+    );
+
+    const [panelSize = DEFAULT_PANEL_SIZE, setPanelSize] = useControllable(
```
pan-kot, line 64: nit: can we use `useControllableState()` from the component toolkit here instead?

src/panel-layout/internal.tsx
```diff
@@ -0,0 +1,155 @@
...
+        setPanelSize(size);
+        fireNonCancelableEvent(onPanelResize, { totalSize: containerWidth, panelSize: size });
+      },
+    });
+
+    const mergedRef = useMergeRefs(rootRef, __internalRootRef, ref);
+
+    const wrappedPanelContent = (
+      <div
+        className={styles['panel-content']}
```
pan-kot: The `testStyles.panel` is missing here.

pages/app-layout/utils/external-global-left-panel-widget.tsx
```diff
@@ -37,17 +61,52 @@ const AIDrawer = () => {
...
+        <div key={index}>Tela chat content</div>
       ))}
     </Box>
   );
+  const artifactContent = (
+    <Box padding="m">
+      <Box variant="h2" padding={{ bottom: 'm' }}>
+        Artifact
+      </Box>
+      <Button
```
pan-kot, line 70: nit: There is no margins between these buttons.

src/i18n/messages/all.en.json
gethinwebster: good catch, there seem to be strange things going on with our i18n process, removed this from this PR for now

pages/app-layout/utils/external-global-left-panel-widget.tsx
```diff
@@ -37,17 +61,52 @@ const AIDrawer = () => {
...
+        }}
+      >
+        Close artifact panel
+      </Button>
+      <Button onClick={() => setArtifactLoaded(true)}>Load artifact details</Button>
+      {artifactLoaded && new Array(100).fill(null).map((_, index) => <div key={index}>Tela content</div>)}
+    </Box>
+  );
+  return (
+    <div ref={ref} style={{ width: '100%', overflow: 'hidden' }}>
```
pan-kot: nit: I don't mind inline styles, but since we already use class names e.g. for the ai-artifact-panel, it would be nice to use it here, too - for consistency

src/panel-layout/interfaces.ts
gethinwebster: Updated the wording within `panelSize`/`defaultPanelSize` to make this clearer

src/panel-layout/index.tsx
gethinwebster: Good point, updated. However left `hasPanelSize` as it is, I think it's clearer that way in case we add other controllable properties in the future.

src/panel-layout/internal.tsx
gethinwebster: we don't use that toolkit version anywhere within the components repo so far, I'd rather keep consistency with other components especially as the signatures are slightly different

pages/panel-layout/nested.page.tsx
pan-kot: oh, and it has a slot for settings

pages/panel-layout/app-layout-panel.page.tsx
```diff
@@ -0,0 +1,203 @@
...
+  longMainContent,
+  minContentSize,
+  minPanelSize,
+  maxPanelSize,
+  display,
+  panelPosition,
+}: PanelLayoutContentProps) => {
+  const [size, setSize] = useState(Math.max(200, minPanelSize));
+
+  const [_actualMaxPanelSize, ref] = useContainerQuery(
```
pan-kot: The onPanelResize.detail does include totalSize, but it looks like we cannot use it as the callback only fires on user-issued resize. Would it make senes to maybe add another callback that fires when the panel size changes?

We already use the container query inside the component, and if we expose the callback - the consumers won't have to use another container query outside.

src/panel-layout/internal.tsx
gethinwebster: it was on the parent element (further down the file), but agree it makes more sense to be here for consistency, updated

src/app-layout/utils/use-pointer-events.ts
```diff
@@ -33,8 +33,7 @@ export const usePointerEvents = ({ position, panelRef, handleRef, onResize }: Si
         // The handle offset aligns the cursor with the middle of the resize handle.
         const handleOffset = getLogicalBoundingClientRect(handleRef.current).inlineSize / 2;
         const panelBoundingClientRect = getLogicalBoundingClientRect(panelRef.current);
-        const width =
-          panelBoundingClientRect.insetInlineEnd + mouseClientX + handleOffset - panelBoundingClientRect.inlineSize;
+        const width = mouseClientX + handleOffset - panelBoundingClientRect.insetInlineStart;
```
pan-kot, line 36: why this change?

src/panel-layout/internal.tsx
```diff
@@ -0,0 +1,153 @@
...
+        setPanelSize(size);
+        fireNonCancelableEvent(onPanelResize, { totalSize: containerWidth, panelSize: size });
+      },
+    });
+
+    const mergedRef = useMergeRefs(rootRef, __internalRootRef, ref);
+
+    const wrappedPanelContent = (
+      <div
+        className={clsx(styles['panel-content'], display !== 'main-only' && testStyles.panel)}
```
pan-kot, line 104: Why do we choose to hide but render the panels depending on the display prop instead of skip them from rendering? Should we then remove the ARIA props and tab index, too?

src/app-layout/utils/use-pointer-events.ts
gethinwebster: There was a bug in the old implementation depending how exactly the elements were positioned. This handles more different scenarios (plus is easier to understand IMO)

pages/panel-layout/nested.page.tsx
```diff
@@ -0,0 +1,146 @@
...
+        <Button>Outer Action</Button>
+        {Array.from({ length: 20 }, (_, i) => (
+          <div key={i}>Simple content line {i + 1}</div>
+        ))}
+      </SpaceBetween>
+    </Container>
+  );
+
+  return (
+    <div style={{ height: '100%', overflow: 'hidden' }}>
```
pan-kot: I noticed that we add these styles to all test pages. Would it make sense to support that on the component itself with props? E.g. some `stretchWidth` and `stretchHeight` flags?

pages/panel-layout/app-layout-panel.page.tsx
```diff
@@ -0,0 +1,212 @@
...
+      ) : (
+        <PanelLayout
+          panelSize={actualSize}
+          minPanelSize={minPanelSize}
+          maxPanelSize={actualMaxPanelSize}
+          resizable={true}
+          onPanelResize={({ detail }) => setSize(detail.panelSize)}
+          display={display}
+          panelPosition={panelPosition}
+          mainFocusable={longMainContent && !buttons ? { ariaLabel: 'Main content' } : undefined}
```
pan-kot: nit: logically it probably makes more sense to only depend on `!buttons`, as unless the content is measured or expected to be very small - there is a chance that it will not fit the screen when the viewport is small.

src/panel-layout/internal.tsx
gethinwebster: We hide rather than skipping so that contents can remain in-DOM for use-cases where that is useful. If teams _don't_ want that behavior, they can conditionally render on their side.

Regarding the other props, these seem to be ignored by browsers and screen readers as the element is `display: none`, and I don't see any accessibility scanner warnings about this either, so would rather reduce the amount of conditional logic unless we strictly need it.

pages/panel-layout/nested.page.tsx
```diff
@@ -0,0 +1,146 @@
...
+        <Container header={<Header variant="h2">Second Level Main Content</Header>}>
+          <p>This is the main content area of the second panel layout.</p>
+          <SpaceBetween size="xs" direction="vertical">
+            {Array.from({ length: 15 }, (_, i) => (
+              <div key={i}>Content line {i + 1}</div>
+            ))}
+          </SpaceBetween>
+        </Container>
+      }
+      mainFocusable={{ ariaLabel: 'Level 2 main' }}
```
pan-kot: There are two issues with the focus outline: in the app layout page, the focus outline is partially hidden behind the scrollbars. In the nested page, the focus outline is barely visible as of being overlayed by container styles.

Image

pan-kot: I also wonder, if it is possible to implement the panel in such a way that the header is sticky (or effectively sticky), so that only the content part is scrollable. It not or if hard - it might make sense to offer separate slots for header and content, potentially.

pages/panel-layout/nested.page.tsx
gethinwebster: It's not actually necessary on this dev page, will remove it. Whether (and what exactly) this styling is necessary is very dependent on the specific use-case, so I don't see a strong case for adding this as I think it would end up adding a fair amount of complexity to the API to handle multiple different cases.

pan-kot: Ok, it's fair. Let's then remove it from test pages where it is not needed

pages/panel-layout/app-layout-panel.page.tsx
gethinwebster: Nice idea, it also gets rid of a lot of places where teams have to add styled containers (because they no longer need a parent element to attach a ref to). Adding as `onLayoutChange`

pages/panel-layout/nested.page.tsx
gethinwebster: Pushed an update that seems to work better to me: it's no longer obscured by the scrollbars. I still don't have a solution for the scenario on the nested page (other than adding padding), but I also think in that scenario (as mentioned in a thread above) the ideal solution would be scrolling _inside_ the container which currently would need to be implemented inside the panel contents (i.e. by the consumer), so I think this is a reasonable solution to the most common use-cases

pan-kot: I agree, when a container-like content is added - having the scroll outside does not look right. Is it straightforward to implement an internal scroll, though? There might be challenges are the panels already have overflow styles, added unconditionally.

pages/app-layout/utils/external-global-left-panel-widget.tsx
gethinwebster: styles have been removed in latest version

src/panel-layout/interfaces.ts
```diff
@@ -0,0 +1,121 @@
...
+  onLayoutChange?: NonCancelableEventHandler<PanelLayoutProps.PanelResizeDetail>;
+}
+
+export namespace PanelLayoutProps {
+  export type PanelPosition = 'side-start' | 'side-end';
+  export type Display = 'all' | 'panel-only' | 'main-only';
+
+  export interface PanelResizeDetail {
+    totalSize: number;
+    panelSize: number;
```
pan-kot, line 102: Can main content size be obtained as `totalSize - panelSize`, or one must account for the resize handle when `resizable=true`?

src/panel-layout/internal.tsx
```diff
@@ -0,0 +1,158 @@
...
+      }
+    );
+
+    const actualMaxSize = Math.min(maxPanelSize ?? containerWidth, containerWidth);
+    const actualMinSize = minPanelSize ?? 0;
+    const actualPanelSize = Math.max(Math.min(panelSize, actualMaxSize), actualMinSize);
+
+    useEffect(() => {
+      fireNonCancelableEvent(onLayoutChange, { totalSize: containerWidth, panelSize: actualPanelSize });
+    }, [containerWidth, actualPanelSize, onLayoutChange]);
```
pan-kot: Should we wrap onLayoutChange with stable callback?

pan-kot: Without that, the onLayoutChange is called on every render, unless the callback is memoized on the consumer end.

pan-kot: Also, it might make sense to keep the prev `totalSize` and `panelSize` in a ref - and only call the onLayoutChange if either value changes. In that case stable callback util is not even needed.

src/panel-layout/internal.tsx
```diff
@@ -0,0 +1,158 @@
...
+      >
+        {panelPosition === 'side-end' && wrappedMainContent}
+        <div
+          className={clsx(styles.panel)}
+          ref={panelRef}
+          style={display === 'all' ? { inlineSize: `${actualPanelSize}px` } : undefined}
+        >
+          {panelPosition === 'side-start' && wrappedPanelContent}
+          {resizable && display === 'all' && (
+            <div className={styles.handle}>
```
pan-kot: nit: you can put the jsx for handle in a variable, same as wrappedMainContent/wrappedPanelContent. It should make the conditional logic here easier to follow

src/panel-layout/interfaces.ts
gethinwebster: `panelSize` includes the resize handle, so actual available panel content size is smaller if resizable

pan-kot: Should we compensate for that?

E.g. if I have some content which min size is precisely 400px - I should want to know that it can fit into the panel. Unless we do this - the consumer will have to account for the resize handle size, which would require to check for resizable and display props also to figure out if the resize handle size should be subtracted from the panel size.

pan-kot: As we know when the resize handle is shown and its size, we can do the compensation, and then expose `totalSize`, `panelSize`, and `mainSize` in the event handler details (in which case the sum of panel and main size will only equal total size if there is no resize handle).

gethinwebster: I think that makes it actually harder for teams. For example, the main use-case for that callback is adjusting the `maxPanelSize` to have an effective `minContentSize`. But for that to work properly the values have to add up consistently, and if we're compensating for the handle I can't see a way that we can do that that will be easy to understand.

src/panel-layout/internal.tsx
gethinwebster: good suggestion, updated

src/panel-layout/internal.tsx
gethinwebster: updated

src/panel-layout/interfaces.ts
pan-kot: Can we expose an additional `panelContentSize` (that is panel size minus handle size)?

pan-kot: That can actually be an incremental change, unlike the one I suggested earlier - so we can do it later as well.

PR #4115, opened 2025-12-17
chore: Shift focus to active ref by clicking / pressing on the recovery CTA

georgylobko:
### Description

Return focus on the prev active element on clicking to Retry CTA
AWSUI-61532

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/multiselect/__tests__/multiselect.test.tsx
+++ b/src/multiselect/__tests__/multiselect.test.tsx
@@ -7,6 +7,7 @@ import { warnOnce } from '@cloudscape-design/component-toolkit/internal';
 import { KeyCode } from '@cloudscape-design/test-utils-core/utils';
 
 import '../../__a11y__/to-validate-a11y';
+import TestI18nProvider from '../../../lib/components/i18n/testing';
 import Multiselect, { MultiselectProps } from '../../../lib/components/multiselect';
 import createWrapper from '../../../lib/components/test-utils/dom';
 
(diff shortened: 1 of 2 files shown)


src/select/internal.tsx
```diff
@@ -197,7 +198,10 @@ const InternalSelect = React.forwardRef(
       noMatch,
       filteringResultsText: filteredText,
       errorIconAriaLabel,
-      onRecoveryClick: handleRecoveryClick,
+      onRecoveryClick: () => {
+        handleRecoveryClick();
+        focusActiveRef();
```
pan-kot, line 203: I believe we should do the same for multiselect

src/select/__tests__/select.test.tsx
```diff
@@ -210,6 +210,7 @@ describe.each([false, true])('expandToViewport=%s', expandToViewport => {
       wrapper.openDropdown();
       onLoadItems.mockClear();
       wrapper.findErrorRecoveryButton({ expandToViewport })!.click();
+      expect(wrapper.findFilteringInput()!.findNativeInput()!.getElement()).toHaveFocus();
```
pan-kot, line 213: Can we also validate the highlighted option?

georgylobko: Not necessarily - the way it works is once the retry button is clicked, the focus is shifted to the search area and we don't expect any option to be highlighted at this point. However, if the user presses arrow down key, the closest option will be highlighted. The main point of this fix is to keep focus within the select component

src/select/internal.tsx
georgylobko: Fair enough

PR #4123, opened 2025-12-19
fix: Fix AppLayoutToolbar breadcrumbs SSR glitch (take three)

TrevorBurnham:
### Description

Previously: https://github.com/cloudscape-design/components/pull/3856, https://github.com/cloudscape-design/components/pull/4065

Embarrassingly, that second PR was missing one of the two changes needed (as seen [on the original branch](https://github.com/TrevorBurnham/cloudscape-components/commit/cbc5ea2ce8a6def5709e474ae7093c7cee0abfbf)), so it was insufficient to fix the glitch. This PR brings over that other change and adds test coverage for it.

Fixes https://github.com/cloudscape-design/components/issues/3848 cc @pan-kot 

### How has this been tested?

I've confirmed this fix locally by cherry-picking [this SSR dev server commit](https://github.com/TrevorBurnham/cloudscape-components/commit/e8b635e0954a1c289ee91cd59a6299690f709b40) and running it.

<details>
   <summary>Review checklist</summary>

_The following items are to be evaluated by the author(s) and the reviewer(s)._

#### Correctness

- _Changes include appropriate documentation updates._
- _Changes are backward-compatible if not indicated, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#public-apis)._
- _Changes do not include unsupported browser features, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#browsers-support)._
- _Changes were manually tested for accessibility, see [accessibility guidelines](https://cloudscape.design/foundation/core-principles/accessibility/)._

#### Security

- _If the code handles URLs: all URLs are validated through [the `checkSafeUrl` function](https://github.com/cloudscape-design/components/blob/main/src/internal/utils/check-safe-url.ts)._

#### Testing

- _Changes are covered with new/existing unit tests?_
- _Changes are covered with new/existing integration tests?_
</details>

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #4065, opened 2025-11-23
fix: Fix AppLayoutToolbar breadcrumbs SSR glitch (take two)

TrevorBurnham:
### Description

I previously submitted a fix for a visual glitch when rendering the `AppLayoutToolbar` breadcrumbs with SSR: #3856 Unfortunately, although that fix was merged, I'm still seeing the glitch.

I investigated and found that while the original PR commit fixed the glitch, it was no longer effective after making updates in response to suggestions. That's my fault; I should've validated those updates. This PR restores the key fixes.

Fixes #3848

### How has this been tested?

I have an [updated branch](https://github.com/TrevorBurnham/cloudscape-components/tree/applayouttoolbar-ssr-updated) with an SSR server. Check out that branch and run `npm install && npm run dev:ssr`. Load http://localhost:3000 in a browser and you can see that the glitch is fixed (thanks to the change in this PR). Check out `HEAD~1` and you can see that the glitch occurs without this change.

<details>
   <summary>Review checklist</summary>

_The following items are to be evaluated by the author(s) and the reviewer(s)._

#### Correctness

- _Changes include appropriate documentation updates._
- _Changes are backward-compatible if not indicated, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#public-apis)._
- _Changes do not include unsupported browser features, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#browsers-support)._
- _Changes were manually tested for accessibility, see [accessibility guidelines](https://cloudscape.design/foundation/core-principles/accessibility/)._

#### Security

- _If the code handles URLs: all URLs are validated through [the `checkSafeUrl` function](https://github.com/cloudscape-design/components/blob/main/src/internal/utils/check-safe-url.ts)._

#### Testing

- _Changes are covered with new/existing unit tests?_
- _Changes are covered with new/existing integration tests?_
</details>

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #3856, opened 2025-09-13
fix: Fix AppLayoutToolbar breadcrumbs SSR glitch

TrevorBurnham:
### Description

When rendering the `AppLayoutToolbar` component under the visual refresh theme with SSR, there's a visual glitch where the breadcrumbs are positioned incorrectly before hydration:

https://github.com/user-attachments/assets/dfca4390-0bd4-4d02-93d0-d4a424c10ea3

This PR addresses the issue by making the SSR render more consistent with the client-side render. This is a relatively small change that should have no effect on applications that use client-side rendering only.

There are several other issues with `AppLayoutToolbar` in SSR (see #3848); this PR should  provide a starting point for fixing them.

### How has this been tested?

I have [a branch](https://github.com/TrevorBurnham/cloudscape-components/tree/applayouttoolbar-ssr-demo) where I've set up a local SSR dev server. You can pull that branch down and run `npm run start:watch` in one shell and `npm run start:ssr` in another.

<details>
   <summary>Review checklist</summary>

_The following items are to be evaluated by the author(s) and the reviewer(s)._

#### Correctness

- _Changes include appropriate documentation updates._
- _Changes are backward-compatible if not indicated, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#public-apis)._
- _Changes do not include unsupported browser features, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#browsers-support)._
- _Changes were manually tested for accessibility, see [accessibility guidelines](https://cloudscape.design/foundation/core-principles/accessibility/)._

#### Security

- _If the code handles URLs: all URLs are validated through [the `checkSafeUrl` function](https://github.com/cloudscape-design/components/blob/main/src/internal/utils/check-safe-url.ts)._

#### Testing

- _Changes are covered with new/existing unit tests?_
- _Changes are covered with new/existing integration tests?_
</details>

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #4127, opened 2025-12-20
chore: New demo server featuring SSR

TrevorBurnham:
### Description

I've been trying to build a Cloudscape app that uses SSR. In doing so, I've found a number of inconsistencies between the visual state when components are rendered server-side vs. when they're rendered client-side. However, these bugs are difficult to replicate and fix, which has made it difficult to contribute patches (see e.g. https://github.com/cloudscape-design/components/pull/3856, https://github.com/cloudscape-design/components/pull/4065, and #4123). 

This PR replaces the existing demo server (`npm run start:dev`) with a new one that features SSR. This should make it much easier to spot SSR bugs, and to confirm fixes.

The new server should be backward-compatible with the old one, though a couple of (hopefully minor) issues are worth pointing out:

#### Time to hydration

It takes ~2s for client-side rendering to kick in, mainly because of the large number of JS and CSS files served on the page. That means that the page is non-interactive during that time. It also means that any screenshot taken before that time would reflect an SSR bug, not what a component looks like when rendered client-side.

For Cloudscape's visual diff tests, we'll need to ensure that those visual diff tests wait for hydration before taking a screenshot. If they already wait for all resources on the page to load, that should be sufficient.

#### URL redirection

The old demo server used hash-based routing, e.g.

```
http://localhost:8080/#/app-layout-toolbar/with-content-layout
```

Hashes can't be read by the server, so supporting SSR requires path-based routing, e.g.

```
http://localhost:8080/app-layout-toolbar/with-content-layout
```

There's a client-side redirect for hashes, so hash-based routes will still take you to the correct page, but the initial SSR-rendered page will be the home page.

### How has this been tested?

I've run the server and poked around quite a bit, but I'd welcome more suggestions for testing.

<details>

#### Correctness

#### Security

#### Testing

</details>

---


SpyZzey:
We are working on the migration to Vite for the components package, which requires some more work in other packages too. I am closing this PR to add the SSR demo server on top of that migration instead once it is ready, as we'd need to delete/refactor most of these changes anyways.

PR #4131, opened 2025-12-23
fix: restore focus to header when exiting tutorial

NathanZlion:
### Description

Fix Accessibility issue for focus on the tutorial panel component. When a user exits/dismisses a tutorial, focus is automatically restored to the tutorial panel header to maintain proper keyboard navigation flow.

Related links, issue AWSUI-61570

### How has this been tested?
- Tested Locally
- Added integration test to check behavior and run it locally

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/dropdown/list-with-sticky-item.page.tsx
+++ b/pages/dropdown/list-with-sticky-item.page.tsx
@@ -90,7 +90,7 @@ export default function MultiselectPage() {
             >
               <ListComponent
                 menuProps={{ statusType: 'finished', ref, open }}
-                getOptionProps={(option, index) => ({ option: { ...option }, key: index, open })}
+                getOptionProps={(option, index) => ({ option: { ...option }, key: index, index: index, open })}
                 filteredOptions={options}
                 filteringValue={''}
                 firstOptionSticky={true}
--- a/pages/multiselect/custom-render-option.page.tsx
+++ b/pages/multiselect/custom-render-option.page.tsx
@@ -1,0 +1,67 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import * as React from 'react';
+
+import { Multiselect, MultiselectProps } from '~components';
+import { SelectProps } from '~components/select';
+
+import { SimplePage } from '../app/templates';
+import { i18nStrings } from './constants';
+const lotsOfOptions: SelectProps.Options = [...Array(50)].map((_, index) => ({
+  value: `Option ${index}`,
+  label: `Option ${index}`,
+}));
+const options: SelectProps.Options = [
+  { value: 'first', label: 'Simple' },
+  { value: 'second', label: 'With small icon', iconName: 'folder' },
+  {
+    value: 'third',
+    label: 'With big icon icon',
+    description: 'Very big option',
+    iconName: 'heart',
+    disabled: false,
+    disabledReason: 'disabled reason',
+    tags: ['Cool', 'Intelligent', 'Cat'],
+  },
+  {
+    label: 'Option group',
+    options: [{ value: 'forth', label: 'Nested option' }],
+    disabledReason: 'disabled reason',
+  },
+  ...lotsOfOptions,
+  { label: 'Last option', disabled: false, disabledReason: 'disabled reason' },
+];
+
+export default function SelectPage() {
+  const [selectedOptions, setSelectedOptions] = React.useState<MultiselectProps.Options>([]);
+  const renderOptionItem: MultiselectProps.MultiselectOptionItemRenderer = ({ item }) => {
+    if (item.type === 'select-all') {
+      return <div>Select all? {item.selected ? 'Yes' : 'No'}</div>;
+    } else if (item.type === 'group') {
+      return <div>Group: {item.option.label}</div>;
+    } else if (item.type === 'item') {
+      return <div>Item: {item.option.label}</div>;
+    }
+
+    return null;
+  };
+
+  return (
+    <SimplePage title="Multiselect with custom item renderer" screenshotArea={{}}>
+      <div style={{ inlineSize: '400px' }}>
+        <Multiselect
+          enableSelectAll={true}
+          i18nStrings={{ ...i18nStrings, selectAllText: 'Select all' }}
+          filteringType={'auto'}
+          renderOption={renderOptionItem}
+          placeholder="Choose option"
+          selectedOptions={selectedOptions}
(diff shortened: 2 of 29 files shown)


src/tutorial-panel/index.tsx
```diff
@@ -28,6 +28,16 @@ export default function TutorialPanel({
 
   const baseProps = getBaseProps(restProps);
   const context = useContext(hotspotContext);
+  const headerRef = useRef<HTMLHeadingElement>(null);
+  const previousTutorialRef = useRef(context.currentTutorial);
+
+  // Restore focus to the tutorial panel header when exiting a tutorial
```
gethinwebster: I think it would be better to use the `onExitTutorial` event more directly rather than observing for changes here, it should make the code a bit clearer to read and avoid needing `useEffect`

src/tutorial-panel/components/tutorial-list/index.tsx
```diff
@@ -45,9 +47,11 @@ export default function TutorialList({
     <>
       <InternalSpaceBetween size="s">
         <InternalSpaceBetween size="m">
-          <InternalBox variant="h2" fontSize={isRefresh ? 'heading-m' : 'heading-l'} padding={{ bottom: 'n' }}>
-            {i18nStrings.tutorialListTitle}
-          </InternalBox>
+          <div ref={headerRef} tabIndex={-1}>
```
avinashbot: We shouldn't move focus to an unlabelled generic element like this. You could add `role="region"` and `aria-labelledby` to this div, wrap the title in a `<span id="...">` and provide the id to the `aria-labelledby` of the div. You would also need to add custom focus ring styling (see uses of `focus-visible.when-visible` in other files).

Can you make this element always be present across all "modes" (like when a tutorial is active or completed or not started yet)? This way you can avoid that `requestAnimationFrame` hack.

src/tutorial-panel/__integ__/tutorial-panel-focus.test.ts
```diff
@@ -0,0 +1,65 @@
...
+
+const setupTest = (testFn: (page: TutorialPanelFocusPage) => Promise<void>) =>
+  useBrowser(async browser => {
+    const page = new TutorialPanelFocusPage(browser);
+    await browser.url('#/light/onboarding/with-app-layout');
+    await page.waitForTutorialListToLoad();
+    await testFn(page);
+  });
+
+describe('TutorialPanel Focus Management', () => {
```
avinashbot: Since the focus moves here are programmatic, I think you can get away with making this a simple unit test! (You just have to provide a mock implementation for `requestAnimationFrame`)

src/tutorial-panel/index.tsx
```diff
@@ -36,7 +37,10 @@ export default function TutorialPanel({
           <TutorialDetailView
             i18nStrings={i18nStrings}
             tutorial={context.currentTutorial}
-            onExitTutorial={context.onExitTutorial}
+            onExitTutorial={e => {
+              context.onExitTutorial(e);
+              requestAnimationFrame(() => headerRef.current?.focus({ preventScroll: true }));
```
avinashbot: Why `requestAnimationFrame`? Is it because the element to focus won't be available until the next render? This might not always work on more complex applications.

NathanZlion: yeah, yhe requestAnimationFrame was because the element to focus on wasn't available until the rerender happened but I see how that might be unreliable.

src/tutorial-panel/components/tutorial-list/index.tsx
NathanZlion: I see, it was because of this piece of code in `src/tutorial-panel/index.tsx` which either shows the list view (which has the title we want to focus on when tutorial's dismissed) or the detail view. 

```
  return (
    <>
      <div {...baseProps} className={clsx(baseProps.className, styles['tutorial-panel'])} ref={__internalRootRef}>
        {context.currentTutorial ? (
          <TutorialDetailView
          ...
          />
        ) : (
          <TutorialList
            ...
          />
        )}
      </div>
    </>
  );
```

so when we dismiss the title will be available only after the TutorialList gets rendered again.

src/tutorial-panel/index.tsx
```diff
@@ -29,14 +32,34 @@ export default function TutorialPanel({
...
+  const headerCallbackRef = useCallback((node: HTMLDivElement | null) => {
+    if (node && shouldFocusRef.current) {
+      node.focus({ preventScroll: true });
+      shouldFocusRef.current = false;
+    }
+  }, []);
+
+  const handleExitTutorial = useCallback(
+    (e: NonCancelableCustomEvent<TutorialPanelProps.TutorialDetail>) => {
+      shouldFocusRef.current = true;
```
gethinwebster, line 52: would it be possible to directly call `ref.current.focus()` on a ref attached to the header in this callback? I think that would be a cleaner solution if possible

NathanZlion: yeah... that's the tricky problem, at the time when the handleExitTutorial is called from the TutorialDetialView component the TutorialList is not rendered yet, so the header which is inside the tutoriallist is not available to focus on.

NathanZlion: So we need to do some kind of coordinating in the tutorialpanel

gethinwebster: got it, makes sense. I think a comment in the code would probably be useful for that then to explain why its done in a slightly unusual way

src/tutorial-panel/__tests__/tutorial-panel.test.tsx
```diff
@@ -293,5 +294,55 @@ describe('URL sanitization', () => {
...
+
+    test('header has correct accessibility attributes', () => {
+      const { container } = renderTutorialPanelWithContext();
+      const wrapper = createWrapper(container).findTutorialPanel()!;
+      const headerRegion = wrapper.find('[role="region"]')!.getElement();
+      expect(headerRegion).toHaveAttribute('tabIndex', '-1');
+      expect(headerRegion).toHaveAttribute('role', 'region');
+      expect(headerRegion).toHaveAttribute('aria-labelledby');
+
+      const ariaLabelledBy = headerRegion.getAttribute('aria-labelledby');
```
gethinwebster: [minor] it should be possible to test using `toHaveAccessibleName` rather than manually testing the aria-labelledby like this

NathanZlion: Done,

NathanZlion:
#4167

PR #4136, opened 2025-12-30
feat: add dev page utils from build-tools

amanabiy:
### Description
This PR allows us to use dev-page utils in components repository.

How does it work?
- In package.json we have `@cloudscape-design/build-tools` which gets the package from GitHub - this works for running locally.
- To allow it running in the pipeline, I added a copy during build time to copy it to `/lib/dev-pages/internal/build-tools`

Tested by:
- Running through the pipeline and testing locally github components

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -28,15 +28,7 @@ const {
 
 const quickBuild = series(
   clean,
-  parallel(
-    packageJSON,
-    generateI18nMessages,
-    generateEnvironment,
-    generateIcons,
-    generateIndexFile,
-    copyFiles,
-    copyBuildTools
-  ),
+  parallel(packageJSON, generateI18nMessages, generateEnvironment, generateIcons, generateIndexFile, copyFiles),
   parallel(generateCustomCssPropertiesMap, styles, typescript, testUtils),
   bundleVendorFiles
 );
@@ -44,7 +36,7 @@ const quickBuild = series(
 exports.clean = clean;
 exports['quick-build'] = quickBuild;
 exports.i18n = generateI18nMessages;
-exports.build = series(quickBuild, parallel(buildPages, themeableSource, docs, sizeLimit));
+exports.build = series(quickBuild, parallel(buildPages, themeableSource, docs, sizeLimit, copyBuildTools));
 exports.test = series(unit, integ, a11y);
 exports['test:unit'] = unit;
 exports['test:integ'] = integ;


scripts/setup-build-tools.js
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env node
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// Can be used in postinstall script like so:
+// "postinstall": "node ./scripts/install-peer-dependency.js collection-hooks:property-filter-token-groups"
```
pan-kot: this description is a leftover, I believe we don't need any for this script as all

scripts/setup-build-tools.js
```diff
@@ -0,0 +1,33 @@
...
+import process from 'node:process';
+import path from 'path';
+
+const branch = 'add-test-pages-util-permutation-view';
+const packageName = 'build-tools';
+const targetRepository = `https://github.com/cloudscape-design/${packageName}.git`;
+const copyBuildToolsPath = path.join(process.cwd(), 'shared', 'build-tools');
+execCommand(`mkdir -p ${copyBuildToolsPath}`);
+execCommand(`rm -rf ${copyBuildToolsPath}`);
+execCommand(`git clone --branch ${branch} --single-branch ${targetRepository} ${copyBuildToolsPath}`);
```
pan-kot: For such a simple command - can we inline that in the package.json?

E.g. "rimraf shared/build-tools && mkdir -p shared/build-tools && git clone --branch add-test-pages-util-permutation-view --single-branch https://github.com/cloudscape-design/build-tools.git shared/build-tools"

pages/alert/permutations.page.tsx
```diff
@@ -111,7 +111,10 @@ export default function AlertScenario() {
     <article>
       <h1>Alert permutations</h1>
       <ScreenshotArea>
-        <PermutationsView permutations={permutations} render={permutation => <Alert {...permutation} />} />
+        <PermutationsView
+          permutations={permutations}
+          render={(permutation: AlertProps) => <Alert {...permutation} />}
```
pan-kot: why the param is explicitly annotated here - it is necessary?

package.json
```diff
@@ -53,7 +54,8 @@
     "@babel/core": "^7.23.7",
     "@babel/plugin-syntax-typescript": "^7.23.3",
     "@cloudscape-design/browser-test-tools": "^3.0.0",
-    "@cloudscape-design/build-tools": "github:cloudscape-design/build-tools#main",
+    "@cloudscape-design/build-tools-repo": "github:cloudscape-design/build-tools#add-test-pages-util-permutation-view",
```
pan-kot: what is @cloudscape-design/build-tools-repo for?

scripts/setup-build-tools.js
amanabiy: I have updated the implementation like the following: instead of using a script we just add it to the package.json devDependencies, and then during build time we copy it to the `/lib/dev-pages/internal/build-tools` that way the package.json will be consistent with the other repositories as well.

pan-kot (approved):

amanabiy:
This PR is going to be closed as we are going to start using the dev-pages utils from the build-tools after migration to vite for the dev pages. This is because right now using from build-tools would require us to make changes to the webpack configuration and we don't want to do that to keep things simple.

PR #4141, opened 2025-12-31
fix: Cache Intl.DateTimeFormat instances for better performance

TrevorBurnham:
### Description

Currently, every call to `formatDateLocalized` (used in `DateInput`) allocates new `Intl.DateTimeFormat` instances, which have significant initialization overhead. This PR adds caching for those formatters so that they're only created the first time `formatDateLocalized` is called.

### How has this been tested?

I've added thorough unit test coverage.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/utils/date-time/intl-date-time-format-cache.ts
+++ b/src/internal/utils/date-time/intl-date-time-format-cache.ts
@@ -1,0 +1,50 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * Cache for Intl.DateTimeFormat instances.
+ *
+ * Creating Intl.DateTimeFormat objects is expensive because the browser must
+ * parse locale data and resolve options. This cache stores formatter instances
+ * keyed by locale and options, allowing reuse across multiple format calls.
+ *
+ * The cache uses a simple Map with a string key derived from the locale and
+ * serialized options. Since the number of unique locale/option combinations
+ * in a typical application is small and bounded, we don't implement cache
+ * eviction.
+ */
+
+const formatterCache = new Map<string, Intl.DateTimeFormat>();
+
+/**
+ * Returns a cached Intl.DateTimeFormat instance for the given locale and options.
+ * If no cached instance exists, creates one and stores it in the cache.
+ */
+export function getDateTimeFormat(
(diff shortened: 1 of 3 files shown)


src/internal/utils/date-time/intl-date-time-format-cache.ts
```diff
@@ -0,0 +1,50 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * Cache for Intl.DateTimeFormat instances.
+ *
+ * Creating Intl.DateTimeFormat objects is expensive because the browser must
```
amanabiy: How much performance gain are we expecting from this caching mechanism?

TrevorBurnham: It's tiny on a per-call basis: something on the order of 10-100 microseconds. But it's cumulative per localized date. On a page with a table full of formatted dates, the savings could add up to 10s of milliseconds.

src/internal/utils/date-time/intl-date-time-format-cache.ts
```diff
@@ -0,0 +1,50 @@
...
+ * parse locale data and resolve options. This cache stores formatter instances
+ * keyed by locale and options, allowing reuse across multiple format calls.
+ *
+ * The cache uses a simple Map with a string key derived from the locale and
+ * serialized options. Since the number of unique locale/option combinations
+ * in a typical application is small and bounded, we don't implement cache
+ * eviction.
+ */
```
amanabiy: The comment here is verbose, The file and function names are descriptive. Let's try to simplify the comment, removing redundancy. we could keep why we don't implement cache eviction (and keep it close to the `formatterCache`.

TrevorBurnham: Good call. I've trimmed the docblocks down to just the *why*:
- Top comment reduced to a two-line rationale for the cache existing.
- Moved the no-eviction note directly onto the `formatterCache` declaration, as you suggested.
- Dropped the `getDateTimeFormat` docblock since it only restated the signature.
- Kept the one non-obvious line on `createCacheKey` (sorting keys for order-independent hits).

amanabiy (approved):

PR #4142, opened 2025-12-31
fix: Use Map for O(1) column lookups in table width calculations

TrevorBurnham:
### Description

Currently the table width calculations use repeated `find` calls across columns in `getColumnStyles`, which is called for each column, so worst-case performance is `O(n^2)`. This PR switches to creating a `Map` for that lookup, making performance consistently `O(n)`.

### How has this been tested?

This functionality is well-covered by existing tests.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/table/use-column-widths.tsx
+++ b/src/table/use-column-widths.tsx
@@ -1,6 +1,6 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
-import React, { createContext, useContext, useEffect, useRef, useState } from 'react';
+import React, { createContext, useContext, useEffect, useMemo, useRef, useState } from 'react';
 
 import { useResizeObserver, useStableCallback } from '@cloudscape-design/component-toolkit/internal';
 import { getLogicalBoundingClientRect } from '@cloudscape-design/component-toolkit/internal';
@@ -35,18 +35,17 @@ function readWidths(
 }
 
 function updateWidths(
-  visibleColumns: readonly ColumnWidthDefinition[],
+  column: ColumnWidthDefinition | undefined,
   oldWidths: Map<PropertyKey, number>,
   newWidth: number,
   columnId: PropertyKey
 ) {
-  const column = visibleColumns.find(column => column.id === columnId);
   let minWidth = DEFAULT_COLUMN_WIDTH;
   if (typeof column?.width === 'number' && column.width < DEFAULT_COLUMN_WIDTH) {
-    minWidth = column?.width;
+    minWidth = column.width;
   }
   if (typeof column?.minWidth === 'number') {
-    minWidth = column?.minWidth;
+    minWidth = column.minWidth;
   }
   newWidth = Math.max(newWidth, minWidth);
   if (oldWidths.get(columnId) === newWidth) {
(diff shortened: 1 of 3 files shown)


src/table/use-column-widths.tsx
```diff
@@ -83,6 +83,9 @@ export function ColumnWidthsProvider({ visibleColumns, resizableColumns, contain
   const containerWidthRef = useRef(0);
   const [columnWidths, setColumnWidths] = useState<null | Map<PropertyKey, number>>(null);
 
+  // Pre-build a Map for column lookups
+  const columnById = useMemo(() => new Map(visibleColumns.map(column => [column.id, column])), [visibleColumns]);
```
pan-kot, line 85: The visibleColumns is created on every render, so there is no reason to memoize this computation - it will still recompute every time.

src/table/use-column-widths.tsx
```diff
@@ -83,6 +83,9 @@ export function ColumnWidthsProvider({ visibleColumns, resizableColumns, contain
   const containerWidthRef = useRef(0);
   const [columnWidths, setColumnWidths] = useState<null | Map<PropertyKey, number>>(null);
 
+  // Pre-build a Map for column lookups
```
pan-kot: nit: this comment seems redundant

src/table/use-column-widths.tsx
pan-kot: btw, it is not O(1) as specified in the PR description but O(n) - which is the cost of dictionary creation

src/table/use-column-widths.tsx
```diff
@@ -35,12 +35,12 @@ function readWidths(
 }
 
 function updateWidths(
-  visibleColumns: readonly ColumnWidthDefinition[],
+  columnById: Map<PropertyKey, ColumnWidthDefinition>,
```
pan-kot: Should it take the `column` instead?

src/table/use-column-widths.tsx
TrevorBurnham: Good catch! I've updated this PR to memoize `visibleColumns` as well.

src/table/use-column-widths.tsx
TrevorBurnham: Removed, thanks.

src/table/use-column-widths.tsx
TrevorBurnham: Good catch. I've updated `updateWidths` to just take the individual column instead of the whole map.

src/table/use-column-widths.tsx
pan-kot: But why all this memoization is needed? It adds a lot of clumsiness to the code but the performance impact of transforming columns is negligible even if there are 100 columns, which is rare.

TrevorBurnham: So, the memoization is actually a much bigger win than the `Map` change: Every time [this `useEffect`](https://github.com/cloudscape-design/components/blob/6e275009c30de280d2a272451f70e2a3790e2937/src/table/use-column-widths.tsx#L156-L180) gets triggered, it calls `updateColumnWidths`, which is potentially very expensive because it [makes DOM measurements with `getBoundingClientRect()`](https://github.com/cloudscape-design/components/blob/6e275009c30de280d2a272451f70e2a3790e2937/src/table/use-column-widths.tsx#L108). Doing that on table renders that don't affect the column widths is a recipe for making interactions feel slow. I'm sure you've experienced times where there's noticeable lag when typing in a Cloudscape table's search box; the `updateColumnWidths` call is likely a big contributor to that lag.

TrevorBurnham: I've updated this PR with a "does not measure column widths when re-rendering with unchanged columns" test, which I think is useful for validating the performance improvement here and preventing future regressions.

pan-kot:

PR #4143, opened 2025-12-31
fix: Memoize context provider values

TrevorBurnham:
Note: This branch includes #4142, which should be merged first. Submitting the changes separately would cause merge conflicts.

### Description

When the `value` object in a context provider changes, every component that consumes that context is updated. That means that something as simple as

```
<MyContext.Provider value={{x}}>
```

can create unnecessary updates: If the provider re-renders but `x` is the same, all context consumers will nonetheless update because a new `{x}` object is produced.

This PR adds memoization to all context providers to reduce unnecessary updates. This could have significant performance benefits in tables, where `WidthsContext` has been causing `Thead` to update even when no column widths changed.

### How has this been tested?

Existing test coverage should be sufficient for this change.

<details>

#### Correctness

#### Security

#### Testing

</details>

---


pan-kot:

PR #4144, opened 2025-12-31
fix: Cache measured column widths to avoid getBoundingClientRect during render

TrevorBurnham:
Note: This branch includes #4142 and #4143. Those PRs should be merged first.

### Description

I noticed that `getBoundingClientRect` was being called directly in `useColumnWidths`, which adds significant overhead to table renders. With this change, `getBoundingClientRect` is only called during `updateColumnWidths`, which is called outside of the React render flow. This should make table updates smoother.

### How has this been tested?

Since this is a render performance issue, it's difficult to test directly. I'd welcome suggestions.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/table/use-column-widths.tsx
+++ b/src/table/use-column-widths.tsx
@@ -91,6 +91,8 @@ export function ColumnWidthsProvider({ visibleColumns, resizableColumns, contain
 
   const cellsRef = useRef(new Map<PropertyKey, HTMLElement>());
   const stickyCellsRef = useRef(new Map<PropertyKey, HTMLElement>());
+  // Cache measured widths from real cells to avoid getBoundingClientRect() during render
+  const measuredWidthsRef = useRef(new Map<PropertyKey, number>());
   const getCell = (columnId: PropertyKey): null | HTMLElement => cellsRef.current.get(columnId) ?? null;
   const setCell = useCallback((sticky: boolean, columnId: PropertyKey, node: null | HTMLElement) => {
     const ref = sticky ? stickyCellsRef : cellsRef;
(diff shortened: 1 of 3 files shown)


src/table/use-column-widths.tsx
```diff
@@ -83,61 +83,84 @@ export function ColumnWidthsProvider({ visibleColumns, resizableColumns, contain
...
 
-    if (sticky) {
-      return {
-        width:
-          cellsRef.current.get(column.id)?.getBoundingClientRect().width ||
-          (columnWidths?.get(column.id) ?? column.width),
-      };
-    }
+      if (sticky) {
+        // Use cached measured width to avoid getBoundingClientRect() during render.
```
pan-kot, line 114: What is the cost of calling `getBoundingClientRect()` during render exactly?

TrevorBurnham: Calling `getBoundingClientRect()` forces the browser to recompute layout. Calling it multiple times in a row is fine, since (if nothing in the DOM has changed) the layout computations are cached. But calling it in a React render is unpredictable: React may or may not batch different component renders together, so it's possible for the DOM state to change in between different `getBoundingClientRect()` calls, causing [layout thrashing](https://webperf.tips/tip/layout-thrashing/). It's also considered generally bad practice to perform expensive operations that block React updates.

The key change here is that all of the `getBoundingClientRect()` calls for the table are batched together to minimize potential overhead, and performed outside of render so they don't block React.

pan-kot:

TrevorBurnham:
I think this is a significant optimization. I've re-submitted it in self-contained form here: https://github.com/cloudscape-design/components/pull/4609

PR #4151, opened 2026-01-03
refactor: replace Box spacing CSS classes with inline styles

TrevorBurnham:
### Description

I noticed that the styles for `Box` emitted 40KB of CSS, with a separate classname for every possible combination of spacing styles. This PR switches to declaring the spacing with inline styles, reducing the size of the CSS that gets shipped to the browser by ~20KB.

### How has this been tested?

I've update the unit tests, but ultimately what matters here is what the screenshot tests say.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/box/__tests__/box.test.tsx
+++ b/src/box/__tests__/box.test.tsx
@@ -14,6 +14,11 @@ function renderBox(props: BoxProps = {}): BoxWrapper {
   return new BoxWrapper(element);
 }
 
+// Helper to get computed style property from element
+function getStyleProperty(element: HTMLElement, property: string): string {
+  return element.style.getPropertyValue(property) || (element.style as any)[property] || '';
+}
+
 function testClassNamesForProperty<PropertyType>(propertyName: string, values: Array<PropertyType>, prefix: string) {
   test(`sets the right className for ${propertyName} property`, () => {
     const expectedClassNames = values.map(suffix => styles[`${prefix}-${suffix}`]);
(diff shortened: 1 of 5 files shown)


SpyZzey:
Hey,
thank you for your effort to create these changes. 
We have discussed this PR in our meeting yesterday and decided, we won't proceed with merging it into prod for now.
The main reason being, it adds some unnecessary complexity and won't work in SSR with our [recommended Content-Security-Policy](https://cloudscape.design/get-started/dev-guides/csp/).

TrevorBurnham:
@SpyZzey I'm glad to hear this was discussed. But as far as CSP goes, I think some important context may have been missed:

It's already the case that several component use inline styles, causing errors during SSR if the CSP you referred to is used. (You can see this in internal tickets like V1294454283 and V1698235459.) In many cases, there's no practical alternative to such styles. You can see that AWS consoles use a CSP policy that's compatible with inline styles in SSR.

For that reason, I've been working internally to get the guidelines changed, specifically to include `style-src-attr: unsafe-inline` in the recommended CSP policy. You can see more details of that proposal in Aristotle-Feedback-870.

SpyZzey:
@TrevorBurnham I don't think changing the CSP to allow for 'unsafe-inline' is the right approach as it removes a security layer to prevent already injected styles. In Aristotle-Feedback-870, you have shared a link to the consoles broader CSP docs.   It refers to dmqfAgZs1cjd (quite outdated though), which aims at improving the security in more consoles by using a stricter CSP. There is also a link to sXkrAsBzrVOp that lists some related TTs at the bottom. Unfortunately, I don't have access to it to see what caused them and if CSP could have had an impact on it, but it might be interesting to analyze.

We are currently investigating different possibilities of expanding the SSR support and solving problems regarding the inline styles in SSR as this also causes issues with our Style API. Because SSR is not explicitly supported yet, this will not have high priority. Some solutions, e.g. using a nonce and injecting inline style tags, might work for this solution too. We are keeping this PR in mind and will come back once a promising solution is found.

PR #4153, opened 2026-01-05
feat: Add custom option renderer for autosuggest

SpyZzey:
### Description
This pull request introduces support for custom rendering of dropdown items in the `Autosuggest` component by adding a new `renderOption` prop. This allows consumers to provide a custom function to render dropdown options. The implementation ensures that custom content replaces the default rendering and removes styling. 

Related links, issue #, if available: [EVgjAHAVR6A7], [9o4IAwsl4Rmu], [D349419297]

### How has this been tested?
- Manual testing
- Unit testing

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/autosuggest/custom-render-option.page.tsx
+++ b/pages/autosuggest/custom-render-option.page.tsx
@@ -38,7 +38,6 @@ export default function AutosuggestPage() {
   const ref = useRef<AutosuggestProps.Ref>(null);
 
   const renderOption: AutosuggestProps.ItemRenderer = ({ item }) => {
-    console.log(item);
     if (item.type === 'entered-text') {
       return <div>Entereted Text: {item.option.value}</div>;
     } else if (item.type === 'group') {


src/autosuggest/__tests__/render-option.test.tsx
```diff
@@ -0,0 +1,342 @@
...
+    const renderAutosuggest = (props?: Partial<AutosuggestProps>) =>
+      renderAutosuggestWrapper(props, listType === 'virtual');
+
+    describe(`${listType} list`, () => {
+      test('renders custom option content', () => {
+        const renderOption = jest.fn(() => <div>Custom</div>);
+        const wrapper = renderAutosuggest({ options: defaultOptions, renderOption });
+        wrapper.focus();
+
+        expect(renderOption).toHaveBeenCalled();
```
pan-kot: nit: Since "Custom" content is only defined by the renderOption - it is not really necessary to wrap it in fn() and check if it was called.

src/autosuggest/__tests__/render-option.test.tsx
```diff
@@ -0,0 +1,342 @@
...
+        const option = { label: 'Test', value: '1' };
+        const wrapper = renderAutosuggest({
+          options: [option],
+          value: '1',
+          renderOption,
+        });
+        wrapper.focus();
+        expect(renderOption).toHaveBeenCalledWith(
+          expect.objectContaining({
+            item: expect.objectContaining({ selected: true }),
```
pan-kot, line 140: nit: In this test we render custom content but check for `selected: true` in the renderOption call. In the test above we check the content instead. I think either is ok, but it would be nice to keep the assertions consistent and avoid unnecessary code (e.g. no need in fn() when checking content or no need in custom content when checking function call).

src/autosuggest/__tests__/render-option.test.tsx
```diff
@@ -0,0 +1,342 @@
...
+        const renderOption = jest.fn(props => <div>{props.item.disabled ? 'disabled' : 'enabled'}</div>);
+        const wrapper = renderAutosuggest({
+          options: [{ label: 'Test', value: '1', disabled: true }],
+          renderOption,
+        });
+        wrapper.focus();
+        expect(renderOption).toHaveBeenCalledWith(
+          expect.objectContaining({
+            item: expect.objectContaining({
+              disabled: true,
```
pan-kot, line 194: same as above - let's either check function call or content and remove irrelevant code

src/autosuggest/__tests__/render-option.test.tsx
```diff
@@ -0,0 +1,342 @@
...
+            item: expect.objectContaining({
+              type: 'group',
+              disabled: true,
+            }),
+          })
+        );
+      });
+    });
+  }
+});
```
pan-kot, line 357: There seems to be no test to assert that the option rendering defers to default when renderOption returns null.

src/autosuggest/interfaces.ts
```diff
@@ -56,6 +56,44 @@ export interface AutosuggestProps
...
+   * Specifies a render function to render custom options in the dropdown menu.
+   *
+   * The item inside the props has a different shape depending on its type:
+   *
+   *
+   * ### item
+   *
+   * - `type` ('item') - The item type.
+   * - `index` (number) - The item's absolute position in the dropdown.
+   * - `option` (Option) - The original item configuration.
```
pan-kot: Should we say "option" instead of "item" here?

src/autosuggest/interfaces.ts
```diff
@@ -56,6 +56,44 @@ export interface AutosuggestProps
...
+   *
+   * ### use-entered
+   *
+   * - `type` ('use-entered') - The item type.
+   * - `option` (Option) - The use-entered item configuration.
+   *
+   * When providing a custom `renderOption` implementation, it fully replaces the default visual rendering and content for that item.
+   * The component still manages focus, keyboard interactions, and selection state, but it no longer applies its default item layout or typography.
+   *
+   * When returning `null`, the default styling will be applied.
```
pan-kot: It is not just styling, but also content. Should we say: "When renderOption returns null, the default rendering will apply" - or something like that?

src/autosuggest/virtual-list.tsx
```diff
@@ -63,8 +66,29 @@ const VirtualList = ({
           hasDropdownStatus
         );
 
+        let parentProps: AutosuggestRenderItemParentProps | undefined = undefined;
```
pan-kot: This block is shared between plain-list and virtual-list. Should we move it to a util or compute before passing to list?

pages/autosuggest/custom-render-option.page.tsx
```diff
@@ -0,0 +1,87 @@
...
+    virtualScroll?: boolean;
+  }>
+>;
+
+export default function AutosuggestPage() {
+  const [value, setValue] = useState('');
+  const ref = useRef<AutosuggestProps.Ref>(null);
+
+  const renderOption: AutosuggestProps.ItemRenderer = ({ item }) => {
+    console.log(item);
```
pan-kot:
```suggestion
```

PR #4157, opened 2026-01-05
fix: Make truncated filename tooltip keyboard accessible (file-token-group)

Who-is-PS:
### Description

Adds keyboard accessibility to the file token tooltip. The filename container is now focusable (tabIndex={0}) with onFocus/onBlur handlers, allowing keyboard users to view the full filename when text is truncated.

`AWSUI-61600`

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
     "start:integ": "cross-env NODE_ENV=development webpack serve --config pages/webpack.config.integ.cjs",
     "start:react18": "npm-run-all --parallel start:watch start:react18:dev",
     "start:react18:dev": "cross-env NODE_ENV=development REACT_VERSION=18 webpack serve --config pages/webpack.config.cjs",
+    "postinstall": "prepare-package-lock",
     "prepare": "husky"
   },
   "dependencies": {
--- a/src/button-dropdown/__integ__/performance-marks.test.ts
+++ b/src/button-dropdown/__integ__/performance-marks.test.ts
@@ -1,39 +1,27 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
+
 import { BasePageObject } from '@cloudscape-design/browser-test-tools/page-objects';
 import useBrowser from '@cloudscape-design/browser-test-tools/use-browser';
 
-function setupTest(
-  pageName: string,
-  testFn: (parameters: {
-    page: BasePageObject;
-    getMarks: () => Promise<PerformanceMark[]>;
-    getElementPerformanceMarkText: (id: string) => Promise<string>;
-  }) => Promise<void>
-) {
-  return useBrowser(async browser => {
-    const page = new BasePageObject(browser);
-    await browser.url(`#/light/button-dropdown/${pageName}`);
-    const getMarks = async () => {
-      await new Promise(r => setTimeout(r, 200));
-      const marks = await browser.execute(() => performance.getEntriesByType('mark') as PerformanceMark[]);
-      return marks.filter(m => m.detail?.source === 'awsui');
-    };
-    const getElementPerformanceMarkText = (id: string) => page.getText(`[data-analytics-performance-mark="${id}"]`);
-
-    await testFn({ page, getMarks, getElementPerformanceMarkText });
-  });
+function getMarkSelector(mark: PerformanceMark) {
+  return `[data-analytics-performance-mark="${mark.detail.instanceIdentifier}"]`;
 }
 
-describe('ButtonDropdown', () => {
-  test(
-    'Emits a single mark',
-    setupTest('main-action', async ({ getMarks, getElementPerformanceMarkText }) => {
-      const marks = await getMarks();
+test(
+  'ButtonDropdown emits a single mark',
+  useBrowser(async browser => {
+    const page = new BasePageObject(browser);
+    await browser.url('#/light/button-dropdown/main-action');
+    await new Promise(r => setTimeout(r, 200));
 
-      expect(marks).toHaveLength(1);
-      expect(marks[0].name).toBe('primaryButtonRendered');
-      expect(marks[0].detail).toMatchObject({
+    await page.waitForAssertion(async () => {
+      const allMarks = await browser.execute(() => performance.getEntriesByType('mark') as PerformanceMark[]);
+      const awsuiMarks = allMarks.filter(m => m.detail?.source === 'awsui');
+
+      expect(awsuiMarks).toHaveLength(1);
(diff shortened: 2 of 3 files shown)


src/file-token-group/file-token.tsx
```diff
@@ -135,14 +149,19 @@ function InternalFileToken({
           >
             <InternalSpaceBetween direction="vertical" size="xxxs">
               <div
+                className={styles['file-name-button']}
                 onMouseOver={() => setShowTooltip(true)}
                 onMouseOut={() => setShowTooltip(false)}
+                onFocus={() => setShowTooltip(true)}
+                onBlur={() => setShowTooltip(false)}
+                role={isTruncated ? 'button' : undefined}
```
avinashbot, line 149: In addition to the role, I think it would help to communicate the state information when the tooltip is visible, something like

```suggestion
                role={isTruncated ? 'button' : undefined}
                aria-expanded={isTruncated ? showTooltip : undefined}
```

We discussed the scenario of the tooltip being confusing for non-sighted screen readers, so I think this should help that case - the role (buttons) explains that it is interactive, and the aria-expanded explains that something related to this control (a role=tooltip) is open. If we need to, we can later add a really obvious announcement, like "Tooltip opened/visible", but this additional property should be enough to first solve this.

Who-is-PS: Thanks @avinashbot! I've added the `aria-expanded` attribute as you recommended:

```jsx
role={isTruncated ? 'button' : undefined}
aria-expanded={isTruncated ? showTooltip : undefined}
```

This should help screen reader users understand when the tooltip is visible, complementing the `role="button"` that indicates interactivity.

Who-is-PS:
Received translations from i18n package (CR-252051089) and merged them into main of AWS-UI-Components-I18n

src/file-token-group/file-token.tsx
```diff
@@ -90,6 +91,19 @@ function InternalFileToken({
...
+  useEffect(() => {
+    const checkTruncation = () => {
+      setIsTruncated(isEllipsisActive());
+    };
+
+    // Check immediately after mount
+    checkTruncation();
+
+    // Check on window resize
+    window.addEventListener('resize', checkTruncation);
```
avinashbot: There are things other than a full window resize that can cause the truncation to trigger (opening panels, content changes, layout updates) so a ResizeObserver (we have a `useResizeObserver`) might be more "correct" here. But I think it's fine — the max width of the token is 230px, it's probably very unlikely that the component will be placed in a layout that narrow, so there's no need to pay that performance tax unless we have to.

avinashbot (dismissed):

src/file-token-group/__tests__/file-token-group.test.tsx
```diff
@@ -243,21 +243,55 @@ describe('Tooltip', () => {
     fireEvent.keyDown(window, { key: 'Escape', code: 'Escape' });
     expect(document.querySelector(`.${tooltipStyles.root}`)).toBeNull();
   });
+
+  test('Should show tooltip on keyboard focus', () => {
+    const wrapper = render({ items: [{ file: file3 }], alignment: 'horizontal' });
+
+    fireEvent.focus(wrapper.findFileToken(1)!.findFileName().getElement());
+    expect(document.querySelector(`.${tooltipStyles.root}`)).not.toBeNull();
```
georgylobko: should you check the content of the tooltip instead of 'not null'?

src/file-token-group/__tests__/file-token-group.test.tsx
```diff
@@ -243,21 +243,55 @@ describe('Tooltip', () => {
...
+    expect(fileNameElement).toHaveFocus();
   });
 
   test('Focus is dispatched to the previous token when removing the token at the end', () => {
     const wrapper = renderStateful({ items: [{ file: file1 }, { file: file2 }] });
     wrapper.findFileToken(2)!.findRemoveButton().click();
 
-    expect(wrapper.findFileToken(1)!.findRemoveButton().getElement()).toHaveFocus();
+    // After dismissal, focus moves to the filename container (first focusable element in the token)
+    const fileNameElement = wrapper.findFileToken(1)!.findFileName().getElement().parentElement;
```
georgylobko: [non-blocking]: I'd be easier to search by the container classname - `file-name-button`

src/file-token-group/file-token.tsx
```diff
@@ -135,14 +149,20 @@ function InternalFileToken({
           >
             <InternalSpaceBetween direction="vertical" size="xxxs">
               <div
+                className={styles['file-name-button']}
```
georgylobko: [nit] "button" in the classname is a bit misleading. I'd suggest to rename it to `file-name-container` instead

src/file-token-group/file-token.tsx
georgylobko: I would recommend to switch to the resize observer as it has a better performance than the `resize` event

Who-is-PS: Done! Switched to useResizeObserver to observe the filename container element directly. This gives us better performance and also handles container-level resizes (panels, layout changes) that the window resize event would miss.

src/file-token-group/__tests__/file-token-group.test.tsx
Who-is-PS: Good point! Updated the assertions to verify the tooltip displays the correct file name.

src/file-token-group/__tests__/file-token-group.test.tsx
Who-is-PS: Done! Updated to use the file-name-container classname selector instead of parentElement.

src/file-token-group/file-token.tsx
Who-is-PS: Done! Renamed to file-name-container

src/file-token-group/__tests__/file-token-group.test.tsx
```diff
@@ -228,7 +228,7 @@ describe('Tooltip', () => {
     const wrapper = render({ items: [{ file: file3 }], alignment: 'horizontal' });
 
     fireEvent.mouseEnter(wrapper.findFileToken(1)!.findFileName().getElement());
-    expect(document.querySelector(`.${tooltipStyles.root}`)).not.toBeNull();
+    expect(document.querySelector(`.${tooltipStyles.root}`)).toHaveTextContent(file3.name);
```
georgylobko, line 231: [non-blocking, could be a follow-up] Should you use test wrapper instead such as `createWrapper().findTooltip()`?

PR #4159, opened 2026-01-06
fix: Fixes trembling of table columns when resizing

pan-kot:
### Description

This fixes a table headers trembling when resizing. The issue was reproducible on complex tables and especially those with sticky headers. 

Before without sticky header:

https://github.com/user-attachments/assets/c7b9c82b-7de8-4bb0-8d7b-13c56baa9683

Before with sticky header:

https://github.com/user-attachments/assets/6a7f8a42-a395-46ec-ad9d-b7791317d386

After:

https://github.com/user-attachments/assets/40fafb63-844f-457f-8538-8cfb2eba8876

Rel: AWSUI-61443

Resolves: https://github.com/cloudscape-design/components/discussions/4010

### How has this been tested?

* Existing unit and integration tests, screenshot tests

The fix can be tested manually on any table with resizable columns but especially those with lots of content. It is worth testing it in both React 16 and 18 and also ensure the sticky headers column sync works fine in both resizable and not resizable tables.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/table/resizer/index.tsx
+++ b/src/table/resizer/index.tsx
@@ -1,6 +1,7 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
-import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react';
+
+import React, { useCallback, useEffect, useRef, useState } from 'react';
 import clsx from 'clsx';
 
 import { useStableCallback, useUniqueId } from '@cloudscape-design/component-toolkit/internal';
@@ -8,11 +9,11 @@ import { getIsRtl, getLogicalBoundingClientRect, getLogicalPageX } from '@clouds
 import { useSingleTabStopNavigation } from '@cloudscape-design/component-toolkit/internal';
 
 import DragHandleWrapper from '../../internal/components/drag-handle-wrapper';
+import { useThrottleCallback } from '../../internal/hooks/use-throttle-callback';
 import { useVisualRefresh } from '../../internal/hooks/use-visual-mode';
 import { KeyCode } from '../../internal/keycode';
 import handleKey, { isEventLike } from '../../internal/utils/handle-key';
 import { scrollElementIntoView } from '../../internal/utils/scrollable-containers';
-import { throttle } from '../../internal/utils/throttle';
 import { DEFAULT_COLUMN_WIDTH } from '../use-column-widths';
 import { getHeaderWidth, getResizerElements } from './resizer-lookup';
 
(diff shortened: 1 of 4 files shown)


src/table/resizer/index.tsx
```diff
@@ -168,21 +170,22 @@ export function Resizer({
     [minWidth, onWidthUpdate, updateTrackerPosition]
   );
 
-  const resizeColumn = useCallback(
-    (offset: number) => {
-      const elements = getResizerElements(resizerToggleRef.current);
-      if (!elements) {
-        return;
-      }
+  const resizeColumn = useMemo(
```
pan-kot: This function is called on every pointer move - so throttling helps by ignoring intermediate updates.

src/table/use-sticky-header.ts
```diff
@@ -24,11 +24,6 @@ export const useStickyHeader = (
       secondaryTableRef.current &&
       tableWrapperRef.current
     ) {
-      // Using the tableRef getBoundingClientRect().width instead of the theadRef because in VR
-      // the tableRef adds extra padding to the table and by default the theadRef will have a width
-      // without the padding and will make the sticky header width incorrect.
-      secondaryTableRef.current.style.inlineSize = `${tableRef.current.getBoundingClientRect().width}px`;
```
pan-kot, line 30: The table column width utils set widths of both normal and sticky headers at the same time, so there is no need to sync table widths.

NathanZlion (approved):

src/table/use-sticky-header.ts
georgylobko: > The table column width utils set widths of both normal and sticky headers at the same time

Could you point out to the code that does that and why did you change it in this PR?

src/table/resizer/index.tsx
georgylobko: [non-blocking] I checked our codebase and notice multiple instance using throttle with useMemo/useCallback. I'd suggest to introduce an abstraction `useThrottle` for them. WDYT?

pan-kot: Makes sense. I added useThrottleCallback (similar to existing useDebounceCallback).

src/table/use-sticky-header.ts
pan-kot: Certainly! Here is where the col widths are synchronised: https://github.com/cloudscape-design/components/blob/main/src/table/use-column-widths.tsx#L141

When researching columns trembling, I realised that the large part of it was due to this line of code (only when sticky header is on). Here we measure the total size of the real table and set it to the sticky table. Due to the async nature of the code, this measurement is not that precise - and often happens before the column widths are settled - and thus contributing to the trembling. I tried to optimise it first, but then realise that this piece of code seems to be no longer relevant, as the same explicit sizes are used for all columns anyways.

PR #4167, opened 2026-01-07
fix: Restore focus to tutorial panel on dismiss

NathanZlion:
### Description
Fix Accessibility issue for focus on the tutorial panel component. When a user exits/dismisses a tutorial, focus is automatically restored to the tutorial panel header to maintain proper keyboard navigation flow.

Related links, issue AWSUI-61570

### How has this been tested?
Tested Locally
Added integration test to check behavior and run it locally

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/tutorial-panel/index.tsx
+++ b/src/tutorial-panel/index.tsx
@@ -48,6 +48,8 @@ export default function TutorialPanel({
         className={clsx(baseProps.className, styles['tutorial-panel'])}
         ref={mergedRef}
         tabIndex={-1}
+        // Adding attributes conditionally since we don't want to point to
+        // a non existent header in the aria-labelledby
         {...(!context.currentTutorial && {
           role: 'region',
           'aria-labelledby': headingId,


src/tutorial-panel/index.tsx
```diff
@@ -29,14 +30,24 @@ export default function TutorialPanel({
   const baseProps = getBaseProps(restProps);
   const context = useContext(hotspotContext);
 
+  function handleExitTutorial(event: NonCancelableCustomEvent<TutorialPanelProps.TutorialDetail>) {
+    context.onExitTutorial(event);
+    __internalRootRef.current?.focus();
```
gethinwebster: I think we generally avoid using the `__internalRootRef` directly, and instead should create a new ref here and then `mergeRefs`

src/tutorial-panel/styles.scss
```diff
@@ -5,10 +5,18 @@
...
   padding-block-end: awsui.$space-m;
   padding-inline: awsui.$space-l;
+
+  &:focus {
+    outline: none;
+  }
+  @include focus-visible.when-visible {
+    @include styles.focus-highlight(0px);
+  }
+  @include styles.styles-reset;
```
gethinwebster: why is the styles reset moved? typically it should be the first entry, so it doesn't override any styles that we want to add

NathanZlion: oh my bad... that's right, Fixed

src/tutorial-panel/index.tsx
NathanZlion: Done,

src/tutorial-panel/index.tsx
```diff
@@ -25,18 +28,26 @@ export default function TutorialPanel({
...
+    context.onExitTutorial(event);
+    panelRef.current?.focus();
+  }
+
+  const mergedRef = useMergeRefs(panelRef, __internalRootRef);
+
   return (
     <>
-      <div {...baseProps} className={clsx(baseProps.className, styles['tutorial-panel'])} ref={__internalRootRef}>
+      <div {...baseProps} className={clsx(baseProps.className, styles['tutorial-panel'])} ref={mergedRef} tabIndex={-1}>
```
avinashbot: We need a couple more props here; we need to move focus to an element with a role and a label, so this needs a role (`role="region"` is probably my best bet) and an aria-labelledby (maybe using a random id provided to the heading? you'll also have to make sure the aria-labelledby is removed when the heading isn't visible)

PR #4174, opened 2026-01-08
fix: Fix sticky state listener logic to listen size change

NathanZlion:
### Description

The drawer footer would incorrectly remain sticky on initial page load when the screen height was small (in the 190-250px range). The footer would only correct itself to non-sticky after triggering a resize event. This was caused by getBoundingClientRect() returning stale/incorrect measurements (for exapmle. 235px instead of the actual 190px) during the initial layout calculation.

Removed window resize event listener and added `useResizeObserver` to observe parent element and drawer element and make sure the checks are done. This fixes the inital load on small screens to have the correct footer stickiness value.

### How has this been tested?

Tested on small screen height (240px) and made sure that when reload the footer becomes non-sticky.

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/drawer/use-sticky-footer.ts
```diff
@@ -41,15 +43,12 @@ export function useStickyFooter({
     STICKY_STATE_CHECK_THROTTLE_DELAY
   );
 
-  useLayoutEffect(() => {
-    window.addEventListener('resize', checkStickyState);
-    checkStickyState();
+  const getParentElement = useCallback(() => {
+    return drawerRef.current?.parentElement ?? null;
+  }, [drawerRef]);
```
georgylobko: You won't get any benefits using the ref as a dependency here as it's a mutable object. You can use it directly without `useCallback`

src/drawer/use-sticky-footer.ts
```diff
@@ -41,15 +43,12 @@ export function useStickyFooter({
...
 
-    return () => {
-      window.removeEventListener('resize', checkStickyState);
-      checkStickyState.cancel();
-    };
-  }, [checkStickyState]);
+  useResizeObserver(getParentElement, checkStickyState);
+  useResizeObserver(() => drawerRef.current, checkStickyState);
```
georgylobko: Do you need to subscribe to both `drawerRef` and its parent? Does changing one element entail changing another one?

NathanZlion: I think we can just observe the Parent Element only since that's what's containing the drawer inside it, I think it would be enough

PR #4186, opened 2026-01-14
feat: Feature notifications

georgylobko:
### Description

Introduced a feature notifications API that allows dynamically injecting a feature notifications drawer on pages using the AppLayout component.

The new API includes 3 methods:
- `registerFeatureNotifications` - registers a new drawer
- `showFeaturePromptIfPossible` - manually shows a feature prompt next to the drawer's trigger button
- `clearFeatureNotifications` - clears the feature notifications drawer

Also enhanced the existing feature prompt internal component by adding dismissal context to the onDismiss method (blur, close-button, outside-click, etc.).

### How has this been tested?

U tests / manually

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/i18n/messages-types.ts
+++ b/src/i18n/messages-types.ts
@@ -330,7 +330,7 @@ export interface I18nFormatArgTypes {
     'ariaLabels.jumpToPageButtonLabel': never;
     'i18nStrings.jumpToPageInputLabel': never;
     'i18nStrings.jumpToPageError': never;
-    "i18nStrings.jumpToPageLoadingText": never;
+    'i18nStrings.jumpToPageLoadingText': never;
   };
   'panel-resize-handle': {
     'i18nStrings.resizeHandleAriaLabel': never;


src/app-layout/visual-refresh-toolbar/state/use-feature-notifications.tsx
```diff
@@ -0,0 +1,254 @@
...
+    if (
+      !shouldShowFeaturePrompt.current ||
+      !featureNotificationsData ||
+      !drawersIds?.length ||
+      !featurePromptRef.current
+    ) {
+      return;
+    }
+    if (drawersIds.includes(featureNotificationsData.id)) {
+      featurePromptRef?.current?.show();
```
georgylobko: Why not set focus immediately after opening the drawer? The drawer opens on the next react rendering cycle, so if we set focus right after registering, the dom element doesn't exist yet. Same issue in use-focus-control: https://github.com/cloudscape-design/components/blob/main/src/app-layout/utils/use-focus-control.ts#L145

src/app-layout/visual-refresh-toolbar/state/use-feature-notifications.tsx
```diff
@@ -0,0 +1,254 @@
...
+        ref={featurePromptRef}
+        onShow={() => {
+          triggerRef.current!.dataset!.awsuiSuppressTooltip = 'true';
+        }}
+        onDismiss={event => {
+          if (event.detail?.method !== 'blur') {
+            triggerRef?.current!.focus();
+          }
+          setFeaturePromptDismissed(true);
+          Promise.resolve().then(() => {
```
georgylobko, line 171: Async problem - we need to unset tooltip suppress after shifting focus to it.

src/internal/do-not-use/feature-prompt/interfaces.ts
```diff
@@ -14,7 +14,7 @@ export interface FeaturePromptProps {
    * Called when a user closes the prompt by using the close icon button,
    * clicking outside the prompt, shifting focus out of the prompt or pressing ESC.
    */
-  onDismiss?: NonCancelableEventHandler<null>;
+  onDismiss?: NonCancelableEventHandler<{ method?: string }>;
```
georgylobko: The method is a way the feature prompt was closed - `blur`, `escape`, `close-button`, `click-outside`.

src/plugins/index.ts
```diff
@@ -0,0 +1,9 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+'use client';
+export * from '../internal/plugins/widget/interfaces';
+export {
+  registerFeatureNotifications,
```
pan-kot: Should we group the APIs, same as we do for internal ones?

Like:
```
const api = {
  featureNotifications: {
    registerFeatureNotifications,
    showFeaturePromptIfPossible,
    clearFeatureNotifications,
  }
}

export default api;
```

pages/feature-notifications/feature-prompt.page.tsx
```diff
@@ -2,17 +2,130 @@
...
 import { Breadcrumbs, Containers, Navigation, Tools } from '../app-layout/utils/content-blocks';
 import labels from '../app-layout/utils/labels';
 import * as toolsContent from '../app-layout/utils/tools-content';
 import ScreenshotArea from '../utils/screenshot-area';
 
+registerFeatureNotifications({
+  id: 'local-feature-notifications',
+  suppressFeaturePrompt: false,
+  featuresPageLink: '/#/feature-notifications/feature-prompt?appLayoutToolbar=true',
+  filterFeatures: () => true,
```
pan-kot, line 88: nit: The `filterFeatures` is unnecessary here

pages/feature-notifications/feature-prompt.page.tsx
```diff
@@ -2,17 +2,130 @@
...
+  ],
+  mountItem: (container, data) => {
+    mount(data, container);
+
+    return () => unmount(container);
+  },
+  persistenceConfig: {
+    uniqueKey: 'feature-notifications',
+  },
+  // DON'T USE
```
pan-kot: What is this comment about?

src/app-layout/__tests__/runtime-feature-notifications.test.tsx
```diff
@@ -0,0 +1,368 @@
...
+  jest.useFakeTimers();
+  jest.setSystemTime(mockCurrentDate);
+});
+
+afterEach(() => {
+  jest.useRealTimers();
+});
+
+function delay() {
+  return act(() => {
```
pan-kot, line 100: Is act() really needed? I think the delay() function can potentially be replaced with `jest.runAllTimersAsync()` or, better, `jest.advanceTimersByTimeAsync(time)`.

Also, the name "delay()" it not semantically sound as we do not use it to delay anything - we use it to wait for the component delays.

src/app-layout/__tests__/runtime-feature-notifications.test.tsx
```diff
@@ -0,0 +1,368 @@
...
+    jest.runAllTimers();
+    return Promise.resolve();
+  });
+}
+
+async function renderComponent(jsx: React.ReactElement) {
+  const { container, rerender, ...rest } = render(jsx);
+  const wrapper = createWrapper(container).findAppLayout()!;
+  await delay();
```
pan-kot: why do we call delay() here?

pan-kot: Even if that is needed for all tests to wait until the initial appearance of feature notifications - I would call it explicitly in each test suite to represent the intent better.

src/app-layout/__tests__/runtime-feature-notifications.test.tsx
```diff
@@ -0,0 +1,368 @@
...
+    const activeDrawerWrapper = wrapper.findActiveDrawer()!;
+    const featureItems = activeDrawerWrapper.findList()!.findItems()!;
+
+    expect(activeDrawerWrapper.getElement()).toBeTruthy();
+
+    // We expect all features to be rendered since the default filter has been overridden
+    expect(featureItems).toHaveLength(3);
+  });
+
+  test('shows feature prompt for a latest features', async () => {
```
pan-kot: typo: We should say "the latest feature" or "latest features" here

src/app-layout/__tests__/runtime-feature-notifications.test.tsx
```diff
@@ -0,0 +1,368 @@
...
+
+    featurePromptWrapper.findDismissButton()!.click();
+    expect(featurePromptWrapper.findContent()).toBeFalsy();
+
+    expect(wrapper.findDrawerTriggerById(featureNotificationsDefaults.id)!.getElement()).toHaveFocus();
+    // ensure the tooltip on the trigger button is suppressed after switching the focus
+    expect(wrapper.findDrawerTriggerTooltip()).toBeFalsy();
+    fireEvent.pointerEnter(wrapper.findDrawerTriggerById(featureNotificationsDefaults.id)!.getElement());
+    // ensure the tooltip on the trigger button is suppressed after switching the focus
+    expect(wrapper.findDrawerTriggerTooltip()).toBeFalsy();
```
pan-kot, line 292: when should it be un-suppressed, do we have a test for that?

src/app-layout/__tests__/runtime-feature-notifications.test.tsx
```diff
@@ -0,0 +1,368 @@
...
+  });
+
+  test('shows feature prompt for a latest unseen features', async () => {
```
pan-kot, line 330: How is that different from the 'shows feature prompt for a latest features' test?

src/app-layout/__tests__/runtime-feature-notifications.test.tsx
```diff
@@ -0,0 +1,368 @@
...
+    expect(activeDrawerWrapper.getElement()).toBeTruthy();
+
+    // We expect all features to be rendered since the default filter has been overridden
+    expect(featureItems).toHaveLength(3);
+  });
+
+  test('shows feature prompt for a latest features', async () => {
+    awsuiWidgetPlugins.registerFeatureNotifications(featureNotificationsDefaults);
+    const { container } = await renderComponent(<AppLayout />);
+    await delay();
```
pan-kot, line 261: We call delay() inside renderComponent, so this call should not be necessary - there are tests where it is called and some where it is not called immediately after render. I recommend to remove it from the render helper and always call it explicitly.

src/app-layout/__tests__/runtime-feature-notifications.test.tsx
```diff
@@ -0,0 +1,368 @@
...
+    await waitFor(() => {
+      expect(mockPersistFeatureNotifications).toHaveBeenCalled();
+    });
+
+    const persistedFeaturesMap = mockPersistFeatureNotifications.mock.calls[0][1];
+
+    expect(persistedFeaturesMap).toHaveProperty('feature-1');
+    expect(persistedFeaturesMap).toHaveProperty('feature-2');
+    expect(persistedFeaturesMap).toHaveProperty('recent-seen-feature');
+    expect(persistedFeaturesMap).not.toHaveProperty('old-seen-feature');
```
pan-kot, line 412: nit: would it makes sense to move persistence-related tests to a separate file?

src/app-layout/visual-refresh-toolbar/drawer/feature-notifications-drawer-content.tsx
```diff
@@ -0,0 +1,93 @@
...
+}
+
+export function RuntimeContentPart<T>({ content, mountContent }: RuntimeContentPartProps<T>) {
+  const ref = useRef<HTMLDivElement>(null);
+
+  useEffect(() => {
+    if (!mountContent) {
+      return;
+    }
+    const container = ref.current!;
```
pan-kot: instead of expecting the content to be there to satisfy TS - should we include it to the if check above?

```
if (!mountContent || !ref.content) {
  return;
}
```

pages/feature-notifications/feature-prompt.page.tsx
```diff
@@ -34,48 +147,81 @@ export default function () {
...
+          navigation={<Navigation />}
+          tools={<Tools>{toolsContent.long}</Tools>}
+          content={
+            <>
+              <div style={{ marginBlockEnd: '1rem' }}>
+                <Header variant="h1" description="Basic demo">
+                  Demo page
+                </Header>
+              </div>
+              <Button
```
pan-kot: nit: these buttons do not have margins between them

pages/feature-notifications/feature-prompt.page.tsx
```diff
@@ -2,17 +2,130 @@
...
+  showFeaturePromptIfPossible,
+} from '~components/internal/plugins/widget';
 import { mount, unmount } from '~mount';
 
 import { Breadcrumbs, Containers, Navigation, Tools } from '../app-layout/utils/content-blocks';
 import labels from '../app-layout/utils/labels';
 import * as toolsContent from '../app-layout/utils/tools-content';
 import ScreenshotArea from '../utils/screenshot-area';
 
+registerFeatureNotifications({
```
pan-kot, line 82: These only work when using `appLayoutToolbar=true`. I suggest to test if the non-toolbar layout does not crash when using this feature in unit tests, and always render the toolbar variant on this test page - it is confusing otherwise.

pan-kot: The feature does not seem to be working correctly. The prompt shows when I open dev tools, but not upon page reload. I am testing it in Chrome.

https://github.com/user-attachments/assets/c896eb24-dc03-426c-b857-44f417f5b27a

pan-kot: When feature notifications are registered (I added a delay for this to be better noticeable) - the tools disappear.

https://github.com/user-attachments/assets/9bf11adc-bc8d-45a0-8d97-99ae99c8564b

src/app-layout/visual-refresh-toolbar/drawer/feature-notifications-drawer-content.tsx
```diff
@@ -0,0 +1,93 @@
...
+  featuresPageLink,
+}: {
+  features: Array<Feature<T>>;
+  mountItem?: MountContentPart<T>;
+  featuresPageLink?: string;
+}) {
+  const i18n = useInternalI18n('features-notification-drawer');
+
+  return (
+    <InternalDrawer header={i18n('i18nStrings.title', undefined)} disableContentPaddings={true}>
```
pan-kot, line 53: This means that the feature cannot be used w/o built-in i18n, and that changing the default messages is only possible with a workaround (by overriding our messages objects).

Should we include (optional) i18nStrings to the runtime API instead?

src/app-layout/visual-refresh-toolbar/drawer/feature-notifications-drawer-content.tsx
```diff
@@ -0,0 +1,93 @@
...
+          renderItem={item => ({
+            id: item.id,
+            content: (
+              <Box variant="h3">
+                <RuntimeContentPart mountContent={mountItem} content={item.header} />
+              </Box>
+            ),
+            secondaryContent: (
+              <>
+                {!!item.releaseDate && (
```
pan-kot: The `releaseDate` is required as per our types.

src/app-layout/visual-refresh-toolbar/drawer/feature-notifications-drawer-content.tsx
```diff
@@ -0,0 +1,93 @@
...
+                  <Box margin={{ top: 'xs' }} fontSize="body-s" color="text-body-secondary">
```
pan-kot: There is already a default padding between content and secondary content - why do we add some more?

pan-kot: Can we use space-between instead of boxes with margins?

src/app-layout/visual-refresh-toolbar/drawer/feature-notifications-drawer-content.tsx
```diff
@@ -0,0 +1,93 @@
...
+      return;
+    }
+    const container = ref.current!;
+    const destructor = mountContent(container, content);
+
+    return () => {
+      destructor?.();
+    };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
pan-kot, line 36: I would add a unit tests where registerFeatureNotifications is called multiple times - sync or async, to assert that the latest content and the latest mountItem are used.

src/app-layout/visual-refresh-toolbar/drawer/feature-notifications-drawer-content.tsx
```diff
@@ -0,0 +1,93 @@
...
+  mountItem?: MountContentPart<T>;
+  featuresPageLink?: string;
+}) {
+  const i18n = useInternalI18n('features-notification-drawer');
+
+  return (
+    <InternalDrawer header={i18n('i18nStrings.title', undefined)} disableContentPaddings={true}>
+      <Box
+        padding={{ top: 'm', left: 'xl', right: 'xl', bottom: 'm' }}
+        className={styles['runtime-feature-notifications-drawer-content']}
```
pan-kot: Overriding styles on our components is not recommended. Can we use a custom div here instead?

pan-kot: The same applies to the `.runtime-feature-notifications-footer`

src/app-layout/visual-refresh-toolbar/drawer/styles.scss
```diff
@@ -488,3 +488,11 @@ $ai-drawer-heider-height: 42px;
     }
   }
 }
+
+.runtime-feature-notifications-drawer-content {
```
pan-kot, line 492: Do we have plans to support test-utils for this feature? I believe we should, as the consumers should have a reliable way to assert the feature prompt and drawer content.

src/app-layout/visual-refresh-toolbar/state/use-app-layout.tsx
```diff
@@ -136,6 +137,9 @@ export const useAppLayout = (
     bottomDrawersFocusControl.setFocus();
   };
 
+  const { featureNotificationsProps, onOpenFeatureNotificationsDrawer, featureNotificationsMessageHandler } =
+    useFeatureNotifications({ getDrawersIds: () => drawers?.map(drawer => drawer.id) ?? [] });
```
pan-kot: Why is it necessary to pass the getDrawersIds to the hook? The feature notifications should not care about other drawers.

If I understand it correctly, the `registerFeatureNotifications` API is responsible for rendering the drawer and also the feature prompt. Can this two actions be joined? For instance, the `featureNotificationsProps.drawer` can include an optional `featurePrompt` object - describing the prompt. When it is present - the prompt is shown, but only during the initial render (app layout would handle that).

pan-kot: Otherwise, there is a bi-directional connection between the app layout state and feature notifications state, which can cause race conditions and is overall difficult to understand.

src/app-layout/visual-refresh-toolbar/state/use-feature-notifications.tsx
```diff
@@ -0,0 +1,254 @@
...
+  Feature,
+  FeatureNotificationsPayload,
+  FeatureNotificationsPersistenceConfig,
+  WidgetMessage,
+} from '../../../internal/plugins/widget/interfaces';
+import { AppLayoutProps } from '../../interfaces';
+import RuntimeFeaturesNotificationDrawer, { RuntimeContentPart } from '../drawer/feature-notifications-drawer-content';
+
+const DEFAULT_PERSISTENCE_CONFIG = {
+  uniqueKey: 'feature-notifications',
```
pan-kot: Should we make the key more unique, like 'awsui-feature-notifications'?

src/app-layout/visual-refresh-toolbar/state/use-feature-notifications.tsx
```diff
@@ -0,0 +1,254 @@
...
+function subtractDaysFromDate(currentDate: Date, daysToSubtract: number) {
+  daysToSubtract = daysToSubtract || 0;
+  const pastDate = new Date(currentDate);
+  pastDate.setDate(pastDate.getDate() - daysToSubtract);
+
+  return pastDate;
+}
+
+function filterOutdatedFeatures(features: Record<string, string>): Record<string, string> {
+  const cutoffDate = subtractDaysFromDate(new Date(), 180);
```
pan-kot: let's extract 180 to a named constant

src/app-layout/visual-refresh-toolbar/state/use-feature-notifications.tsx
```diff
@@ -0,0 +1,254 @@
...
+
+function subtractDaysFromDate(currentDate: Date, daysToSubtract: number) {
+  daysToSubtract = daysToSubtract || 0;
+  const pastDate = new Date(currentDate);
+  pastDate.setDate(pastDate.getDate() - daysToSubtract);
+
+  return pastDate;
+}
+
+function filterOutdatedFeatures(features: Record<string, string>): Record<string, string> {
```
pan-kot: The `Record<string, string>` is confusing. If I get it right - the first prop is the feature id and the second one is the stringified release date, but that is hard to track. I would create a named type for that to improve readability, e.g.:

```
// Code comment explaining the structure
type PersistedFeaturesDict = Record<string, string>;
```

or 

```
type FeatureId = string;
type FeatureReleaseDateString = string;
type PersistedFeaturesDict = Record<FeatureId, FeatureReleaseDateString>;
```

src/app-layout/visual-refresh-toolbar/state/use-feature-notifications.tsx
```diff
@@ -0,0 +1,254 @@
...
+    if (event.type === 'registerFeatureNotifications') {
+      const { payload } = event;
+      const features = getFeaturesToDisplay(payload);
+      if (features.length === 0) {
+        return;
+      }
+
+      setFeatureNotificationsData({ ...payload, features });
+
+      const persistenceConfig = payload.persistenceConfig ?? DEFAULT_PERSISTENCE_CONFIG;
```
pan-kot, line 120: nit: I would add some code comments here explaining the relation between the event payload and the data we extract from the persistence mechanism.

src/app-layout/visual-refresh-toolbar/state/use-feature-notifications.tsx
```diff
@@ -0,0 +1,254 @@
...
+      const __retrieveFeatureNotifications:
+        | ((persistenceConfig: FeatureNotificationsPersistenceConfig) => Promise<Record<string, string>>)
+        | undefined = (payload as any)?.__retrieveFeatureNotifications;
```
pan-kot: Can this `__retrieveFeatureNotifications` be used from the public API? If so - do we have use cases besides testing? If no - I recommend doing some internal version of the API that allows it and make sure that everything starting from `__` is omitted from the public plugin (using `getExternalProps` util).

src/app-layout/visual-refresh-toolbar/state/use-feature-notifications.tsx
```diff
@@ -0,0 +1,254 @@
...
+          } else {
+            setMarkAllAsRead(true);
+          }
+        }
+      );
+      return;
+    }
+
+    if (event.type === 'showFeaturePromptIfPossible') {
+      if (markAllAsRead) {
```
pan-kot: Why do we need this extra state? Can we check the featureNotificationsData instead?

src/app-layout/visual-refresh-toolbar/state/use-feature-notifications.tsx
pan-kot: If the feature prompt was rendered by the app layout (with an extra internal prop on the drawer) - then we could suppress the tooltip whenever the feature prompt is present, w/o introducing the data-attributes.

Is there a reason why the feature cannot be fully controlled with the React state alone?

src/app-layout/visual-refresh-toolbar/state/use-feature-notifications.tsx
```diff
@@ -0,0 +1,254 @@
...
+
+  const onOpenFeatureNotificationsDrawer = () => {
+    if (!featureNotificationsData || markAllAsRead) {
+      return;
+    }
+    const persistenceConfig = featureNotificationsData.persistenceConfig ?? DEFAULT_PERSISTENCE_CONFIG;
+    const featuresMap = featureNotificationsData.features.reduce((acc, feature) => {
+      return {
+        ...acc,
+        [feature.id]: feature.releaseDate.toString(),
```
pan-kot: the Date.toString() is browser and locale dependent. If the persistence is cross-browser - the keys won't match. If the user would change their locale - the keys won't match.

src/plugins/index.ts
georgylobko: We normally do not follow this notation, just expose methods without nesting as they are descriptive enough

pages/feature-notifications/feature-prompt.page.tsx
georgylobko: It is necessary, because otherwise they will be filtered out when `releaseDate` for features become outdated - 90 days or older. But we need to keep them consistent on the testing page

pages/feature-notifications/feature-prompt.page.tsx
georgylobko: This comment is for anyone using this page as a code example. Since this feature won't be used directly but will be proxied through a different layer (happy to discuss in DM), the intermediate layer needs to inject its own persistence - though this isn't part of the public API. We use a similar approach here: https://github.com/cloudscape-design/components/blob/main/pages/app-layout/multi-layout-with-hidden-instances.page.tsx#L18

src/app-layout/__tests__/runtime-feature-notifications.test.tsx
georgylobko: > Is act() really needed? 

yes, because the purpose of the `delay` is to test pending react updates - the ones coming after calling the runtime api

> I think the delay() function can potentially be replaced with jest.runAllTimersAsync() or, better, jest.advanceTimersByTimeAsync(time).

no, it cannot be replaced. here we need to wait for 2 things:
1. wait until react pending updates
2. wait until all macro- and micro-tasks are done, such as focus control etc.

I agree delay probably isn't the best name, but we use it across many test files for the runtime API. Open to suggestions for renaming it.

src/app-layout/__tests__/runtime-feature-notifications.test.tsx
georgylobko: > wait until the initial appearance of feature notifications

Yes, this is what it is used for. As mentioned above, this is a typical use case for the runtime api testing, I don't see much value calling it explicitly, because it's more like a way the runtime api works. As mentioned, I'm open for suggestions for renaming.

src/app-layout/__tests__/runtime-feature-notifications.test.tsx
georgylobko: I see the confusion. This is about ensuring the trigger's built-in tooltip doesn't show, like in the picture below. I've renamed it to be more explicit:
> // ensure the built-in tooltip on the trigger button (not the feature prompt) doesn't appear after focus changes.
image

src/app-layout/__tests__/runtime-feature-notifications.test.tsx
georgylobko: The difference is that in this test I mocked persistence - `feature-1` has already been seen by the user, so it's marked as seen. In this scenario, we show the _latest unseen feature_ - `feature-2` rather than just the
latest feature

src/app-layout/__tests__/runtime-feature-notifications.test.tsx
georgylobko: Fair enough. Removed from the render function and call it explicitly in each test

src/app-layout/__tests__/runtime-feature-notifications.test.tsx
georgylobko: I don't think so, because the persistence is tightly coupled with feature notifications functionality

pages/feature-notifications/feature-prompt.page.tsx
georgylobko: The first one is really weird - I reloaded the page 30 times (15 with dev tools, 15 without) and never hit this condition (the video attached). Is there anything special about your environment?

https://github.com/user-attachments/assets/f5c78fc4-eecf-4fd0-87e5-4caa74e91722

georgylobko: For the second one I realized that it's not an issue, but expected behavior:

https://cloudscape.design/components/app-layout/?tabId=api
> Drawers property. If you set both drawers and tools, drawers will take precedence.

We show either tools or drawers, but not both at the same time

pan-kot: I upgraded my Chrome to the newest version, turned off the greasemonkey plugin, and disabled focused page emulation - I do not know what else can it be. When I test it now - the feature prompt often appears upon page reload, but not consistently so. I appears more often when I do not record my screen - so it can be performance-dependant.


https://github.com/user-attachments/assets/613a35cc-6e67-43ba-9634-893de3b3cadb

pan-kot: @georgylobko, regarding the drawers and tools. Despite the behaviour is expected - there is an important nuance: when the drawers are added explicitly - the relation between drawers and tools will likely be observed and an action taken to replace tools with drawers. When one calls the runtime API - then the tools disappearing might come unnoticed and cause trouble.

I suggest to render a dev warning and prevent feature notifications from overriding tools when the tools feature is used - this will give the necessary feedback to the implementer so that they can adjust the implementation in order to use feature notifications in a non-breaking way.

src/app-layout/visual-refresh-toolbar/drawer/feature-notifications-drawer-content.tsx
georgylobko: No, we should not. These messages are pre-defined and cannot be changed by users, because we want to keep consistent UX for all use-cases

pages/feature-notifications/feature-prompt.page.tsx
georgylobko: @pan-kot there is the evidence it works - https://github.com/cloudscape-design/components/blob/feat/feature-notitications/src/app-layout/__tests__/runtime-feature-notifications.test.tsx#L240

I can add an additional integ test to make sure it works in a real browser. At this point I have no clue what could be a reason behind this. Any additional info would be helpful

georgylobko: Found the root cause and fixed it. The issue was a race condition on React 18 pages. Fixed by introducing a ref to check when the ref mounts

src/app-layout/visual-refresh-toolbar/drawer/feature-notifications-drawer-content.tsx
georgylobko: > There is already a default padding between content and secondary content - why do we add some more?

Because default ones don't meet our design reference

> Can we use space-between instead of boxes with margins? 

Yes

src/app-layout/visual-refresh-toolbar/drawer/feature-notifications-drawer-content.tsx
georgylobko: Good call! Added a test and an example to the demo page

src/app-layout/visual-refresh-toolbar/drawer/feature-notifications-drawer-content.tsx
georgylobko: yes. for both

src/app-layout/visual-refresh-toolbar/drawer/styles.scss
georgylobko: We don't. Consumers could use existing test utils - for the active drawer and the list component inside. That's the way the runtime api works

src/app-layout/visual-refresh-toolbar/state/use-app-layout.tsx
georgylobko: > Why is it necessary to pass the getDrawersIds to the hook?

Removed it and refactored the entire async code:
https://github.com/cloudscape-design/components/blob/feat/feature-notitications/src/app-layout/visual-refresh-toolbar/state/use-feature-notifications.tsx#L129

Introduced a new util useMountRefPromise that returns a ref callback and promise that resolves when the ref attaches to a mounted element. This eliminates the need to leak this abstraction across multiple places.
The ref resolves the promise as soon as it mounts - no race conditions and easier to read.
https://github.com/cloudscape-design/components/blob/feat/feature-notitications/src/internal/utils/promises.ts#L50

> featureNotificationsProps.drawer can include an optional featurePrompt object - describing the prompt

We could technically do that, but using the feature prompt is a very specific use case while the drawer object is generic across the application. Given the refactoring above, I don't think it makes sense anymore

pages/feature-notifications/feature-prompt.page.tsx
pan-kot: Got it! I would include a code comment to explain this bit

src/app-layout/visual-refresh-toolbar/state/use-feature-notifications.tsx
georgylobko: No, it's not designed for the public API. The only use case is for an intermediate layer to override the persistence engine (happy to discuss further in DM).

It's not exposed in the public API, and need to be declared via spread operator to avoid typescript complaining: https://github.com/cloudscape-design/components/blob/feat/feature-notitications/pages/feature-notifications/feature-prompt.page.tsx#L95
It allows us to avoid types duality

src/app-layout/visual-refresh-toolbar/state/use-feature-notifications.tsx
georgylobko: > If the feature prompt was rendered by the app layout (with an extra internal prop on the drawer) - then we could suppress the tooltip whenever the feature prompt is present, w/o introducing the data-attributes.

> Is there a reason why the feature cannot be fully controlled with the React state alone?

Yes, drawers are generic types while feature notifications are domain specific. I'd rather keep all the abstractions in one place, in this hook

src/app-layout/visual-refresh-toolbar/state/use-feature-notifications.tsx
georgylobko: Turned this into a calculated value: https://github.com/cloudscape-design/components/blob/feat/feature-notitications/src/app-layout/visual-refresh-toolbar/state/use-feature-notifications.tsx#L74

pages/feature-notifications/feature-prompt.page.tsx
srungta08: QQ: Shouldn't this be used within useEffect?

georgylobko: No, it shouldn't. This API is designed for micro-frontend environments where a team doesn't have control over or access to the outer AppLayout. The API needs to dynamically inject a drawer into that AppLayout, so it's intentionally designed to be called outside of react scope

src/internal/do-not-use/feature-prompt/interfaces.ts
pan-kot: Why is it typed as a string and not a union like you described in the comment?

src/app-layout/visual-refresh-toolbar/state/use-feature-notifications.tsx
```diff
@@ -0,0 +1,246 @@
...
+
+const FEATURE_NOTIFICATIONS_RETENTION_DAYS = 180;
+
+function subtractDaysFromDate(currentDate: Date, daysToSubtract: number) {
+  daysToSubtract = daysToSubtract || 0;
+  const pastDate = new Date(currentDate);
+  pastDate.setDate(pastDate.getDate() - daysToSubtract);
+
+  return pastDate;
+}
```
srungta08: Nit: Better approach
```
function subtractDaysFromDate(currentDate: Date, daysToSubtract: number = 0): Date {
  const pastDate = new Date(currentDate);
  pastDate.setDate(pastDate.getDate() - daysToSubtract);
  return pastDate;
}

```

src/plugins/index.ts
pan-kot: I don't think it is fair to compare it with a few other utils that we expose from components like `useFilesDragging` etc. that are focused on a single thing.

The closest thing to compare this for me is the internal plugin APIs - which is exposed as an object. I expect the public APIs to grow over time: we can add methods related to other features or more methods related to feature notifications etc. At some point we might as well deprecate the entire feature notifications in favour of something else. In all cases, having a more structured approach should be a win.

That does not have to be an object - we can as well do a dedicated namespace like `/plugins/feature-notifications`.

srungta08 (approved):

pages/feature-notifications/feature-prompt.page.tsx
pan-kot: When we do such things in components - we test them using the internal exports instead. For instance, there are test pages for internal dropdown, popover container - and other components, which are either not available publicly or offer richer internal APIs. Can we do the same here?

Or, we can e.g. introduce a private React context to inject persistence and use it on this test page.

src/app-layout/__tests__/runtime-feature-notifications.test.tsx
pan-kot: The use of runTimers() and act() everywhere can make tests pass by accident - hiding potential issues.

When I updated the delay method to this form:
```
async function delay() {
  await jest.runAllTimersAsync();
}
```

all tests passed except one:
```
test('shows feature prompt via triggering showFeaturePromptIfPossible when possible', async () => {
  awsuiWidgetPlugins.registerFeatureNotifications({ ...featureNotificationsDefaults, suppressFeaturePrompt: true });
  const { container } = renderComponent(<AppLayout />);
  await delay();

  const featurePromptWrapper = new FeaturePromptWrapper(container);
  expect(featurePromptWrapper.findContent()).toBeFalsy();

  awsuiWidgetPlugins.showFeaturePromptIfPossible();

  await delay();

  // ↓↓↓ FAILS HERE ↓↓↓
  await waitFor(() => {
    expect(featurePromptWrapper.findContent()!.getElement()).toHaveTextContent(
      'This is the first new feature content'
    );
  });
});
```

In that case, there was no issue with the implementation, but the delay was unnecessary either: the prompt is actually shown synchronously. This version of the test works:

```
test('shows feature prompt via triggering showFeaturePromptIfPossible when possible', async () => {
  awsuiWidgetPlugins.registerFeatureNotifications({ ...featureNotificationsDefaults, suppressFeaturePrompt: true });
  const { container } = renderComponent(<AppLayout />);
  await delay();

  const featurePromptWrapper = new FeaturePromptWrapper(container);
  expect(featurePromptWrapper.findContent()).toBeFalsy();

  act(() => awsuiWidgetPlugins.showFeaturePromptIfPossible());

  expect(featurePromptWrapper.findContent()!.getElement().textContent).toBe('This is the first new feature content');
});
```

Notice that I also removed the `waitFor()` - which also makes the test less precise. If we change the implementation so that the prompt will start showing with a bit of an async delay - these tests must fail, indicating the change in behaviour.

pan-kot: Regarding the name, I recommend calling it meaningfully to represent what is it exactly that we are waiting for, see example: https://github.com/cloudscape-design/chart-components/blob/main/src/core/__tests__/highcharts-utils.ts#L46

Notice how all different delays that can happen inside are deliberately called, and how `advanceTimersByTimeAsync` is used instead of `runAllTimers` so that the expectations are set clearly.

src/app-layout/__tests__/runtime-feature-notifications.test.tsx
pan-kot: I meant something different here. The intent of the test is clear: once the feature notifications drawer trigger has focus - the tooltip must be suppressed for either focus or hover action.

However, once the focus is removed - the tooltip shall no longer be suppressed - but there seems to be no test to verify that.

src/app-layout/__tests__/runtime-feature-notifications.test.tsx
pan-kot: ok, got it

src/app-layout/__tests__/runtime-feature-notifications.test.tsx
pan-kot: I get it, my suggestion is to separate the tests so that the navigation is a bit easier: there are still different areas to test, such as:
1. Feature notifications drawer content
2. Feature notifications drawer open/close behaviour
3. Feature prompt content
4. Feature prompt content focus and open/close behaviours
6. Filtering logic
7. Logic to determine the latest unseen feature (involves persistence)
8. Logic to mark all features as seen (involves persistence)
9. Runtime API to clear notifications
10. Runtime API to show prompt if possible
11. some more?

All these can share the same file and describe block or can be split into groups for the ease of maintenance. Say, if I make a UX change to the notifications drawer - I am not interested much in the persistence tests, unless they fail.

pan-kot: That is a minor suggestion though.

pages/feature-notifications/feature-prompt.page.tsx
pan-kot: @georgylobko the tools panel is still overridden by the runtime API - that is a blocker for me.

src/app-layout/visual-refresh-toolbar/drawer/feature-notifications-drawer-content.tsx
pan-kot: We expect different notifications to be used on different application pages. I think the default messages are fine, but my bet is that there will soon be the ask to customise it so that instead of saying "Latest feature releases" I can say e.g. "Latest {page name} releases".

In the worst case, the consumers will inject the messages they need by merging them into the i18n messages objects imported from us - which is a major risk, as there is no promise for those messages to stay backward-compatible.

src/app-layout/__tests__/runtime-feature-notifications.test.tsx
```diff
@@ -0,0 +1,422 @@
...
+} from '../../../lib/components/internal/persistence';
+import awsuiPlugins from '../../../lib/components/internal/plugins';
+import * as awsuiWidgetPlugins from '../../../lib/components/internal/plugins/widget';
+import * as awsuiWidgetInternal from '../../../lib/components/internal/plugins/widget/core';
+import { FeatureNotificationsPayload } from '../../../lib/components/internal/plugins/widget/interfaces';
+import createWrapper from '../../../lib/components/test-utils/dom';
+import FeaturePromptWrapper from '../../../lib/components/test-utils/dom/internal/feature-prompt';
+import { describeEachAppLayout } from './utils';
+
+const i18nMessages = {
```
pan-kot, line 20: What is the purpose of defining these messages if none gets tested?

src/app-layout/visual-refresh-toolbar/drawer/feature-notifications-drawer-content.tsx
pan-kot: Should we then use the `secondaryContent` at all? Why not using our custom layout and put everything to the `content` slot instead?

src/app-layout/__tests__/runtime-feature-notifications.test.tsx
```diff
@@ -0,0 +1,422 @@
...
+    expect(featureItems[1].findContent().getElement()).toHaveTextContent('New Feature 2 Header');
+    expect(featureItems[1].findSecondaryContent()!.getElement()).toHaveTextContent(
+      'This is the second new feature content'
+    );
+    expect(featureItems[1].findSecondaryContent()!.getElement()).toHaveTextContent('2024-12-01');
+
+    expect(activeDrawerWrapper!.find('a')!.getElement()).toHaveAttribute('href', '/features-page');
+  });
+
+  test('registerFeatureNotifications should override previous call when called multiple time', async () => {
```
pan-kot:
```suggestion
  test('registerFeatureNotifications should override previous call when called multiple times', async () => {
```

src/app-layout/__tests__/runtime-feature-notifications.test.tsx
```diff
@@ -0,0 +1,422 @@
...
+    awsuiWidgetPlugins.registerFeatureNotifications({ ...featureNotificationsDefaults, mountItem: undefined });
+    const { wrapper } = renderComponent(<AppLayout />);
+    await delay();
```
pan-kot: Is this delay necessary? The test also passes w/o it.

src/app-layout/__tests__/runtime-feature-notifications.test.tsx
```diff
@@ -0,0 +1,422 @@
...
+          content: 'This is the first new feature content',
+          contentCategory: 'Category A',
+          releaseDate: mockDate2025,
+        },
+      ],
+    });
+
+    await delay();
+
+    await waitFor(() => {
```
pan-kot, line 219: this waitFor is not needed - and I believe the same is true for many, if not all, other tests here

src/app-layout/visual-refresh-toolbar/drawer/styles.scss
pan-kot: The consumers should not rely on the fact that we use our list component in there - the implementation can as well change. There is also no deterministic way to find the "view all notifications link", and I believe the feature prompt test utils are all private, too.

src/app-layout/visual-refresh-toolbar/state/use-feature-notifications.tsx
pan-kot: Right, but a consumer can technically still call it if ignoring the "DON'T USE" comment - right?

src/app-layout/__tests__/runtime-feature-notifications.test.tsx
```diff
@@ -0,0 +1,422 @@
...
+
+const i18nMessages = {
+  'features-notification-drawer': {
+    'i18nStrings.title': 'Latest feature releases',
+    'i18nStrings.viewAll': 'View all feature releases',
+    'ariaLabels.closeButton': 'Close notifications',
+    'ariaLabels.content': 'Feature notifications',
+    'ariaLabels.triggerButton': 'Show feature notifications',
+    'ariaLabels.resizeHandle': 'Resize feature notifications',
+  },
```
pan-kot, line 28: I would give more descriptive names to i18n props, like `featureNotificationsDrawerTitle`, `featureNotificationsDrawerToggle` etc. Otherwise it is hard to understand what do these attributes represent: if the close button related to the drawer or the feature prompt, what is content, etc.

src/app-layout/visual-refresh-toolbar/state/use-feature-notifications.tsx
pan-kot: @georgylobko what is with toString() serialization? The new code passes the date object directly w/o calling toString() on it - but it works just the same and has the same flaw.

pan-kot: Btw, why do we need it to be a dictionary at all - isn't it enough to store the keys only?

pages/feature-notifications/feature-prompt.page.tsx
pan-kot: @georgylobko the `appLayoutToolbar=true` is still required to be explicitly set on the page so that the feature can be seen.

pages/feature-notifications/feature-prompt.page.tsx
```diff
@@ -34,48 +147,154 @@ export default function () {
...
+                      ref={triggerRef}
+                      onClick={() => {
+                        clearFeatureNotifications();
+                      }}
+                    >
+                      clear feature notifications
+                    </Button>
+                    <Button
+                      onClick={() => {
+                        registerFeatureNotifications({
```
pan-kot, line 298: This override button features too much duplicate code especially around the persistence part - can we extract it to a method? Like:

```
function registerDemoNotifications({ features }) {/* calls register api with given features */}
```

pages/feature-notifications/feature-prompt.page.tsx
```diff
@@ -34,48 +147,154 @@ export default function () {
...
+                                  150
+                                )
+                              );
+                              return result;
+                            },
+                          },
+                        });
+                      }}
+                    >
+                      Override
```
pan-kot, line 361: When I click override - the feature prompt is dismissed. Why?

src/app-layout/__tests__/runtime-feature-notifications.test.tsx
```diff
@@ -0,0 +1,422 @@
...
+  featuresPageLink: '/features-page',
+  mountItem: (container, data) => {
+    container.textContent = data;
+  },
+  persistenceConfig: {
+    uniqueKey: 'feature-notifications',
+  },
+};
+
+jest.mock('@cloudscape-design/component-toolkit', () => ({
```
pan-kot, line 70: when this mock is removed - all tests still pass

src/app-layout/__tests__/runtime-feature-notifications.test.tsx
```diff
@@ -0,0 +1,422 @@
...
+
+  test('clears feature notifications correctly', async () => {
+    awsuiWidgetPlugins.registerFeatureNotifications({ ...featureNotificationsDefaults, mountItem: undefined });
+    const { wrapper } = renderComponent(<AppLayout />);
+    await delay();
+    expect(wrapper.findDrawerTriggerById(featureNotificationsDefaults.id)).toBeTruthy();
+
+    awsuiWidgetPlugins.clearFeatureNotifications();
+
+    await delay();
```
pan-kot: is this delay needed?

src/app-layout/__tests__/runtime-feature-notifications.test.tsx
```diff
@@ -0,0 +1,422 @@
...
+    });
+  });
+
+  test('supports custom filterFeatures function', async () => {
+    awsuiWidgetPlugins.registerFeatureNotifications({
+      ...featureNotificationsDefaults,
+      filterFeatures: feature => {
+        const halfYearAgo = new Date();
+        halfYearAgo.setDate(halfYearAgo.getDate() - 180);
+        return feature.releaseDate >= halfYearAgo;
```
pan-kot, line 241: nit: this implementation is unnecessary as using `filterFeatures: () => true` would give the same result.

If you also want to verify that the payload is correct, it is best to use:
```
const filterFeatures = jest.fn().returns(() => true);
```

, in which case you can explicitly check that the function was called the correct amount of times and with correct payload

src/app-layout/__tests__/runtime-feature-notifications.test.tsx
```diff
@@ -0,0 +1,422 @@
...
+
+    await waitFor(() => {
+      const featurePromptWrapper = new FeaturePromptWrapper(container);
+      expect(featurePromptWrapper.findContent()!.getElement()).toHaveTextContent(
+        'This is the second new feature content'
+      );
+    });
+  });
+
+  test('should not show feature prompt for unseen features is suppressed', async () => {
```
pan-kot:
```suggestion
  test('should not show feature prompt for unseen features when suppressed', async () => {
```

src/app-layout/__tests__/runtime-feature-notifications.test.tsx
```diff
@@ -0,0 +1,422 @@
...
+
+    await waitFor(() => {
+      expect(featurePromptWrapper.findContent()!.getElement()).toHaveTextContent(
+        'This is the first new feature content'
+      );
+    });
+  });
+
+  test('should not show feature prompt if all feature are seen', async () => {
+    mockRetrieveFeatureNotifications.mockResolvedValue({
```
pan-kot, line 368: I changed this to:
```
    mockRetrieveFeatureNotifications.mockResolvedValue({
      'feature-1': 'mockDate2025.toString()',
      'feature-2': 'mockDate2024.toString()',
      'feature-old': 'mockDateOld.toString()',
    });
```

and the test still passed. Why do we need these values?

src/app-layout/visual-refresh-toolbar/drawer/feature-notifications-drawer-content.tsx
```diff
@@ -0,0 +1,94 @@
...
+                  <RuntimeContentPart mountContent={mountItem} content={item.header} />
+                </Box>
+              ),
+              secondaryContent: (
+                <SpaceBetween size="xs" direction="vertical">
+                  <Box fontSize="body-s" color="text-body-secondary">
+                    {formatDate(item.releaseDate)}
+                  </Box>
+                  {!!item.contentCategory && (
+                    <Box>
```
pan-kot, line 71: The `RuntimeContentPart` already wraps content with a div or span. Why do we need the surrounding Box elements for content category and content?

src/app-layout/visual-refresh-toolbar/drawer/feature-notifications-drawer-content.tsx
```diff
@@ -0,0 +1,94 @@
...
+                  <Box>
+                    <RuntimeContentPart mountContent={mountItem} content={item.content} />
+                  </Box>
+                </SpaceBetween>
+              ),
+            })}
+          />
+
+          {!!featuresPageLink && (
+            <footer className={styles['runtime-feature-notifications-footer']}>
```
pan-kot, line 84: Should we use drawer's footer for that?

https://cloudscape.design/components/drawer/?tabId=playground&example=with-footer

src/popover/body.tsx
```diff
@@ -85,14 +87,14 @@ const PopoverBody = React.forwardRef(
...
-          onClick={() => onDismiss?.()}
+          onClick={() => onDismiss?.('close-button')}
           ref={dismissButtonRef}
         />
       </div>
     );
 
     const isDialog = showDismissButton;
-    const shouldTrapFocus = showDismissButton && variant !== 'annotation';
+    const shouldTrapFocus = trapFocus || (showDismissButton && variant !== 'annotation');
```
pan-kot: Can we introduce a distinct variant for feature notifications instead?

src/internal/persistence/index.ts
```diff
@@ -58,3 +72,21 @@ export let retrieveAlertDismiss = async function (
 ): Promise<boolean> {
   return Promise.resolve(false);
 };
+
+// eslint-disable-next-line require-await
```
pan-kot: nit: should we disable these two eslint rules for the entire file instead?

pan-kot:
When register API is used from inside an iframe - the initial feature prompt is not shown. The "show a standalone prompt" still works - but focus is not handled correctly and click outside does not cause a dismiss.

I tested that by wrapping the new test page with the IframeWrapper util.

src/internal/do-not-use/feature-prompt/internal.tsx
```diff
@@ -53,6 +54,26 @@ function InternalFeaturePrompt(
...
+    const clickListener = (event: MouseEvent) => {
+      // Since the listener is registered on the window, `event.target` will incorrectly point at the
+      // shadow root if the component is rendered inside shadow DOM.
+      const target = event.composedPath ? event.composedPath()[0] : event.target;
+      if (!nodeBelongs(popoverBodyRef.current, target)) {
+        setShow(false);
+        fireNonCancelableEvent(onDismiss, { method: 'click-outside' });
+      }
+    };
+    window.addEventListener('click', clickListener, true);
```
pan-kot: This and the nodeBelongs can be a problem when the API is called from inside an iframe. I managed to confirm it by rendering the test page inside an iframe helper.

src/drawer/use-sticky-footer.ts
```diff
@@ -23,7 +23,7 @@ export function useStickyFooter({
 
   const checkStickyState = throttle(
     useCallback(() => {
-      if (!drawerRef.current || !footerRef.current) {
+      if (!drawerRef?.current || !footerRef?.current) {
```
pan-kot, line 26: why is this change needed?

src/app-layout/visual-refresh-toolbar/state/use-feature-notifications.tsx
```diff
@@ -0,0 +1,246 @@
...
+  const [seenFeatures, setSeenFeatures] = useState<PersistedFeaturesDict>({});
+  const featurePromptRef = useRef<FeaturePromptProps.Ref>(null);
+  const { ref: featurePromptMountRef, promise: featurePromptMountPromise } = useMountRefPromise();
+  const featurePromptMergedRef = useMergeRefs(featurePromptRef, featurePromptMountRef);
+  const allNotificationsSeen = useMemo(() => {
+    return featureNotificationsData?.features.every(feature => !!seenFeatures[feature.id]);
+  }, [featureNotificationsData, seenFeatures]);
+
+  const defaultFeaturesFilter = (feature: Feature<unknown>) => {
+    return feature.releaseDate >= subtractDaysFromDate(new Date(), 90);
```
pan-kot: let's use a constant for 90 as well

src/app-layout/visual-refresh-toolbar/state/use-feature-notifications.tsx
```diff
@@ -0,0 +1,246 @@
...
+
+export function useFeatureNotifications() {
+  const i18n = useInternalI18n('features-notification-drawer');
+  const [featurePromptDismissed, setFeaturePromptDismissed] = useState(false);
+  const [featureNotificationsData, setFeatureNotificationsData] = useState<FeatureNotifications | null>(null);
+  const [seenFeatures, setSeenFeatures] = useState<PersistedFeaturesDict>({});
+  const featurePromptRef = useRef<FeaturePromptProps.Ref>(null);
+  const { ref: featurePromptMountRef, promise: featurePromptMountPromise } = useMountRefPromise();
+  const featurePromptMergedRef = useMergeRefs(featurePromptRef, featurePromptMountRef);
+  const allNotificationsSeen = useMemo(() => {
```
pan-kot, line 70: nit: this memo is not really needed - the cost of checking the data (which is already filtered) is negligible.

src/app-layout/visual-refresh-toolbar/state/use-feature-notifications.tsx
```diff
@@ -0,0 +1,246 @@
...
+        iconName: 'suggestions',
+      },
+      ariaLabels: {
+        closeButton: i18n('ariaLabels.closeButton', undefined),
+        drawerName: i18n('ariaLabels.content', undefined) ?? '',
+        triggerButton: i18n('ariaLabels.triggerButton', undefined),
+        resizeHandle: i18n('ariaLabels.resizeHandle', undefined),
+      },
+      resizable: true,
+      defaultSize: 320,
```
pan-kot, line 105: nit: let's create a constant for default drawer size, too

src/app-layout/visual-refresh-toolbar/state/use-feature-notifications.tsx
```diff
@@ -0,0 +1,246 @@
...
+      badge: payload.badge,
+    };
+  };
+
+  function featureNotificationsMessageHandler(event: WidgetMessage) {
+    if (event.type === 'registerFeatureNotifications') {
+      warnOnce('feature notifications', 'Feature notifications drawer is overriding the tools panel');
```
pan-kot: This renders a warning when the feature notifications feature is used correctly - why?

We should only warn if the tools feature is actually used, and in that case the feature notifications must not be rendered.

pan-kot: Also, we should add a unit test for the warning.

src/internal/utils/promises.ts
```diff
@@ -37,3 +39,33 @@ export function makeCancellable<T>(promise: Promise<T>): {
...
 }
+
+/**
+ * Returns a ref callback and promise that resolves when the ref is attached to a mounted element.
+ * Useful for coordinating async operations that depend on DOM element availability, such as
+ * setting focus or measuring dimensions after mount.
+ *
+ * @returns An object containing a ref callback and a promise that resolves with the mounted element
+ */
+export function useMountRefPromise<T extends HTMLElement>() {
```
pan-kot, line 50: nice util!

src/app-layout/visual-refresh-toolbar/state/use-feature-notifications.tsx
```diff
@@ -0,0 +1,246 @@
...
+            }
+            setFeatureNotificationsData(data => (data ? { ...data, badge: true } : data));
+          }
+        }
+      );
+      return;
+    }
+
+    if (event.type === 'showFeaturePromptIfPossible') {
+      if (allNotificationsSeen) {
```
pan-kot: nit: we can simplify the check to:

```
if (event.type === 'showFeaturePromptIfPossible' && !allNotificationsSeen) {
  featurePromptRef.current?.show();
}
```

Alternatively, the event mapping can be made more explicit like:
```
switch (event.type) {
  case 'registerFeatureNotifications':
    return handleRegisterFeatureNotifications(...);
  case 'showFeaturePromptIfPossible':
    return showFeaturePromptIfApplicable();
  case 'clearFeatureNotifications':
    return setFeatureNotificationsData(null);
}
```

src/app-layout/visual-refresh-toolbar/state/use-feature-notifications.tsx
```diff
@@ -0,0 +1,246 @@
...
+  }
+
+  function getLatestUnseenFeature(): Feature<unknown> | null {
+    if (!featureNotificationsData) {
+      return null;
+    }
+
+    // Features array is already sorted in reverse chronological order (most recent first)
+    // Find the first feature that hasn't been seen
+    for (const feature of featureNotificationsData.features) {
```
pan-kot: nit: this can be simplified to:

```
featureNotificationsData.features.find(feature => !seenFeatures[feature.id]) ?? null;
```

or even one line:

```
function getLatestUnseenFeature() {
  return featureNotificationsData?.features.find(feature => !seenFeatures[feature.id]) ?? null;
}
```

src/app-layout/visual-refresh-toolbar/state/use-feature-notifications.tsx
```diff
@@ -0,0 +1,246 @@
...
+        return feature;
+      }
+    }
+
+    return null;
+  }
+
+  function renderLatestFeaturePrompt({ triggerRef }: RenderLatestFeaturePromptProps) {
+    const latestFeature = getLatestUnseenFeature();
+    if (!(triggerRef?.current && featureNotificationsData && latestFeature)) {
```
pan-kot: this can be simplified to:
```
    if (!triggerRef.current || !latestFeature) {
      return null;
    }
```

src/app-layout/visual-refresh-toolbar/state/use-feature-notifications.tsx
```diff
@@ -0,0 +1,246 @@
...
+    }
+    return (
+      <FeaturePrompt
+        ref={featurePromptMergedRef}
+        onShow={() => {
+          triggerRef.current!.dataset!.awsuiSuppressTooltip = 'true';
+        }}
+        onDismiss={event => {
+          if (event.detail?.method !== 'blur') {
+            triggerRef?.current!.focus();
```
pan-kot, line 168: This `.focus()` already accepts options object. Can we add support for `suppressTooltip` in there?

src/app-layout/visual-refresh-toolbar/state/use-feature-notifications.tsx
```diff
@@ -0,0 +1,246 @@
...
+    const persistenceConfig = featureNotificationsData.persistenceConfig ?? DEFAULT_PERSISTENCE_CONFIG;
+    const featuresMap = featureNotificationsData.features.reduce((acc, feature) => {
+      return {
+        ...acc,
+        [feature.id]: feature.releaseDate,
+      };
+    }, {});
+    const filteredSeenFeaturesMap = filterOutdatedFeatures(seenFeatures);
+    const allFeaturesMap = { ...featuresMap, ...filteredSeenFeaturesMap };
+    const __persistFeatureNotifications:
```
pan-kot: nit: I would abstract the persistence related code into a dedicated function, so that the logic inside onOpenFeatureNotificationsDrawer is more clear:

```
const onOpenFeatureNotificationsDrawer = () => {
  if (!featureNotificationsData || allNotificationsSeen) {
    return;
  }
  const persistenceConfig = featureNotificationsData.persistenceConfig ?? DEFAULT_PERSISTENCE_CONFIG;
  persistFeatureNotificationsDismissState(persistenceConfig, featureNotificationsData).then((seenFeatures) => {
    setSeenFeatures(seenFeatures);
    setFeatureNotificationsData(data => (data ? { ...data, badge: false } : data));
  });
};
```

pages/feature-notifications/feature-prompt.page.tsx
georgylobko: addressed the issue with the tools panel. resolving this thread

src/app-layout/visual-refresh-toolbar/state/use-app-layout.tsx
georgylobko: addressed. resolving this thread

src/app-layout/visual-refresh-toolbar/state/use-feature-notifications.tsx
georgylobko: it's not enough as we filter outdated seen features from the storage here: https://github.com/cloudscape-design/components/blob/feat/feature-notitications/src/app-layout/visual-refresh-toolbar/state/use-feature-notifications.tsx#L224

src/app-layout/__tests__/runtime-feature-notifications.test.tsx
georgylobko: added a test for them

src/plugins/index.ts
georgylobko: exported them as 

```
featureNotifications = {
  registerFeatureNotifications,
  showFeaturePromptIfPossible,
  clearFeatureNotifications,
};
```

src/app-layout/visual-refresh-toolbar/state/use-feature-notifications.tsx
pan-kot: Oh, I did not realise that at first. Are there tests for that?

georgylobko: yes: https://github.com/cloudscape-design/components/blob/feat/feature-notitications/src/app-layout/__tests__/runtime-feature-notifications.test.tsx#L404

src/app-layout/visual-refresh-toolbar/state/use-feature-notifications.tsx
georgylobko: Feature prompt belongs to feature notifications and isn't really related to drawers - it's a domain-specific entity. Even if we add it to the drawers array, we'd still need to control when it opens and closes. I want
to keep that logic inside this hook since it's domain-specific, but then we'd have to forward those states to the triggers component, which leaks abstractions. So I wouldn't consider that a better solution

src/app-layout/visual-refresh-toolbar/drawer/feature-notifications-drawer-content.tsx
georgylobko: Because `secondaryContent` has different styles

pages/feature-notifications/feature-prompt.page.tsx
georgylobko: Introduced public and private interfaces for the API. The public one is used directly, and the private one goes through an intermediate proxy layer

src/internal/do-not-use/feature-prompt/interfaces.ts
georgylobko: introduced a type `export type DismissMethod = 'click-outside' | 'blur' | 'escape' | 'close-button';`

src/app-layout/__tests__/runtime-feature-notifications.test.tsx
georgylobko: added an additional scenario for it

src/app-layout/visual-refresh-toolbar/drawer/feature-notifications-drawer-content.tsx
georgylobko: It's a 2-way door decision. Let's collect use cases first. If we'll need to add it - we can always do so

src/app-layout/visual-refresh-toolbar/state/use-feature-notifications.tsx
georgylobko: added public and private interfaces

src/app-layout/__tests__/runtime-feature-notifications.test.tsx
georgylobko: no, it does not pass

src/app-layout/__tests__/runtime-feature-notifications.test.tsx
georgylobko: removed unnecessary delay and waitFor

src/app-layout/__tests__/runtime-feature-notifications.test.tsx
georgylobko: refactored the whole test file - removed all unnecessary delays and waitFors for

src/app-layout/__tests__/runtime-feature-notifications.test.tsx
georgylobko: refactored the whole test file - removed all unnecessary delays and waitFors for

src/app-layout/__tests__/runtime-feature-notifications.test.tsx
georgylobko: All these strings belong to the drawer. The feature prompt uses its own aria label for the dismiss button, so it's should not be confusing. Even if we need to introduce custom i18n strings for it, we can always rename them

pages/feature-notifications/feature-prompt.page.tsx
georgylobko: Discussed in DM - because the override uses the same feature id

src/app-layout/__tests__/runtime-feature-notifications.test.tsx
georgylobko: removed

src/app-layout/__tests__/runtime-feature-notifications.test.tsx
georgylobko: the purpose of this test is to provide a usage example, so I would keep it as is

src/app-layout/__tests__/runtime-feature-notifications.test.tsx
georgylobko: Yes, we need this test. However the dates are not important here, because we check only features ids - we check dates only when we persist values to filter out outdated values

src/app-layout/visual-refresh-toolbar/drawer/feature-notifications-drawer-content.tsx
georgylobko: yes, we need them for right paddings

src/app-layout/visual-refresh-toolbar/drawer/feature-notifications-drawer-content.tsx
georgylobko: No, I tried. The footer is sticky, which we don't need here

src/popover/body.tsx
georgylobko: yes

src/internal/do-not-use/feature-prompt/internal.tsx
georgylobko: fixed

src/drawer/use-sticky-footer.ts
georgylobko: otherwise it throws an error

src/app-layout/visual-refresh-toolbar/state/use-feature-notifications.tsx
georgylobko: not needed anymore as the drawer works along with tools

src/app-layout/visual-refresh-toolbar/state/use-feature-notifications.tsx
georgylobko: The suppress should be controlled not only when focus is set, but also when we show the feature popover. In that case, focus goes to the feature popover while the tooltip for its trigger is suppressed

src/internal/plugins/widget/index.ts
```diff
@@ -21,20 +29,44 @@ export function registerLeftDrawer(drawer: DrawerPayload) {
...
+export function registerFeatureNotifications<T>(payload: FeatureNotificationsPayload<T>) {
+  const message: RegisterFeatureNotificationsMessage<T> = {
+    type: 'registerFeatureNotifications',
+    payload,
+  };
+  pushInitialMessage(message);
+  getAppLayoutMessageHandler()?.(message as WidgetMessage<unknown>);
+}
+export function registerFeatureNotificationsPublic<T>(payload: FeatureNotificationsPayloadPublic<T>) {
+  registerFeatureNotifications(payload);
```
pan-kot: Let's use `getExternalProps` util here to ensure the `__` cannot be actually set on the public API.

srungta08 (approved):

PR #4194, opened 2026-01-20
feat: Make progress bar color tokens public

amanabiy:
### Description
Exposes two progress bar color tokens as public and themeable. Tokens:
- `colorBackgroundProgressBarContentDefault` to `colorBackgroundProgressBarValueDefault` (progress fill)
- `colorBackgroundProgressBarLayoutDefault` to `colorBackgroundProgressBarDefault` (progress track)

The following tokens have been renamed for consistency but not exposed as public tokens.
- `colorBackgroundProgressBarContentInFlash` to `colorBackgroundProgressBarValueInFlash` (in-flash fill)
- `colorBackgroundProgressBarLayoutInFlash` to `colorBackgroundProgressBarInFlash` (in-flash track)

Tested by:
- running through pipeline

Related links, issue #, if available: AWSUI-61668

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


style-dictionary/visual-refresh/metadata/colors.ts
```diff
@@ -207,6 +207,26 @@ const metadata: StyleDictionary.MetadataIndex = {
...
+    description: 'The default background color of the progress bar fill.',
+    public: true,
+    themeable: true,
+  },
+  colorBackgroundProgressBarContentInFlash: {
+    description: 'The background color of the progress bar fill when used in a flash component.',
+    public: true,
+    themeable: true,
+  },
+  colorBackgroundProgressBarLayoutDefault: {
```
mxschll: Should we call this `colorBackgroundProgressBarDefault` instead? And the "content" would be `colorBackgroundProgressBarValueDefault`? That matches the CSS selectors for progress bar a bit better: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/::-webkit-progress-value

mxschll: We discussed this naming already for the Style API (Quip: xHjOAgqjh0jY) and ended up with this convention:

```
export interface Style {
  progressBar?: { // Track
    backgroundColor?: string;
    borderRadius?: string;
    height?: string;
  };
  progressValue?: { // Colored bar within the track
    backgroundColor?: string;
    borderRadius?: string;
  };
  progressPercentage?: { // Text that shows percentage
    color?: string;
    fontSize?: string;
    fontWeight?: string;
  };
}
``` 

Maybe we should figure out how we can tie that together a bit nicer?

amanabiy: I see, thanks. I think this makes more sense as well.

jkuelz: I was going to leave a similar comment, suggesting to use "track" for the background color and "bar" for the filled part. But aligning with the style API makes even more sense. 

One thing I am wondering about -- are the `InFlash` tokens needed? It seems like this use case should be supported via visual contexts rather than additional tokens. We could just redefine the non-flash tokens within the [flashbar context](https://github.com/cloudscape-design/components/blob/main/style-dictionary/visual-refresh/contexts/flashbar.ts), like we recently did in [#4096](https://github.com/cloudscape-design/components/pull/4096) to support additional components. 

It might've been that the progress bar was created before visual contexts existed in the system, so it wasn't considered as an option. But it looks like we also made [this update](https://github.com/cloudscape-design/components/pull/2652) for progress bars in [warning flashbar context](https://github.com/cloudscape-design/components/blob/main/style-dictionary/visual-refresh/contexts/flashbar-warning.ts#L23-L26) since the initial release too. I would be in favor of doing the same for flashbar context and removing the `colorBackgroundProgressBarValueInFlash` and `colorBackgroundProgressBarInFlash` tokens all together.

amanabiy: Agreed. I was initially thinking that removing the InFlash tokens would be confusing, as builders might get confused when the styles they provided are not applied when the progress bar is in a flashbar. However, looking back at it now, I also think we should keep them private, because we are already applying a different color for the progress bar in the warning flashbar context.

mxschll (dismissed):

src/progress-bar/styles.scss
```diff
@@ -12,19 +12,19 @@ $progress-height: var(#{custom-props.$progressBarHeight}, 0.4 * styles.$base-siz
...
 $progress-background-color: var(
   #{custom-props.$progressBarBackgroundColor},
-  awsui.$color-background-progress-bar-layout-default
+  awsui.$color-background-progress-bar-default
 );
 $progress-background-color-in-flash: var(
   #{custom-props.$progressBarBackgroundColor},
-  awsui.$color-background-progress-bar-layout-in-flash
+  awsui.$color-background-progress-bar-in-flash
 );
```
jkuelz, line 20: We shouldn't need this at all if we just use `colorBackgroundProgressBarDefault` in the flashbar context

src/progress-bar/styles.scss
```diff
@@ -12,19 +12,19 @@ $progress-height: var(#{custom-props.$progressBarHeight}, 0.4 * styles.$base-siz
...
 $progress-value-background-color: var(
   #{custom-props.$progressValueBackgroundColor},
-  awsui.$color-background-progress-bar-content-default
+  awsui.$color-background-progress-bar-value-default
 );
 $progress-value-background-color-in-flash: var(
   #{custom-props.$progressValueBackgroundColor},
-  awsui.$color-background-progress-bar-content-in-flash
+  awsui.$color-background-progress-bar-value-in-flash
 );
```
jkuelz, line 28: Same here, we shouldn't need this at all if we just use `colorBackgroundProgressBarValueDefault` in the flashbar context

PR #4199, opened 2026-01-22
fix: Patch Screan reader issue in multiselect selected items

NathanZlion:
### Description

Fixed MultiSelect component accessibility issue where screen readers incorrectly announced "not-selected" for all options, including those that were actually selected. This prevented users relying on assistive technology from distinguishing between selected and unselected options.

- Added `aria-multiselectable="true"`__ to the listbox element when in multiselect mode
- Added both aria-selected attributes to multiselect options
- Retained selectedAriaLabel in live region announcements  AWSUI-61639

AWSUI-61639

### How has this been tested?
- Tested with VoiceOver (macOS) - announcements work correctly, no more "not selected" false announcements
- Done tests for MAC as well with the help of A11Y team, that's how we decided to keep the aria-live region selectedAriaLabel because NVDA and early version of JAWS (e.g. 2022, which is still pretty commonly used) have a bug around announcing the selected state.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/select/utils/use-announcement.ts
+++ b/src/select/utils/use-announcement.ts
@@ -58,6 +58,14 @@ export function useAnnouncement<Option extends OptionHolder>({
   }
 
   // Use default renderer with selected ARIA label if defined and relevant.
+  // Note: We intentionally keep selectedAriaLabel in the live region announcement even though
+  // we now use proper ARIA attributes (aria-selected, aria-checked, aria-multiselectable).
+  // This is because NVDA and older versions of JAWS (e.g., 2022, still commonly used) have bugs
+  // where they don't properly announce the selected state from ARIA attributes alone.
+  // For newer screen readers (modern JAWS, VoiceOver), this results in redundant "Selected"
+  // announcements, but this is preferable to the alternative where users hear "Selected"
+  // followed by "Not Selected" due to missing ARIA attribute support.
+  // See AWSUI-61639 for details.
   const selectedAnnouncement = announceSelected && selectedAriaLabel ? selectedAriaLabel : '';
   const defaultDescription = defaultOptionDescription({ option, parentGroup: group, highlightText });
   return [selectedAnnouncement, defaultDescription].filter(Boolean).join(' ');


src/select/parts/multiselect-item.tsx
```diff
@@ -159,6 +159,7 @@ const MultiSelectItem = (
   return (
     <SelectableItem
       disableContentStyling={!!renderResult}
+      ariaSelected={isParent && indeterminate ? undefined : Boolean(selected)}
```
SpyZzey, line 162: Having ariaSelected and ariaChecked is usually not the right approach, see:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-selected

Is there a reason to use both?

src/internal/components/selectable-item/interfaces.ts
```diff
@@ -28,7 +28,9 @@ export type SelectableItemProps = BaseComponentProps & {
   sticky?: boolean;
   afterHeader?: boolean;
   withScrollbar?: boolean;
-} & ({ ariaSelected?: boolean; ariaChecked?: never } | { ariaSelected?: never; ariaChecked?: boolean | 'mixed' });
+  ariaSelected?: boolean | never;
+  ariaChecked?: boolean | 'mixed' | never;
```
SpyZzey, line 32: See my comment below: https://github.com/cloudscape-design/components/pull/4199/changes#r2773565144

The type before discouraged using both.

src/select/parts/multiselect-item.tsx
NathanZlion: > [🔗](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-selected#option) Don't specify both aria-selected and aria-checked on option elements contained by the same listbox unless the meaning and purpose of aria-selected is different from the meaning and purpose of aria-checked in the user interface, the meaning and purpose of each state apparent, and the UI provides separate methods for controlling each state.

You're right in that it's discouraged to use both aria-selected and aria-checked, but some screen readers like NVDA and older versions of Jaws don't have support and based on some testing I did they announce checked and unchecked instead of selected so they give priority to that. It's a limitation because of screen readers not compatible with the best practice. We can remove the aria-checked when we have multiselect but with the risk of affecting older screen readers which currently announce "checked" and don't announce "selected" even when provided with `aria-multiselectable` and `aria-selected`. I've you sent more context in chat.

PR #4200, opened 2026-01-22
feat:  Add expandable step navigation for narrow viewports in Wizard Component

Who-is-PS:
### Description

Adds an expandable step navigation for the Wizard component that appears on narrow viewports (< 688px).

### Changes

- __New `WizardStepNavigationExpandable` component__: Displays step navigation inside an expandable section when the sidebar is hidden on narrow viewports
- __New `WizardStepList` component__: Shared step list rendering with grid-based layout matching the desktop sidebar visual style (circles, connecting lines, step labels)
- __Refactored shared utilities__: `getStepStatus`, `handleStepNavigation`, `canSkip` functions extracted for reuse between desktop and expandable navigation
- __Accessibility enhancements__: `<nav>` landmark wrapper with `headerAriaLabel` prop support for screen readers
- __Expandable section state management__: Navigation stays open during step transitions, with controlled state lifted to `InternalWizard`

https://github.com/user-attachments/assets/619eabe6-7777-45a3-8772-cb77ba7332ee

Related links, issue #, if available: `AWSUI-19175`

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/wizard/styles.scss
+++ b/src/wizard/styles.scss
@@ -274,7 +274,6 @@
 }
 
 /* Override fixed sidebar width for expandable navigation */
-/* stylelint-disable selector-max-type */
 .navigation.refresh.collapsed-steps-navigation {
   grid-column: unset;
   grid-row: unset;
@@ -284,7 +283,6 @@
     inline-size: 100%;
   }
 }
-/* stylelint-enable selector-max-type */
 
 .form-header-component {
   &-wrapper {


src/wizard/__integ__/wizard.test.ts
```diff
@@ -94,6 +94,29 @@ describe('Wizard keyboard navigation', () => {
...
+    useBrowser(async browser => {
+      const page = new WizardPageObject(browser);
+      // Set narrow viewport first using page object with object syntax
+      await page.setWindowSize({ width: 320, height: 600 });
+      await browser.url('/#/light/wizard/simple?visualRefresh=true');
+      await page.waitForVisible(wizardWrapper.findPrimaryButton().toSelector());
+
+      // Collapsed steps container should exist at narrow viewport
+      // Note: Using attribute selector because CSS class names are hashed in the build output
+      // Using isExisting() because ExpandableSection header has screenreader-only styling when collapsed
```
avinashbot: Hmm, that doesn't sound right to me. Are you sure you can't use `isDisplayed()` here? The expandable section header is always visible, so it doesn't use screenreader-only styling.

And instead of relying on `[class*=...]`, you can use the exact selectors directly from the style file. Check the unit tests for how we import it; it's find to do in the integ tests too.

```typescript
const collapsedSteps = wizardWrapper.find(styles['collapsed-steps']).findExpandableSection().toSelector();
```

Same for the sidebar navigation.

src/wizard/__tests__/wizard.test.tsx
```diff
@@ -1,18 +1,40 @@
...
+    removeListener: jest.fn(),
+    addEventListener: jest.fn(),
+    removeEventListener: jest.fn(),
+    dispatchEvent: jest.fn(),
+  }));
+  return () => {
+    window.matchMedia = originalMatchMedia;
+  };
+};
+
```
avinashbot: It's better to just mock useMobile entirely rather than the underlying implementation. It's also simpler to manage since you just need to change true/false.

```typescript
jest.mock('../../../lib/components/internal/hooks/use-mobile', () => ({
  ...jest.requireActual('../../../lib/components/internal/hooks/use-mobile'),
  useMobile: jest.fn().mockReturnValue(true),
}));
```

src/wizard/internal.tsx
```diff
@@ -53,10 +54,8 @@ export default function InternalWizard({
 }: InternalWizardProps) {
   const baseProps = getBaseProps(rest);
 
-  const [breakpoint, breakpointsRef] = useContainerBreakpoints(['xs']);
-  const ref = useMergeRefs(breakpointsRef, __internalRootRef);
-
-  const smallContainer = breakpoint === 'default';
+  const isMobile = useMobile();
```
avinashbot: This is a change in implementation; did we talk about doing this? `useMobile` is a global hook, so it returns true/false depending on the width of the viewport. `useContainerBreakpoints` is a container hook, so it changes depends on the width of the wizard itself.

For example, take a look at the [playground page for wizard](https://cloudscape.design/components/wizard/). It switches to mobile even before the page goes to mobile size because the wizard is in a small container.

src/wizard/wizard-step-list.tsx
```diff
@@ -0,0 +1,209 @@
...
+
+  const handleInteraction = () => handleStepNavigation(index, status, onStepClick, onSkipToClick);
+
+  return (
+    <li className={clsx(styles[`${state}`], styles['navigation-link-item'])}>
+      <hr />
+
+      <span className={clsx(styles.number, styles['navigation-link-label'])}>
+        {i18nStrings.stepNumberLabel?.(index + 1)}
+        {step.isOptional && <i id={optionalDescriptionId}>{` - ${i18nStrings.optional}`}</i>}
```
avinashbot, line 166: Nice!

src/wizard/internal.tsx
Who-is-PS: This was raised as a bug during bug bash. When resizing the viewport, the Wizard collapses to mobile steps first, then a few pixels later AppLayout switches to mobile - creating a jarring staggered effect.

This happens because Wizard uses `useContainerBreakpoints()` (triggers earlier as container shrinks) while AppLayout uses `useMobile()` (triggers at viewport breakpoint).

Which behavior should we prioritize?

- __(A) Keep container-based:__ Preserves narrow-container collapse, but Wizard collapses before AppLayout
- __(B) Switch to viewport-based:__ Syncs with AppLayout collapse timing, but loses narrow-container collapse

avinashbot: I don't think it's so important for the responsive logic to smoothly transition from big to small screen and never go up again — the idea behind container queries is so that it switches depending on both the screen size and if panels are toggled, which is more of a common action.

I would prefer not making the change in the PR and independently checking with a designer — there's a reason we have container queries in a number of components, especially when the available width on a page can change so frequently (especially now, with left panels, right panels, bottom panels, etc)

Who-is-PS: I've reverted the change back to `useContainerBreakpoints`. The container-responsive behavior is preserved - Wizard will collapse based on its own container width, not the viewport.

I'll mark this bug bash issue as "works as designed" and note that it needs designer input if we want to revisit the transition timing between AppLayout and Wizard.

src/wizard/__tests__/wizard.test.tsx
Who-is-PS: Done - now mocking useContainerBreakpoints directly instead of window.matchMedia.

src/wizard/__integ__/wizard.test.ts
Who-is-PS: Done - using isDisplayed() and importing styles directly from styles.selectors.js

src/wizard/styles.scss
```diff
@@ -261,13 +269,22 @@
...
-  font-weight: styles.$font-weight-bold;
-  padding-block-start: awsui.$space-scaled-xxs;
-  &-hidden {
-    display: none;
+  grid-column: 1 / span 2;
+  grid-row: 1;
+}
+
+/* Override fixed sidebar width for expandable navigation */
+/* stylelint-disable selector-max-type */
```
avinashbot: Looks like this selector was disabled unnecesarily - unless I'm wrong, you should be able to get rid of these comments without any issues. (h/t @timogasda)

Who-is-PS: Removed the unnecessary comments.

PR #4220, opened 2026-02-03
fix: Inline isFragment function to remove react-is dependency

amanabiy:
### Description
This PR inlines the isFragment function in the internal vendored react-keyed-flatten-children implementation to completely eliminate the react-is dependency. This builds on [PR #3997](https://github.com/cloudscape-design/components/pull/3997).

### Why?
React 19 renamed the element symbol from react.element to react.transitional.element to detect version mismatches in React and react-is ([React PR #28813](https://github.com/facebook/react/pull/28813)). This means we can't use react-is if we want to support both React 19 and maintain backward compatibility. Therefore, we need to replace the isFragment function with an internal implementation to make it work with both forward and backward compatibility.

[Here](https://codesandbox.io/p/sandbox/x7gmdq) is a minimal example of the problem using sandbox

### Changes
It makes it easy for review if you go through the commits
- The first commit reverses this [PR #3997](https://github.com/cloudscape-design/components/pull/3997)
- The second commit makes this isFragment inline removing react-is
- The third commit adds bigint support to handle bigint as a valid child type for React 19 compatibility ([React PR #24580](https://github.com/facebook/react/pull/24580)). TypeScript will error on bigint children with React <19 before reaching this conditional statement.
 
### How I tested it:
- Verified across multiple React versions from 16 to 19 using a simple application to test the flatten works as expected.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/vendor/react-keyed-flatten-children/index.ts
+++ b/src/internal/vendor/react-keyed-flatten-children/index.ts
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 // Vendored from react-keyed-flatten-children@3.2.0 (MIT License)
-// Original source: https://github.com/grrowl/react-keyed-flatten-children
+// Original source: https://github.com/grrowl/react-keyed-flatten-children/tree/120a286144c7fdc76b64daadaeae0df7d255b497
 
 import { Children, cloneElement, isValidElement, ReactElement, ReactNode } from 'react';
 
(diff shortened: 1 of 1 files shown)


amanabiy:
Closing this PR in favor of a different approach. After further discussion, we've decided to deprecate fragment flattening rather than maintain a custom isFragment implementation.

Follow-up PR: [Link](https://github.com/cloudscape-design/components/pull/4249)
Reference: PptYApl47Z5P

PR #4225, opened 2026-02-04
fix: Runtime content page function not found runtime exception

SpyZzey:
The initialCheck on our alert pages has a bug that throws a "t.match is not a function" runtime error. We are trying to call "match" on the AlertFlashContentInitialContext `context.content.props.children`. However, we are missing to check if it is a string. Apparently, it can sometimes be an array, therefore throwing an error. You can test this by opening the dev pages, then select the alert runtime-content and afterwards the alert permutations page.

-> As it only appears when switching from "runtime-content" to the "permutations"-page, I assume there is a deeper bug in our dev pages. But this is blocking me, so this should be treated as a hotfix until further research was done.
 
Screenshot 2026-02-04 at 18 04 43

### How has this been tested?
Tested locally.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/alert/runtime-content.page.tsx
+++ b/pages/alert/runtime-content.page.tsx
@@ -71,7 +71,7 @@ awsuiPlugins.alertContent.registerContentReplacer({
         typeof context.content === 'object' &&
         'props' in context.content &&
         typeof context.content.props.children === 'string' &&
-        context.content.props.children?.match('Access denied')
+        context.content.props.children.match('Access denied')
       )
     );
   },


pages/alert/runtime-content.page.tsx
```diff
@@ -70,6 +70,7 @@ awsuiPlugins.alertContent.registerContentReplacer({
         context.content &&
         typeof context.content === 'object' &&
         'props' in context.content &&
+        typeof context.content.props.children === 'string' &&
         context.content.props.children?.match('Access denied')
```
gethinwebster: `?` is no longer needed in this line now we have the new check
```suggestion
        context.content.props.children.match('Access denied')
```

PR #4237, opened 2026-02-10
chore: Allow dynamic deferred loading of internationalization

avinashbot:
### Description

Partial implementation of dynamically loaded internationalization. This lets us more readily embed the I18nProvider into performance-sensitive parts of the code (like the app layout toolbar) without worrying so much about added bundle size of the intl library, intl-messageformat (34.3 kB, 9.83 kB gzip) and the locale files (465kB, 68kB gzip). This also avoids dealing with any additional build-time considerations (intl-messageformat raises our bare minimum TypeScript version requirement, which we don't document publicly, though we do call out the raised minimum in our [documentation](https://cloudscape.design/get-started/for-developers/internationalization/#requirements)).

The added cost is one asynchronously triggered full-application rerender.

```
components/LocalI18nProvider  <- The application may provide this already (this is fine)
↳ components/AppLayout
  ↳ components/AppLayout/skeleton
    ↳ components/RemoteI18nProvider <-------+
      ↳ application code                    | 
      ↳ widgetized/AppLayout/toolbar        | --- Once messages and formatter are loaded,
      ↳ widgetized/AppLayout/drawers        |     the content is re-rendered.
    ↳ widgetized/loadRemoteI18nMessages() --+
```

"Dynamic loading" of the internationalization bundle isn't all that complicated, it's just a runtime import of `I18nFormatter` (which this PR refactors out) preloaded with the appropriate locale file (which are already exposed under `/messages/all.<locale>.js`).

---

Possible optimizations (either later in this PR or after release):
- ~~Wrapping the app rerender in a React transition: For React versions that support it, it avoids our rerender from blocking user actions and affecting web vitals.~~
- ~~Removing ICU parser from the remote bundle ([reference](https://formatjs.github.io/docs/guides/advanced-usage#react-intl-without-parser-40-smaller)): If we can safely guarantee that the remote provider only accepts and handles pre-parsed AST-only strings, we can remove the parser from the dynamically loaded bundle and reduce the dynamic bundle's file size by 26.34 kB (7.1 kB gzip). But this is a non-blocking network request, so it's not high priority.~~ See CR-253557644.

Related links, issue #, if available: AWSUI-61508

### How has this been tested?

Existing tests for locally (i.e. application) provided strings, additional tests for remote inclusion of logic will follow.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/i18n/providers/__tests__/remote-provider.test.tsx
+++ b/src/i18n/providers/__tests__/remote-provider.test.tsx
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import React, { useContext } from 'react';
-import { render, waitFor } from '@testing-library/react';
+import { render } from '@testing-library/react';
 
 import { I18nProvider } from '../../../../lib/components/i18n';
 import { FormatFunction, InternalI18nContext } from '../../../../lib/components/i18n/context';
@@ -41,9 +41,8 @@ describe('RemoteI18nProvider', () => {
       </RemoteI18nProvider>
     );
 
-    await waitFor(() => {
-      expect(getByTestId('locale')).toHaveTextContent('es');
-    });
+    await jest.runAllTimersAsync();
+    expect(getByTestId('locale')).toHaveTextContent('es');
     expect(loadFormatter).toHaveBeenCalledWith({ locale: 'es' });
   });
 
@@ -56,9 +55,8 @@ describe('RemoteI18nProvider', () => {
       </RemoteI18nProvider>
     );
 
-    await waitFor(() => {
-      expect(getByTestId('locale')).toHaveTextContent('en');
-    });
+    await jest.runAllTimersAsync();
+    expect(getByTestId('locale')).toHaveTextContent('en');
     expect(loadFormatter).toHaveBeenCalledWith({ locale: 'en' });
   });
 
@@ -71,9 +69,8 @@ describe('RemoteI18nProvider', () => {
       </RemoteI18nProvider>
     );
 
-    await waitFor(() => {
-      expect(loadFormatter).toHaveBeenCalled();
-    });
+    await jest.runAllTimersAsync();
+    expect(loadFormatter).toHaveBeenCalled();
     expect(getByTestId('locale')).toHaveTextContent('no-context');
   });
 
@@ -86,9 +83,8 @@ describe('RemoteI18nProvider', () => {
       </RemoteI18nProvider>
     );
 
-    await waitFor(() => {
-      expect(loadFormatter).toHaveBeenCalled();
-    });
+    await jest.runAllTimersAsync();
+    expect(loadFormatter).toHaveBeenCalled();
     expect(getByTestId('locale')).toHaveTextContent('no-context');
   });
 
--- a/src/app-layout/__tests__/toolbar.test.tsx
+++ b/src/app-layout/__tests__/toolbar.test.tsx
@@ -1,13 +1,26 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 import React from 'react';
+import { waitFor } from '@testing-library/react';
 
 import AppLayout from '../../../lib/components/app-layout';
+import { loadFormatter } from '../../../lib/components/app-layout/visual-refresh-toolbar/internal';
+import { I18nProvider } from '../../../lib/components/i18n';
+import { I18nFormatter } from '../../../lib/components/i18n/utils/i18n-formatter';
 import SplitPanel from '../../../lib/components/split-panel';
 import { describeEachAppLayout, manyDrawers, renderComponent } from './utils';
 
 // no-op function to suppress controllability warnings
 function noop() {}
+
+jest.mock('../../../lib/components/app-layout/visual-refresh-toolbar/internal', () => ({
+  ...jest.requireActual('../../../lib/components/app-layout/visual-refresh-toolbar/internal'),
+  loadFormatter: jest.fn(() => Promise.resolve(null)),
+}));
+
+afterEach(() => {
+  jest.clearAllMocks();
+});
 
 describe('toolbar mode only features', () => {
   describeEachAppLayout({ themes: ['refresh-toolbar'] }, ({ size }) => {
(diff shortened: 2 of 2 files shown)


src/app-layout/visual-refresh-toolbar/index.tsx
```diff
@@ -62,36 +64,65 @@ const AppLayoutStateProvider: React.FC<{
...
   (props, forwardRef) => {
-    const stateManager = useRef<StateManager>({ setState: undefined, hasToolbar: false, setToolbar: undefined });
-    const { __forceDeduplicationType: forceDeduplicationType, __embeddedViewMode: embeddedViewMode } = props as any;
+    const stateManagerRef = useRef<StateManager>({ setState: undefined, hasToolbar: false, setToolbar: undefined });
 
     return (
-      <>
-        <AppLayoutStateProvider
-          forceDeduplicationType={forceDeduplicationType}
+      <RemoteI18nProvider loadFormatter={loadFormatter}>
```
georgylobko, line 71: Why do you need to wrap the whole thing with this provider? We can only wrap `SkeletonLayout` and call `useAriaLabels` inside

src/i18n/providers/__tests__/remote-provider.test.tsx
```diff
@@ -0,0 +1,109 @@
...
+function TestConsumer() {
+  const context = useContext(InternalI18nContext);
+  return <div data-testid="locale">{context?.locale || 'no-context'}</div>;
+}
+
+describe('RemoteI18nProvider', () => {
+  it('loads the formatter and provides the context to the children', async () => {
+    document.documentElement.lang = 'es';
+
+    const loadFormatter = jest.fn().mockResolvedValue(createMockFormatter());
```
georgylobko, line 36: Why do you provide this function as a prop? Could it be an internal value in the provider?

src/i18n/providers/__tests__/remote-provider.test.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+  return { format };
+};
+
+function TestConsumer() {
+  const context = useContext(InternalI18nContext);
+  return <div data-testid="locale">{context?.locale || 'no-context'}</div>;
+}
+
+describe('RemoteI18nProvider', () => {
```
georgylobko, line 32: Let's add more realistic scenarios: 
- what happens with provider's children i18n strings until loadFormatter is loaded?
- let's test it with our AppLayout component, especially with dynamically injected content
- let's make sure it works correctly if consumers wrap the content area with our i18nprovider

src/i18n/providers/__tests__/remote-provider.test.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+    const loadFormatter = jest.fn().mockResolvedValue(createMockFormatter());
+
+    const { getByTestId } = render(
+      <RemoteI18nProvider loadFormatter={loadFormatter}>
+        <TestConsumer />
+      </RemoteI18nProvider>
+    );
+
+    await waitFor(() => {
```
georgylobko: [non-blocking] Let's make it more explicit: `await jest.runAllTimersAsync()` waits until the async function is resolved

src/i18n/providers/remote-provider.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+    staticLoadFormatter({ locale })
+      .then(formatter => {
+        if (formatter) {
+          setFormatFunction(() => formatter.format.bind(formatter));
+        }
+        // If formatter isn't available, do nothing.
+      })
+      .catch(() => {
+        // Do nothing. Failure in fetching the formatter should not be fatal.
```
georgylobko, line 56: Do we want to send an operational metric here? same as here: https://github.com/cloudscape-design/components/blob/main/src/app-layout/visual-refresh-toolbar/state/use-app-layout.tsx#L227

src/i18n/utils/i18n-formatter.ts
```diff
@@ -0,0 +1,95 @@
...
+    };
+  };
+}
+
+/**
+ * A stateful container for formatting internal strings. Caches formatters
+ * where possible; a new instance must be created if locale or messages may
+ * have changed.
+ */
+export class I18nFormatter {
```
georgylobko, line 31: Should we name it more explicitly like `RemoveI18nFormatter` and put to the same folder as other widget parts?

avinashbot: It's the common formatting logic — it's imported by both the local provider and the widget (which passes it to remote provider). So it's not specific to either scenario.

src/app-layout/visual-refresh-toolbar/index.tsx
avinashbot: It seemed cleaner and safer this way. I wasn't sure if the weird "hidden copy of breadcrumbs" thing should also be wrapped by the provider or if it's safe for its strings to not match the skeleton's strings, and if the deduplication breadcrumbs are included, the wrapped component ends up requiring passing along a lot of internal props and state tracking, which seemed unmaintainable to me.

If the breadcrumbs are 100% safe to exclude, I can go for this approach.

src/i18n/providers/__tests__/remote-provider.test.tsx
avinashbot: Not sure what you mean, could you expand? The `loadFormatter` function is a widget export — locally it just returns null; in production, it would return a promise with a formatter.

src/i18n/providers/remote-provider.tsx
avinashbot: Oh absolutely, thanks for the callout; was wishing there was a way we could send a metric with included data. Let me see what the limits are, maybe I can include the error message.

src/i18n/providers/__tests__/remote-provider.test.tsx
georgylobko: I mean, why do you provide this function as a prop for the `RemoteI18nProvider`? It could be a part of its implementation

avinashbot: It can, but this is better for bundle size (because intl-messageformat is not in the bundle):
- `RemoteI18nProvider` is loaded by the application
- `loadFormatter` is part of the loader (in the application but not this package)
- `loadFormatter` then fetches `I18nFormatter` from the widget

src/i18n/providers/__tests__/remote-provider.test.tsx
avinashbot: Added some more tests!

georgylobko (dismissed):

PR #4238, opened 2026-02-11
fix: UAP buttons misplaced when drawer opens via keyboard

Who-is-PS:
### Description

UAP buttons appeared in wrong position when opening drawer via keyboard navigation. Position was calculated during CSS transition before it settled.

Changed to continuous position monitoring while buttons are visible, matching the existing `PortalOverlay` pattern.

Related links, issue #, if available: `AWSUI-61721`

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/components/drag-handle-wrapper/__tests__/drag-handle-wrapper.test.tsx
+++ b/src/internal/components/drag-handle-wrapper/__tests__/drag-handle-wrapper.test.tsx
@@ -33,8 +33,14 @@ const position = (inlineStart: number, blockStart: number) => ({
 });
 
 // Store RAF callbacks globally so flush() can be used in tests
-let rafCallbacks: Set<FrameRequestCallback>;
-let flushAnimationFrames: () => void;
+const rafCallbacks = new Set<FrameRequestCallback>();
+const flushAnimationFrames = () => {
+  act(() => {
+    const callbacks = [...rafCallbacks]; // Snapshot before clearing to preserve re-registered callbacks
+    rafCallbacks.clear();
+    callbacks.forEach(cb => cb(performance.now()));
+  });
+};
 
 beforeAll(() => {
   (window as any).PointerEvent ??= PointerEventMock;
@@ -47,17 +53,10 @@ beforeEach(() => {
   Object.defineProperty(window, 'innerWidth', { value: viewport.width, writable: true });
   Object.defineProperty(window, 'innerHeight', { value: viewport.height, writable: true });
 
-  // Mock requestAnimationFrame to collect callbacks for explicit flushing
-  // This gives tests control over when position calculations run
-  rafCallbacks = new Set<FrameRequestCallback>();
-  flushAnimationFrames = () => {
-    act(() => {
-      const callbacks = [...rafCallbacks]; // Snapshot before clearing
-      rafCallbacks.clear();
-      callbacks.forEach(cb => cb(performance.now()));
-    });
-  };
+  // Clear any leftover callbacks from previous tests
+  rafCallbacks.clear();
 
+  // Mock requestAnimationFrame to collect callbacks for explicit flushing
   jest.spyOn(window, 'requestAnimationFrame').mockImplementation((cb: FrameRequestCallback) => {
     rafCallbacks.add(cb);
     return rafCallbacks.size;


src/internal/components/drag-handle-wrapper/index.tsx
```diff
@@ -184,8 +184,20 @@ export default function DragHandleWrapper({
...
-    if (showButtons && dragHandleRef.current) {
+  // Continuously monitor position while buttons are shown to handle CSS transitions/animations.
+  // The position needs to be recalculated as the element may animate into its final position.
+  useEffect(() => {
+    if (!showButtons || !dragHandleRef.current) {
+      setForcedPosition(null);
+      return;
+    }
+
+    let cleanedUp = false;
```
georgylobko: Does this code ever run? It's in the same function scope, so once the useEffect cleans up, this variable is already out of scope

jperals: My understanding is that since `checkPosition` is called with `requestAnimationFrame`, there is a non-zero chance that the component unmounted in-between —that is what `cleanedUp` tracks.

Who-is-PS: The cleanedUp variable is captured by closure in both the checkPosition function and the cleanup function. When cleanup runs and sets cleanedUp = true, the pending rAF callback still has access to it via closure and will see the updated value, causing it to exit.

src/internal/components/drag-handle-wrapper/index.tsx
```diff
@@ -199,7 +211,14 @@ export default function DragHandleWrapper({
...
         setForcedPosition(null);
       }
-    }
+      requestAnimationFrame(checkPosition);
+    };
+
+    requestAnimationFrame(checkPosition);
+
+    return () => {
+      cleanedUp = true;
```
pan-kot: why do we introduce cleanedUp flag instead of cancelling the requested animation frames?

src/internal/components/drag-handle-wrapper/__tests__/drag-handle-wrapper.test.tsx
```diff
@@ -42,6 +42,18 @@ beforeEach(() => {
   jest.mocked(getLogicalBoundingClientRect).mockReturnValue(position(viewport.width / 2, viewport.height / 2));
   Object.defineProperty(window, 'innerWidth', { value: viewport.width, writable: true });
   Object.defineProperty(window, 'innerHeight', { value: viewport.height, writable: true });
+
+  // Mock requestAnimationFrame to execute the callback once synchronously
+  // but prevent infinite recursion from the continuous position monitoring loop
+  let rafExecuting = false;
+  jest.spyOn(window, 'requestAnimationFrame').mockImplementation((cb: FrameRequestCallback) => {
```
pan-kot, line 60: nit: you can instead store the cb to a global var so that you can execute it on demand:

```
let afCb = () => {};
jest.spyOn(window, 'requestAnimationFrame').mockImplementation((cb) => {
  afCb = cb;
});

test("...", () => {
  // render drag handle
  // execute afCb and ensure the position is reset
})
```

src/internal/components/drag-handle-wrapper/index.tsx
Who-is-PS: Valid point, I will update it with cancelAnimationFrame. It is cleaner

src/internal/components/drag-handle-wrapper/__tests__/drag-handle-wrapper.test.tsx
Who-is-PS: This would require refactoring all the forced position tests to call afCb() manually. Do you have have a strong opinion on this ?

pan-kot: I think that's fair. After all, in the new impl the forces position is only set in the animation frame.

Oh, I think that my above suggestion does not handle the recursive nature of it. It can probably be rather something like this:
```
const registeredCallbacks = new Set();
const flush = () => [...registeredCallbacks].map(cb => cb());
jest.spyOn(window, 'requestAnimationFrame').mockImplementation((cb) => {
  registeredCallbacks.add(cb);
});

test("...", () => {
  // render drag handle
  // expect no forced position
})

test("...", () => {
  // render drag handle
  // flush
  // expect forced position
})

test("...", () => {
  // render drag handle
  // flush
  // expect forced position
  // flush
  // expect no longer forced position
})
```

Who-is-PS: Done. I've implemented the Set-based approach with explicit flushing.

src/internal/components/drag-handle-wrapper/__tests__/drag-handle-wrapper.test.tsx
```diff
@@ -32,6 +32,10 @@ const position = (inlineStart: number, blockStart: number) => ({
   insetBlockEnd: blockStart + size.blockSize,
 });
 
+// Store RAF callbacks globally so flush() can be used in tests
+let rafCallbacks: Set<FrameRequestCallback>;
```
jperals: Nit: could these be consts?

src/internal/components/drag-handle-wrapper/__tests__/drag-handle-wrapper.test.tsx
```diff
@@ -42,6 +46,23 @@ beforeEach(() => {
   jest.mocked(getLogicalBoundingClientRect).mockReturnValue(position(viewport.width / 2, viewport.height / 2));
   Object.defineProperty(window, 'innerWidth', { value: viewport.width, writable: true });
   Object.defineProperty(window, 'innerHeight', { value: viewport.height, writable: true });
+
+  // Mock requestAnimationFrame to collect callbacks for explicit flushing
+  // This gives tests control over when position calculations run
+  rafCallbacks = new Set<FrameRequestCallback>();
+  flushAnimationFrames = () => {
+    act(() => {
+      const callbacks = [...rafCallbacks]; // Snapshot before clearing
```
jperals: Why assign to a snapshot instead of just calling the callbacks first and clearing `rafCallbacks` afterwards?

jperals:
> I see that a mock for requestAnimationFrame is added - but where is the test to ensure that the position is recomputed on every animation frame?

I see that a unit test for this has now been added, I'm OK with it but think that visual regression test is better here —testing the actual result and not depending so much on implementation details.

src/internal/components/drag-handle-wrapper/__tests__/drag-handle-wrapper.test.tsx
Who-is-PS: Because callbacks re-register themselves (animation loop). If we clear after executing, the newly registered callback is lost.

src/internal/components/drag-handle-wrapper/__tests__/drag-handle-wrapper.test.tsx
Who-is-PS: Refactored it to const.

PR #4241, opened 2026-02-11
fix: Respect minWidth for dynamically added columns in resizable mode

### LINKED ISSUE #4236: [Bug]: resizableColumns: dynamically added columns ignore minWidth, fall back to DEFAULT_COLUMN_WIDTH
### Browser

Chrome

### Package version

v3.0.0 (@amzn/awsui-components-console)

### React version

v19.2.3

### Description

When using `resizableColumns` on a `<Table>`, columns that are dynamically added to `columnDefinitions` (e.g. swapping visible columns based on a filter/toggle) do not respect their `minWidth`. They are instead assigned `DEFAULT_COLUMN_WIDTH` (120px).

The issue is in [`use-column-widths.tsx` L171](https://github.com/cloudscape-design/components/blob/9a1b7c5bf8e597df72f40a152f58bee9d09d2ac8/src/table/use-column-widths.tsx#L171):

```ts
newColumnWidths.set(column.id, column.width || DEFAULT_COLUMN_WIDTH);
```

This only checks `column.width` and ignores `column.minWidth`. A column defined with `{ minWidth: 250 }` but no explicit `width` will get 120px when dynamically added.

The initial render handles this correctly in `readWidths` ([L28-L41](https://github.com/cloudscape-design/components/blob/9a1b7c5bf8e597df72f40a152f58bee9d09d2ac8/src/table/use-column-widths.tsx#L28-L41)) by doing `Math.max(width, minWidth)`, but the dynamic column path does not.

### Expected behavior

Dynamically added columns should respect `minWidth`, consistent with the initial render behavior. The fix would be something like:

```ts
const minWidth = column.minWidth || column.width || DEFAULT_COLUMN_WIDTH;
newColumnWidths.set(column.id, Math.max(column.width || DEFAULT_COLUMN_WIDTH, minWidth));
```

### Workaround

Set an explicit `width` equal to `minWidth` on affected columns:

```ts
{
 id: 'myColumn',
 header: 'My Column',
 width: 250,    // workaround: must set width explicitly
 minWidth: 250,
 cell: (item) => item.value,
}
```

### Source code

[`use-column-widths.tsx` L171](https://github.com/cloudscape-design/components/blob/9a1b7c5bf8e597df72f40a152f58bee9d09d2ac8/src/table/use-column-widths.tsx#L171):

```ts
newColumnWidths.set(column.id, column.width || DEFAULT_COLUMN_WIDTH);
```

### Reproduction

_No response_

### Code of Conduct

kaigritun:
## Description

Fixes #4236

When columns are dynamically added to a table with `resizableColumns` enabled, the `minWidth` property was being ignored. Instead, columns defaulted to `DEFAULT_COLUMN_WIDTH` (120px) even when they had a `minWidth` property set.

### Problem

In `use-column-widths.tsx`, the code path for dynamically added columns (line ~171) only checked `column.width`:

```ts
newColumnWidths.set(column.id, (column.width as number) || DEFAULT_COLUMN_WIDTH);
```

This is inconsistent with the initial render behavior in `readWidths()` (lines 27-32), which correctly handles `minWidth`:

```ts
let width = (column.width as number) || 0;
const minWidth = (column.minWidth as number) || width || DEFAULT_COLUMN_WIDTH;
result.set(column.id, Math.max(width, minWidth));
```

### Solution

Apply the same logic to dynamically added columns, ensuring `minWidth` is respected consistently:

```ts
const width = (column.width as number) || 0;
const minWidth = (column.minWidth as number) || width || DEFAULT_COLUMN_WIDTH;
newColumnWidths.set(column.id, Math.max(width, minWidth));
```

### Testing

Added a new test case `should respect minWidth for dynamically added columns without explicit width` that:
1. Renders a table with only one visible column
2. Dynamically adds columns that have `minWidth` but no explicit `width`
3. Verifies the columns respect their `minWidth` values instead of falling back to `DEFAULT_COLUMN_WIDTH`

All existing tests continue to pass.

### DIFF
--- a/src/table/use-column-widths.tsx
+++ b/src/table/use-column-widths.tsx
@@ -168,7 +168,10 @@ export function ColumnWidthsProvider({ visibleColumns, resizableColumns, contain
         const column = visibleColumns[index];
         if (!columnWidths?.get(column.id) && lastVisible.indexOf(column.id) === -1) {
           updated = true;
-          newColumnWidths.set(column.id, (column.width as number) || DEFAULT_COLUMN_WIDTH);
+          // Handle minWidth consistently with readWidths() for initial render
+          const width = (column.width as number) || 0;
+          const minWidth = (column.minWidth as number) || width || DEFAULT_COLUMN_WIDTH;
+          newColumnWidths.set(column.id, Math.max(width, minWidth));
         }
       }
       if (updated) {
--- a/src/table/__tests__/columns-width.test.tsx
+++ b/src/table/__tests__/columns-width.test.tsx
@@ -137,6 +137,33 @@ test('should use the fallback value for columns without explicit width', () => {
   ]);
 });
 
+test('should respect minWidth for dynamically added columns without explicit width', () => {
+  const columns: TableProps.ColumnDefinition<Item>[] = [
+    { id: 'id', header: '', cell: item => item.id, width: 100 },
+    { id: 'name', header: '', cell: () => '-', minWidth: 250 },
+    { id: 'description', header: '', cell: () => '-', minWidth: 300 },
+  ];
+  const { wrapper, rerender } = renderTable(
+    <Table columnDefinitions={columns} visibleColumns={['id']} items={defaultItems} resizableColumns={true} />
+  );
+  expect(wrapper.findColumnHeaders().map(column => column.getElement().style.width)).toEqual(['100px']);
+  // Dynamically add columns that have minWidth but no explicit width
+  rerender(
+    <Table
+      columnDefinitions={columns}
+      visibleColumns={['id', 'name', 'description']}
+      items={defaultItems}
+      resizableColumns={true}
+    />
+  );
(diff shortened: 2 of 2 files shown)


src/table/use-column-widths.tsx
```diff
@@ -168,7 +168,10 @@ export function ColumnWidthsProvider({ visibleColumns, resizableColumns, contain
         const column = visibleColumns[index];
         if (!columnWidths?.get(column.id) && lastVisible.indexOf(column.id) === -1) {
           updated = true;
-          newColumnWidths.set(column.id, (column.width as number) || DEFAULT_COLUMN_WIDTH);
+          // Handle minWidth consistently with readWidths() for initial render
+          const width = (column.width as number) || 0;
+          const minWidth = (column.minWidth as number) || width || DEFAULT_COLUMN_WIDTH;
+          newColumnWidths.set(column.id, Math.max(width, minWidth));
```
SpyZzey, line 174: If width is not specified, but column.minWidth is, this should probably fallback to DEFAULT_COLUMN_WIDTH. The proposed changes will set the width to column.minWidth.

PR #4249, opened 2026-02-13
feat: support React 19 bigint, and deprecate fragment flattening

amanabiy:
### Description
Cloudscape components (ColumnLayout, Grid, and SpaceBetween) currently flatten React Fragments using react-keyed-flatten-children (v2.2.1), which depends on react-is (^18.2.0). React 19 renamed internal type elements, breaking fragment detection and flattening when using older versions of react-is.

### Changes
- React 16-18: Uses react-keyed-flatten-children for backward compatibility (fragments still flattened)
- React 19+: Uses Children.toArray() directly (fragments NOT flattened, BigInt support added)

Tested by
- using our column-layout in different versions of react from 16 - 19 application locally.
- A simple [sandbox test](https://codesandbox.io/p/devbox/dawn-waterfall-forked-kdyrvt?workspaceId=ws_8KrQCeJzJf8bLXhxVgzDPt) showing toArray behaves as we expect in react 19

Related links, issue #, if available: PptYApl47Z5P

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/internal/utils/flatten-children.ts
```diff
@@ -0,0 +1,17 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import React, { ReactNode } from 'react';
+import flattenChildrenLegacy from 'react-keyed-flatten-children';
+
+export function flattenChildren(children: ReactNode): ReactNode[] {
+  const majorVersion = parseInt(React.version.split('.')[0], 10);
```
jperals: Since `React.version` is undocumented, should we make this line more failproof? And if checking the version fails, then fall back to the new behavior (i.e, not handling fragments, same as version >= 19)

src/space-between/internal.tsx
```diff
@@ -52,10 +52,11 @@ const InternalSpaceBetween = forwardRef(
         ref={mergedRef}
       >
         {flattenedChildren.map(child => {
-          const key = typeof child === 'object' ? child.key : undefined;
+          // If this react child is a primitive value, the key will be undefined
+          const key = (child as Record<'key', unknown>).key;
```
jperals: Is there a possible case where this will throw an error? Let's say some conditional rendering logic where a child can be rendered as `undefined` or a number.

amanabiy: if the child is a number, bigint (in react 19), or string the key will be undefined. so we will be assigning undefined key. But it shouldn't throw an error.

src/internal/utils/flatten-children.ts
amanabiy: Good point, I will do it like that.

src/space-between/internal.tsx
jperals: And what about `undefined`?

Before the changes, the code was safe because `typeof undefined` is not `'object'`, therefore we render `undefined` according to the ternary operator, but after the changes, we will try to access a property of `undefined`, which throws an error.

amanabiy: You're right that accessing a property on undefined would throw an error. However, this is safe because both Children.toArray() and react-keyed-flatten-children filter out undefined, null, and boolean values.

Looking at the type signature of toArray it returns the following: `toArray(children: ReactNode | ReactNode[]): Array<Exclude<ReactNode, boolean | null | undefined>>;` The return type explicitly excludes undefined, null, and boolean. React's Children.toArray() implementation ([test example](https://github.com/facebook/react/blob/6066c782fe06b65e98a410741d994f620321fe11/packages/react/src/__tests__/ReactChildren-test.js#L830)). Under the hood, react-keyed-flatten-children also uses Children.toArray so they are filtered automatically, in both cases.

Here is another place where we also use similar [code](https://github.com/cloudscape-design/components/blob/38a9eb8c08fe59efdc201b1f656f2f260a03aee2/src/grid/internal.tsx#L85) (line 85). That said, if you'd prefer more safe approach, we could add a guard and/or I can change the other implementation as well. What do you think?

amanabiy: I believe it is safe to update it to handle if it is undefined, or null. I will update the ones in here, and the others as well. Thank you.

jperals: OK, then I think both approaches should be fine. Thanks for the deep dive.

PR #4243, opened 2026-02-11
feat: add position prop to modal

amanabiy:
### Description
This PR adds a prop `position` to the modal to allow builders choose position of center or top. This is a follow up to [this pull request](https://github.com/cloudscape-design/components/pull/4240) which adds x-large and xx-large.

Related links, issue #, if available: gfq0AFhyrah9

### How has this been tested?

<details>
   <summary>Review checklist</summary>

_The following items are to be evaluated by the author(s) and the reviewer(s)._

#### Correctness

- _Changes include appropriate documentation updates._
- _Changes are backward-compatible if not indicated, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#public-apis)._
- _Changes do not include unsupported browser features, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#browsers-support)._
- _Changes were manually tested for accessibility, see [accessibility guidelines](https://cloudscape.design/foundation/core-principles/accessibility/)._

#### Security

- _If the code handles URLs: all URLs are validated through [the `checkSafeUrl` function](https://github.com/cloudscape-design/components/blob/main/src/internal/utils/check-safe-url.ts)._

#### Testing

- _Changes are covered with new/existing unit tests?_
- _Changes are covered with new/existing integration tests?_
</details>

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #4252, opened 2026-02-13
feat: Add custom width and height props to modal

amanabiy:
### Description
This PR extends the Modal component with custom dimension support by adding width and height props, building on the position prop introduced in the [previous PR](https://github.com/cloudscape-design/components/pull/4243).

**Note:** The base branch is `dev-v3-amanabiy-modal-top-position` to isolate changes from the position prop PR for easier review.

New Features:
- **width prop**: Allows specifying custom modal width (minimum 320px). Takes precedence over the size prop when provided.
- **height prop**: Enables custom modal height with automatic content scrolling when content exceeds the specified height.

The base is made to `dev-v3-amanabiy-modal-top-position` so that you can review the main changes only related to this PR, and not the previous prs which adds position prop and different sizes.

Behavior:
- Both props respect viewport constraints - if specified dimensions exceed available space, the modal uses maximum available space
- When custom height is set and content overflows, the content is scrollable
- Minimum width constraint ensures that the minimum width builders can set is equal to the smallest `size`
- Minimum height constraint ensures content area is at least 60px for accessibility
- Proper ARIA attributes (role="region", aria-label) are applied to scrollable content regions for screen reader support

Related links, issue #, if available: gfq0AFhyrah9

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/modal/__tests__/modal.test.tsx
+++ b/src/modal/__tests__/modal.test.tsx
@@ -3,8 +3,6 @@
 import * as React from 'react';
 import { act, fireEvent, render } from '@testing-library/react';
 
-import { warnOnce } from '@cloudscape-design/component-toolkit/internal';
-
 import Autosuggest from '../../../lib/components/autosuggest';
 import Button from '../../../lib/components/button/index.js';
 import ButtonDropdown from '../../../lib/components/button-dropdown';
@@ -22,15 +20,6 @@ import { PerformanceMetrics } from '../../internal/analytics';
 import { KeyCode } from '../../internal/keycode';
 
 import styles from '../../../lib/components/modal/styles.css.js';
-
-jest.mock('@cloudscape-design/component-toolkit/internal', () => ({
-  ...jest.requireActual('@cloudscape-design/component-toolkit/internal'),
-  warnOnce: jest.fn(),
-}));
-
-afterEach(() => {
-  (warnOnce as jest.Mock).mockReset();
-});
 class ModalInternalWrapper extends ModalWrapper {
   findDialog(): ElementWrapper {
     return this.findByClassName(styles.dialog)!;
@@ -126,46 +115,6 @@ describe('Modal component', () => {
       (['small', 'medium', 'large', 'x-large', 'xx-large', 'max'] as ModalProps.Size[]).forEach(size => {
         const wrapper = renderModal({ size });
         expect(wrapper.findDialog().getElement()).toHaveClass(styles[size]);
-      });
-    });
-
-    it('width overrides size', () => {
-      const wrapper = renderModal({ size: 'large', width: 500 });
-      expect(wrapper.findDialog().getElement()).not.toHaveClass(styles.large);
-      expect(wrapper.findDialog().getElement()).toHaveClass(styles['custom-width']);
-    });
-
-    it('applies custom width class', () => {
-      const wrapper = renderModal({ width: 600 });
-      expect(wrapper.findDialog().getElement()).toHaveClass(styles['custom-width']);
-    });
-
-    it('applies custom height class', () => {
-      const wrapper = renderModal({ height: 400, footer: <div>Footer</div> });
-      expect(wrapper.findDialog().getElement()).toHaveClass(styles['custom-height']);
-      expect(wrapper.findFooter()!.getElement()).toHaveClass(styles['custom-height']);
-      expect(wrapper.findByClassName(styles.container)!.getElement()).toHaveClass(styles['custom-height-container']);
-      expect(wrapper.findContent().getElement()).toHaveClass(styles['custom-height-content']);
-    });
-
-    it('applies custom height class when footer is absent', () => {
-      const wrapper = renderModal({ height: 400 });
-      expect(wrapper.findDialog().getElement()).toHaveClass(styles['custom-height']);
-      expect(wrapper.findByClassName(styles.container)!.getElement()).toHaveClass(styles['custom-height-container']);
-      expect(wrapper.findContent().getElement()).toHaveClass(styles['custom-height-content']);
-    });
-  });
-
-  describe('position property', () => {
-    it('defaults to center position', () => {
-      const wrapper = renderModal();
-      expect(wrapper.findFocusLock().getElement()).toHaveClass(styles['position-center']);
-    });
-
-    it('displays correct position', () => {
-      (['center', 'top'] as ModalProps.Position[]).forEach(position => {
-        const wrapper = renderModal({ position });
-        expect(wrapper.findFocusLock().getElement()).toHaveClass(styles[`position-${position}`]);
       });
     });
   });


pages/modal/custom-dimensions.page.tsx
```diff
@@ -0,0 +1,60 @@
...
+}
+
+export default function () {
+  const urlParams = getUrlParams();
+  const [visible, setVisible] = useState(false);
+  const [width, setWidth] = useState(urlParams.width);
+  const [height, setHeight] = useState(urlParams.height);
+
+  return (
+    <SimplePage title="Modal with custom dimensions">
```
jperals: You can use the `settings` prop of the SimplePage to separate the settings form from the page content.

If the content is to be used in visual regression, then you can either use the `screenshotArea` prop or wrap the content in a ScreenshotArea.

pages/modal/custom-dimensions.page.tsx
```diff
@@ -0,0 +1,60 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import React, { useState } from 'react';
+
+import { Button, FormField, Input, Modal, SpaceBetween } from '~components';
+
+import { SimplePage } from '../app/templates';
+
+function getUrlParams() {
```
jperals: You can use the existing AppContexProvider which provides `urlParams` and `setUrlParams`. This way the parameters get synced directly with the URL instead of internal state. This can be practical for visual regression or integration testing to load the page in a desired state directly without interacting with the page (or just for direct links to a specific state).

src/modal/interfaces.ts
```diff
@@ -24,12 +24,37 @@ export interface BaseModalProps {
...
+   * The minimum height is constrained to ensure the content area is at least 60px for accessibility.
+   */
+  height?: number;
+
   /**
    * Sets the width of the modal. `max` uses variable width up to the
-   * largest size allowed by the design guidelines. Other sizes
-   * (`small`/`medium`/`large`) have fixed widths.
+   * largest size allowed by the design guidelines. Other sizes:
+   * `small` (320px), `medium` (600px), `large` (820px), `x-large` (1024px), `xx-large` (1280px).
```
jperals, line 45: The last part of the sentence ("have fixed widths") is gone now. I think the most conservative option without losing any information would be to add it to the beginning:

> Other sizes have fixed widths: `small` (320px), `medium` (600px), `large` (820px), `x-large` (1024px), `xx-large` (1280px),

src/modal/interfaces.ts
```diff
@@ -24,12 +24,37 @@ export interface BaseModalProps {
...
    */
   size?: ModalProps.Size;
+  /**
+   * Controls the vertical positioning of the modal.
+   *
+   * - `center` (default) - Modal is vertically centered in viewport and re-centers
```
jperals: I'd suggest to add the article "the" in a few places: "_The_ modal", "_the_ viewport"

src/modal/interfaces.ts
```diff
@@ -24,12 +24,37 @@ export interface BaseModalProps {
...
+   *   when content height changes. Use for dialogs with static, predictable content.
+   *
+   * - `top` - Modal anchors at fixed distance and grows downward
```
jperals: Also here I'd add "the" to "_The_ modal" and probably also to "_the_ content" (the latter is a bit more ambiguous though, also works without it I'd say)

src/modal/interfaces.ts
```diff
@@ -24,12 +24,37 @@ export interface BaseModalProps {
...
+   * Specifies the width of the modal. When provided, takes precedence over the `size` property.
+   * If the specified width exceeds available viewport space, the modal will use the maximum available space.
+   * The minimum width is 320px (equivalent to the `small` size).
+   */
+  width?: number;
+
+  /**
+   * Specifies the height of the modal. When provided, the modal content becomes scrollable if it exceeds the specified height.
+   * If the specified height exceeds available viewport space, the modal will use the maximum available space.
+   * The minimum height is constrained to ensure the content area is at least 60px for accessibility.
```
jperals: "constrained" suggests that a maximum is applied to a given value, but here it is a minimum. Can we find a different phrasing? Is this the vertical equivalent of the minimum 320px width mentioned in the `width` prop above? If so, could we harmonize both descriptions?

src/modal/styles.scss
```diff
@@ -85,6 +125,12 @@ $modal-z-index: 5000;
   border-end-start-radius: awsui.$border-radius-container;
   border-end-end-radius: awsui.$border-radius-container;
   box-shadow: awsui.$shadow-modal;
+
+  &.custom-height-container {
```
jperals, line 114: Is it possible to make the entire document be scrollable instead of the modal content, as the current behavior when custom height is not used?

I noticed that I can achieve this by removing this ruleset, but it is needed when the modal is shorter than the viewport. Maybe this can be used conditionally?

amanabiy: When switching between scroll modes (content overflow vs document scroll), the layout shift causes the content to jump unexpectedly. This creates a poor user experience. So, I think dynamically toggling between these two overflow behaviors isn't the right approach. I think the best thing would be to remove the stickiness of the footer, to make it consistent with when there is no footer provided. I have made the modal to behave consistently when there is a footer and without footer as well, now we always have a fixed padding on top and bottom.

pages/modal/custom-dimensions.page.tsx
```diff
@@ -0,0 +1,65 @@
...
+              type="number"
+            />
+          </FormField>
+        </SpaceBetween>
+      }
+    >
+      <Button data-testid="modal-trigger" onClick={() => setVisible(true)}>
+        Show modal
+      </Button>
+      <ScreenshotArea>
```
jperals, line 49: Nitpick: actually the ScreenshotArea in these pages does nothing, because the modal is positioned visually outside of it. For this reason we need to use `captureViewport` in the modal visual regression tests and not `captureScreenshotArea`.

pages/modal/custom-height-no-footer.page.tsx
```diff
@@ -0,0 +1,43 @@
...
+// SPDX-License-Identifier: Apache-2.0
+import React, { useState } from 'react';
+
+import { Button, Modal } from '~components';
+
+import { SimplePage } from '../app/templates';
+import ScreenshotArea from '../utils/screenshot-area';
+
+function getUrlParams() {
+  const hashParts = window.location.hash.split('?');
```
jperals: Still using the custom URL parameters logic here instead of the existing context hooks :)

Actually could this page be instead implemented as an extra parameter in the other dev page (`custom-dimensions.page.tsx`), which adds or removes the footer?

src/modal/interfaces.ts
```diff
@@ -24,6 +24,20 @@ export interface BaseModalProps {
...
+   * Specifies the width of the modal. When provided, takes precedence over the `size` property.
+   * If the specified width exceeds available viewport space, the modal will use the maximum available space.
+   * The minimum width is 320px (equivalent to the `small` size).
+   */
+  width?: number;
+
+  /**
+   * Specifies the height of the modal. When provided, the modal content becomes scrollable if it exceeds the specified height.
+   * If the specified height exceeds available viewport space, the modal will use the maximum available space.
+   * The minimum height is header height + footer height + 60px content height (to ensure content is accessible).
```
jperals: I think the last paragraph is a not very easy to understand. What about something like:
> To make sure the content is accessible, the component will apply a minimum height of 60px for the content area, plus the header height and the footer height.

src/modal/styles.scss
```diff
@@ -33,9 +34,8 @@ $modal-z-index: 5000;
 
 .focus-lock {
   align-self: flex-start;
-  margin-block: auto;
   margin-inline: auto;
-  padding-block: awsui.$space-s;
+  padding-block: awsui.$space-static-s;
```
jperals: Is it still necessary to replace `$space-s` with `$space-static-s`?

pages/modal/custom-height-no-footer.page.tsx
amanabiy: Yes. Done

pages/modal/custom-dimensions.page.tsx
amanabiy: I would keep the `ScreenshotArea` because in `captureViewport` when opening the page we use `openScenario` and it includes a check for the screenshotArea to be visible.

src/modal/interfaces.ts
amanabiy: Good point

src/modal/styles.scss
amanabiy: I have updated it to $space-s.

src/modal/internal.tsx
```diff
@@ -217,9 +223,47 @@ function PortaledModal({
...
+  // Minimum content height is twice the height of the largest type style in Cloudscape (60px)
+  // to ensure content remains scrollable and accessible even with custom heights.
+  const MIN_CONTENT_HEIGHT = 60;
+  const MIN_MODAL_WIDTH = 320; // Matches smallest predefined size (small)
+
+  // Calculate minimum modal height based on header, footer, and content
+  const minModalHeight = (headerHeight ?? 0) + (footer ? (footerHeight ?? 0) : 0) + MIN_CONTENT_HEIGHT;
+
+  // Constrain dimensions to minimum values
+  const constrainedHeight = height && height < minModalHeight ? minModalHeight : height;
```
jperals: If I understand this correctly, can this be simplified a bit, like

```
const constrainedHeight = Math.max(height ?? 0, minModalHeight);
const constrainedWidth = Math.max(width ?? 0, MIN_MODAL_WIDTH);

const hasCustomHeight = height !== undefined;
const hasCustomWidth = width !== undefined;
```

pages/modal/custom-dimensions.page.tsx
```diff
@@ -0,0 +1,66 @@
...
+        <Modal
+          header="Custom dimensions modal"
+          visible={visible}
+          onDismiss={() => setVisible(false)}
+          width={urlParams.width ? Number(urlParams.width) : undefined}
```
jperals: Tried the page locally and for some reason the dimensions are not applied to the modal (once any value is set, it is as if it would be set to infinite). Something might be wrong either with the page or with the component logic

src/modal/internal.tsx
amanabiy: Yes that's right. updated it accordingly.

pages/modal/custom-dimensions.page.tsx
amanabiy: Thanks for catching that. I have used the check [Number.isFinite()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isFinite) to check the number is finite and custom widths has been provided.

src/modal/styles.scss
```diff
@@ -82,12 +87,21 @@ $modal-z-index: 5000;
     max-inline-size: 1280px;
   }
 
+  &.custom-width {
+    max-inline-size: var(#{custom-props.$modalCustomWidth, $modal-size-medium});
```
jperals: Why do we default to `$modal-size-medium`? Is it for extra safety for edge cases like Inifinity or exponential number expressions (e.g, `1e9`)?

amanabiy: Defaulting to `$modal-size-medium`, is just done as a precaution for in any case if the variable doesn't exist. An Infinity, or exponential number should be handled when we are calculating the constrainedWidth. I will keep it simple, and remove it since we only apply this class if there is a width.

src/modal/internal.tsx
```diff
@@ -216,9 +223,46 @@ function PortaledModal({
   // Add extra scroll padding to account for the height of the sticky footer,
   // to prevent it from covering focused elements.
   const [footerHeight, footerRef] = useContainerQuery(rect => rect.borderBoxHeight);
-  const headerRef = useRef<HTMLDivElement>(null);
+  const [headerHeight, headerRef] = useContainerQuery(rect => rect.borderBoxHeight);
```
jperals: Not a blocker, but I wonder if we could abstract away most of this new logic, e.g, in a hook.

src/modal/internal.tsx
```diff
@@ -247,19 +291,27 @@ function PortaledModal({
...
+              disabled={!visible}
+              autoFocus={true}
+              restoreFocus={true}
+              className={clsx(styles['focus-lock'], styles[`position-${position}`])}
+            >
               <div
                 className={clsx(
                   styles.dialog,
-                  styles[size],
+                  !hasCustomWidth && styles[size],
```
NathanZlion, line 268: the width takes precedence over size, we've already mentioned it in the docs but should we perhaps add a simple dev warning above ,just like we do for the too small height case, just to simply say "hey you used size and width at the same time, width is used here..."

src/modal/internal.tsx
amanabiy: Done, Implemented it in this PR.

src/modal/internal.tsx
amanabiy: Thanks for the suggestion! I'd prefer to hold off on the dev warning for using size and width together. The documentation already states width takes precedence. Plus doing this would remove the default value from our documenter snapshot. We can always add it later if we see builders getting confused in practice.    We can also add it in a separate PR, without blocking this one (if we decide to add it). What do you think?

NathanZlion: okay, if that's the case it can be a follow-up.

src/modal/styles.scss
```diff
@@ -141,6 +162,11 @@ $modal-z-index: 5000;
   inset-block-end: 0;
   z-index: 800;
 
+  &.custom-height {
```
jperals, line 165: Doesn't the footer have the same border styles when the modal has custom height as when it doesn't but it is "stuck"? Can we merge these two rulesets under one single combined selector? Like

```
&--stuck,
&.custom-height {
    border-start-start-radius: 0;
    border-start-end-radius: 0;
    border-end-start-radius: awsui.$border-radius-container;
    border-end-end-radius: awsui.$border-radius-container;
}
```

src/modal/use-modal-dimensions.ts
```diff
@@ -0,0 +1,55 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
jperals, line 1: Thanks for extracting this into a separate hook, this makes the component code easier to deal with 👍

src/modal/__tests__/use-modal-dimensions.test.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+    test('does not warn when dimensions are not provided', () => {
+      renderHook(useModalDimensions, {
+        initialProps: { hasFooter: false },
+      });
+
+      expect(warnOnce).not.toHaveBeenCalled();
+    });
+
+    test('does not warn when dimensions are NaN', () => {
```
jperals, line 132: I'd say passing NaN this is quite an edge case which we don't necessarily want to deal with on our end, but for at the same time and kind of for the same reason I don't see why we would _intentionally_ not want to warn to the point of writing a test about it. If we really care about this case and are willing to take the effort to take it into account, maybe we just go ahead and issue this warning?

amanabiy: This makes sense. If dimensions are NaN, we should treat them as unset and warn the user. I'll address this in a follow-up PR along with consolidating the `--stuck` and `custom-height` footer styles, and adding a dev warning when size and width/height are used together.

PR #4253, opened 2026-02-13
refactor: move theme validation after render setup in tests

srungta08:
### Description

Improved test organization by moving validateThemeCompatibility calls to after component render and variable destructuring in 17 tests.

Benefits:
- Better code organization: setup → render → validate → test
- Allows test setup code (registerDrawer, delay) to run for all themes
- Validates component state after setup completes
- More consistent test structure

Pattern change:
Before:
  registerDrawer(...) await delay() if (await validateThemeCompatibility(...)) return; const { wrapper } = await renderComponent(...)

After:
  registerDrawer(...) await delay() const { wrapper } = await renderComponent(...) if (await validateThemeCompatibility(...)) return;

This ensures drawer registration and delays execute for all themes, then validates the rendered state before proceeding with assertions.

Tests affected: 17 tests in toolbar mode suite
Net change: +142/-129 lines (formatting improvements)

### How has this been tested?

Ran unit tests with:

```npm run test```

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/__tests__/runtime-drawers.test.tsx
+++ b/src/app-layout/__tests__/runtime-drawers.test.tsx
@@ -837,12 +837,8 @@ describe('toolbar mode only features', () => {
    * Returns true if the component is compatible with non-toolbar themes (test should be skipped),
    * false if running in toolbar mode (test should continue with full assertions).
    */
-  async function validateThemeCompatibility(currentTheme: string, component: React.ReactElement): Promise<boolean> {
-    if (currentTheme !== 'refresh-toolbar') {
-      await expect(renderComponent(component)).resolves.not.toThrow();
-      return true;
-    }
-    return false;
+  function isNotRefreshToolbarTheme(currentTheme: string): boolean {
+    return currentTheme !== 'refresh-toolbar';
   }
   describeEachAppLayout({ sizes: ['desktop'] }, ({ theme }) => {
     test('should contain overridden in AWS-UI-Widget-Global-Navigation css classes for drawers', async () => {
@@ -857,7 +853,7 @@ describe('toolbar mode only features', () => {
 
       await delay();
 
-      if (await validateThemeCompatibility(theme, <AppLayout drawers={[testDrawer]} />)) {
+      if (isNotRefreshToolbarTheme(theme)) {
         return;
       }
 
@@ -878,7 +874,7 @@ describe('toolbar mode only features', () => {
 
       await delay();
 
-      if (await validateThemeCompatibility(theme, <AppLayout drawers={[testDrawer]} />)) {
+      if (isNotRefreshToolbarTheme(theme)) {
         return;
       }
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/app-layout/__tests__/runtime-drawers.test.tsx [hunk 1]
```diff
@@ -1129,13 +1129,13 @@ describe('toolbar mode only features', () => {
             mountContent: container => (container.textContent = 'global drawer content 1'),
           });
 
+          const renderProps = await renderComponent(<AppLayout drawers={[testDrawer]} />);
```
gethinwebster, line 1138: Given that `validateThemeCompatibility` tests whether `renderComponent` throws an error I would suggest keeping it before any non-wrapped calls to `renderComponent`. That way the resulting error messages if something goes wrong will be more descriptive.

src/app-layout/__tests__/runtime-drawers.test.tsx
```diff
@@ -1236,10 +1237,6 @@ describe('toolbar mode only features', () => {
         });
 
         test('should render trigger buttons for global drawers even if local drawers are not present', async () => {
-          if (await validateThemeCompatibility(theme, <AppLayout toolsHide={true} />)) {
```
gethinwebster, line 1239: however this is a (partially) valid change: for `validateThemeCompatibility` to be useful it needs to run after `registerDrawer`, otherwise there's not really anything for it to test

src/app-layout/__tests__/runtime-drawers.test.tsx
georgylobko: Why? The purpose of all of these tests is to make sure the runtime api does not crash the page, same as here: https://github.com/cloudscape-design/components/blob/main/src/app-layout/__tests__/runtime-drawers.test.tsx#L81

If the page is crashed because of AppLayout props, it will fail on `renderComponent` anyway. The test condition with early return ensures that we don't test features parity for classic theme, only the face that it's not crashing

srungta08:
I have update some of the scenarios which were missed in previous revision

src/app-layout/__tests__/runtime-drawers.test.tsx
gethinwebster: If `validateThemeCompatibility` is run first, then a clearer error message will be thrown, that's traceable back to that function. Alternatively, if we want the tests to follow the pattern of always running `renderComponent` first, then we should remove `renderComponent` from `validateThemeCompatibility`, as it makes no sense to run that twice

gethinwebster: (reading through the other tests again, I think the latter approach would make more sense: remove all rendering from `validateThemeCompatibility`, and rename/replace that with something like:

```
if (!themeSupportsDrawersFunctionality(theme) {
  return;
}
```

So long as that is always after all of the functional `renderComponent` and `registerDrawer` calls then it should give us exactly the same test coverage with simpler and easier to follow code

PR #4256, opened 2026-02-16
fix: Addresses styles-check flakiness

pan-kot:
### Description

The styles check util validates the presence of the component styles on the page and logs an error like `Missing AWS-UI CSS for theme...` to the console when the styles are not present.

This util causes integ tests flakiness with over 10 recorded occurrences.

The PR addresses the problem by making changes to the util so that it validates that the stylesheets (present upon document readiness) are loaded.

### How has this been tested?

* New unit tests to cover the util changes
* Dry-run to live
* 400+ test runs with React v16 and v18 of the most affected tests: there was not a single occurrence of the issue in the logs (I checked for "AWS-UI CSS" in GitHub actions logs instead of looking at the test failures as we have retries. Before the fix these messages were consistently present).

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/app-layout/__integ__/multi-app-layout-navigation.test.ts
+++ b/src/app-layout/__integ__/multi-app-layout-navigation.test.ts
@@ -18,59 +18,57 @@ class PageObject extends BasePageObject {
   }
 }
 
-for (let i = 0; i < 200; i++) {
-  describe('Multi app layout navigation', () => {
-    describe.each([[true], [false]])('iframe=%s', iframe => {
-      const mainLayout = createWrapper().find('[data-testid="main-layout"]').findAppLayout();
-      const secondaryLayout = createWrapper().find('[data-testid="secondary-layout"]').findAppLayout();
-      const setupTest = (testFn: (page: PageObject) => Promise<void>) =>
-        useBrowser(async browser => {
-          const page = new PageObject(browser);
-          await browser.url(
-            `#/light/app-layout/multi-layout-with-hidden-instances${iframe ? '-iframe' : ''}?${getUrlParams(theme)}`
-          );
-          await testFn(page);
+describe('Multi app layout navigation', () => {
+  describe.each([[true], [false]])('iframe=%s', iframe => {
+    const mainLayout = createWrapper().find('[data-testid="main-layout"]').findAppLayout();
+    const secondaryLayout = createWrapper().find('[data-testid="secondary-layout"]').findAppLayout();
+    const setupTest = (testFn: (page: PageObject) => Promise<void>) =>
+      useBrowser(async browser => {
+        const page = new PageObject(browser);
+        await browser.url(
+          `#/light/app-layout/multi-layout-with-hidden-instances${iframe ? '-iframe' : ''}?${getUrlParams(theme)}`
+        );
+        await testFn(page);
+      });
+
+    test(
+      'should clean up and restore previous route state',
+      setupTest(async page => {
+        await page.click(mainLayout.findToolsToggle().toSelector());
+        await expect(page.isToolsToggleActive()).resolves.toBe(true);
+        await page.runInsideIframe('#page1', !!iframe, async () => {
+          await expect(page.isClickable(secondaryLayout.findTools().toSelector())).resolves.toEqual(true);
         });
 
-      test(
-        'should clean up and restore previous route state',
-        setupTest(async page => {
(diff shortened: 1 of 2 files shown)


src/internal/hooks/use-base-component/__tests__/styles-check-hook.test.tsx
```diff
@@ -1,47 +0,0 @@
-// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
pan-kot, line 1: I renamed this file to "styles-check.test.tsx" as now there is a single util that we test.

src/internal/hooks/use-base-component/__tests__/styles-check-pure.test.ts
```diff
@@ -1,185 +0,0 @@
-// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
pan-kot, line 1: I removed this file where the tests were done by calling `checkMissingStyles` and `documentReadyAndIdle` utils directly. Instead, all tests are now done against the `useMissingStylesCheck`.

I ensured that every test from this file is represented in the new tests.

src/internal/hooks/use-base-component/styles-check.ts
```diff
@@ -25,18 +26,41 @@ function documentReady(document: Document, callback: () => void) {
...
     documentReady(document, () => {
-      setTimeout(() => {
-        requestIdleCallback(() => resolve());
-      }, 1000);
+      setTimeout(() => requestIdleCallback(() => resolve()), 1000);
     });
   });
+
+  const stylesheets = Array.from(document.querySelectorAll<HTMLLinkElement>('link[rel="stylesheet"]'));
+  await Promise.all(
```
jperals, line 38: Should we use `Promise.allSettled` instead? So that if some stylesheet aborts to load, we still wait for the rest.

src/internal/hooks/use-base-component/styles-check.ts
```diff
@@ -25,18 +26,41 @@ function documentReady(document: Document, callback: () => void) {
...
+          link.removeEventListener('load', onLoad);
+          link.removeEventListener('error', onError);
+          signal.removeEventListener('abort', onAbort);
+          done();
+        }
+        const onLoad = () => cleanup(resolve);
+        const onError = () => cleanup(resolve);
+        const onAbort = () => cleanup(() => reject(new DOMException('Aborted', 'AbortError')));
+
+        link.addEventListener('load', onLoad);
```
jperals, line 55: Is there a possibility that a stylesheet never loads, leading the base component to never do the check?

(Or in more practical terms, that it takes way too long, preventing the check in practice)

src/internal/hooks/use-base-component/styles-check.ts
```diff
@@ -25,18 +26,41 @@ function documentReady(document: Document, callback: () => void) {
...
-export function documentReadyAndIdle(document: Document, signal: AbortSignal) {
-  return new Promise<void>((resolve, reject) => {
+async function documentReadyAndIdle(document: Document, signal: AbortSignal) {
+  await new Promise<void>((resolve, reject) => {
     signal.addEventListener('abort', () => reject(new DOMException('Aborted', 'AbortError')));
     documentReady(document, () => {
-      setTimeout(() => {
-        requestIdleCallback(() => resolve());
-      }, 1000);
+      setTimeout(() => requestIdleCallback(() => resolve()), 1000);
```
jperals, line 33: Is this hard-coded timeout still valuable?

pan-kot: I would not remove this. Technically, the stylesheets can be added to the document after its initial load, and be not available if we query them immediately.

src/internal/hooks/use-base-component/styles-check.ts
pan-kot: It can technically take a while to load, but should still be resolved before the rest of the content - so I don't think the check should fail in that case.

src/internal/hooks/use-base-component/styles-check.ts
pan-kot: I don't think that is needed: the abort signal is only sent when the util un-mounts, in which case there is no need to wait for the rest. When it aborts, the error is not logged.

PR #4259, opened 2026-02-16
feat: add `onFocusEnter`, `onFocusLeave`, `onEscape` callback props

mxschll:
### Description

- This PR refactors adds `onFocusEnter`,  `onFocusLeave`, `onEscape` event handlers to the Dropdown component 
- Renames `onDropdownClose` to `onOutsideClick`

See proposal `RGJgABPXohP8` for more context.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/internal/components/dropdown/interfaces.ts
```diff
@@ -82,9 +82,11 @@ export interface DropdownProps extends ExpandToViewport {
   open?: boolean;
 
   /**
-   * Called when a user clicks outside of the dropdown content, when it is open.
+   * Called when the user clicks outside the dropdown and trigger.
+   * The dropdown does not close automatically - the parent component
+   * must update the `open` prop to close the dropdown.
    */
-  onDropdownClose?: NonCancelableEventHandler<null>;
+  onOutsideClick?: NonCancelableEventHandler<null>;
```
pan-kot, line 96: Should we also include `onEscape` - like we have in the tooltip component?

Alternatively, we can parametrise onDropdownClose e.g. with some `reason: "click-outside" | "escape"`.

src/internal/components/dropdown/index.tsx
```diff
@@ -287,6 +295,19 @@ const Dropdown = ({
...
 
+  const isOutsideDropdownContent = (element: Element) =>
+    !dropdownRef.current || !nodeBelongs(dropdownRef.current, element);
+
+  const focusInHandler = (event: React.FocusEvent) => {
+    fireNonCancelableEvent(onFocusIn, event);
+  };
+
+  const focusOutHandler = (event: React.FocusEvent) => {
+    if (!event.relatedTarget || isOutsideDropdownContent(event.relatedTarget)) {
```
pan-kot, line 316: Why do we need this check?

mxschll: Internally we use `onBlur` which fires any time an element inside the dropdown content looses focus. So we check if the focus actually leaves the dropdown.

src/internal/components/dropdown/interfaces.ts
mxschll: I can go with `onEscape` since it matches the tooltips API.

src/internal/components/dropdown/__tests__/dropdown.test.tsx
```diff
@@ -159,6 +154,231 @@ describe('Dropdown Component', () => {
...
+      // Focus on button in dropdown content
+      screen.getByTestId('button1').focus();
+      expect(handleFocusOut).not.toHaveBeenCalled();
+
+      // Move focus to trigger - onFocusOut SHOULD fire because trigger is outside dropdown content
+      screen.getByTestId('trigger').focus();
+      expect(handleFocusOut).toHaveBeenCalledTimes(1);
+    });
+
+    test('onFocusIn fires for each focus event on nested interactive elements', async () => {
```
pan-kot: this does not look much different from the "dropdown content focus events (onFocusIn/onFocusOut)" test - is there an opportunity to simplify?

src/internal/components/dropdown/index.tsx
pan-kot: Oh, I see. Doesn't it create inconsistency with focus-in then? The focus-in fires anytime an element inside the dropdown gets focus, even if the focus already was inside the dropdown. The focus-out only fires when the focus leaves the dropdown though.

I also researched the diff between focus/focusin and blur/focusout - it looks like there is no actual diff in React. I think what we might do if we want to add some customisation (checking targets) - then we better call the handlers somewhat distinctly to avoid confusion, e.g. onFocusEnter/onFocusLeave (similar to mouseenter/mouseleave) - wdyt?

mxschll: Good catch, thanks! It was supposed to get triggered only once the focus enters the dropdown. I think renaming is actually a good point, as they behave differently than the focusin and focusout events.

src/internal/components/dropdown/__tests__/dropdown.test.tsx
mxschll: Done.

src/internal/components/dropdown/index.tsx
mxschll: Done.

src/internal/components/dropdown/index.tsx
```diff
@@ -294,6 +303,21 @@ const Dropdown = ({
...
+    !dropdownRef.current || !nodeBelongs(dropdownRef.current, element);
+
+  const focusEnterHandler = (event: React.FocusEvent) => {
+    if (!event.relatedTarget || isOutsideDropdownContent(event.relatedTarget)) {
+      fireNonCancelableEvent(onFocusEnter, event);
+    }
+  };
+
+  const focusLeaveHandler = (event: React.FocusEvent) => {
+    if (!event.relatedTarget || isOutsideDropdownContent(event.relatedTarget)) {
```
pan-kot, line 316: I like how symmetrical these two handlers now are :)

PR #4267, opened 2026-02-19
feat: add `fontFamilyHeading` design token

mxschll:
### Description

Adds a new `fontFamilyHeading` design token the theme header fonts. If not provided, it falls back to `fontFamilyBase`. 

### How has this been tested?

Verified locally by applying a custom theme.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/__tests__/snapshot-tests/__snapshots__/design-tokens.test.ts.snap
+++ b/src/__tests__/snapshot-tests/__snapshots__/design-tokens.test.ts.snap
@@ -2393,7 +2393,7 @@ exports[`Design tokens artifacts Design tokens JSON for classic matches the snap
           "$description": "The default font family that will be applied globally to the product interface.",
           "$value": "'Noto Sans', 'Helvetica Neue', Roboto, Arial, sans-serif",
         },
-        "font-family-header": {
+        "font-family-heading": {
           "$description": "The font family for headings and component headers. Defaults to the base font family.",
           "$value": "'Noto Sans', 'Helvetica Neue', Roboto, Arial, sans-serif",
         },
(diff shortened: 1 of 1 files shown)


src/internal/styles/typography/mixins.scss
```diff
@@ -42,6 +43,7 @@
 }
 
 @mixin font-heading-s($weight: true) {
+  font-family: constants.$font-family-header;
   font-size: awsui.$font-size-heading-s;
   line-height: awsui.$line-height-heading-s;
   letter-spacing: awsui.$letter-spacing-heading-s;
```
jkuelz: We use 'heading' naming for all other generic heading tokens, should we call this `fontFamilyHeading` instead? It would also match the corresponding color token as well: `colorTextHeadingDefault`

mxschll: Good catch, totally agree!

mxschll: Updated.

src/internal/styles/typography/constants.scss
```diff
@@ -10,6 +10,7 @@ $base-size: 10px;
 
 // Font families
 $font-family-base: awsui.$font-family-base;
+$font-family-heading: awsui.$font-family-heading;
```
jkuelz, line 13: Nit: We probably don't need this extra assignment layer for a new token. I think most of this file was legacy before we had design tokens.

mxschll: I think we should remove [constants.scss](https://github.com/cloudscape-design/components/pull/4267/files/e07b91b765f0047bfb0a9dfad2626f23721c31aa#diff-e878ecfde98f275d4849b47da85b403d559c2c6b272e259176e6f752a5034082) entirely in that case...

PR #4247, opened 2026-02-12
feat: Adds dataAttributes support to ButtonDropdown items

giridharsaday:
### Description

This PR adds support for custom data attributes on ButtonDropdown items, enabling teams to add analytics tracking and custom metadata without DOM manipulation workarounds.

**Key Changes:**
- Add `dataAttributes` property to Item interface (CheckboxItem inherits it)
- Automatically prefix attribute names with 'data-'
- Prevent 'testid' override to preserve existing behavior
- Add comprehensive unit tests including checkbox items

**Use Case:**
Teams building analytics-instrumented UIs (like NGS Hub with Tricorder) need to add data-* attributes to track user interactions consistently across multiple instances of the same action.

**API Design:**
```
<ButtonDropdown
 items={[
   {
     id: 'edit',
     text: 'Edit',
     dataAttributes: {
       'analytics-action': 'edit-product',
       'item-key': 'product-123'
     }
   }
 ]}
/>
// Renders as: data-analytics-action="edit-product" data-item-key="product-123"
```

Related links: 
- SIM Ticket: AWSUI-61657
- API Proposal: [AdJuAxOmsoQW]

### How has this been tested?

- ✅ Added comprehensive unit tests covering:
  - Custom data attributes rendering
  - Automatic "data-" prefix addition
  - "testid" key exclusion
  - Undefined value handling
  - Multiple items
  - Disabled items
  - Checkbox items
- ✅ Manually tested in dev environment - attributes render correctly on `<li>` elements
- ✅ Verified existing tests still pass
- ✅ Tested keyboard navigation and accessibility - no regressions

Reviewers can test by:
1. Run unit tests: `npm test -- src/button-dropdown/__tests__/data-attributes.test.tsx`
2. Start dev server: `npm start`
3. Inspect dropdown items in browser DevTools to verify data-* attributes

<details>
   <summary>Review checklist</summary>

#### Correctness

- ✅ Changes include appropriate documentation updates (JSDoc comments on interface).
- ✅ Changes are backward-compatible - new optional property.
- ✅ Changes do not include unsupported browser features - uses standard data-* attributes.
- ✅ Changes were manually tested for accessibility - no impact on screen readers or keyboard navigation.

#### Security

- ✅ No URL handling in this change.

#### Testing

- ✅ Changes are covered with new unit tests (7 test cases).
- ⚠️ Integration tests not added yet - can be added if requested.
</details>

---

[...]

PR #4269, opened 2026-02-19
feat: Add dataAttributes support to ButtonDropdown MainAction

giridharsaday:
# Add dataAttributes support to ButtonDropdown MainAction

## Description

This PR adds `dataAttributes` support to ButtonDropdown's `MainAction` interface, completing the feature that was partially implemented in #4247.

**What was missing:**
The original PR #4247 added `dataAttributes` to dropdown items but overlooked the main action button. This meant teams couldn't apply consistent analytics tracking across all ButtonDropdown actions.

**Key Changes:**
* Add `dataAttributes` property to `MainAction` interface
* Apply data attributes to main action button via `nativeButtonAttributes`
* Add unit tests for mainAction dataAttributes

**Use Case:**
Teams building analytics-instrumented UIs need to track user interactions on both dropdown items AND the main action button. Without this, the main action button requires DOM manipulation workarounds.

**API Design:**
```typescript
<ButtonDropdown
  items={[{ id: 'item1', text: 'Item 1' }]}
  mainAction={{
    text: 'Main Action',
    dataAttributes: {
      'analytics-action': 'primary-action',
      'button-type': 'main'
    }
  }}
>
  Actions
</ButtonDropdown>
// Renders as: data-analytics-action="primary-action" data-button-type="main"
```

## Implementation Details

1. **Interface Update**: Added `dataAttributes?: Record<string, string>` to `MainAction` interface
2. **Rendering**: Applied attributes via `nativeButtonAttributes` using the existing `getDataAttributes()` utility
3. **Consistency**: Uses the same utility function and behavior as dropdown items (auto-prefix with 'data-', exclude 'testid')

## How has this been tested?

* ✅ Added unit tests covering:
  * Custom data attributes rendering on mainAction
  * Automatic "data-" prefix addition
  * "testid" key exclusion
  * No duplicate "data-" prefix when already present
* ✅ Used ButtonDropdown test-utils for element selection
* ✅ Verified existing tests still pass

Reviewers can test by:
1. Run unit tests: `npm test -- src/button-dropdown/__tests__/data-attributes.test.tsx`

### Correctness
* ✅ Changes include appropriate documentation
* ✅ Changes are backward-compatible - new optional property
* ✅ Changes do not include unsupported browser features

### Security
* ✅ No URL handling in this change
* ✅ Reuses existing validated utility function

### Testing
* ✅ Changes are covered with new unit tests

## Related

* Original PR: #4247 (Added dataAttributes to items, but missed mainAction)

### DIFF
--- a/src/button-dropdown/internal.tsx
+++ b/src/button-dropdown/internal.tsx
@@ -18,6 +18,7 @@ import { useVisualRefresh } from '../internal/hooks/use-visual-mode/index.js';
 import { isDevelopment } from '../internal/is-development';
 import { spinWhenOpen } from '../internal/styles/motion/utils';
 import { checkSafeUrl } from '../internal/utils/check-safe-url';
+import { getDataAttributes } from '../internal/utils/data-attributes';
 import {
   GeneratedAnalyticsMetadataButtonDropdownCollapse,
   GeneratedAnalyticsMetadataButtonDropdownExpand,
@@ -227,7 +228,7 @@ const InternalButtonDropdown = React.forwardRef(
         </div>
       );
     } else if (hasMainAction) {
-      const { text, iconName, iconAlt, iconSvg, iconUrl, external, externalIconAriaLabel, ...mainActionProps } =
+      const { text, iconName, iconAlt, iconSvg, iconUrl, external, externalIconAriaLabel, dataAttributes, ...mainActionProps } =
         mainAction;
       const mainActionIconProps = external
         ? ({ iconName: 'external', iconAlign: 'right', target: '_blank', rel: 'noopener noreferrer' } as const)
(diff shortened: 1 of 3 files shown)


src/button-dropdown/internal.tsx
```diff
@@ -252,7 +253,10 @@ const InternalButtonDropdown = React.forwardRef(
           ariaLabel={mainActionAriaLabel}
           formAction="none"
           nativeAnchorAttributes={nativeMainActionAttributes?.anchor}
-          nativeButtonAttributes={nativeMainActionAttributes?.button}
+          nativeButtonAttributes={{
+            ...getDataAttributes(dataAttributes, ['testid']),
```
pan-kot, line 257: Why excluding `testid` in this case?

pan-kot: Also, why do we pass data attributes to native attributes? It should work just fine when adding them to the internal button directly.

pan-kot: waait.. Can't we already use the `nativeMainActionAttributes` to pass the data-* attributes? Why is a separate prop needed?

giridharsaday:
You're absolutely right! I didn't realize `nativeMainActionAttributes.button` already allows passing data-* attributes directly.

Users can achieve this today with:
```
typescript
<ButtonDropdown
 items={[...]}
 mainAction={{ text: 'Main Action' }}
 nativeMainActionAttributes={{
   button: {
     'data-analytics-action': 'primary-action',
     'data-button-type': 'main'
   }
 }}
/>
```

This makes the new `dataAttributes` property on `MainAction` unnecessary. Closing this PR since the functionality already exists through `nativeMainActionAttributes`.

Thank you for catching this!

PR #4272, opened 2026-02-19
refactor: replace legacy vars with design tokens

mxschll:
### Description

Replaces reference to legacy token with `awsui.$*`.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/styles/typography/mixins.scss
+++ b/src/internal/styles/typography/mixins.scss
@@ -32,8 +32,10 @@
   line-height: awsui.$line-height-body-m;
 }
 
-@mixin font-heading-xs($weight: true) {
-  font-family: awsui.$font-family-heading;
+@mixin font-heading-xs($weight: true, $family: true) {
+  @if $family {
+    font-family: awsui.$font-family-heading;
+  }
   font-size: awsui.$font-size-heading-xs;
   line-height: awsui.$line-height-heading-xs;
   @if $weight {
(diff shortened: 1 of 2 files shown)


src/internal/styles/typography/mixins.scss
```diff
@@ -16,7 +16,7 @@
   @include font-body-m;
   color: awsui.$color-text-body-default;
   font-weight: constants.$font-weight-normal;
```
NathanZlion, line 18: nit: Is the reason we're sparing this, because it's not part of awsui?

NathanZlion: Would it be possible to completely move these 3 font-weights to the awsui:tokens?
```
$font-weight-bold: 700;
$font-weight-normal: 400;
$font-weight-lighter: 300;
```

I see they're still being used in only 2 other places:
- Date Range Picker : https://github.com/cloudscape-design/components/blob/0dd82106eb2456f80c2281d5fe0a7ca5e39ad118/src/date-range-picker/styles.scss#L66

- Calendar component: https://github.com/cloudscape-design/components/blob/0dd82106eb2456f80c2281d5fe0a7ca5e39ad118/src/calendar/styles.scss#L44

mxschll: These font weights are used in too many places, so I'll leave them as is for now.

NathanZlion (approved):

src/internal/styles/typography/mixins.scss
NathanZlion: yeah, makes sense. could do a follow-up. Looks good for now.

NathanZlion (approved):

PR #4239, opened 2026-02-11
feat: introduce `minWidth` and `maxWidth` properties

mxschll:
### Description  
This PR introduces **`minWidth`** and **`maxWidth`** sizing props for the Dropdown component, replacing the previous boolean-based approach.

**Changes Made**

- Added `minWidth` and `maxWidth` props to the Dropdown component, allowing to define explicit width constraints instead of relying on boolean flags
- Updated all components that consume the Dropdown to utilize the new `minWidth` and `maxWidth` props
- Moved legacy logic related to `expandToViewport=true` out of the Dropdown component and into the respective consuming components. This ensures that width expansion does not exceed defined constraints of `xxs`
  **Context:** `AWSUI-19898`, https://github.com/cloudscape-design/components/pull/1423
- Migrated styling logic that previously depended on boolean flags into dedicated prop (`hideBlockBorder`)
  **Context:** `AWSUI-21049`, `AWSUI-32517`, https://github.com/cloudscape-design/components/pull/1961

**Related**

[Quip proposal: RGJgABPXohP8]

### How has this been tested?

- Manual testing
- Unit tests green
- Integration tests green
- Pipeline green (though some flakiness)

<details>
   <summary>Review checklist</summary>

_The following items are to be evaluated by the author(s) and the reviewer(s)._

#### Correctness

- _Changes include appropriate documentation updates._
- _Changes are backward-compatible if not indicated, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#public-apis)._
- _Changes do not include unsupported browser features, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#browsers-support)._
- _Changes were manually tested for accessibility, see [accessibility guidelines](https://cloudscape.design/foundation/core-principles/accessibility/)._

#### Security

- _If the code handles URLs: all URLs are validated through [the `checkSafeUrl` function](https://github.com/cloudscape-design/components/blob/main/src/internal/utils/check-safe-url.ts)._

#### Testing

- _Changes are covered with new/existing unit tests?_
- _Changes are covered with new/existing integration tests?_
</details>

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #4283, opened 2026-02-23
fix: restore legacy dropdown sizing behavior 

mxschll:
### Description

This PR restores legacy behavior in Select, Multiselect and Autosuggest when `expandToViewport=true`.

**Current behavior:**  When `expandToViewport=true`, the dropdown’s width scales up to the `xss` breakpoint but can be smaller than the trigger if the content inside is narrower.

**Expected behavior:**  When `expandToViewport=true`, the dropdown’s width matches the width of the trigger element, but only up to the `xss` breakpoint. This behavior is consistent with the legacy implementation (see `AWSUI-19898`).

image

image

We were missing a specific test case where the dropdown content is narrower than the trigger, which caused this behavior change to slip through in [#4239](https://github.com/cloudscape-design/components/pull/4239).

### How has this been tested?

Pipeline is green (will add tests for that specific scenario).

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/components/autosuggest-input/__tests__/autosuggest-input.test.tsx
+++ b/src/internal/components/autosuggest-input/__tests__/autosuggest-input.test.tsx
@@ -11,11 +11,6 @@ import AutosuggestInput, {
 import Dropdown from '../../../../../lib/components/internal/components/dropdown';
 import AutosuggestInputWrapper from '../../../../../lib/components/test-utils/dom/internal/autosuggest-input';
 
-jest.mock('@cloudscape-design/component-toolkit/internal', () => ({
-  ...jest.requireActual('@cloudscape-design/component-toolkit/internal'),
-  useResizeObserver: jest.fn(),
-}));
-
 function render(jsx: React.ReactElement) {
   const { container, rerender, getByTestId } = renderJsx(jsx);
   const wrapper = new AutosuggestInputWrapper(container);


src/internal/utils/get-dropdown-min-width.ts
```diff
@@ -0,0 +1,31 @@
...
+import { getBreakpointValue } from '../breakpoints';
+import { DropdownWidthConstraint } from '../components/dropdown/interfaces';
+
+/**
+ * AWSUI-19898
+ * If `expandToViewport` is true, we want to constrain the dropdown width to the smallest of:
+ * 1. The trigger width
+ * 2. The xxs breakpoint width
+ *
+ * If `expandToViewport` is false, we want to constrain the dropdown width to the trigger width.
```
jperals, line 13: I'd suggest a different word than "constrain", because it suggest applying a minimum but it is used as a minimum, if I understand correctly.

src/internal/utils/get-dropdown-min-width.ts
```diff
@@ -0,0 +1,31 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { getBreakpointValue } from '../breakpoints';
+import { DropdownWidthConstraint } from '../components/dropdown/interfaces';
+
+/**
+ * AWSUI-19898
+ * If `expandToViewport` is true, we want to constrain the dropdown width to the smallest of:
```
jperals, line 9: Why is the behavior different depending on the value of `expandToViewport`?
If they used to behave differently, I think it could be because when `expandToViewport` is true it was not easy to make the dropdown width match the trigger width via CSS, because it is detached from it. But now you are gathering the trigger width with a resize observer. I might be wrong, but if this is the case and there is no UX reason to make the two cases behave differently, I'd suggest to either
1. take this chance and bring them to the same behavior, using the value we have from the resize observer, or
2. as a more conservative option, if we keep the different behavior, then I'd consider if the resize observer is necessary or CSS is enough

PR #4288, opened 2026-02-24
fix: dynamically added columns ignore minWidth

### LINKED ISSUE #4236: [Bug]: resizableColumns: dynamically added columns ignore minWidth, fall back to DEFAULT_COLUMN_WIDTH

SpyZzey:
### Description
Fixes #4236.
Currently, when adding columns to a table with minWidth > DEFAULT_COLUMN_WIDTH, they will still use DEFAULT_COLUMN_WIDTH, even though the minimum width is higher. This PR fixes this issue by checking if minWidth > DEFAULT_COLUMN_WIDTH and if so, setting the width to minWidth.

### How has this been tested?
- Manually tested.
- Added unit tests.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/table/__tests__/columns-width.test.tsx
+++ b/src/table/__tests__/columns-width.test.tsx
@@ -143,6 +143,10 @@ test('should respect minWidth when dynamically adding columns via visibleColumns
     { id: 'small-width', header: '', cell: () => '-', width: 80, minWidth: 150 },
     { id: 'width-only', header: '', cell: () => '-', width: 180 },
     { id: 'minWidth-larger', header: '', cell: () => '-', width: 180, minWidth: 200 },
+    { id: 'no-width-no-minWidth', header: '', cell: () => '-' },
+    { id: 'no-width-minWidth-large', header: '', cell: () => '-', minWidth: 200 },
+    { id: 'no-width-minWidth-small', header: '', cell: () => '-', minWidth: 80 },
+    { id: 'width-larger-than-minWidth', header: '', cell: () => '-', width: 200, minWidth: 100 },
   ];
   const { wrapper, rerender } = renderTable(
     <Table columnDefinitions={columns} visibleColumns={['id']} items={defaultItems} resizableColumns={true} />
@@ -153,7 +157,16 @@ test('should respect minWidth when dynamically adding columns via visibleColumns
   rerender(
     <Table
       columnDefinitions={columns}
-      visibleColumns={['id', 'small-width', 'width-only', 'minWidth-larger']}
+      visibleColumns={[
+        'id',
+        'small-width',
+        'width-only',
+        'minWidth-larger',
+        'no-width-no-minWidth',
+        'no-width-minWidth-large',
+        'no-width-minWidth-small',
+        'width-larger-than-minWidth',
+      ]}
       items={defaultItems}
       resizableColumns={true}
     />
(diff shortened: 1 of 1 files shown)


src/table/use-column-widths.tsx
```diff
@@ -168,7 +168,9 @@ export function ColumnWidthsProvider({ visibleColumns, resizableColumns, contain
         const column = visibleColumns[index];
         if (!columnWidths?.get(column.id) && lastVisible.indexOf(column.id) === -1) {
           updated = true;
-          newColumnWidths.set(column.id, (column.width as number) || DEFAULT_COLUMN_WIDTH);
+          const width = (column.width as number) || 0;
+          const minWidth = (column.minWidth as number) || width || DEFAULT_COLUMN_WIDTH;
+          newColumnWidths.set(column.id, Math.max(width || DEFAULT_COLUMN_WIDTH, minWidth));
```
NathanZlion: nit: I think this can be simplified,
instead of doing `|| DEFAULT_COLUMN_WIDTH` 2 times here we can just default widths to that if not provided. And minWidth can be defaulted to width if not provided. then we just have to take max of the width and the minWidth.

```
const width = (column.width as number) || DEFAULT_COLUMN_WIDTH;
const minWidth = (column.minWidth as number) || width;
newColumnWidths.set(column.id, Math.max(width, minWidth));
```

same thing, but just makes it a bit simpler to read and understand.

SpyZzey: Actually, my previous logic was flaws. The "|| 0" shouldn't have been there. Will update with your code, this simplification makes sense.

PR #4296, opened 2026-02-25
feat: AppLayout error boundaries v2

georgylobko:
### Description
Error boundaries for AppLayout areas - all areas that load as widgets now have error boundary coverage.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/app-layout/with-error-boundaries.page.tsx
+++ b/pages/app-layout/with-error-boundaries.page.tsx
@@ -1,0 +1,377 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import React, { useContext, useEffect, useRef, useState } from 'react';
+
+import { AppLayout, Button, Container, ContentLayout, Header, SpaceBetween, SplitPanel, Toggle } from '~components';
+import { AppLayoutProps } from '~components/app-layout';
+import BreadcrumbGroup from '~components/breadcrumb-group';
+import ErrorBoundary from '~components/error-boundary';
+import awsuiPlugins from '~components/internal/plugins';
+import { registerBottomDrawer, registerLeftDrawer } from '~components/internal/plugins/widget';
+import { mount, unmount } from '~mount';
+
+import AppContext, { AppContextType } from '../app/app-context';
+import { drawerLabels } from './utils/drawers';
+import appLayoutLabels from './utils/labels';
+import { splitPaneli18nStrings } from './utils/strings';
+
+type DemoContext = React.Context<
+  AppContextType<{
+    hasParentErrorBoundary: boolean | undefined;
+    hasDrawers: boolean | undefined;
+    splitPanelPosition: AppLayoutProps.SplitPanelPreferences['position'];
+  }>
+>;
+
+export default function WithErrorBoundariesPage() {
+  const { urlParams, setUrlParams } = useContext(AppContext as DemoContext);
+  const hasParentErrorBoundary = urlParams.hasParentErrorBoundary ?? false;
+  const [isBrokenNavigation, setIsBrokenNavigation] = useState(false);
+  const appLayoutRef = useRef<AppLayoutProps.Ref>(null);
+  const [breadcrumbsItems, setBreadcrumbsItems] = useState([
+    { text: 'Home', href: '#' },
+    { text: 'Service', href: '#' },
+  ]);
+
+  useEffect(() => {
+    window.addEventListener(
+      'error',
+      error => {
+        console.log('The error gets caught by the global event handler: ', error);
+      },
+      true
+    );
+  }, []);
+
+  const appLayout = (
+    <AppLayout
+      ariaLabels={{ ...appLayoutLabels, ...drawerLabels }}
+      ref={appLayoutRef}
+      breadcrumbs={<BreadcrumbGroup items={breadcrumbsItems} />}
+      content={
+        <AppLayout
+          navigationHide={true}
+          toolsHide={true}
+          content={
+            <div data-testid="app-layout-content-area">
+              <ContentLayout
+                disableOverlap={true}
+                header={
+                  <SpaceBetween size="m">
+                    <Header variant="h1">Error boundaries in app layout slots</Header>
+                  </SpaceBetween>
+                }
+              >
+                <SpaceBetween size="xs">
+                  <Toggle
+                    checked={hasParentErrorBoundary}
+                    onChange={({ detail }) => setUrlParams({ hasParentErrorBoundary: detail.checked })}
+                  >
+                    Has parent error boundary
+                  </Toggle>
+                  <Container header={<Header>Toolbar</Header>}>
+                    <Button
+                      onClick={() => {
+                        registerLeftDrawer({
+                          id: 'ai-panel',
+                          resizable: true,
+                          isExpandable: true,
+                          defaultSize: 300,
+                          preserveInactiveContent: true,
+
+                          ariaLabels: {
+                            closeButton: 'Close AI Panel drawer',
+                            content: 'AI Panel',
+                            triggerButton: 'AI Panel',
+                            resizeHandle: 'Resize handle',
+                            expandedModeButton: 'Expanded mode button',
+                            exitExpandedModeButton: 'Console',
+                          },
+
+                          trigger: {
+                            // @ts-expect-error testing error boundaries with invalid react nodes
+                            iconSvg: Symbol(),
+                          },
+
+                          mountContent: container => {
+                            mount(<div>left drawer</div>, container);
+                          },
+                          unmountContent: container => unmount(container),
+                        });
+                      }}
+                      data-testid="break-left-drawer-trigger"
+                    >
+                      Break left drawer trigger
+                    </Button>
+                    <Button
+                      onClick={() => {
+                        // @ts-expect-error testing error boundaries with invalid react nodes
+                        setBreadcrumbsItems(null);
+                      }}
+                      data-testid="break-breadcrumbs"
+                    >
+                      Break breadcrumbs
+                    </Button>
+                    <Button
+                      onClick={() => {
+                        awsuiPlugins.appLayout.registerDrawer({
+                          id: 'broken-local-drawer',
+                          type: 'local',
+                          trigger: {
+                            // @ts-expect-error testing error boundaries with invalid react nodes
+                            iconSvg: Symbol(),
+                          },
+                          defaultActive: false,
+
+                          ariaLabels: {
+                            closeButton: 'Close button',
+                            content: 'Content',
+                            triggerButton: 'Trigger button',
+                            resizeHandle: 'Resize handle',
+                          },
+
+                          mountContent: container => {
+                            mount(<div>sdfsdf</div>, container);
+                          },
+                          unmountContent: () => {},
+                        });
+                      }}
+                      data-testid="break-local-drawer-trigger"
+                    >
+                      Break a local drawer
+                    </Button>
+                    <Button
+                      onClick={() => {
+                        awsuiPlugins.appLayout.registerDrawer({
+                          id: 'broken-global-drawer',
+                          type: 'global',
+                          defaultActive: false,
+                          resizable: true,
+                          defaultSize: 320,
+                          orderPriority: -1,
+
+                          ariaLabels: {
+                            closeButton: 'Close button',
+                            content: 'Content',
+                            triggerButton: 'Trigger button',
+                            resizeHandle: 'Resize handle',
+                          },
+
+                          trigger: {
+                            // @ts-expect-error testing error boundaries with invalid react nodes
+                            iconSvg: Symbol(),
+                          },
+
+                          mountContent: container => {
+                            mount(<div>Hello!</div>, container);
+                          },
+                          unmountContent: container => unmount(container),
+                        });
+                      }}
+                      data-testid="break-global-drawer-trigger"
+                    >
+                      Break a global drawer
+                    </Button>
+                  </Container>
+                  <Container header={<Header>Panels</Header>}>
+                    <Button
+                      onClick={() => {
+                        registerLeftDrawer({
+                          id: 'ai-panel',
+                          resizable: true,
+                          isExpandable: true,
+                          defaultSize: 300,
+                          preserveInactiveContent: true,
+                          defaultActive: true,
+
+                          ariaLabels: {
+                            closeButton: 'Close AI Panel drawer',
+                            content: 'AI Panel',
+                            triggerButton: 'AI Panel',
+                            resizeHandle: 'Resize handle',
+                            expandedModeButton: 'Expanded mode button',
+                            exitExpandedModeButton: 'Console',
+                          },
+
+                          trigger: {
+                            customIcon: '',
+                          },
+
+                          mountContent: () => {
+                            throw new Error('Mount error in drawer content');
+                          },
+                          unmountContent: container => unmount(container),
+                        });
+                      }}
+                      data-testid="break-left-drawer-content"
+                    >
+                      Break left drawer content on mount
+                    </Button>
+                    <Button
+                      onClick={() => {
+                        registerLeftDrawer({
+                          id: 'ai-panel',
+                          resizable: true,
+                          isExpandable: true,
+                          defaultSize: 300,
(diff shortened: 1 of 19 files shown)


src/error-boundary/internal.tsx
```diff
@@ -87,6 +87,37 @@ export function BuiltInErrorBoundary({ wrapper, suppressNested = false, children
...
+      className={styles['app-layout-part-fallback']}
+    >
+      <ErrorBoundariesContext.Provider value={{ ...context, suppressed: nextSuppressed, renderFallback }}>
+        {children}
+      </ErrorBoundariesContext.Provider>
+    </ErrorBoundaryImpl>
+  ) : (
+    <ErrorBoundaryImpl
+      renderFallback={renderFallback}
+      onError={() => null}
```
georgylobko: We don't need to do anything here as we expect such errors to be handled in global event listeners to avoid over-reporting them.

src/app-layout/visual-refresh-toolbar/widget-areas/before-main-slot.tsx
```diff
@@ -55,7 +56,15 @@ export const BeforeMainSlotImplementation = ({ toolbarProps, appLayoutState, app
   const drawerExpandedModeInChildLayout = !!toolbarProps?.expandedDrawerId;
   const { __embeddedViewMode: embeddedViewMode } = appLayoutProps as any;
   return (
-    <>
+    <AppLayoutBuiltInErrorBoundary
+      renderFallback={() => (
```
georgylobko: I don't know how to ensure code coverage here - I expect this code to be executed as a last resort - when (if) something really unexpected happens

pages/app-layout/with-error-boundaries.page.tsx
```diff
@@ -0,0 +1,401 @@
...
+  const AppLayoutWrapper = hasParentErrorBoundary ? ErrorBoundary : 'div';
+  const appLayoutWrapperProps = hasParentErrorBoundary
+    ? {
+        onError: (error: any) =>
+          console.log('The error gets caught by the parent error boundary wrapping app layout: ', error),
+      }
+    : {};
+
+  return (
+    <AppLayoutWrapper {...(appLayoutWrapperProps as any)}>
```
just-boris: Optional comment

How to avoid that many `any`

```js
const appLayout = <AppLayout ... />;

return hasParentErrorBoundary ? <ErrorBoundary>{appLayout}<ErrorBoundary> : appLayout
```

pages/app-layout/with-error-boundaries.page.tsx
```diff
@@ -0,0 +1,401 @@
...
+                          closeButton: 'Close AI Panel drawer',
+                          content: 'AI Panel',
+                          triggerButton: 'AI Panel',
+                          resizeHandle: 'Resize handle',
+                          expandedModeButton: 'Expanded mode button',
+                          exitExpandedModeButton: 'Console',
+                        },
+
+                        trigger: {
+                          iconSvg: Symbol() as any,
```
just-boris: What is happening here?

pages/app-layout/with-error-boundaries.page.tsx
```diff
@@ -0,0 +1,401 @@
...
+                    onClick={() => {
+                      awsuiPlugins.appLayout.registerDrawer({
+                        id: 'broken-local-drawer',
+                        type: 'local',
+                        trigger: {
+                          iconSvg: `<svg viewBox="0 0 16 16" focusable="false">
+          <circle stroke-width="2" stroke="currentColor" fill="none" cx="8" cy="8" r="7" />
+          <circle stroke-width="2" stroke="currentColor" fill="none" cx="8" cy="8" r="3" />
+        </svg>`,
+                        } as any,
```
just-boris: Why is this needed?

src/__a11y__/a11y-app-layout-toolbar.test.ts
```diff
@@ -10,6 +10,7 @@ const EXCLUDED_PAGES = [
   // Test page for an app layout nested inside another through an iframe.
   // Not a use case that's encouraged.
   'app-layout/multi-layout-global-drawer-child-layout',
+  'app-layout/with-error-boundaries',
```
just-boris, line 13: I do not see any nested app layouts, why is it skipped?

src/app-layout/__tests__/widget-areas-error-boundaries.test.tsx
```diff
@@ -0,0 +1,244 @@
...
+    expect(errorBoundaryWrapper.findHeader()).toBeFalsy();
+    expect(errorBoundaryWrapper.findDescription()).toBeFalsy();
+    expect(errorBoundaryWrapper.findAction()).toBeFalsy();
+  };
+
+  describeEachAppLayout({ themes: ['refresh-toolbar'] }, () => {
+    describe.each([true, false])(
+      'entire AppLayout wrapped with error boundary component : %p',
+      (wrappedWithErrorBoundary: boolean) => {
+        const onError = jest.fn();
```
just-boris, line 58: Who is going to reset this spy function between tests?

src/app-layout/__tests__/widget-areas-error-boundaries.test.tsx
```diff
@@ -0,0 +1,244 @@
...
+
+  const expectInvisibleErrorBoundary = (errorBoundaryWrapper: ErrorBoundaryWrapper) => {
+    expect(errorBoundaryWrapper.getElement()).toBeInTheDocument();
+    expect(errorBoundaryWrapper.findHeader()).toBeFalsy();
+    expect(errorBoundaryWrapper.findDescription()).toBeFalsy();
+    expect(errorBoundaryWrapper.findAction()).toBeFalsy();
+  };
+
+  describeEachAppLayout({ themes: ['refresh-toolbar'] }, () => {
+    describe.each([true, false])(
```
just-boris, line 55: This with and without error boundary repetition is not needed. 

You only need `wrappedWithErrorBoundary = true` scenario, where you assert that the root error boundary was reached when expected and was not reached when not expected

src/app-layout/__tests__/widget-areas-error-boundaries.test.tsx
```diff
@@ -0,0 +1,244 @@
...
+          const { errorBoundaryWrapper } = renderComponent(
+            <AppLayoutWrapper {...(appLayoutWrapperProps as any)}>
+              <AppLayout />
+            </AppLayoutWrapper>
+          );
+
+          expectInvisibleErrorBoundary(errorBoundaryWrapper);
+          expectErrorCallbacksToBeCalled();
+        });
+      }
```
just-boris, line 281: More tests and assertions needed
1. All tests needs to ensure that the `content` slot is intact after errors happen
    * Maybe also check other combinations like "error in navigation does not break tools", "error in breadcrumbs does not break navigation", etc
3. We need to have a test for `<AppLayout content={<ThrowsError />}>`. In this case the app layout crashes completely, only the wrapper error boundary is triggered

src/app-layout/visual-refresh-toolbar/drawer/global-drawer.tsx
```diff
@@ -112,104 +113,106 @@ function AppLayoutGlobalDrawerImplementation({
...
-                    className={testutilStyles['drawers-slider']}
-                    ariaLabel={activeGlobalDrawer?.ariaLabels?.resizeHandle}
-                    tooltipText={activeGlobalDrawer?.ariaLabels?.resizeHandleTooltipText}
-                    ariaValuenow={resizeProps.relativeSize}
-                    onKeyDown={resizeProps.onKeyDown}
-                    onDirectionClick={resizeProps.onDirectionClick}
-                    onPointerDown={resizeProps.onPointerDown}
-                  />
-                </div>
+          <AppLayoutBuiltInErrorBoundary>
```
just-boris: Why is it inside `<Transition>` component and not outside?

There could be some problems because `drawerRef` becomes undefined in case of error and the `<Transition>` component might be unprepared for this and crash as a result

Same question for all such code patterns

src/app-layout/visual-refresh-toolbar/widget-areas/bottom-content-slot.tsx
```diff
@@ -18,7 +19,7 @@ export const BottomContentSlotImplementation = ({ appLayoutState, appLayoutProps
   const { splitPanelPosition, placement, activeGlobalBottomDrawerId, bottomDrawerReportedSize, isMobile } =
     appLayoutState.widgetizedState;
   return (
-    <>
+    <AppLayoutBuiltInErrorBoundary>
```
just-boris: This is not enough, because the code above is outside the boundary. if there will be any undefined value, the page will crash

Expected implementation is how you did `AppLayoutStateProvider`. Zero code outside the boundary

You can create a reusable utility for it

```js
function withErrorBoundary(Component) {
    return function WithBoundaryComponent() {
        return <AppLayoutBuiltInErrorBoundary><Component /></AppLayoutBuiltInErrorBoundary>
    }
}
```

src/error-boundary/internal.tsx
```diff
@@ -87,6 +87,37 @@ export function BuiltInErrorBoundary({ wrapper, suppressNested = false, children
...
+export function AppLayoutBuiltInErrorBoundary({
+  wrapper,
+  suppressNested = false,
+  children,
+  renderFallback = () => <></>,
+}: AppLayoutBuiltInErrorBoundaryProps) {
+  const context = useContext(ErrorBoundariesContext);
+  const thisSuppressed = context.suppressed === true || context.suppressed === RootSuppressed;
+  const nextSuppressed = suppressNested || thisSuppressed;
+  return !thisSuppressed ? (
```
just-boris: Who is enabling this suppressing condition? Us or the consuming teams?

The in-widget boundaries should be only controlled by us and only emit metrics to our telemetry sink

src/error-boundary/internal.tsx
```diff
@@ -87,6 +87,37 @@ export function BuiltInErrorBoundary({ wrapper, suppressNested = false, children
   );
 }
 
+export function AppLayoutBuiltInErrorBoundary({
+  wrapper,
+  suppressNested = false,
+  children,
+  renderFallback = () => <></>,
```
just-boris, line 94: Is this argument ever used? I cannot find an example

src/error-boundary/internal.tsx
just-boris: What are these global listeners? Which telemetry sink do they forward events to?

src/error-boundary/internal.tsx
```diff
@@ -87,6 +87,37 @@ export function BuiltInErrorBoundary({ wrapper, suppressNested = false, children
...
+}: AppLayoutBuiltInErrorBoundaryProps) {
+  const context = useContext(ErrorBoundariesContext);
+  const thisSuppressed = context.suppressed === true || context.suppressed === RootSuppressed;
+  const nextSuppressed = suppressNested || thisSuppressed;
+  return !thisSuppressed ? (
+    <ErrorBoundaryImpl
+      {...context}
+      wrapper={wrapper}
+      renderFallback={renderFallback}
+      className={styles['app-layout-part-fallback']}
```
just-boris, line 104: Does this class goes anywhere? Didn't we discuss that this component renders no DOM? What happened to that discussion?

pages/app-layout/with-error-boundaries.page.tsx
georgylobko: I added a drawer with a broken trigger to test how the error boundary handles it

pages/app-layout/with-error-boundaries.page.tsx
georgylobko: removed

src/__a11y__/a11y-app-layout-toolbar.test.ts
georgylobko: Removed. I initially tested it with nested app layouts

src/app-layout/__tests__/widget-areas-error-boundaries.test.tsx
georgylobko: done

src/app-layout/__tests__/widget-areas-error-boundaries.test.tsx
georgylobko: This test verifies that errors don't bubble up to the top level and won't crash the entire page, even without a parent error boundary

pages/app-layout/with-error-boundaries.page.tsx
just-boris: Makes sense in hindsight, but it needs an explaining comment that all of this is intentional. Good case for `// @ts-expect-error` by the way

and same for all intentionally invalid content

src/__a11y__/a11y-app-layout-toolbar.test.ts
just-boris: Actually, this is a good idea for a test. That one crashing app layout does not break another

src/app-layout/__tests__/widget-areas-error-boundaries.test.tsx
just-boris: If you would like to test a simple `try/catch` logic, you do not need to test with with different types of handlers. 

Same here. You need to have one catch-all wrapper to run assertions.

Feel free to keep your code if you really want, but just so you know, this setup is redundant.

src/error-boundary/internal.tsx
georgylobko: discussed in DM

pages/app-layout/with-error-boundaries.page.tsx
georgylobko: good point. addressed

src/__a11y__/a11y-app-layout-toolbar.test.ts
georgylobko: yes, brought back nested app layouts

src/app-layout/__tests__/widget-areas-error-boundaries.test.tsx
georgylobko: added both

src/app-layout/visual-refresh-toolbar/drawer/global-drawer.tsx
georgylobko: brought it outside the transition component

src/app-layout/visual-refresh-toolbar/widget-areas/bottom-content-slot.tsx
georgylobko: done

just-boris:
Still see 3 unresolved discussions

Also forgot to call it on the previous review, but we need at least one integration test here. Because this is the only way to test against React 18 in our build

src/error-boundary/internal.tsx
georgylobko: Yes, it is used here: https://github.com/cloudscape-design/components/blob/974254f526ca83a957f2fab05e51a1e30a1d1259/src/app-layout/visual-refresh-toolbar/drawer/global-ai-drawer.tsx#L195

The empty div there instead of a react fragment ensures correct positioning for elements in the header

src/error-boundary/internal.tsx
georgylobko: > Who is enabling this suppressing condition? 

Consumers. This does not suppress in-widget EBs, but forwards suppress conditions down to the components tree. Refactored this code to make it more readable

src/error-boundary/internal.tsx
georgylobko: Yes, it applies display: contents to the EB root. Without this, components that depend on specific element ordering would break. This ensures they render correctly without adding extra styling.

https://github.com/cloudscape-design/components/blob/d11007edbdea2ed15a5338d0a490497c69670391/src/error-boundary/styles.scss#L11

src/error-boundary/internal.tsx
just-boris: This comment and the one below cancel each other: https://github.com/cloudscape-design/components/pull/4296/changes#r2882872385

Fallback is a `<div>` by default which ensures correct positioning. However it receives `app-layout-part-fallback` with `display: contents` which breaks correct positioning so you need to add an extra element.

This chain "correct" -> "break" -> "restore" looks weird and error prone, but up to you how you are planning to maintain this code

PR #4299, opened 2026-02-26
chore: Make tests compatible with type changes in Webdriver

jperals:
### Description

In the latest version of Webdriver, the return type of `browser.getElementAttribute` changed from `Promise<string>` to `Promise<string | null>`, so we need to account for that.

This is necessary for https://github.com/cloudscape-design/browser-test-tools/pull/179

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/build-tools/tasks/package-json.js
+++ b/build-tools/tasks/package-json.js
@@ -24,6 +24,8 @@ function getComponentsExports() {
     // Plugin api
     './internal/plugins': './internal/plugins/index.js',
     './internal/plugins/widget': './internal/plugins/widget.js',
+    // Public plugin api
+    './plugins': './plugins/index.js',
     // Public internal components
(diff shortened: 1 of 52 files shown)


src/cards/__integ__/selection.test.ts
```diff
@@ -11,11 +11,6 @@ class CardsPage extends BasePageObject {
   isSelectionInputFocused = (index: number) => {
     return this.isFocused(this.wrapper.findItems().get(index).find('input').toSelector());
   };
-
-  getActiveElement = async () => {
```
jperals, line 15: This function was also giving me issues because the return type of `browser.activeElement` does not correspond anymore to the input type of `browser.$`, but it is not used, so I just removed it.

PR #4304, opened 2026-02-26
fix: Replace raw "in" operator checks with type guards in property-filter

Who-is-PS:
### Description

Replaces raw `'prop' in obj` checks in the property-filter component with type guard functions that use `keyof` for compile-time validation of discriminant property names.

This is a COE action item `AWSUI-59006` from the (YpPBAPz2BrqF) where a type change during refactoring silently broke a `'propertyKey' in newToken` check because TypeScript doesn't validate the `in` operator — `'banana' in obj` compiles without error.

__Changes:__

- Added `isInternalToken()` and `isInternalTokenGroup()` type guards to `interfaces.ts`
- Added `isAbstractToken()` type guard to `utils.ts` for the generic `tokenGroupToTokens` helper
- Updated `isGroup()` in `filter-options.ts` to use `keyof`
- Added inline `isTokenGroup()` in `internal.tsx` for public `Token | TokenGroup` discrimination
- Replaced all raw `in` checks in `token.tsx` and `controller.ts`

__Not changed:__ `'keepOpenOnSelect' in option` (Category 2 — runtime property not in TS type, will get an `eslint-disable` comment in a follow-up PR).

No behavioral changes. Pure refactor for type safety.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/property-filter/utils.ts
+++ b/src/property-filter/utils.ts
@@ -10,10 +10,6 @@ import {
   Token,
 } from './interfaces';
 
-// Type guards for InternalToken | InternalTokenGroup union discrimination.
-// Using keyof ensures compile-time validation of the discriminant property name.
-// See AWSUI-59006 for context.
-
 export function isInternalToken(tokenOrGroup: InternalToken | InternalTokenGroup): tokenOrGroup is InternalToken {
   const key: keyof InternalToken = 'operator';
   return key in tokenOrGroup;


src/property-filter/interfaces.ts
```diff
@@ -416,3 +416,19 @@ export type ParsedText =
   | { step: 'property'; property: InternalFilteringProperty; operator: ComparisonOperator; value: string }
   | { step: 'operator'; property: InternalFilteringProperty; operatorPrefix: string }
   | { step: 'free-text'; operator?: ComparisonOperator; value: string };
+
+// Type guards for InternalToken | InternalTokenGroup union discrimination.
```
pan-kot: We can add a custom comment to the eslint rule is necessary - I don't think this comment here is needed.

src/property-filter/interfaces.ts
```diff
@@ -416,3 +416,19 @@ export type ParsedText =
...
+// Using keyof ensures compile-time validation of the discriminant property name.
+// See AWSUI-59006 for context.
+
+export function isInternalToken(tokenOrGroup: InternalToken | InternalTokenGroup): tokenOrGroup is InternalToken {
```
pan-kot: why do we define these in the interfaces file?

pan-kot: these are also similar checks defined in other files - can we create them all inside utils instead?

src/property-filter/interfaces.ts
Who-is-PS: I removed the comment.

src/property-filter/interfaces.ts
Who-is-PS: Good point, they should not be in the interfaces file. I moved them to utils.

PR #4305, opened 2026-02-26
fix: Replace raw "in" checks with type-safe alternatives in autosuggest and option utils

Who-is-PS:
### Description

Replaces raw `'prop' in obj` checks with type-safe alternatives in autosuggest, internal option utilities, and select validation.

Part of COE action item `AWSUI-59006` and Tech Proposal. (k9wvA3eNMiZ9).

### Changes

__Removed redundant `in` checks__ (4 occurrences):

- `src/autosuggest/utils/utils.ts` — `'type' in option && option.type === 'parent'` → `option.type === 'parent'`

- `src/autosuggest/autosuggest-option.tsx` — 3× `'type' in option && option.type === X` → `option.type === X`

  `type` is already declared as an optional property on `AutosuggestItem`, so direct property access is equivalent and safer (`undefined === 'parent'` → `false`).

__Applied `keyof` pattern__ (2 occurrences):

- `src/autosuggest/options-controller.ts` — `'options' in optionOrGroup` → `keyof AutosuggestProps.OptionGroup`

- `src/internal/components/option/utils/filter-options.ts` — `'options' in option` → `keyof OptionGroup`

  These discriminate `Option` from `OptionGroup`. The `keyof` pattern ensures the string stays in sync with the type at compile time.

__Annotated Category 2 check__ (1 occurrence):

- `src/select/utils/check-option-value-field.ts` — `'options' in element || 'value' in element` left as-is with explanatory comment. `element` is typed as `any`, so compile-time `keyof` validation is not possible

Related links, issue #, if available: `AWSUI-59006` and Tech Proposal. (k9wvA3eNMiZ9).

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/property-filter/controller.ts
+++ b/src/property-filter/controller.ts
@@ -14,7 +14,6 @@ import {
   InternalQuery,
   InternalToken,
   InternalTokenGroup,
-  isInternalToken,
   JoinOperation,
   ParsedText,
   Query,
@@ -22,6 +21,7 @@ import {
   TokenGroup,
 } from './interfaces';
 import {
+  isInternalToken,
   matchFilteringProperty,
   matchOperator,
   matchOperatorPrefix,
--- a/src/property-filter/interfaces.ts
+++ b/src/property-filter/interfaces.ts
@@ -416,19 +416,3 @@ export type ParsedText =
   | { step: 'property'; property: InternalFilteringProperty; operator: ComparisonOperator; value: string }
   | { step: 'operator'; property: InternalFilteringProperty; operatorPrefix: string }
   | { step: 'free-text'; operator?: ComparisonOperator; value: string };
-
-// Type guards for InternalToken | InternalTokenGroup union discrimination.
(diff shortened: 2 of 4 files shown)


src/property-filter/internal.tsx
```diff
@@ -161,7 +161,11 @@ const PropertyFilterInternal = React.forwardRef(
         tokenOrGroup: Token | TokenGroup,
         standaloneIndex?: number
       ): InternalToken | InternalTokenGroup {
-        return 'operation' in tokenOrGroup
+        const isTokenGroup = (t: Token | TokenGroup): t is TokenGroup => {
```
jperals, line 164: Not a blocker, but note that this function is within another function which is already within another function.

Everywhere else we have such functions (`isGroup`, `isInternalToken`, `isInternalTokenGroup`, `isInternalTokenGroup`) defined at the root level.

PR #4306, opened 2026-02-26
fix: Replace raw `in` operator checks with `keyof`-validated type guards

Who-is-PS:
### Description

Replaces raw `'prop' in obj` checks with `keyof`-validated alternatives across 6 components to prevent a class of bugs where TypeScript silently accepts invalid property name strings in `in` expressions.

This is a COE action item AWSUI-59006 from the property filter bug (YpPBAPz2BrqF) where a type change during refactoring silently broke a `'propertyKey' in newToken` check because TypeScript doesn't validate the `in` operator — `'banana' in obj` compiles without error.

### Changes:

- Added `isStackableItem()` and `isRefObject()` type guards to `flashbar/utils.ts`
- Updated `collapsible-flashbar.tsx` to use `isStackableItem()` (2 checks) and `isRefObject()` (1 check)
- Updated `non-collapsible-flashbar.tsx` to use `isRefObject()` (1 check)
- Updated `mixed-line-bar-chart/utils.ts` to use `keyof` in `isYThreshold()` and `isXThreshold()` (2 checks)
- Updated `side-navigation/util.tsx` to use inline `keyof` variables for `'href'` and `'items'` checks (2 checks)
- Updated `top-navigation/parts/utility.tsx` to use inline `keyof` for `'items'` check (1 check)
- Updated `button-dropdown/internal.tsx` to use inline `keyof` for `'items'` and `'badge'` checks (2 checks)
- Updated `button-group/item-element.tsx` to use inline `keyof` for `'popoverFeedback'` check (1 check)

12 checks across 7 files, 6 components. No behavioral changes. Pure refactor for type safety.

Related links, issue #, if available: `AWSUI-59006` and Tech Proposal. (k9wvA3eNMiZ9).

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/autosuggest/utils/utils.ts
```diff
@@ -5,7 +5,7 @@ import { AutosuggestItem } from '../interfaces';
 type SearchableFields = 'value' | 'label' | 'description' | 'labelTag';
 type SearchableTagFields = 'tags' | 'filteringTags';
 
-const isGroup = (option: AutosuggestItem) => 'type' in option && option.type === 'parent';
+const isGroup = (option: AutosuggestItem) => option.type === 'parent';
```
georgylobko: the `type` is optional

```suggestion
const isGroup = (option: AutosuggestItem) => option?.type === 'parent';
```

src/autosuggest/autosuggest-option.tsx
```diff
@@ -108,9 +108,9 @@ const AutosuggestOption = (
   ref: React.Ref<HTMLDivElement>
 ) => {
   const baseProps = getBaseProps(rest);
-  const useEntered = 'type' in option && option.type === 'use-entered';
-  const isParent = 'type' in option && option.type === 'parent';
-  const isChild = 'type' in option && option.type === 'child';
+  const useEntered = option.type === 'use-entered';
```
georgylobko:
```suggestion
  const useEntered = option?.type === 'use-entered';
```

georgylobko: And the same for `isParent` and `isChild`

src/flashbar/utils.ts
```diff
@@ -13,6 +15,16 @@ export interface StackableItem extends FlashbarProps.MessageDefinition {
   collapsedIndex?: number;
 }
 
+export function isStackableItem(item: StackableItem | FlashbarProps.MessageDefinition): item is StackableItem {
+  const key: keyof StackableItem = 'expandedIndex';
+  return key in item;
+}
+
+export function isRefObject<T>(ref: React.Ref<T>): ref is React.RefObject<T> {
```
georgylobko: Should the ref argument be typed as any? When you call this function, you don't know the type of ref yet - that's exactly what this function is supposed to check

src/internal/components/option/utils/filter-options.ts
```diff
@@ -73,5 +73,7 @@ export const isInteractive = (option?: DropdownOption) => !!option && !option.di
 
 export const isGroupInteractive = (option?: DropdownOption) => !!option && !option.disabled;
 
-export const isGroup = (option?: OptionDefinition | OptionGroup): option is OptionGroup =>
-  !!option && 'options' in option && !!option.options;
+export const isGroup = (option?: OptionDefinition | OptionGroup): option is OptionGroup => {
+  const key: keyof OptionGroup = 'options';
```
georgylobko, line 77: I'm not following what is fundamentally different in this function after refactoring. Could you add a u test that fails before the refactoring and passes after?

src/autosuggest/utils/utils.ts
Who-is-PS: Updated with optional chaining

src/autosuggest/autosuggest-option.tsx
Who-is-PS: Done, applied `?.` to all three.

src/flashbar/utils.ts
Who-is-PS: Changed to `unknown` instead of `any` since we already do `!== null && typeof ref === 'object'` runtime checks inside.

src/internal/components/option/utils/filter-options.ts
Who-is-PS: The runtime behavior is identical — no unit test can distinguish before/after because both produce the same output for every input.

The difference is at __compile time__. If someone renames `options` to `items` on `OptionGroup`:

- __Before:__ `'options' in option` — compiles fine, silently returns `false` for all groups (the exact COE bug pattern)
- __After:__ `const key: keyof OptionGroup = 'options'` — __fails to compile__: *Type '"options"' is not assignable to type '"items" | ...'*

PR #4310, opened 2026-02-27
fix: Improve live region behavior for short-lived content updates

avinashbot:
### Description

One of the common use-cases for the live region component is for announcing content loading/refreshing changes. Something like "Reloading things" -> "Last updated: <XXX>" (see [loading and refreshing](https://cloudscape.design/patterns/general/loading-and-refreshing/)). However, the live region deduplicates and throttles updates, which means that this information may not be properly conveyed under certain circumstances.

- (Refresh button clicked)
- Annoucement: `Loading issues` (suppressed if loading takes < 2 seconds)
- Announcement: `Last updated: December 21, 2023, 14:29` (announced)
- (Refresh button clicked again)
- Annoucement: `Loading issues` (suppressed if loading takes < 2 seconds)
- Announcement: `Last updated: December 21, 2023, 14:29` (suppressed since the text is the same as the last announcement, **this is bad**)

So I updated the controller to force an announcement even if the text is the same if there was an update in between (like the text `Loading issues`). Also reduced the delay to 1 second to make the announcements snappier (but I don't feel comfortable going any lower than that in case the live region is used in conjuction with user input)

Related links, issue #, if available: AWSUI-61749, AWSUI-61345

### How has this been tested?

Updated unit tests.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/button/loading-announcements.page.tsx
+++ b/pages/button/loading-announcements.page.tsx
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 import React, { useEffect, useState } from 'react';
 
-import { Input, LiveRegion } from '~components';
+import { FormField, Input, LiveRegion } from '~components';
 import Button from '~components/button';
 
 export default function ButtonIntegrationPage() {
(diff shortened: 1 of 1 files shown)


gethinwebster:
Added a demo page that I was playing around with locally. It seems like the fix causes the "Loading" message to take priority over "Loading complete" (or at least something prevents "Loading complete" from being announced, which I think would be optimal here)

src/live-region/controller.ts
```diff
@@ -48,16 +49,26 @@ export class LiveRegionController {
     }
   }
 
-  announce({ message, forceReannounce = false }: { message?: string; delay?: number; forceReannounce?: boolean }) {
```
avinashbot, line 51: Realized while refactoring that the delay parameter here actually does nothing. I removed it from `src/live-region/internal.tsx` (and moved it to the class constructor).

PR #4314, opened 2026-02-27
feat: action card

SpyZzey:
### Description
This pr adds the new action card component, including ActionCardWrapper, unit tests, dev pages.

Related links, issue #, if available:  [hiEUAPCbkCZS], [L0ZFA09lhlzj], [D347816619]

### How has this been tested?
- Manual tests
- Unit tests
- A11y tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---


pages/action-card/padding-permutations.page.tsx
```diff
@@ -0,0 +1,71 @@
...
+    disabled: [true],
+    disableHeaderPaddings: [true],
+    disableContentPaddings: [true],
+    onClick: [onClick],
+  },
+]);
+
+export default function ActionCardPaddingPermutations() {
+  return (
+    <>
```
amanabiy: [Not blocking] consider using SimplePage here

src/action-card/interfaces.ts
```diff
@@ -0,0 +1,160 @@
...
+  ariaLabel?: string;
+
+  /**
+   * Adds an aria-describedby reference for the action card.
+   */
+  ariaDescribedby?: string;
+
+  /**
+   * Determines whether the action card is disabled.
+   * @default false
```
mxschll: We don't use `@default` I think

src/action-card/interfaces.ts
```diff
@@ -0,0 +1,160 @@
...
+   * We do not support using this attribute to apply custom styling.
+   *
+   * @awsuiSystem core
+   */
+  nativeButtonAttributes?: NativeAttributes<React.ButtonHTMLAttributes<HTMLButtonElement>>;
+}
+
+export namespace ActionCardProps {
+  // eslint-disable-next-line @typescript-eslint/no-empty-object-type
+  export interface ClickDetail {}
```
mxschll, line 88: Why do we need this?

src/action-card/interfaces.ts
SpyZzey: I think that's something we should do. This will help JSDoc to determine default values. We do that in other places too (e.g., date-picker, prompt-input).

src/action-card/interfaces.ts
SpyZzey: It's for backwards compatibility. If we decide to add ClickDetail properties in the future, we'd need to change the type of onClick, e.g., from `CancelableEventHandler` to `CancelableEventHandler<ActionCardProps.ClickDetail>`. Though CancelableEventHandler is internal, it's still imported by builders.

pages/action-card/padding-permutations.page.tsx
```diff
@@ -0,0 +1,68 @@
...
+import React from 'react';
+
+import ActionCard, { ActionCardProps } from '~components/action-card';
+
+import { PermutationsPage } from '../app/templates';
+import createPermutations from '../utils/permutations';
+import PermutationsView from '../utils/permutations-view';
+import { longHeader, onClick, reactNodeContent, shortContent, shortDescription } from './common';
+
+/* Disable-padding permutations for header and content */
```
pan-kot: What does this comment mean?

pan-kot: Ah, I got it now.

I think I was a bit surprised because: 1) this comment uses a different format from the rest of comments in the doc; 2) the comment suggests there are only permutations with disabled paddings, while the file name suggests there are all combinations.

Do we even need such comments if the info is already conveyed by the file names?

pages/action-card/permutations.page.tsx
```diff
@@ -0,0 +1,108 @@
...
+  },
+]);
+
+export default function ActionCardPermutations() {
+  return (
+    <PermutationsPage title="Action card permutations" i18n={{}}>
+      <PermutationsView
+        permutations={permutations}
+        render={permutation => (
+          <div style={{ maxInlineSize: '400px' }}>
```
pan-kot: Why do we need it for permutations?

src/action-card/__tests__/action-card.test.tsx
```diff
@@ -0,0 +1,339 @@
...
+    });
+  });
+
+  describe('header', () => {
+    test('renders header text when provided', () => {
+      const wrapper = renderActionCard({ header: 'Test Header' });
+      expect(wrapper.findHeader()!.getElement()).toHaveTextContent('Test Header');
+    });
+
+    test('renders header as ReactNode', () => {
```
pan-kot, line 28: Do we really need the 'renders header text when provided' test  when we have this one?

pan-kot: same for description and content tests

src/action-card/__tests__/action-card.test.tsx
```diff
@@ -0,0 +1,339 @@
...
+
+  describe('onClick', () => {
+    test('calls onClick when clicked', () => {
+      const onClickSpy = jest.fn();
+      const wrapper = renderActionCard({ onClick: onClickSpy });
+      wrapper.click();
+      expect(onClickSpy).toHaveBeenCalledTimes(1);
+    });
+
+    test('fires onClick exactly once per click', () => {
```
pan-kot, line 104: how is it different from 'calls onClick when clicked' test?

src/action-card/__tests__/action-card.test.tsx
```diff
@@ -0,0 +1,339 @@
...
+      const onClickSpy = jest.fn();
+      const wrapper = renderActionCard({ onClick: onClickSpy });
+      wrapper.click();
+      expect(onClickSpy).toHaveBeenCalledTimes(1);
+    });
+
+    test('does not call onClick when event.defaultPrevented', () => {
```
pan-kot: What do we need this test for? Should we make the onClick event cancelable instead?

src/action-card/__tests__/action-card.test.tsx
```diff
@@ -0,0 +1,339 @@
...
+  });
+
+  describe('focus management', () => {
+    test('can be focused through the ref API', () => {
+      let actionCard: ActionCardProps.Ref | null = null;
+      const renderResult = render(<ActionCard ref={el => (actionCard = el)} />);
+      const wrapper = createWrapper(renderResult.container);
+      actionCard!.focus();
+      const button = wrapper.findActionCard()!.getElement().querySelector('button')!;
+      expect(document.activeElement).toBe(button);
```
pan-kot: nit: there is a toHaveFocus() assertion that can be used here instead.

src/action-card/__tests__/action-card.test.tsx
```diff
@@ -0,0 +1,339 @@
...
+    test('renders with embedded variant', () => {
+      const wrapper = renderActionCard({ variant: 'embedded', header: 'Header' });
+      expect(wrapper.getElement()).toHaveClass(styles['variant-embedded']);
+    });
+  });
+
+  describe('iconVerticalAlignment', () => {
+    test('renders icon with iconVerticalAlignment=top', () => {
+      const wrapper = renderActionCard({ icon: <span>icon</span>, iconVerticalAlignment: 'top', header: 'Header' });
+      expect(wrapper.findIcon()).not.toBeNull();
```
pan-kot: I think this test does not assert anything, really. For such placement variants the screenshot tests will be enough.

src/action-card/__tests__/action-card.test.tsx
```diff
@@ -0,0 +1,339 @@
...
+    test('applies disabled styling to header and description', () => {
+      const wrapper = renderActionCard({ header: 'Header', description: 'Description', disabled: true });
+      expect(wrapper.findHeader()).not.toBeNull();
+      expect(wrapper.findDescription()).not.toBeNull();
+      expect(wrapper.isDisabled()).toBe(true);
+    });
+  });
+
+  describe('style prop', () => {
+    test('applies all style properties to elements', () => {
```
pan-kot: what this test asserts?

src/action-card/test-classes/styles.scss
```diff
@@ -0,0 +1,24 @@
+/*
+ Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ SPDX-License-Identifier: Apache-2.0
+*/
+
+.header {
```
pan-kot: nit: we can use `.header,.description,.body,.icon,.disabled { /* used in test-utils* / }` here for brevity - I think that is what we use most often for the test-classes.

src/action-card/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+      disableHeaderPaddings = false,
+      disableContentPaddings = false,
+      iconVerticalAlignment = 'top',
+      variant = 'default',
+      ...props
+    }: ActionCardProps,
+    ref: React.Ref<ActionCardProps.Ref>
+  ) => {
+    const baseComponentProps = useBaseComponent('ActionCard', {
+      props: {
```
pan-kot, line 27: nit: we can consider including metadata props such as hasHeader, hasDescritption, hasContent, hasIcon

src/action-card/interfaces.ts
pan-kot: We use Documenter to extract the default value automatically from the component definition inside index.ts. I suggest we remove this annotation from here as using it is not a part of our dev conventions as of yet. It does improve the experience for TS users - but we do not have mechanism to ensure it matches the actual default value of the component.

src/action-card/interfaces.ts
pan-kot: The ActionCard is basically a button - should we make the event handling consistent with what we use in the button component?

src/action-card/internal.tsx
```diff
@@ -0,0 +1,210 @@
...
+    const descriptionId = useUniqueId('action-card-description-');
+    const bodyId = useUniqueId('action-card-body-');
+
+    useImperativeHandle(ref, () => ({
+      focus: () => {
+        buttonRef.current?.focus();
+      },
+    }));
+
+    if (isDevelopment) {
```
pan-kot: The warnOnce already has a check for dev environment inside - we don't need a separate one.

src/action-card/internal.tsx
```diff
@@ -0,0 +1,210 @@
...
+  ) => {
+    const baseProps = getBaseProps(rest);
+    const buttonRef = useRef<HTMLButtonElement>(null);
+    const rootRef = useRef<HTMLDivElement>(null);
+    const headerId = useUniqueId('action-card-header-');
+    const standaloneButtonId = useUniqueId('action-card-button-');
+    const descriptionId = useUniqueId('action-card-description-');
+    const bodyId = useUniqueId('action-card-body-');
+
+    useImperativeHandle(ref, () => ({
```
pan-kot: Can we use the useForwardFocus util here?

src/action-card/internal.tsx
```diff
@@ -0,0 +1,210 @@
...
+        warnOnce(
+          'ActionCard',
+          'An accessible name is required. Provide either a `header` or an `ariaLabel` prop so the action card has a meaningful label for screen reader users.'
+        );
+      }
+    }
+
+    const handleButtonClick = (event: React.MouseEvent<HTMLButtonElement>) => {
+      event.stopPropagation();
+      nativeButtonAttributes?.onClick?.(event);
```
pan-kot: Should we be firing the native onClick when the action card is disabled?

src/action-card/internal.tsx
```diff
@@ -0,0 +1,210 @@
...
+        return;
+      }
+      if (disabled) {
+        return event.preventDefault();
+      }
+      fireCancelableEvent(onClick, {}, event);
+    };
+
+    // We are adding a root click so we can have test-utils .click on the wrapper itself.
+    const handleRootClick = () => {
```
pan-kot: Can we solve it by overriding the click() even in the action card wrapper? It feels wrong to write special code here to support the test-utils.

src/action-card/internal.tsx
```diff
@@ -0,0 +1,210 @@
...
+      ...nativeButtonAttributes,
+      type: 'button',
+      className: clsx(
+        styles['header-button'],
+        disabled && styles.disabled,
+        variant && styles[`variant-${variant}`],
+        nativeButtonAttributes?.className
+      ),
+      onClick: handleButtonClick,
+      'aria-describedby': ariaDescribedby || (description && (ariaLabel || header) ? descriptionId : undefined),
```
pan-kot: Why this requires both description and ariaLabel/header to set the descriptionId?

src/action-card/styles.scss
```diff
@@ -0,0 +1,235 @@
...
+  &.disabled {
+    color: map.get(constants.$states, 'disabled', 'text-color');
+  }
+}
+
+.icon {
+  flex-shrink: 0;
+  display: flex;
+  color: map.get(constants.$states, 'default', 'icon-color');
+  pointer-events: none;
```
pan-kot, line 55: Why do we need to disabled pointer events here?

pages/action-card/permutations.page.tsx
SpyZzey: I added it to limit the size, so that it doesn't expand to the whole page. Potentially, we can remove it the screenshot tests can capture the sizes in different viewport widths.

pages/action-card/padding-permutations.page.tsx
SpyZzey: Will remove it.

src/action-card/internal.tsx
SpyZzey: If there is no header, the "standaloneButton"/"overlay-button" is used, which will set the description as aria-labelledby. Thus, it would announce the description twice, once as `aria-describedby` and once as `aria-labelledby`.

src/action-card/styles.scss
SpyZzey: The icon adds its own cursor css value, which will overwrite the action-card cursor: pointer. We don't need any pointer-events on the icon, so I just removed them.

src/action-card/__tests__/action-card.test.tsx
SpyZzey: Good catch. I forgot to add the assertions here...
Added them.

src/action-card/internal.tsx
SpyZzey: Good catch, Moved the disabled call earlier.

src/action-card/__tests__/action-card.test.tsx
SpyZzey: Removed it

src/action-card/__tests__/action-card.test.tsx
SpyZzey: This test asserts that when the onClick from the native button uses `event.preventDefault()`, the onClick on the ActionCard won't fire.

SpyZzey: The onClick is already a cancelable event. Do you mean making it non-cancelable?

src/action-card/interfaces.ts
SpyZzey: @pan-kot I think we can potentially do that in the future, but keep it as is for now.

src/action-card/__tests__/action-card.test.tsx
pan-kot: No, I guess I just didn't understand the use case - and I still don't. Will that work on other direction - when calling preventDefault() on the top-level click event handler?

src/action-card/internal.tsx
pan-kot: This is useful info - I recommend documenting that in code, e.g.:

```
// ...explanation
if (!ariaDescribedby && description && (ariaLabel || header)) {
  ariaDescribedby = descriptionId;
}
//...
'aria-describedby': ariaDescribedby,
```

src/action-card/__tests__/action-card.test.tsx
SpyZzey: I expected nativeButtonAttributes.onClick to work in parallel with the component-level onClick, in which case 
I'd expect the component-level onClick to be cancelled when nativeButtonAttributes.onClick calls 
preventDefault() - I'm not sure if there is a use case for it, but that is how I'd expect it to behave if 
someone had an actual use case for that.

I have replaced the logic with the WithNativeAttributes component, which should take care of this. This is how button handles it.

pan-kot (approved):

PR #4315, opened 2026-02-27
feat: expose card component

SpyZzey:
### Description
This pr exposes our card component externally. 
Introduces the public ItemCard component in our public API, which exposes the previously internal Card component.
The internal Card is removed and replaced by ItemCard, including instances in the Cards component.

- Adds public ItemCard component
- Migrates Cards component to use `InternalItemCard` instead of `Card`,
- Adds test utils (ItemCardWrapper), unit tests, and dev pages for visual regression tests,

Related links, issue #, if available: hiEUAPCbkCZS

### How has this been tested?
- Manual tests,
- Unit tests,
- Screenshot tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/item-card/styles.scss
+++ b/src/item-card/styles.scss
@@ -63,35 +63,25 @@ $action-padding-horizontal: awsui.$space-xxs;
   position: relative;
   background-color: var(#{custom-props.$styleItemCardBackgroundDefault}, awsui.$color-background-card);
   min-inline-size: 0;
+  box-shadow: var(#{custom-props.$styleItemCardBoxShadowDefault}, awsui.$shadow-card);
 
   &:before {
     @include styles.base-pseudo-element;
     box-shadow: none;
     border-color: transparent;
+    border-block: solid var(#{custom-props.$styleItemCardBorderWidthDefault}, awsui.$border-width-card)
+      var(#{custom-props.$styleItemCardBorderColorDefault}, awsui.$color-border-card);
+    border-inline: solid var(#{custom-props.$styleItemCardBorderWidthDefault}, awsui.$border-width-card)
+      var(#{custom-props.$styleItemCardBorderColorDefault}, awsui.$color-border-card);
   }
 
   &::after {
     @include styles.base-pseudo-element;
   }
 
-  &:not(.refresh) {
-    &::before {
-      border-block-start: awsui.$border-container-top-width solid awsui.$color-border-container-top;
-    }
-    &::after {
-      box-shadow: awsui.$shadow-container;
-    }
-  }
-
-  &.refresh {
-    box-shadow: var(#{custom-props.$styleItemCardBoxShadowDefault}, awsui.$shadow-card);
-
-    &:before {
-      border-block: solid var(#{custom-props.$styleItemCardBorderWidthDefault}, awsui.$border-width-card)
-        var(#{custom-props.$styleItemCardBorderColorDefault}, awsui.$color-border-card);
-      border-inline: solid var(#{custom-props.$styleItemCardBorderWidthDefault}, awsui.$border-width-card)
-        var(#{custom-props.$styleItemCardBorderColorDefault}, awsui.$color-border-card);
-    }
+  &:not(.refresh)::before {
+    border-block-start: solid var(#{custom-props.$styleItemCardBorderWidthDefault}, awsui.$border-container-top-width)
+      var(#{custom-props.$styleItemCardBorderColorDefault}, awsui.$color-border-container-top);
   }
 
   &.highlighted {


src/test-utils/dom/item-card/index.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  /**
+   * Finds the header slot of the card.
+   */
+  findHeader(): ElementWrapper | null {
+    return this.findByClassName(styles.header);
+  }
+
+  /**
+   * Finds the footer slot of the card
```
jperals: Missing dot at the end

pages/item-card/header-permutations.page.tsx
```diff
@@ -0,0 +1,88 @@
...
+    footer: [shortFooter, longFooter],
+  },
+  // With custom paddings
+  {
+    children: [reactNodeContent],
+    disableHeaderPaddings: [true],
+    disableContentPaddings: [true],
+    disableFooterPaddings: [true],
+  },
+  // Variant permutations
```
jperals: Can we test the `embedded` variant in combination with actions and icons?

SpyZzey: Good point, I will add that

pages/item-card/header-permutations.page.tsx
```diff
@@ -0,0 +1,88 @@
...
+  },
+  // Without header but with description
+  {
+    header: [undefined],
+    children: [longContent],
+    description: [longDescription],
+    actions: [undefined, actions],
+    iconName: [undefined],
+  },
+  // With footer
```
jperals: The permutations defined after this point on this page do not relate much to the card header. Would they better belong to the other page `permutations.page.tsx`?

pages/item-card/style-custom.page.tsx
```diff
@@ -0,0 +1,152 @@
...
+import { SpaceBetween } from '~components';
+import ItemCard from '~components/item-card';
+
+import ScreenshotArea from '../utils/screenshot-area';
+
+export default function CustomCard() {
+  const background = 'light-dark(#ffe1e1, rgb(40, 0, 0))';
+
+  return (
+    <ScreenshotArea>
```
jperals: The h1 header doesn't need to be in the screenshot area.

Why actually not use the CardPage, and PermutationsView?

SpyZzey: The CardPage shows everything vertically, which is good for the permutations. But I think we can just take a screenshot of the whole screenshot area here. We do similar things in other components (e.g., container, flashbar, toggle, ...). I have updated the page to use the SimplePage instead, but I don't think permutations are necessary here (also adds more delay to our visual regression tests).

pages/item-card/header-permutations.page.tsx
SpyZzey: Added with the variants permutations page

pages/item-card/header-permutations.page.tsx
SpyZzey: Removed header-permutations in favor for a more detailed test setup (general permutations, padding and variant permutations)

src/item-card/styles.scss
```diff
@@ -0,0 +1,161 @@
...
+    background-color: awsui.$color-background-item-selected;
+
+    &:before {
+      border-block: solid awsui.$border-width-card-highlighted awsui.$color-border-card-highlighted;
+      border-inline: solid awsui.$border-width-card-highlighted awsui.$color-border-card-highlighted;
+    }
+  }
+
+  > .inner-card {
+    overflow: hidden;
```
jperals: Not sure it's a great idea to hide overflowing content, especially as the current cards don't do this (nor does the container).

src/item-card/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const externalProps = getExternalProps(props);
+
+  return (
+    <InternalItemCard
+      disableHeaderPaddings={disableHeaderPaddings}
+      disableContentPaddings={disableContentPaddings}
+      disableFooterPaddings={disableFooterPaddings}
+      iconName={iconName}
+      variant={variant}
+      {...externalProps}
```
jperals, line 37: Does TypeScript still type-check the `externalProps` and `baseComponentProps` if we pass them destructured?

src/item-card/interfaces.ts
```diff
@@ -0,0 +1,183 @@
...
+
+  /**
+   * Specifies alternate text for a custom icon. We recommend that you provide this for accessibility.
+   * This property is ignored if you use a predefined icon or if you set your custom icon using the `iconSvg` slot.
+   */
+  iconAlt?: string;
+
+  /**
+   * Removes the default padding from the header area.
+   * @default false
```
jperals, line 45: These `@default` annotations are not currently used by Documenter. To display default values in the documentation, specify in the function signature:
```
{
  ...,
  disableHeaderPaddings = true,
  ...,
}
```

src/item-card/interfaces.ts
```diff
@@ -0,0 +1,183 @@
...
+
+  /**
+   * Main content of the card.
+   */
+  children?: React.ReactNode;
+
+  /**
+   * Displays an icon next to the content.
+   */
+  iconName?: IconProps.Name;
```
jperals: Why not an icon slot?

src/item-card/interfaces.ts
```diff
@@ -0,0 +1,183 @@
...
+   */
+  actions?: React.ReactNode;
+
+  /**
+   * Main content of the card.
+   */
+  children?: React.ReactNode;
+
+  /**
+   * Displays an icon next to the content.
```
jperals: This suggests that it is displayed next to the main content, but it is displayed next to the header, right?

src/item-card/styles.scss
```diff
@@ -0,0 +1,161 @@
...
+
+    display: flex;
+    flex-direction: column;
+    block-size: 100%;
+    inline-size: 100%;
+  }
+
+  // Generate variant styles from the configuration maps
+  @each $variant in ('embedded', 'default') {
```
jperals, line 122: This will create quite a bit of duplicated code. Can we include in the loop only the parts that depend on the variant? I think having the same selector mentioned more than once in a stylesheet is fine.

src/item-card/styles.scss
SpyZzey: We need to have overflow hidden for some scenarios (e.g., when displaying an image), where otherwise builders would need to adjust their code to match our border radius for cards. I think not having it would cause more friction - especially because we are changing the border radius depending on the variant.

 I don't see a scenario, where it something should be generated outside of the card that cannot be attached to the body instead (e.g., tooltip uses portals to be rendered on another level, dropdown components have 'expandToViewport' properties to get rendered on the page level).

src/item-card/interfaces.ts
SpyZzey: I think having `@default` makes sense from a JSDoc standpoint, but yes, for our documenter, I added it to the function props.

src/item-card/styles.scss
SpyZzey: Agree, moved some of the parts outside of the `@each`.

src/item-card/interfaces.ts
SpyZzey: Reached out to team, using icon slot instead.

src/item-card/index.tsx
SpyZzey: We pass them like this in other components too, don't we?

pages/item-card/common.tsx
```diff
@@ -64,4 +76,10 @@ export const actions = (
   />
 );
 
-export const icon = <Icon name="settings" />;
+export const icon = <Icon name="settings"></Icon>;
+
+export const iconSvg = (
```
jperals: Nit: this element doesn't seem to be used

src/item-card/styles.scss
jperals: Even if dropdown components have the `expandToViewport` property, there are probably current customers using them in the Cards component without it, and this CSS will change the dropdown's behavior, as dropdowns avoid growing beyond parents with hidden overflow. Replacing the button group with a button dropdown in the permutations reveals some issues:

Screenshot 2026-03-23 at 18 35 16

Screenshot 2026-03-23 at 18 33 28

Screenshot 2026-03-23 at 18 38 11

One possible way to work around this is to apply `overflow: hidden` only when `disableContentPadding` is set to `true` (since this property didn't exist before, this wouldn't break any existing customer), although it's not a very elegant solution, and remains an implicit behavior. We might also be able to constrain this behavior to the body area instead of the entire card. In any case, if we set `overflow: hidden` in any way to support cases like the image preview I think we should add visual regression test for this.

We can also think of adding a `media` prop to Card, pretty much like in Container (but I know this would need further discussion).

SpyZzey: Good catch! I changed it up to adjust the border radius for inner children. This will require builders to inherit the border-radius and set overflow: hidden themselves when using media.

pages/item-card/common.tsx
SpyZzey: Removed it.

src/item-card/styles.scss
SpyZzey: Updated it to require builders to use the design tokens for now. We might want to add a simpler approach as a follow-up.

src/test-utils/dom/item-card/index.ts
```diff
@@ -0,0 +1,52 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import { ComponentWrapper, ElementWrapper } from '@cloudscape-design/test-utils-core/dom';
+
+import structuredItemStyles from '../../../internal/components/structured-item/test-classes/styles.selectors.js';
+import styles from '../../../item-card/styles.selectors.js';
```
jperals, line 6: It is recommended to separate test-util classes from the actual styling CSS classes. You can do this by putting test-util classes into `src/<component-name>/test-classes/styles.scss`. Examples: https://github.com/cloudscape-design/components/pull/4345/changes#diff-92a138c364e1e87c5b03630ce578bd4d8f8fef4dbc95eeb555798fe1f1a4b1e2

After a component is already released this is not that easy to refactor.

style-dictionary/classic/shadows.ts
```diff
@@ -7,6 +7,11 @@ import { StyleDictionary } from '../utils/interfaces.js';
 import { tokens as parentTokens } from '../visual-refresh/shadows.js';
 
 const tokens: StyleDictionary.ShadowsDictionary = {
+  shadowCard: {
```
jperals: Until now, the cards in Classic were inheriting the shadow style from the container (see [lines 39 to 41 in the removed file `src/internal/components/card/styles.scss`](https://github.com/cloudscape-design/components/pull/4315/changes#diff-00ea818a9ccfb97ccb0ada293327e1707780d520997ec2f0bcc408516433947eL39-L41)):
```
  &:not(.refresh)::after {
    box-shadow: awsui.$shadow-container;
  }
```
So I think to keep backwards compatibility this should be `{shadowContainer}`.

src/item-card/__tests__/card.test.tsx
```diff
@@ -0,0 +1,357 @@
...
+});
+
+test('renders header when provided', () => {
+  const { itemCard } = renderItemCard(<ItemCard header="Card title">content</ItemCard>);
+  const header = itemCard.findHeader();
+  expect(header).not.toBeNull();
+  expect(header!.getElement()).toHaveTextContent('Card title');
+});
+
+test('renders header-inner element for header text', () => {
```
jperals: Why is this relevant? And why do we use the `header-inner` class in other tests too? This class doesn't seem relevant for customers, just for us to apply some style to the header. Customers should use `findHeader` to find and inspect the header.

src/item-card/__tests__/card.test.tsx
```diff
@@ -0,0 +1,357 @@
...
+    <ItemCard header="Title" actions={<button>Action</button>}>
+      content
+    </ItemCard>
+  );
+  const actions = itemCard.findActions();
+  expect(actions).not.toBeNull();
+  expect(actions!.getElement()).toHaveTextContent('Action');
+});
+
+test('renders header with with-actions class when actions are provided', () => {
```
jperals: Similar here, I would not assert presence of classes or other internal implementation details unless there is a good reason for it.

PR #4318, opened 2026-03-02
chore: improve modal custom dimensions dev warning

amanabiy:
### Description
This PR is a follow-up to [#4252](https://github.com/cloudscape-design/components/pull/4252) with the following improvements:
- Consolidate footer border radius styles: merge `--stuck` and `custom-height` into a single `--rounded` modifier
- Add dev warning when NaN is passed as width or height, so that builders know the width and the height are not set.

### How has this been tested?

- Added unit tests for NaN validation warnings
- Manually tested modal with custom dimensions
- Verified footer styling with stuck state and custom height

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/__tests__/snapshot-tests/__snapshots__/documenter.test.ts.snap
+++ b/src/__tests__/snapshot-tests/__snapshots__/documenter.test.ts.snap
@@ -17571,11 +17571,10 @@ of the most recent getModalRoot call as an argument.",
       "type": "((container: HTMLElement | null) => void)",
     },
     {
-      "description": "Sets the width of the modal. Defaults to \`medium\`.
-
-\`max\` uses variable width up to the largest size allowed by the design guidelines.
-Other sizes have fixed widths: \`small\` (320px), \`medium\` (600px), \`large\` (820px),
-\`x-large\` (1024px), \`xx-large\` (1280px).",
+      "defaultValue": "'medium'",
+      "description": "Sets the width of the modal. \`max\` uses variable width up to the
+largest size allowed by the design guidelines. Other sizes have fixed widths:
+\`small\` (320px), \`medium\` (600px), \`large\` (820px), \`x-large\` (1024px), \`xx-large\` (1280px).",
       "inlineType": {
         "name": "ModalProps.Size",
         "type": "union",
--- a/src/modal/index.tsx
+++ b/src/modal/index.tsx
@@ -3,7 +3,7 @@
 'use client';
 import React from 'react';
 
-import { useUniqueId, warnOnce } from '@cloudscape-design/component-toolkit/internal';
+import { useUniqueId } from '@cloudscape-design/component-toolkit/internal';
 
 import {
   AnalyticsFunnel,
@@ -70,12 +70,7 @@ function ModalWithAnalyticsFunnel({
   );
 }
 
-export default function Modal({ position = 'center', width, height, ...rest }: ModalProps) {
-  const { size = 'medium', ...props } = rest;
-  if (rest.size !== undefined && width !== undefined) {
-    warnOnce('Modal', 'Both `size` and `width` exist. `size` will not be used for width calculations.');
-  }
-
+export default function Modal({ size = 'medium', position = 'center', width, height, ...props }: ModalProps) {
   const { isInFunnel } = useFunnel();
   const analyticsMetadata = getAnalyticsMetadataProps(props as BasePropsWithAnalyticsMetadata);
   const baseComponentProps = useBaseComponent(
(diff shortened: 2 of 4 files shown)


src/__tests__/snapshot-tests/__snapshots__/documenter.test.ts.snap
```diff
@@ -17571,7 +17571,6 @@ of the most recent getModalRoot call as an argument.",
       "type": "((container: HTMLElement | null) => void)",
     },
     {
-      "defaultValue": "'medium'",
```
avinashbot, line 17574: This is unfortunate — the default size is still medium but is no longer documented. A less ideal option is to "manually" document the default in the description.

src/modal/index.tsx
```diff
@@ -70,7 +70,12 @@ function ModalWithAnalyticsFunnel({
   );
 }
 
-export default function Modal({ size = 'medium', position = 'center', width, height, ...props }: ModalProps) {
+export default function Modal({ position = 'center', width, height, ...rest }: ModalProps) {
+  const { size = 'medium', ...props } = rest;
+  if (hasSizeProp(rest) && width !== undefined) {
```
avinashbot: You don't have do create a "type guard" style function here, do you? `rest` is already typed to be `ModalProps`, so why not just:

```suggestion
  if (rest.size && width !== undefined) {
```

src/__tests__/snapshot-tests/__snapshots__/documenter.test.ts.snap
NathanZlion: I agree, ideally it would be great to keep the defaultValue from the Modal parameter value but I see how we can't do that because we need to get the user given value for the dev warning, before setting our default value. so I think the 'manual' way is okay,

src/modal/index.tsx
jperals: In that case wouldn't `rest.size !== undefined` be a better condition because it would include the case of `size` being zero (as much as it's a bad value to use)?

amanabiy: Agreed. The type guard is overkill here. Since size is typed as `'small' | 'medium' | 'large' | 'x-large' | 'xx-large' | 'max'`, it can't be zero or an empty string. Using `rest.size !== undefined` is semantically clearer so I will go with that.

src/__tests__/snapshot-tests/__snapshots__/documenter.test.ts.snap
amanabiy: I've added "Defaults to medium" to the size property documentation in the interface to document.

jperals: You can let this default value be in the typings directly by adding this to the API doc comment:
```
@default 'medium'
```
Example where we do this: https://github.com/cloudscape-design/components/blob/a9682ae3230edca7d71ed79d4939f157a1284b9b/src/date-picker/interfaces.ts#L78

avinashbot: @jperals But the documenter won't pick it up, right? You'll still need to put it in the description.

amanabiy: Yes, the documenter doesn't pick up the `@default`. In the example above, the default value is being fetched from the destructuring pattern in the index file here:

https://github.com/cloudscape-design/components/blob/a9682ae3230edca7d71ed79d4939f157a1284b9b/src/date-picker/index.tsx#L65

jperals: I think that's right, it used to pick it up but it doesn't anymore after recent changes.

amanabiy: I'd suggest removing the warning when `size` and `width` are used together. The behavior that `width` overrides `size` is already well-documented, and the website has always shown the default value for `medium`, which is generated from the component definition. Adding the warning would require us to remove the `defaultValue` from the component definition, which breaks consistency. For the sake of maintaining that consistency, I'd encourage us not to add the warning. What do you think? If we see the need in the future, we can add `@default` tag support to the documenter to handle cases like this.

jperals: I support that, I prefer to keep the documented default value than adding the new dev warning, especially given that the behavior when both `size` and `width` are defined is already documented, as you say.

jperals (approved):

PR #4331, opened 2026-03-04
fix: Fix ariaLabel support for date range picker

avinashbot:
### Description

Turns out the `ariaLabel` attribute doesn't actually work on the date range picker — because there's always an `aria-labelledby` coming from the trigger itself, which always overrides `aria-label`. Whoops. Not breaking any new ground with the implementation here; keeping the implementation consistent with what we do in the select component.

One thing I did change is moving `ariaLabel` out of `i18nStrings` and putting it in the top level. This is also consistent with what we did in other components; it was just missed in this one. `i18nStrings` are for properties that are always static, not situation-dependent.

Related links, issue #, if available: AWSUI-61769

### How has this been tested?

Added unit tests.

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/internal/components/button-trigger/index.tsx
```diff
@@ -28,7 +28,6 @@ export interface ButtonTriggerProps extends BaseComponentProps {
   inlineTokens?: boolean;
   ariaHasPopup?: 'true' | 'listbox' | 'dialog';
   ariaControls?: string;
-  ariaLabel?: string;
```
avinashbot, line 31: Not used anymore, since the select trigger also switched to the `ariaLabelledby`-only implementation.

PR #4332, opened 2026-03-04
fix: Add characterCountText to form field for improved screen reader support

avinashbot:
### Description

Apparently, NVDA/JAWS don't fare well when you update the accessible description of an input you're focused on after every keypress, and end up interrupting everything else, like the keypress announcements themselves.

This introduces a new string property, `characterCountText`, whose updates are debounced for screen readers. This also means that only the updated section of the constraint text ("Character count: x/y") is announced when it changes.

Related links, issue #, if available: AWSUI-61742

### How has this been tested?

Added unit tests and a test util method.

Provisioning a Windows machine right now, but I'll also let our accessibility specialists take a crack at it before merging. But in the meantime, the code itself is reviewable.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/form-field/internal.tsx
+++ b/src/form-field/internal.tsx
@@ -92,7 +92,7 @@ export function FormFieldWarning({ id, children, warningIconAriaLabel }: FormFie
   );
 }
 
-export function ConstraintTextArea({
+export function ConstraintTextRegion({
   id,
   hasValidationText,
   children,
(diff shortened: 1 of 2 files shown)


src/form-field/internal.tsx
```diff
@@ -88,7 +92,7 @@ export function FormFieldWarning({ id, children, warningIconAriaLabel }: FormFie
   );
 }
 
-export function ConstraintText({
+export function ConstraintTextArea({
```
ernst-dev: NP: Might make searching easier if it were named ConstraintTextRegion instead of ...TextArea in the case of looking for textareas

I believe we use Region in other places in the package for sections or blocks which are not slots

PR #4336, opened 2026-03-05
fix: apply gap only when both pagination && prefrences are present

amanabiy:
### Description
Fixes unnecessary gap in table tools header when pagination and preferences slots are empty.

When a table has only a filter (no pagination and preferences), an unwanted horizontal gap appeared after the filter due to the empty `tools-align-right` wrapper being rendered. The flex `gap` property applies spacing between all flex children, including empty elements. The `tools-align-right` wrapper was always rendered, even when both pagination and preferences were absent.

Large Screen
Before          |  After
:-------------------------:|:-------------------------:
image  |  image

Small Screen
Before          |  After
:-------------------------:|:-------------------------:
image |  image
Tested by:
- Running through pipeline and only expected changes are visible which is the above one

Related links, issue #, if available: AWSUI-61752

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/table/tools-header.tsx
+++ b/src/table/tools-header.tsx
@@ -28,6 +28,7 @@ export default function ToolsHeader({ header, filter, pagination, preferences, s
   }
   const isSmall = breakpoint === 'default';
   const hasTools = filter || pagination || preferences;
+  const hasRightAlignedTools = !!(pagination || preferences);
   return (
     <>
       {isHeaderString ? <span id={headingId}>{header}</span> : header}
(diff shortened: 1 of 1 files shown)


mxschll:
Do we have screenshot tests for that?

src/table/tools-header.tsx
```diff
@@ -42,18 +43,20 @@ export default function ToolsHeader({ header, filter, pagination, preferences, s
...
-            {preferences && (
-              <div className={styles['tools-preferences']} onClickCapture={() => setLastUserAction?.('preferences')}>
-                {preferences}
-              </div>
-            )}
-          </div>
+          {hasRightAlignedTools && (
+            <div className={styles['tools-align-right']}>
+              {pagination && (
+                <div className={styles['tools-pagination']} onClickCapture={() => setLastUserAction?.('pagination')}>
```
jperals, line 49: Note that this can lead to the test utils `findPagination` and `findCollectionPreferences` to return `null` in situations in which previously it was returning an element.

IMO this new behavior is actually better, but I recommend to first assess this and whether there is customer impact to mitigate.

amanabiy: Thank you for sharing. since both pagination, and preferences are rendered conditionally, and hasRightAligned is supposed to be present if either of the two is there, it shouldn't break any tests.

mxschll: Might be worth doing a dry-run to live before merging.

amanabiy: Already done. I will send it to you and @jperals

jperals (approved):

amanabiy:
> Do we have screenshot tests for that?

Yes, we do. We have a some permutations that has filtering tool, but not pagination, and collection prefrences.

PR #4337, opened 2026-03-05
feat: add dev warning for missing field definition labels in attribute-editor

Who-is-PS:
### Description

Adds a console warning via `warnOnce` when an `AttributeEditor` field definition is missing a `label` prop. The `label` is used as `aria-label` for the inputs in the column, so omitting it degrades the screen reader experience.

This change helps guide developers toward providing accessible labels without introducing a breaking change (making `label` required).

Related links, issue #, if available: `AWSUI-61432`

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/attribute-editor/__tests__/warnings.test.tsx
```diff
@@ -0,0 +1,53 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import React from 'react';
+import { render } from '@testing-library/react';
+
+import AttributeEditor from '../../../lib/components/attribute-editor';
+
+jest.mock('@cloudscape-design/component-toolkit', () => ({
+  ...jest.requireActual('@cloudscape-design/component-toolkit'),
+  useContainerQuery: jest.fn().mockImplementation(() => ['m', () => {}]),
```
cansuaa, line 12: Do we need this line? If you mock '@cloudscape-design/component-toolkit/internal' we shouldn't need this since we are only interested in 'warnOnce'. I'd check how the other warnOnce usages are tested e.g. https://github.com/cloudscape-design/components/blob/8093e3f2654b92ebfd592edd3c30ba978a309d38/src/button-dropdown/__tests__/button-dropdown.test.tsx#L16-L23

Who-is-PS: Yeah true, the `useContainerQuery` mock is unnecessary here. I've switched to mocking `@cloudscape-design/component-toolkit/internal` with just `warnOnce: jest.fn()`, matching the pattern in `button-dropdown` and `table`. This also lets us drop the `consoleWarnSpy` entirely since we can assert directly on the `warnOnce` mock and use `mockReset()` between tests instead of relying on a separate file for a clean `warnOnce` cache

src/attribute-editor/__tests__/warnings.test.tsx
```diff
@@ -39,7 +39,6 @@ describe('AttributeEditor component', () => {
       />
     );
 
-    expect(warnOnce).toHaveBeenCalledTimes(1);
```
cansuaa: Could you check why it was being triggered twice before deleting it?

Who-is-PS: The warning check runs directly in the render body (not in a `useEffect`/`useMemo`), so it fires on every render. The original test had `useContainerQuery` mocked to return a stable value (`['m', () => {}]`), which prevented the re-render. When I refactored the test to mock `@cloudscape-design/component-toolkit/internal` instead of `@cloudscape-design/component-toolkit`, I lost that `useContainerQuery` mock — so the real hook runs, resolves the container query asynchronously, and triggers a second render → second `warnOnce` call.

PR #4344, opened 2026-03-11
chore: add optional parent boundary clamping for popover trigger positioning

jsilll:
## Summary of Changes

Adds optional parent boundary clamping for popover trigger positioning. This allows popovers to respect parent container boundaries when positioning their arrows.

**New Props:**
- Added `parentRef?: React.RefObject<HTMLElement>` to `ChartPopover` and `PopoverContainer` components

**Implementation Details:**
- Introduced `getClampedTrackRect()` function in `use-popover-position.ts` that:
  - Clamps the track center point to parent boundaries when `parentRef` is provided
  - Calculates adjusted track position while preserving original dimensions
  - Uses logical properties (insetInlineStart, blockSize, etc.) for RTL/writing-mode support
  - Falls back to normal behavior when `parentRef` is absent (backward compatible)

https://github.com/user-attachments/assets/27a93b25-952c-460f-918f-39763d3ffee6


src/popover/use-popover-position.ts
```diff
@@ -262,3 +265,42 @@ function isBoundary(element: HTMLElement) {
...
+  if (!parentRef) {
+    return trackRect;
+  }
+
+  const parentRect = getLogicalBoundingClientRect(parentRef);
+  const parentInlineEnd = parentRect.insetInlineStart + parentRect.inlineSize;
+  const parentBlockEnd = parentRect.insetBlockStart + parentRect.blockSize;
+
+  // Find where the arrow should point (closest visible point to actual track center)
+  const trackCenterInline = trackRect.insetInlineStart + trackRect.inlineSize / 2;
```
jperals: Why do we need to take the center of the target into account? Wouldn't it be enough to return the rectangle resulting from intersecting the track and the parent?

src/popover/use-popover-position.ts
```diff
@@ -262,3 +265,42 @@ function isBoundary(element: HTMLElement) {
   const computedStyle = getComputedStyle(element);
   return !!computedStyle.clipPath && computedStyle.clipPath !== 'none';
 }
+
+function clamp(value: number, min: number, max: number): number {
+  return Math.max(min, Math.min(value, max));
+}
+
+function getClampedTrackRect(track: HTMLElement | SVGElement, parentRef?: HTMLElement | null) {
```
jperals, line 269: Can you add unit testing coverage for this new logic?

jperals:
I updated the PR title to fit the format accepted by our release notes script (just `feat:` instead of `feat(popover):`)

src/popover/use-popover-position.ts
jsilll: Yes, sure! I just wanted to get your feedback on the implementation first.

src/popover/use-popover-position.ts
jsilll: I made it in such a way that the popover arrow always points to the closest visible point to the actual track center. I think we could do that yes but it would behave slightly differently.

jsilll: I'll try to make a much simpler version of this

src/popover/utils/positions.ts
```diff
@@ -345,6 +345,16 @@ function isTopOrBottom(internalPosition: InternalPosition) {
   return ['top', 'bottom'].includes(internalPosition.split('-')[0]);
 }
 
+export function clampRectStart<T extends BoundingBox>(rect: T, parentRect: BoundingBox): T {
```
jperals: Why the generic `T`? Can we be more specific?

src/popover/utils/positions.ts
```diff
@@ -345,6 +345,16 @@ function isTopOrBottom(internalPosition: InternalPosition) {
...
+  const parentInlineEnd = parentRect.insetInlineStart + parentRect.inlineSize;
+  const parentBlockEnd = parentRect.insetBlockStart + parentRect.blockSize;
+  return {
+    ...rect,
+    insetInlineStart: Math.max(parentRect.insetInlineStart, Math.min(rect.insetInlineStart, parentInlineEnd)),
+    insetBlockStart: Math.max(parentRect.insetBlockStart, Math.min(rect.insetBlockStart, parentBlockEnd)),
+  } as T;
```
jperals: Can we get rid of the `as T` for better type safety?

src/popover/utils/positions.ts
```diff
@@ -345,6 +345,16 @@ function isTopOrBottom(internalPosition: InternalPosition) {
   return ['top', 'bottom'].includes(internalPosition.split('-')[0]);
 }
 
+export function clampRectStart<T extends BoundingBox>(rect: T, parentRect: BoundingBox): T {
+  const parentInlineEnd = parentRect.insetInlineStart + parentRect.inlineSize;
+  const parentBlockEnd = parentRect.insetBlockStart + parentRect.blockSize;
+  return {
```
jperals, line 363: Although in the case we are trying to solve we only care about the "inline start" side, and the new API is internal, nothing in it indicates that the `parent` will be used to clamp only from that side. Unless it adds considerable complexity, would it be possible to clamp all sides?

src/popover/use-popover-position.ts
jperals: Thanks! Note, Codecov is still complaining about line 280 below not being covered. You might want to either add coverage for `getClampedTrackRect` specifically, or some case(s) for the popover component using a value for `parent` (if not too complicated, I'd personally prefer the latter, because it would test the integration of the logic in the component rather than in isolation).

src/popover/utils/positions.ts
jsilll: We're already clamping both the `inline` and `block` directions. While working on this I found out that doing it for the `block` direction actually fixes a problem in the current version of the `chart-components` library.

https://github.com/user-attachments/assets/3b2f65d1-0e41-4ee6-b7b9-64b00d7d62a5

src/popover/use-popover-position.ts
jsilll: I agree with the latter approach as well!

src/popover/utils/positions.ts
jsilll: Oh I missed the already defined `Rect` type

src/popover/utils/positions.ts
jsilll: In this new simplified implementation I'm not clamping `inlineSize` or `blockSize` which means the center of the track can still overflow by a very tiny amount outside of the container in some cases. I think this is small enough issue for us to ignore.

src/popover/utils/positions.ts
jperals: Thanks! Now the tests complain about the requiring `insetBlockEnd` and `insetInlineEnd` not being passed (see, TypeScript doing its job 😛)

src/popover/utils/positions.ts
jperals: What I meant is that we are clamping on the "start" (of both inline and block directions) but not on the "end".

src/popover/utils/positions.ts
```diff
@@ -345,6 +345,14 @@ function isTopOrBottom(internalPosition: InternalPosition) {
   return ['top', 'bottom'].includes(internalPosition.split('-')[0]);
 }
 
+export function clampRectStart(rect: Rect, bounds: BoundingBox) {
+  const parentInlineEnd = bounds.insetInlineStart + bounds.inlineSize;
+  const parentBlockEnd = bounds.insetBlockStart + bounds.blockSize;
+  rect.insetInlineStart = Math.max(bounds.insetInlineStart, Math.min(rect.insetInlineStart, parentInlineEnd));
```
jperals: Any particular reason to modify the original passed-in object rather than creating a new one?

jsilll: Looking again at the function I preferred to have it mutate the argument instead of creating a new one. I know this is mostly a matter of personal preference so if you prefer to create the new object I'm also fine with that.

jperals: In general it is preferred to not mutate objects, because from within the function you don't know what else is the object is used for, so its mutation might be unexpected to a user outside of it, leading to hard-to-debug issues.

src/popover/use-popover-position.ts
jperals: Thanks for adding the test, although Codecov still complains about line 280 not being covered. Could this be because by asserting
```
  expect(usePopoverPositionSpy.mock.calls[0][0].parentRef?.current).toBe(parentRef);
```
you are only ever checking the first call, which corresponds to when the ref is null?

src/popover/utils/positions.ts
jperals: Thanks for the changes! Should the function be renamed accordingly, e.g., to just `clampRect`? :)

src/popover/container.tsx
```diff
@@ -27,6 +27,8 @@ interface PopoverContainerProps {
   */
   trackKey?: string | number;
   minVisibleBlockSize?: number;
+  /** Optional parent element to clamp popover position within its bounds */
+  parentRef?: React.RefObject<HTMLElement>;
```
jperals: Not blocking, but this prop name does not convey very well what it is used for. What about something like `triggerClipRef`, `triggerBoundaryRef` or `triggerClampRef`?

jsilll: Let's go with `triggerClampRef` :) Thanks for the suggestion!

src/popover/__tests__/positions.test.ts
```diff
@@ -1,7 +1,9 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
+import { Rect } from '../../../lib/components/popover/interfaces';
 import {
   calculatePosition,
+  clampRectStart,
```
jperals: This needs to be updated to the new function name

src/popover/__tests__/positions.test.ts
```diff
@@ -280,3 +282,56 @@ describe('isCenterOutside', () => {
     ).toBe(false);
   });
 });
+
+describe('clampStart', () => {
```
jperals: Also update the function name in this test description

jperals:
Actually I'd suggest prefixing the PR title with `chore:` instead of `feat:` because this is not a public feature

src/popover/__tests__/positions.test.ts
jsilll: Sorry for missing this

PR #4058, opened 2025-11-20
fix: Restore legacy test util selectors for Radio Group

jperals:
### Description

Ticket: `AWSUI-61492`

Related doc: `zOhkAQWpDXKq`

### How has this been tested?

1. Updated test-utils selectors with `npm run build && npx jest -c jest.unit.config.js -u src/__tests__/snapshot-tests/test-utils-selectors.test.tsx` and verified the change in the snapshot (part of this PR's diff)
2. Manually verified in my local environment that both the radio group and the inner radio button have the legacy classnames

<details>
   <summary>Review checklist</summary>

_The following items are to be evaluated by the author(s) and the reviewer(s)._

#### Correctness

- _Changes include appropriate documentation updates._
- _Changes are backward-compatible if not indicated, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#public-apis)._
- _Changes do not include unsupported browser features, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#browsers-support)._
- _Changes were manually tested for accessibility, see [accessibility guidelines](https://cloudscape.design/foundation/core-principles/accessibility/)._

#### Security

- _If the code handles URLs: all URLs are validated through [the `checkSafeUrl` function](https://github.com/cloudscape-design/components/blob/main/src/internal/utils/check-safe-url.ts)._

#### Testing

- _Changes are covered with new/existing unit tests?_
- _Changes are covered with new/existing integration tests?_
</details>

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #4345, opened 2026-03-11
feat: make dropdown component public

mxschll:
### Description

- Exposes the internal dropdown component and refactors existing components/pages to use public version.
- Introduces test utils for dropdown.
- Adds style api to dropdown.

Related links, issue #, if available: `RGJgABPXohP8`

### How has this been tested?

Changes went through the pipeline without visual regressions.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/dropdown/__tests__/dropdown.test.tsx
+++ b/src/dropdown/__tests__/dropdown.test.tsx
@@ -58,6 +58,32 @@ describe('Dropdown Component', () => {
       expect(handleOutsideClick).toHaveBeenCalled();
     });
 
+    test('does not fire event when clicking the trigger element', async () => {
+      const handleOutsideClick = jest.fn();
+      renderDropdown(
+        <Dropdown
+          trigger={<button data-testid="trigger">Trigger</button>}
+          onOutsideClick={handleOutsideClick}
+          open={true}
+        />
+      );
+      await runPendingEvents();
+
+      act(() => screen.getByTestId('trigger').click());
+      expect(handleOutsideClick).not.toHaveBeenCalled();
+    });
+
+    test('fires event when clicking the trigger wrapper div', async () => {
+      const handleOutsideClick = jest.fn();
+      const [wrapper] = renderDropdown(
+        <Dropdown trigger={<button>Trigger</button>} onOutsideClick={handleOutsideClick} open={true} />
+      );
+      await runPendingEvents();
+
+      act(() => wrapper.findTrigger().getElement().click());
+      expect(handleOutsideClick).toHaveBeenCalled();
+    });
+
     test('does not fire event when a portaled element inside dropdown is clicked', async () => {
       const handleOutsideClick = jest.fn();
       renderDropdown(
@@ -515,6 +541,33 @@ describe('Dropdown Component', () => {
       await runPendingEvents();
 
       act(() => screen.getByTestId('external-trigger').click());
+      expect(handleOutsideClick).not.toHaveBeenCalled();
+    });
+
+    test('does not fire onOutsideClick when clicking a child of the external trigger', async () => {
+      const handleOutsideClick = jest.fn();
+      function TestComponent() {
+        const triggerRef = useRef<HTMLDivElement>(null);
+        return (
+          <div>
+            <button data-testid={outsideId} />
+            <div ref={triggerRef}>
+              <button data-testid="trigger-child">Child</button>
+            </div>
+            <Dropdown
+              trigger={null}
+              triggerRef={triggerRef as React.RefObject<HTMLElement>}
+              open={true}
+              onOutsideClick={handleOutsideClick}
+              content={<div>Content</div>}
+            />
+          </div>
+        );
+      }
+      render(<TestComponent />);
+      await runPendingEvents();
+
+      act(() => screen.getByTestId('trigger-child').click());
       expect(handleOutsideClick).not.toHaveBeenCalled();
     });
   });
--- a/src/test-utils/dom/internal/dropdown.ts
+++ b/src/test-utils/dom/internal/dropdown.ts
@@ -3,9 +3,14 @@
 import { ElementWrapper } from '@cloudscape-design/test-utils-core/dom';
 
 import styles from '../../../dropdown/styles.selectors.js';
+import testUtilStyles from '../../../dropdown/test-classes/styles.selectors.js';
 import legacyStyles from '../../../internal/components/dropdown/styles.selectors.js';
 export default class DropdownWrapper extends ElementWrapper {
   static rootSelector: string = styles.root;
+
+  findTrigger(): ElementWrapper {
+    return this.findByClassName(testUtilStyles.trigger)!;
+  }
 
   findOpenDropdown(): ElementWrapper | null {
     return this.find(`:is(.${styles.dropdown}, .${legacyStyles.dropdown})[data-open=true]`);


pages/date-range-picker/month-calendar-permutations.page.tsx
```diff
@@ -5,7 +5,7 @@ import * as React from 'react';
 
 import Box from '~components/box';
 import Calendar, { DateRangePickerCalendarProps } from '~components/date-range-picker/calendar';
-import Dropdown from '~components/internal/components/dropdown';
+import Dropdown from '~components/dropdown/internal';
```
jperals, line 8: Why not just `~components/dropdown`? In dev pages we use the root file instead of the internal one (with very few exceptions, which we should probably address), and I think this makes sense because that is where customers will import the component from.

(The same applies for the rest of modified dev pages)

src/test-utils/dom/dropdown/index.ts
```diff
@@ -0,0 +1,115 @@
...
+    return this.findOpenDropdown(options) !== null;
+  }
+
+  /**
+   * Clicks the trigger element.
+   */
+  @usesDom
+  clickTrigger(): void {
+    act(() => {
+      const trigger = this.findTrigger().getElement().firstElementChild as HTMLElement | null;
```
mxschll: This is necessary to click the trigger button inside the wrapper. Not sure if there's a better way.

pages/date-range-picker/month-calendar-permutations.page.tsx
mxschll: Some dropdowns on the dev pages still use internal props like `stretchHeight={true}` in this case. I'll check again if it is really necessary on the dev pages.

jperals: I see.. Well, for those cases I think we can use the internal imports (these pages are testing internal features), but otherwise default to the public imports

mxschll: Addressed in new revision.

src/dropdown/interfaces.ts
```diff
@@ -59,118 +82,64 @@ export interface BaseDropdownHostProps extends ExpandToViewport {
 
 export interface DropdownProps extends ExpandToViewport {
   /**
-   * Trigger element.
```
jperals, line 62: We have the convention of ending all API prop description with a dot.

src/dropdown/interfaces.ts
```diff
@@ -59,118 +82,64 @@ export interface BaseDropdownHostProps extends ExpandToViewport {
...
    * Open state of the dropdown.
    */
   open?: boolean;
 
   /**
-   * Called when the user clicks outside the dropdown and trigger.
-   * The dropdown does not close automatically - the parent component
-   * must update the `open` prop to close the dropdown.
+   * Called when the user clicks outside the dropdown. The dropdown does not
+   * close automatically - the parent component must update the `open` prop to
```
jperals: Not sure taking about "the parent component" helps or just can add confusion. What about
> Called when the user clicks outside the dropdown. The dropdown does not close automatically - the `open` prop needs to be updated to actually close the dropdown.

As a side note, I see that we will have this mentioned in the dev guidelines, so this is overlaps with that, but I'm OK with having this in both places as this seems to be quite an important point for builders in order to use the component correctly.

src/dropdown/interfaces.ts
```diff
@@ -59,118 +82,64 @@ export interface BaseDropdownHostProps extends ExpandToViewport {
...
-   */
-  stretchHeight?: boolean;
-
-  /**
-   * Stretches the trigger to the height of the dropdown container.
-   */
-  stretchTriggerHeight?: boolean;
-
   /**
    * Minimum width constraint for the dropdown.
```
jperals: I find this enumeration of the allowed type and undefined quite an atypical format for our API docs. What about:
> Minimum width for the dropdown in pixels. If no value is specified, the dropdown will shrink to fit its content.

src/dropdown/interfaces.ts
```diff
@@ -59,118 +82,64 @@ export interface BaseDropdownHostProps extends ExpandToViewport {
...
    * - Number: minimum width in pixels
-   * - 'trigger': dropdown will be at least as wide as the trigger
-   * - undefined: no minimum constraint (fits content)
+   * - undefined: no maximum constraint (natural content sizing)
    */
-  minWidth?: DropdownWidthConstraint;
+  minWidth?: number;
 
   /**
    * Maximum width constraint for the dropdown.
```
jperals: Similar here. Suggestion:
> Maximum width for the dropdown in pixels. If no value is specified, the dropdown will expand to fit its content.

src/dropdown/interfaces.ts
```diff
@@ -59,118 +82,64 @@ export interface BaseDropdownHostProps extends ExpandToViewport {
...
 
   /**
-   * Called when focus enters the trigger or dropdown content.
-   */
-  onFocus?: NonCancelableEventHandler<Pick<React.FocusEvent, 'target' | 'relatedTarget'>>;
-
-  /**
-   * Called when focus leaves the trigger or dropdown content.
+   * Called when the user presses the Escape key while the dropdown is open.
+   * The dropdown does not close automatically - the parent component
```
jperals: Same comment here about "the parent component"

src/dropdown/interfaces.ts
```diff
@@ -180,36 +149,30 @@ export interface DropdownProps extends ExpandToViewport {
...
    */
   ariaLabelledby?: string;
 
   /**
-   * Describedby for the dropdown (recommended when role="dialog")
+   * Adds `aria-describedby` to the dropdown content element.
    */
   ariaDescribedby?: string;
+
+  /**
```
jperals, line 188: Is this necessary since the entire component is Core-only anyway?

src/dropdown/interfaces.ts
mxschll: That makes it better understandable 👍

src/dropdown/interfaces.ts
mxschll: I'd leave it here for the sake of consistency, as well as for future use in case we want to expose the component to the console (probably unlikely), so we can keep the style API core only. Or is this concern unjustified?

src/dropdown/interfaces.ts
mxschll: Addressed in new revision.

src/dropdown/interfaces.ts
mxschll: Addressed in new revision.

src/dropdown/interfaces.ts
mxschll: Addressed in new revision.

src/dropdown/interfaces.ts
mxschll: Addressed in new revision.

src/dropdown/interfaces.ts
mxschll: Addressed in new revision.

src/dropdown/__tests__/__snapshots__/dropdown.test.tsx.snap
```diff
@@ -0,0 +1,34 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Dropdown Component Style API all style properties 1`] = `
+CSSStyleDeclaration {
+  "0": "background",
```
jperals: This snapshot is a bit noisy. Can we reduce it to just the style attributes that we expect and their values?

src/dropdown/interfaces.ts
```diff
@@ -2,10 +2,33 @@
...
 import React from 'react';
 
-import { NonCancelableEventHandler } from '../../events';
+import { NonCancelableEventHandler } from '../internal/events';
 
 export type OptionsFilteringType = 'none' | 'auto' | 'manual';
 
+export namespace DropdownProps {
+  export interface Style {
+    dropdown?: {
```
jperals, line 11: Why are the properties actually nested under `dropdown`? Is it because there is also the trigger, although we don't consider styling it for now?

src/dropdown/interfaces.ts
jperals: Not blocking, as this doesn't break anything

src/test-utils/dom/dropdown/index.ts
jperals: Maybe we don't actually need to provide this method? Since customers can trigger events on the elements that they render inside the trigger.

src/dropdown/__tests__/dropdown.test.tsx
```diff
@@ -418,6 +416,34 @@ describe('Dropdown Component', () => {
     });
   });
 
+  describe('Style API', () => {
+    test('all style properties', () => {
```
jperals: Does this cover Core and non-Core?

src/dropdown/interfaces.ts
mxschll: Yes, I wanted to go with `root` initially but that's also not ideal as we need to style the `:after` element which is used to display the border as well as the dropdown wrapper to change border radius and background. So I grouped it in `dropdown`. Since the trigger is custom, we probably could remove this nesting entirely...

src/test-utils/dom/dropdown/index.ts
mxschll: I agree, let's remove it. I thought it would be convenient to have, but since we don't know what builders render as the trigger, the behavior if this method is unpredictable.

mxschll: Removed in latest revision.

src/dropdown/__tests__/__snapshots__/dropdown.test.tsx.snap
mxschll: Agree. Also moving style api tests to `styles.test.tsx` to be consistent with other components.

src/dropdown/__tests__/dropdown.test.tsx
mxschll: Good catch, made it consistent with other style api tests in `styles.test.tsx`

src/dropdown/internal.tsx
```diff
@@ -409,7 +442,14 @@ const Dropdown = ({
       // Since the listener is registered on the window, `event.target` will incorrectly point at the
       // shadow root if the component is rendered inside shadow DOM.
       const target = event.composedPath ? event.composedPath()[0] : event.target;
-      if (!nodeBelongs(dropdownRef.current, target) && !nodeBelongs(triggerRef.current, target)) {
+      // When using an internal trigger, triggerRef is the wrapper div around the trigger element.
+      // A click on the wrapper directly (in the gap between the wrapper boundary and the trigger element inside)
+      // must fire onOutsideClick. Only clicks on strict descendants (the actual trigger element) should suppress it.
+      // When using an external triggerRef, the ref points directly at the trigger element, so all clicks on it are inside.
+      const isInsideTriggerElement = externalTriggerRef
```
jperals: Wouldn't this be the other way around?
```
const isInsideTriggerElement = externalTriggerRef
  ? nodeBelongs(triggerRef.current, target) && target !== triggerRef.current
  : nodeBelongs(triggerRef.current, target);
```

Because it is in the case that the external trigger is provided that we need to check that "gray area" in the slot and consider it as "outside".

I also wonder if there is alternative way of expressing this logic that make it more understandable. Something like this at least removes the extra cognitive load of negating `isInsideTriggerElement` in the condition:

```
const isOutsideTrigger = externalTriggerRef
  ? !nodeBelongs(triggerRef.current, target) || target === triggerRef.current
  : !nodeBelongs(triggerRef.current, target);
if (!nodeBelongs(dropdownRef.current, target) && isOutsideTrigger) {
```

In any case, can we have an integration test that clicks on that trigger slot area and asserts that the event is triggered (e.g, by checking that the dropdown closes)?

src/dropdown/internal.tsx
```diff
@@ -24,9 +27,31 @@ import {
...
+  minWidth?: DropdownWidthConstraint;
+  maxWidth?: DropdownWidthConstraint;
+  preferredAlignment?: DropdownAlignment;
+  hideBlockBorder?: boolean;
+  onFocus?: NonCancelableEventHandler<Pick<React.FocusEvent, 'target' | 'relatedTarget'>>;
+  onBlur?: NonCancelableEventHandler<Pick<React.FocusEvent, 'target' | 'relatedTarget'>>;
+}
 
 import styles from './styles.css.js';
+import testutilStyles from './test-classes/styles.css.js';
```
jperals: Nit: `testutilStyles` does not fully follow camelCase convention (`testUtilStyles`)

src/__tests__/snapshot-tests/__snapshots__/test-utils-selectors.test.tsx.snap
```diff
@@ -378,7 +386,6 @@ exports[`test-utils selectors 1`] = `
     "awsui_direction-button-inline-start_8k1rt",
     "awsui_direction-button-visible_8k1rt",
     "awsui_disabled_15o6u",
-    "awsui_dropdown_qwoo0",
```
jperals, line 381: Can removing these selectors cause failures in consumer tests that find dropdowns in e.g, selects, multiselects or autosuggests? See `4Kj2A35lgyvA`.

One way to mitigate this is to temporarily leave the old stylesheet, only containing the relevant empty rulesets for these selectors, and use these old selectors as `legacyRootSelector` in the test utils.

Example PR: https://github.com/cloudscape-design/components/pull/4058

src/test-utils/dom/internal/dropdown-host.ts
```diff
@@ -31,10 +31,10 @@ export default abstract class DropdownHostComponentWrapper extends ComponentWrap
    * @param options
    * * expandToViewport (boolean) - Use this when the component under test is rendered with an `expandToViewport` flag.
    */
-  findDropdown(options = { expandToViewport: false }): DropdownContentWrapper {
+  findDropdown(options = { expandToViewport: false }): OptionsDropdownContentWrapper {
```
jperals, line 35: Note that renaming this type will break some consumers that import it. Not blocking because they shouldn't, but FYI

src/dropdown/internal.tsx
mxschll: Addressed in latest revision.

src/__tests__/snapshot-tests/__snapshots__/test-utils-selectors.test.tsx.snap
mxschll: Restored the legacy selector and using `:is(.${styles.dropdown}, .${legacyStyles.dropdown}` in internal test utils that relied them.

src/test-utils/dom/dropdown/index.ts
```diff
@@ -0,0 +1,69 @@
...
+// SPDX-License-Identifier: Apache-2.0
+import { ComponentWrapper, createWrapper, ElementWrapper } from '@cloudscape-design/test-utils-core/dom';
+
+import styles from '../../../dropdown/styles.selectors.js';
+import testUtilStyles from '../../../dropdown/test-classes/styles.selectors.js';
+import legacyDropdownStyles from '../../../internal/components/dropdown/styles.selectors.js';
+
+export class DropdownContentWrapper extends ComponentWrapper {
+  static legacyRootSelector: string = legacyDropdownStyles.root;
+  static legacyDropdownSelector: string = legacyDropdownStyles.dropdown;
```
jperals: Is this second static declaration (`legacyDropdownSelector`) needed?

mxschll: No, I can remove it again since it's being used by utils now.

src/dropdown/internal.tsx
mxschll: I think it might be a bit confusing having `externalTriggerRef` which is provided from the outside and `triggerRef` which is a ref to the trigger wrapper and not the actual trigger IF `externalTriggerRef` was not provided.

When `externalTriggerRef` is provided, `triggerRef` is the trigger element, so clicking it means `target === triggerRef.current`. We want that to count as "inside" and not fire `onOutsideClick`.

If `externalTriggerRef` is *not* provided, `triggerRef` is the wrapper div. Clicking the wrapper itself should count as outside "gray area" and only clicks on its children are inside. So we exclude `target === triggerRef.current`.

But I agree it's not very readable, I'll flip it to check for `isOutsideTrigger` as it is more intuitive.

mxschll: Updated the logic to this:

```js
// For internal triggers, the wrapper div itself counts as outside — only its children are the trigger.
// For external triggers, the ref is the trigger element directly, so it counts as inside.
const isOutsideTrigger = !nodeBelongs(triggerRef.current, target) || (!externalTriggerRef && target === triggerRef.current);
```

It's still not ideal because `triggerRef` can be either the trigger element itself or the wrapper which leads to the confusion I think, but it might be a bit more readable.

jperals: I think what was not clear to me was the exact meaning of "internal" and "external" trigger and their mapping to the `trigger` and `triggerRef` props respectively. I'd say this part is still not super obvious, but I agree that the code is a bit more clear now.

PR #4352, opened 2026-03-12
feat: icon provider custom icons

bmv437:
### Description

Adds custom icon support to `IconProvider`, allowing consumers to register arbitrary icon names (beyond the built-in set) and use them across all Cloudscape components that accept icon names.

**Notable changes:**

- **`IconRegistry` / `IconMap` types** — Consumers use TypeScript module augmentation (`declare module`) to register custom icon names, making them type-safe across `Icon`, `Button`, and all other icon-accepting components.
- **`defineIcons` helper** — A type-identity function that lets consumers define a custom icon map (`defineIcons({ rocket: <RocketSvg /> })`).
- **Single source of truth for icon names** — The duplicated `IconProps.Name` union type is replaced by a shared `IconName` type defined in `icon-provider/interfaces.ts`. `IconProviderProps.Icons` is now a mapped type over `IconName` instead of a hand-maintained interface with ~130 explicit keys. This is now possible due to [this update to documenter](https://github.com/cloudscape-design/documenter/pull/111).
- **Runtime handling** — `InternalIcon` renders custom icons from context and warns when a name isn't found. Nested `IconProvider`s inherit custom icons from parent context, and `null` values correctly fall back to either the built-in default or the inherited custom icon.
- **Exports** — `defineIcons`, `IconRegistry`, and `IconMap` are exported from `icon-provider/index.tsx`.

API proposal: oY6SACNg4aOu

### How has this been tested?

- **Unit tests** (`src/icon-provider/__tests__/define-icons.test.ts`): Validates that `defineIcons` returns the same object reference and accepts built-in, custom, and mixed icon keys.
- **Unit tests** (`src/icon-provider/__tests__/icon-provider.test.tsx`): New test cases for rendering custom icons through context, inheriting across nested providers, null-reset fallback behavior, and console warnings for missing icons.
- **Integration tests** (`src/icon-provider/__integ__/custom-icons.test.ts`): Browser tests verifying custom icon rendering in `Icon` and `Button`, built-in icon coexistence, nested provider inheritance, and null-reset behavior.
- **Demo page** (`pages/icon-provider/custom-icons/`): Visual test page exercising all scenarios with a separate `tsconfig.json` to isolate module augmentation.

<details>

#### Correctness

#### Security

#### Testing

</details>

---


pages/icon-provider/custom-icons/custom-icons.page.tsx
```diff
@@ -0,0 +1,103 @@
...
+// across all Cloudscape components. Isolated to this tsconfig project
+// so it doesn't leak into other pages.
+declare module '~components/icon-provider' {
+  // eslint-disable-next-line @typescript-eslint/no-empty-object-type -- needed for module augmentation to work correctly.
+  interface IconRegistry extends IconMap<typeof customIcons> {}
+}
+
+export default function CustomIconsPage() {
+  return (
+    <article>
```
pan-kot: nit: You can use SimplePage helper for this page instead. Passing `screenshotArea={{}}` to it will enable the internal screenshot area for it, too.

pages/icon-provider/custom-icons/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
```
pan-kot, line 1: good thinking 👍

src/icon-provider/__tests__/icon-provider.test.tsx
```diff
@@ -137,4 +137,83 @@ describe('Icon Provider', () => {
...
+      );
+
+      const firstIcon = wrapper(container).find('[data-testid=firstIcon]')!.find('svg');
+      const secondIcon = wrapper(container).find('[data-testid=secondIcon]')!.find('svg');
+      const expectedSvg = wrapper(container).find('[data-testid=expectedSvg]')!.find('svg');
+
+      expect(firstIcon).toStrictEqual(expectedSvg);
+      expect(secondIcon).toStrictEqual(expectedSvg);
+    });
+  });
```
pan-kot, line 272: should we add tests for the following cases:
1. Custom icon is defined using provider and is then overridden with a nested provider
2. Custom icon override behaviour when passing `{ rocket: undefined }` to the nested provider

src/icon-provider/__tests__/define-icons.test.ts
```diff
@@ -0,0 +1,32 @@
...
+    expect(result).toBe(icons);
+  });
+
+  it('accepts built-in icon names', () => {
+    const icons = defineIcons({ 'add-plus': 'test' as any, close: 'test' as any });
+    expect(icons).toHaveProperty('add-plus');
+    expect(icons).toHaveProperty('close');
+  });
+
+  it('accepts arbitrary string keys', () => {
```
pan-kot: What is this test for, isn't it captured by e.g. the first one?

src/icon-provider/__tests__/define-icons.test.ts
```diff
@@ -0,0 +1,32 @@
...
+    const icons = defineIcons({ rocket: 'test' as any, zap: 'test' as any });
+    expect(icons).toHaveProperty('rocket');
+    expect(icons).toHaveProperty('zap');
+  });
+
+  it('accepts a mix of built-in and custom keys', () => {
```
pan-kot: Is it not the same as first?

src/icon-provider/__integ__/custom-icons.test.ts
```diff
@@ -0,0 +1,68 @@
...
+
+const pageUrl = '#/light/icon-provider/custom-icons/custom-icons';
+
+class CustomIconsPage extends BasePageObject {
+  svgInside(testId: string) {
+    return `[data-testid="${testId}"] svg`;
+  }
+}
+
+function setupTest(testFn: (page: CustomIconsPage) => Promise<void>) {
```
pan-kot, line 15: I think these tests are not really needed, because:
1. The logic can be and is tested by the unit tests
2. We will provide an internal screenshot test to capture the actual render output of the page - it will capture the presence of the icons and also the way they render, which can include colour inheritance and size variants, once supported.

pages/icon-provider/custom-icons/custom-icons.page.tsx
bmv437: Updated.

src/icon-provider/__tests__/icon-provider.test.tsx
bmv437: Added.

src/icon-provider/__tests__/define-icons.test.ts
bmv437: Good point, I simplified/reworked this, and added a type assertion test to validate the object keys would return from `defineIcons` as exact values, are weren't widened to `string`.

PR #4395, opened 2026-03-30
chore: Add themeable tokens for button

at-susie:
### Description
This PR adds independently themeable design tokens for the button component. Specifically it introduces dedicated color tokens for the link button variant to achieve parity with normal and primary variants, and adds a new spaceButtonVertical spacing token so button padding is themeable in both directions. No visual changes — all new tokens default to the previous values.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


pages/theming/themed-stroke-width.page.tsx
```diff
@@ -73,7 +73,14 @@ export default function ThemedStrokeWidthPage() {
...
           borderWidthIconBig: strokeBig ? `${strokeBig}px` : '1.5px',
           borderWidthIconLarge: strokeLarge ? `${strokeLarge}px` : '2px',
+<<<<<<< feat/theme-button
+          spaceButtonHorizontal: '12px',
+          spaceButtonVertical: '4px',
           borderRadiusButton: '8px',
+          borderWidthButton: '1px',
+=======
+          borderRadiusButton: '8px',
+>>>>>>> main
```
jkuelz: Looks like some leftover merge markers that need to be removed?

src/app-layout/visual-refresh-toolbar/toolbar/trigger-button/styles.scss
jkuelz: Does the placement of the notification dot need to be calculated in some way now that the border radius can be changed?

src/app-layout/visual-refresh-toolbar/toolbar/trigger-button/styles.scss
```diff
@@ -40,10 +40,10 @@
   }
 
   &.circle {
```
jkuelz: Does the classname have any specific reason for being called circle? Previously this was always 50%, which always resulted in a circle, now that it is configurable, should we change this classname to something more about its semantic purpose? 

I see we have two options for trigger button shape in the interface: `variant?: 'circle' | 'custom';` Seems like `default` would be more appropriate than `circle`. Can you check with @georgylobko and at the very least, add a comment to clarify?

style-dictionary/visual-refresh/metadata/colors.ts
```diff
@@ -550,6 +565,21 @@ const metadata: StyleDictionary.MetadataIndex = {
...
+  colorTextButtonInlineIconDisabled: {
+    description: 'The color of inline button icons in disabled state.',
+    themeable: true,
+    public: true,
+  },
+  colorTextButtonInlineIconHover: {
+    description: 'The color of inline button icons in hover state.',
+    themeable: true,
+    public: true,
+  },
```
jkuelz, line 592: These weren't explicitly mentioned in the PR description, so just making sure they are supposed to be here.

pages/theming/themed-stroke-width.page.tsx
jkuelz, line 1: Is the stroke width dev page the right place for these examples? Nothing on the page indicates that there will be other tokens other than the icon stroke widths applied. Your change in particular is about button height and colors. Should we create a new page that allows you to toggle on/off the entire core theme?

style-dictionary/visual-refresh/metadata/spacing.ts
jkuelz, line 142: Do we need a vertical one for this as well since buttons and fields should be the same height?

pages/theming/themed-stroke-width.page.tsx
at-susie: Good call. I will create another page or repurpose this page to be more generic

style-dictionary/visual-refresh/metadata/spacing.ts
at-susie: Changing vertical padding doesn't change the height of inputs, since the height of inputs is specified using [the token `sizeVerticalInput`](https://github.com/cloudscape-design/components/blob/dbcec2d678db5f87c32ca9cc8fa9cd467f3e6d3b/src/input/styles.scss#L56), so we might need to make it themeable. I'll look into it.

src/app-layout/visual-refresh-toolbar/toolbar/trigger-button/styles.scss
at-susie: I agree. `default` sounds more appropriate. Will discuss with @georgylobko and update accordingly

pages/theming/themed-stroke-width.page.tsx
at-susie: Removed.

src/app-layout/visual-refresh-toolbar/toolbar/trigger-button/styles.scss
at-susie: I updated the structure a bit here. Instead of clip-path, I added an outline to the dot and I slightly updated the dot position. This way, we don't need to adjust the dot position based on how the trigger button is themed. 
I added a dev page example called themed-components where you can see how it looks.

pages/theming/themed-stroke-width.page.tsx
at-susie: Now, I created a separate page collecting components to see theming

style-dictionary/visual-refresh/metadata/colors.ts
at-susie: These existing tokens currently defaults to `colorTextLinkXxx`, which are currently not themeable, but need to be individually themeable.

src/app-layout/visual-refresh-toolbar/toolbar/trigger-button/styles.scss
at-susie: I updated to be `variant?: 'default' | 'custom'`. @georgylobko Would love to get your thoughts.

georgylobko: Looks good to me

style-dictionary/visual-refresh/metadata/spacing.ts
at-susie: Update: 
1. I made the `sizeVerticalInput` token themeable
2. I introduced `spaceControlVertical` token to be able to individually theme

pages/theming/themed-components.page.tsx
```diff
@@ -0,0 +1,396 @@
...
+    if (themed) {
+      const theme: Theme = {
+        tokens: {
+          spaceButtonHorizontal: '12px',
+          spaceButtonVertical: '4px',
+          borderRadiusButton: '8px',
+          borderWidthButton: '1px',
+          sizeVerticalInput: '30px',
+          borderWidthToken: '1px',
+          spaceControlVertical: '3px',
```
jkuelz: Why is 3px padding needed for inputs to achieve 30px overall height, but buttons require 4px?

src/app-layout/visual-refresh-toolbar/toolbar/trigger-button/styles.scss
```diff
@@ -87,6 +75,7 @@
   border-end-start-radius: 8px;
   border-end-end-radius: 8px;
   background-color: awsui.$color-background-badge-icon;
-  inset-block-start: 1px;
+  inset-block-start: -1px;
   inset-inline-end: -1px;
+  outline: solid 2px awsui.$color-background-layout-panel-content;
```
jkuelz, line 80: perhaps this deserves its own token for toolbar-background?

pages/theming/themed-components.page.tsx
```diff
@@ -0,0 +1,396 @@
...
+  );
+}
+
+export default function ThemedComponentsPage() {
+  const [themed, setThemed] = useState<boolean>(false);
+  const [strokeSmall] = useState<string>('2');
+  const [strokeNormal] = useState<string>('2');
+  const [strokeMedium] = useState<string>('2');
+  const [strokeBig] = useState<string>('3');
+  const [strokeLarge] = useState<string>('4');
```
jkuelz: Can these be removed?

pages/theming/themed-components.page.tsx
```diff
@@ -0,0 +1,396 @@
...
+        baseThemeId: 'visual-refresh',
+      });
+      reset = result.reset;
+    }
+    return reset;
+  }, [themed, strokeSmall, strokeNormal, strokeMedium, strokeBig, strokeLarge]);
+
+  return (
+    <div style={{ padding: 15 }}>
+      <h1>Themed icon stroke width</h1>
```
jkuelz: Perhaps a different page title?

pages/theming/themed-components.page.tsx
```diff
@@ -0,0 +1,396 @@
...
+  const [strokeBig] = useState<string>('3');
+  const [strokeLarge] = useState<string>('4');
+
+  // Reload page once after initial load to fix theme application
+  useLayoutEffect(() => {
+    const hasReloaded = sessionStorage.getItem('themed-stroke-width-reloaded');
+    if (!hasReloaded) {
+      sessionStorage.setItem('themed-stroke-width-reloaded', 'true');
+      window.location.reload();
+    }
```
jkuelz: Is it a problem we are using the same storage key from the other page? Why is this needed?

style-dictionary/visual-refresh/spacing.ts
```diff
@@ -10,6 +10,8 @@ const tokens: StyleDictionary.SpacingDictionary = {
   spaceAlertVertical: '{spaceFlashbarVertical}',
   spaceButtonFocusOutlineGutter: '4px',
   spaceButtonHorizontal: '{spaceScaledL}',
+  spaceButtonVertical: '{spaceScaledXxs}',
```
jkuelz, line 13: should this default to spaceControlVertical? Like `spaceButtonVertical: '{spaceControlVertical}'` since they should be the same most of the time?

src/internal/styles/forms/constants.scss
```diff
@@ -19,7 +19,7 @@ $group-option-padding-with-border-placeholder-vertical: calc(
   #{$group-option-padding-vertical} + (#{awsui.$border-item-width} - #{awsui.$border-divider-list-width})
 );
 
-$control-padding-vertical: awsui.$space-scaled-xxs;
+$control-padding-vertical: awsui.$space-control-vertical;
 $control-padding-horizontal: awsui.$space-field-horizontal;
```
jkuelz: I know these are kinda inconsistent in general, but since the horizontal token uses "field" naming, should we keep it consistent? `awsui.$space-field-vertical`?

src/input/styles.scss
```diff
@@ -53,7 +53,7 @@
   box-shadow: var(#{custom-props.$styleBoxShadowDefault});
 
   @include styles.font-body-m;
-  block-size: awsui.$size-vertical-input;
+  min-block-size: awsui.$size-vertical-input;
```
jkuelz, line 56: Doesn't this change make the field respect `awsui.$space-control-vertical` when themed? If so, do we need to expose both? I would also double check that this change doesn't have any adverse effects across browsers or on fields that have multiple lines of text (e.g. dropdowns with additional metadata). IIRC, this used to cause issues in Safari, but it could be fixed by now.

src/property-filter/styles.scss
```diff
@@ -117,7 +117,7 @@ $operator-field-width: 120px;
 
   &-multiselect-wrapper {
     position: relative;
-    block-size: awsui.$size-vertical-input;
+    min-block-size: awsui.$size-vertical-input;
```
jkuelz, line 120: Same comment here about the cross-browser behavior and redundant tokens.

style-dictionary/visual-refresh/metadata/spacing.ts
jkuelz: With the switch from 'inline-size` to `min-inline-size` change, is the `sizeVerticalInput` token redundant?

pages/theming/themed-components.page.tsx
at-susie: This is not needed. Removed.

pages/theming/themed-components.page.tsx
at-susie: This was a mistake. It's supposed to be 4px

src/app-layout/visual-refresh-toolbar/toolbar/trigger-button/styles.scss
at-susie: It can, but that feels a bit out of scope for this PR. Happy to revisit whenever needed.

pages/theming/themed-components.page.tsx
at-susie: Removed.

pages/theming/themed-components.page.tsx
at-susie: Renamed.

src/input/styles.scss
at-susie: If we keep the `block-size`, it will not respect the `awsui.$space-control-vertical` when themed. That's why we need to make it `min-block-size`. This way, the `min-block-size` works as a guardrail while the `awsui.$space-control-vertical` is themeable.

at-susie: But it makes sense to double-check different examples.

src/property-filter/styles.scss
at-susie: As mentioned in [the above reply,](https://github.com/cloudscape-design/components/pull/4395#discussion_r3057763058) we would rather need to make it min-block-size

pages/theming/themed-components.page.tsx
at-susie: Fixed

style-dictionary/visual-refresh/spacing.ts
at-susie: I would rather keep it separated from `spaceControlVertical` as button can have different stroke-widths And depending on that they need to adjust vertical padding independently. So they are different.

src/internal/styles/forms/constants.scss
at-susie: Updated to be `space-field-vertical`

style-dictionary/visual-refresh/metadata/spacing.ts
at-susie: The sizeVerticalInput is still needed as min-block-size. We can also change from `sizeVerticalInput` to something like `sizeMinVerticalInput` to be explicit. WDYT?

jkuelz: Do we need to expose it though? Could we just calculate the height based on the spacing that is set? Or vice versa? Because right now you need to explicitly define both otherwise there is no effect, which could be confusing. Is there a reason when you would define one different from the other?

at-susie: I totally agree with the confusion if we expose both. I think I will keep the `sizeVerticalInput` token unpublished. And we should be able to adjust the height only by adjusting the spaceFieldVertical token.

PR #4417, opened 2026-04-08
fix: allow text in read-only components to be selected

### LINKED ISSUE #4411: [Bug]: Text in read-only date range picker is unselectable
### Browser

_No response_

### Package version

v3.0.1267

### React version

v19.2.3

### Description

I'm displaying a date range using a `DateRangePicker` with `readOnly={true}`:

Image

When I do this, the text shown in the widget is unselectable by the user. This is perhaps expected for a `disabled` component, but not for a `readOnly` one. It's a usability issue, and potentially an accessibility issue as well.

### Source code

_No response_

### Reproduction

_No response_

### Code of Conduct

TrevorBurnham:
Fixes #4411

This PR modifies the styles on `DateRangePicker`, `FileTokenGroup`, `Select` and `Multiselect` to make text selectable.

#### Testing

I've added a read-only toggle to the demo pages and was able to manually test each of these.

- _Changes are covered with new/existing unit tests?_ No
- _Changes are covered with new/existing integration tests?_ No

---

### DIFF
--- a/src/date-range-picker/__integ__/date-range-picker.test.ts
+++ b/src/date-range-picker/__integ__/date-range-picker.test.ts
@@ -98,4 +98,38 @@ describe('Date Range Picker', () => {
       }, granularity)
     );
   });
+
+  describe('Read-only mode', () => {
+    const setupReadOnlyTest = (testFn: (page: DateRangePickerPage, browser: WebdriverIO.Browser) => Promise<void>) => {
+      return useBrowser(async browser => {
+        const params = new URLSearchParams({ readOnly: 'true' });
+        const page = new DateRangePickerPage(createWrapper().findDateRangePicker().getElement(), browser);
+        await browser.url(`#/light/date-range-picker/with-value?${params}`);
+        await page.waitForLoad();
+        await testFn(page, browser);
+      });
+    };
+
+    test(
+      'trigger text is selectable',
+      setupReadOnlyTest(async (page, browser) => {
+        const triggerSelector = page.dateRangePickerTrigger;
+        const triggerText = await page.getTriggerText();
+        expect(triggerText).toContain('2018-01-09');
+
+        // Simulate a real user selection. Unlike programmatically adding a
+        // Range via the Selection API (which works regardless of CSS
+        // `user-select`), a double-click is gated by `user-select: text`
+        // and therefore actually exercises the fix.
+        const element = await browser.$(triggerSelector);
+        await element.doubleClick();
+
+        // Double-click selects a single word, so for a value like
+        // `2018-01-09T00:00:00Z` the selected text will just be `2018`.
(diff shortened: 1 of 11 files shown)


src/internal/components/button-trigger/styles.scss
```diff
@@ -105,6 +105,7 @@ $padding-block-inner-filtering-token: 0px;
   // should retain standard borders, only the icon changes
   &.readonly:not(&.disabled):not(&.in-filtering-token) {
```
jperals, line 106: Note: probably a minor use case, but the `:not(&.in-filtering-token)` excludes from this fix read-only operator selectors in property filter:
Image

At the same time, we would prefer placeholder text to _not_ be selectable, because that comes with collateral a11y issues (if it is interactive, it will have other requirements such as contrast):

Image

jperals:
Would it be possible to add coverage, maybe with the Selection API? https://developer.mozilla.org/en-US/docs/Web/API/Selection

TrevorBurnham:
I've rebased the branch, added test coverage (see the new "Read-only mode" describe block), and addressed your comments

src/internal/components/button-trigger/styles.scss
TrevorBurnham: I've moved the `user-select: text` to a separate `&.readonly:not(&.disabled)` rule that also covers filtering tokens. I've also added `user-select: none` to `.placeholder`.

src/date-range-picker/__integ__/date-range-picker.test.ts
```diff
@@ -98,4 +98,40 @@ describe('Date Range Picker', () => {
...
+        const params = new URLSearchParams({ readOnly: 'true' });
+        const page = new DateRangePickerPage(createWrapper().findDateRangePicker().getElement(), browser);
+        await browser.url(`#/light/date-range-picker/with-value?${params}`);
+        await page.waitForLoad();
+        await testFn(page, browser);
+      });
+    };
+
+    test(
+      'trigger text is selectable via the Selection API',
```
jperals: The test should assert that the text is selectable via user interaction, not via the Selection API. The Selection API can be used for verification.

One way you can select text in tests is by double-clicking, with the caveat that the will only select part of the text (in this case, "2018") when it contains separators such as dashes.

Proposed:
```
    test(
      'trigger text is selectable',
      setupReadOnlyTest(async (page, browser) => {
        const triggerSelector = page.dateRangePickerTrigger;
        const triggerText = await page.getTriggerText();
        expect(triggerText).toContain('2018-01-09');

        const element = browser.$(triggerSelector);
        await element.doubleClick();

        const selectedText = await browser.execute(() => window.getSelection()?.toString());
        expect(selectedText).not.toBeFalsy();
        expect(triggerText).toContain(selectedText);
      })
```

jperals (approved):

PR #4433, opened 2026-04-13
chore: Make vertical space themeable for input related components

at-susie:
### Description
This PR introduces a new spaceFieldVertical design token to control vertical padding inside form input components, and removes the fixed block-size constraint from input and property-filter so that input height is driven by padding + content rather than a hard-coded value. Also makes sizeVerticalInput publicly themeable.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


pages/theming/themed-components.page.tsx
```diff
@@ -422,7 +498,9 @@ export default function ThemedComponentsPage() {
           spaceButtonVertical: '4px',
           borderRadiusButton: '8px',
           borderWidthButton: '1px',
+          //sizeVerticalInput: '30px',
```
georgylobko: Why is this line commented out?

src/flashbar/collapsible.scss
```diff
@@ -16,7 +16,7 @@ $notification-bar-top-overlap: (
   'expanded': awsui.$space-scaled-s,
 );
 $notification-bar-line-height: awsui.$line-height-body-m;
-$notification-bar-padding-vertical: styles.$control-padding-vertical;
+$notification-bar-padding-vertical: awsui.$space-scaled-xxs;
```
georgylobko: why?

src/input/styles.scss
```diff
@@ -53,7 +53,6 @@
   box-shadow: var(#{custom-props.$styleBoxShadowDefault});
 
   @include styles.font-body-m;
-  block-size: awsui.$size-vertical-input;
```
georgylobko, line 56: Why did you remove the block size?

src/prompt-input/__integ__/prompt-input.test.ts
```diff
@@ -87,7 +87,7 @@ describe('Prompt input', () => {
     'Should has one row height in Split Panel',
     setupTest({}, async page => {
       await page.click(createWrapper().findAppLayout().findSplitPanelOpenButton().toSelector());
-      await expect(page.getPromptInputHeight('Prompt-input-in-split-panel')).resolves.toEqual(30);
+      await expect(page.getPromptInputHeight('Prompt-input-in-split-panel')).resolves.toEqual(32);
```
georgylobko, line 90: I want to double check if it's expected

src/flashbar/collapsible.scss
at-susie: Now that `styles.$control-padding-vertical` resolves to the `awsui.$space-field-vertical`, which is now themeable, my intention was to decouple this from that. But I'll double-check this again.

src/input/styles.scss
at-susie: You can see the reason in the description of this PR

src/prompt-input/__integ__/prompt-input.test.ts
at-susie: Yes, it's intentional

pages/theming/themed-components.page.tsx
at-susie: This is a mistake, I will fix

src/input/styles.scss
at-susie: @georgylobko I ended up bringing it back as `min-block-size` after noticing a regression in screenshot tests on my pipeline. When inputs are empty and don’t have placeholder text or values, their height becomes smaller than expected.

Using `$size-vertical-input` as a minimum helps prevent this, acting as a guardrail while still remaining themeable.

src/flashbar/collapsible.scss
at-susie: I ended up bringing the original style back here as I found it's necessary - When the `awsui.$space-field-vertical` is themed, it needs to be applied here as well

pages/theming/themed-components.page.tsx
at-susie: Fixed.

src/input/styles.scss
georgylobko: why `min-block-size` and not `block-size`? do you expect the input field block size to be stretched?

src/prompt-input/__integ__/prompt-input.test.ts
```diff
@@ -38,7 +38,7 @@ describe('Prompt input', () => {
   test(
     'Height should update based on maxRows property',
     setupTest({}, async page => {
-      await expect(page.getPromptInputHeight()).resolves.toEqual(30);
+      await expect(page.getPromptInputHeight()).resolves.toEqual(32);
```
georgylobko, line 41: Doesn't it introduce a regression for existing components?

src/input/styles.scss
at-susie: Regarding the min-block-size vs. block-size, yes, in this PR we are making vertical space on inputs themeable to be able to change the height of them.

src/prompt-input/__integ__/prompt-input.test.ts
at-susie: The height change from 30→32px aligns prompt-input with the standard form control height (sizeVerticalInput: 32px). Previously, prompt-input used spaceScaledXxs (4px / 2px) for its vertical padding while the input height token assumed 5px padding, causing a 2px mismatch. By switching to the shared spaceFieldVertical token, all form controls now derive their height from the same source — which is also what makes the height themeable. The 30px was the inconsistency; 32px is the intended baseline.

georgylobko (approved):

PR #4446, opened 2026-04-15
fix: associate selection inputs with section content

mxschll:
### Description

Selectable cards had no programmatic link between the checkbox input and the card's body content, so screen readers couldn't announce the section text when the input was focused.

Each section content element now gets a unique `id`, and the selection input's `aria-describedby` is set to those IDs. To generate the IDs using `useUniqueId`, we extract the internal `CardItem` component.

Related links, issue #, if available: `AWSUI-61870`

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/cards/index.tsx
+++ b/src/cards/index.tsx
@@ -373,18 +373,18 @@ const CardItem = <T,>({
       >
         {visibleSectionsDefinition.length > 0 &&
           visibleSectionsDefinition.map(({ width = 100, header, content, id }, sectionIndex) => (
-            <div key={id || sectionIndex} className={styles.section} style={{ width: `${width}%` }}>
+            <div
+              key={id || sectionIndex}
+              id={
+                selectionProps && sectionIndex === firstContentSectionIndex
+                  ? `${cardId}-section-${sectionIndex}`
+                  : undefined
+              }
+              className={styles.section}
+              style={{ width: `${width}%` }}
+            >
               {header ? <div className={styles['section-header']}>{header}</div> : ''}
-              {content ? (
-                <div
-                  id={selectionProps ? `${cardId}-section-${sectionIndex}` : undefined}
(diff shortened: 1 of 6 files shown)


gethinwebster:
I wonder whether this approach might result in overly-verbose labelling (with cards with many sections, for example). Instead, perhaps with a `useAsDescription` boolean within each `cardDefinition.sections[]` item. That would match the ActionCard implementation, where only the description gets assigned as aria-describedby, but actual card contents are not

mxschll:
> I wonder whether this approach might result in overly-verbose labelling (with cards with many sections, for example). Instead, perhaps with a `useAsDescription` boolean within each `cardDefinition.sections[]` item. That would match the ActionCard implementation, where only the description gets assigned as aria-describedby, but actual card contents are not

Would it make sense to just use the first section instead? I wonder what use cases would want, e.g. the last section or some section in between to be the description.

gethinwebster:
true, I think first section would also be a reasonable solution, let's go with that

src/cards/index.tsx
```diff
@@ -247,6 +247,151 @@ const Cards = React.forwardRef(function <T = any>(
...
+                selectionProps?.onChange();
+                // Manually move focus to the native input (checkbox or radio button)
+                event.currentTarget.querySelector('input')?.focus();
+              }
+            : undefined
+        }
+      >
+        {visibleSectionsDefinition.length > 0 &&
+          visibleSectionsDefinition.map(({ width = 100, header, content, id }, sectionIndex) => (
+            <div key={id || sectionIndex} className={styles.section} style={{ width: `${width}%` }}>
```
gethinwebster: I think it would make sense to add the 'description' id here, so that the section name is also included (that often gives important context)

gethinwebster (approved):

PR #4448, opened 2026-04-16
chore: Make the font-weight themeable for display large text

at-susie:
### Description
Make the font-weight themeable for the display large text.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


style-dictionary/visual-refresh/typography.ts
```diff
@@ -41,6 +41,7 @@ export const tokens: StyleDictionary.TypographyDictionary = {
   fontWeightHeadingM: '700',
   fontWeightHeadingS: '700',
   fontWeightHeadingXs: '700',
+  fontWeightDisplayLBold: '700',
```
mxschll: Why do we add `Bold` to the token name? This suggests I should not set it to some other value than 700.

src/box/text.scss
```diff
@@ -50,7 +50,6 @@
     }
     &.value-large-variant {
       @include styles.font-display-l;
-      font-weight: awsui.$font-box-value-large-weight;
```
mxschll, line 53: Did you run this through the pipeline? I am unsure if this will cause a regression.

mxschll: Actually, I think it's ok - but still :)

style-dictionary/visual-refresh/typography.ts
at-susie: Good call. I will just simply remove "Bold"

src/box/text.scss
at-susie: Since this change is narrowly scoped—only affecting box and link components when `variant="awsui-value-large"`—it may not be necessary to run it through the full pipeline before merging. Additionally, according to the component feature usage in our dashboard, usage of the `awsui-value-large` variant is minimal, which suggests a low regression risk:
- Links using the awsui-value-large variant account for ~0.3% of all link variants
- Boxes using the awsui-value-large variant account for ~0.1% of all box variants

mxschll (approved):

PR #4459, opened 2026-04-21
chore: Drawer extra features, internal-only

pan-kot:
### Description

A follow-up for https://github.com/cloudscape-design/components/pull/4405

API: [ZuPnAAdHwmum]

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/drawer/__tests__/backdrop.test.tsx
```diff
@@ -0,0 +1,164 @@
...
+  ...jest.requireActual('@cloudscape-design/component-toolkit/internal'),
+  warnOnce: (...args: unknown[]) => warnOnceMock(...args),
+}));
+
+afterEach(() => warnOnceMock.mockReset());
+
+function renderDrawer(jsx: React.ReactElement) {
+  const { container } = render(jsx);
+  const drawer = createWrapper(container).findDrawer()!;
+  const backdrop = createWrapper(container).findByClassName(testClasses.backdrop);
```
pan-kot: Note for the reviewers: the test util methods like findCloseAction() or findBackdrop() are missing yet because all features are added as internal for now. I will refactor all tests to use the new test util methods once the new features are ready for the release.

src/__tests__/snapshot-tests/__snapshots__/test-utils-selectors.test.tsx.snap
```diff
@@ -252,10 +252,10 @@ exports[`test-utils selectors 1`] = `
     "awsui_validation-error_1afkv",
   ],
   "drawer": [
-    "awsui_drawer_1sxt8",
     "awsui_footer_1sxt8",
     "awsui_header-actions_1sxt8",
     "awsui_header_1sxt8",
+    "awsui_root_1sr4m",
```
pan-kot, line 258: This has a risk of breaking consumer e2e tests when the testing infrastructure and the application use different versions of Cloudscape. This problem will get mitigated on its own after both packages are updated.

In this case, I do not see a way how changing selectors can be avoided. The reason the new `root` class replaces the old `drawer` is because now the root element has to include both the drawer and its backdrop.

src/drawer/implementation.tsx
```diff
@@ -17,8 +25,12 @@ import { useStickyFooter } from './use-sticky-footer';
 import { getPositionStyles } from './utils';
 
 import styles from './styles.css.js';
+import testClasses from './test-classes/styles.css.js';
 
-type DrawerInternalProps = NextDrawerProps & InternalBaseComponentProps;
+type DrawerInternalProps = NextDrawerProps &
+  InternalBaseComponentProps & {
+    __ref?: React.Ref<NextDrawerProps.Ref>;
```
pan-kot, line 32: Normally, component refs are propagated using forwardRef util. That does not work for the drawer due to the use of createWidgetizedDrawer in the internal.tsx.

src/drawer/implementation.tsx
```diff
@@ -28,30 +40,93 @@ export function DrawerImplementation({
...
+      : {};
   const containerProps = {
-    ...baseProps,
+    ref: containerRef,
+    ...roleProps,
     style: positionStyles.style,
     className: clsx(
-      baseProps.className,
       styles.drawer,
+      loading && styles['content-with-paddings'],
```
pan-kot, line 83: Previously, the drawer used `loading ? <div ...> : <div ...>` - thus using not just different content and styles, but also rendering separate root nodes. This conflicts with the new semantic drawer roles and focus behaviours. Now, when a drawer gets open - its container element can receive focus. This focus should stay even if the drawer content needs time to load.

src/drawer/implementation.tsx
```diff
@@ -28,30 +40,93 @@ export function DrawerImplementation({
...
     ),
   };
-  const footerRef = useRef<HTMLDivElement>(null);
+
+  const focusIn = useStableCallback(() => {
+    returnFocusTargetRef.current = document.activeElement as HTMLElement;
+    if (containerRef.current && isFocusable(containerRef.current)) {
+      containerRef.current.focus();
+    } else if (containerRef.current) {
+      getAllFocusables(containerRef.current)[0]?.focus();
```
pan-kot, line 96: By default, the drawer auto-focuses its region element when it opens. However, in case the drawer's role is "presentation" - the container can no longer be focused. As a fallback, we focus the first interactive element instead (we do the same in the modal but via using autoFocus={true} prop on the FocusLock).

src/drawer/__tests__/backdrop.test.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+const warnOnceMock = jest.fn();
+jest.mock('@cloudscape-design/component-toolkit/internal', () => ({
+  ...jest.requireActual('@cloudscape-design/component-toolkit/internal'),
+  warnOnce: (...args: unknown[]) => warnOnceMock(...args),
+}));
+
+afterEach(() => warnOnceMock.mockReset());
+
+function renderDrawer(jsx: React.ReactElement) {
```
georgylobko: Let's create a reusable function in utils with all the wrappers like `backdrop`, `closeAction`, etc., so you can swap them out for official test utils in one place later

src/drawer/__tests__/close-action.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+import testClasses from '../../../lib/components/drawer/test-classes/styles.selectors.js';
+
+function renderDrawer(jsx: React.ReactElement) {
+  const { container } = render(jsx);
+  const drawer = createWrapper(container).findDrawer()!;
+  const closeAction = drawer.findByClassName(testClasses['close-action']);
+  return { drawer, closeAction };
+}
+
+test('close button is not rendered when closeAction is not set', () => {
```
georgylobko, line 11: This condition contradicts your interface description:

```
 /**
   * Hides the close action slot next to the header actions, which is present even
   * when close action is not set. Use it when a close action is not needed, or a
   * custom close action implementation is used.
   */
  hideCloseAction?: boolean;
```

georgylobko: In that case, you'd render the close button only when the closeAction prop is provided and skip it otherwise, which satisfies this test condition. But then `hideCloseAction` becomes redundant

src/drawer/__tests__/close-action.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+test('close button is not rendered when hideCloseAction=true', () => {
+  const { closeAction } = renderDrawer(
+    <NextDrawer header="Title" closeAction={{ ariaLabel: 'Close' }} hideCloseAction={true}>
+      content
+    </NextDrawer>
+  );
+  expect(closeAction).toBeNull();
+});
+
+test('close button is not rendered when header is not provided', () => {
```
georgylobko: Why? What is a user wants to show the close button without the header?

src/drawer/__tests__/focus-behaviour.test.tsx
```diff
@@ -0,0 +1,191 @@
...
+      <NextDrawer open={false} onClose={jest.fn()} focusBehavior={{ autoFocus: false }}>
+        content
+      </NextDrawer>
+    );
+    rerender(
+      <NextDrawer open={true} onClose={jest.fn()} focusBehavior={{ autoFocus: false }}>
+        content
+      </NextDrawer>
+    );
+    expect(getDrawerElement(container)).not.toHaveFocus();
```
georgylobko, line 64: Where the focus is expected to be in this case? In the place it was before?

src/drawer/__tests__/focus-behaviour.test.tsx
```diff
@@ -0,0 +1,191 @@
...
+    const trigger = document.createElement('button');
+    document.body.appendChild(trigger);
+    trigger.focus();
+
+    const { rerender } = render(
+      <NextDrawer open={false} onClose={jest.fn()}>
+        content
+      </NextDrawer>
+    );
+    rerender(
```
georgylobko, line 105: let's make sure the focus is shifting to the drawers after re-rendering

src/drawer/__tests__/role.test.tsx
georgylobko, line 1: [nit]: should this file be named as `a11y.test.tsx` instead as it checks not only roles?

src/drawer/interfaces.ts
```diff
@@ -103,6 +132,66 @@ export interface NextDrawerProps extends DrawerProps {
    * Applicable when using `position="sticky"`, `position="absolute"`, or `position="fixed"`.
    */
   zIndex?: number;
+
+  /**
+   * Renders a close button in the header with the provided configuration.
```
georgylobko: It's not clear from the description if it renders a default button is this prop is not defined

src/drawer/__tests__/close-action.test.tsx
pan-kot: When closeAction is not set - we still render the slot for it (a blank space where app layout renders its own close button using absolute position). The hideCloseAction hides that slot.

src/drawer/__tests__/focus-behaviour.test.tsx
pan-kot: Yes, the .focus() action will not get performed - so that focus will stay where it was (i.e. on the trigger)

src/drawer/__tests__/role.test.tsx
pan-kot: I would keep it "role" as it is related to the "role" property behaviour of the drawer component and because it does not test for all a11y issues - only the role-related ones.

src/drawer/__tests__/close-action.test.tsx
pan-kot: Fair point - let me explore

pan-kot: Updated!

src/drawer/__tests__/close-action.test.tsx
georgylobko: what is the value of `hideCloseAction` then?

pan-kot: See a drawer with header + actions in the playground. By default, it has a blank space next to the actions slot, reserved for the close action rendered by app layout as an overlay.

That is the default component behaviour that we cannot change. This slot is not an issue in case `closeAction` is used. However, if the consumer prefers to e.g. have no close action in the header - then the extra blank slot should be hidden.

Screenshot 2026-04-29 at 12 04 31

georgylobko: Understood. Is there a value to keep the empty slot when the close button is not present?

georgylobko: Discussed in DM: the reason for having both hideCloseAction and closeAction is to maintain backward compatibility with existing Drawer use cases

georgylobko (approved):

PR #4467, opened 2026-04-24
fix: Iframe case for tooltip user component

NathanZlion:
Fix Iframe tooltip and Portal usage bugs.

### Description

Related links, issue AWSUI-61932

### How has this been tested?

Created Test Pages and did manual test and added Integ tests too.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/hooks/use-portal-container/index.ts
+++ b/src/internal/hooks/use-portal-container/index.ts
@@ -1,19 +1,10 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
-import { useLayoutEffect, useRef, useState } from 'react';
 
 /**
  * Resolves the portal container to the ownerDocument.body of the referenced element.
  * Ensures portaled content renders in the correct document context (e.g. inside iframes).
  */
-export function usePortalContainer(getElement: () => HTMLElement | SVGElement | null | undefined): HTMLElement | null {
-  const [container, setContainer] = useState<HTMLElement | null>(null);
-  const getElementRef = useRef(getElement);
-  getElementRef.current = getElement;
-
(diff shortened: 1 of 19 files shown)


src/internal/hooks/use-portal-container/index.ts
```diff
@@ -0,0 +1,12 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * Resolves the portal container to the ownerDocument.body of the referenced element.
+ * Ensures portaled content renders in the correct document context (e.g. inside iframes).
+ */
+export function usePortalContainer(
```
jperals, line 8: Why not let the Portal from component-toolkit use this by default? Is it not desired in general to append portal elements to the owner document instead of the child document if using an iframe?

This fix does not include feature prompt and popover, which also use Portal without arguments. Can they not be affected too? Fixing this as default behavior would address them out of the box.

NathanZlion: yes, I noticed the feature prompt and the popover components. Wanted to minimize the PR size and add those as a follow up to make it easy to review.

NathanZlion: > Why not let the Portal from component-toolkit use this by default?

Interesting, I don't see any argument against that. I think that would fix every single usecase no?

NathanZlion: But still defaulting to ownerDocument in the toolkit like [here](https://github.com/cloudscape-design/component-toolkit/commit/7d2d57dca8a5e7f8c25062866186b3af8c4eef6e) solves only half the problem. We still have to attach scroll listener to the right document, instead of glbal window.

PR #4482, opened 2026-04-30
feat: Table Column Groups Table

NathanZlion:
### Description
Add Table Column Grouping Feature support for Table Component.

Related links, issue AWSUI-9594, if available: n/a

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/table/index.tsx
gethinwebster, line 54: we should add a metadata flag here so we can gather metrics on the feature usage

src/table/internal.tsx
```diff
@@ -300,6 +302,28 @@ const InternalTable = React.forwardRef(
       visibleColumns,
     });
 
+    // Build visible column IDs set for grouping
+    const visibleColumnIds = new Set(visibleColumnDefinitions.map((col, idx) => col.id || `column-${idx}`));
+
+    const hierarchicalStructure = useColumnGroups(columnDefinitions, groupDefinitions, visibleColumnIds, columnDisplay);
+
+    const groupLeafMap = useMemo(() => {
```
gethinwebster: I think it would make sense to have this map calculation inside `useColumnGroups` too, to keep as much code out of `internal.tsx` as possible

src/table/sticky-scrolling.ts
```diff
@@ -27,7 +27,11 @@ export default function stickyScrolling(
     if (!item || !containerRef.current || !stickyRef.current) {
       return;
     }
-    const stickyBottom = getLogicalBoundingClientRect(stickyRef.current).insetBlockEnd;
+    // For grouped headers with multiple <tr> rows, stickyRef points to the first <tr>.
+    /* istanbul ignore next: requires DOM scroll measurements */
```
gethinwebster: [minor] it would be more readable to have the two explanatory comments together instead of the istanbul ignore in between them

src/table/thead.tsx
```diff
@@ -80,7 +90,18 @@ const Thead = React.forwardRef(
...
     outerRef: React.Ref<HTMLTableRowElement>
   ) => {
-    const { getColumnStyles, columnWidths, updateColumn, setCell } = useColumnWidths();
+    const { getColumnStyles, columnWidths, updateColumn, updateGroup, setCell } = useColumnWidths();
+
+    /* istanbul ignore next: resize requires real DOM measurements */
+    const handleSplitGroupResize = (leafIds: string[], newWidth: number) => {
+      const lastLeaf = leafIds[leafIds.length - 1];
+      if (lastLeaf) {
+        const currentHalfWidth = leafIds.reduce((sum, id) => sum + (columnWidths.get(id) || 120), 0);
```
gethinwebster: I don't understand what the 'Half' here refers to, isn't this the full width of the group? (ok, now I read further down the file it makes more sense, but I think within the context of this standalone function something like `currentGroupWidth` would be more readable)

src/test-utils/dom/table/index.ts
```diff
@@ -56,7 +74,10 @@ export default class TableWrapper extends ComponentWrapper {
    * @param columnIndex 1-based index of the column containing the resizer.
    */
   findColumnResizer(columnIndex: number): ElementWrapper | null {
-    return this.findActiveTHead().find(`th:nth-child(${columnIndex}) .${resizerStyles.resizer}`);
+    return (
+      this.findActiveTHead().find(`th[data-column-index="${columnIndex}"] .${resizerStyles.resizer}`) ??
```
gethinwebster: I don't think this syntax will work for the auto-generated selector utils? Is the `data-column-index` always present for all grouped and non-grouped columns after the code-change?

src/test-utils/dom/table/index.ts
```diff
@@ -105,8 +126,16 @@ export default class TableWrapper extends ComponentWrapper {
...
 
+  /**
+   * Returns the clickable sorting area of a column header.
+   *
+   * @param colIndex 1-based index of the column.
+   */
   findColumnSortingArea(colIndex: number): ElementWrapper | null {
-    return this.findActiveTHead().find(`tr > *:nth-child(${colIndex}) [role=button]`);
+    return (
+      this.findActiveTHead().find(`th[data-column-index="${colIndex}"] [role=button]`) ??
```
gethinwebster: (same comment as above here)

src/table/table-role/utils.ts
```diff
@@ -63,19 +63,81 @@ export function findTableRowByAriaRowIndex(table: null | HTMLTableElement, targe
 /**
  * Finds the closest column to the targetAriaColIndex+delta in the direction of delta.
  */
+/* istanbul ignore next: requires real DOM layout */
```
gethinwebster: did this previously have no test coverage too?

src/table/resizer/styles.scss
```diff
@@ -46,26 +46,47 @@ th:not(:last-child) > .divider,
...
-// stylelint-disable-next-line selector-combinator-disallowed-list
-th:last-child > .resizer-wrapper.visual-refresh.is-borderless .divider-interactive {
-  inset-inline-end: 0;
+.divider-active {
+  /* used in test-utils */
 }
 
-.divider-active {
-  border-inline-start: $active-separator-width solid awsui.$color-border-divider-active;
+/* stylelint-disable selector-combinator-disallowed-list */
```
gethinwebster: why is this required? is there a possible alternative approach?

src/table/resizer/index.tsx
```diff
@@ -30,15 +30,36 @@ interface ResizerProps {
...
 const RESIZE_THROTTLE = 25;
 const AUTO_GROW_START_TIME = 10;
 const AUTO_GROW_INTERVAL = 10;
 const AUTO_GROW_INCREMENT = 5;
 
-export function Divider({ className }: { className?: string }) {
-  return <span className={clsx(styles.divider, styles['divider-disabled'], className)} />;
+export type DividerPosition = 'default' | 'top' | 'bottom' | 'full';
+
+/* istanbul ignore next */
```
gethinwebster: why is this ignored? (and why didn't it need to be ignored before?)

src/table/header-cell/styles.scss
```diff
@@ -160,6 +174,26 @@ $cell-horizontal-padding: awsui.$space-scaled-l;
...
+  vertical-align: bottom;
+
+  > .header-cell-content {
+    block-size: 100%;
+    box-sizing: border-box;
+    display: flex;
+    flex-direction: column;
+    justify-content: flex-end;
+
+    // stylelint-disable-next-line no-descending-specificity
```
gethinwebster, line 188: strange, this doesn't look like descending specificity, is it needed?

pages/table/column-groups.page.tsx
```diff
@@ -0,0 +1,449 @@
...
+      empty: <Box textAlign="center">No instances</Box>,
+      noMatch: <Box textAlign="center">No matches</Box>,
+    },
+    pagination: { pageSize: 10 },
+    sorting: {},
+    selection: {},
+  });
+
+  return (
+    <SimplePage title="Column Grouping - Feature Combinations" i18n={{}} screenshotArea={{}}>
```
pan-kot: nit: the SimplePage has a `settings` slot - to be used for configuration inputs.

pages/table/column-groups.page.tsx
```diff
@@ -0,0 +1,449 @@
...
+    empty: boolean;
+    cellVerticalAlign: string;
+    sortingDisabled: boolean;
+  }>
+>;
+
+export default function ColumnGroupsPage() {
+  const {
+    urlParams: {
+      direction = 'ltr' as 'ltr' | 'rtl',
```
pan-kot: nit: I recommend not including direction as a config setting - it is best to support it for all pages instead (as a separate improvement), as it is a global setting - not page specific.

pages/table/column-groups.page.tsx
```diff
@@ -0,0 +1,449 @@
...
+              checked={direction === 'rtl'}
+              onChange={({ detail }) => setUrlParams({ direction: detail.checked ? 'rtl' : 'ltr' })}
+            >
+              RTL
+            </Toggle>
+          </SpaceBetween>
+        </SpaceBetween>
+
+        {/* Table */}
+        <div dir={direction}>
```
pan-kot: We can apply direction globally - setting direction=rtl in the url bar already works on other pages. Let's remove this custom solution.

pages/table/column-groups.page.tsx
```diff
@@ -0,0 +1,449 @@
...
+                countText={`${filteredItemsCount} ${filteredItemsCount === 1 ? 'match' : 'matches'}`}
+                filteringPlaceholder="Find instances"
+              />
+            }
+            pagination={<Pagination {...paginationProps} />}
+            empty={<Box textAlign="center">No instances</Box>}
+          />
+        </div>
+      </SpaceBetween>
+      <div style={{ height: '2000px' }} />
```
pan-kot: why is that needed?

src/table/__tests__/column-grouping-rendering.test.tsx
```diff
@@ -0,0 +1,1009 @@
...
+      />
+    );
+    const wrapper = createWrapper(container).findTable()!;
+    const thead = wrapper.find('thead')!;
+    const firstRow = thead.findAll('tr')[0];
+
+    // Focus a header cell
+    const th = firstRow.findAll('th')[0];
+    th.getElement().dispatchEvent(new FocusEvent('focus', { bubbles: true }));
+    // No error thrown — focus handler executed
```
pan-kot: What is this test for - what error can be thrown here?

pan-kot: Let's assert that the focus lands on the expected element - same is true for the onBlur test below

src/table/__tests__/column-grouping-rendering.test.tsx
```diff
@@ -0,0 +1,1009 @@
...
+    );
+    const table = container.querySelector('table')!;
+    const thead = container.querySelector('thead')!;
+    const firstTh = thead.querySelector('th')!;
+
+    // Focus the first header cell
+    firstTh.focus();
+
+    // Press arrow down to navigate to body
+    table.dispatchEvent(new KeyboardEvent('keydown', { key: 'ArrowDown', keyCode: 40, bubbles: true }));
```
pan-kot: Can we use `fireEvent.keydown()`? See src/table/table-role/__tests__/grid-navigation.test.tsx for example

src/table/__tests__/column-grouping-rendering.test.tsx
```diff
@@ -0,0 +1,1009 @@
...
+
+    // Press arrow right to navigate across columns
+    table.dispatchEvent(new KeyboardEvent('keydown', { key: 'ArrowRight', keyCode: 39, bubbles: true }));
+
+    // No errors thrown — navigation handlers executed
+    expect(document.activeElement).toBeDefined();
```
pan-kot: Can we assert the specific element to be focused instead? The `.toBeDefined()` works even if the focus ends up on the document.body

src/table/__tests__/column-grouping-rendering.test.tsx
```diff
@@ -0,0 +1,1009 @@
...
+        groupDefinitions={groupDefinitions}
+        columnDisplay={singleLevelDisplay}
+        resizableColumns={true}
+        {...props}
+      />
+    );
+    return createWrapper(container).findTable()!;
+  }
+
+  test('group resizer triggers updateGroup on drag', () => {
```
pan-kot: nit: updateGroup is an internal util name - the test description should explain the expected behaviour, e.g. "a group can be resized with pointer drag-and-drop"

src/table/column-groups/split-utils.ts
```diff
@@ -0,0 +1,57 @@
...
+  }
+  return childIds;
+}
+
+/**
+ * Determines if a group header cell is split by a sticky column boundary.
+ * Returns null if no split, or the split details if the group straddles a boundary.
+ */
+export function getGroupSplit(
+  col: ColumnInRow<any>,
```
pan-kot: nit: let's update this util API to be an object - otherwise it is hard to read (e.g. getGroupSplit(col, 1, 0, 5))

src/table/column-groups/split-utils.ts
```diff
@@ -0,0 +1,57 @@
...
+
+import { ColumnInRow, HierarchicalStructure } from './utils';
+
+export interface GroupSplit {
+  stickyColspan: number;
+  nonStickyColspan: number;
+  side: 'first' | 'last';
+}
+
+export function getChildColumnIds(hierarchicalStructure: HierarchicalStructure<any>, groupId: string): string[] {
```
pan-kot: nit: should we call this `getGroupColumnIds` (IDs of columns that belong to the given group)?

src/table/column-groups/utils.ts
```diff
@@ -0,0 +1,302 @@
...
+  parentGroupIds: string[];
+  rowIndex: number;
+  colIndex: number;
+}
+
+export interface HeaderRow<T> {
+  columns: ColumnInRow<T>[];
+}
+
+export interface HierarchicalStructure<T> {
```
pan-kot: This name describes the data structure shape (hierarchical) but not its purpose. I recommend changing it to a better meaningful name e.g. "ColumnHeadersDefinition" etc.

src/table/column-groups/split-utils.ts
```diff
@@ -0,0 +1,57 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { ColumnInRow, HierarchicalStructure } from './utils';
+
+export interface GroupSplit {
```
pan-kot: Can this be simplified to `{ stickyColpan: number; staticColspan: number }` - and used for first and last sticky cols separately?

I also recommend to work on the data structure naming to include "sticky" so that its purpose is better clear. Alternatively we can add code comments to help readers understand the purpose and constraints of this structure.

src/table/column-groups/split-utils.ts
```diff
@@ -0,0 +1,57 @@
...
+  stickyColspan: number;
+  nonStickyColspan: number;
+  side: 'first' | 'last';
+}
+
+export function getChildColumnIds(hierarchicalStructure: HierarchicalStructure<any>, groupId: string): string[] {
+  const leafRow = hierarchicalStructure.rows[hierarchicalStructure.rows.length - 1];
```
pan-kot: nit: This is not just "leaf" row - as there are two possible leaf rows - at the beginning/end of the thead. I think we should rename it to "lastRow" or maybe "columnsRow" (meaning it includes only columns, not groups).

src/table/column-groups/split-utils.ts
pan-kot: I also recommend to add a code comment or change the name to emphasise that the function returns not direct but deep group columns.

src/table/column-groups/split-utils.ts
```diff
@@ -0,0 +1,57 @@
...
+ * Returns null if no split, or the split details if the group straddles a boundary.
+ */
+export function getGroupSplit(
+  col: ColumnInRow<any>,
+  stickyColumnsFirst: number,
+  stickyColumnsLast: number,
+  totalLeafColumns: number
+): GroupSplit | null {
+  if (!col.isGroup) {
+    return null;
```
pan-kot: Can we make this more generalized and always return the structure here instead of null (in that case the stickyColumns will equal 0, but then the consumer code will not have to do a null check - so one code branch less).

src/table/column-groups/use-column-groups.tsx
```diff
@@ -0,0 +1,27 @@
...
+export function useColumnGroups<T>(
+  columnDefinitions: ReadonlyArray<TableProps.ColumnDefinition<T>>,
+  groupDefinitions?: ReadonlyArray<TableProps.GroupDefinition>,
+  visibleColumns?: Set<string>,
+  columnDisplay?: ReadonlyArray<TableProps.ColumnDisplayProperties>
+) {
+  return useMemo(() => {
+    const visibleIds = visibleColumns
+      ? Array.from(visibleColumns)
+      : columnDefinitions.map((col, idx) => col.id || `column-${idx}`);
```
pan-kot: We already have the `getColumnKey()` to be used instead of a custom `col.id || column-${idx}`

src/table/column-groups/use-column-groups.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+    return calculateHierarchyTree(
+      [...columnDefinitions],
+      visibleIds,
+      [...(groupDefinitions ?? [])],
+      columnDisplay ? [...columnDisplay] : undefined
```
pan-kot: We do we copy columnDisplay here? Can we change calculateHierarchyTree to accept it as readonly instead? Same for column definitions above.

src/table/column-groups/use-column-groups.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { TableProps } from '../interfaces';
+import { calculateHierarchyTree } from './utils';
+
+export function useColumnGroups<T>(
+  columnDefinitions: ReadonlyArray<TableProps.ColumnDefinition<T>>,
+  groupDefinitions?: ReadonlyArray<TableProps.GroupDefinition>,
+  visibleColumns?: Set<string>,
+  columnDisplay?: ReadonlyArray<TableProps.ColumnDisplayProperties>
+) {
+  return useMemo(() => {
```
pan-kot: Is useMemo really needed here - do we expect calculateHierarchyTree to be heavy in computations?

src/table/column-groups/utils.ts
```diff
@@ -0,0 +1,302 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import { warnOnce } from '@cloudscape-design/component-toolkit/internal';
+
+import { isDevelopment } from '../../internal/is-development';
+import { TableProps } from '../interfaces';
+import { getVisibleColumnDefinitions } from '../utils';
+
+export interface ColumnInRow<T> {
```
pan-kot: nit: should we rename it to HeaderRowColumn? Then HeaderRow and HeaderRowColumn are better symmetrical

src/table/column-groups/utils.ts
```diff
@@ -0,0 +1,302 @@
...
+  id: string;
+  header?: React.ReactNode;
+  colSpan: number;
+  rowSpan: number;
+  isGroup: boolean;
+  columnDefinition?: TableProps.ColumnDefinition<T>;
+  groupDefinition?: TableProps.GroupDefinition;
+  parentGroupIds: string[];
+  rowIndex: number;
```
pan-kot: should rowIndex belong to the HeaderRow?

src/table/column-groups/utils.ts
```diff
@@ -0,0 +1,302 @@
...
+function buildTreeFromColumnDisplay<T>(
+  displayItems: ReadonlyArray<TableProps.ColumnDisplayProperties>,
+  nodeMap: Map<string, TableHeaderNode<T>>,
+  parent: TableHeaderNode<T>
+): void {
+  for (const item of displayItems) {
+    if (item.type === 'group') {
+      const groupNode = nodeMap.get(item.id);
+      if (!groupNode) {
+        if (isDevelopment) {
```
pan-kot: the warnOnce already checks for isDevelopment - there is no need to add a separate check

src/table/column-groups/utils.ts
```diff
@@ -0,0 +1,302 @@
...
+          warnOnce(
+            '[Table]',
+            `Group "${item.id}" referenced in columnDisplay not found in groupDefinitions. Skipping.`
+          );
+        }
+        continue;
+      }
+      buildTreeFromColumnDisplay(item.children, nodeMap, groupNode);
+      if (groupNode.children.length > 0) {
```
pan-kot, line 106: I think this line deserves a code comment to explain why we check for children to be non-empty. Btw, can children array include group nodes - or only column nodes? If it can include groups - then this check is not sufficient.

pan-kot: Also, can groupNode.children include column nodes that are not visible with column display?

src/table/column-groups/utils.ts
```diff
@@ -0,0 +1,302 @@
...
+/**
+ * Fallback when no columnDisplay is provided: all visible columns attach directly to root.
+ */
+function connectFlatColumns<T>(
+  visibleColumns: Readonly<TableProps.ColumnDefinition<T>[]>,
+  nodeMap: Map<string, TableHeaderNode<T>>,
+  root: TableHeaderNode<T>
+): void {
+  for (const col of visibleColumns) {
+    /* istanbul ignore next */
```
pan-kot: Why do we ignore it? We once had a high-sev issue, only reproducible for cases where col.id was not set.

src/table/column-groups/utils.ts
```diff
@@ -0,0 +1,302 @@
...
+    columnDisplay,
+    visibleColumns: visibleColumnIds,
+    columnDefinitions,
+  });
+
+  // Build node map
+  const nodeMap = new Map<string, TableHeaderNode<T>>();
+
+  for (const col of visibleColumns) {
+    if (col.id) {
```
pan-kot, line 197: I would add a code comment explaining why ignoring columns w/o IDs is ok

src/table/column-groups/utils.ts
```diff
@@ -0,0 +1,302 @@
...
+      nodeMap.set(col.id, new TableHeaderNode<T>(col.id, { columnDefinition: col }));
+    }
+  }
+
+  for (const group of groupDefinitions) {
+    nodeMap.set(group.id, new TableHeaderNode<T>(group.id, { groupDefinition: group }));
+  }
+
+  // Build tree
```
pan-kot: nit: The comments like "Build tree" or "Compute layout" are redundant - the corresponding code (`new TableHeaderNode(...)`, `computeSubTreeHeights(root)`) is clear enough.

src/table/column-groups/utils.ts
```diff
@@ -0,0 +1,302 @@
...
+  const root = new TableHeaderNode<T>('*', { isRoot: true });
+
+  if (columnDisplay && columnDisplay.length > 0) {
+    buildTreeFromColumnDisplay(columnDisplay, nodeMap, root);
+  } else {
+    connectFlatColumns(visibleColumns, nodeMap, root);
```
pan-kot: Why do we call this `connectFlatColumns`? Should we call it `buildTreeFromVisibleColumns` - for the symmetry with the other method? This one involves much simpler transformations, but these are transformations nevertheless.

src/table/column-groups/utils.ts
```diff
@@ -0,0 +1,302 @@
...
+
+    if (node.isLeaf && node.columnDefinition && parentChain.length > 0) {
+      columnToParentIds.set(node.id, parentChain);
+    }
+
+    queue.push(...node.children);
+  }
+
+  const rows: HeaderRow<T>[] = Array.from(rowsMap.keys())
+    .sort((a, b) => a - b)
```
pan-kot, line 277: why do we perform sorting twice on the same array?

src/table/header-cell/group-header-cell.tsx
```diff
@@ -0,0 +1,170 @@
...
+import { Divider, Resizer } from '../resizer';
+import { StickyColumnsModel, useStickyCellStyles } from '../sticky-columns';
+import { TableRole } from '../table-role';
+import { DEFAULT_COLUMN_WIDTH, useColumnWidths } from '../use-column-widths';
+import { getStickyClassNames } from '../utils';
+import { TableThElement } from './th-element';
+
+import styles from './styles.css.js';
+
+export interface TableGroupHeaderCellProps {
```
pan-kot: Can we share common props with TableHeaderCellProps?

src/table/header-cell/index.tsx
```diff
@@ -51,6 +51,14 @@ export interface TableHeaderCellProps<ItemType> {
...
   variant: TableProps.Variant;
   tableVariant?: TableProps.Variant;
+  colSpan?: number;
+  rowSpan?: number;
+  /** ID of the direct parent group, forwarded to the <th> as data-column-group-id for test-utils. */
+  columnGroupId?: string;
+  /** When true, this cell is the rightmost child within its parent group. */
+  isLastChildOfGroup?: boolean;
+  /** Determine if the cell is the right most cell of the header */
+  isRightmost?: boolean;
```
pan-kot: We should not call it isRightmost - as it will be left-most when using RTL direction. I suggest calling this isLastColumn - we already use this name inside table's code.

pan-kot: If the word "column" does not quite apply here semantically - maybe we can just use "isLast"

src/table/header-cell/group-header-cell.tsx
```diff
@@ -0,0 +1,170 @@
...
+  // The offset/position comes from stickyColumnId (first child); this only adds boundary classes.
+  const boundaryStyles = useStickyCellStyles({
+    stickyColumns: stickyState,
+    columnId: stickyBoundaryColumnId ?? stickyColumnId ?? groupId,
+    getClassName: props => getStickyClassNames(styles, props),
+    classOnly: true,
+  });
+
+  // Extract only the shadow classes from the boundary subscription
+  /* istanbul ignore next: requires real sticky column state */
```
pan-kot: What is real sticky column state? Does it require some browser behaviours?

src/table/header-cell/th-element.tsx
```diff
@@ -38,6 +38,15 @@ export interface TableThElementProps {
   variant: TableProps.Variant;
   tableVariant?: TableProps.Variant;
   ariaLabel?: string;
+  colSpan?: number;
+  rowSpan?: number;
+  scope?: 'col' | 'colgroup';
+  columnGroupId?: string;
+  isRightmost?: boolean;
+  /** Additional className to merge (e.g. boundary shadow classes from a secondary sticky subscription). */
+  extraClassName?: string;
```
pan-kot: We should call it just `className` - that is commonly used when we need to attach some classes from the outside.

src/table/header-cell/th-element.tsx
```diff
@@ -38,6 +38,15 @@ export interface TableThElementProps {
...
+  /** Additional ref for boundary sticky subscription (imperatively updates shadow classes). */
+  extraRef?: React.RefCallback<HTMLElement>;
```
pan-kot: Why is this called extraRef? Is it not just the ref we attach to the th el?

src/table/sticky-columns/use-sticky-columns.ts
```diff
@@ -139,6 +139,7 @@ interface UseStickyCellStylesProps {
   stickyColumns: StickyColumnsModel;
   columnId: PropertyKey;
   getClassName: (styles: null | StickyColumnsCellState) => Record<string, boolean>;
+  classOnly?: boolean;
```
pan-kot: is this one used?

src/table/table-role/grid-navigation.tsx
```diff
@@ -330,16 +332,49 @@ export class GridNavigationProcessor {
...
+    // Find next cell to focus or move focus into.
+    // Use getAllCellsInRow to include cells from earlier rows that span into the target row via rowspan.
     const targetAriaColIndex = from.colIndex + delta.x;
-    const targetCell = findTableRowCellByAriaColIndex(targetRow, targetAriaColIndex, delta.x);
+    const targetRowAriaIndex = parseInt(targetRow.getAttribute('aria-rowindex') ?? '');
+    let allVisibleCells = getAllCellsInRow(this.table, targetRowAriaIndex);
+    let targetCell =
+      allVisibleCells.length > 0
+        ? findClosestCellByAriaColIndex(allVisibleCells, targetAriaColIndex, delta.x)
+        : /* istanbul ignore next */ findTableRowCellByAriaColIndex(targetRow, targetAriaColIndex, delta.x);
```
pan-kot: Why is this ignored? The keyboard nav should be testable and tested with unit tests

src/table/table-role/grid-navigation.tsx
```diff
@@ -330,16 +332,49 @@ export class GridNavigationProcessor {
...
+      const skipRow = findTableRowByAriaRowIndex(this.table, skipToRowIndex, delta.y);
+      /* istanbul ignore next */ if (!skipRow) {
+        return null;
+      }
+      const skipRowAriaIndex = parseInt(skipRow.getAttribute('aria-rowindex') ?? '');
+      allVisibleCells = getAllCellsInRow(this.table, skipRowAriaIndex);
+      targetCell =
+        allVisibleCells.length > 0
+          ? findClosestCellByAriaColIndex(allVisibleCells, targetAriaColIndex, delta.x)
+          : /* istanbul ignore next */ findTableRowCellByAriaColIndex(skipRow, targetAriaColIndex, delta.x);
```
pan-kot: same here

src/table/table-role/grid-navigation.tsx
```diff
@@ -330,16 +332,49 @@ export class GridNavigationProcessor {
...
+
+    // When vertical movement lands on the same cell (due to rowspan), skip past it.
+    if (targetCell === cellElement && delta.y !== 0 && cellElement) {
+      const cellRow = cellElement.closest('tr');
+      const cellRowIndex = parseInt(cellRow?.getAttribute('aria-rowindex') ?? '0');
+      const cellRowSpan = (cellElement as HTMLTableCellElement).rowSpan || 1;
+      // Jump to the first row after this cell's span (↓) or one row before the cell's start (↑).
+      const skipToRowIndex = delta.y > 0 ? cellRowIndex + cellRowSpan : cellRowIndex - 1;
+      const skipRow = findTableRowByAriaRowIndex(this.table, skipToRowIndex, delta.y);
+      /* istanbul ignore next */ if (!skipRow) {
```
pan-kot: same here

src/table/table-role/grid-navigation.tsx
```diff
@@ -330,16 +332,49 @@ export class GridNavigationProcessor {
...
-      return null;
+    // When horizontal movement lands on the same cell (due to colspan), skip past it.
+    if (targetCell === cellElement && delta.x !== 0 && cellElement) {
+      const cellColIndex = parseInt(cellElement.getAttribute('aria-colindex') ?? '0');
+      const cellColSpan = (cellElement as HTMLTableCellElement).colSpan || 1;
+      const skipToColIndex = delta.x > 0 ? cellColIndex + cellColSpan : cellColIndex - 1;
+      targetCell = findClosestCellByAriaColIndex(allVisibleCells, skipToColIndex, delta.x);
+      if (!targetCell || targetCell === cellElement) {
+        return null;
+      }
```
pan-kot: The changes in this file adds lots of complexity. What is the behaviour w/o this changes?

pan-kot: I think there is an opportunity to significantly simplify the code here. Theoretically, the findCell by next col/row index utils should locate the correct element to focus, and might require changes to be aware of col/row spans. It is not clear why changes in the grid nav logic are necessary.

src/table/table-role/utils.ts
pan-kot: What is the issue with test DOM layout? This does not involve any measurements. E.g. all tests here `src/table/table-role/__tests__/grid-navigation.test.tsx` do not even use our Table component - but run on a custom `<table>` for a better encapsulation.

src/table/table-role/utils.ts
```diff
@@ -63,19 +63,81 @@ export function findTableRowByAriaRowIndex(table: null | HTMLTableElement, targe
...
+  return findClosestCellByAriaColIndex(cellElements, targetAriaColIndex, delta);
+}
+
+/**
+ * Collects all cells visually present in a row, including cells from earlier rows
+ * that span into this row via rowspan. This is needed because cells with rowspan > 1
+ * are only in one <tr> in the DOM but visually occupy multiple rows.
+ */
+export function getAllCellsInRow(table: null | HTMLTableElement, targetAriaRowIndex: number): HTMLTableCellElement[] {
+  /* istanbul ignore next */ if (!table) {
```
pan-kot: We should test this, not ignore

pan-kot: I guess the problem is that we expect table to always be there - so it is kind of hard to make it be null for this util call. In that case we might check for the table before calling the util instead.

src/table/internal.tsx
```diff
@@ -72,7 +73,7 @@ import headerStyles from '../header/styles.css.js';
 import styles from './styles.css.js';
 
 const GRID_NAVIGATION_PAGE_SIZE = 10;
-const SELECTION_COLUMN_WIDTH = 54;
+const SELECTION_COLUMN_WIDTH = 40;
```
pan-kot: Why did we change that?

src/table/internal.tsx
```diff
@@ -300,6 +302,28 @@ const InternalTable = React.forwardRef(
       visibleColumns,
     });
 
+    // Build visible column IDs set for grouping
+    const visibleColumnIds = new Set(visibleColumnDefinitions.map((col, idx) => col.id || `column-${idx}`));
```
pan-kot: We should use getColumnKey() here

src/table/sticky-header.tsx
```diff
@@ -83,11 +97,22 @@ function StickyHeader(
...
           styles.table,
           styles['table-layout-fixed'],
+          hasGroupedColumns && styles['has-grouped-header'],
           contentDensity === 'compact' && getVisualContextClassname('compact-table')
         )}
         ref={secondaryTableRef}
         {...getTableRoleProps({ tableRole })}
       >
+        {hasGroupedColumns && columnDefinitions && (
+          <colgroup>
```
pan-kot: should we also use `<TableColGroup />` here?

src/table/thead.tsx
```diff
@@ -80,7 +90,18 @@ const Thead = React.forwardRef(
     }: TheadProps,
     outerRef: React.Ref<HTMLTableRowElement>
   ) => {
-    const { getColumnStyles, columnWidths, updateColumn, setCell } = useColumnWidths();
+    const { getColumnStyles, columnWidths, updateColumn, updateGroup, setCell } = useColumnWidths();
+
+    /* istanbul ignore next: resize requires real DOM measurements */
```
pan-kot: I would try to find a way and cover it with unit tests to be safe

src/table/thead.tsx
pan-kot: What is 120?

src/table/column-groups/utils.ts
NathanZlion: makes sense,

pages/table/column-groups.page.tsx
NathanZlion: Because wanted to give some space for us to scroll and test sticky header for the table.

src/table/column-groups/use-column-groups.tsx
NathanZlion: yes, calculateHierarchyTree builds a tree via BFS, computes rowSpan/colSpan/colIndex for every node, and constructs the output rows. For tables with many columns and nested groups this might be a bit computation heavy.

src/table/column-groups/utils.ts
NathanZlion: I agree, ColumnGroupsLayout sounds like a descriptive name, wdyt?

src/table/column-groups/utils.ts
NathanZlion: We're not ordering it twice, first make sure the rows are ordered by sorting the row indixes and then we sort the columns within the rows

src/table/header-cell/styles.scss
NathanZlion: It's less specific than the sorting-icon selector above in line 171:
```
  .header-cell-disabled.header-cell-sorted > & {
    & > .sorting-icon {
```

src/table/header-cell/th-element.tsx
NathanZlion: It's not the primary ref, primary as in the only one, so I think it shouldn't be called just ref. we have multiple refs including the `cellRef` and this ref is one or many which get merged with useMergeRefs  and attached to the th. I agree extraRef is maybe too generic and doesn't mean anything, so renamed it to boundaryRef for better clarity.

src/table/sticky-columns/use-sticky-columns.ts
NathanZlion: good catch, my original idea was to not add the inline offset styles from the subscription, and but the group header cells already handles this by only consuming .className and .ref from the result. Removed.

src/table/table-role/utils.ts
NathanZlion: I think that's a good Idea, I'll do that. So that when the util functions are called table is always true.

src/table/internal.tsx
NathanZlion: I don't think it is supposed to be 54 because Initially even though this constant was 54 the selection column had a 40px width using the _awsui.$size-table-selection-horizontal_ token which is 40px. But the thing is with the current implementation I'm using the value.

```
TableColGroup
...
    <colgroup>
      {hasSelection && <col style={{ width: selectionColumnWidth }} />}
      {visibleColumnDefinitions.map((column, colIndex) => {
        const columnId = getColumnKey(column, colIndex);
        return <col key={columnId} data-column-id={String(columnId)} ref={node => setCol(columnId, node)} />;
      })}
    </colgroup>
```

Removing the inline style `style={{ width: selectionColumnWidth }}` for the **col** for selection cell will restore previous behavior, but not sure why this value is set to 54 still? Perhaps any context on this @pan-kot ?

NathanZlion: with that I think I can avoid making this change to the constant by the way,

pan-kot: I don't have a context - but I'd avoid changing this as part of the larger PR. In you need to change that - it is best to open a separate PR and merge it prior to this one to ensure there are no regressions.

src/table/thead.tsx
NathanZlion: It's the DEFAULT_COLUMN_WIDTH from use-column-widths, I imported it and used it the variable instead to make it more clear

src/test-utils/dom/table/index.ts
NathanZlion: Yes, data-column-index is always present on all leaf column headers after this change (both grouped and non-grouped). the nth-child is sort of a fallbcak for backwards compatibilty @gethinwebster

gethinwebster: this won't work for selector utils though, as they need to generate a static selector at runtime without access to the results in order to be able to switch between the two modes

src/table/column-groups/utils.ts
NathanZlion: yeah, children can include both column nodes and nested group nodes. And it wll only attaches a child group if that group also has children.length > 0 because it's recursive. if a group only has hidden columns then it gets discluded.

src/table/utils.ts
```diff
@@ -55,7 +55,7 @@ export function getVisibleColumnDefinitions<T>({
   columnDefinitions,
 }: {
   columnDisplay?: ReadonlyArray<TableProps.ColumnDisplayProperties>;
-  visibleColumns?: ReadonlyArray<string>;
+  visibleColumns?: ReadonlyArray<string | number>;
```
pan-kot: why is this string | number now?

src/table/use-column-widths.tsx
```diff
@@ -195,13 +240,73 @@ export function ColumnWidthsProvider({ visibleColumns, resizableColumns, contain
     setColumnWidths(columnWidths => updateWidths(visibleColumns, columnWidths ?? new Map(), newWidth, columnId));
   }
 
+  /* istanbul ignore next: covered by integration tests, requires real DOM measurements */
```
pan-kot: We have some unit tests for column widths here: src/table/__tests__/columns-width.test.tsx

Can these tests be extended to support the groups use cases?

src/table/thead.tsx
```diff
@@ -80,7 +90,17 @@ const Thead = React.forwardRef(
     }: TheadProps,
     outerRef: React.Ref<HTMLTableRowElement>
   ) => {
-    const { getColumnStyles, columnWidths, updateColumn, setCell } = useColumnWidths();
+    const { getColumnStyles, columnWidths, updateColumn, updateGroup, setCell } = useColumnWidths();
+
+    const handleSplitGroupResize = (leafIds: string[], newWidth: number) => {
```
pan-kot: what are "leaf ids" here - is that columns ids or can it be groups, too?

src/table/selection/selection-control.tsx
```diff
@@ -123,7 +126,12 @@ export function SelectionControl({
         onMouseUp={setShiftState}
         onClick={handleClick}
         htmlFor={controlId}
-        className={clsx(styles.label, styles.root, verticalAlign === 'top' && styles['label-top'])}
+        className={clsx(
+          styles.label,
+          styles.root,
+          verticalAlign === 'top' && !spansRows && styles['label-top'],
+          spansRows && styles['label-bottom']
```
pan-kot: Will this create an inconsistent look? I think the reason we have 'lavel-top' is to ensure the selection element is at the top of the cell when the column headers are multiline (with wrap-lines: true). Now, with col groups + wrap-lines it will look different.

src/table/utils.ts
NathanZlion: https://github.com/cloudscape-design/components/blob/1d90d180c5e1cfefd71edcce6118e9e92c9da830/src/table/utils.ts#L30-L32

The getColumnKey returns column.id || index when a column definition doesn't have an explicit id, it falls back to the numeric array index.

The column groups code path (useColumnGroups > calculateHierarchyTree > getVisibleColumnDefinitions), which previously only accepted string.

https://github.com/cloudscape-design/components/blob/1d90d180c5e1cfefd71edcce6118e9e92c9da830/src/table/column-groups/use-column-groups.tsx#L10-L22

If it's okay I could just cast the type only in the useColumnGroups, and we din't have to do all the | number in all the props.

```
    const visibleIds = visibleColumns
      ? Array.from(visibleColumns)
      : columnDefinitions.map((col, idx) => getColumnKey(col, idx).toString());
```

I'm fine with both

src/table/thead.tsx
NathanZlion: It's a column only, I will rename the leaf to colmn to make it less confusing

src/table/selection/selection-control.tsx
NathanZlion: > Discussed internally to remove this change, and keep the original alignment behavior.

pages/table/column-groups.page.tsx
```diff
@@ -0,0 +1,443 @@
...
+              Loading state
+            </Toggle>
+            <Toggle checked={empty} onChange={({ detail }) => setUrlParams({ empty: detail.checked })}>
+              Empty state
+            </Toggle>
+          </SpaceBetween>
+        </SpaceBetween>
+      }
+    >
+      {/* Table */}
```
pan-kot: nit: I think this comment is not very useful :)

pages/table/column-groups.page.tsx
pan-kot: would it work to increase page size to 20 or 25 instead?

src/table/__tests__/column-groups.test.tsx
```diff
@@ -0,0 +1,1009 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
pan-kot, line 1: nit: these tests verify not only rendering but also other features in combination with column groups, I think we can call the file column-groups.test.tsx, wdyt?

src/table/__tests__/column-groups.test.tsx
```diff
@@ -0,0 +1,1009 @@
...
+    const thead = wrapper.find('thead')!;
+    const groupCells = thead.findAll('th[scope="colgroup"]');
+
+    // Split group should have resizers
+    expect(groupCells.length).toBe(3);
+    const splitGroupCell = groupCells[0];
+    const resizerBtn = splitGroupCell.find('button');
+    if (resizerBtn) {
+      resizerBtn.fireEvent(new PointerEvent('pointerdown', { pointerType: 'mouse', button: 0, bubbles: true }));
+      document.body.dispatchEvent(new PointerEvent('pointerup', { pointerType: 'mouse', bubbles: true }));
```
pan-kot: why is there no assertion for these actions?

src/table/__tests__/columns-width.test.tsx
```diff
@@ -271,3 +334,119 @@ describe('with stickyHeader=true', () => {
...
+    const rightResizerBtn = rightSplitCell.find('button');
+    if (rightResizerBtn) {
+      firePointerdown(new ElementWrapper(rightResizerBtn.getElement()));
+      firePointermove(300);
+      firePointerup(300);
+      expect(onColumnWidthsChange).toHaveBeenCalledTimes(1);
+    }
+  });
+
+  test('renders colgroup with selection col for grouped table', () => {
```
pan-kot: why do we assert this in column-width.test?

src/table/column-groups/__tests__/use-column-groups.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+    test('handles columns without IDs gracefully', () => {
+      const cols: TableProps.ColumnDefinition<any>[] = [{ header: 'No ID', cell: () => 'x' } as any];
+      const { result } = renderHook(() => useColumnGroups(cols, []));
+      expect(result.current.rows).toBeDefined();
+    });
+
+    test('warns in dev when a group referenced in columnDisplay is not in groupDefinitions', () => {
+      const originalEnv = process.env.NODE_ENV;
+      process.env.NODE_ENV = 'development';
+      const warnSpy = jest.spyOn(console, 'warn').mockImplementation();
```
pan-kot: We normally test console warnings by mocking the warnOnce util:

```
import { warnOnce } from '@cloudscape-design/component-toolkit/internal';

jest.mock('@cloudscape-design/component-toolkit/internal', () => ({
  ...jest.requireActual('@cloudscape-design/component-toolkit/internal'),
  warnOnce: jest.fn(),
}));

afterEach(() => {
  (warnOnce as jest.Mock).mockReset();
});

// ...

expect(warnOnce).toHaveBeenCalledWith('Table', '{expected warning message}');
```

src/table/column-groups/use-column-groups.tsx
```diff
@@ -0,0 +1,41 @@
...
+    const visibleIds = visibleColumns
```
pan-kot: Why do we compute visibleIds like that? How is it different from visibleColumns - is it not a set of visible column ids either?

src/table/column-groups/use-column-groups.tsx
```diff
@@ -0,0 +1,41 @@
...
+      ? Array.from(visibleColumns)
+      : columnDefinitions.map((col, idx) => getColumnKey(col, idx).toString());
+
+    const layout = calculateHierarchyTree(columnDefinitions, visibleIds, groupDefinitions ?? [], columnDisplay);
+
+    let groupLeafMap: Map<string, string[]> | undefined;
```
pan-kot: What is group leaf map? It is not quite clear from the var name or the code that generates it.

src/table/column-groups/utils.ts
```diff
@@ -0,0 +1,294 @@
...
+  isRoot?: boolean;
+  colSpan?: number;
+  rowSpan?: number;
+  rowIndex?: number;
+  colIndex?: number;
+}
+
+/**
+ * A node in the table header tree.
+ * - Leaf nodes map to column definitions.
```
pan-kot: Should we call it column instead of leaf in this case? So the isLeaf can become isColumn

src/test-utils/dom/table/index.ts
```diff
@@ -105,7 +126,15 @@ export default class TableWrapper extends ComponentWrapper {
     return this.findByClassName(styles.loading);
   }
 
-  findColumnSortingArea(colIndex: number): ElementWrapper | null {
+  /**
+   * Returns the clickable sorting area of a column header.
+   *
+   * @param colIndex 1-based index of the column.
```
pan-kot, line 131: what options.grouped is for?

pan-kot: can we find one or the other, w/o requiring this param?

src/table/utils.ts
```diff
@@ -104,3 +102,17 @@ export function getStickyClassNames(styles: Record<string, string>, props: Stick
     [styles['sticky-cell-last-inline-end']]: !!props?.lastInsetInlineEnd,
   };
 }
+
+function flattenVisibleColumnIds(items: ReadonlyArray<TableProps.ColumnDisplayProperties>): string[] {
+  const ids: string[] = [];
+  for (const item of items) {
+    if (item.type === 'group') {
+      // ColumnDisplayGroup — recurse into children
```
pan-kot: What is ColumnDisplayGroup and ColumnDisplayItem - these do not correspond to any types we have

src/table/header-cell/group-header-cell.tsx
```diff
@@ -0,0 +1,146 @@
...
+  // Subscribe to the boundary leaf's sticky state to inherit shadow/clip-path classes.
+  // The offset/position comes from stickyColumnId (first child); this only adds boundary classes.
+  const boundaryStyles = useStickyCellStyles({
+    stickyColumns: stickyState,
+    columnId: stickyBoundaryColumnId ?? stickyColumnId ?? groupId,
+    getClassName: props => getStickyClassNames(styles, props),
+  });
+
+  // boundaryStyles.className is populated by scroll/intersection observers in the browser.
+  // In JSDOM these observers don't fire, so this branch is only exercised in integration tests.
```
pan-kot, line 86: the line is covered - only the branches are not, which do not contribute to the required codecov nubmer, let's remove the istanbul ignore.

src/table/thead.tsx
```diff
@@ -91,69 +114,336 @@ const Thead = React.forwardRef(
...
+                />
+              );
+            })}
+          </tr>
+        </thead>
+      );
+    }
+
+    // Grouped columns
+    const totalLeafColumns = columnDefinitions.length;
```
pan-kot: why do we call this "leaf columns"?

src/table/thead.tsx
```diff
@@ -91,69 +114,336 @@ const Thead = React.forwardRef(
...
+            {...(rowIndex === 0 ? focusMarkers.all : {})}
+            ref={rowIndex === 0 ? outerRef : undefined}
+            aria-rowindex={rowIndex + 1}
+            {...getTableHeaderRowRoleProps({ tableRole, rowIndex })}
+            onFocus={event => {
+              const focusControlElement = findUpUntil(event.target, element => !!element.getAttribute('data-focus-id'));
+              const focusId = focusControlElement?.getAttribute('data-focus-id') ?? null;
+              onFocusedComponentChange?.(focusId);
+            }}
+            onBlur={() => onFocusedComponentChange?.(null)}
```
pan-kot: these properties are largely the same as those passed to the single-row header. Should we organise all shared props into some sharedTrProps var?

src/table/thead.tsx
```diff
@@ -91,69 +114,336 @@ const Thead = React.forwardRef(
...
+                  // Group is bisected by the sticky boundary — render two <th> elements.
+                  // Both halves get resizers. Each resizes its own rightmost column child.
+                  const isSplitFirst = splitFirst.stickyColspan > 0;
+
+                  // Left half is sticky for 'first', non-sticky for 'last'
+                  const leftColspan = isSplitFirst ? split.stickyColspan : split.staticColspan;
+                  const leftColIndex = col.colIndex;
+                  const leftGroupId = isSplitFirst ? col.id : `${col.id}__split`;
+                  // Left half's child IDs for resize
+                  const leftChildIds = childIds.filter((_, i) => col.colIndex + i < leftColIndex + leftColspan);
```
pan-kot: instead of filtering by index to find left and right halves, can we do childIds.split(splitIndex)?

src/table/thead.tsx
```diff
@@ -91,69 +114,336 @@ const Thead = React.forwardRef(
...
+                const isAtStickyLastBoundary =
+                  isFullyStickyLast && col.colIndex === columnDefinitions.length - stickyColumnsLast;
+                const fullyStickyBoundaryColumnId = isAtStickyFirstBoundary
+                  ? childIds[childIds.length - 1]
+                  : isAtStickyLastBoundary
+                    ? childIds[0]
+                    : undefined;
+
+                return (
+                  <TableGroupHeaderCell
```
pan-kot, line 363: let's also reuse props via some sharedCellProps and sharedColumnProps

src/table/use-column-widths.tsx
```diff
@@ -195,13 +239,72 @@ export function ColumnWidthsProvider({ visibleColumns, resizableColumns, contain
...
   );
 }
 
+/*
+ * Renders a <colgroup> with <col> elements for each leaf column.
+ * With table-layout:fixed, <col> widths control actual column widths,
+ * which makes colspan headers automatically span the correct width.
+ * Must be rendered inside ColumnWidthsProvider.
+ */
+export function TableColGroup({
```
pan-kot: why do we define this component inside use-column-widths.tsx?

src/table/__tests__/column-groups.test.tsx
NathanZlion: I agree, I'll rename it .

src/table/thead.tsx
NathanZlion: was first meant to distinguish between the columns and group 'columns' but I think it would be okay to just call them columns since groups can called just groups.

pages/table/column-groups.page.tsx
NathanZlion: aah, true, cool will do that.

src/table/column-groups/use-column-groups.tsx
NathanZlion: I think I should find out the place where I used the leaf terminology, and rename them all to Column because it will be less ambiguous.

src/table/column-groups/use-column-groups.tsx
NathanZlion: They're the same thing, accepting the string[] and using it as is.

src/table/use-column-widths.tsx
NathanZlion: It's existence is purely for width purpose mainly. Not sure where to put it, it was initially somewhere else. Moved it here for the sake of keeping it closer to why it's needed. I considered creating a separate file for it,but it's a very small component. I'm open to suggestions,

src/test-utils/dom/table/index.ts
NathanZlion: I don't think there's a better way to do this. data-column-index is only rendered on columns within grouped tables (where :nth-child can't reliably target a specific column due to multi-row headers with varying colspans). For non-grouped tables, :nth-child is the correct approach. Auto-detection isn't reliable for the selector utils (they generate CSS strings, not runtime queries), so the explicit grouped param is needed to pick the right selector strategy. Tagged you into the thread where we discussed this internally.

src/table/use-column-widths.tsx
pan-kot: I would move it into a separate file still

src/table/column-groups/use-column-groups.tsx
pan-kot: Alright, but in which cases this memoization actually works? For instance, the visibleColumns passed to useColumnGroups are computed as:
```
const visibleColumnIds = visibleColumnDefinitions.map((col, idx) => getColumnKey(col, idx).toString());
```
which means its reference is different on every table render

NathanZlion: it's a different reference, true... It doesn't work.In that case I would be more inclined to remove the useMemo then, it's highly unlikely that this calculation takes place frequently because the visible columns has to change for this to happen.

NathanZlion: wdyt @pan-kot ?

src/table/__tests__/column-groups.test.tsx
```diff
@@ -539,7 +539,7 @@ describe('Column grouping sorting', () => {
       />
     );
     const tableWrapper = createWrapper(container).findTable()!;
-    const sortArea = tableWrapper.findColumnSortingArea(3);
+    const sortArea = tableWrapper.find('thead th[data-focus-id="header-type"] [role="button"]');
```
pan-kot: why is this change needed?

NathanZlion: Because of the changes to the findColumnSortingArea to remove the option prop:

https://github.com/cloudscape-design/components/pull/4482/changes/0ae35f1050b6269dfbcb625c641644a69c8b1c78#diff-38a10b6a63280407d9046a04072b6da294a97a0beb7f81db9a70cd87c85c5365

We're now using `tr:not([data-group-level]) > *:nth-child(${colIndex}) [role=button]` which means it would not work for grouped ones right now. Second option, like you mentioned, is `[data-column-index="${colIndex}"] [role=button]` which wouldn't be old version compatible but if failed consumers should upgrade to latest version.
@pan-kot

pan-kot: That is only the 2nd part of the selector that we discussed - this fix was needed so that it does not accidentally match the group use case. I believe the full one should be:
```
th[data-column-index="${colIndex}"] [role=button], tr:not([data-group-level]) > *:nth-child(${colIndex}) [role=button]
```
where one part reliable matches grouped headers, and another part reliably matches the non-grouped headers.

pan-kot (approved):

gethinwebster (approved):

PR #4493, opened 2026-05-07
refactor: use public namespace-qualified types for onLoadItems event …

mxschll:
### Description

The documenter was outputting OptionsLoadItemsDetail (internal interface) as the event detail type for onLoadItems on select, multiselect, and autosuggest. Builders should see the public namespace-qualified types instead.

Override onLoadItems in each public component interface with a namespace-qualified type (SelectProps.LoadItemsDetail, MultiselectProps.LoadItemsDetail, AutosuggestProps.LoadItemsDetail).

**Before**

image

**After**

image

Related links, issue #, if available: `D421195737`

### How has this been tested?

Change went through pipelines and deployed to website.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/autosuggest/interfaces.ts
+++ b/src/autosuggest/interfaces.ts
@@ -19,6 +19,7 @@ export interface AutosuggestProps
     InputClearLabel,
     FormFieldValidationControlProps,
     DropdownStatusProps {
+  onLoadItems?: NonCancelableEventHandler<AutosuggestProps.LoadItemsDetail>;
   /**
    * Specifies an array of options that are displayed to the user as a dropdown list.
    * The options can be grouped using `OptionGroup` objects.
@@ -179,7 +180,8 @@ export namespace AutosuggestProps {
     label?: string;
     options: ReadonlyArray<Option>;
   }
-  export type LoadItemsDetail = OptionsLoadItemsDetail;
+  // eslint-disable-next-line @typescript-eslint/no-empty-object-type
+  export interface LoadItemsDetail extends OptionsLoadItemsDetail {}
   export type StatusType = DropdownStatusProps.StatusType;
   export interface SelectDetail {
     value: string;
--- a/src/autosuggest/interfaces.ts
+++ b/src/autosuggest/interfaces.ts
@@ -180,7 +180,9 @@ export namespace AutosuggestProps {
     label?: string;
     options: ReadonlyArray<Option>;
   }
-  // eslint-disable-next-line @typescript-eslint/no-empty-object-type
+  /* eslint-disable-next-line @typescript-eslint/no-empty-object-type --
+   * Required to create a distinct named type for the documenter.
+   **/
   export interface LoadItemsDetail extends OptionsLoadItemsDetail {}
   export type StatusType = DropdownStatusProps.StatusType;
   export interface SelectDetail {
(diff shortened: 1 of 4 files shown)


src/autosuggest/interfaces.ts
```diff
@@ -179,7 +180,8 @@ export namespace AutosuggestProps {
     label?: string;
     options: ReadonlyArray<Option>;
   }
-  export type LoadItemsDetail = OptionsLoadItemsDetail;
+  // eslint-disable-next-line @typescript-eslint/no-empty-object-type
```
mxschll: This is on purpose.

jperals: Nit: you can explain the reason in the comment: https://eslint.org/docs/latest/use/configure/rules#comment-descriptions

src/multiselect/interfaces.ts
```diff
@@ -86,6 +88,8 @@ export namespace MultiselectProps {
   export type Option = OptionDefinition;
   export type OptionGroup = OptionGroupDefinition;
   export type Options = ReadonlyArray<Option | OptionGroup>;
+  // eslint-disable-next-line @typescript-eslint/no-empty-object-type
```
jperals: Same here

src/select/interfaces.ts
```diff
@@ -224,7 +225,8 @@ export namespace SelectProps {
   export type SelectItem = SelectOptionItem | SelectOptionGroupItem | SelectTriggerOptionItem;
   export type SelectOptionItemRenderer = (props: { item: SelectItem; filterText?: string }) => ReactNode | null;
 
-  export type LoadItemsDetail = OptionsLoadItemsDetail;
+  // eslint-disable-next-line @typescript-eslint/no-empty-object-type
```
jperals: And here

src/autosuggest/interfaces.ts
jperals: Actually, would it be an alternative to turn this into a type?
```
export type LoadItemsDetail = OptionsLoadItemsDetail;
```
It's what we do in other places (for example, `src/multiselect/interfaces.ts`, with Option, OptionGroup and Options).

Docs about types vs interfaces: https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces There seem to be no practical reason to prefer one or the other, and actually the fact that types cannot be extended is a plus.
 
Or would this break existing customers?

mxschll: It was a type before, which didn't work. From my understanding, `type LoadItemsDetail = OptionsLoadItemsDetail` creates a type alias pointing to a named interface. TypeScript resolves through such aliases and outputs the original interface name `OptionsLoadItemsDetail`, not the alias. I think [this part here](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-2.html#:~:text=In%20TypeScript%204.2%2C%20our%20internals%20are%20a%20little%20smarter.%20We%20keep%20track%20of%20how%20types%20were%20constructed%20by%20keeping%20around%20parts%20of%20how%20they%20were%20originally%20written%20and%20constructed%20over%20time.%20We%20also%20keep%20track%20of%2C%20and%20differentiate%2C%20type%20aliases%20to%20instances%20of%20other%20aliases!) might explain the behaviour, but not 100% certain.

jperals (approved):

PR #4512, opened 2026-05-12
feat: Add skeleton component

gethinwebster:
### Description

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


pages/skeleton/components-examples.page.tsx
```diff
@@ -0,0 +1,338 @@
...
+    date: '2026-04-12',
+  },
+];
+
+export default function SkeletonComponentsExamples() {
+  const [isLoading, setIsLoading] = useState(true);
+
+  return (
+    <Box padding="l">
+      <div style={{ position: 'sticky', top: 0, zIndex: 1000, backgroundColor: 'white' }}>
```
jperals: Why the hardcoded background color? And why does it has no space with the next element?

pages/skeleton/permutations.page.tsx
```diff
@@ -0,0 +1,63 @@
...
+    height: ['100px'],
+    width: ['250px'],
+  },
+]);
+
+export default function SkeletonPermutations() {
+  return (
+    <>
+      <h1>Skeleton permutations</h1>
+      <ScreenshotArea disableAnimations={true}>
```
jperals, line 51: Note: assuming this is the only dev page that will be used for visual regression testing, because the other ones do not have animations disabled

pages/skeleton/simple.page.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+export default function SkeletonSimple() {
+  return (
+    <article>
+      <h1>Skeleton demo page</h1>
+      <SpaceBetween size="l">
+        <Container header={<h2>Default skeleton</h2>}>
+          <SpaceBetween size="m">
+            <Skeleton />
+            <p>Default skeleton with no properties (100% width, 3em height)</p>
```
jperals: Isn't the default 1em, not 3em? https://github.com/cloudscape-design/components/pull/4512/changes#diff-cd788456cab9743991aafed593d1d6566087c943f3ae4bed6b7c89cd6580244cR34

src/skeleton/__tests__/skeleton.test.tsx
```diff
@@ -0,0 +1,162 @@
...
+  return createWrapper(container).findSkeleton()!;
+}
+
+describe('Skeleton Component', () => {
+  describe('Basic rendering', () => {
+    it('renders with default props', () => {
+      const wrapper = renderSkeleton();
+      expect(wrapper.getElement()).toBeInTheDocument();
+      expect(wrapper.getElement().tagName).toBe('DIV');
+      expect(wrapper.getElement()).toHaveAttribute('aria-hidden', 'true');
```
jperals, line 25: Nit: this assertion is more specific than "renders with default props". It could have its own ARIA-related test.

src/skeleton/__tests__/skeleton.test.tsx
```diff
@@ -0,0 +1,162 @@
...
+    it('applies height prop to inner element', () => {
+      const wrapper = renderSkeleton({ height: '100px' });
+      const innerElement = wrapper.getElement().querySelector(`.${styles.inner}`);
+      expect(innerElement).toHaveStyle({ blockSize: '100px' });
+    });
+
+    it('applies both width and height', () => {
+      const wrapper = renderSkeleton({ width: '50%', height: '4em' });
+      expect(wrapper.getElement()).toHaveStyle({ inlineSize: '50%' });
+      const innerElement = wrapper.getElement().querySelector(`.${styles.inner}`);
```
jperals, line 44: We are testing some internal implementation in some of these tests —this is for example illustrated by the fact that for height we assert the outer element and for width the inner element. Not a blocker but IMO visual regression testing would be enough for asserting dimensions and even variants.

src/skeleton/styles.scss
```diff
@@ -0,0 +1,121 @@
...
+    @include styles.with-motion {
+      &::after {
+        content: '';
+        position: absolute;
+        inset-block-start: 0;
+        inset-inline-end: 0;
+        inset-block-end: 0;
+        inset-inline-start: 0;
+        background: linear-gradient(90deg, transparent, awsui.$color-background-skeleton-wave, transparent);
+        @include styles.with-motion {
```
jperals, line 57: Is it intended that this statement (`@include styles.with-motion`) is nested within itself?

src/skeleton/styles.scss
```diff
@@ -0,0 +1,121 @@
...
+          animation: skeleton-wave $skeleton-wave-duration $skeleton-wave-easing infinite;
+        }
+      }
+      @include styles.with-direction('rtl') {
+        &::after {
+          @include styles.with-motion {
```
jperals, line 63: Same here, is this additional nested statement needed?

src/skeleton/styles.scss
```diff
@@ -0,0 +1,121 @@
...
+.display-inline-block {
+  display: inline-block;
+  vertical-align: middle;
+}
+.display-inline {
+  display: inline;
+}
+
+.variant-text-body-s {
+  font-size: awsui.$font-size-body-s;
```
jperals, line 84: - Are both font size and line height necessary?
- Will this approach reliably give the box the desired size even if empty spaces are stripped out from the HTML output? This might not work if our code gets transformed from
```
<tag>
  <tag/>
</tag>
```

to
```
<tag><tag/></tag>
```

src/skeleton/styles.scss
gethinwebster: yes it's needed to satisfy the linter (we want the outer statement, so that the pseudoelement doesn't get added if not needed, but then the linter doesn't recognize that the actual animation is "protected" in a with-motion block)

src/skeleton/__tests__/skeleton.test.tsx
gethinwebster: unit tests typically do test internal implementation to some extent, and I think it's better to potentially learn about this sooner rather than having to wait for visual regression tests

src/skeleton/styles.scss
gethinwebster: Yes, both are needed to get the correct size. I've tested with spaces stripped and it seems to also work fine in that scenario.

pages/skeleton/permutations.page.tsx
gethinwebster: yes, the other ones are more for demo/dev purposes

pages/skeleton/components-examples.page.tsx
gethinwebster: will update

pages/skeleton/simple.page.tsx
gethinwebster: the page was overall quite outdated actually, removing it for now as I think it didn't have much value

src/skeleton/__tests__/skeleton.test.tsx
gethinwebster: will update

src/skeleton/styles.scss
jperals: OK. Looks like these rule pairs (font size and line height) should almost always go together across the system, plus there is some overlap between this code and the styles defined in `src/internal/styles/typography/mixins.scss`. What do you tink about moving them there and reuse them?

For example, in `src/internal/styles/typography/mixins.scss`:

```
// New mixin
@mixin text-body-s {
  font-size: awsui.$font-size-body-s;
  line-height: awsui.$line-height-body-s;
}

// New mixin
@mixin .text-body-m {
  font-size: awsui.$font-size-body-m;
  line-height: awsui.$line-height-body-m;
}

(...)

// Existing mixin, refactored
@mixin font-body-s {
  @include text-body-s;
  // Some of the existing mixins contain extra rules besides font size and line height
  letter-spacing: awsui.$letter-spacing-body-s;
}

// Existing mixin, refactored
@mixin font-body-m {
  @include text-body-m;
}
```

And in this new file:

```
.variant-text-body-s {
  @include text-body-s;
}

.variant-text-body-m {
  @include text-body-m;
}

(...)
```

gethinwebster: looking at `font-size-body-s` for example, there are more places where it's used without the lineheight than with, so I think there's not much to be gained from putting this into mixins

jperals (approved):

PR #4514, opened 2026-05-13
chore: Add icon size and stroke width props

at-susie:
### Description
This PR adds:
#### 1. New sizes and strokeWidths props on IconProvider:
Extend the existing IconProvider component with two new props — sizes and strokeWidths — that allow overriding icon dimensions and stroke widths at any level of the component tree.
In a nutshell, this proposal enables icon size and stroke-width customization by overriding the inline size and stroke-width of both the icon SVG and its wrapper <span> element via the existing IconProvider. 

Because the icon-provider already establishes a predictable scoping mechanism, icon sizes can be flexibly adjusted at different levels of the UI while maintaining consistent behavior.
This approach:

* Leverages the existing provider/context pattern used for icon customization
* Supports scoped overrides (global → local)
* Provides independent control over both size and stroke-width per variant

#### 2. A new x-small icon size variant (12px) for the icon component:
While it’s not currently used by any existing components, it supports future use cases we’ve already identified, such as icons inside badge components and icons used in button groups within prompt inputs.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


pages/icon-provider/icon-scale-props.page.tsx
```diff
@@ -0,0 +1,1201 @@
...
+    category: 'Security',
+  },
+  { name: 'Tempor LLP', description: 'aliquet odio. Etiam ligula tortor, dictum eu, placerat eget.', category: 'AI' },
+  { name: 'Egestas Corp', description: 'ridiculus mus. Donec dignissim magna a tortor.', category: 'Serverless' },
+  { name: 'Aenean Inc', description: 'Vivamus molestie dapibus ligula. Aliquam erat volutpat.', category: 'Security' },
+];
+
+function TableAndCards() {
+  const [selectedTableItems, setSelectedTableItems] = useState([tableItems[2]]);
+  const { items, collectionProps } = useCollection(tableItems, { sorting: {} });
```
pan-kot: nit: Do we really need sorting and selection in this demo page?

pages/icon-provider/icon-scale-props.page.tsx
```diff
@@ -0,0 +1,1201 @@
...
+          onChange={({ detail }) => setFilteringText(detail.filteringText)}
+        />
+        <PropertyFilter
+          query={query}
+          onChange={({ detail }) => setQuery(detail)}
+          countText="5 matches"
+          enableTokenGroups={true}
+          expandToViewport={true}
+          filteringAriaLabel="Find distributions"
+          i18nStrings={{
```
pan-kot: We can use I18nProvider instead. The simplest option is to use the SimplePage template as such:
```
<SimplePage title="Icon scale overview" i18n={{}} screenshotArea={{}}>
  ...
</SimplePage>
```

pages/icon-provider/icon-scale-props.page.tsx
```diff
@@ -0,0 +1,1201 @@
...
+            tokenEditorTokenActionsAriaLabel: () => 'Token actions',
+            tokenEditorTokenRemoveAriaLabel: () => 'Remove filter',
+            tokenEditorTokenRemoveLabel: 'Remove filter',
+            tokenEditorTokenRemoveFromGroupLabel: 'Remove filter from group',
+            tokenEditorAddNewTokenLabel: 'Add new filter',
+            tokenEditorAddTokenActionsAriaLabel: 'Add filter actions',
+            tokenEditorAddExistingTokenAriaLabel: () => 'Add filter to group',
+            tokenEditorAddExistingTokenLabel: () => 'Add filter to group',
+          }}
+          filteringOptions={[
```
pan-kot: I believe we only need property filter to show the effect on the close button icons inside the tokens. I recommend removing unneeded filtering properties and filtering options from this test page.

pages/icon-provider/icon-scale-props.page.tsx
```diff
@@ -0,0 +1,1201 @@
...
+
+function TableAndCards() {
+  const [selectedTableItems, setSelectedTableItems] = useState([tableItems[2]]);
+  const { items, collectionProps } = useCollection(tableItems, { sorting: {} });
+
+  return (
+    <SpaceBetween size="l">
+      <Table
+        {...collectionProps}
+        items={items}
```
pan-kot, line 526: nit: you can use shared table configs instead of the custom ones:

```
import { generateItems } from './table/generate-data';
import { columnsConfig } from './table/shared-configs';
```

pages/icon-provider/icon-scale-props.page.tsx
```diff
@@ -0,0 +1,1201 @@
...
+  );
+}
+
+interface ConnectorLinesItem {
+  id: string;
+  label: string;
+  children?: ConnectorLinesItem[];
+}
+
+const connectorLinesItems: ConnectorLinesItem[] = [
```
pan-kot: nit: why do we call these connectorLinesItems and not treeItems, and why is a custom type needed for these?

pages/icon-provider/icon-scale-props.page.tsx
```diff
@@ -0,0 +1,1201 @@
...
+        icon: (
+          <Icon
+            name={expandedItems.includes(item.id) && item.children ? 'folder-open' : item.children ? 'folder' : 'file'}
+          />
+        ),
+        content: item.label,
+      })}
+      getItemId={item => item.id}
+      getItemChildren={item => item.children}
+      onItemToggle={({ detail }) =>
```
pan-kot: nit: for the purposes of this demo I think we can remove the interactivity from the tree-view, wdyt?

pages/icon-provider/icon-scale-props.page.tsx
pan-kot: the same applies to other i18nStrings on this page

pages/icon-provider/icon-scale-props.page.tsx
```diff
@@ -0,0 +1,1201 @@
...
+  const [strokeWidthSmall, setStrokeWidthSmall] = useState('2');
+  const [strokeWidthNormal, setStrokeWidthNormal] = useState('2');
+  const [strokeWidthMedium, setStrokeWidthMedium] = useState('2');
+  const [strokeWidthBig, setStrokeWidthBig] = useState('3');
+  const [strokeWidthLarge, setStrokeWidthLarge] = useState('4');
+
+  // Split panel state
+  const [splitPanelOpen, setSplitPanelOpen] = useState(true);
+
+  const sizeXSmallValue = parseFloat(iconSizeXSmall);
```
pan-kot: nit: this value is used in a single place - you can call parseFloat before passing it to the provider in order to avoid creation of a new var here

src/icon/mixins.scss
```diff
@@ -66,25 +73,37 @@ $_icon-sizes: (
     $padding: calc((#{$supportedLineHeight} - #{$size}) / 2);
 
     &.size-#{$name} {
-      inline-size: $size;
+      /* stylelint-disable-next-line custom-property-pattern */
```
pan-kot: This rule exists in order to avoid using un-hashed css props that consumers can start relying on in their code. See approach used for style API (/internal/generated/custom-css-properties) - can use use it here, too?

src/icon/mixins.scss
```diff
@@ -66,25 +73,37 @@ $_icon-sizes: (
...
+      inline-size: var(--icon-size-override, #{$size});
+      /* stylelint-disable-next-line custom-property-pattern */
+      block-size: max(#{$size}, var(--icon-size-override, #{$size}));
       box-sizing: border-box;
       &-mapped-height {
-        block-size: $supportedLineHeight;
+        //block-size: $supportedLineHeight;
```
pan-kot: is this line a leftover?

pages/icon-provider/icon-scale-props.page.tsx
at-susie: Good point — switched to generateItems + columnsConfig from the shared table configs, which removes the custom selection/sorting setup.

pages/icon-provider/icon-scale-props.page.tsx
at-susie: Thanks for the suggestion. Removed the verbose i18nStrings from PropertyFilter and SplitPanel for now. Will look into adopting SimplePage / I18nProvider as a follow-up.

pages/icon-provider/icon-scale-props.page.tsx
at-susie: Agreed — trimmed the `filteringOptions` and `filteringProperties` down to just what's needed to show the token effect on the close button icons.

pages/icon-provider/icon-scale-props.page.tsx
at-susie: Done — now using generateItems and `columnsConfig` from `pages/table/shared-configs`.

pages/icon-provider/icon-scale-props.page.tsx
at-susie: Good catch — renamed to `treeItems` and `TreeItem`. The custom interface was unnecessary since the shape is the same as what `TreeView` expects.

pages/icon-provider/icon-scale-props.page.tsx
at-susie: Done — removed the `useState` and `onItemToggle` handler, `expandedItems` is now a plain constant.

pages/icon-provider/icon-scale-props.page.tsx
at-susie: Done — removed all the intermediate *Value variables and inlined `parseFloat()` directly in the `IconProvider` props.

src/icon/mixins.scss
at-susie: Good call. Added `iconSizeOverride`, `iconStrokeWidthOverride`, and `iconStrokeScale` to 
`custom-css-properties.js`, regenerated the map, and updated `mixins.scss` to use the hashed variables via `@use '../internal/generated/custom-css-properties'` as custom-props. Also updated `internal.tsx` and the tests to use the map. No more stylelint-disable comments needed.

src/icon/mixins.scss
at-susie: Yes, removed it.

pages/icon-provider/icon-scale-props.page.tsx
pan-kot: nit: I think we can improve it even further by importing shared settings from pages/property-filter/common-props.tsx

src/icon-provider/internal.tsx
```diff
@@ -27,7 +35,75 @@ function InternalIconProvider({ children, icons }: IconProviderProps) {
     iconsToProvide = { ...contextIcons, ...clonedIcons };
   }
 
-  return <InternalIconContext.Provider value={iconsToProvide}>{children}</InternalIconContext.Provider>;
+  // Build the size override map by merging parent context with this provider's sizes prop.
+  const sizeOverridesToProvide = useMemo<IconSizeOverrideMap>(() => {
```
pan-kot: how is that different from `{ ...contextSizeOverrides, ...sizes }`?

src/icon/internal.tsx
```diff
@@ -54,14 +90,69 @@ const InternalIcon = ({
...
+    ...(contextualSize && parentHeight !== null ? { height: `${parentHeight}px` } : {}),
+    ...(targetSizePx !== undefined
+      ? ({ [customCSSPropertiesMap.iconSizeOverride]: `${targetSizePx}px` } as React.CSSProperties)
+      : {}),
+    ...(strokeWidthOverride
+      ? ({ [customCSSPropertiesMap.iconStrokeWidthOverride]: strokeWidthOverride } as React.CSSProperties)
+      : strokeScale
+        ? ({ [customCSSPropertiesMap.iconStrokeScale]: strokeScale } as React.CSSProperties)
+        : {}),
+  };
```
pan-kot: nit, the core above is difficult to read. I recommend refactoring it e.g. as:

```
const styles = {}
if (contextualSize && parentHeight !== null) {
  styles.height = ...
}
if (strokeWidthOverride) {
  styles[customCSSPropertiesMap.iconStrokeWidthOverride] = strokeWidthOverride
}
// ...
```

src/icon-provider/internal.tsx
at-susie: You're right. Simplified both the sizes and strokeWidths merge blocks to { ...contextSizeOverrides, ...sizes } and { ...contextStrokeWidthOverrides, ...strokeWidths } respectively.

src/icon/internal.tsx
at-susie: Done — refactored to the imperative style you suggested:

src/icon-provider/internal.tsx
```diff
@@ -27,7 +35,27 @@ function InternalIconProvider({ children, icons }: IconProviderProps) {
...
+  // Build the stroke-width override map by merging parent context with this provider's strokeWidths prop.
+  const strokeWidthOverridesToProvide = useMemo<IconStrokeWidthOverrideMap>(
+    () => (strokeWidths ? { ...contextStrokeWidthOverrides, ...strokeWidths } : contextStrokeWidthOverrides),
+    [contextStrokeWidthOverrides, strokeWidths]
+  );
+
+  return (
+    <InternalIconContext.Provider value={iconsToProvide}>
+      <InternalIconSizeOverrideContext.Provider value={sizeOverridesToProvide}>
+        <InternalIconStrokeWidthOverrideContext.Provider value={strokeWidthOverridesToProvide}>
```
pan-kot: Why do we need 3 contexts? As icon provider context is internal, we should be able to change its type to include icons, size overrides, and stroke width overrides.

src/icon/internal.tsx
```diff
@@ -54,14 +90,70 @@ const InternalIcon = ({
...
-  const inlineStyles = contextualSize && parentHeight !== null ? { height: `${parentHeight}px` } : {};
+
+  // Compute the target pixel size and stroke-width scale from the override.
+  // Instead of CSS `scale`, we directly set inline-size on both the span and SVG.
+  // The stroke-width scale compensates so the themed stroke remains visually consistent.
+  // The CSS uses: stroke-width: calc(token / cssScaleFactor * var(--icon-stroke-scale, 1))
+  // By setting --icon-stroke-scale to basePx/targetPx, the visual stroke stays at the themed value.
+  let targetSizePx: number | undefined;
+  let strokeScale: number | undefined;
+  if (!contextualSize) {
```
pan-kot: This logic is difficult to read. The strokeScale is only computed when size is not contextual. This is clear from this block, but not from the one below where the inline style is created. Can we refactor the code e.g. as:

```
const inlineStyles: React.CSSProperties = {};
const setVar = (key: string, value: string) => ((inlineStyles as Record<string, unknown>)[key] = value);

if (contextualSize && parentHeight !== null) {
  inlineStyles.height = `${parentHeight}px`;
}
if (!contextualSize) {
  const override = computeSizeOverrides({ sizeOverrides, strokeWidthOverrides, iconSize });
  if (override.size) {
    setVar(customCSSPropertiesMap.iconSizeOverride, `${override.size}px`);
  }
  if (override.strokeWidth) {
    setVar(customCSSPropertiesMap.iconStrokeWidthOverride, override.strokeWidth);
  } else if (override.strokeScale) {
    setVar(customCSSPropertiesMap.iconStrokeScale, override.strokeScale);
  }
}
```

src/icon-provider/internal.tsx
at-susie: That's a great call. I consolidated them into one context

src/icon/internal.tsx
at-susie: Integrated your feedback, extracting a `computeSizeOverrides` helper that takes `{ sizeOverrides, strokeWidthOverrides, iconSize }` and returns `{ size?, strokeWidth?, strokeScale? }`, consolidating all the math in one place. The inline styles block now has a single `if (!contextualSize)` guard that calls the helper and applies the results via a setVar helper, matching your suggestion.

pan-kot (dismissed):

pan-kot (approved):

PR #4531, opened 2026-05-18
feat: add link decoration design tokens

mxschll:
### Description

Adds new tokens for theming link decoration.

### How has this been tested?

~Pipeline is green.~ ~Was green.~ We're green again.

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/link/constants.scss
```diff
@@ -10,39 +10,39 @@
...
   'heading-xl': (
     'text-underline-offset': 0.25em,
-    'text-decoration-thickness': 1px,
+    'text-decoration-thickness': awsui.$font-decoration-thickness-link,
   ),
   'display-l': (
     'text-underline-offset': 0.25em,
-    'text-decoration-thickness': 2px,
+    'text-decoration-thickness': awsui.$font-decoration-thickness-link-display-l,
   ),
```
mxschll, line 42: This one uses 2px, so we use a different token. If necessary, we can split the others up too, but there's currently no need for it.

style-dictionary/visual-refresh/metadata/colors.ts
```diff
@@ -866,6 +866,16 @@ const metadata: StyleDictionary.MetadataIndex = {
     public: true,
     themeable: true,
   },
+  colorTextLinkDecorationDefault: {
+    description: 'The default color of the text decoration (underline) of links. Defaults to the link text color.',
+    public: false,
```
at-susie: What are the reason why we don't have public true?

src/internal/styles/links.scss
```diff
@@ -48,6 +48,7 @@
 @mixin link-font-size-style($font-size) {
   text-underline-offset: map.get($font-size, 'text-underline-offset');
   text-decoration-thickness: map.get($font-size, 'text-decoration-thickness');
+  text-decoration-style: awsui.$font-decoration-style-link;
```
NathanZlion, line 51: I can't find the value underline in the possible set of values for text-decoration-style, but for text-decoration-line.
https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/text-decoration-style

have you tested if this specific part works?

mxschll: This mixin is part of other mixins, where `text-decoration`line` gets set: https://github.com/cloudscape-design/components/blob/88febec316fd462795095a231da8df01b2140c74/src/internal/styles/links.scss#L54-L71

So the underline is there, we want to be able to change how it looks.

mxschll: Changed default to `solid`.

NathanZlion (approved):
Looks Solid 😉  to me

PR #4549, opened 2026-05-22
feat: add one theme style dictionary

mxschll:
### Description

Adds One Theme as a overlay on visual-refresh, gated behind `INCLUDE_ONE_THEME=true`. Set on dev scripts, not on build, so published packages don't include it. Dev page theme switcher has a new toggle for local testing.

Related links, issue #, if available:`aQokArCTx9ab` 

### How has this been tested?

- Verified in local workspace that build output does not include any one theme references.
- Changes wend through pipeline.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/build-tools/utils/themes.js
+++ b/build-tools/utils/themes.js
@@ -3,6 +3,9 @@
 const path = require('path');
 const workspace = require('./workspace');
 
+// One Theme is gated on NODE_ENV so it does not ship in published packages yet.
+const INCLUDE_ONE_THEME = process.env.NODE_ENV !== 'production';
+
 const themes = [
   // This is the default Cloudscape theme, which is best used with Visual Refresh enabled (by default)
   {
@@ -13,7 +16,10 @@ const themes = [
     designTokensPackageJson: { name: '@cloudscape-design/design-tokens' },
     outputPath: path.join(workspace.targetPath, 'components'),
     primaryThemePath: './classic/index.js',
-    secondaryThemePaths: ['./visual-refresh-secondary/index.js', './one-theme/index.js'],
+    secondaryThemePaths: [
+      './visual-refresh-secondary/index.js',
+      ...(INCLUDE_ONE_THEME ? ['./one-theme/index.js'] : []),
+    ],
   },
 ];
 
--- a/build-tools/utils/themes.js
+++ b/build-tools/utils/themes.js
@@ -13,7 +13,7 @@ const themes = [
     designTokensPackageJson: { name: '@cloudscape-design/design-tokens' },
     outputPath: path.join(workspace.targetPath, 'components'),
     primaryThemePath: './classic/index.js',
-    secondaryThemePaths: ['./visual-refresh-secondary/index.js'],
+    secondaryThemePaths: ['./visual-refresh-secondary/index.js', './one-theme/index.js'],
   },
 ];
 
--- a/build-tools/utils/themes.js
+++ b/build-tools/utils/themes.js
@@ -3,7 +3,6 @@
 const path = require('path');
 const workspace = require('./workspace');
 
-// One Theme is gated on this env var so it does not ship in published packages.
 const INCLUDE_ONE_THEME = process.env.INCLUDE_ONE_THEME === 'true';
 
 const themes = [
--- a/build-tools/utils/themes.js
+++ b/build-tools/utils/themes.js
@@ -3,8 +3,8 @@
 const path = require('path');
 const workspace = require('./workspace');
 
-// One Theme is gated on NODE_ENV so it does not ship in published packages yet.
-const INCLUDE_ONE_THEME = process.env.NODE_ENV !== 'production';
+// One Theme is gated on this env var so it does not ship in published packages.
+const INCLUDE_ONE_THEME = process.env.INCLUDE_ONE_THEME === 'true';
 
 const themes = [
   // This is the default Cloudscape theme, which is best used with Visual Refresh enabled (by default)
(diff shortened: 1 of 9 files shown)


style-dictionary/one-theme/colors.ts
```diff
@@ -0,0 +1,14 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import { expandColorDictionary } from '../utils/index.js';
+import { StyleDictionary } from '../utils/interfaces.js';
+
+// Anything not listed here falls back to the visual-refresh value via ThemeBuilder.addTokens in ./index.ts.
+const tokens: StyleDictionary.ColorsDictionary = {
+  colorBackgroundButtonPrimaryDefault: '{colorBlack}',
```
mxschll, line 8: Subsequent PRs will fill this dictionary.

pages/app/components/theme-switcher.tsx
```diff
@@ -21,11 +30,8 @@ export default function ThemeSwitcher() {
     vrSwitchProps.checked = true;
     vrSwitchProps.readOnly = true;
   } else {
-    vrSwitchProps.checked = urlParams.visualRefresh;
-    vrSwitchProps.onChange = event => {
-      setUrlParams({ visualRefresh: event.target.checked });
-      window.location.reload();
-    };
+    vrSwitchProps.checked = urlParams.visualRefresh && !urlParams.oneTheme;
+    vrSwitchProps.onChange = event => activateTheme(event.target.checked ? 'visualRefresh' : 'classic');
```
mxschll, line 33: Same behavior as before: un-toggling visual refresh falls back to classic.

pages/app/components/theme-switcher.tsx
```diff
@@ -34,6 +40,17 @@ export default function ThemeSwitcher() {
         <input {...vrSwitchProps} />
         Visual refresh
       </label>
+      {INCLUDE_ONE_THEME && (
+        <label>
+          <input
+            id="one-theme-toggle"
+            type="checkbox"
+            checked={urlParams.oneTheme}
+            onChange={event => activateTheme(event.target.checked ? 'oneTheme' : 'classic')}
```
mxschll, line 48: Same behavior as with visual refresh: un-toggling falls back to classic.

build-tools/utils/themes.js
```diff
@@ -13,7 +15,10 @@ const themes = [
     designTokensPackageJson: { name: '@cloudscape-design/design-tokens' },
     outputPath: path.join(workspace.targetPath, 'components'),
     primaryThemePath: './classic/index.js',
-    secondaryThemePaths: ['./visual-refresh-secondary/index.js'],
+    secondaryThemePaths: [
+      './visual-refresh-secondary/index.js',
+      ...(INCLUDE_ONE_THEME ? ['./one-theme/index.js'] : []),
```
mxschll, line 20: Without this guard, one theme would be included in all -themeable targets automatically. We don't want this for now.

pages/app/components/theme-switcher.tsx
jperals: VR and One theme are mutually exclusive, right? We'll need some other type of UI for the selection, which doesn't allow enabling both.

pages/app/app-context.tsx
```diff
@@ -12,6 +12,7 @@ interface AppUrlParams {
   density: Density;
   direction: 'ltr' | 'rtl';
   visualRefresh: boolean;
+  oneTheme: boolean;
```
jperals, line 15: Similarly, could we refactor these booleans into a `theme` prop which can take several string values?

pages/app/index.tsx
```diff
@@ -111,6 +111,11 @@ window[awsuiGlobalFlagsSymbol].appLayoutWidget = appLayoutWidget;
 window[awsuiGlobalFlagsSymbol].appLayoutToolbar = appLayoutToolbar;
 window[awsuiCustomFlagsSymbol].appLayoutDelayedWidget = appLayoutDelayedWidget;
 
+// Visual Refresh and One Theme are mutually exclusive — manage both classes here so they never coexist.
+// useRuntimeVisualRefresh() detects .awsui-visual-refresh on body and short-circuits before its Symbol fallback.
+document.body.classList.toggle('awsui-one-theme', oneTheme);
```
jperals, line 116: Similarlt, sohuld we have it be `awsui-theme` with the theme name being the value, so that we don't need to come up with a new class name for each theme in the future and custom logic to make them mutually exclusive?

mxschll: Switching to a `data-awsui-theme` attribute makes more sense than creating a new class per theme. The focus of this PR is adding the style dictionary, so I'll add this in when I enable the new theme in the dev pages more broadly in a follow-up.

pages/app/components/theme-switcher.tsx
mxschll: For now, I want to introduce the style dict without affecting / changing any existing infra or dev pages. That's the reason I kept this separate. When enabling it more broadly, I'll introduce a selector.

pages/app/app-context.tsx
mxschll: Agreed. Pairs naturally with the `data-awsui-theme` switch above. I'll do both together when I enable One Theme generally in the dev pages.

pages/app/components/theme-switcher.tsx
jperals: Can we then include only the style dictionary changes in this PR?

jperals (approved):

PR #4553, opened 2026-05-26
feat: Left panel without toolbar in AppLayout component

georgylobko:
### Description

`AWSUI-62013`
Enabled the left panel for AppLayout and AppLayoutToolbar when no toolbar is present. Adjusted the design so borders and rounded corners render consistently with the toolbar version

image
image

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


pages/app-layout-toolbar/without-toolbar-nested.page.tsx
georgylobko, line 1: I will introduce a screenshot test on this page

src/app-layout/visual-refresh-toolbar/state/props-merger.ts
```diff
@@ -42,7 +42,7 @@ export const mergeProps: MergeProps = (ownProps, additionalProps) => {
       toolbar.onActiveGlobalDrawersChange = props.onActiveGlobalDrawersChange;
     }
     if (
-      props.aiDrawer &&
+      props.aiDrawer?.trigger &&
```
georgylobko, line 45: these are toolbar only props, so if the drawer is trigger-less, its props shouldn't be forwarded to the toolbar

src/app-layout/visual-refresh-toolbar/state/use-app-layout.tsx
```diff
@@ -239,7 +240,7 @@ export const useAppLayout = (
     }
   };
 
-  useWidgetMessages(hasToolbar, message => {
+  useWidgetMessages(hasToolbar || forceEnableRuntimeMessages, message => {
```
georgylobko, line 243: Introduced an opt-in flag here. I considered detecting whether the AppLayout is the outermost instance in nested layouts, but that doesn't work with isolated iframes and would introduce a regression

src/app-layout/visual-refresh-toolbar/skeleton/styles.scss
```diff
@@ -214,6 +214,64 @@
...
+
+.pseudo-toolbar-content {
+  position: relative;
+
+  @include theming.dark-mode-only {
+    border-block-start: awsui.$border-divider-section-width solid awsui.$color-border-layout;
+    border-start-start-radius: awsui.$space-xxs;
+  }
+
+  &:before,
```
georgylobko, line 248: These styles are for showing a round-corner for content area. Since the pseudo toolbar should not occupy any space, there is no way to apply styles to it directly

pages/app-layout-toolbar/without-toolbar-nested-iframe.page.tsx
```diff
@@ -0,0 +1,149 @@
...
+  };
+
+  return (
+    <ScreenshotArea gutters={false}>
+      <AppLayoutToolbar
+        {...{ __forceEnableRuntimeMessages: true }}
+        ariaLabels={{ ...appLayoutLabels, ...drawerLabels }}
+        ref={pageLayoutRef}
+        content={
+          <IframeWrapper
```
pan-kot: Can we add iframe wrapper with a query parameter to avoid having two near-identical pages?

src/__a11y__/run-a11y-tests.ts
```diff
@@ -34,6 +34,8 @@ export default function runA11yTests(theme: Theme, mode: Mode, skip: string[] =
         // this page intentionally has issues to test the helper
         'undefined-texts',
         'app-layout/with-error-boundaries',
+        'app-layout-toolbar/without-toolbar-nested',
```
pan-kot, line 38: what is the reason for excluding these?

georgylobko: Because nested app layouts aren't accessible and there is no good way to fix them

pages/app-layout-toolbar/without-toolbar-nested-iframe.page.tsx
georgylobko: yes, added a query param to `pages/app-layout-toolbar/without-toolbar-nested.page.tsx` and removed this page

src/__a11y__/run-a11y-tests.ts
pan-kot: should we add a comment explaining the accessibility issue?

georgylobko: added

pan-kot (approved):

PR #4559, opened 2026-05-27
feat: Global theming APIs

fralongo:
### Description

### How has this been tested?

<details>

#### Correctness

- _Changes are backward-compatible if not indicated, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#public-apis)._ Yes
- _Changes do not include unsupported browser features, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#browsers-support)._ Yes
- _Changes were manually tested for accessibility, see [accessibility guidelines](https://cloudscape.design/foundation/core-principles/accessibility/)._ N/A

#### Security

#### Testing

- _Changes are covered with new/existing unit tests?_ Yes
- _Changes are covered with new/existing integration tests?_ Yes
</details>

---

### DIFF
--- a/src/theming/index.ts
+++ b/src/theming/index.ts
@@ -67,9 +67,6 @@ export function setGlobalTheme(theme: Theme): void {
 }
 
 function getGlobalTheme(): Theme | undefined {
-  if (typeof window === 'undefined') {
-    return undefined;
-  }
   const topWindow = getTopWindow();
   return topWindow[themeStorageKey];
 }
--- a/src/theming/index.ts
+++ b/src/theming/index.ts
@@ -38,3 +38,75 @@ export function generateThemeStylesheet({ theme, baseThemeId }: GenerateThemeSty
     baseThemeId,
   });
 }
+
+const themeStorageKey = Symbol.for('awsui-global-theme');
+const themeChangeEvent = 'awsui-global-theme-change';
+
+interface WindowWithTheme extends Window {
+  [themeStorageKey]?: Theme;
+}
+
+function getTopWindow(): WindowWithTheme {
+  try {
+    if (window.top) {
+      return window.top as unknown as WindowWithTheme;
+    }
+  } catch {
+    // Cross-origin access error — fall back to current window.
+  }
+  return window as unknown as WindowWithTheme;
+}
+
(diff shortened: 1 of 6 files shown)


src/theming/index.ts
```diff
@@ -38,3 +38,72 @@ export function generateThemeStylesheet({ theme, baseThemeId }: GenerateThemeSty
...
+function getTopWindow(): WindowWithTheme {
+  try {
+    if (window.top) {
+      return window.top as WindowWithTheme;
+    }
+  } catch {
+    // Cross-origin access error — fall back to current window.
+  }
+  return window as WindowWithTheme;
+}
```
jkuelz, line 58: Is there a reason we go straight to the top vs using the [`findUpAPI`](https://github.com/cloudscape-design/components/blob/9b5fbfbbf37f95c828c793fb1c400e3345fe183a/src/internal/plugins/api.ts#L38-L58) that already exists for plugins?

We also use [`getTopWIndow`](https://github.com/cloudscape-design/components/blob/9b5fbfbbf37f95c828c793fb1c400e3345fe183a/src/error-boundary/utils.ts#L24-L27) in our ErrorBoundary component.

There is also a key piece missing here that both of those have: the cross-origin SecurityError check -- which is only thrown when accessing writable properties on the window directly. We need to add that to the first statement to make the catch have any effect. 

```
function getTopWindow(): WindowWithTheme {
    try {
      if (window.top && window.top.document) { // <-- this forces cross-orign to return the security error here
        return window.top as WindowWithTheme;
      }
    } catch {
      // Cross-origin top — fall back to current window.
    }
    return window as WindowWithTheme;
```

Should we use a shared utility here? We could've prevented this bug (and any future fixes that we may need to make).

jkuelz (commented):
Looks good, just one comment on the getTopWindow fn.

src/theming/index.ts
fralongo: I think that findUp would only work for the getter, not for the setter. I did take the recommendation to add an extra check.

I can definitely create a shared utility

jkuelz (approved):

PR #4566, opened 2026-05-31
chore: Fix downstream build failures introduced by #4514

at-susie:
### Description
Fixes a compatibility issue that caused a downstream package build to fail after #4514 added `size` and `strokeWidth` props to `IconProvider`.

When the `IconProvider` context is not present, `sizeOverrides` and `strokeWidthOverrides` can be undefined, causing `computeSizeOverrides` to crash with TypeError: Cannot read properties of undefined. To fix this, I added a null guard that returns early when either map is undefined.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/icon/internal.tsx
```diff
@@ -79,6 +79,11 @@ function computeSizeOverrides({
 }): SizeOverrideResult {
   const result: SizeOverrideResult = {};
 
+  // Guard against stale context consumers that don't provide these maps yet.
```
SpyZzey: I would remove this comment

src/icon/mixins.scss
```diff
@@ -75,18 +76,18 @@ $_icon-sizes: (
 
     &.size-#{$name} {
       inline-size: var(#{custom-props.$iconSizeOverride}, #{$size});
-      block-size: max(#{$size}, var(#{custom-props.$iconSizeOverride}, #{$size}));
+      block-size: string.unquote('max(#{$size}, var(#{custom-props.$iconSizeOverride}, #{$size}))');
```
SpyZzey: I don't think this will fix the issue on the consumers end. This file already gets converted to a css file (icon/styles.scoped.css) after building the package. Therefore, it will result in the same css as before.

src/icon/internal.tsx
at-susie: Removed the comment

src/icon/internal.tsx
```diff
@@ -79,6 +79,11 @@ function computeSizeOverrides({
...
+  if (!sizeOverrides || !strokeWidthOverrides) {
+    return result;
+  }
```
SpyZzey: Is it expected that it will return an empty object for when either sizeOverrides or strokeWidthOverrides is undefined? Maybe this should be something like

```suggestion
  if (!sizeOverrides) {
    sizeOverrides = {};
  }
  
  if (!strokeWidthOverrides) {
    strokeWidthOverrides = {};
  }
```
instead. This way, when sizeOverrides is not defined, the strokeWidthOverrides will still be applied (and vice-versa).

src/icon/mixins.scss
at-susie: Reverted the change.

src/icon/internal.tsx
at-susie: That's better. Integrated your suggestion

src/icon-provider/__tests__/icon-provider.test.tsx
```diff
@@ -417,4 +417,31 @@ describe('Icon Provider', () => {
       expect(secondIcon).toStrictEqual(expectedSvg);
     });
   });
+
+  describe('missing context fields (null guard)', () => {
+    it('renders without crashing when no IconProvider is present', () => {
+      const { container } = render(<Icon name="calendar" size="normal" />);
+      expect(container.querySelector('[class*="icon"]')).not.toBeNull();
```
SpyZzey: Can we use the Cloudscape test utils here? e.g., 
```
const wrapper = createWrapper(container)
wrapper.findIcon()
```

So that we don't depend on internals.

src/icon-provider/__tests__/icon-provider.test.tsx
```diff
@@ -417,4 +417,31 @@ describe('Icon Provider', () => {
...
+    });
+
+    it('still applies strokeWidths when sizeOverrides is an empty map', () => {
+      const { container } = render(
+        <IconProvider icons={{}} strokeWidths={{ normal: 2 }}>
+          <Icon name="calendar" size="normal" />
+        </IconProvider>
+      );
+      const iconEl = container.querySelector('[class*="icon"]') as HTMLElement;
```
SpyZzey: same here

src/icon-provider/__tests__/icon-provider.test.tsx
```diff
@@ -417,4 +417,31 @@ describe('Icon Provider', () => {
...
+      expect(iconEl.style.getPropertyValue(customCSSPropertiesMap.iconStrokeWidthOverride)).toBe('2px');
+    });
+
+    it('still applies sizes when strokeWidthOverrides is an empty map', () => {
+      const { container } = render(
+        <IconProvider icons={{}} sizes={{ normal: 12 }}>
+          <Icon name="calendar" size="normal" />
+        </IconProvider>
+      );
+      const iconEl = container.querySelector('[class*="icon"]') as HTMLElement;
```
SpyZzey: and same here

src/icon/internal.tsx
```diff
@@ -79,6 +79,13 @@ function computeSizeOverrides({
 }): SizeOverrideResult {
   const result: SizeOverrideResult = {};
 
+  if (!sizeOverrides) {
```
jperals, line 82: Why can `sizeOverrides` and `strokeWidthOverrides` be undefined in the first place if they are required in the function signature?

src/icon-provider/__tests__/icon-provider.test.tsx
at-susie: Good call. Updated to use `expect(wrapper(container).findIcon())`

src/icon-provider/__tests__/icon-provider.test.tsx
at-susie: Good call. Updated

src/icon-provider/__tests__/icon-provider.test.tsx
at-susie: Good call. Updated

src/icon/internal.tsx
at-susie: Good point. Updated the parameter types to | undefined to make this explicit.

PR #4579, opened 2026-06-03
feat: property filter custom operator support

revezard:
### Description

  Adds end-to-end support for custom operators in the Property Filter component. Consumers can define their own operator strings with custom matching logic, display descriptions, and accessible labels.

  Key changes:
  - `getAllowedOperators` now preserves custom operators (appended after known operators in standard display order)
  - `getAllowedFreeTextOperators` updated to handle extended operator objects and preserve custom operators
  - `operatorToDescription` checks custom `description` before built-in i18n, allowing consumers to override predefined operator labels (e.g. `>` described as "After" for date properties)
  - `matchOperator` requires word boundary after text-based operators to prevent false matches (e.g. "in" not matching "internal")
  - Token editor resolves custom operator descriptions from filtering properties when in free-text mode
  - Added `getOperatorDescription` accessor to `InternalFilteringProperty`

  Depends on: https://github.com/cloudscape-design/collection-hooks/pull/151

  ### How has this been tested?

  - Unit tests for `operatorToDescription` (custom description override, fallback to i18n, unknown operators)
  - Unit tests for `getAllowedOperators` (custom operators preserved, ordering, duplicates)
  - Unit tests for `matchOperator` (text-based operator word boundary, case-insensitive matching, symbolic operators unaffected)
  - Unit tests for `parseText` (text-based operators with space, case-insensitive, no false match on value prefix)
  - Integration test for autosuggest options (custom description in dropdown, override of predefined operator descriptions)
  - Demo page at `/#/light/property-filter/property-filter-custom-operator` exercising all scenarios

  <details>

  #### Correctness

  #### Security

  #### Testing
  
  </details>

---

### DIFF
--- a/src/property-filter/controller.ts
+++ b/src/property-filter/controller.ts
@@ -91,20 +91,20 @@ export const getQueryActions = ({
 
 const knownOperatorOrder = ['=', '!=', ':', '!:', '^', '!^', '>=', '<=', '<', '>'] as const;
 
+function orderOperators(operatorSet: Set<ComparisonOperator>): ComparisonOperator[] {
+  const knownOperators = knownOperatorOrder.filter(op => operatorSet.has(op));
+  const customOperators = [...operatorSet].filter(op => !(knownOperatorOrder as readonly string[]).includes(op));
+  return [...knownOperators, ...customOperators];
+}
+
 export const getAllowedOperators = (property: InternalFilteringProperty): ComparisonOperator[] => {
   const { operators = [], defaultOperator } = property;
-  const operatorSet = new Set([defaultOperator, ...operators]);
-  const knownOperators = knownOperatorOrder.filter(op => operatorSet.has(op));
-  const customOperators = operators.filter(op => !(knownOperatorOrder as readonly string[]).includes(op));
-  return [...knownOperators, ...customOperators];
+  return orderOperators(new Set([defaultOperator, ...operators]));
 };
 
 export const getAllowedFreeTextOperators = (freeText: InternalFreeTextFiltering): ComparisonOperator[] => {
   const operators = freeText.operators.map(op => (typeof op === 'string' ? op : op.operator));
-  const operatorSet = new Set(operators);
-  const knownOperators = knownOperatorOrder.filter(op => operatorSet.has(op));
-  const customOperators = operators.filter(op => !(knownOperatorOrder as readonly string[]).includes(op));
-  return [...knownOperators, ...customOperators];
+  return orderOperators(new Set(operators));
 };
 
 /*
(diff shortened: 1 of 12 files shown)


src/property-filter/controller.ts
```diff
@@ -94,17 +94,22 @@ export const getQueryActions = ({
...
 
-const operatorOrder = ['=', '!=', ':', '!:', '^', '!^', '>=', '<=', '<', '>'] as const;
+const knownOperatorOrder = ['=', '!=', ':', '!:', '^', '!^', '>=', '<=', '<', '>'] as const;
 
 export const getAllowedOperators = (property: InternalFilteringProperty): ComparisonOperator[] => {
   const { operators = [], defaultOperator } = property;
   const operatorSet = new Set([defaultOperator, ...operators]);
-  return operatorOrder.filter(op => operatorSet.has(op));
+  const knownOperators = knownOperatorOrder.filter(op => operatorSet.has(op));
+  const customOperators = operators.filter(op => !(knownOperatorOrder as readonly string[]).includes(op));
```
gethinwebster: maybe we should use `operatorSet` instead of `operators` here: technically `operators` could have duplicates in it which we probably want to avoid

src/property-filter/controller.ts
```diff
@@ -94,17 +94,22 @@ export const getQueryActions = ({
...
+  return [...knownOperators, ...customOperators];
 };
 
 export const getAllowedFreeTextOperators = (freeText: InternalFreeTextFiltering): ComparisonOperator[] => {
-  const operatorSet = new Set(freeText.operators);
-  return operatorOrder.filter(op => operatorSet.has(op));
+  const operators = freeText.operators.map(op => (typeof op === 'string' ? op : op.operator));
+  const operatorSet = new Set(operators);
+  const knownOperators = knownOperatorOrder.filter(op => operatorSet.has(op));
```
gethinwebster: as this joining/ordering isn't entirely straightforward and is repeated twice, would be good to refactor it into a helper

gethinwebster (commented):
Also looks like there are some axe a11y violations on this page:     ✕ #/light/property-filter/property-filter-editor-permutations?visualRefresh=false (1689 ms)

(and/or possibly "// Report if there are any occurrences of string "undefined" in HTML.", the failure outputs are a bit hard to read...)

You can test this locally by running `npm start` in one terminal, then `npx jest -c jest.integ.config.js a11y` in another, narrowing down which pages are tested by temporarily editing this file locally https://github.com/cloudscape-design/components/blob/ff19e2c2adbafc4bb615c3e0c04e5dee93c51a94/src/__integ__/utils.ts#L9

gethinwebster (approved):

Issue #4551, opened 2026-05-26
[Bug]: allowSkipTo doesn't enable navigation to immediately-next step when current step is optional


abrasive519mars:
### Browser

Safari

### Package version

3.0.1011

### React version

18.2.0

### Description

- canSkip(fromIndex, toIndex) returns false when fromIndex >= toIndex (line: if (fromIndex >= toIndex) return false)
- When on step N-1 targeting step N, it calls canSkip(N, N) which hits this guard
- The fallback only allows it if the target is isOptional, but the target being mandatory shouldn't prevent navigation when the current step is optional
- This creates inconsistency: Step 4 is clickable from Steps 1 and 2 (via intermediate optional checks) but not from Step 3 (the step right before it)
- Expected: If I can skip to Step 4 from Step 1 (because Steps 2/3 are optional), I should also be able to skip to Step 4 from Step 3 (since the current step is optional and there are no intermediates to block)
- Workaround: Mark the target step as isOptional: true, but this adds an unwanted "- optional" label

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/cloudscape-design/components/blob/main/CODE_OF_CONDUCT.md)
- [x] I checked the [current issues](https://github.com/cloudscape-design/components/issues) for duplicate problems

SpyZzey:
Hey,
Thanks for submitting this bug report.

I was able to reproduce the bug. Can you verify that this is what you mean (see video below)?

Best regards,

Simon

SpyZzey:
https://github.com/user-attachments/assets/82628e2b-5053-4242-b384-68f618539c10

abrasive519mars:
Hello,
Yes, this is exactly it. Please do look into it as this is a component in our production console. Thanks for the quick repro!

SpyZzey:
@abrasive519mars We will track this ticket internally and update this GitHub issue once it's
  resolved.

amanabiy:
Hi @abrasive519mars, this is now fixed and released in version 3.0.1308. Thanks for reporting this!

PR #4581, opened 2026-06-03
fix: Allow skip-to navigation to adjacent step when current step is optional

amanabiy:
### Description
Fixes an inconsistency in the wizard's allowSkipTo navigation where the immediately-next step was unreachable from an optional step, even though it was reachable from earlier steps. 

Related links, issue #, if available: https://github.com/cloudscape-design/components/issues/4551, AWSUI-62034

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing
Unit test + Manual testing with a wizard where steps 2 and 3 are optional and step 4 is required. Verified step 4 is clickable from all preceding steps.
</details>

---

### DIFF
--- a/src/wizard/wizard-step-list.tsx
+++ b/src/wizard/wizard-step-list.tsx
@@ -49,12 +49,12 @@ export function getStepStatus(
     return StepStatusValues.Visited;
   }
   if (allowSkipTo && index > activeStepIndex) {
-    // Can we skip to this step? (all steps between current and this one are optional)
+    // All steps between current and target are optional — can skip over them
     if (canSkip(activeStepIndex + 1, index, steps)) {
       return StepStatusValues.Next;
     }
-    // Immediate next step is also navigable if it's optional
-    if (index === activeStepIndex + 1 && steps[index]?.isOptional) {
+    // If either the current step or the next step is optional, we can navigate to it
+    if (index === activeStepIndex + 1 && (steps[index]?.isOptional || steps[activeStepIndex]?.isOptional)) {
       return StepStatusValues.Next;
     }
   }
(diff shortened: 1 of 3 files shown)


src/wizard/wizard-step-list.tsx
```diff
@@ -49,12 +49,12 @@ export function getStepStatus(
...
   if (allowSkipTo && index > activeStepIndex) {
-    // Can we skip to this step? (all steps between current and this one are optional)
+    // All steps between current and target are optional — can skip over them
     if (canSkip(activeStepIndex + 1, index, steps)) {
       return StepStatusValues.Next;
     }
-    // Immediate next step is also navigable if it's optional
-    if (index === activeStepIndex + 1 && steps[index]?.isOptional) {
+    // If either the current step or the next step is optional, we can navigate to it
+    if (index === activeStepIndex + 1 && (steps[index]?.isOptional || steps[activeStepIndex]?.isOptional)) {
```
jperals, line 57: Is it expected that we can navigate to the next step if the next step is optional but not the current one? Why should we be able to skip the current step if it is not optional?

Should it be instead
```
    // All steps between current and target are optional — can skip over them
    if (canSkip(activeStepIndex, index, steps)) {
      return StepStatusValues.Next;
    }
    // If either the current step or the next step is optional, we can navigate to it
    if (index === activeStepIndex + 1 && steps[activeStepIndex]?.isOptional) {
      return StepStatusValues.Next;
    }
```

amanabiy: The ability to navigate to the next step when it's optional (but the current one is not) was already there before this PR, so I didn't want to change it as it might cause unexpected issues if anyone is relying on it. The `steps[index]?.isOptional` check was the existing condition for adjacent steps; I only added `steps[activeStepIndex]?.isOptional`. 

Logically, it makes sense not to allow skipping to the next step if the current step is required. I'm just concerned it could cause a behavior difference since it's been the existing behavior.

jperals (approved):

PR #4583, opened 2026-06-04
feat: Add skeleton loading support to table

gethinwebster:
Add skeleton prop ({ totalRows: number }) that renders skeleton placeholder rows to fill the table while data is loading.

- skeleton.totalRows defines total visible rows (typically page size)
- Skeleton rows rendered = max(0, totalRows - items.length)
- Standard loading: items=[] + loading=true + skeleton -> all skeleton rows
- Progressive loading: items=[partial] + skeleton -> real rows + trailing skeleton
- A11y: visually-hidden row with ScreenreaderOnly loadingText before skeleton rows
- Skeleton rows use aria-hidden=true
- Includes interactive demo pages for both patterns

### Description

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/table/internal.tsx
+++ b/src/table/internal.tsx
@@ -50,6 +50,7 @@ import { ResizeTracker } from './resizer';
 import { focusMarkers, useSelection, useSelectionFocusMove } from './selection';
 import { TableBodySelectionCell } from './selection/selection-cell';
 import { useGroupSelection } from './selection/use-group-selection';
+import { SkeletonRows } from './skeleton-rows';
 import { useStickyColumns } from './sticky-columns';
 import StickyHeader, { StickyHeaderRef } from './sticky-header';
 import { StickyScrollbar } from './sticky-scrollbar';
@@ -113,6 +114,7 @@ const InternalTable = React.forwardRef(
       trackBy,
       loading,
       loadingText,
+      skeleton,
       selectionType: externalSelectionType,
       selectedItems,
       isItemDisabled,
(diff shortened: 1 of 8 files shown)


src/table/skeleton-row.tsx
```diff
@@ -0,0 +1,94 @@
...
+  stickyState: any;
+  tableRole: any;
+  ariaLabels: any;
+  cellVerticalAlign: any;
+  computedVariant: any;
+  visibleColumnDefinitions: readonly any[];
+  wrapLines: any;
+  resizableColumns: any;
+  colIndexOffset: number;
+}
```
NathanZlion: There's types for tableRole, stickyState and wrapLines and resizableColumns here can be a booleans. I think we can type these instead of using any.

src/table/internal.tsx
```diff
@@ -603,7 +606,33 @@ const InternalTable = React.forwardRef(
...
-                      {loading || allItems.length === 0 ? (
+                      {skeleton && allItems.length === 0 && loading ? (
+                        <>
+                          <tr aria-hidden="false">
+                            <td colSpan={totalColumnsCount} className={styles['skeleton-loading-cell']}>
+                              <ScreenreaderOnly>{loadingText}</ScreenreaderOnly>
+                            </td>
+                          </tr>
+                          {Array.from({ length: skeleton.totalRows }, (_, rowIndex) => (
+                            <SkeletonRow
```
NathanZlion: This block and the one below look similar and have a lot in common, can we extract them to avoid repetetion.

src/table/internal.tsx
```diff
@@ -797,6 +829,33 @@ const InternalTable = React.forwardRef(
...
+                        <>
+                          <tr aria-hidden="false">
+                            <td colSpan={totalColumnsCount} className={styles['skeleton-loading-cell']}>
+                              <ScreenreaderOnly>{loadingText}</ScreenreaderOnly>
+                            </td>
+                          </tr>
+                          {Array.from({ length: Math.max(0, skeleton.totalRows - allItems.length) }, (_, rowIndex) => (
+                            <SkeletonRow
+                              key={`skeleton-row-${rowIndex}`}
+                              rowIndex={rowIndex + 1}
```
NathanZlion: I don't understand 1 here to the rowIndex?

src/table/skeleton-row.tsx
```diff
@@ -0,0 +1,94 @@
...
+  tableRole,
+  ariaLabels,
+  cellVerticalAlign,
+  computedVariant,
+  visibleColumnDefinitions,
+  wrapLines,
+  resizableColumns,
+  colIndexOffset,
+}: SkeletonRowProps) {
+  const isFirstRow = rowIndex === 0;
```
NathanZlion: Following up: https://github.com/cloudscape-design/components/pull/4583/changes#r3379168550
If we do the +1 to row index then I think this would not be true

src/table/internal.tsx
gethinwebster: extracted

src/table/internal.tsx
gethinwebster: updated the naming so it's clearer what's going on

src/table/skeleton-row.tsx
gethinwebster: it can be true in the other code path (when _only_ skeleton rows are displayed), but have updated naming so hopefully the logic is clearer now

NathanZlion (approved):

PR #4585, opened 2026-06-05
feat: Make color-background-badge-icon token public and themeable

avinashbot:
### Description

For core theme — the top navigation is going to be white, and the red dot doesn't meet contrast. So for the new theme to be applied, it needs to be themeable.

Related links, issue #, if available: D460257194

### How has this been tested?

Updated snapshots

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/__tests__/snapshot-tests/__snapshots__/design-tokens.test.ts.snap
+++ b/src/__tests__/snapshot-tests/__snapshots__/design-tokens.test.ts.snap
@@ -232,7 +232,7 @@ exports[`Design tokens artifacts Design tokens JSON for classic matches the snap
           },
         },
         "color-background-badge-icon": {
-          "$description": "The background color of the badge indicator on icons and icon buttons.",
+          "$description": "The background color of the badge indicator on navigation and panel buttons.",
           "$value": {
             "dark": "#ff5d64",
             "light": "#d13212",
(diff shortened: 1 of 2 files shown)


src/__tests__/snapshot-tests/__snapshots__/design-tokens.test.ts.snap
```diff
@@ -6999,6 +7013,13 @@ exports[`Design tokens artifacts Design tokens JSON for classic matches the snap
             "light": "radial-gradient(circle farthest-corner at top right, #b8e7ff 0%, #0099ff 25%, #5c7fff 40% , #8575ff 60%, #962eff 80%)",
           },
         },
+        "color-background-badge-icon": {
+          "$description": "The background color of the badge indicator on icons and icon buttons.",
```
gethinwebster: I'm wondering about the description.

1) do we actually have this directly on icons anywhere?
2) it _kinda_ suggests that the icon and icon button components support a badge property, but I don't think they do: it's internal-only?

Maybe "The background color of the badge indicator on navigation and panel buttons."?

avinashbot: It's implemented in icon, but you're right, publicly it's only used in the context of buttons. But since it's implemented in icon, I wanted to play it safe. There's no backwards compat issue with the descriptions, so for now your suggestion is better. Will update.

gethinwebster (approved):

PR #4593, opened 2026-06-09
feat: Side navigation collapsed state and icon properties

jkuelz:
### Description

This commit contains changes to enable a collapsed (icon-only) variant of Side Navigation.
 
Added two new properties to Side Navigation - `icon` on item types and `collapsed` on the parent.
This change also refactors the css of side navigation to be able to support alternative visual styles (e.g. background color to highlight the active state). The previous version relied on margins for all spacing, but an active background requires padding around text. Hence, all spacing needed to be adjusted in order to keep it as backwards compatible (visually) as possible. These spacing values have been changed in a way to make the density easily themeable with just a couple tokens (by changing the height/size of the item or the gap between them).

There are a few very minor expected visual updates that have been tested and verified in my dev-pipeline. All are within 4px diff. The only change worth noting here is the following:

Increased space between text and info slot because of change in wrapping. The info slot can't be part of the <a> because we recommend using the info slot for feature announcement popovers, which would result in nested click targets. To simplify, the info slot will appear the same as it did, except if the text wraps to multiple lines. In this case, the info slot will be pulled to the right, hence the addition of a little more space to help differentiate. 
Screenshot 2026-06-18 at 12 02 03 AM

Screenshot 2026-06-17 at 11 16 27 PM

Related links, issue #, if available: aA2oAPaWUrMb

### How has this been tested?

Pipeline: dev-v3-jkuelz

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/test-utils/dom/side-navigation/index.ts
```diff
@@ -27,7 +27,7 @@ export default class SideNavigationWrapper extends ComponentWrapper {
   }
 
   findActiveLink(): ElementWrapper<HTMLAnchorElement> | null {
-    return this.findByClassName(styles['link-active']);
+    return this.findByClassName(styles['link--active']);
```
georgylobko: This change is not backward compatible and will introduce a regression. Is this intentional?

src/side-navigation/__tests__/collapsed-side-navigation.test.tsx
```diff
@@ -0,0 +1,431 @@
...
+          items={[
+            {
+              type: 'section-group',
+              title: 'My Group',
+              items: [iconLink('Item', '#/item')],
+            },
+          ]}
+        />
+      );
+      const innerList = container.querySelector('ul[aria-label="My Group"]');
```
georgylobko: How is this test different from the `applies aria-label from section text to the inner list'`?

src/side-navigation/__tests__/collapsed-side-navigation.test.tsx
```diff
@@ -0,0 +1,431 @@
...
+            {
+              type: 'section',
+              text: 'Resources',
+              icon: <Icon name="settings" />,
+              items: [plainLink('Child', '#/child')],
+            },
+          ]}
+        />
+      );
+      expect(container.querySelector('[class*="section-header-text"]')).toBeTruthy();
```
georgylobko: Should we check for the icon additionally?

src/side-navigation/__tests__/collapsed-side-navigation.test.tsx
```diff
@@ -0,0 +1,431 @@
...
+
+  describe('tooltip on collapsed items', () => {
+    it('shows tooltip on focus of a collapsed link', () => {
+      const { container } = render(<SideNavigation collapsed={true} items={[iconLink('Dashboard', '#/dashboard')]} />);
+      const link = container.querySelector('a')!;
+      fireEvent.focus(link);
+      expect(document.body.querySelector('[role="tooltip"]')).toBeTruthy();
+    });
+
+    it('hides tooltip on blur of a collapsed link', () => {
```
georgylobko: I'd suggest merging this with the previous test such as:

1. Check the tooltip is not shown initially
2. Focus the link
3. Check the tooltip is shown
4. Blur
5. Check the tooltip is hidden

src/side-navigation/__tests__/collapsed-side-navigation.test.tsx
```diff
@@ -0,0 +1,431 @@
...
+      const { container } = render(<SideNavigation collapsed={true} items={[iconLink('Dashboard', '#/dashboard')]} />);
+      const link = container.querySelector('a')!;
+      fireEvent.focus(link);
+      fireEvent.blur(link);
+      expect(document.body.querySelector('[role="tooltip"]')).toBeNull();
```
georgylobko: General comment: for consistency, let's use container.querySelector instead of document.body.querySelector

src/side-navigation/__tests__/collapsed-side-navigation.test.tsx
```diff
@@ -0,0 +1,431 @@
...
+                  icon: <Icon name="settings" />,
+                  items: [iconLink('Child', '#/child')],
+                },
+              ],
+            },
+          ]}
+        />
+      );
+      const icons = container.querySelectorAll('[class*="item-icon"]');
+      expect(icons.length).toBeGreaterThan(0);
```
georgylobko: Can we check the actual value here?

src/side-navigation/__tests__/collapsed-side-navigation.test.tsx
```diff
@@ -0,0 +1,431 @@
...
+                  text: 'Nested',
+                  icon: <Icon name="settings" />,
+                  items: [iconLink('Child', '#/child')],
+                },
+              ],
+            },
+          ]}
+        />
+      );
+      const icons = container.querySelectorAll('[class*="item-icon"]');
```
georgylobko: I'd suggest to import the actual test class instead of this selector such as `testStyles['icon-icon']`

src/side-navigation/parts.tsx
```diff
@@ -219,15 +337,37 @@ export function NavigationItemsList({
...
-      {lists.map((list, index) => {
+      {filteredLists.map((list, index) => {
         if (!list.items || list.items.length === 0) {
           return (
             <div
               key={`hr-${index}`}
               className={clsx(styles.list, styles[`list-variant-${variant}`], {
                 [styles['list-variant-root--first']]: list.listVariant === 'root' && index === 0,
+                [styles['list-variant-root--collapsed']]: list.listVariant === 'root' && collapsed,
+                [styles['list-variant-root--symmetric']]: list.listVariant === 'root' && collapsed,
```
georgylobko: the prev class is shown on the same condition. do we need to keep both?

src/side-navigation/parts.tsx
```diff
@@ -239,6 +379,8 @@ export function NavigationItemsList({
               key={`list-${index}`}
               className={clsx(styles.list, styles[`list-variant-${list.listVariant}`], {
                 [styles['list-variant-root--first']]: list.listVariant === 'root' && index === 0,
+                [styles['list-variant-root--collapsed']]: list.listVariant === 'root' && collapsed,
+                [styles['list-variant-root--symmetric']]: list.listVariant === 'root' && collapsed,
```
georgylobko: same here

src/side-navigation/parts.tsx
```diff
@@ -338,17 +547,43 @@ function Section({ definition, activeHref, fireFollow, fireChange, variant, posi
...
+
   return (
     <InternalExpandableSection
       variant="footer"
       expanded={expanded}
       onChange={onExpandedChange}
       className={clsx(
         styles.section,
-        variant === 'section-group' && styles['section--no-ident'],
+        isNestedInSectionGroup && styles['section--no-ident'],
```
georgylobko: [nit] why this variable has `nested` in its name?

src/test-utils/dom/side-navigation/index.ts
jkuelz: How is this not backwards compatible if the class name changed along with it?

src/side-navigation/__tests__/collapsed-side-navigation.test.tsx
jkuelz: one tests section-group the other tests section. Combined into one test since they test the same thing.

src/side-navigation/__tests__/collapsed-side-navigation.test.tsx
jkuelz: good call, updated

src/side-navigation/parts.tsx
jkuelz: yes, removed. originally they served two different purposes, but I since removed the reason the symmetric class is needed.

src/side-navigation/parts.tsx
jkuelz: Because this variant is referring to its parent, which I found confusing at first. So added a variable to make it clearer. Also because the classname is a little misleading, since this class actually adds indent, it removes the negative margin that would move it outward.

src/test-utils/dom/side-navigation/index.ts
georgylobko: Responded in DM

georgylobko (approved):

PR #4594, opened 2026-06-09
feat: add children slot for custom content to top navigation

amanabiy:
### Description
Adds a children slot to TopNavigation that lets consumers render fully custom navigation content, replacing the structured layout (identity, search, utilities).
```typescript []
<TopNavigation>
  <MyCustomNav />
</TopNavigation>
```
When children is provided, the component renders the custom content inside the navigation bar and skips the structured layout entirely. It still respects the visualContext prop, so custom content adapts to the dark/branded context or uses page colors with visualContext="none".

**Note**: Depends on the [visualContext PR](https://github.com/cloudscape-design/components/pull/4631)

### How has this been tested?

- Unit tests for custom content rendering, structured fallback, and visualContext across both modes
- Snapshot updates for the new prop

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/top-navigation/optional-props.permutations.page.tsx
+++ b/pages/top-navigation/optional-props.permutations.page.tsx
@@ -1,0 +1,93 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import React from 'react';
+
+import Input from '~components/input';
+import TopNavigation, { TopNavigationProps } from '~components/top-navigation';
+
+import { SimplePage } from '../app/templates';
+import { I18N_STRINGS } from './common';
+import logo from './logos/simple-logo.svg';
+
+const utilities: TopNavigationProps['utilities'] = [
+  { type: 'button', iconName: 'notification', ariaLabel: 'Notifications', badge: true },
+  { type: 'button', text: 'Settings', iconName: 'settings' },
+  {
+    type: 'menu-dropdown',
+    text: 'Jane Doe',
+    description: 'jane.doe@example.com',
+    iconName: 'user-profile',
+    items: [
+      { id: 'profile', text: 'Profile' },
+      { id: 'signout', text: 'Sign out' },
+    ],
+  },
+];
+
+const identity: TopNavigationProps['identity'] = {
+  href: '#',
+  title: 'My Service',
+  logo: { src: logo, alt: 'Logo' },
+};
+
+const search = <Input type="search" placeholder="Search..." value="" onChange={() => {}} ariaLabel="Search" />;
+
+interface Permutation {
+  label: string;
+  props: TopNavigationProps;
+}
+
+const permutations: Permutation[] = [
+  {
+    label: 'Identity + search + utilities',
+    props: { identity, search, utilities, i18nStrings: I18N_STRINGS },
+  },
+  {
+    label: 'Identity + utilities (no search)',
+    props: { identity, utilities, i18nStrings: I18N_STRINGS },
+  },
+  {
+    label: 'Identity + search (no utilities)',
+    props: { identity, search, i18nStrings: I18N_STRINGS },
+  },
+  {
+    label: 'Identity only',
+    props: { identity, i18nStrings: I18N_STRINGS },
+  },
+  {
+    label: 'Utilities only (no identity)',
+    props: { utilities, i18nStrings: I18N_STRINGS },
+  },
+  {
+    label: 'Search + utilities (no identity)',
+    props: { search, utilities, i18nStrings: I18N_STRINGS },
+  },
+  {
+    label: 'Search only (no identity, no utilities)',
+    props: { search, i18nStrings: I18N_STRINGS },
+  },
+  {
+    label: 'Empty (no identity, no search, no utilities)',
+    props: { i18nStrings: I18N_STRINGS },
+  },
+];
+
+const visualContexts: Array<TopNavigationProps['visualContext']> = ['top-navigation', 'none'];
+
+export default function OptionalPropsPermutations() {
+  return (
+    <SimplePage title="TopNavigation optional props permutations" screenshotArea={{}}>
+      {visualContexts.map(visualContext => (
+        <div key={visualContext} style={{ marginBottom: 32 }}>
+          <h2>visualContext=&quot;{visualContext}&quot;</h2>
+          {permutations.map(({ label, props }) => (
+            <div key={label} style={{ marginBottom: 16 }}>
+              <h3>{label}</h3>
+              <TopNavigation {...props} visualContext={visualContext} />
+            </div>
+          ))}
+        </div>
+      ))}
+    </SimplePage>
+  );
+}
--- a/pages/top-navigation/optional-props.permutations.page.tsx
+++ b/pages/top-navigation/optional-props.permutations.page.tsx
@@ -6,6 +6,8 @@ import Input from '~components/input';
 import TopNavigation, { TopNavigationProps } from '~components/top-navigation';
 
 import { SimplePage } from '../app/templates';
+import createPermutations from '../utils/permutations';
+import PermutationsView from '../utils/permutations-view';
 import { I18N_STRINGS } from './common';
 import logo from './logos/simple-logo.svg';
 
@@ -32,62 +34,38 @@ const identity: TopNavigationProps['identity'] = {
 
 const search = <Input type="search" placeholder="Search..." value="" onChange={() => {}} ariaLabel="Search" />;
 
-interface Permutation {
-  label: string;
-  props: TopNavigationProps;
-}
+const contentPermutations: Record<string, TopNavigationProps> = {
+  'Identity + search + utilities': { identity, search, utilities, i18nStrings: I18N_STRINGS },
+  'Identity + utilities (no search)': { identity, utilities, i18nStrings: I18N_STRINGS },
+  'Identity + search (no utilities)': { identity, search, i18nStrings: I18N_STRINGS },
+  'Identity only': { identity, i18nStrings: I18N_STRINGS },
+  'Utilities only (no identity)': { utilities, i18nStrings: I18N_STRINGS },
+  'Search + utilities (no identity)': { search, utilities, i18nStrings: I18N_STRINGS },
+  'Search only (no identity, no utilities)': { search, i18nStrings: I18N_STRINGS },
+  'Empty (no identity, no search, no utilities)': { i18nStrings: I18N_STRINGS },
+};
 
-const permutations: Permutation[] = [
+const permutations = createPermutations<{ visualContext: TopNavigationProps['visualContext']; label: string }>([
   {
-    label: 'Identity + search + utilities',
-    props: { identity, search, utilities, i18nStrings: I18N_STRINGS },
+    visualContext: ['top-navigation', 'none'],
+    label: Object.keys(contentPermutations),
   },
-  {
-    label: 'Identity + utilities (no search)',
-    props: { identity, utilities, i18nStrings: I18N_STRINGS },
-  },
-  {
-    label: 'Identity + search (no utilities)',
-    props: { identity, search, i18nStrings: I18N_STRINGS },
-  },
-  {
-    label: 'Identity only',
-    props: { identity, i18nStrings: I18N_STRINGS },
-  },
-  {
-    label: 'Utilities only (no identity)',
-    props: { utilities, i18nStrings: I18N_STRINGS },
-  },
-  {
-    label: 'Search + utilities (no identity)',
-    props: { search, utilities, i18nStrings: I18N_STRINGS },
-  },
-  {
-    label: 'Search only (no identity, no utilities)',
-    props: { search, i18nStrings: I18N_STRINGS },
-  },
-  {
-    label: 'Empty (no identity, no search, no utilities)',
-    props: { i18nStrings: I18N_STRINGS },
-  },
-];
-
-const visualContexts: Array<TopNavigationProps['visualContext']> = ['top-navigation', 'none'];
+]);
 
 export default function OptionalPropsPermutations() {
   return (
     <SimplePage title="TopNavigation optional props permutations" screenshotArea={{}}>
-      {visualContexts.map(visualContext => (
-        <div key={visualContext} style={{ marginBottom: 32 }}>
-          <h2>visualContext=&quot;{visualContext}&quot;</h2>
-          {permutations.map(({ label, props }) => (
-            <div key={label} style={{ marginBottom: 16 }}>
-              <h3>{label}</h3>
-              <TopNavigation {...props} visualContext={visualContext} />
-            </div>
-          ))}
-        </div>
-      ))}
+      <PermutationsView
+        permutations={permutations}
+        render={({ visualContext, label }) => (
+          <div style={{ marginBottom: 16 }}>
+            <h2>
+              visualContext=&quot;{visualContext}&quot; — {label}
+            </h2>
+            <TopNavigation {...contentPermutations[label]} visualContext={visualContext} />
+          </div>
+        )}
+      />
     </SimplePage>
   );
 }
--- a/src/test-utils/dom/top-navigation/index.ts
+++ b/src/test-utils/dom/top-navigation/index.ts
@@ -9,12 +9,17 @@ import LinkWrapper from '../link';
 import buttonDropdownStyles from '../../../button-dropdown/styles.selectors.js';
 import menuDropdownStyles from '../../../internal/components/menu-dropdown/styles.selectors.js';
 import styles from '../../../top-navigation/styles.selectors.js';
+import testUtilStyles from '../../../top-navigation/test-classes/styles.selectors.js';
 
 export default class TopNavigationWrapper extends ComponentWrapper {
   static rootSelector = `${styles['top-navigation']}:not(.${styles.hidden})`;
 
-  findIdentityLink(): ElementWrapper {
-    return this.find(`.${styles.identity} a`)!;
+  findContent(): ElementWrapper | null {
+    return this.find(`.${testUtilStyles['custom-content']}`);
+  }
+
+  findIdentityLink(): ElementWrapper | null {
+    return this.find(`.${styles.identity} a`);
   }
 
   findLogo(): ElementWrapper | null {
--- a/src/top-navigation/__integ__/top-navigation.test.ts
+++ b/src/top-navigation/__integ__/top-navigation.test.ts
@@ -14,9 +14,9 @@ class TopNavigationPage extends BasePageObject {
   }
 }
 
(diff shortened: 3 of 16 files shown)


src/top-navigation/styles.scss
```diff
@@ -42,6 +42,10 @@
   inline-size: 9000px;
 }
 
+.custom-content {
```
pan-kot: For new components and features we use `/src/{component-name}/test-classes/styles.scss` for test-utils selectors.

src/top-navigation/internal.tsx
```diff
@@ -222,30 +245,38 @@ export default function InternalTopNavigation({
...
-                (utility, i) =>
-                  (!responsiveState.hideUtilities || responsiveState.hideUtilities.indexOf(i) !== -1) &&
-                  !utility.disableUtilityCollapse
-              )}
-              onClose={toggleOverflowMenu}
-            />
-          </div>
-        )}
-      </VisualContext>
+      {visualContext === 'top-navigation' ? (
```
pan-kot: nit: should we instead do:
```
visualContext === 'none' ? content :  <VisualContext contextName={visualContext}>{structuredContent}</VisualContext>
```
?

This makes the relation between props clearer and automatically allows for extension with more visual contexts.

src/top-navigation/internal.tsx
```diff
@@ -88,9 +111,9 @@ export default function InternalTopNavigation({
       >
         <div className={styles['padding-box']}>
           {showIdentity && (
```
pan-kot: nit: I would change this to `showIndentity && identity` to avoid null checks in the code below

src/top-navigation/internal.tsx
```diff
@@ -63,12 +58,40 @@ export default function InternalTopNavigation({
...
+  // Custom content mode: render a simple container with no structured layout
+  if (children !== undefined) {
+    const header = (
+      <header className={styles['top-navigation']}>
+        <div className={styles['custom-content']}>{children}</div>
+      </header>
+    );
+    return (
+      <div {...baseProps} ref={__internalRootRef}>
+        {visualContext === 'top-navigation' ? (
```
pan-kot: This block is repeated. Can we do:

```
let content = null;
if (children !== undefined) {
  content = ...
} else {
  content = ...
}
return <VisualContext>{content}</VisualContext>
```
?

src/top-navigation/internal.tsx
```diff
@@ -27,12 +27,17 @@ export default function InternalTopNavigation({
...
 }: InternalTopNavigationProps) {
-  checkSafeUrl('TopNavigation', identity.href);
+  if (identity) {
+    checkSafeUrl('TopNavigation', identity.href);
+  }
   const baseProps = getBaseProps(restProps);
+
   const { mainRef, virtualRef, breakpoint, responsiveState, isSearchExpanded, onSearchUtilityClick } = useTopNavigation(
-    { identity, search, utilities }
+    { identity: identity ?? { href: '' }, search, utilities }
```
pan-kot: Let's not pass default that does not make any practical sense. Instead, let's ensure that useTopNavigation can handle the missing identity.

Alternatively, we can avoid calling this hook at all:

```
if (children !== null) {
  return <CustomTopNavigation {...props} />
} else {
  return <StructuredTopNavigation {...props} />
}
```

src/top-navigation/interfaces.ts
```diff
@@ -15,7 +15,21 @@ export interface TopNavigationProps extends BaseComponentProps {
...
+
+  /**
+   * Specifies custom navigation content.
+   * When provided, replaces all structured content (identity, search, utilities are ignored).
+   */
+  children?: React.ReactNode;
+
+  /**
+   * Controls the color scheme of the navigation bar and its contents.
+   * - "top-navigation": Applies the top-navigation visual context. The component and its contents use dark, branded colors in both light and dark mode.
```
pan-kot: what is "branded colors"?

src/top-navigation/interfaces.ts
```diff
@@ -15,7 +15,21 @@ export interface TopNavigationProps extends BaseComponentProps {
    * * `href` (string) - Specifies the `href` that the header links to.
    * * `onFollow` (() => void) - Specifies the event handler called when the identity is clicked without any modifier keys.
    */
-  identity: TopNavigationProps.Identity;
+  identity?: TopNavigationProps.Identity;
```
pan-kot, line 18: What happens if a top nav is rendered without custom content and also without identity? Will the component maintain its layout and responsiveness or appear broken?

src/top-navigation/__tests__/top-navigation-custom-content.test.tsx
```diff
@@ -0,0 +1,93 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import React from 'react';
+import { render } from '@testing-library/react';
+
+import createWrapper from '../../../lib/components/test-utils/dom';
+import TopNavigation, { TopNavigationProps } from '../../../lib/components/top-navigation';
+
+const I18N_STRINGS: TopNavigationProps.I18nStrings = {
```
pan-kot: Why are these strings needed in the custom content tests?

src/top-navigation/__tests__/top-navigation-custom-content.test.tsx
```diff
@@ -0,0 +1,93 @@
...
+  searchIconAriaLabel: 'Search',
+  searchDismissIconAriaLabel: 'Close search',
+  overflowMenuTriggerText: 'More',
+  overflowMenuTitleText: 'All',
+  overflowMenuBackIconAriaLabel: 'Back',
+  overflowMenuDismissIconAriaLabel: 'Close',
+};
+
+const renderTopNavigation = (props: TopNavigationProps, children?: React.ReactNode) => {
```
pan-kot: nit: children is already a part of TopNavigationProps - why do we define it separately?

src/top-navigation/__tests__/top-navigation-custom-content.test.tsx
```diff
@@ -0,0 +1,93 @@
...
+};
+
+describe('children', () => {
+  test('renders custom content when children are provided', () => {
+    const wrapper = renderTopNavigation({}, <div data-testid="custom">Custom Nav</div>);
+    expect(wrapper.findContent()).not.toBeNull();
+    expect(wrapper.findContent()!.getElement()).toHaveTextContent('Custom Nav');
+  });
+
+  test('does not render identity when children are provided', () => {
```
pan-kot: nit: should we add these assertions to the 'renders custom content when children are provided' test instead:
```
expect(wrapper.findTitle()).toBeNull();
expect(wrapper.findIdentityLink()).toBeNull();
expect(wrapper.findSearch()).toBeNull();
expect(wrapper.findUtilities()).toHaveLength(0);
```
?

As custom content replaces everything else, I think it makes sense to check for everything in one place rather than creating separate tests for that.

src/top-navigation/__tests__/top-navigation-custom-content.test.tsx
```diff
@@ -0,0 +1,93 @@
...
+  });
+
+  test('renders structured mode when children are not provided', () => {
+    const wrapper = renderTopNavigation({ identity: { href: '#', title: 'Structured' } });
+    expect(wrapper.findContent()).toBeNull();
+    expect(wrapper.findTitle()!.getElement()).toHaveTextContent('Structured');
+  });
+});
+
+describe('visualContext', () => {
```
pan-kot: Why do visual context related tests live inside top-navigation-custom-content.test?

src/top-navigation/__tests__/top-navigation-custom-content.test.tsx
```diff
@@ -0,0 +1,93 @@
...
+  test('defaults to top-navigation (dark visual context)', () => {
+    const { container } = render(<TopNavigation identity={{ href: '#', title: 'Title' }} i18nStrings={I18N_STRINGS} />);
+    expect(container.querySelector('[class*="awsui-context-top-navigation"]')).not.toBeNull();
+  });
+
+  test('applies visual context when visualContext is "top-navigation"', () => {
```
pan-kot: This test also checks custom content, but that is not explicitly called out. I recommend making test names better uniform to ensure we capture all 4 states explicitly. Alternatively, we can group tests together:

```
test('defaults to top-navigation visual context', () => {
  render(
    <>
      <TopNavigation identity={{ href: '#', title: 'Title' }} />
      <TopNavigation>custom</TopNavigation>
    </>
  );
  expect(createWrapper().findAll('[class*="awsui-context-top-navigation"]')).toHaveLength(2);
});

test('uses top-navigation visual context explicitly', () => {
  // ...
});

test('uses no visual context', () => {
  // ...
});
```

src/top-navigation/__integ__/top-navigation.test.ts
```diff
@@ -233,3 +233,21 @@ describe('Top navigation', () => {
...
+  const setupCustomContentTest = (testFn: (page: TopNavigationPage) => Promise<void>) => {
+    return useBrowser(async browser => {
+      await browser.url('#/light/top-navigation/custom-content');
+      const page = new TopNavigationPage(browser);
+      await page.waitForVisible(wrapper.toSelector());
+      await testFn(page);
+    });
+  };
+
+  test(
```
pan-kot: why is this test necessary?

src/test-utils/dom/top-navigation/index.ts
```diff
@@ -13,8 +13,12 @@ import styles from '../../../top-navigation/styles.selectors.js';
 export default class TopNavigationWrapper extends ComponentWrapper {
   static rootSelector = `${styles['top-navigation']}:not(.${styles.hidden})`;
 
-  findIdentityLink(): ElementWrapper {
-    return this.find(`.${styles.identity} a`)!;
+  findContent(): ElementWrapper | null {
+    return this.find(`.${styles['custom-content']}`);
+  }
+
+  findIdentityLink(): ElementWrapper | null {
```
pan-kot, line 21: Should we make this change separately?

As it can be breaking - we can do this change after releasing the feature. The chance of hitting the issue is very low - it is only possible in cases where the new custom content property is used and the test checks for identity for some reason (there is little reason to do that in consumer tests).

pages/top-navigation/custom-content.page.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+      <h2>Custom content (visualContext=&quot;none&quot;)</h2>
+      <TopNavigation visualContext="none">
+        <CustomNav searchValue={searchValue} onSearchChange={setSearchValue} />
+      </TopNavigation>
+
+      <br />
+
+      <h2>Structured mode (unchanged)</h2>
+      <TopNavigation
```
pan-kot: why do we render this variant in the page called custom-content?

pages/top-navigation/custom-content.page.tsx
```diff
@@ -0,0 +1,68 @@
...
+    <SimplePage title="TopNavigation children" screenshotArea={{}}>
+      <h2>Custom content (default visual context)</h2>
+      <TopNavigation>
+        <CustomNav searchValue={searchValue} onSearchChange={setSearchValue} />
+      </TopNavigation>
+
+      <br />
+
+      <h2>Custom content (visualContext=&quot;none&quot;)</h2>
+      <TopNavigation visualContext="none">
```
pan-kot: I would also add one or two examples with visualContext="none" to some existing top-nav pages that use structured content

pages/top-navigation/custom-content.page.tsx
```diff
@@ -0,0 +1,68 @@
...
+import React, { useState } from 'react';
+
+import Input from '~components/input';
+import TopNavigation from '~components/top-navigation';
+
+import { SimplePage } from '../app/templates';
+import { I18N_STRINGS } from './common';
+import logo from './logos/simple-logo.svg';
+
+function CustomNav({ searchValue, onSearchChange }: { searchValue: string; onSearchChange: (value: string) => void }) {
```
pan-kot: nit: this custom nav is not that custom. I would extend this example or even create a few more examples to feature more Cloudscape components that can be used inside the top-nav, such as button group, button dropdown, links, and more.

src/top-navigation/__integ__/top-navigation.test.ts
amanabiy: I Agreed, I removed it. The unit test already covers this, and this integ test wasn't verifying any browser-specific behavior.

src/test-utils/dom/top-navigation/index.ts
amanabiy: Good point, I considered separating it too. I did a dry run and all live builds pass, but I agree it's safer to keep it in a separate follow-up to avoid blocking this PR. I will share the internal dry run with you.

pages/top-navigation/custom-content.page.tsx
amanabiy: I was using it for visual comparison during development. Removed.

src/top-navigation/internal.tsx
amanabiy: Done, I have created a wrapper function `wrapWithVisualContext` to reuse it.

src/top-navigation/internal.tsx
amanabiy: Done 👍

src/top-navigation/interfaces.ts
amanabiy: Yes, it should be responsive and shouldn't be broken. I have added a `optional-props.permutations` page to also test how it looks like. This should help us to catch any future regressions as well.

pages/top-navigation/optional-props.permutations.page.tsx
amanabiy, line 1: The optional-props.permutations.page.tsx helps to catch visual regression for different permutation of optional props in structured top navigation.

src/top-navigation/use-top-navigation.ts
```diff
@@ -171,6 +171,16 @@ function getContentBoxWidth(element: Element): number {
   );
 }
 
+/**
+ * Measures the width of an optional element. Returns 0 when the element is not rendered
+ * (e.g. there is no `identity`, `title` or `search`). The `number` return type makes the
```
pan-kot: nit: I find the zero fallback part not worth detailed explanation - this is a common approach that we use.

src/top-navigation/internal.tsx
```diff
@@ -18,19 +21,84 @@ import Utility from './parts/utility';
...
+
+function CustomContentTopNavigation({
+  __internalRootRef,
+  children,
+  visualContext,
+  ...restProps
+}: CustomContentTopNavigationProps) {
+  const baseProps = getBaseProps(restProps);
+
+  if (isDevelopment) {
```
pan-kot: this explicit isDevelopment check is not needed: the warnOnce util has a check for it inside

src/top-navigation/internal.tsx
```diff
@@ -18,19 +21,84 @@ import Utility from './parts/utility';
...
+    const { identity, search, utilities } = restProps;
+    if (identity || search || (utilities && utilities.length > 0)) {
+      warnOnce(
+        'TopNavigation',
+        'You have provided `children` along with structured props (`identity`, `search`, and/or `utilities`). ' +
```
pan-kot: nit: I think it is enough to say: "When `children` is set, the structured props (`identity`, `search`, and `utilities`) are ignored" - wdyt?

src/top-navigation/internal.tsx
```diff
@@ -18,19 +21,84 @@ import Utility from './parts/utility';
...
+      <CustomContentTopNavigation __internalRootRef={__internalRootRef} visualContext={visualContext} {...restProps}>
+        {children}
+      </CustomContentTopNavigation>
+    );
+  }
+
+  return <StructuredTopNavigation __internalRootRef={__internalRootRef} visualContext={visualContext} {...restProps} />;
+}
+
+interface CustomContentTopNavigationProps extends InternalTopNavigationProps {
```
pan-kot: Why do we extent InternalTopNavigationProps with visualContext for custom and structured? This prop should be already included as part of the updated public api

pan-kot: If we want to ensure the internal prop is required - check usages of the utility `SomeRequired` type - that is our go-to approach to use for internal props that receive default values in index.tsx

src/top-navigation/__tests__/top-navigation-responsiveness.test.tsx
```diff
@@ -189,3 +190,20 @@ describe('TopNavigation responsiveness', () => {
     expect(responsiveState).toEqual({});
   });
 });
+
+describe('getOptionalElementWidth', () => {
```
pan-kot, line 194: is this really necessary to have a test for this simple util, can we ensure it is covered as part of the component-level tests instead?

src/top-navigation/__tests__/common.tsx
```diff
@@ -0,0 +1,23 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import React from 'react';
+import { render } from '@testing-library/react';
+
+import createWrapper from '../../../lib/components/test-utils/dom';
+import TopNavigation, { TopNavigationProps } from '../../../lib/components/top-navigation';
+
+export const I18N_STRINGS: TopNavigationProps.I18nStrings = {
```
pan-kot: Do we import this object somewhere?

src/top-navigation/use-top-navigation.ts
amanabiy: Agreed, Removed the detailed explanation.

src/top-navigation/internal.tsx
amanabiy: Yes, I have removed it.

src/top-navigation/internal.tsx
amanabiy: Yes, that's cleaner. I have changed it.

src/top-navigation/internal.tsx
amanabiy: Thanks, I have used `InternalTopNavigationProps` and for the `StructuredTopNavigationProps` I have omitted the children since we don't need it to make it explicit.

src/top-navigation/__tests__/common.tsx
amanabiy: Fixed

src/top-navigation/__tests__/top-navigation-responsiveness.test.tsx
amanabiy: Agreed, I have removed the simple test for the function, and added integration test that search and utilities behave as expected when identity is not there. The code coverage might complain though since we don't have unit test covering those lines, but they are covered on the integration test.

pan-kot (dismissed):

pan-kot (dismissed):

pan-kot (approved):

PR #4609, opened 2026-06-11
fix: Avoid redundant table column measurements on re-render

TrevorBurnham:
This PR reduces unnecessary DOM measurements when a table with "sticky" features (`stickyColumns` or `stickyHeader`) re-renders.

`ColumnWidthsProvider` runs a `useEffect` that depends on its `visibleColumns` array. That effect calls `updateColumnWidths()`, which synchronizes sticky cells by reading layout from the DOM via `getBoundingClientRect()` (the `sticky` branch of `getColumnStyles`). In `InternalTable`, the arrays passed to the provider — `visibleColumnDefinitions` and the derived `visibleColumnWidthsWithSelection` / `visibleColumnIdsWithSelection` — are rebuilt as fresh arrays of fresh objects on every render, causing the provider's effect to re-fire. These unnecessary calculations contribute to input lag when users type in a sticky table's searchbox.

**Memoize the derivations in `InternalTable`:**

- `visibleColumnDefinitions` — keyed on `columnDefinitions`, `columnDisplay`, and `visibleColumns`.
- `visibleColumnWidthsWithSelection` / `visibleColumnIdsWithSelection` — keyed on `hasSelection` and `visibleColumnDefinitions`.

**Re-sync sticky headers on content-driven resizes:**

Memoizing `visibleColumns` means the provider's effect no longer re-fires on every render, so sticky header copies are no longer re-synced from the primary cells on unrelated re-renders. That's the point for fixed-width and resizable columns, but auto-layout columns (no explicit width) can change width when their content changes without any change to the column definitions (e.g. paginating or filtering). To keep the sticky header copies correct in that case, `ColumnWidthsProvider` now observes the primary cells with a `ResizeObserver` and re-syncs the copies when a primary cell resizes. The observer re-attaches when the set of visible columns changes, and no-ops in environments without `ResizeObserver` (e.g. SSR).

## Benefits

Measured `getBoundingClientRect` calls per re-render with unchanged columns:

| Configuration | Before | After |
| --- | --- | --- |
| `stickyColumns` | 3 | 0 |
| `stickyHeader` | 5 | 3 |

The remaining reads come from `StickyHeader`'s own per-render synchronization, a separate mechanism this PR does not change (a candidate for a follow-up).

Non-sticky tables benefit in a smaller way from reduced array allocations.

### How has this been tested?

Added unit tests in `src/table/__tests__/columns-width.test.tsx`:

- `does not measure column widths when re-rendering with unchanged columns` — re-renders with new `items` but identical `columnDefinitions` and asserts no `getBoundingClientRect` calls occur. Fails on `main`, passes with this change.
- `does measure column widths when the columns change` — a negative control that re-renders with an added column and asserts measurement still occurs.
- `re-syncs the sticky header copy when a primary cell changes width without a column change` — simulates an auto-layout cell growing and asserts the observed resize re-syncs the sticky copy.
- `renders without a ResizeObserver (e.g. server-side rendering)` — asserts the provider renders without throwing when `ResizeObserver` is absent.

---


ernst-dev (approved):

ernst-dev (changes requested):
Looks like there's a valid failing test relating to table measurements, please take a look at the failure

TrevorBurnham:
> Looks like there's a valid failing test relating to table measurements, please take a look at the failure

Fixed!

ernst-dev (changes requested):
Sorry Trevor, I know this is annoying but can we hit the codecov target for these changes?

TrevorBurnham:
> Sorry Trevor, I know this is annoying but can we hit the codecov target for these changes?

Done!

ernst-dev (approved):

PR #4611, opened 2026-06-12
refactor: speed up option selection state in useSelect

TrevorBurnham:
This PR speeds up the selection-state computation that runs on every render in `useSelect`, which backs both `Select` and `Multiselect`.

Previously, `useSelect` recomputed which options are selected using two nested scans that grew quadratically with the size of the options and selection lists:

- **`connectOptionsByValue`** scanned all dropdown options to find a match for *each* selected option — `O(selected × options)`.
- **`getOptionProps`** then ran up to three `__selectedOptions.indexOf()` scans per *rendered* option (the option itself plus its two neighbours, used to compute `selected` / `isNextSelected` / `isPreviousSelected`) — `O(filtered × selected)`. On the default, non-virtualized `PlainList` this scans the entire list, so the cost is not bounded by the visible window.

For a large multiselect, this adds up to hundreds of microseconds of avoidable work on every render.

This change makes both lookups linear:

- `connectOptionsByValue` builds a `Map` of dropdown options keyed by value once, then resolves each selected option in `O(1)` → `O(selected + options)`.
- `useSelect` builds a `Set` of the selected dropdown options once per render and tests membership in `O(1)` per option, replacing the per-option `indexOf` scans.

The `Set` is keyed by the same `DropdownOption` *references* that `connectOptionsByValue` returns, so `Set.has()` is exactly equivalent to the previous reference-based `indexOf() > -1` checks — including the `undefined`-neighbour case at list boundaries.

Benchmarks (Node, isolated microbenchmarks with an equivalence check):

| Function | Inputs | Before | After | Speedup |
| --- | --- | --- | --- | --- |
| `connectOptionsByValue` | 1000 options, 50 selected | 147µs | 13µs | ~11x |
| `connectOptionsByValue` | 2000 options, 200 selected | 993µs | 27µs | ~37x |
| `getOptionProps` (over all options) | 1000 options, 100 selected | 50µs | 12µs | ~4x |

### How has this been tested?

- New unit tests in `src/select/__tests__/connect-options.test.ts` cover the edge cases that must be preserved: empty selection, order preservation, parent options skipped, first-match-wins on duplicate values, and matching options with an `undefined` value.
- The selection-state change in `getOptionProps` is covered by the existing `src/select/__tests__/use-select.test.ts` assertions on `selected` / `isNextSelected` / `isPreviousSelected`, including the group-neighbour case.
- Full `Select`, `Multiselect`, and `Autoselect` unit suites pass (`src/select/`, `src/multiselect/`, `src/autosuggest/` — 552 tests green).

---


src/select/utils/connect-options.ts
```diff
@@ -6,17 +6,23 @@ export const connectOptionsByValue = (
...
-        (dropdownOption.option as OptionDefinition).value === selectedOption.value
-      ) {
-        return dropdownOption;
+  if (!selectedOptions || selectedOptions.length === 0) {
+    return [];
+  }
+  // Index the dropdown options by value so each selected option resolves in O(1), turning the
+  // overall cost from O(selected × options) into O(selected + options). This matters for large
+  // multiselects, which re-run this on every render. The first non-parent option for a given value
+  // wins, preserving the previous linear-scan order.
```
gethinwebster:
```suggestion
```

src/select/utils/use-select.ts
```diff
@@ -76,6 +76,9 @@ export function useSelect({
   const hasFilter = filteringType !== 'none' && !embedded;
   const activeRef = hasFilter ? filterRef : menuRef;
   const __selectedOptions = connectOptionsByValue(options, selectedOptions);
+  // Membership set for the selected dropdown options so getOptionProps can test selection in O(1)
+  // instead of scanning __selectedOptions for every (filtered) option, which is O(filtered × selected).
```
gethinwebster:
```suggestion
```

gethinwebster (commented):
Looks good but please remove the inline comments, they're adding unnecessary explanations.

gethinwebster (approved):

PR #4615, opened 2026-06-12
fix: Fix container-query driven layout oscillation caused by scrollbars

avinashbot:
### Description

Cleanest solution I could find for the issue in the ticket — make breakpoint changes "stick" or "lag" into a boundary between two breakpoints. This prevents any added scrollbars from making the breakpoint calculation unstable. Of course, this now means that adjacent breakpoints values are implicitly dependent on the current breakpoint, but only by 20px, so we're not messing up people's layouts too badly.

I use term "stickiness" in the code; I imagine it like a kind of magnet that doesn't let go until you pass a certain distance in either direction of the numerical boundary values.

```   
-----------<-20px-|------------|-20px->------
           ^      ^            ^
           |      |            |
           |   breakpoint boundaries
           |
  breakpoint actually
    switches here
```

Related links, issue #, if available: AWSUI-62065

### How has this been tested?

Kiro wrote me up some unit tests, but there's a dev page recreating the original ticket scenario.

<details>

#### Correctness

#### Security

#### Testing

</details>

---


pages/container-query-scrollbar.page.tsx
```diff
@@ -0,0 +1,46 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
SpyZzey, line 1: This page needs to follow our *.page.tsx convention to be registered as a page.

src/internal/breakpoints.ts
```diff
@@ -30,15 +30,44 @@ export function matchBreakpointMapping<T>(subset: Partial<Record<Breakpoint, T>>
...
+/**
+ * The width (in px) by which a measured container can switch into a breakpoint boundary without
+ * actually triggering that breakpoint switch. This makes each breakpoint edge "sticky", making
+ * you have to travel a little further into the breakpoint to "lose" the previous one.
+ *
+ * When a JS-resolved breakpoint sits within a scrollbar-width of a boundary, switching the layout
+ * can grow/shrink the page enough to toggle the viewport scrollbar, which in turn changes the
+ * measured width and flips the breakpoint back — an infinite layout loop (see AWSUI-62065).
+ */
+const BREAKPOINT_SWITCH_OFFSET = 20;
```
SpyZzey: While this is probably fine, I am wondering if we could/should calculate the width of the scrollbar. This way, the approach would also work for cases where the scrollbar is bigger than 20px. Note: As of now, all default scrollbar sizes are <= 17px (https://codepen.io/sambible/post/browser-scrollbar-widths) - but could theoretically be changed by browser extensions (or even ourselves through CSS).

SpyZzey (changes requested):

src/internal/breakpoints.ts
avinashbot: I could go either way on that, I was conflicted about this.

We actually already have a utility for it ([browser-scrollbar-size.ts](https://github.com/cloudscape-design/components/blob/main/src/internal/utils/browser-scrollbar-size.ts)) but the only reliable way of doing is to modify the DOM and measure the elements, which is a bit of a worrying performance trap (not something I noticed on my Macbook on a simple dev page, but breakpoint measurements are pretty foundational and used in a lot of places). Plus, having a fixed number makes it more predictable between test and real environments (for example, we don't have any testing for overlay scrollbars, even browser/screenshot tests).

pages/container-query-scrollbar.page.tsx
avinashbot: Phew, well spotted, thanks!

src/internal/breakpoints.ts
SpyZzey: I think it'd be fine having it once when the application loads instead of re-calculating the scrollbar width every time the breakpoints gets checked. 

And regarding testing, this should probably be abstracted away anyway, right? But I am fine with both.

SpyZzey (dismissed):

src/internal/breakpoints.ts
avinashbot: Okay, changed my mind, let's just start with scrollbar measurements, and if any performance issues come up, I'll switch to the static value. This keeps the overlay scrollbar users unaffected.

SpyZzey (approved):

PR #4626, opened 2026-06-15
feat: Add one-theme styling support for button-dropdown split trigger

at-susie:
### Description

This PR adds one-theme-specific styles and icon behavior to the button-dropdown component's split trigger and expandable category elements.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/button-dropdown/styles.scss
```diff
@@ -114,6 +114,18 @@ $dropdown-trigger-icon-offset: 2px;
       margin-inline-start: calc(#{awsui.$border-width-button} * -1);
     }
   }
+
+  & > .trigger-item:not(:first-child).one-theme {
+    & > .trigger-button {
+      padding-inline: calc(#{awsui.$space-xs} - #{$dropdown-trigger-icon-offset});
```
mxschll, line 122: This rule is the same as in line 98. Do we actually need to duplicate this rule?

at-susie: It's actually necessary. So the logic is that the `one-theme` trigger doesn't follow the `visual-refresh` padding-end adjustment. It's not redundant with the base rule, but it's a reset over the `visual-refresh` sub-rule.
But, to make it clear, I added a comment above the line.

mxschll: I see, @ernst-dev consideration for the future, should we remove the visual refresh scoping at some point and make it the default?

mxschll (approved):

src/button-dropdown/styles.scss
at-susie: We can discuss the approach. I will write down how we can tackle that.

PR #4630, opened 2026-06-15
fix: ProgressBar calling a throttle assertion after unmounting

ernst-dev:
### Description

Cancels a call that is throttled when the component has unmounted preventing the call from firing.

Related links, issue #, if available: `AWSUI-62081`

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/progress-bar/index.tsx
```diff
@@ -64,6 +64,9 @@ export default function ProgressBar({
     throttledAssertion(value);
   }, [throttledAssertion, value]);
 
+  // Cancel pending throttled invocation on unmount to avoid setting state on an unmounted component.
+  useEffect(() => () => throttledAssertion.cancel(), [throttledAssertion]);
```
pan-kot: can we add the throttledAssertion.cancel() to the prev useEffect instead?

```
useEffect(() => {
    throttledAssertion(value);
    return () => throttledAssertion.cancel()
}, [throttledAssertion, value]);
```

src/progress-bar/__tests__/progress-bar.test.tsx
```diff
@@ -286,4 +300,22 @@ describe('Progress updates', () => {
...
+    // Drop the leading invocation that runs synchronously on mount.
+    callback.mockClear();
+
+    // Schedule a trailing invocation, then unmount before it fires.
+    rerender(<ProgressBar label="progress" value={42} />);
+    unmount();
+    act(() => jest.advanceTimersByTime(6000));
+
+    expect(callback).not.toHaveBeenCalled();
+    jest.useRealTimers();
```
pan-kot: Should we use before/after for setting fake timers? Otherwise, if there is an exception prior to this line - the real timers might not be re-established - thus potentially affecting other tests.

src/progress-bar/index.tsx
ernst-dev: Yes

src/progress-bar/__tests__/progress-bar.test.tsx
ernst-dev: Good catch, I'll add the safety nets

pan-kot (approved):

PR #4633, opened 2026-06-15
feat: Update token component styles for one-theme

at-susie:
### Description
This PR: 
- Updated the icon size in the Token component for the dismiss button and inline variant
- Introduce border radius token for the Token component's inline variant which replaces the hardcoded approach

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/token/__tests__/token.test.tsx
```diff
@@ -152,6 +153,22 @@ describe('Token', () => {
       expect(inlineIcon).toHaveClass(styles.icon);
       expect(inlineIcon).toHaveClass(styles['icon-inline']);
     });
+
+    test('overrides icon size to x-small for inline variant in OneTheme', () => {
```
georgylobko: suggestion: let's create a section "One theme" under `describe` and set `document.body.classList.add('awsui-one-theme')` before each and `document.body.classList.remove('awsui-one-theme')` after each test.

src/token/interfaces.ts
```diff
@@ -33,6 +33,8 @@ export interface TokenProps extends BaseComponentProps {
    * When `variant="normal"`, if a description or tags are set, icon size should be `normal`.
    *
    * When `variant="inline"`, icon size should be `small`.
+   *
+   * For the OneTheme with `variant="inline"`, the icon size will automatically be overridden to `x-small`.
```
georgylobko: This strings will end up on the external website as well. Is that intentional?

src/token/internal.tsx
```diff
@@ -74,6 +81,13 @@ function InternalToken({
     }
   });
 
+  const sizedIcon = (iconNode: React.ReactNode) => {
+    if (isInline && isOneTheme && React.isValidElement(iconNode)) {
```
georgylobko: You also need to check if the `iconNode` is an instance of the `Icon` component, not just any react component

src/token/internal.tsx
```diff
@@ -74,6 +81,13 @@ function InternalToken({
...
+      return React.cloneElement(iconNode as React.ReactElement<{ size?: string }>, { size: 'x-small' });
```
georgylobko: Why do you need to cast the data here?

src/token/__tests__/token.test.tsx
at-susie: Great suggestion. I moved the OneTheme test out of `Icons` describe block into its own `describe('One theme', ...)` with beforeEach/afterEach for class management.

src/token/interfaces.ts
at-susie: Removed the OneTheme-specific sentence from the icon prop.

src/token/internal.tsx
at-susie: Added `iconNode.type === InternalIcon` so `sizedIcon` only resizes actual `InternalIcon` elements

src/token/internal.tsx
at-susie: Good call. Removed the `as React.ReactElement<{ size?: string }>` cast. Once the type guard `iconNode.type === InternalIcon` is in place, TypeScript can infer enough to make React.cloneElement work cleanly.

src/token/dismiss-button.tsx
```diff
@@ -41,7 +43,8 @@ function DismissButton(
         styles['dismiss-button'],
         legacyTestingStyles['dismiss-button'],
         testUtilStyles['dismiss-button'],
-        inline && styles['dismiss-button-inline']
+        inline && styles['dismiss-button-inline'],
+        isOneTheme && styles['one-theme']
```
georgylobko: Does it make sense to declare this variable in every component that needs one-theme styles?

let's create a new one-theme mixin here (similar to what we do for dark mode) and use it for css only, as it's less verbose and more scalable

https://github.com/cloudscape-design/components/blob/main/src/internal/styles/utils/theming.scss#L5

at-susie: That makes sense. I added the `one-theme-only` mixin in the [src/internal/styles/utils/theming.scss](https://github.com/cloudscape-design/components/pull/4633/changes#diff-d1597da1fdb35cefba576370462f9210c35a478e38f5f8c74ac0c80959e814acR15).
This way I don't have to have `isOneTheme && styles['one-theme']`.

georgylobko (approved):

PR #4637, opened 2026-06-16
feat: Add filtering to button dropdown

avinashbot:
### Description

Button dropdown action filtering implemented a similar way to select/multiselect filtering. The big underlying change is that filtering requires the use of `aria-activedescendant` rather than simply moving focus from item to item. So most of the internal changes are about disabling focusing logic when filtering is active, so that the real focus can stay on the input (and "accessibility" focus can move around using `aria-activedescendant`).

Rel: AWSUI-61960 (Chorus: qodG8KXdJXXk)

### How has this been tested?

Updated unit tests, a bit of manual accessibility testing, but more validation with other screen readers is needed.

<details>

#### Correctness

#### Security

#### Testing

</details>

---


pages/button-dropdown/filtering.page.tsx
```diff
@@ -0,0 +1,297 @@
...
+  { itemType: 'checkbox', id: 'auto-deploy', text: 'Auto-deploy on commit', checked: false },
+  { itemType: 'checkbox', id: 'verbose-logs', text: 'Verbose logging', checked: true },
+];
+
+export default function ButtonDropdownFilteringPage() {
+  const [checkboxItems, setCheckboxItems] = React.useState(withCheckboxItems);
+  const onItemClick = (event: CustomEvent<ButtonDropdownProps.ItemClickDetails>) => console.log(event.detail);
+
+  return (
+    <article>
```
pan-kot: nit: You can use SimplePage helper. It comes with a slot for page's title and simplifies the addition of screenshot area and i18n.

src/test-utils/dom/button-dropdown/index.ts
```diff
@@ -119,6 +122,27 @@ export default class ButtonDropdownWrapper extends ComponentWrapper {
     return createWrapper().find(`[data-testid="button-dropdown-disabled-reason"]`);
   }
 
+  /**
+   * Finds the filtering input rendered inside the open dropdown when `filteringType` is set.
```
pan-kot: nit: filteringType can as well be "none", so the comment is not quite accurate.

src/button-dropdown/interfaces.ts
pan-kot: Should we pass `filterText` as a separate argument here, adjacent to `item` instead of adding it to each item?

src/button-dropdown/utils/filter-items.ts
```diff
@@ -0,0 +1,61 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import { ButtonDropdownProps } from '../interfaces';
+import { isItemGroup } from './utils';
+
+function matchesString(value: string | undefined, searchText: string): boolean {
```
pan-kot: nit: matchesString, matchesItem, and collectMatchingLeaves implicitly rely on the searchText being lowercased. I think we should either explicitly document it e.g. with a var name, or instead move all three function definitions under filterItems - so that searchText can also be taken from context:

```
export function filterItems(...) {
  if (!filterText) {
    return items;
  }
  const searchText = filterText.toLowerCase();

  function matchesString(value) {...}
  function matchesItem(item) {...}
  function collectMatchingLeaves(items) {...}

  return items.reduce(...);
}
```

src/button-dropdown/utils/use-button-dropdown.ts
```diff
@@ -49,20 +66,47 @@ export function useButtonDropdown({
...
+    }
     openStateProps.toggleDropdown();
   };
 
+  // When filtering is enabled the dropdown holds several focusable elements (the filter input
+  // and its clear button), so closing on Tab is driven by focus actually leaving the dropdown
+  // rather than by the Tab keydown. The Dropdown's onBlur only fires once focus moves outside
+  // the trigger and the dropdown content, which is exactly when we want to close.
+  const onDropdownBlur = () => {
+    if (hasFiltering && isOpen) {
```
pan-kot, line 104: Do we actually need to check for hasFiltering and isOpen - can we close it unconditionally instead?

src/button-dropdown/utils/use-button-dropdown.ts
```diff
@@ -135,26 +179,38 @@ export function useButtonDropdown({
...
+        if (!hasFiltering) {
+          event.preventDefault();
+        }
         break;
       }
       case KeyCode.enter: {
+        // While filtering, pressing Enter without a highlighted item should do nothing
+        // (matching select/multiselect) rather than closing the dropdown with no selection.
+        if (hasFiltering && !targetItem) {
+          event.preventDefault();
```
pan-kot: This makes clear filter button irresponsive when pressing Enter

src/button-dropdown/utils/use-button-dropdown.ts
```diff
@@ -135,26 +179,38 @@ export function useButtonDropdown({
...
+          break;
+        }
         if (!targetItem?.disabled) {
           activate(event, true);
         }
         break;
       }
       case KeyCode.left:
       case KeyCode.right: {
+        if (hasFiltering && filteringValue) {
```
pan-kot, line 221: nit: it would make sense to add code comments explaining this change. I assume in this case we want that left/right keys would move the filter cursor, while expanding sections is no longer relevant once there is input.

src/button-dropdown/utils/use-button-dropdown.ts
```diff
@@ -174,8 +236,13 @@ export function useButtonDropdown({
         break;
       }
       case KeyCode.tab: {
-        // When expanded to viewport the focus can't move naturally to the next element.
```
pan-kot, line 178: Why was this comment removed?

src/button-dropdown/utils/use-button-dropdown.ts
```diff
@@ -185,10 +252,7 @@ export function useButtonDropdown({
     }
   };
   const onKeyUp = (event: React.KeyboardEvent) => {
-    // We need to handle activating items with Space separately because there is a bug
```
pan-kot, line 189: why was this comment removed?

src/button-dropdown/internal.tsx
```diff
@@ -381,8 +433,19 @@ const InternalButtonDropdown = React.forwardRef(
           expandToViewport={expandToViewport}
           preferredAlignment={preferCenter ? 'center' : 'start'}
           onOutsideClick={() => toggleDropdown()}
+          // In filtering mode the dropdown is a dialog with several focusable elements, so we
+          // close it once focus leaves the trigger and the dropdown content rather than on Tab.
+          onBlur={hasFiltering ? onDropdownBlur : undefined}
```
pan-kot: can this be un-conditonal?

src/button-dropdown/tooltip.tsx
```diff
@@ -14,20 +14,22 @@ interface TooltipProps {
...
 }
 
 const DEFAULT_OPEN_TIMEOUT_IN_MS = 120;
 
-export default function Tooltip({ children, content, position = 'right', className }: TooltipProps) {
+export default function Tooltip({ children, content, position = 'right', className, show }: TooltipProps) {
   const ref = useRef<HTMLSpanElement | null>(null);
   const isReducedMotion = useReducedMotion(ref);
   const { open, triggerProps } = useTooltipOpen(isReducedMotion ? 0 : DEFAULT_OPEN_TIMEOUT_IN_MS);
+  const isOpen = open || !!show;
```
pan-kot: The original visibility logic includes some delays, but the new "show" property does not come with anything like that - what is the problem that visibility state addresses with delayed open?

pan-kot: Is there a chance that we can use a shared solution, no matter if filtering is used? Ideally, that solution should use the src/tooltip component instead of this custom one.

src/button-dropdown/__tests__/button-dropdown-filtering.test.tsx
```diff
@@ -0,0 +1,598 @@
...
+  const wrapper = createWrapper(result.container).findButtonDropdown()!;
+  return { ...result, wrapper };
+}
+
+function getFilterInput(container: HTMLElement): HTMLInputElement | null {
+  return createWrapper(container).findButtonDropdown()!.findFilteringInput()?.findNativeInput().getElement() ?? null;
+}
+
+function getMenuItems(container: HTMLElement): HTMLElement[] {
+  return Array.from(container.querySelectorAll('[role="menuitem"], [role="menuitemcheckbox"]'));
```
pan-kot: Why don't we use test-utils for this selector?

src/button-dropdown/__tests__/button-dropdown-filtering.test.tsx
```diff
@@ -0,0 +1,598 @@
...
+}
+
+describe('ButtonDropdown filtering', () => {
+  beforeEach(() => {
+    jest.mocked(scrollElementIntoView).mockClear();
+  });
+
+  describe('filter input rendering', () => {
+    test('does not render filter input when filteringType is not set', () => {
```
pan-kot, line 80: Should we check for both filteringType=undefined and filteringType="none"?

src/button-dropdown/__tests__/button-dropdown-filtering.test.tsx
```diff
@@ -0,0 +1,598 @@
...
+    });
+
+    test('aria-activedescendant updates when navigating with arrow keys', () => {
+      const { container, wrapper } = renderDropdown({ filteringType: 'auto' });
+      wrapper.openDropdown();
+      const input = getFilterInput(container)!;
+
+      const dropdownEl = wrapper.findOpenDropdown()!.getElement();
+      act(() => {
+        fireEvent.keyDown(dropdownEl, { keyCode: KeyCode.down });
```
pan-kot: nit: alternatively, we can do `wrapper.findOpenDropdown()!.keyDown(KeyCode.down)`

src/button-dropdown/__tests__/button-dropdown-filtering.test.tsx
```diff
@@ -0,0 +1,598 @@
...
+      wrapper.openDropdown();
+      wrapper.findFilteringInput()!.setInputValue('Cut');
+
+      const input = getFilterInput(container)!;
+      const clearButton = wrapper.findFilteringInput()!.findClearButton()!.getElement();
+      act(() => {
+        input.focus();
+      });
+      act(() => {
+        clearButton.focus();
```
pan-kot: I believe this does not capture the regression. Wasn't the issue due to the Tab keypress handling? I think we should rather secure this with an integration test then.

src/button-dropdown/__tests__/button-dropdown-filtering.test.tsx
```diff
@@ -0,0 +1,598 @@
...
+      const input = getFilterInput(container)!;
+      const clearButton = wrapper.findFilteringInput()!.findClearButton()!.getElement();
+      act(() => {
+        clearButton.focus();
+      });
+      act(() => {
+        input.focus();
+      });
+
+      // Shift+Tabbing back to the input stays within the dropdown, so it remains open.
```
pan-kot: The code above does not technically Shift+Tab. This test and the one before can be translated to the integ tests with the same design, but with proper Tab/Shift+Tab calls.

src/button-dropdown/__tests__/button-dropdown-filtering.test.tsx
```diff
@@ -0,0 +1,598 @@
...
+
+      const input = getFilterInput(container)!;
+      act(() => {
+        fireEvent.keyDown(input, { keyCode: KeyCode.right });
+      });
+      act(() => {
+        fireEvent.keyDown(input, { keyCode: KeyCode.left });
+      });
+      // Filtering renders groups flat, so the nested items are visible regardless of expansion.
+      expect(wrapper.findOpenDropdown()).not.toBeNull();
```
pan-kot, line 382: This assertion can work if the test was: "Left and Right arrow keys do not close the dropdown".

In the original button dropdown the expandable sections can be not only expanded, but also collapsed. One thing we can check is:

```
expect(getItemsCount()).toBe(filteredItemsCount);

// Make item inside a group highlighted
input.keyDown(KeyCode.down);
input.keyDown(KeyCode.down);

input.keyDown(KeyCode.right);
expect(getItemsCount()).toBe(filteredItemsCount);

input.keyDown(KeyCode.left);
expect(getItemsCount()).toBe(filteredItemsCount);
```

this would ensure that the item count stays the same when issuing left/right keypresses.

pan-kot: I would also include this test:

```
// do not set filtering input

expect(getItemsCount()).toBe(collapsedItemsCount);

// Make category item highlighted
input.keyDown(KeyCode.down);
input.keyDown(KeyCode.down);

// Expands a group
input.keyDown(KeyCode.right);
expect(getItemsCount()).toBe(expandedItemsCount);

// Collapses a group
input.keyDown(KeyCode.left);
expect(getItemsCount()).toBe(collapsedItemsCount);
```

to ensure the Left/Right works normally when filtering is used, but filter is empty.

pan-kot: I also found a bug, but not sure if it is related to the new changes. When expandable groups include more groups - the Left/Right does not work correctly.

In that case, pressing Left/Right does not close the open group once it is opened. It is also not possible to exit the expandable groups with keyboard nav anymore.

https://github.com/user-attachments/assets/b8a494f9-7d85-4693-afe3-da0f3e868cc9

src/button-dropdown/__tests__/filter-items.test.ts
```diff
@@ -0,0 +1,155 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import { ButtonDropdownProps } from '../interfaces';
+import { filterItems } from '../utils/filter-items';
```
pan-kot: All these tests are technically possible to run from the component level: render button dropdown with certain items, set filtering - check the actual rendered items. Should we do this instead of testing the helper directly? This can give us better real coverage, as it not only ensures that the helper works, but also that it is used correctly by the component.

src/button-dropdown/__tests__/render-item.test.tsx
```diff
@@ -40,6 +40,19 @@ describe('ButtonDropdown renderItem', () => {
     expect(elementWrapper).toHaveTextContent('Custom');
   });
 
+  test('renders custom item content for link items', () => {
```
pan-kot, line 43: Why did we add this test? Is that also the expected behaviour? I find it strange that we still render a link when the item renderer is overridden.

src/button-dropdown/utils/use-button-dropdown.ts
```diff
@@ -91,7 +135,6 @@ export function useButtonDropdown({
   };
 
   const actOnParentDropdown = (event: React.KeyboardEvent) => {
-    // if there is no highlighted item we act on the trigger by opening or closing dropdown
```
amanabiy, line 95: Did the behavior change here? or we don't need the comment?

src/button-dropdown/utils/use-button-dropdown.ts
```diff
@@ -110,7 +153,6 @@ export function useButtonDropdown({
   const activate = (event: React.KeyboardEvent, isEnter?: boolean) => {
     setIsUsingMouse(false);
 
-    // if item is a link we rely on default behavior of an anchor, no need to prevent
```
amanabiy, line 114: Same, Did the behavior change here? or we don't need the comment?

src/button-dropdown/utils/use-button-dropdown.ts
```diff
@@ -49,20 +66,47 @@ export function useButtonDropdown({
...
   } = useHighlightedMenu({
-    items,
-    hasExpandableGroups,
+    items: filteredItems,
+    hasExpandableGroups: effectiveHasExpandableGroups,
     isInRestrictedView,
   });
 
-  const { isOpen, closeDropdown, ...openStateProps } = useOpenState({ onClose: reset });
+  const prevFilteringValue = useRef(filteringValue);
```
amanabiy, line 77: Should we use usePrevious here?

`const prevFilteringValue = usePrevious(filteringValue);`

src/button-dropdown/__tests__/render-item.test.tsx
avinashbot: Spent a lot of time fighting with codecov because of the lines I touched; a few stray tests is the end result. I need this to keep my patch coverage check from failing 😓

src/button-dropdown/utils/use-button-dropdown.ts
avinashbot: Considered it, but not much would get simpler here; I still need the "reset" function to be called, so I still need to have an effect and compare previous/current.

src/button-dropdown/utils/use-button-dropdown.ts
avinashbot: I _think_ so (focus leaving the dropdown should close it), but I would rather keep the logic separate between the two "modes" (menu focus and aria-activedescendant), even if it looks safe now to have some code run for both.

src/button-dropdown/__tests__/button-dropdown-filtering.test.tsx
avinashbot: Oh yeah, I discovered this too. It's a pre-existing bug and it's reproducible in production. Don't want to bloat up this PR further, working on this separately.

src/button-dropdown/__tests__/filter-items.test.ts
avinashbot: These were mostly AI-written as an "in depth" solution to keep code coverage passing. Ended up rewriting a chunk of the unit tests, restructuring the code slightly; looks like I no longer need this. Deleted.

src/button-dropdown/__tests__/button-dropdown-filtering.test.tsx
avinashbot: For the keyboard functionality comments; it's difficult to ensure these things work as expected over unit testing. Decided to just move focus-move tests to integration tests instead.

src/button-dropdown/__tests__/button-dropdown-filtering.test.tsx
avinashbot: The test utils select the parent containers rather than the aria role elements themselves. But these can be updated to use test utils as much as possible.

src/button-dropdown/tooltip.tsx
avinashbot: Refactored this. Still keeping the "controlled" mode that aria-activedescendant needs, but it now respects the open delay.

src/button-dropdown/__tests__/button-dropdown-filtering.test.tsx
```diff
@@ -0,0 +1,727 @@
...
+    });
+
+    test('opening with ArrowUp does not highlight an item when filtering is enabled', () => {
+      const { wrapper } = renderDropdown({ filteringType: 'auto' });
+      wrapper.findNativeButton().keydown(KeyCode.up);
+      expect(wrapper.findOpenDropdown()).not.toBeNull();
+      expect(wrapper.findHighlightedItem()).toBeNull();
+    });
+
+    test('Escape closes the dropdown', () => {
```
pan-kot, line 305: nit: should we include a test to ensure the dropdown closes also when filtering input is not empty?

src/button-dropdown/__tests__/button-dropdown-filtering.test.tsx
```diff
@@ -0,0 +1,727 @@
...
+      const { wrapper } = renderDropdown({ expandToViewport: true });
+      wrapper.openDropdown();
+      expect(wrapper.findOpenDropdown()).not.toBeNull();
+
+      wrapper.findOpenDropdown()!.keydown(KeyCode.tab);
+      expect(wrapper.findOpenDropdown()).toBeNull();
+      expect(document.activeElement).toBe(wrapper.findNativeButton().getElement());
+    });
+
+    test('Tab does not close the dropdown when filtering is enabled', () => {
```
pan-kot, line 404: What is the purpose of this test? The actual component does close the dropdown when using Tab, unless the filtering input is non-empty. This is already captured with integ tests.

src/button-dropdown/__tests__/button-dropdown-filtering.test.tsx
```diff
@@ -0,0 +1,727 @@
...
+
+    test('does not highlight anything when there is no filtering value', () => {
+      const { container, wrapper } = renderDropdown({ filteringType: 'auto', items: richItems });
+      wrapper.openDropdown();
+      expect(container.querySelectorAll('mark')).toHaveLength(0);
+    });
+  });
+
+  describe('filtered expandable groups', () => {
+    test('renders matching nested items with ids and non-focusable tab indexes', () => {
```
pan-kot: Why do we check for tab indices here? Is there a dependency between items matching and tab indices?

src/button-dropdown/__tests__/button-dropdown-filtering.test.tsx
```diff
@@ -0,0 +1,727 @@
...
+    test('prevents focus stealing on mouse down for expandable categories while filtering is enabled', () => {
+      const { wrapper } = renderDropdown({
+        filteringType: 'auto',
+        items: expandableItems,
+        expandableGroups: true,
+      });
+      wrapper.openDropdown();
+      const categoryEl = wrapper.findExpandableCategoryById('states')!.getElement();
+      // With filtering enabled the focus stays on the input, so mouse down is prevented on items.
+      expect(fireEvent.mouseDown(categoryEl)).toBe(false);
```
pan-kot: Should we check for the effect instead? E.g. checking that the focus remains on the input.

src/button-dropdown/__tests__/button-dropdown-filtering.test.tsx
```diff
@@ -0,0 +1,727 @@
...
+    });
+
+    test('does not prevent mouse down for expandable categories without filtering', () => {
+      const { wrapper } = renderDropdown({ items: expandableItems, expandableGroups: true });
+      wrapper.openDropdown();
+      const categoryEl = wrapper.findExpandableCategoryById('states')!.getElement();
+      expect(fireEvent.mouseDown(categoryEl)).toBe(true);
```
pan-kot: Same here - I would rather check that the item does indeed receive focus

src/button-dropdown/__tests__/button-dropdown-filtering.test.tsx
```diff
@@ -0,0 +1,727 @@
...
+      const footer = wrapper.findFooterRegion();
+      expect(footer).not.toBeNull();
+      expect(footer!.getElement()).toHaveTextContent('1 of 4 items');
+    });
+
+    test('does not render filtering results text when there are no matches', () => {
+      const { wrapper } = renderDropdown({
+        filteringType: 'auto',
+        filteringResultsText: (matchesCount, totalCount) => `${matchesCount} of ${totalCount} items`,
+        noMatch: 'No items found',
```
pan-kot, line 555: Why do we need both, actually? In autosuggest we use filteringResultsText for both cases.

src/button-dropdown/__tests__/button-dropdown-filtering.test.tsx
```diff
@@ -0,0 +1,727 @@
...
+      });
+      wrapper.openDropdown();
+      wrapper.findFilteringInput()!.setInputValue('zzz');
+      const footer = wrapper.findFooterRegion();
+      // The noMatch content is shown, not the filteringResultsText
+      expect(footer).not.toBeNull();
+      expect(footer!.getElement()).toHaveTextContent('No items found');
+    });
+
+    test('counts leaf items in groups for totalCount', () => {
```
pan-kot, line 565: This is inconsistent with multiselect - that does include groups to filtering results text. Most importantly, the filtering also works inconsistently: in multiselect one can filter for the group name, e.g. "Group 1" - it then shows all items that belong to this group. Why is this different in the button dropdown? E.g. in the demo I type "Edit" to the filter to find all items that belong to the "Edit" category. The result is empty though.

src/button-dropdown/__tests__/button-dropdown-filtering.test.tsx
```diff
@@ -0,0 +1,727 @@
...
+
+      expect(renderItem).toHaveBeenCalledWith(
+        expect.objectContaining({
+          item: expect.objectContaining({ type: 'action' }),
+          filterText: 'Cu',
+        })
+      );
+    });
+
+    test('passes filterText to renderItem for checkbox items', () => {
```
pan-kot: what is the purpose of checking for different item types now that filtering text is passed as a separate arg?

src/button-dropdown/item-element/index.tsx
```diff
@@ -114,20 +125,42 @@ interface MenuItemProps {
...
   const isCheckbox = isCheckboxItem(item);
   const isCurrentBreadcrumb = !isCheckbox && item.isCurrentBreadcrumb;
 
   useEffect(() => {
     if (highlighted && menuItemRef.current) {
-      menuItemRef.current.focus();
+      if (filteringEnabled) {
+        // In filtering mode focus stays on the filter input, so we scroll the
+        // highlighted item into view manually instead of relying on focus().
+        scrollElementIntoView(menuItemRef.current);
```
pan-kot, line 156: Should we include a test for this behaviour?

src/button-dropdown/item-element/index.tsx
```diff
@@ -171,12 +213,10 @@ function MenuItem({ index, item, disabled, highlighted, linkStyle, renderItem, p
     'aria-current': isCurrentBreadcrumb,
     lang: item.lang,
     ref: menuItemRef,
-    // We are using the roving tabindex technique to manage the focus state of the dropdown.
-    // The current element will always have tabindex=0 which means that it can be tabbed to,
-    // while all other items have tabindex=-1 so we can focus them when necessary.
-    tabIndex: highlighted ? 0 : -1,
+    // When filteringEnabled is true (filtering mode), focus stays on the filter input
```
pan-kot: Why was the original description removed? This makes it inconsistent with https://github.com/cloudscape-design/components/pull/4637/changes#diff-7ce4272800a73b0bfecd20f86b5aa9443684efb16601ca729dba254e8349f7e6R89

src/button-dropdown/internal-interfaces.ts
```diff
@@ -25,6 +25,10 @@ export interface CategoryProps extends HighlightProps {
   variant?: ItemListProps['variant'];
   position?: string;
   renderItem?: ButtonDropdownProps.ItemRenderer;
+  filteringText?: string;
+  filteringEnabled?: boolean;
```
pan-kot, line 29: nit: I would either make filteringText required, or remove filteringEnabled in favour of `filteringText !== undefined` check.

src/button-dropdown/internal.tsx
```diff
@@ -344,9 +373,49 @@ const InternalButtonDropdown = React.forwardRef(
...
+    const filteredText = isFiltered ? filteringResultsText?.(matchesCount, totalCount) : undefined;
+
+    const dropdownStatus = useDropdownStatus({
+      statusType: 'finished',
+      isNoMatch,
+      noMatch,
+      filteringResultsText: filteredText,
+    });
+
+    // Only create a filteringDescription element if filtering is actually enabled,
```
pan-kot, line 392: nit: I think this comment is redundant as the code is clear enough

src/test-utils/dom/button-dropdown/index.ts
```diff
@@ -119,6 +122,27 @@ export default class ButtonDropdownWrapper extends ComponentWrapper {
...
+   *
+   * This utility does not open the dropdown. To find the filtering input, call `openDropdown()` first.
+   */
+  findFilteringInput(): InputWrapper | null {
+    return this.findOpenDropdown()?.findComponent(`.${inputStyles['input-container']}`, InputWrapper) ?? null;
+  }
+
+  /**
+   * Finds the footer region rendered at the bottom of the open dropdown. When filtering is enabled and there
+   * are no matching items, this contains the `noMatch` content. Returns null if there is no open dropdown or the
```
pan-kot: This description needs to be updated to mention filteringResultsText

src/button-dropdown/__tests__/button-dropdown-filtering.test.tsx
avinashbot: The test here (and others near this one) is linked to all the keyboard handlers we have in our useButtonDropdown hook. They listen to arbitrary keypresses, so this coverage helps avoid unexpected breakage. I see integ tests as a place to test things that rely on built-in browser behavior (like focus navigation), but for explicit keypress listeners, we should still have unit test coverage.

src/button-dropdown/__tests__/button-dropdown-filtering.test.tsx
avinashbot: There is a dependency between filtering and tabindexes, yeah. The value depends on whether filtering is enabled (-1) or not (0/-1). The code for setting tabindexes is in a couple of places: items and expandable category items, so this makes sure everything is consistent.

src/button-dropdown/__tests__/button-dropdown-filtering.test.tsx
avinashbot: Good callout, was too focused on direct coverage at this point and the exact code behavior. Updating.

src/button-dropdown/__tests__/button-dropdown-filtering.test.tsx
avinashbot: This was a decision I had to make: you're right that autosuggest doesn't do this, but select and multiselect do. So for consistency, I chose to stick with the conventions established by select/multiselect for the APIs added for this whole feature.

src/button-dropdown/item-element/index.tsx
avinashbot: I accidentally dropped these comments in the previous version of the PR because I used Claude to refactor the code around and didn't notice it. I brought back that comment where I thought things were deleted, but missed this one. Fixing!

src/button-dropdown/internal.tsx
avinashbot: The comment describes what the hook does internally; easy to check the source, but I felt it important to call out here, since a seemingly innocent refactor to not match the existence of the element with the existence of the id could create an accessibility issue.

src/button-dropdown/internal-interfaces.ts
avinashbot: Felt it was cleaner this way to have a boolean rather than combining these two attributes into an `| undefined` union. `hasFiltering` is easy to determine at the top-level from the `filteringType` prop, and for child elements, it's easier to pass the boolean straight through to effect dependencies (for example: `expandable-category-element.tsx`) rather than create a separate variable (`const hasFiltering = filteringText !== undefined`) just for effects (reversing additional logic in `use-button-dropdown.tsx`, which would need to set `filteringText: hasFiltering ? filteringText : undefined`).

Don't feel super strongly about this though, it's just a minor refactor. Let me know if that makes sense or you'd still prefer condensing this into one state variable.

src/button-dropdown/__tests__/button-dropdown-filtering.test.tsx
avinashbot: Test coverage! This logic is actually duplicated three times, and each test tests a different place where the check is defined.

Not sure how much sense deduplicating it makes, it's just one additional prop in the event handler call.

src/button-dropdown/item-element/index.tsx
avinashbot: There are a number of tests that mock and verify `scrollElementIntoView` in the test file already (main one being "scrolls the highlighted item into view when navigating with arrow keys while filtering"). Should be fine, I think.

src/button-dropdown/__tests__/button-dropdown-filtering.test.tsx
avinashbot: Oh, good callout. I don't know what the intended behavior should be here actually, need to confirm with design. We had this behavior implicltly from the original prototype and signoff but I don't know if this was ever discussed.

But I would imagine behaving like select/multiselect could be pretty unhelpful here, especially for really large dropdowns (like with EC2), where typing in "instance" would get you an really big list of items rather than a specific action you were maybe looking for.

Let's assume this is correct for now, but I won't merge until I get an approval on this specific behavior from the designer.

src/button-dropdown/__tests__/button-dropdown-filtering.test.tsx
pan-kot: I mean, what is the point of doing: `wrapper.findOpenDropdown()!.keydown(KeyCode.tab)`, since as we know the focused element is actually the input? This test emulates an action that the user cannot trigger.

src/button-dropdown/__tests__/button-dropdown-filtering.test.tsx
pan-kot: Yes, I saw the dependency between filtering and tab indices - I was not sure if there is a dependency between matching items and tab indices. As tab indices are added to all items simply when filtering feature is used - we should check it for all items, not just those matching when filtering input is non-empty.

src/button-dropdown/__tests__/button-dropdown-filtering.test.tsx
pan-kot: I think this can be misleading. E.g. when I search for "Security" in your demo - it matches one item our of two that are defined inside the "Security" group - because that one happens to have "security" keyword in the description. It is true that matching groups can result in many items in the output - but this can be mitigated by clarifying the input, e.g. saying "change security" or "security groups" would then find the item alone, no longer matching every item in the group.

src/button-dropdown/__tests__/button-dropdown-filtering.test.tsx
pan-kot: Can we have a single test then that defines a button dropdown featuring all item types?

src/button-dropdown/__tests__/button-dropdown-filtering.test.tsx
avinashbot: Ohh, I see what you mean. There's already an existing test for that; I'll just replace this with a simple filtering test.

src/button-dropdown/__tests__/button-dropdown-filtering.test.tsx
avinashbot: Gotcha. That's sensible; it works either way because the listener is defined at the top level, but I'll pick out the elements that should have focus and call handlers on those.

src/button-dropdown/__tests__/button-dropdown-filtering.test.tsx
avinashbot: Sure. But to make sure callback assertions are correct, I still need to trigger each item individually. So not much is deduplicated.

src/button-dropdown/__tests__/button-dropdown-filtering.test.tsx
avinashbot: The trouble is also that it's not consistent. For nested expandable groups (for example: "Networking" in the nested expandable group demo in the filtering dev page), the inner group name is intentionally discarded when filtering, because we don't want to have nested flattened groups. So the outer group name is filterable, but the inner one isn't. For example, typing in "Diagnostics" in the nested demo gets you no results. The alternative is to match the items under the nested group, but not show the parent group, which makes it equally confusing why they were matched.

Screenshot 2026-07-09 at 14 38 31

avinashbot: Had a chat with the designer and decided to keep this behavior.

- This (group filtering) wasn't part of the initial scope internal or in discussion with the building team. Adding this (especially having to re-address the group "collapsing" topic we already had an agreement on earlier) is going to expand the scope for no clear use-case.
- The component is different to select and multiselect; the use-case of getting to specific actions without digging through menus is different to the select/multiselect components more "exploratory" use-case.
- Because the API for this is pretty opaque, we could treat this as a two-way door to introduce in the future if we really had strong data pushing us to do so. And this ("auto" filtering) is just a stepping stone to supporting more complex filtering in the future.

pan-kot (approved):

PR #4640, opened 2026-06-17
feat: Add `stretch` option to space between's `alignItems` property

timogasda:
### Description
This change adds a new option of `"stretch"` to the `alignItems` property on space between.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/space-between/__tests__/space-between.test.tsx
+++ b/src/space-between/__tests__/space-between.test.tsx
@@ -85,6 +85,29 @@ describe('SpaceBetween', () => {
     expect(content[0].getElement()).toHaveAttribute('id', 'button-one');
     expect(content[1].getElement()).toHaveAttribute('id', 'button-two');
   });
+
+  it.each(['center', 'start', 'end', 'stretch'] as const)('applies align-%s class', alignItems => {
+    const { container } = render(
+      <SpaceBetween size="s" alignItems={alignItems}>
+        <button />
+        <button />
+      </SpaceBetween>
+    );
+    expect(container.firstChild).toHaveClass(styles[`align-${alignItems}`]);
+  });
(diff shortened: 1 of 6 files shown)


src/space-between/__tests__/space-between.test.tsx
```diff
@@ -85,6 +85,29 @@ describe('SpaceBetween', () => {
...
+    const { container } = render(
+      <SpaceBetween size="s" alignItems={alignItems}>
+        <button />
+        <button />
+      </SpaceBetween>
+    );
+    expect(container.firstChild).toHaveClass(styles[`align-${alignItems}`]);
+  });
+
+  it('does not apply an alignment class when alignItems is undefined', () => {
```
SpyZzey, line 99: When we don't define the alignItems props, it basically falls back to "stretched" (browser default). Should we explicitly set it to "stretch" instead of relying on the implicit browser defaults?

timogasda: I'm not sure I understand. The default value for `align-items` is `normal`, not `stretch`.

SpyZzey: Yes, which in our case behaves like `stretch` (https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/align-items#values). Except that I missed 

> behaves like start on replaced absolutely-positioned boxes

which would change the behavior on rare cases were SpaceBetween has a absolutely-positioned replaced element (e.g., img, canvas, input?) as a direct child. So I think we are good, keeping it as is to ensure we don't break some niche cases.

SpyZzey (approved):

PR #4646, opened 2026-06-18
refactor: Public types restructuring, part 1

pan-kot:
### Description

This PR lifts shared public types to the src/types folder - thus helping to make the separation between public and internal types clearer. In the follow-up PRs I will also remove all internal types from the public paths and establish a contract to be used by all component repositories (secured with ESLint rules).

Rel: D475262969, iu7zAZtdKN3Z

### How has this been tested?

* Dry-run to live (7896903128)

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/internal/analytics/interfaces.ts
```diff
@@ -1,48 +1,9 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+import { AnalyticsMetadata, FlowType } from '../../types/analytics';
+
 export type FunnelType = 'single-page' | 'multi-page' | 'modal';
-export type FlowType = 'create' | 'edit' | 'delete' | 'home' | 'dashboard' | 'view-resource';
```
SpyZzey, line 5: Why do we move FlowType but not FunnelType?

pan-kot: Because FlowType is referenced in public interfaces (of app-layout). The FunnelType is (so far) internal.

src/internal/components/cartesian-chart/interfaces.ts
```diff
@@ -1,259 +1,15 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
-import { BaseComponentProps } from '../../base-component';
-import { NonCancelableEventHandler } from '../../events';
 
-export interface CartesianChartProps<T extends ChartDataTypes, Series> extends BaseComponentProps {
```
SpyZzey, line 6: Why do we move CartesianChartProps interface + namespace, but not other internal component interfaces (e.g., src/internal/components/masked-input/interfaces.ts)

src/internal/components/radio-button/index.tsx
```diff
@@ -7,12 +7,12 @@ import { useMergeRefs } from '@cloudscape-design/component-toolkit/internal';
 import { useSingleTabStopNavigation } from '@cloudscape-design/component-toolkit/internal';
 import { copyAnalyticsMetadataAttribute } from '@cloudscape-design/component-toolkit/internal/analytics-metadata';
 
+import { RadioButtonProps } from '../../../radio-button/interfaces';
```
SpyZzey, line 10: This does not use the types/ subfolder - do we really want to use the "public" RadioButtonProps interface? Why do we stlil have the internal component then?

SpyZzey: Why do we have this internal component in the first place?

src/internal/components/cartesian-chart/interfaces.ts
pan-kot: The masked input component is consumed by date- and time inputs internally via compositions - there is no types inheritance.

On the contrary, area-, bar-, line- and mixed charts inherit from CartesianChartProps.

src/internal/components/radio-button/index.tsx
pan-kot: There is no reason to have it as internal, I think. It is just previously we only exposed public radio group, but not radio button. At the same time, the radio button was still used internally by tiles, table, and other components. We can refactor that as a follow-up and bring internal radio button impl to the src/radio-button - so that it is consistent with other components.

SpyZzey (approved):

ernst-dev (approved):

PR #4663, opened 2026-06-23
refactor: introduce reusable theme url param

mxschll:
### Description

Introduces a `theme` url param for dev pages and replaces the `oneTheme` param. Does not affect any visual refresh / classic logic. 

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


pages/app/components/theme-switcher.tsx
```diff
@@ -7,15 +7,15 @@ import { Density, Mode } from '@cloudscape-design/global-styles';
...
 import SpaceBetween from '~components/space-between';
 
-import AppContext from '../app-context';
+import AppContext, { Theme } from '../app-context';
 
 export default function ThemeSwitcher() {
   const { mode, urlParams, setUrlParams, setMode } = useContext(AppContext);
 
-  function activateTheme(theme: 'visualRefresh' | 'oneTheme' | 'classic') {
+  function activateTheme(selected: 'visualRefresh' | 'oneTheme' | 'classic') {
```
jperals: Nitpick: the parameter could still be called `theme`. `selected` suggests a boolean

pages/app/app-context.tsx
```diff
@@ -54,12 +74,17 @@ export function parseQuery(query: string) {
...
-  return mapValues(queryParams, value => {
+  const parsed = mapValues(queryParams, value => {
     if (value === 'true' || value === 'false') {
       return value === 'true';
     }
     return value;
   });
+
+  const themeValues = Object.values(Theme) as string[];
+  parsed.theme = themeValues.includes(parsed.theme) ? parsed.theme : Theme.Default;
```
jperals: I think this extra logic would be better merged with the `if` block above. Its purpose above is to transform values as needed, currently only transform specific strings into booleans, but could be extended to include the new case. In this new case we need not only the value but also the value, but we can use lodash's `mapKey` instead of `mapValue` (or switch ti native functions such as `Object.keys(queryParams).map` or `Object.entries(queryParams).map`):

```
  const themeValues = Object.values(Theme) as string[];

  return mapKeys(queryParams, k => {
    const value = queryParams[k];
    if (k === 'theme') {
      return themeValues.includes(value) ? value : Theme.Default;
    }
    if (value === 'true' || value === 'false') {
      return value === 'true';
    }
    return value;
  });
```

pages/app/components/theme-switcher.tsx
mxschll: I will change it back

pages/app/app-context.tsx
mxschll: Good point, merged them.

jperals (approved):

PR #4671, opened 2026-06-26
chore: Apply flashbar new style in one-theme

at-susie:
### Description

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


style-dictionary/visual-refresh/contexts/flashbar-warning.ts
```diff
@@ -16,9 +16,15 @@ const baseTokens: StyleDictionary.ColorsDictionary = {
   // Focus outline matches the text color to align with other flashbar types and to ensure color contrast.
   colorBorderItemFocused: '{colorTextNotificationYellow}',
 
-  // Dismiss button
-  colorTextInteractiveInvertedDefault: '{colorNeutral650}',
```
mxschll, line 20: I don't understand why this existed before and we're removing it now.

src/flashbar/styles.scss
```diff
@@ -36,6 +36,7 @@
   border-start-end-radius: awsui.$border-radius-flashbar;
   border-end-start-radius: awsui.$border-radius-flashbar;
   border-end-end-radius: awsui.$border-radius-flashbar;
+  border-inline-start: awsui.$border-width-flashbar-inline-start solid transparent;
```
mxschll, line 39: If this is always transparent, why have the border at all?

style-dictionary/visual-refresh/contexts/flashbar-warning.ts
at-susie: It's actually not a true removal but it's retargeted. The `flashbar-icon` button variant was switched from the inverted interactive tokens to the plain `colorTextInteractive*` ones in the `constants.scss`. 

So the warning override had to move from .`..InvertedDefault/Hover` to `...Default/Hover` the inverted overrides are now dead. 

The warning flash is light in both modes, so we still pin to `colorNeutral650/950` for dismiss-icon contrast.

src/flashbar/styles.scss
at-susie: As we chatted, the transparent is just as the base default. Each `.flash-type-*` overrides `border-inline-start-color` with its status color, so this is the colored left accent (new one-theme style, matching alerts). Width comes from `border-width-flashbar-inline-start` which is 2px in one-theme, 0px in visual-refresh

mxschll (approved):

PR #4676, opened 2026-06-30
fix: Ignore IME composition enter in select and multiselect filtering

mahirhir:
### Description

The filtering input of `Select` and `Multiselect` selects the highlighted option when Enter is pressed, but it does not account for IME composition. When a user composes text with an IME (for example Japanese, Chinese or Korean) and presses Enter to confirm a candidate, that Enter is also treated as an option selection: the highlighted option is selected and the dropdown closes, instead of the keypress just committing the composed text into the filter.

The other text inputs in this repository already guard against this. `autosuggest-input` reads `useIMEComposition` and ignores Enter while `isComposing()` is true, and `prompt-input` checks `event.nativeEvent.isComposing || props.isComposing?.()`. The select filter input was passing its keydown straight to `useMenuKeyboard` without any such guard.

This applies the same guard to the filter handler in `useSelect` (shared by `Select` and `Multiselect`). `event.detail.isComposing` covers the keydown fired during composition, and `useIMEComposition(filterRef)` covers the brief window after `compositionend` where some browsers fire a spurious Enter, matching `autosuggest-input` and `prompt-input`. Other key handling is unchanged.

### How has this been tested?

Added a unit test in `src/select/__tests__/use-select.test.ts`: with an option highlighted, an Enter whose `isComposing` detail is `true` no longer calls `updateSelectedOption` and keeps the dropdown open, while a regular Enter still selects the option. The existing `use-select` unit tests continue to pass.

### DIFF
--- a/src/select/__tests__/use-select.test.ts
+++ b/src/select/__tests__/use-select.test.ts
@@ -10,7 +10,7 @@ import { KeyCode } from '../../internal/keycode';
 import { BaseKeyDetail } from '../../types/events';
 import { useSelect } from '../utils/use-select';
 
-const createTestEvent = (keyCode: KeyCode): CustomEvent<BaseKeyDetail> =>
+const createTestEvent = (keyCode: KeyCode, isComposing = false): CustomEvent<BaseKeyDetail> =>
   createCustomEvent({
     cancelable: true,
     detail: {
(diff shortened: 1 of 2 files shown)


mahirhir:
Closing this. It's a small IME composition guard and the checks aren't passing on my side, so I'd rather take it off your plate than leave it lingering. Thanks.

PR #4679, opened 2026-07-01
chore: Introduce visualAccent prop to box

at-susie:
### Description
This PR introduces a new property `visualAccent` prop to the box component. The new prop allows builders to visually emphasize specific elements through a curated set of background and content color combinations. 

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


style-dictionary/visual-refresh/colors.ts
```diff
@@ -370,6 +370,44 @@ const tokens: StyleDictionary.ColorsDictionary = {
   colorTextBadgeBlue: '{colorTextNotificationDefault}',
   colorTextBadgeRed: '{colorTextNotificationDefault}',
   colorBackgroundCodeView: { light: '#f8f8f8', dark: '#282c34' },
+
+  // ── Visual accent ───────────────────────────────────
+  colorBackgroundAccentRed: { light: '{colorError100}', dark: 'rgba(255, 61, 61, 0.2)' },
```
mxschll, line 375: Why do we need these tokens in visual refresh? This is a 1T feature.

style-dictionary/visual-refresh/metadata/colors.ts
```diff
@@ -276,6 +276,186 @@ const metadata: StyleDictionary.MetadataIndex = {
     public: true,
     themeable: true,
   },
+  colorBackgroundAccentRed: {
+    description: 'The background color of the red accent in the Box `awsui-accent` variant.',
+    public: true,
```
mxschll: Do we want to make them themeable? The point of accent colors is, that they provide a predefined set of colors. Builders already can customize using theming / style api. I am not sure if this feature should expose themeable tokens.

src/box/internal.tsx
```diff
@@ -12,6 +13,11 @@ import styles from './styles.css.js';
 
 type InternalBoxProps = BoxProps & InternalBaseComponentProps;
 
+// Curated t-shirt size keywords mirror the Box spacing scale and are applied via CSS classes so
+// they resolve to the corresponding space tokens. Any other value is treated as a raw CSS
+// `border-radius` string and applied through a custom property.
+const RADIUS_KEYWORDS: ReadonlyArray<string> = ['n', 'xxxs', 'xxs', 'xs', 's', 'm', 'l', 'xl', 'xxl', 'xxxl'];
```
mxschll: I find `n` not really clear. Can we just not set the border radius (`undefined`)? I think the box has no default border radius, so that would work.

src/box/internal.tsx
```diff
@@ -73,11 +96,15 @@ const getClassNamesSuffixes = (value: BoxProps.SpacingSize | BoxProps.Spacing) =
   return sides.filter(side => !!value[side]).map(side => `${side}-${value[side]}`);
 };
 
-const getTag = (variant: BoxProps.Variant, tagOverride: BoxProps['tagOverride']) => {
+const getTag = (variant: BoxProps.Variant, tagOverride: BoxProps['tagOverride'], hasVisualAccent: boolean) => {
   if (tagOverride) {
     return tagOverride;
   }
 
+  if (hasVisualAccent) {
```
mxschll: You can merge this: if `(hasVisualAccent || variant === 'awsui-value-large')`

mxschll: Why does it have to be a span if `hasVisualAccent === true`?

src/box/visual-accent.scss
```diff
@@ -0,0 +1,179 @@
...
+// overrides are needed here.
+//
+// The selectors chain the `.visual-accent` base class with the `.visual-accent-{color}`
+// modifier (both are always applied together whenever `visualAccent` is set) to reach a
+// specificity of (0,3,0). This intentionally beats the `color` prop's `.box.color-*`
+// selector (0,2,0) so the accent's coordinated foreground color always wins, regardless of
+// stylesheet source order. Winning here preserves the guaranteed background/foreground color
+// contrast of the accent pair. Descendants (nested Box, Icon) inherit this color.
+
+.box.visual-accent.visual-accent-red {
```
mxschll: Nit: it's more maintainable like this:

```
$accents: ...;

@each $accent in $accents {
  .box.visual-accent.visual-accent-#{$accent} {
    background-color: awsui.$color-background-accent-#{$accent};
    color: awsui.$color-text-accent-#{$accent};
  }
}
```

style-dictionary/visual-refresh/color-palette.ts
```diff
@@ -68,6 +68,91 @@ const tokens: StyleDictionary.ColorPaletteDictionary = {
     'colorPurple700',
     'colorAmber400',
     'colorAmber500',
+    'colorNeutralGrey50',
```
mxschll: Why do we need this in VR?

src/box/internal.tsx
```diff
@@ -32,11 +39,26 @@ export default function InternalBox({
   const marginsClassNamesSuffices = getClassNamesSuffixes(margin);
   const paddingsClassNamesSuffices = getClassNamesSuffixes(padding);
 
+  const accentAspectRatio = visualAccent?.aspectRatio ?? 'auto';
+  const accentBorderRadius = visualAccent?.borderRadius;
+  const isRadiusKeyword = accentBorderRadius !== undefined && RADIUS_KEYWORDS.includes(accentBorderRadius);
+  // Arbitrary (non-keyword) values are applied through a custom CSS property, which the base
+  // `.visual-accent` rule reads. This keeps custom rounding CSP-safe. Keyword values are handled
```
mxschll: I don't understand this comment.

src/box/interfaces.ts
```diff
@@ -156,6 +178,70 @@ export namespace BoxProps {
...
+     * The coordinated background and foreground color pair applied to the accent wrapper.
+     * Each color works in both light and dark modes.
+     *
+     * The background color is applied directly to the wrapper. The foreground color is applied
+     * as the wrapper's CSS `color` and is therefore only picked up by content that inherits the
+     * current color, such as an `Icon` or a nested `Box` with `color="inherit"`. Content that
+     * defines its own color (for example a `Box` with an explicit `color`, or a `Link`) keeps that
+     * color and is not recolored by the accent.
+     */
+    color: BoxProps.VisualAccent.Color;
```
mxschll, line 193: Why is color required, but aspectRatio and borderRadius are optional? Could there be a case where I don't want to set a color but want to set the borderRadius?

at-susie: color is essential for the visualAccent prop. aspectRatio and borderRadius are not visible without color.

mxschll: I'd expect aspect ratio to impact the size of the element though?

style-dictionary/visual-refresh/colors.ts
at-susie: Moving these to 1T style-dictionary

at-susie: I looked into moving the values into one-theme, but it turned out that they need to stay rooted in `visual-refresh` as the default package is built with `classic` as the primary theme (which inherits from visual-refresh), while one-theme is only ever a secondary theme.

style-dictionary/visual-refresh/metadata/colors.ts
at-susie: Making all of these non-themeable and non-public.

src/box/internal.tsx
at-susie: I like your suggestion and integrated it. So now default is 0px border-radius. No need to have 'n'

src/box/internal.tsx
at-susie: This was a leftover from an old experiment. Updated not to make it span by default when `hasVisualAccent === true`. Now it just follows the box variant

src/box/internal.tsx
at-susie: Updated

src/box/interfaces.ts
at-susie: aspectRatio impacts size, but they don't have to be required.

src/box/visual-accent.scss
at-susie: Good suggestion. Integrated. It helps reducing repeating css

pages/box/visual-accent.page.tsx
```diff
@@ -0,0 +1,170 @@
...
+  { id: 'multi-session', content: 'Multi-session data', icon: 'multiscreen', color: 'purple' },
+  { id: 'alert', content: 'Alert center', icon: 'security', color: 'red' },
+  { id: 'communication', content: 'Communication', icon: 'contact', color: 'mint' },
+];
+
+// ─── Page ──────────────────────────────────────────────────────────────────────
+
+export default function StyleBoxPage() {
+  return (
+    <div style={{ display: 'flex', justifyContent: 'center' }}>
```
mxschll: Can we use SimplePage and have a screenshot area inside, so we can take screenshot tests?

src/box/internal.tsx
```diff
@@ -32,11 +40,28 @@ export default function InternalBox({
   const marginsClassNamesSuffices = getClassNamesSuffixes(margin);
   const paddingsClassNamesSuffices = getClassNamesSuffixes(padding);
 
+  const accentAspectRatio = visualAccent?.aspectRatio ?? 'auto';
+  const accentBorderRadius = visualAccent?.borderRadius;
+  const isRadiusKeyword = accentBorderRadius !== undefined && RADIUS_KEYWORDS.includes(accentBorderRadius);
```
mxschll: Can we use types instead of having an array `RADIUS_KEYWORDS`? That is a bit awkward and I don't think we do that elsewhere?

mxschll: It's also hard to debug, if I pass an arbitrary string but nothing happens, I can't know why.

style-dictionary/visual-refresh/color-palette.ts
at-susie: Removed

src/box/visual-accent.scss
```diff
@@ -0,0 +1,120 @@
...
+  inline-size: auto;
+}
+
+.visual-accent-aspect-equal {
+  // Equal padding on all sides.
+  padding-block: awsui.$space-xs;
+  padding-inline: awsui.$space-xs;
+  // Force the element to be a square regardless of content dimensions.
+  // The width is constrained to match the height (content + padding).
+  aspect-ratio: 1 / 1;
```
mxschll, line 70: With long text, the element grows very large. How should this feature behave with text, or is not not supposed to be used with text?

Image

src/box/internal.tsx
```diff
@@ -32,11 +40,28 @@ export default function InternalBox({
...
+  // Keyword values are applied via dedicated `.visual-accent-radius-*` classes that map to the
+  // spacing tokens. Any other value is treated as a raw CSS `border-radius` string and passed to
+  // the stylesheet through a custom property, which the base `.visual-accent` rule reads and
+  // applies to all four corners. When unset, no custom property is set and the base rule falls
+  // back to a `0` border radius.
+  const accentStyle =
```
mxschll: Nit: you could remove this const and move this directly to the style prop since it's not used anywhere else. Also, related to the comment above: it's weird to have this isRadiusKeyword check, instead of using types.

src/box/visual-accent.scss
at-susie: Such extreme things can happen in buttons today as well. We will provide guidances

pages/box/visual-accent.page.tsx
at-susie: Adding screenshot area

src/box/internal.tsx
at-susie: Good call. `borderRadius` is now a strict keyword union (xxxs…xxxl plus full), matching padding/margin. Dropped the runtime keyword array and check. Invalid values are caught by the type system now.

src/box/internal.tsx
at-susie: Resolved by the same change. There is no more `accentStyle` const or `isRadiusKeyword` branch. The keyword maps directly to a `.visual-accent-radius-*` class.

mxschll (approved):

PR #4680, opened 2026-07-01
fix: Correctly align side navigation link group items when info slot is present

jperals:
### Description

Issue: AWSUI-62124

### How has this been tested?

Pipeline: AwsUi-v3-jotresse

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/side-navigation/styles.scss
```diff
@@ -187,16 +187,12 @@ $expandable-icon-negative-margin: awsui.$space-l;
...
-  &--info {
+  &--link-with-info {
     display: flex;
     align-items: baseline;
   }
 }
 
-.list-item--info > .list {
-  inline-size: 100%;
-}
```
amanabiy, line 198: I don't see this moved anywhere else do we have this behavior covered?

jperals: I think this was supposed to give the nested list full width so that it would wrap to the next line instead of staying in the same line, but, as we have seen, it didn't work (double-checked by inspecting in the browser: this selector hits the nested list but this element stays in the same flex line).

With the small refactor in this PR, this selector doesn't even hit any element anymore, so I removed it. The same-row issue is now prevented by wrapping the link and the info slot in a wrapper of their own, separate from the nested list.

pages/side-navigation/link-group.page.tsx
```diff
@@ -0,0 +1,63 @@
...
+  const [activeHref, setActiveHref] = useState('#/dashboards/overview');
+
+  const onFollow = useCallback((e: CustomEvent<SideNavigationProps.FollowDetail>) => {
+    setActiveHref(e.detail.href);
+    e.preventDefault();
+  }, []);
+
+  return (
+    <>
+      <h1>Side navigation — Link group</h1>
```
amanabiy: nit: use SimplePage here

jperals: Done

amanabiy (approved):

PR #4692, opened 2026-07-03
fix: Emit funnelStepComplete before content is removed in react 18

georgylobko:
### Description

On React 18, funnelStepComplete came through with `stepName: undefined`. We fire it from an effect cleanup, and the analytics client reads the step's selectors off the DOM at that point, but on React 18 useEffect cleanups run after the DOM is already gone, so nothing matches

Switched the funnel and step unmount effects to useLayoutEffect so cleanup runs while the content is still on the page. Both are moved together to keep the funnel cleanup running before the step cleanup, which the cancel check relies on.

Also re-enabled the "Form submission" funnel integ test on React 18, which was skipped because of this bug.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/analytics/__tests__/components/analytics-funnel.test.tsx
+++ b/src/internal/analytics/__tests__/components/analytics-funnel.test.tsx
@@ -412,6 +412,37 @@ describe('AnalyticsFunnelStep', () => {
     });
   });
 
+  test('resolves the step DOM selectors while emitting funnelStepComplete when the step unmounts', () => {
+    // Regression test for AWSUI-61222. The analytics client re-resolves
+    // `stepNameSelector`/`subStepAllSelector` against the live DOM at the moment
+    // `funnelStepComplete` is emitted. In React 18, a passive `useEffect` cleanup
+    // runs *after* the step content has been detached from the document, so those
+    // selectors would resolve to nothing. Emitting from a layout-effect cleanup
+    // keeps the content attached while the event is dispatched, matching React 16.
+    const stepNumber = 1;
+    const stepNameSelector = '.step-name-selector';
+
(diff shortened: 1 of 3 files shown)


src/internal/analytics/__tests__/components/analytics-funnel.test.tsx
```diff
@@ -412,6 +412,37 @@ describe('AnalyticsFunnelStep', () => {
     });
   });
 
+  test('resolves the step DOM selectors while emitting funnelStepComplete when the step unmounts', () => {
+    // Regression test for AWSUI-61222. The analytics client re-resolves
+    // `stepNameSelector`/`subStepAllSelector` against the live DOM at the moment
```
pan-kot: nit: I would remove the backticks and star formatting from the code comments as rather unnecessary

src/internal/analytics/__tests__/components/analytics-funnel.test.tsx
```diff
@@ -412,6 +412,37 @@ describe('AnalyticsFunnelStep', () => {
     });
   });
 
+  test('resolves the step DOM selectors while emitting funnelStepComplete when the step unmounts', () => {
+    // Regression test for AWSUI-61222. The analytics client re-resolves
```
pan-kot: This test does not secure a regression: it passes also when the change is reverted. Is the existing integ test not enough for this? I guess it exists because of the challenges to test this behaviour accurately in the unit tests env.

georgylobko: Good catch! Removed this test. Yes, existing integ tests that I have uncommented covers this behavior more accurately

pan-kot (approved):

PR #4698, opened 2026-07-06
fix: Use ButtonWrapper for pagination buttons

gethinwebster:
## Description

Pagination's page-button finders returned `ElementWrapper`, so consumers couldn't call `isDisabled()` — the buttons use `aria-disabled` rather than the native `disabled` attribute (see #4576). This adds a pagination-scoped `PaginationButtonWrapper` (exposing `isDisabled()`) and returns it from `findPreviousPageButton`, `findNextPageButton`, `findPageNumberByIndex`, `findCurrentPage`, and `findPageNumbers`. Pagination renders bare `<button>`s rather than the `Button` component, so a dedicated wrapper is used instead of reusing `ButtonWrapper` (`findJumpToPageButton` still returns `ButtonWrapper`, as it wraps an actual `Button`). Non-breaking, since `PaginationButtonWrapper extends ElementWrapper`.

## How has this been tested?

Updated the pagination tests to assert disabled state via `isDisabled()`, and added prev/next focus-retention tests (#4576 switched to `aria-disabled` to keep focus but shipped without a test for it).

### DIFF
--- a/src/__tests__/snapshot-tests/__snapshots__/documenter.test.ts.snap
+++ b/src/__tests__/snapshot-tests/__snapshots__/documenter.test.ts.snap
@@ -37640,12 +37640,7 @@ Returns the current value of the input.",
           "parameters": [],
           "returnType": {
             "isNullable": false,
-            "name": "ElementWrapper",
-            "typeArguments": [
-              {
-                "name": "HTMLElement",
-              },
-            ],
+            "name": "PaginationButtonWrapper",
           },
         },
         {
@@ -37680,12 +37675,7 @@ Returns the current value of the input.",
           "parameters": [],
           "returnType": {
             "isNullable": false,
-            "name": "ElementWrapper",
-            "typeArguments": [
-              {
-                "name": "HTMLElement",
-              },
-            ],
+            "name": "PaginationButtonWrapper",
           },
         },
         {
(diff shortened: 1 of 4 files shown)


avinashbot (commented):
> Note: pagination renders bare <button> elements rather than the Button component, so ButtonWrapper methods that depend on Button-component internals (findLoadingIndicator, findTextRegion, findDisabledReason) are not meaningful for pagination buttons. The commonly needed methods (isDisabled, click, getElement, keyboard/focus helpers) all work.

Doesn't this seem brittle? It seems like a sensible assumption for a Cloudscape dev that component wrapper classes only map to the components exactly, so this might subtly break in the future, especially for methods that we don't test.

This is definitely prettier than my idea though (#4673). Do you think a new "NativeButtonWrapper" that ButtonWrapper could extend from could be possible?

gethinwebster:
> > Note: pagination renders bare  elements rather than the Button component, so ButtonWrapper methods that depend on Button-component internals (findLoadingIndicator, findTextRegion, findDisabledReason) are not meaningful for pagination buttons. The commonly needed methods (isDisabled, click, getElement, keyboard/focus helpers) all work.
> 
> Doesn't this seem brittle? It seems like a sensible assumption for a Cloudscape dev that component wrapper classes only map to the components exactly, so this might subtly break in the future, especially for methods that we don't test.
> 
> This is definitely prettier than my idea though (#4673). Do you think a new "NativeButtonWrapper" that ButtonWrapper could extend from could be possible?

oh interesting, I'd not realized we use plain `<button>` and not even <InternalButton> here. Yes, in that case I agree it's not the best solution, will update it.

avinashbot (approved):
I guess there's a backwards compatibility possibility when moving from ElementWrapper to ComponentWrapper, but realistically, I don't see how anyone could be using those, so it should be fine.

PR #4699, opened 2026-07-06
chore: Expose app layout error-boundary catches for canary detection

gethinwebster:
## Summary

Lets our integration/canary tests detect errors that an App Layout error boundary caught.

## What changed

- **Record + expose caught errors.** `AppLayoutBuiltInErrorBoundary` records each caught error (`{ appLayoutPart, message }`) into a bounded module registry. A new `getAppLayoutErrors()` / `clearAppLayoutErrors()` accessor is exposed on the App Layout root element's `__awsui__` (via `attachAppLayoutErrorBoundaryTestHooks`, wired through the skeleton root ref). No console logging — teams that wrap App Layout in their own `<ErrorBoundary>` keep errors handled quietly, and the signal is a passive DOM query invisible to users.
- **Production positive-control hook.** Adds a test-only `throwInAppLayoutPart(part)` that forces a *genuine* boundary catch, so a canary can verify — against the real minified prod bundle — that detection still works and hasn't been broken by a build change. It fans out to every mounted boundary for the part, so multiple App Layouts on one page do not clobber each other.

## Testing

- Unit tests in `error-boundary.test.tsx`: records + exposes errors, `throwInAppLayoutPart` triggers a real recorded catch, `clearAppLayoutErrors`, `attach` no-ops on null and merges (does not clobber) an existing `__awsui__`, a non-app-layout boundary catch is not recorded, and multi-instance fan-out.
- Integration tests in `widget-areas-error-boundaries.test.tsx` that go through the real skeleton-wired App Layout root element: a genuine boundary catch is readable via `getAppLayoutErrors()`, and `throwInAppLayoutPart` drives a real catch into the registry.
- eslint + prettier clean; quick-build clean.

### DIFF
--- a/package.json
+++ b/package.json
@@ -176,7 +176,7 @@
     {
       "path": "lib/components/internal/plugins/index.js",
       "brotli": false,
-      "limit": "15 kB"
+      "limit": "16 kB"
     },
     {
       "path": "lib/components/internal/widget-exports.js",
--- a/src/app-layout/__tests__/widget-areas-error-boundaries.test.tsx
+++ b/src/app-layout/__tests__/widget-areas-error-boundaries.test.tsx
@@ -66,7 +66,7 @@ beforeEach(() => {
   // masked the exact appLayoutPart values asserted below.
   awsuiPlugins.appLayout.clearRegisteredDrawersForTesting();
   awsuiWidgetInternal.clearInitialMessages();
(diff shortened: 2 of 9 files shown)


gethinwebster:
Superseded by a simpler console-prefix approach in a new PR (dropping the DOM hook + plugin-API controller).

PR #4700, opened 2026-07-06
fix: disable jump-to-page control when pagination is disabled

amanabiy:
### Description
When `disabled` was set, the arrows and page buttons were disabled but the jump-to-page input and go-button stayed interactive, so users could still navigate. Disable both, and mark the root `aria-disabled` so the whole control is conveyed as disabled. Adds a unit test.

Related links, issue #, if available: AWSUI-62134

### How has this been tested?
Using unit test and dev pages locally

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/pagination/internal.tsx
+++ b/src/pagination/internal.tsx
@@ -221,13 +221,14 @@ const InternalPagination = React.forwardRef(
         loadingText={jumpToPageLoadingText}
         ariaLabel={jumpToPage?.loading ? jumpToPageLoadingText : jumpToPageButtonLabel}
         onClick={() => handleJumpToPageClick(Number(jumpToPageValue))}
-        disabled={!jumpToPageValue || Number(jumpToPageValue) === currentPageIndex}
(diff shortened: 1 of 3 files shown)


src/pagination/internal.tsx
```diff
@@ -221,13 +221,14 @@ const InternalPagination = React.forwardRef(
...
         onClick={() => handleJumpToPageClick(Number(jumpToPageValue))}
-        disabled={!jumpToPageValue || Number(jumpToPageValue) === currentPageIndex}
+        disabled={disabled || !jumpToPageValue || Number(jumpToPageValue) === currentPageIndex}
       />
     );
 
     return (
       <ul
         aria-label={paginationLabel}
+        aria-disabled={disabled}
```
pan-kot: what is the purpose of this?

If we need it - please include it to the test, too

amanabiy: I discovered the issue when I was working on the the compact variant. It is a residual from that. I have removed it.

georgylobko (approved):

PR #4706, opened 2026-07-08
chore: Log app layout error-boundary catches

gethinwebster:
## Summary

App layout error boundaries now emit a fixed `[AwsUiAppLayoutError]` prefix to the console on each catch. This lets the integration/canary tests detect real App Layout breakage by matching the prefix, instead of maintaining an allowlist of every other service's console noise. The prefix is a string literal, so it survives minification.

## Testing

Unit tests assert the marker fires per app layout part on a real boundary catch. quick-build, unit suites, and lint all pass.

### DIFF
--- a/src/error-boundary/internal.tsx
+++ b/src/error-boundary/internal.tsx
@@ -110,6 +110,7 @@ export function AppLayoutBuiltInErrorBoundary({
           errorMessage: error?.error?.message ?? '',
           appLayoutPart: appLayoutPart ?? '',
         });
+        console.log(`[AwsUiAppLayoutError] appLayoutPart=${appLayoutPart ?? ''}`, error?.error);
       }}
     >
       <ErrorBoundariesContext.Provider value={{ ...context, suppressed: nextSuppressed, renderFallback }}>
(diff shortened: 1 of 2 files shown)


src/error-boundary/internal.tsx
```diff
@@ -110,6 +110,7 @@ export function AppLayoutBuiltInErrorBoundary({
           errorMessage: error?.error?.message ?? '',
           appLayoutPart: appLayoutPart ?? '',
         });
+        console.log(`[AwsUiAppLayoutError] appLayoutPart=${appLayoutPart ?? ''}`, error?.error);
```
georgylobko, line 113: Why not `console.error`?

gethinwebster: Because the purpose of error boundaries is, at least in part, to allow errors to be caught and handled gracefully. At this point, we don't know whether this error will be caught by a higher-level <ErrorBoundary> and handled in some way there or not, so IMO it's better to be rather cautious and use `log` instead of `error` here to prevent creating `error` messages when the error is actually caught and handled gracefully.

georgylobko (approved):

PR #4713, opened 2026-07-10
feat: add multi-column sort to Table

gethinwebster:
### Description

Adds opt-in multi-column sorting for `Table` through `multiColumnSort`.

Customers can sort by multiple columns using Shift+click or Shift+keyboard activation, manage each column through a sort-options menu, see numbered sort priorities, and clear the active sort. The change also adds screen-reader announcements, localized strings, public test-utils support, generated API documentation, and demo/permutation pages.

The public API is additive and opt-in. Existing single-column sorting behavior is unchanged.

Related links, issue #: n/a

### How has this been tested?

Added new unit tests

<details>

#### Correctness

- [x] Changes include public API and test-utils documentation updates.
- [x] Changes are backward-compatible: `multiColumnSort` is optional and single-column sorting is unchanged.
- [x] Changes do not introduce unsupported browser features.
- [ ] Manual browser accessibility validation was not run locally because ChromeDriver is unavailable; targeted accessibility behavior is covered by unit tests and browser validation requires CI or a ChromeDriver-enabled environment.

#### Security

- [x] The change does not add URL handling.

#### Testing

- [x] Changes are covered by unit tests, including keyboard interaction, comparator columns, screen-reader priority text, sort-menu actions, clearing, warnings, live announcements, and test utilities.
- [ ] Browser integration testing was not run locally because ChromeDriver is unavailable; verify in CI or a ChromeDriver-enabled environment.
</details>

---


pages/table/multi-column-sort.page.tsx
```diff
@@ -0,0 +1,173 @@
...
+const initialSingleColumnSort: TableProps.SortingState<Item> = initialMultiColumnSort[0];
+const PAGE_SIZE = 4;
+
+export default function MultiColumnSortPage() {
+  const { urlParams, setUrlParams } = useAppContext<
+    'multiSort' | 'selection' | 'grouped' | 'pagination' | 'longNames' | 'resizable'
+  >();
+
+  // `multiSort` defaults to on; the rest default to off.
+  const multiSortEnabled = urlParams.multiSort !== 'false' && urlParams.multiSort !== false;
```
jperals: This chunk repeats the same type of conversion 6 times and we even already have it here: https://github.com/cloudscape-design/components/blob/82568c59e92be9870cd256d29c7b875fa54d78ca/pages/app/app-context.tsx#L83-L85

Can we deduplicate this?

pages/table/multi-column-sort.permutations.page.tsx
```diff
@@ -0,0 +1,123 @@
...
+    wrapLines: [false, true],
+  },
+]);
+
+export default function MultiColumnSortPermutationsPage() {
+  return (
+    <PermutationsPage title="Table multi-column sort permutations" i18n={{}}>
+      <PermutationsView
+        permutations={permutations}
+        render={permutation => (
```
jperals: Minor: `render` takes the array index as second argument, so this could be
```
render={(permutation, index) => (
          <Table
            {...permutation}
            ariaLabels={{ ...ariaLabels, tableLabel: `Items ${index}` }}
          />
        )}
```

Also, is this ARIA label very meaningful?

src/button-dropdown/styles.scss
```diff
@@ -65,6 +65,15 @@ $dropdown-trigger-icon-offset: 2px;
   }
 }
 
+// Removes the trigger's vertical padding and borders so an `icon`-variant trigger matches the
+// compact `inline-icon` footprint. The class is doubled to override the button variant styles
+// (`.button.variant-icon`), which button-dropdown CSS is loaded after.
```
jperals: Are we 100% sure that this stylesheet is always loaded after the button styles in all cases (what if for example components are loaded asynchronously), and that selector specificity is not going to break this?

What about delegating this class to the button component? There we can even do something like this to make sure we don't mix both styles:

```
.variant-icon:not(.compact), (...) { ... }
```

src/table/__tests__/multi-column-sort-behavior.test.tsx
```diff
@@ -0,0 +1,353 @@
...
+function renderTable(
+  sortingColumns: Sort,
+  onChange = jest.fn(),
+  extraColumns?: TableProps.ColumnDefinition<Item>[],
+  ariaLabels?: TableProps.AriaLabels<Item>
+) {
+  const { container } = render(
+    <Table
+      items={items}
+      columnDefinitions={extraColumns ?? columnDefinitions}
```
jperals: Nit: "extra" suggests that they will be added rather than replace the default `columnDefinitions`. What about renaming `extraColumns` as `columnDefinitions` and `columnDefinitions` as `defaultColumnDefinitions`?

pages/table/multi-column-sort.page.tsx
gethinwebster: removed, wasn't needed

src/button-dropdown/styles.scss
gethinwebster: good catch, moved to button styling

src/table/styles.scss
```diff
@@ -37,6 +37,13 @@
     margin-inline-start: auto;
   }
 
+  &-clear-sort {
```
jperals, line 40: The slot only ever contains one element (the clear sort button), so I don't think these flex styling is needed

src/table/multi-column-sort/live-announcement.tsx
```diff
@@ -0,0 +1,107 @@
...
+    // Recompute only when the sort changes; the render functions and formatter are behaviourally stable.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [sortingColumns]);
+
+  if (!announcement) {
+    return null;
+  }
+  return (
+    <InternalLiveRegion hidden={true} tagName="span">
+      <span>{announcement}</span>
```
jperals: Nit: is the span needed?

src/table/header-cell/styles.scss
```diff
@@ -168,28 +185,46 @@ $cell-horizontal-padding: awsui.$space-scaled-l;
...
 
   @include header-cell-focus-outline(awsui.$space-table-header-focus-outline-gutter);
 
-  .header-cell-disabled.header-cell-sorted > & {
+  .header-cell-sortable > .header-cell-main > & {
+    padding-inline-end: calc(#{awsui.$space-xl} + #{awsui.$space-xxs});
+  }
+
+  // The multi-sort priority badge renders alongside the sort arrow inside the
+  // absolutely-positioned sort icon. Reserve extra inline-end space for it so it
```
jperals, line 195: I'm OK with not including this in this scope but it would be nice if the sort icon (and the newly added priority badge) is not rendered with absolute position

src/table/header-cell/styles.scss
```diff
@@ -168,28 +185,46 @@ $cell-horizontal-padding: awsui.$space-scaled-l;
...
     & > .sorting-icon {
       color: awsui.$color-text-interactive-disabled;
     }
   }
 }
 
+.header-cell-main-with-menu {
+  display: flex;
+  align-items: center;
+  // stylelint-disable-next-line no-descending-specificity
```
jperals: Can we avoid overriding our own linting rules, e.g, by adding classnames to the inner element (`.header-cell-content` in this case) instead of relying on the DOM structure? I think this is what we do elsewhere

src/table/header-cell/styles.scss
```diff
@@ -150,6 +150,23 @@ $cell-horizontal-padding: awsui.$space-scaled-l;
   color: awsui.$color-text-column-sorting-icon;
 }
 
+.header-cell-main {
+  /* When the wrapper is present without a sort menu, behave as if it weren't there. */
```
jperals, line 154: How does this selector alone that there is no sort menu? It is added unconditionally: https://github.com/cloudscape-design/components/pull/4713/changes#diff-f948a1d80d749dbe50dd7f1835c64a59e6e5c18ca7b8e380a2a559b65cbb4479R246

src/table/header-cell/styles.scss
```diff
@@ -150,6 +150,23 @@ $cell-horizontal-padding: awsui.$space-scaled-l;
...
+  display: contents;
+}
+
+.sort-priority-badge {
```
jperals, line 158: Flex-related styles are not needed if there is only one child. Same for the ruleset below

src/table/header-cell/sort-menu.tsx
```diff
@@ -0,0 +1,103 @@
...
+            'i18nStrings.sortDropdownRemoveFromSortDisabledReason',
+            i18nStrings?.sortDropdownRemoveFromSortDisabledReason
+          ),
+        },
+      ],
+    },
+  ];
+
+  return (
+    <span className={styles['sort-menu']} onClick={e => e.stopPropagation()}>
```
jperals: Can the stopPropagation be handled by the button dropdown instead? Is this wrapper necessary?

jperals:
Can we test the feature alongside `enableKeyboardNavigation`?

src/table/__tests__/multi-column-sort-behavior.test.tsx
gethinwebster: renamed

src/table/header-cell/styles.scss
gethinwebster: updated

src/table/styles.scss
gethinwebster: removing

src/table/multi-column-sort/live-announcement.tsx
gethinwebster: removing

src/table/header-cell/styles.scss
gethinwebster: fixing

src/table/header-cell/styles.scss
gethinwebster: removing

src/table/header-cell/sort-menu.tsx
gethinwebster: removing wrapper

gethinwebster (commented):
also added keyboard nav toggle and tests for keyboard nav behaviour

src/table/__integ__/multi-column-sort-keyboard.test.ts
```diff
@@ -0,0 +1,161 @@
...
+import createWrapper from '../../../lib/components/test-utils/selectors';
+
+const tableWrapper = createWrapper().findTable();
+
+// Columns on the multi-column-sort dev page, in order: 1=Name, 2=Type, 3=State, 4=CPU %.
+const sortControl = (colIndex: number) => tableWrapper.findColumnSortingArea(colIndex).toSelector();
+// The per-column sort menu renders a single <button> trigger in the header cell (the sort control
+// itself is a role="button" div, not a <button>), so this uniquely targets the sort menu trigger.
+const sortMenuTrigger = (colIndex: number) =>
+  tableWrapper.find(`th[data-column-index="${colIndex}"] button`).toSelector();
```
jperals: Can't we use `findColumnSortMenu` here?

src/table/header-cell/sort-menu.tsx
jperals: There is no stopProgagation now. Was it not needed?

gethinwebster: button dropdown already does it internally

src/table/__integ__/multi-column-sort-keyboard.test.ts
gethinwebster: ideally, yes, but using it uncovered a bug in existing table test utils (related to this line https://github.com/cloudscape-design/components/blob/90c8d77dd14cf0cef3d5aa38919aa9e8b96470f2/src/test-utils/dom/table/index.ts#L78, where the usage of comma-separated selectors breaks any subsequent chaining in integ tests.

So, I suggest leaving this as-is for this PR, and separately investigating how we can fix this more generally.

jperals (dismissed):

src/table/__integ__/multi-column-sort-keyboard.test.ts
gethinwebster: actually, decided to fix it, new commit incoming

jperals:
There seems to be an unnecessary padding at the inline-start of the first column header: https://d21d5uik3ws71m.cloudfront.net/components/d1229c7385ad48926501db7e341ebdfc4ad060f2/allure-report/index.html#1fc1a847e3c9474fdc6859a1a3957176

src/table/__integ__/multi-column-sort-keyboard.test.ts
```diff
@@ -0,0 +1,162 @@
...
+    await expect(page.isFocused(sortControl(1))).resolves.toBe(false);
+  })
+);
+
+test(
+  'left/right arrows traverse each column sort control and its sort menu across the header row',
+  setupTest(async page => {
+    await page.click('h1');
+    await page.keys(TABS_INTO_GRID);
+    console.log(sortControl(1));
```
jperals: Leftover console.log

jperals (approved):

PR #4714, opened 2026-07-10
test: Add visual regression tests for app layout

jperals:
### Description

Migrate all app layout visual regression tests to the new test definitions.

Related ticket: `D417180663`

### How has this been tested?

Manually verified that all visual regression tests executed in the internal pipeline were also run in this PR's checks (see [Allure report](https://d21d5uik3ws71m.cloudfront.net/components/b5219ec8bce53bf17b17c3c613cc77b05e7445c9/allure-report/index.html)).

Also manually verified the screenshots for some of them.

<details>

#### Correctness

#### Security

#### Testing

</details>

---


test/definitions/visual/app-layout-drawers.ts
```diff
@@ -0,0 +1,69 @@
...
+    {
+      description: 'with split panel',
+      path: 'app-layout/with-drawers',
+      screenshotType: 'viewport',
+      setup: async ({ page, wrapper }) => {
+        await page.click(wrapper.findAppLayout().findDrawerTriggerById('pro-help').toSelector());
+      },
+    },
+    {
+      description: 'with tooltip on hover',
```
jperals, line 36: Note: this test description in the internal package is called "with tooltip on hover in visual refresh", but the test is run for all systems and themes without distinction, including Classic , so I am dropping the "in visual refresh" part (which is not the special case anymore anyway)

test/definitions/visual/app-layout-content-paddings.ts
```diff
@@ -0,0 +1,29 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import { TestSuite } from '../types';
+
+const suite: TestSuite = {
+  description: 'Content paddings',
+  componentName: 'app-layout',
+  tests: [
+    ...(['true', 'false'] as const).flatMap(toolsEnabled =>
```
gethinwebster, line 9: why `flatMap` here and `map` 2 lines later? both are already flat arrays

jperals: `['true', 'false']` is a flat array, but the function passed to `flatMap` returns an array instead of a single item, that is why it is needed (same for the `flatMap` in the next line).

`flatMap` is about flattening the result of the passed-in function, not the array on which it is called.

There is an alternative without using `flatMap`, something like this:

```
tests: [
  ...(['true', 'false'] as const).map(toolsEnabled =>
    description: `toolsEnabled=${toolsEnabled}`,
    tests: (['true', 'false'] as const).map(splitPanelEnabled =>
        description: `splitPanelEnabled=${splitPanelEnabled}`,
        tests: (['bottom', 'side'] as const).map(splitPanelPosition => ({
          description: `splitPanelPosition=${splitPanelPosition}`,
          path: 'app-layout/with-split-panel',
          screenshotType: 'viewport' as const,
          queryParams: { toolsEnabled, splitPanelEnabled, splitPanelPosition },
        }))
      )
    ),
 ...
```

By nesting test definitions under `tests`, the [runner](https://github.com/cloudscape-design/components/blob/fcf95ed76c7260846f3bb09da02414f4f8f649dc/test/definitions/utils.ts#L65-L141) will nest tests in `describe` blocks. Actually slightly more readable probably, but the `flatMap` approach just mimics the result of the internal tests.

gethinwebster (approved):

PR #4716, opened 2026-07-11
fix: Prevent HelpPanel content from clipping at drawer bottom edge

TrevorBurnham:
### Description

I noticed that when I scrolled to the bottom of a `HelpPanel`, there was no whitespace below the content.

The cause: `HelpPanel` reserved its bottom whitespace as `margin-block-end` on its last child. A descendant's block-end margin is not added to an ancestor scroll container's scrollable region, so when the panel renders inside a fixed-height, scrollable app-layout drawer, that trailing margin is clipped and content sits flush against the bottom edge.

This fix reserves the space as `padding-block-end` on the panel root instead. An element's own padding *is* included in the scroll extent, so the whitespace renders consistently regardless of the wrapping container.

### How has this been tested?

- Existing `HelpPanel` unit tests pass (6/6); app-layout unit suite passes (1563 tests, 38 suites, 10 snapshots).
- Manually verified in the dev server (`app-layout/default`).

#### Before

Screenshot 2026-07-11 at 5 10 12 PM

#### After

Screenshot 2026-07-11 at 5 09 06 PM

### DIFF
--- a/src/help-panel/styles.scss
+++ b/src/help-panel/styles.scss
@@ -10,7 +10,12 @@
   @include styles.styles-reset;
   word-wrap: break-word;
   padding-block-start: awsui.$space-panel-header-vertical;
-  padding-block-end: 0;
+  // Reserve the bottom whitespace as padding on the panel root rather than as a
+  // margin on the last child. A descendant's block-end margin is not added to an
+  // ancestor scroll container's scrollable region, so when the panel is rendered
+  // inside a fixed-height, scrollable app-layout drawer the trailing margin gets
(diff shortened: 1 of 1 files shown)


gethinwebster:
Here is another possible approach, that potentially solves this more generally for drawers. We need to regression test both approaches against various different scenarios (including checking our visual regression test suite for coverage, as scrolling in app layout is not generally tested thoroughly) https://github.com/cloudscape-design/components/compare/main...dev-v3-gethinw-drawer-padding

georgylobko (approved):

PR #4718, opened 2026-07-14
chore: Stable hashes for border tokens

pan-kot:
### Description

A follow-up for https://github.com/cloudscape-design/theming-core/pull/190

### How has this been tested?

* Existing screenshot tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/build-tools/utils/token-versions.js
+++ b/build-tools/utils/token-versions.js
@@ -1,0 +1,27 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+const DEFAULT_TOKEN_VERSION = 'v3-1';
+
+// Groups map a token-name pattern (matched against the token's CSS variable name) to a version.
+// Tokens matching no group stay version-less and keep the legacy value-based hashes.
+const versionGroups = [{ pattern: /^border-/, version: DEFAULT_TOKEN_VERSION }];
+
+// Builds the token -> version allowlist from the full token -> cssName map.
+function getTokenVersions(variablesMap, groups = versionGroups) {
+  const counters = {};
+  const tokenVersions = {};
+  for (const [token, cssName] of Object.entries(variablesMap)) {
+    const group = groups.find(({ pattern }) => pattern.test(cssName));
+    if (group) {
+      tokenVersions[token] = group.version;
+      counters[group.version] = (counters[group.version] ?? 0) + 1;
+    } else {
+      counters.unset = (counters.unset ?? 0) + 1;
+    }
+  }
+  console.log(`Design token versions: ${JSON.stringify(counters)}`);
+  return tokenVersions;
+}
+
+module.exports = { versionGroups, getTokenVersions };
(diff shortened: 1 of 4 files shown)


build-tools/utils/token-versions.js
```diff
@@ -0,0 +1,27 @@
...
+  for (const [token, cssName] of Object.entries(variablesMap)) {
+    const group = groups.find(({ pattern }) => pattern.test(cssName));
+    if (group) {
+      tokenVersions[token] = group.version;
+      counters[group.version] = (counters[group.version] ?? 0) + 1;
+    } else {
+      counters.unset = (counters.unset ?? 0) + 1;
+    }
+  }
+  console.log(`Design token versions: ${JSON.stringify(counters)}`);
```
amanabiy, line 23: Why do we need this log? I am thinking we could make it separate function. because the it seems lke getTokenVersions is doing two things (logging the count of token version, and returning the token versions).

build-tools/utils/token-versions.js
```diff
@@ -0,0 +1,27 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+const DEFAULT_TOKEN_VERSION = 'v3-1';
+
+// Groups map a token-name pattern (matched against the token's CSS variable name) to a version.
+// Tokens matching no group stay version-less and keep the legacy value-based hashes.
+const versionGroups = [{ pattern: /^border-/, version: DEFAULT_TOKEN_VERSION }];
+
+// Builds the token -> version allowlist from the full token -> cssName map.
+function getTokenVersions(variablesMap, groups = versionGroups) {
```
amanabiy, line 11: Consider separating this function into three functions: getTokenVersions, findVersion, and logVersionCount (if this is needed, if not let's remove the counter, and log). That way we can also test the findVersion separately.

build-tools/utils/token-versions.js
```diff
@@ -0,0 +1,27 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+const DEFAULT_TOKEN_VERSION = 'v3-1';
+
+// Groups map a token-name pattern (matched against the token's CSS variable name) to a version.
+// Tokens matching no group stay version-less and keep the legacy value-based hashes.
+const versionGroups = [{ pattern: /^border-/, version: DEFAULT_TOKEN_VERSION }];
```
amanabiy, line 8: From the comments and implementation it's not clear that the order in our version group matters, for the getTokenVersions (which is a very important thing). Can we make this more robust? so that our tests fail when the order changes? or we don't need the method getTokenVersions to be dependent on the order of the versionGroups.

Do we also need a way to detect a conflicting groups?

pan-kot: There is such test already: https://github.com/cloudscape-design/components/pull/4718/changes#diff-d98db6d8daa40af7ea255a703f2e95003ae2814d6254b622d136d719f523ecb7R23

build-tools/utils/token-versions.js
pan-kot: Why to test findVersion separately? The entire function and the pattern we use for border tokens are already covered with tests. The logging does not need to be tested - that is just for debugging.

build-tools/utils/token-versions.js
pan-kot: Logging is just the debugging - it helps to understand what portion of the tokens is already covered with versions, that's it.

amanabiy (approved):

PR #4724, opened 2026-07-15
feat: Steps connector lines display

pan-kot:
### Description

The PR introduces new API for Steps component:

```
interface StepsProps {
  // ...
  
  /**
   * Determines whether the connector lines between steps are displayed. The following variants are available:
   * * `visible` - Shows connector lines between consecutive steps.
   * * `none` - Hides the connector lines between steps.
   */
  connectorLines?: "visible" | "none";
}

<Steps 
  steps={[...]}
  connectorLines="none"
/>
```

This change is synonymous with the existing Tree view API that declares `connectorLines: "vertical" | "none"`. While there are only two options for now, the union allows us to add more connector lines variants in the future.

Rel: AWSUI-62150

### How has this been tested?

* New unit tests
* New permutations

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/__tests__/snapshot-tests/__snapshots__/documenter.test.ts.snap
```diff
@@ -27782,6 +27776,25 @@ Don't use \`ariaLabel\` and \`ariaLabelledby\` at the same time.",
       "optional": true,
       "type": "string",
     },
+    {
+      "defaultValue": "'default'",
```
cansuaa: Why do we call it "default"? As a customer I wouldn't understand what default is without trying it out

pan-kot: Fair point!

That is the part that is different from the tree-view. How about "visible"?

cansuaa: Better than default! I guess we cannot use "vertical" like tree-view because of different orientation options in the component

pan-kot: The lines are shown or hidden in both horizontal and vertical orientation of steps - that's why it is not "vertical" as it is in the tree view. I will update to "visible" then!

src/steps/internal.tsx
```diff
@@ -89,12 +95,12 @@ const InternalStep = ({
...
               <InternalStatusIcon type={status} iconAriaLabel={statusIconAriaLabel} />
             </InternalBox>
-            <hr className={styles.connector} role="none" />
+            <hr className={connectorClassName} role="none" />
           </>
         )}
       </div>
       {orientation === 'vertical' ? (
-        <hr className={styles.connector} role="none" />
+        <hr className={connectorClassName} role="none" />
```
amanabiy, line 103: Since the background becomes transparent rather than being removed from the DOM, should we rename the API values to "normal" and "hidden"? That would better describe the behavior.

pan-kot: I chose "none" to be aligned with the tree-view component which has the same property. The "normal" has the same flaw as "default" (original proposal) - it is not indicative enough - I will update to `"visible" | "none"`.

pan-kot: For the consumer, the fact that we make lines transparent instead of removing them from the dom should not matter - it is an implementation detail. The reason we hide them like that is to preserve original spacing between items: removing the lines makes items closer in the vertical layout.

src/__tests__/snapshot-tests/__snapshots__/documenter.test.ts.snap
pan-kot: Done!

src/steps/internal.tsx
amanabiy: That looks good to me.

cansuaa (approved):

amanabiy (approved):

PR #4736, opened 2026-07-16
feat: add copy to clipboard as valid item type for button group

shannenigans:
### Description

  Add `icon-copy-to-clipboard` as a valid item type in the ButtonGroup component. This
  allows users to embed a copy-to-clipboard action directly within a button group
  alongside existing item types (icon-button, icon-toggle-button, icon-file-input,
  menu-dropdown).

  Changes:
  - Added `IconCopyToClipboard` interface with properties: `id`, `text`, `textToCopy`,
  `copySuccessText`, `copyErrorText`, `disabled`, and `disabledReason`
  - Added `onCopySuccess` and `onCopyFailure` event handlers to `ButtonGroupProps`
  - Created `CopyToClipboardItem` component that wraps `InternalCopyToClipboard` with
  tooltip support
  - Updated `ItemRuntime`, `Item`, and `InternalItem` union types to include the new
  item type
  - Wired up copy event handlers through `internal.tsx` and `item-element.tsx`

  ### How has this been tested?

  - Verified TypeScript compilation with the new type additions
  - Manually tested rendering a button group with an `icon-copy-to-clipboard` item
  - Confirmed copy success/failure callbacks fire correctly
  - Confirmed tooltip behavior matches other button group item types

  <details>

  #### Correctness

  - _Changes are backward-compatible if not indicated, see [`CONTRIBUTING.md`](https://
  github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#public-apis)._
  - _Changes do not include unsupported browser features, see
  [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIB
  UTING.md#browsers-support)._
  - _Changes were manually tested for accessibility, see [accessibility
  guidelines](https://cloudscape.design/foundation/core-principles/accessibility/)._

  #### Security

  - _If the code handles URLs: all URLs are validated through [the `checkSafeUrl`
  function](https://github.com/cloudscape-design/components/blob/main/src/internal/util
  s/check-safe-url.ts)._

  #### Testing

  </details>


src/button-group/interfaces.ts
```diff
@@ -105,6 +106,14 @@ export interface ButtonGroupProps extends BaseComponentProps {
    * Called when the user uploads files. The event detail object contains the id and files from the file input item.
    */
   onFilesChange?: NonCancelableEventHandler<ButtonGroupProps.FilesChangeDetails>;
+  /**
+   * Called when the text is successfully copied to the clipboard.
+   */
+  onCopySuccess?: NonCancelableEventHandler<CopyToClipboardProps.CopySuccessDetail>;
```
pan-kot: I think these should use custom `ButtonGroupProps.CopySuccessDetail` and `ButtonGroupProps.CopyFailureDetail` - that is because in button group these also need to have the `id` of the target item, as there can be multiple copy buttons.

src/button-group/index.tsx
```diff
@@ -59,7 +59,14 @@ const ButtonGroup = React.forwardRef(
 );
 
 function getItemCounts(allItems: readonly ButtonGroupProps.ItemOrGroup[] = []) {
-  const counters = { 'icon-button': 0, 'icon-toggle-button': 0, 'icon-file-input': 0, 'menu-dropdown': 0, group: 0 };
+  const counters = {
+    'icon-button': 0,
+    'icon-toggle-button': 0,
+    'icon-file-input': 0,
+    'icon-copy-to-clipboard': 0,
```
pan-kot, line 67: The counts are passed to component's metadata. Please include a new field for the copy to clipboard.

src/button-group/__tests__/button-group-copy-to-clipboard.test.tsx
```diff
@@ -0,0 +1,121 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
pan-kot, line 1: Please also add the new button type to `/button-group/permutations` test page. It is used for visual testing (by comparing screenshots), and can also be used to test the changes manually.

pan-kot: Actually, the page already includes copy buttons - implemented as regular icon-buttons. You can update them to use the new variant instead.

src/button-group/item-element.tsx
```diff
@@ -120,6 +126,16 @@ const ItemElement = forwardRef(
       setTooltip(null);
     };
 
+    const onCopyFailureHandler: NonCancelableEventHandler<CopyToClipboardProps.CopyFailureDetail> = event => {
+      fireCancelableEvent(onCopyFailure, event.detail);
+      setTooltip(null);
```
pan-kot: This does not really work. The tooltip is hidden initially, but then it re-appears if the item is hovered.

Image

In button group we use a customised tooltip implementation that transitions from tooltip to popover state if needed - that was used precisely to implement copy-to-clipboard items (which required a bit of extra code to actually do copy to clipboard).

I think reusing the copy to clipboard component here does not really work - it is best to use a custom implementation but reuse relevant utilities from the copy to clipboard component instead.

src/button-group/item-element.tsx
pan-kot, line 62: We need the same for copy to clipboard, too. This forwards programmatic focus to the correct underlying element.

src/button-group/__tests__/button-group-copy-to-clipboard.test.tsx
```diff
@@ -0,0 +1,121 @@
...
+  text: 'Copy code',
+  textToCopy: 'Text to copy',
+  copySuccessText: 'Copied',
+  copyErrorText: 'Failed to copy',
+};
+
+function renderButtonGroup(props: Partial<ButtonGroupProps>) {
+  const { container } = render(<ButtonGroup variant="icon" ariaLabel="Chat actions" items={[copyItem]} {...props} />);
+  const wrapper = createWrapper(container).findButtonGroup()!;
+  const copyToClipboard = createWrapper(container).findCopyToClipboard()!;
```
pan-kot: This is wrong: the button dropdown wrapper must receive its `findCopyToClipboardById()` helper instead.

src/button-group/item-element.tsx
shannenigans: I see. Just want to clarify that the expected behavior is that the success/failure popover and the tooltip should render as one component as opposed to being rendered on top of the other here, which can't be accomplished with simply reusing the copy to clipboard component?

src/button-group/__tests__/button-group-copy-to-clipboard.test.tsx
shannenigans: is this due to the possibility that there can be multiple copy to clipboard buttons?

edit: since I'm updating this to a cusotm implementation using iconbutton, updating this to `findButtonById()`

src/button-group/item-element.tsx
pan-kot: The tooltip does not have to remain the same dom element when transitioning from tooltip to popover state, but we should not display both at the same time.

src/button-group/__tests__/button-group-copy-to-clipboard.test.tsx
pan-kot: Yes, but also to ensure consistency when it comes to different item types. Even if there is only one copy button - the consumers should use the button group's utils for that - as whether we implement it by reusing the copy-to-clipboard component, or use a custom solution - is an implementation detail.

src/button-group/__tests__/button-group-copy-to-clipboard.test.tsx
```diff
@@ -0,0 +1,126 @@
...
+      expect(onCopyFailure).toHaveBeenCalledTimes(1);
+      expect(onCopyFailure).toHaveBeenCalledWith(
+        expect.objectContaining({ detail: { id: 'copy', text: 'Text to copy with error' } })
+      );
+      expect(wrapper.findTooltip()!.getElement()).toHaveTextContent('Failed to copy');
+    });
+  });
+
+  test('falls back to the error feedback when the clipboard API is unavailable', async () => {
+    Object.assign(global.navigator, { clipboard: undefined });
```
pan-kot, line 97: This should have a cleanup

src/button-group/__tests__/button-group-copy-to-clipboard.test.tsx
```diff
@@ -0,0 +1,126 @@
...
+  text: 'Copy code',
+  textToCopy: 'Text to copy',
+  copySuccessText: 'Copied',
+  copyErrorText: 'Failed to copy',
+};
+
+function renderButtonGroup(props: Partial<ButtonGroupProps>) {
+  const { container } = render(<ButtonGroup variant="icon" ariaLabel="Chat actions" items={[copyItem]} {...props} />);
+  const wrapper = createWrapper(container).findButtonGroup()!;
+  const copyToClipboard = wrapper.findButtonById('copy')!;
```
pan-kot: We need to introduce `findCopyToClipboardById()` util. This is needed for consistency but also it gives consumers the ability to validate item's type this way.

src/button-group/__tests__/button-group-copy-to-clipboard.test.tsx
shannenigans: is this not cleaned up by the `afterEach`? and then reset by the `beforeEach`?

pan-kot: Oh, I see now - apologies!

pan-kot (dismissed):

pan-kot (approved):

PR #4745, opened 2026-07-20
chore: Refactor visual accent colors to use tokens

at-susie:
### Description
This PR refactors the visual accent styling for the Box and Badge components.

#### Changes
- Removed some unnecessary color options
- Updated visual accent colors and badge component's style

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


style-dictionary/one-theme/colors.ts
```diff
@@ -130,10 +131,10 @@ const tokens: StyleDictionary.ColorsDictionary = {
   colorBackgroundProgressBarValueDefault: { light: '{colorPrimary600}', dark: '{colorPrimary500}' },
 
   // ── Badge ─────────────────────────────────────────────────────────
-  colorBackgroundBadgeGreen: { light: '{colorSuccess100}', dark: 'rgba(43, 181, 52, 0.2)' },
-  colorBackgroundBadgeBlue: { light: '{colorInfo100}', dark: 'rgba(92, 127, 255, 0.2)' },
-  colorBackgroundBadgeRed: { light: '{colorError100}', dark: 'rgba(255, 61, 61, 0.2)' },
-  colorBackgroundBadgeYellow: { light: '{colorWarning100}', dark: 'rgba(251, 211, 50, 0.2)' },
+  colorBackgroundBadgeGreen: { light: '{colorSuccess100}', dark: hexToRgba(paletteTokens.colorGreen500!, 0.2) },
```
georgylobko: Why do we need to convert hex to rgba? We already have some tokens defined in hex, so why should these be different?

at-susie: In the visual-accent feature, we are introducing a quite a bit of semi-transparent colors that are all associated with primitive color palette like `colorGreen500` with different opacity values. If we keep maintaining these only with the `rgba` it's error prone and hard to maintain. In think in the long term, other rgba values can also use the `hexToRgba` function for better maintainability.

georgylobko: >  it's error prone and hard to maintain

why?

at-susie: We design those styles using primitive color palette like `colorGreen* ` while applying opacities. As soon as we convert them to `rgba` values, it becomes hard to understand which primitive color is associated with. 
Here is an example:
- In design, we define styles like `colorIndigo500` with 20% opacity
- If we convert it to `rgba` it's `rgba(35, 42, 68, 0.2)` - It's almost impossible to remember that is Indigo500

Maintaining such `rgba` values for many styles in the code base is not good idea in general.

georgylobko: > It's almost impossible to remember that is Indigo500

I didn't get it

> Maintaining such rgba values for many styles in the code base is not good idea in general.

But that’s exactly what you’ve done in this function - you’ve converted hex values to rgba.

at-susie: In other words, `rgba` is a hard coded value which we should not have, in the same reason why we should not style using hex code. The function is a layer that that converts primitive values to raw values.

georgylobko (dismissed):

georgylobko (approved):

PR #4750, opened 2026-07-21
chore: Adds package json exports for test-utils dom wrappers

pan-kot:
### Description

Make individual test-utils dom wrappers exported via package.json exports. These are necessary for extending test utils when building component libraries on top of Cloudscape.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/build-tools/tasks/package-json.js
+++ b/build-tools/tasks/package-json.js
@@ -51,6 +51,11 @@ function getComponentsExports() {
     result[`./${component}`] = `./${component}/index.js`;
   }
 
+  // Per-component test-utils DOM wrappers (e.g. `.../test-utils/dom/button`).
+  for (const component of listPublicItems('src/test-utils/dom')) {
+    result[`./test-utils/dom/${component}`] = `./test-utils/dom/${component}/index.js`;
+  }
+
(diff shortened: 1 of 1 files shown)


build-tools/tasks/package-json.js
```diff
@@ -51,6 +51,11 @@ function getComponentsExports() {
     result[`./${component}`] = `./${component}/index.js`;
   }
 
+  // Per-component test-utils DOM wrappers (e.g. `.../test-utils/dom/button`).
```
jperals, line 54: Can't the per-component test utils already be imported from the root? Like
```
import { Component } from "@cloudscape-design/components/test-utils/dom";
```

pan-kot: Ah, good point - I did not realise that we export them from the index file! I will cancel the PR and update my docs :)

pan-kot: Wait, no.

The wrappers are indeed exported - but the whole idea is to avoid importing from `/dom/index` - doing so will augment the ElementWrapper.

I reopened the PR :)

jperals (approved):

PR #4755, opened 2026-07-21
feat: Input prefix suffix

NathanZlion:
### Description

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/input/adornments.page.tsx
+++ b/pages/input/adornments.page.tsx
@@ -1,0 +1,148 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import React, { useContext } from 'react';
+
+import Box from '~components/box';
+import FormField from '~components/form-field';
+import Icon from '~components/icon';
+import Input, { InputProps } from '~components/input';
+import Select, { SelectProps } from '~components/select';
+import SpaceBetween from '~components/space-between';
+import Toggle from '~components/toggle';
+
+import AppContext, { AppContextType } from '../app/app-context';
+import { SimplePage } from '../app/templates';
+
+type AdornmentMode = 'none' | 'short' | 'long' | 'icon';
+
+type PageContext = React.Context<
+  AppContextType<{
+    type?: InputProps.Type;
+    prefix?: AdornmentMode;
+    suffix?: AdornmentMode;
+    disabled?: boolean;
+    readOnly?: boolean;
+    invalid?: boolean;
+    warning?: boolean;
+    value?: string;
+  }>
+>;
+
+const typeOptions: ReadonlyArray<SelectProps.Option> = [
+  { value: 'text', label: 'Text' },
+  { value: 'password', label: 'Password' },
+  { value: 'number', label: 'Number' },
+  { value: 'email', label: 'Email' },
+  { value: 'url', label: 'URL' },
+  { value: 'search', label: 'Search' },
+];
+
+const adornmentOptions: ReadonlyArray<SelectProps.Option> = [
+  { value: 'none', label: 'None' },
+  { value: 'short', label: 'Short text' },
+  { value: 'long', label: 'Long text' },
+  { value: 'icon', label: 'Icon' },
+];
+
+function getPrefix(mode: AdornmentMode): React.ReactNode {
+  switch (mode) {
+    case 'short':
+      return '$';
+    case 'long':
+      return 'https://example.com/very-long-prefix-that-overflows';
+    case 'icon':
+      return <Icon name="search" />;
+    default:
+      return undefined;
+  }
+}
+
+function getSuffix(mode: AdornmentMode): React.ReactNode {
+  switch (mode) {
+    case 'short':
+      return '%';
+    case 'long':
+      return '.ec2.internal.very-long-domain-that-overflows';
+    case 'icon':
+      return <Icon name="settings" />;
+    default:
+      return undefined;
+  }
+}
+
+export default function AdornmentsPage() {
+  const { urlParams, setUrlParams } = useContext(AppContext as PageContext);
+  const type = urlParams.type ?? 'text';
+  const prefixMode = urlParams.prefix ?? 'short';
+  const suffixMode = urlParams.suffix ?? 'short';
+  const disabled = urlParams.disabled ?? false;
+  const readOnly = urlParams.readOnly ?? false;
+  const invalid = urlParams.invalid ?? false;
+  const warning = urlParams.warning ?? false;
+
+  return (
+    <SimplePage
+      title="Input prefix and suffix"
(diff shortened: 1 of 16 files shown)


pages/input/adornments.page.tsx
```diff
@@ -0,0 +1,768 @@
...
+interface StateControlsProps {
+  disabled: boolean;
+  setDisabled: (v: boolean) => void;
+  readOnly: boolean;
+  setReadOnly: (v: boolean) => void;
+  invalid: boolean;
+  setInvalid: (v: boolean) => void;
+  warning: boolean;
+  setWarning: (v: boolean) => void;
+  compact: boolean;
```
jperals: Compact density is already in the top nav controls (and RTL should ideally be as well at some point)

pages/input/adornments.page.tsx
```diff
@@ -0,0 +1,768 @@
...
+          </ColumnLayout>
+        </div>
+      </Box>
+    </Container>
+  );
+}
+
+// ─── Main page ────────────────────────────────────────────────────────────────
+
+export default function AdornmentsPage() {
```
jperals: For visual regression testing, a screenshot area will be needed.

My suggestion would be a separate permutations page, so that only the component to be tested is included in the screenshots and no other extraneous content which lead to unnecessary collateral diffs.

src/input/internal.tsx
```diff
@@ -120,11 +125,29 @@ function InternalInput(
   __rightIcon = __rightIcon ?? searchProps.__rightIcon;
   __onRightIconClick = __onRightIconClick ?? searchProps.__onRightIconClick;
 
+  // Search inputs use built-in search and clear icons that would overlap adornments.
```
jperals, line 126: Sounds like a good case for a dev warning?

src/input/internal.tsx
```diff
@@ -120,11 +125,29 @@ function InternalInput(
...
+
   const formFieldContext = useFormFieldContext(rest);
   const { ariaLabelledby, ariaDescribedby, controlId, invalid, warning } = __inheritFormFieldProps
     ? formFieldContext
     : rest;
 
+  const hasAdornmentContent = (content: React.ReactNode) =>
+    content !== null && content !== undefined && typeof content !== 'boolean' && content !== '';
+  const hasPrefix = hasAdornmentContent(prefix);
+  const hasSuffix = hasAdornmentContent(suffix);
```
jperals: Non blocking: Could this be just
```
const hasPrefix = !!prefix;
const hasSuffix = !!suffix;
```

I think the only possible difference would arise if someone sets `prefix={true}` or `suffix={true}`, but I would not expect that?

src/input/internal.tsx
```diff
@@ -120,11 +125,29 @@ function InternalInput(
...
+  const hasAdornment = hasPrefix || hasSuffix;
```
jperals: I find "adornment" not very clear, although I won't block on this because I can't think of perfect alternatives.

I think I would even prefer `hasPrefixOrSuffix`. Alternatively `hasAffix` ("affix" includes prefix and suffix) though not a common term.

src/input/internal.tsx
```diff
@@ -211,10 +235,50 @@ function InternalInput(
...
+        )}
+      >
+        {mainInput}
+      </div>
+    </div>
+  ) : (
+    mainInput
+  );
+
+  const rightIcon = __rightIcon ? (
```
jperals: Non-blocking because it doesn't technically break anything, but this terminology is not RTL-friendly and it is easy to name these elements like `suffixIcon`, and even clearer IMO.

Same with class names e.g, `input-icon-suffix`.

src/test-utils/dom/input/index.ts
```diff
@@ -12,4 +12,18 @@ export default class InputWrapper extends BaseInputWrapper {
   findClearButton(): ElementWrapper | null {
     return this.find(`.${inputSelectors['input-button-right']}`);
   }
+
+  /**
+   * Returns the prefix adornment element, or null if no prefix is set.
```
jperals: Again I wouldn't call this blocking but an adornment is purely decorational while this is not necessarily the case since it has a functional purpose in helping the user understanding how the input value will be used etc.

src/input/styles.scss
```diff
@@ -224,3 +239,146 @@
...
+    ) {
+      padding-inline-start: 0;
+    }
+
+    &:focus-within {
+      box-shadow: var(#{custom-props.$styleBoxShadowFocus}, foundation.$box-shadow-focused-light-invalid);
+    }
+  }
+
+  &.input-adorned-container-disabled {
```
jperals, line 309: For some reason now the height of an input with prefix or suffix jumps from 20px to 22px when disabled or readOnly.

pages/input/prefix-suffix.page.tsx
gethinwebster, line 1: suggest renaming prefix-suffix or similar to match feature naming

src/input/internal.tsx
gethinwebster: not sure why this was resolved? I think it's valid: the current code is overly complicated (and different to how we treat similar properties in other components)

src/test-utils/dom/input/index.ts
```diff
@@ -10,6 +10,14 @@ export default class InputWrapper extends BaseInputWrapper {
   static rootSelector: string = inputSelectors.root;
 
   findClearButton(): ElementWrapper | null {
-    return this.find(`.${inputSelectors['input-button-right']}`);
+    return this.find(`.${inputSelectors['input-button-end']}`);
```
gethinwebster: Ideally we should try not to change class names on existing test utils, as it causes backwards-incompatibility. As it's just a naming thing here, I suggest changing this one classname back.

src/input/styles.scss
```diff
@@ -224,3 +239,184 @@
...
+.input-suffix {
+  display: flex;
+  align-items: center;
+  flex-shrink: 1;
+  max-inline-size: max(0px, calc((100% - 12rem) / 2));
+  min-inline-size: 0;
+  padding-block: styles.$control-padding-vertical;
+  padding-inline: styles.$control-padding-horizontal;
+  color: inherit;
+  user-select: text;
```
gethinwebster: do we need to specify these, I think they're the default anyway?

src/test-utils/dom/input/index.ts
NathanZlion: Odd edge-case to the !!hasPrefix is for a number 0 for instance. But I guess it's not a real usecase? it's falsy so doesn't render.

gethinwebster: I think it's unlikely, and if it's needed it can be passed as string to make it truthy

gethinwebster (approved):

PR #4772, opened 2026-07-23
feat: add automatic skeleton rows to Table

gethinwebster:
Adds automatic skeleton-row sizing to Table via `skeleton={{ totalRows: 'auto' }}`: during loading the table fills its available viewport with placeholder rows, and a post-render correction removes rows if the rendered page would overflow (never below one row).

**Changes**
- Automatic skeleton rows in `use-auto-skeleton-rows.ts`, wired through `Table` internals and interfaces.
- Unit + browser integration coverage: sticky header, footer, `wrapLines` on/off with long headers, and mixed data + skeleton rows (data rows always precede skeleton rows).
- AppLayout console-shell dev pages (`#/light/app-layout/auto-skeleton-table`, `...-nowrap`) with an external `#h` header and `#f` footer, matching a real console layout.

**Testing**
- `npm run build`, `npm run lint`, and the table unit suites (17/17) pass.
- Browser integration tests run in CI.
- Row-count oscillation risk was investigated with a real headless-Chrome probe (forced overflow, container churn, seed-height mismatch) — no oscillation and no ResizeObserver-loop warnings.


src/internal/utils/scrollable-containers.ts
```diff
@@ -117,6 +119,22 @@ export function scrollRectangleIntoView(box: BoundingBox, scrollableParent?: HTM
   }
 }
 
+/**
+ * Returns the element's ancestors whose computed `overflow-y` is `auto` or `scroll`,
+ * ordered nearest-first. These are the ancestors capable of producing a vertical scrollbar.
+ */
+export function getScrollableParents(element: HTMLElement): HTMLElement[] {
```
NathanZlion, line 126: Do we have a test for this?

src/table/interfaces.tsx
```diff
@@ -706,9 +708,7 @@ export namespace TableProps {
     item: T;
   }
 
-  export interface SkeletonConfig {
-    totalRows: number;
-  }
+  export type SkeletonConfig = { totalRows: number; maxAutoRows?: never } | { totalRows: 'auto'; maxAutoRows?: number };
```
NathanZlion: can we do union types that are named in the namespace for these instead of inline one, it'll make it more readable I think? Just like we do for `ColumnDisplayProperties = ColumnDisplay | GroupDisplay;` Maybe something like AutoConfig and FixedConfig or AutoSkeletonConfig,

src/table/use-auto-skeleton-rows.ts
```diff
@@ -0,0 +1,164 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import React, { useCallback, useLayoutEffect, useRef, useState } from 'react';
+
+import { getScrollableParents } from '../internal/utils/scrollable-containers';
+
+const AUTO_SKELETON_VIEWPORT_BUFFER = 16;
```
NathanZlion, line 7: nit: why this number?

src/internal/utils/scrollable-containers.ts
gethinwebster: implicitly, but now added an explicit test too

src/table/interfaces.tsx
gethinwebster: good point, updated

src/table/use-auto-skeleton-rows.ts
gethinwebster: In my testing it needed a certain amount of buffer away from the bottom to prevent often over-shooting with measurements because things shift around the page a little during loading. So no specific reason for this number rather than another close number, but _some_ buffer is needed

NathanZlion (dismissed):

NathanZlion (approved):

PR #4831, opened 2026-07-24
feat: gap side nav items when icons are present

mxschll:
### Description

Changes spacing between nav items in side nav when icons are present.

Related links, issue #, if available: n/a

### How has this been tested?

Locally + Pipeline. 

<details>
   <summary>Review checklist</summary>

_The following items are to be evaluated by the author(s) and the reviewer(s)._

#### Correctness

- _Changes include appropriate documentation updates._
- _Changes are backward-compatible if not indicated, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#public-apis)._
- _Changes do not include unsupported browser features, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#browsers-support)._
- _Changes were manually tested for accessibility, see [accessibility guidelines](https://cloudscape.design/foundation/core-principles/accessibility/)._

#### Security

- _If the code handles URLs: all URLs are validated through [the `checkSafeUrl` function](https://github.com/cloudscape-design/components/blob/main/src/internal/utils/check-safe-url.ts)._

#### Testing

- _Changes are covered with new/existing unit tests?_
- _Changes are covered with new/existing integration tests?_
</details>

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #4843, opened 2026-07-29
feat: Add expandable-icon-end to collapsible side navigation

jkuelz:
### Description

Adjusts the layout of the collapsible SideNavigation variant based on whether icons are present. When any nav item has an icon, the layout adapts: item spacing widens (separate PR #4831), the expand caret moves to the inline-end, and rows get a full-width hover background. Caret moves to end of the row to not clutter the left side when there are icons present, and hover is added to connect the caret and the text. Added an internal `__expandIconPosition` to the ExpandableSection component which is then used in SideNav.

Also included:
  
  - **Fix:** ExpandableSection's `disableContentPaddings`: Now removes content padding for all variants, not just container (other variants also have default content padding); API description corrected accordingly.
  - New design token: `colorBackgroundSideNavigationItemIconHover` -- it is a known issue that the hover token's dark 1T value isn't fully reference-token-themeable (won't track a primary-color change) — proper fix needs an alpha palette derived from primary, which is a larger change.
  - One-theme section headers: removed the section-header-specific typography overrides so section headers inherit the footer default (16px/500), to distinguish them from active expandable-link-groups.
  
### How has this been tested?

Dev pipeline (dev-v3-jkuelz)

<details>

#### Correctness

#### Security

#### Testing

</details>

---


mxschll:
The collapsed-state behavior feels a bit inconsistent. When a parent has an icon, its children are not shown in the collapsed state:

| Open    | Collapsed |
| -------- | ------- |
| image  | image    |

When the parent has no icon, the collapsed state shows children instead, but only the children that have icons:

| Open    | Collapsed |
| -------- | ------- |
|  image |  image  |

Is this the intended behavior?

My concern is that this could lead to confusing UX: two sections can behave differently in the collapsed state depending on whether the parent has an icon. In one case, the collapsed nav shows the parent; in another, it shows a subset of the children.

Would it be more intuitive to consistently show only parent items in the collapsed state? We could add a dev warning when a parent has no icon set or document in the guidelines that all parents should have icons.

pages/side-navigation/collapsed.page.tsx
mxschll, line 109: Will there be a slot for the trigger in the side nav header as a follow up?

style-dictionary/one-theme/colors.ts
```diff
@@ -192,6 +192,10 @@ const tokens: StyleDictionary.ColorsDictionary = {
   colorTextSideNavigationItemDefault: { light: '{colorTextBodyDefault}', dark: '{colorTextBodyDefault}' },
   colorBackgroundSideNavigationItemActive: { light: 'transparent', dark: 'transparent' },
   colorBackgroundSideNavigationItemActiveCollapsed: { light: '{colorPrimary500}', dark: '{colorPrimary500}' },
+  colorBackgroundSideNavigationItemIconHover: {
```
mxschll, line 195: Does this only apply to the expanded state? Should we have a token for the collapsed hover colors too?

src/side-navigation/styles.scss
```diff
@@ -206,31 +219,16 @@ $expandable-icon-negative-margin: awsui.$space-l;
...
-    display: grid;
-    align-items: flex-start;
-    padding-block: $item-padding-block;
+  &--no-ident {
+    margin-inline-start: 0;
   }
 
-  &--no-ident {
+  &--expand-icon-end {
     margin-inline-start: 0;
```
mxschll:
```suggestion
  &--no-ident,
  &--expand-icon-end {
    margin-inline-start: 0;
  }
```

src/expandable-section/internal.tsx
```diff
@@ -131,7 +152,12 @@ export default function InternalExpandableSection({
         <div
           id={controlId}
           ref={ref}
-          className={clsx(styles.content, styles[`content-${baseVariant}`], expanded && styles['content-expanded'])}
+          className={clsx(
+            styles.content,
+            styles[`content-${baseVariant}`],
+            expanded && styles['content-expanded'],
+            disableContentPaddings && styles['disable-content-paddings']
```
mxschll, line 159: `disableContentPaddings` existed before. Is this a breaking change?

src/expandable-section/__tests__/expandable-section.test.tsx
```diff
@@ -509,3 +512,289 @@ describe('headerText', () => {
...
+    test('icon renders after header text for default variant', () => {
+      const wrapper = renderInternalExpandableSection({ variant: 'default', __expandIconPosition: 'end' });
+      const expandButton = wrapper.findExpandButton().getElement();
+      const children = Array.from(expandButton.children);
+      const iconIndex = children.findIndex(el => el.classList.contains(styles['icon-container']));
+      const textIndex = children.findIndex(el => el.classList.contains(styles['header-text']));
+      expect(iconIndex).toBeGreaterThan(textIndex);
+    });
+
+    test('icon renders after header content for navigation variant', () => {
```
mxschll, line 539: The tests are a quite verbose, e.g. `icon renders after header content for navigation variant` and `applies header-icon-end modifier class for navigation variant` are almost the same. Can we simplify?

src/expandable-section/__tests__/expandable-section.test.tsx
```diff
@@ -509,3 +512,289 @@ describe('headerText', () => {
...
+        </InternalExpandableSection>
+      );
+      const wrapper = createWrapper(container).findExpandableSection()!;
+      const iconButton = wrapper.findExpandIcon().getElement();
+      // Native button handles Enter/Space natively, triggering click
+      iconButton.click();
+      expect(onChange).toHaveBeenCalledTimes(1);
+    });
+  });
+});
```
mxschll, line 828: Should we add a test that ensures the tab order is correct for link groups? When `_expandIconPosition="start"` it should be `previous item -> caret -> link -> next item` and when `_expandIconPosition="end"` it should be `previous item -> link -> caret -> next item`.

src/expandable-section/__tests__/expandable-section.test.tsx
```diff
@@ -509,3 +512,289 @@ describe('headerText', () => {
...
+    test('default variant toggles aria-expanded on click', () => {
+      const wrapper = renderInternalExpandableSection({ variant: 'default', __expandIconPosition: 'end' });
+      const expandButton = wrapper.findExpandButton().getElement();
+      expect(expandButton).toHaveAttribute('aria-expanded', 'false');
+      wrapper.findExpandButton().click();
+      expect(expandButton).toHaveAttribute('aria-expanded', 'true');
+    });
+
+    test('container variant toggles aria-expanded on click', () => {
+      const wrapper = renderInternalExpandableSection({ variant: 'container', __expandIconPosition: 'end' });
```
mxschll: There's a bug when opening / closing the section with Enter / Space, it opens and closes immediately. Clicking with the mouse works fine, because propagation is prevented for click events, not Enter / Space. We should fix that and have a test for it. 

The focus behaves differently when `__expandIconPosition: 'end'`. With `start`, the whole header gets the focus. With `end`, you can tab onto the title, then on the trigger itself, but the trigger does not show a focus ring.

style-dictionary/one-theme/typography.ts
```diff
@@ -31,9 +31,6 @@ const tokens: StyleDictionary.TypographyDictionary = {
   lineHeightHeadingXs: '20px',
   fontWeightHeadingXs: '500',
 
-  fontSideNavigationSectionHeaderSize: '14px',
```
mxschll, line 34: Why are we removing this? Can we set the values required for one-theme? Otherwise it'll fall back to vr again.

src/side-navigation/styles.scss
```diff
@@ -264,7 +263,6 @@ $expandable-icon-negative-margin: awsui.$space-l;
 // Section groups
 // ==========================================================================
 .section-group {
-  @include styles.font-heading-m;
```
mxschll, line 267: Does this change how section groups look? This affects all themes.

src/expandable-section/__tests__/expandable-section.test.tsx
mxschll: We should also add this to the `/expandable-section/container-variant.permutations` and add integration tests for it.

pages/side-navigation/collapsed.page.tsx
jkuelz: That is a separate item, since it requires public API. This PR is just adapting the internal styles.

style-dictionary/one-theme/colors.ts
jkuelz: The hover state is the same for both expanded and collapsed states.

mxschll: Non-blocking: If there's a use case for builders to change this, we could add another token and default it to be the same in both states. Can be a follow up though.

src/expandable-section/internal.tsx
jkuelz: `DisableContentPaddings` will still exist. The only part that is changing is that now `disableContentPaddings` applies to variants besides container. I view this more as a bug fix with a clear resolution rather than a breaking change. It might've only applied to the container variant because we didn't have use cases for the others, or the other variants didn't come with content padding (the previous description said, "default padding is only present for the `container` variant."). Which might've been true at the time, but is not true anymore.

style-dictionary/one-theme/typography.ts
jkuelz: The desired is the same as VR (16px/500), which is why we are removing it.

src/side-navigation/styles.scss
jkuelz: It was redundant because we already use Box h3 variant in the parts.tsx file which would apply these styles by default.

style-dictionary/one-theme/typography.ts
mxschll: Can we defined it here explicitly still? We'll have to untangle VR with One Theme at one point and keeping things separate makes it easier for us in the future.

jkuelz: I can, but when we did VR, we only defined the items that were changed from the base theme. Which helped define the line between them, otherwise you're constantly going between the two to find a diff.

mxschll: Non-blocking, but we should strive for making One Theme independent. Dependency on VR will cause issues should VR be deprecated in the distant future.

mxschll (dismissed):

mxschll (approved):

PR #4854, opened 2026-08-02
feat: Add public useCustomI18n hook

radsacha:
## Summary

Add a public namespace-aware `useI18n(namespace, component)` hook from `@cloudscape-design/components/i18n`.

This lets component libraries built on top of Cloudscape resolve their own translated strings through the existing `I18nProvider` message context, without exposing `InternalI18nContext` or changing Cloudscape's internal `useInternalI18n` hook.

## Changes

- Add `useI18n(namespace, component)` as a thin public wrapper around the existing formatter.
- Export `useI18n` and `I18nFormatFunction` from `@cloudscape-design/components/i18n`.
- Add unit coverage for custom namespaces, provided-value precedence, no-provider fallback, locale widening, custom handlers, multiple namespaces, nested providers, and shared Cloudscape/third-party provider usage.

## Non-goals

- No `I18nProvider` behavior changes.
- No message format or bundle shape changes.
- No changes to `useInternalI18n` or Cloudscape component typing.

## Testing

- `npm run quick-build`
- `node_modules/.bin/eslint src/i18n/context.ts src/i18n/index.ts src/i18n/__tests__/use-i18n.test.tsx`
- `TZ=UTC node_modules/.bin/jest -c jest.unit.config.js src/i18n/__tests__/use-i18n.test.tsx src/i18n/__tests__/i18n.test.tsx`
- `TZ=UTC node_modules/.bin/jest -c jest.unit.config.js src/__tests__/functional-tests/public-exports.test.ts`
- `git diff --check`


src/i18n/context.ts
```diff
@@ -61,6 +61,21 @@ export interface ComponentFormatFunction<ComponentName extends StringKeyOf<I18nF
...
+  key: string,
+  provided: string | undefined,
+  customHandler?: CustomHandler<string | undefined, Record<string, string | number>>
+) => string | undefined;
+
+/**
+ * Public hook for third-party component libraries to resolve translations
+ * through I18nProvider under their own namespace.
+ */
+export function useI18n(namespace: string, component: string): I18nFormatFunction {
```
avinashbot: It looks like the only difference between the public-facing hook and internal hook is the type-safety, which would be a shame to lose when making the API public. Internally, we auto-generate an argument mapping alongside the strings, so this comes for "free" for us, but types can still be maintained by builders making custom components and namespaces.

`useInternalI18n` essentially hardcodes a dependency on `I18nFormatArgTypes`, but we can make this a generic; something like (I didn't test this out, but it looks reasonable):

```ts
export const namespace = 'cloudscape-design-components';

export interface ComponentFormatFunction<
  NamespaceTypes,
  ComponentName extends StringKeyOf<NamespaceTypes>,
> {
  // ...
}

/** The publicly exposed custom i18n function */
export function useCustomI18n<
  NamespaceTypes,
  ComponentName extends StringKeyOf<NamespaceTypes>
>(
  namespace: string,
  componentName: ComponentName
): ComponentFormatFunction<NamespaceTypes, ComponentName> {
  // ...
}

/** The internal convenience version hardcoded for Cloudscape's namespace and types */
export function useInternalI18n<ComponentName extends StringKeyOf<I18nFormatArgTypes>>(
  componentName: ComponentName
): ComponentFormatFunction<I18nFormatArgTypes, ComponentName> {
  return useCustomI18n<I18nFormatArgTypes, ComponentName>(namespace, componentName);
}
```

This way, there's no "internal" and "external" version of the hook, the internal is just the external one with a frequently duplicated argument pre-provided.

src/i18n/__tests__/use-i18n.test.tsx
```diff
@@ -0,0 +1,180 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
avinashbot: These mostly seem to test the same things the existing unit tests test for, except with a different string value? If you can mostly remove these without Codecov yelling at you, I would be fine with it.

But the specific tests concerning interactions between multiple namespaces, either in a single I18nProvider or in nested providers, are worth keeping.

src/i18n/context.ts
avinashbot: (To discuss) `useCustomI18n` might be a better name; we still want to indicate the hook is meant for custom components, not to replace application-wide internationalization support.

Will discuss with the team tomorrow about the naming.

avinashbot: Discussed this with the team; would prefer `useCustomI18n`.

src/i18n/__tests__/use-i18n.test.tsx
radsacha: Updated in the latest revision. I removed the duplicate coverage for existing provider behavior and kept the namespace-specific cases.

src/i18n/context.ts
radsacha: Great suggestion, thanks. Updated the latest revision to follow this shape: the public hook is now `useCustomI18n<NamespaceTypes, ComponentName>()`, and `useInternalI18n()` is the Cloudscape one with `I18nFormatArgTypes` and `cloudscape-design-components` pre-bound.

radsacha: Sounds good, thanks for confirming with the team. Updated the latest revision to rename the public hook to `useCustomI18n`

radsacha: I tested it out, screenshots 
Screenshot 2026-08-05 at 6 37 27 AM

radsacha (commented):
Thanks for the review. I pushed a new revision addressing the feedback:
- Renamed the public hook to `useCustomI18n`
- Made the public hook generic so custom namespaces can keep the same type-safety model as Cloudscape internals
- Updated `useInternalI18n` to delegate to `useCustomI18n` with Cloudscape’s namespace and generated `I18nFormatArgTypes`
- Trimmed the test file to keep only namespace-specific coverage: multiple namespaces in one provider, nested providers, and Cloudscape + custom namespace sharing one provider

radsacha (commented):
.

radsacha (commented):
.

src/i18n/context.ts
```diff
@@ -41,35 +41,64 @@ export function useLocale(): string | null {
...
+  const formatFunction = <MessageKey extends StringKeyOf<NamespaceTypes[ComponentName]>, ValueType>(
     key: MessageKey,
     provided: ValueType,
-    customHandler?: CustomHandler<ValueType, I18nFormatArgTypes[ComponentName][MessageKey]>
-  ) => {
-    return format(namespace, componentName, key, provided, customHandler);
-  };
+    customHandler?: CustomHandler<ValueType, NamespaceTypes[ComponentName][MessageKey]>
+  ) => format(namespace, componentName, key, provided, customHandler);
+  return formatFunction as I18nFormatFunction<NamespaceTypes, ComponentName>;
```
avinashbot: It looks like this cast is unnecesary (it typechecks well without it locally). It might mask any potential type issues in the future, could you remove it?

```suggestion
  return formatFunction;
```

avinashbot (changes requested):

radsacha (commented):
remove cast

avinashbot (approved):

PR #4856, opened 2026-08-03
feat: add dropdown customization slots to Select and Multiselect

gethinwebster:
### Description

Adds render-prop slots to `Select` and `Multiselect` for customizing the dropdown, plus two props for consumers to own the accessibility semantics of that content:

- `renderDropdownHeader` — content pinned at the top of the dropdown, above the filter input and options list.
- `renderDropdownFooter` — content pinned at the bottom, always sticky (never scrolls), additive to the built-in status region.
- `dropdownRole?: 'auto' | 'dialog'` — lets a consumer opt the dropdown into `dialog` semantics for interactive custom content, independent of whether filtering is enabled.
- `dropdownAriaDescribedby?: string` — adds `aria-describedby` to the dropdown's listbox/dialog content.

Each render prop receives `DropdownContentProps { filterText, closeDropdown }`. The props live on the shared `BaseSelectProps`, so `Select` and `Multiselect` inherit them; a small `composeDropdownContent` helper wraps the custom header/footer around the filter and the built-in status region. The two short header/footer wrapper styles are declared in each component's stylesheet (the header is lifted with `z-index` since it can sit directly above the options list; the footer needs no lift).

Backward compatible: all new props are optional. When omitted, the composed output is unchanged from today's behavior (verified — additive-only documenter/selectors snapshots, no removals).

### How has this been tested?

- **Unit tests:** new `dropdown-customization` suites for Select and Multiselect (composition/order, `closeDropdown` closes + returns focus to the trigger exactly once, `filteringType='none'`, `dropdownRole`, `dropdownAriaDescribedby`, `expandToViewport`). Full Select/Multiselect/dropdown scope passes (503 tests).
- **Integration tests (keyboard reachability):** new integ tests confirm on the real browser harness that an interactive **footer** is reachable via a forward `Tab` and an interactive **header** via a backward `Shift+Tab`, with the dropdown staying open, in both `dialog` (`filteringType='auto'`) and plain `listbox` (`filteringType='none'`) modes — so interactive header/footer content is keyboard-reachable without requiring `dropdownRole='dialog'`.
- **Type / build / snapshots:** `tsc --noEmit` clean; `npm run build` succeeds; documenter and `test-utils-selectors` snapshots regenerated (additive only).
- **Manual:** dev pages under `pages/select` and `pages/multiselect` exercise the header/footer slots, including the no-filter header case, forced `dropdownRole="dialog"` with `dropdownAriaDescribedby`, and `expandToViewport`.

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/select/utils/dropdown-customization.tsx
```diff
@@ -0,0 +1,54 @@
...
+export function composeDropdownContent({
+  filter,
+  customDropdownHeader,
+  customDropdownFooter,
+  statusFooter,
+  styles,
+}: ComposeDropdownContentProps): { header: React.ReactNode; footer: React.ReactNode } {
+  // A render prop that is not provided yields undefined; one that opts out can return null. Normalize both
+  // to null so presence checks below use strict equality.
+  const headerNode = customDropdownHeader ?? null;
```
avinashbot, line 28: For what it's worth, the render prop function can also return `undefined`, which would be similarly normalized into `null`. Don't _think_ it's a problem here, but worth pointing out.

src/select/interfaces.ts
```diff
@@ -149,6 +149,34 @@ export interface BaseSelectProps
...
+  renderDropdownFooter?: SelectProps.DropdownContentRenderer;
+
+  /**
+   * Determines the ARIA role of the dropdown.
+   *
+   * - `auto` (default): the dropdown uses a listbox role, or a dialog role when built-in filtering is enabled.
+   * - `dialog`: the dropdown always uses a dialog role, even without built-in filtering. Use this when you render
+   * interactive custom content and want assistive technologies to treat the dropdown as a dialog.
+   */
+  dropdownRole?: SelectProps.DropdownRole;
```
avinashbot, line 175: Just to consider: Technically, the options are always wrapped in a listbox, so the two options are either "listbox" or "dialog + listbox". The doc here just says it's either "listbox"/"dialog" or just "dialog".

Do you think this needs to be clarified? Or is that just too much detail for here?

src/select/utils/dropdown-customization.tsx
```diff
@@ -0,0 +1,54 @@
...
+// SPDX-License-Identifier: Apache-2.0
+import React from 'react';
+
+interface ComposeDropdownContentProps {
+  // The filter node (already includes any custom filtering actions, composed by the Filter component).
+  filter: React.ReactNode;
+  customDropdownHeader: React.ReactNode;
+  customDropdownFooter: React.ReactNode;
+  statusFooter: React.ReactNode;
+  styles: Record<string, string>;
```
avinashbot: Interesting pattern, not sure how I feel about this 😅 

Because it's untyped, I feel like there's a bigger risk of future refactors and find-and-replace changes breaking this. Would prefer just `dropdownHeaderClass` and `dropdownFooterClass` over this pattern.

avinashbot (dismissed):
Left some optional comments to address, but none of them are big enough to block a merge.

src/select/utils/dropdown-customization.tsx
gethinwebster: good point, updated

src/select/interfaces.ts
gethinwebster: adjusting wording

src/select/utils/dropdown-customization.tsx
gethinwebster: removed the comment, as I think that was the main thing that was unclear here

avinashbot (approved):

PR #4867, opened 2026-08-05
feat: Animate side navigation collapse and expandable section reveal

jkuelz:
### Description

Animates `SideNavigation`'s collapsed icon rail transition and `ExpandableSection`'s content reveal, scoped to 1T only as per the animation strategy and to prevent breaking changes.

- Replaces `ExpandableSection`'s `display: none`/`block` content toggle with an animated `grid-template-rows` reveal. The content stays mounted and inert while collapsed, to allow for smooth transition states that weren't possible with the previous `display: none`.
- Refactors SideNavigation's to the same `grid-template-rows` strategy as expandable sections to enable transitions: text fade, spatial collapse/expand, and staggered enter/exit timing for links, sections, section-groups, link-groups, and expandable-link-groups.
 - Adds a one-theme-with-motion mixin (internal/styles/motion/mixins.scss) alongside the existing with-motion, so animation-only CSS additions can be scoped to one-theme without affecting other themes' timing.
- Gives one-theme its own values, reusing the existing token pair rather than introducing new ones. These may change with the upcoming animations strategy.

Other behavior changes:
- Added warnings when a collapsed-rail item (a root-level link, link-group, or expandable-link-group, or a link/link-group/ELG nested inside a root-level section/section-group) has no icon and would therefore be invisible or have no visual indicator in the collapsed rail.
- Expandable-link-group children are now forced closed when the nav is collapsed, regardless of the ELG's own expanded/collapsed state or icon presence. Previously an ELG's would either collapse their children or show their children depending on whether the parent had an icon.
- Sections are now forced open when the nav is collapsed, so their icon-bearing children always flatten into the rail — this existed conceptually before but is now explicit and paired with the ELG behavior above.
- Sections, link-groups, and ELGs use `inert` (not just visual hiding) to remove their headers/parent links from the tab order and assistive technology when they collapse to zero size in the rail.
- Reverted an earlier change that added dividers between sections/groups in the collapsed state and returned to using space instead.
- New `ariaLabel` plumbing and `aria-hidden` on section/section-group header text for collapsed-rail accessibility.

Why animations only in 1T? 
The initial version of this animation applied to all themes and broke the pre-existing contract for ExpandableSection that collapsed/hidden content is visually removed instantly (display: none) — which means any downstream tests that asserted isDisplayed()/isExisting() immediately after a collapse toggle with no wait, could fail. Scoping the animation to one-theme (opt-in) avoids that regression for existing VR/classic consumers while still shipping the new motion for one-theme.

### How has this been tested?

dev pipeline

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/expandable-section/internal.tsx
```diff
@@ -143,29 +176,45 @@ export default function InternalExpandableSection({
           headerActions={headerActions}
           headingTagOverride={headingTagOverride}
           expandIconPosition={__expandIconPosition}
+          hideExpandIcon={__hideExpandIcon}
           {...triggerProps}
         />
       }
       __internalRootRef={__internalRootRef}
     >
-      <CSSTransition in={expanded} timeout={30} classNames={{ enter: styles['content-enter'] }} nodeRef={ref}>
```
jkuelz, line 151: Couldn't use `CSSTransition` for reveal animations because `onExit` fires one render after `expanded` becomes false... so when `!expanded=true` AND `!exitInProgress=true`, the formula yields true momentarily causing a flash when padding was removed.

src/expandable-section/expandable-section-container.tsx
```diff
@@ -59,7 +57,7 @@ export const ExpandableSectionContainer = ({
           className={className}
           header={header}
           variant={variant === 'stacked' ? 'stacked' : 'default'}
-          disableContentPaddings={disableContentPaddings || !expanded}
+          disableContentPaddings={true}
```
jkuelz, line 60: Moved padding to `ExpandableSection`’s `.content-inner-body-container`, which sits inside the clip and collapses naturally with the grid track. Previously, container's content padding lived outside `ExpandableSection`'s grid transition, so toggling it based on expanded state (as before) caused a visible 1-frame padding jump instead of animating smoothly.

src/expandable-section/internal.tsx
```diff
@@ -59,9 +63,13 @@ export default function InternalExpandableSection({
   __injectAnalyticsComponentMetadata,
   __expandIconPosition = 'start',
   __disableHeaderPaddings = false,
+  __hideExpandIcon = false,
   ...props
 }: InternalExpandableSectionProps) {
-  const ref = useRef<HTMLDivElement>(null);
+  const contentInnerRef = useRef<HTMLDivElement>(null);
+  // Starts settled when initially expanded or under reduced motion, since neither case
```
mxschll: > or under reduced motion
Where is that part handled?

src/side-navigation/styles.scss
```diff
@@ -134,26 +171,92 @@ $expandable-icon-negative-margin: awsui.$space-l;
...
+    grid-template-rows: 0fr;
+
+    @include styles.one-theme-with-motion {
+      // EXIT: grid row collapses to zero height.
+      transition: grid-template-rows $link-exit-spatial-duration $link-exit-easing $link-exit-spatial-delay;
+    }
+  }
+}
+
+.link-group-children-inner {
```
mxschll, line 229: If the transition is gated, why do we need to add this rule outside of the one theme rule? Will this cause a regression for VR, e.g. focus ring clipping?

style-dictionary/one-theme/motion.ts
```diff
@@ -0,0 +1,23 @@
...
+import merge from 'lodash/merge.js';
+
+import { expandMotionDictionary } from '../utils/index.js';
+import { StyleDictionary } from '../utils/interfaces.js';
+import { tokens as parentTokens } from '../visual-refresh/motion.js';
+
+const tokens: StyleDictionary.MotionDictionary = {
+  motionDurationFast: { default: '110ms', disabled: '0ms' },
+  motionDurationModerate: { default: '150ms', disabled: '0ms' },
+  motionDurationSlow: { default: '200ms', disabled: '0ms' },
```
mxschll: This values are different from VR. Changing these tokens here will change animations for all components that use these tokens in One Theme. Is this intentional?

src/expandable-section/internal.tsx
jkuelz: Good catch. I think I moved it to css at one point but then didn't add back here. Fixed. The collapse-effect now settles immediately when expanding under reduced motion (previously `contentSettled` would never become true in that case, since `handleContentTransitionEnd` only fires on a real `transitionend`, which never happens when `grid-template-rows` doesn't transition).

style-dictionary/one-theme/motion.ts
jkuelz: Yeas this was intentional, and seen as a two way door. I needed to increase the durations to make the animations more impactful than what we have defined in VR, which are generally very quick and snappy. I decided to change them here so that all animations still work together as a system/unit (scaled them up equally ~20ms each). I did a brief gut check on a few components, but I'll admit I didn't go and manually test every instance of motion one by one. I figured these will likely be changing again with our animation story anyway, so it felt like a low risk change to make. and changes this small are usually not very noticeable to the eye.

src/side-navigation/styles.scss
jkuelz: The focus ring clipping was mostly an issue on expandable section because of how it does its padding/margin, but you're right to flag this here too. These can't be gated to one-theme because we no longer rely on conditional rendering that used to fully unmount `LinkGroup`'s children in every theme. So VR now needs this CSS unconditionally to visually hide the always-mounted content when collapsed.

That said, `overflow-y: clip` was permanent, in every theme and every motion state, with nothing ever switching it back to visible once expanded. That meant any active/hover/focus state on the first/last child that visually extends past its own box was clipped, permanently, any time the group was expanded. Fixed by adding the same settled-state pattern expandable-section uses.

style-dictionary/one-theme/motion.ts
mxschll: Can we decouple the navigation motion from these? We'll have more motion in place. Maybe we should think of motion tokens the same way we do for component specific color tokens. Wdyt?

pages/side-navigation/collapsed.page.tsx
```diff
@@ -117,8 +448,31 @@ export default function SideNavigationCollapsedPage() {
             />
           </div>
           <SideNavigation
+            // header={{
+            //   href: '#/',
```
mxschll: Do we need to keep this block?

src/expandable-section/internal.tsx
mxschll: I don't fully understand. Is this an issue with CSSTransition?

src/expandable-section/motion.scss
```diff
@@ -1,30 +0,0 @@
-/*
```
mxschll, line 1: Why did we move this to [src/expandable-section/styles.scss](https://github.com/cloudscape-design/components/pull/4867/changes#diff-c7307a390be1a6409df97a39fc555975acfe85d72ed74250954f08781cc5b4b7)? It would be good to have a separation if possible to make it more maintainable in the future.

src/expandable-section/styles.scss
```diff
@@ -267,23 +291,82 @@ $icon-total-space-medium: calc(#{$icon-width-medium} + #{$icon-margin-left} + #{
...
+    display: block;
+  }
+
+  &-compact {
+    padding-inline-start: $icon-total-space-normal;
+  }
+
+  // One-theme only: animated grid-template-rows reveal instead of the true/false display:none. VR keeps
+  // instantaneous display toggling, to prevent breaking reliance on synchronous isDisplayed()/isExisting() checks.
+  @include theming.one-theme-only {
```
mxschll, line 306: If our linter rules allow it, can we wrap all one theme specific styles in one big `@include theming.one-theme-only`? Makes it easier for us to keep track and maintain these.

src/expandable-section/internal.tsx
```diff
@@ -103,6 +112,36 @@ export default function InternalExpandableSection({
...
+    if (!expanded) {
+      setContentSettled(false);
+    } else if (reducedMotion) {
+      setContentSettled(true);
+    }
+  }, [expanded, reducedMotion]);
+
+  const handleContentTransitionEnd = useCallback(
+    (event: React.TransitionEvent<HTMLDivElement>) => {
+      if (event.propertyName === 'grid-template-rows' && expanded) {
```
mxschll: I _think_ since we can have nested expandable sections (not sure if we should have, that's a different topic), this condition fires for children too.

src/side-navigation/parts.tsx
```diff
@@ -701,6 +723,40 @@ function LinkGroup({
...
+  const reducedMotion = useReducedMotion(childrenRef);
+  // Starts settled when not collapsed, since that case never transitions
+  // grid-template-rows — the only trigger for handleChildrenTransitionEnd below.
+  const [childrenSettled, setChildrenSettled] = useState(!collapsed);
+
+  // Set inert on the children container when collapsed so they leave
+  // the tab order and assistive technology.
+  useEffect(() => {
+    if (childrenRef.current) {
+      childrenRef.current.inert = !!collapsed;
```
mxschll, line 733: Applicable to all inerts: what happens if I focused a child, and the nav collapses. Where does the focus move?

src/expandable-section/styles.scss
jkuelz: Linter doesn't allow it :/

pages/side-navigation/collapsed.page.tsx
jkuelz: No, good catch, this was just used during testing. Removed.

src/expandable-section/motion.scss
jkuelz: I discovered that the `motion.scss` file had drifted from the intended implementation: it targeted old selectors, which had I guess at some point been removed from the main file, making them obsolete and regress on the animations that should've been present in VR.
In addition to restoring those animations, I moved them to the main file because keeping those transition declarations beside the selectors which enable them makes their relationship explicit and avoids another selector/transition divergence in the future.

src/expandable-section/internal.tsx
jkuelz: Fixed and added tests.

style-dictionary/one-theme/motion.ts
jkuelz: I think ideally any component specific motion tokens are more of composites vs individual values. However, this requires a change to theme-builder to be able to parse multiple `{ }` values inside a token value. For now, I created more specific tokens that are used in these two components.

src/expandable-section/internal.tsx
jkuelz: After looking into this again, the issue was not unique to `CSSTransition` as I originally thought. It was more about how it was configured. The original had a timeout of 30ms which is far shorter than any animation, causing the 'exited' state to render true while the animations were still in progress, resulting in a mixed transient flash state. 

I found that we already had an internal wrapper around React Transition Group’s [Transition](https://reactcommunity.org/react-transition-group/transition) component, so I used that instead. It exposes the transition status, handles reduced motion, and uses the actual `transitionend` event when no timeout is provided. This lets the "settled" state reset synchronously from `expanded` on collapse and move to `entered` only after the root’s `grid-template-rows` transition finishes, without adding another custom listener here.

mxschll:
@jkuelz some remarks offline

mxschll (approved):

PR #4875, opened 2026-08-06
fix: PromptInput minRows in token-mode did not grow input height correctly

ernst-dev:
### Description

Addresses an issue where setting minRows on PromptInput while using `menus` causes the input to stay at the equivalent minimum height of one row.

### How has this been tested?
Added tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/prompt-input/__integ__/prompt-input-token-mode.test.ts
```diff
@@ -593,6 +593,78 @@ const setupTest = (testFn: (page: PromptInputTokenModePage) => Promise<void>, ur
...
+      'increasing minRows grows the input height',
+      setupTest(async page => {
+        // The height-recalculation effect depends on the token list changing, not on minRows
+        // directly, so a keystroke is needed after each minRows change to force a recompute —
+        // matching how a real user's height would update as they keep typing.
+        const heightAtOne = await page.getContentEditableHeight();
+
+        await page.setValue('#min-rows-input', '3');
+        await page.focusInput();
+        await page.keys(['a']);
```
amanabiy: Changing minRows at runtime doesn't reflow the input, the height only updates after you type a character. You can repro it on shortcuts.page.tsx. The test here is also relying on this behavior which is not ideal? Can we update it to automatically adjust, and fix the test accordingly as well?

src/prompt-input/internal.tsx
```diff
@@ -167,7 +167,12 @@ const InternalPromptInput = React.forwardRef(
       // this is required so the scrollHeight becomes dynamic, otherwise it will be locked at the highest value for the size it reached e.g. 500px
       element.style.height = 'auto';
 
-      const minRowsHeight = `calc(${LINE_HEIGHT} +  ${designTokens.spaceFieldVertical} * 2)`;
+      // The native textarea already grows to its `rows={minRows}` attribute, so this
+      // formula only needs to act as a one-line floor there. The contentEditable div used
+      // in token mode has no equivalent native sizing, so it must scale with minRows directly.
+      const minRowsHeight = isTokenMode
+        ? `calc(${minRows} * (${LINE_HEIGHT} + ${PADDING} / 2) + ${PADDING})`
```
amanabiy, line 174: The calculation here changed to using `PADDING` instead of the `designTokens.spaceFieldVertical` is this expected? and does it bring any visual difference?

src/prompt-input/internal.tsx
```diff
@@ -191,11 +196,19 @@ const InternalPromptInput = React.forwardRef(
       }
     }, [isTokenMode, tokens, adjustInputHeight, isCompactMode, placeholder]);
 
+    // When minRows/maxRows change, recalculate height synchronously — these prop
+    // changes don't involve caret position so no RAF deferral is needed.
+    useEffect(() => {
```
amanabiy, line 201: Can we remove this useEffect? since we have already added minRows, and maxRows to the existing useEffect? Both of them call `adjustInputHeight`

amanabiy (approved):

PR #4876, opened 2026-08-06
fix: Increase pointer target of expandable section navigation trigger

Harsh-Anand-Singh:
### Description

The expand chevron in the expandable section `navigation` variant (used by side navigation expandable link groups) has a ~12px pointer target, failing WCAG 2.5.8 Target Size (Minimum, 24x24px).

The header wrapper already shows `cursor: pointer` but had no click handler. This change makes it handle pointer clicks, deferring to interactive descendants (expand button, header link), so the effective target becomes the whole header row.

### How has this been tested?

- All existing unit tests for expandable-section and side-navigation pass
- Verified in browser (one-theme, side navigation permutations): dead-space clicks toggle, button click toggles once, link click navigates without toggling, nothing extends outside the component

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/expandable-section/styles.scss
```diff
@@ -239,6 +240,14 @@ $icon-total-space-medium: calc(#{$icon-width-medium} + #{$icon-margin-left} + #{
       text-decoration: none;
       flex-direction: column;
 
+      // Extends the clickable area to ensure a minimum 24x24px pointer target (WCAG 2.5.8)
+      &::after {
+        content: '';
+        position: absolute;
+        inset-block: min(0px, calc((100% - 24px) / 2));
+        inset-inline: min(0px, calc((100% - 24px) / 2));
```
SpyZzey: This will make the click target leak outside the bounding box of the expandable section and could therefore interfere with another interactive element, which is close to this one, e.g., two expandable sections above each other (reducing the clickable area of one of them to less than 24px).

SpyZzey (changes requested):

src/expandable-section/styles.scss
SpyZzey: Screenshot 2026-08-06 at 18 38 55

Harsh-Anand-Singh: Good catch, thanks. Reworked the approach: instead of extending the hit area with a pseudo-element, the header wrapper (which already had `cursor: pointer` but no handler) now handles pointer clicks itself, deferring to interactive descendants. The enlarged target is the header row itself, so it stays within the component's bounding box and can't overlap adjacent sections.

SpyZzey (changes requested):
I think this approach might have two flaws:
1. It's inconsistent and not transparent to the user. The behavior can change depending content without a visual clue (might need to be checked with a designer).
2. The accessibility issue still persists for places where there is a button or link inside the expandable section.

Harsh-Anand-Singh:
In loop with the designer, will open when a fix is decided.

PR #4878, opened 2026-08-06
feat: Add resize property to textarea

### LINKED ISSUE #4877: [Feature Request]: Support resize options on Textarea
### Description

The `Textarea` component can always be resized both vertically and horizontally. It's common on the web to restrict the axes that a textarea can be resized along. In particular, it's common to see situations where a textarea is side-by-side with other content, so limiting resizing to the y-axis with `resize: vertical` is helpful for users.

Internal tracking: AWSUI-41554

### Code of Conduct

- [x] I checked the [current issues](https://github.com/cloudscape-design/components/issues) for duplicate requests

TrevorBurnham:
### Description

Adds a `resize` property to `Textarea` so consumers can constrain the resize handle. Values mirror the CSS property: `both` (default), `horizontal`, `vertical`, `none`.

Previously the component hard-coded `resize: auto` in its stylesheet with no supported override. The only workaround was `nativeTextareaAttributes.style`, which is internal-only and whose own docs state it must not be used for custom styling.

Two notes on the implementation:

**Physical axis names, not logical.** CSS also offers `block`/`inline`, but resize axes follow `writing-mode`, not `direction` — so in `horizontal-tb` (used throughout Cloudscape) they resolve identically to `vertical`/`horizontal` in both LTR and RTL. I verified this by drag-resizing in Chrome across `dir` × `writing-mode`; the axes only swap under `vertical-rl`/`vertical-lr`, which isn't supported. Offering both spellings would add redundant surface area for identical behavior. Physical names also match every existing axis-valued prop (`Divider.Orientation`, `SpaceBetween.Direction`, `RadioGroup.direction`, `NavigableGroup.navigationDirection`); no component currently uses `block`/`inline` as an axis name.

**`resize: auto` → `resize: both` in the base rule.** `auto` isn't a spec value for `resize`; browsers compute it to `both` for textarea. Rendering is unchanged, the off-spec value is gone, and the default is now stated in the API docs.

Closes #4877

### How has this been tested?

- 4 new unit tests in `src/textarea/__tests__/textarea.test.tsx`; all 71 textarea tests pass, plus 3176 tests across the textarea, input, prompt-input, and snapshot suites.
- Verified end-to-end in Chrome 151 by driving the built `permutations` page with Puppeteer: computed style and actual drag behavior match the expected axis for all four values, in both LTR and RTL (8/8).
- New permutations entry gives visual-regression coverage of all four values.
- `documenter` snapshot updated; stylelint, eslint, and prettier are clean.

<details>

#### Correctness

- [x] Changes include appropriate documentation updates (JSDoc on the new property).
- [x] Changes are backward-compatible — `resize` defaults to `both`, which is what `resize: auto` already computed to.
- [x] No unsupported browser features used — the physical `resize` keywords have been supported since Chrome 4 / Firefox 4 / Safari 4.
- [x] Changes were manually tested for accessibility — the property only affects the native resize handle; keyboard navigation and screen reader behavior are unchanged.

#### Security

- [x] No URL handling introduced.

#### Testing

- [x] Changes are covered with new unit tests.
- [x] Visual regression coverage added via the permutations page.
</details>

---

### DIFF
--- a/pages/textarea/permutations.page.tsx
+++ b/pages/textarea/permutations.page.tsx
@@ -43,6 +43,10 @@ const permutations = createPermutations<TextareaProps>([
     placeholder: ['Placeholder with\nmultiple lines'],
     value: [''],
   },
+  {
+    resize: ['both', 'horizontal', 'vertical', 'none'],
+    value: ['Resizable in the direction given by the resize property'],
+  },
 ]);
 
 export default function TextareaPermutations() {
--- a/src/__tests__/snapshot-tests/__snapshots__/documenter.test.ts.snap
+++ b/src/__tests__/snapshot-tests/__snapshots__/documenter.test.ts.snap
@@ -31111,6 +31111,30 @@ Don't use read-only inputs outside a form.",
       "type": "boolean",
     },
     {
+      "defaultValue": "'both'",
+      "description": "Specifies the directions in which the user can resize the textarea:
+
+* \`both\` (default) - The user can change both the width and the height.
+* \`vertical\` - The user can only change the height. Use this when the textarea
+  is next to other content that a width change would disrupt.
+* \`horizontal\` - The user can only change the width.
+* \`none\` - The user can't resize the textarea. Longer content scrolls instead.
+  Consider using the \`rows\` property to make the initial height fit the expected content.",
+      "inlineType": {
+        "name": "TextareaProps.Resize",
+        "type": "union",
+        "values": [
+          "none",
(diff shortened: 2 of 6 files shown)


TrevorBurnham:
The "Lint PR" check failed with "Service unavailable," not due to any issue with this PR.

gethinwebster (approved):

gethinwebster:
Thanks for the addition!

PR #4896, opened 2026-08-12
fix: Extend resize handle pointer target to 24px minimum

Harsh-Anand-Singh:
### Description

Extends the pointer target of the resize handle (`handle-resize-area` in the internal drag-handle component, used by the code editor's resize control) to a minimum of 24x24px to meet WCAG 2.5.8 Target Size (Minimum).

The extension is a transparent `::after` pseudo-element anchored to the handle's bottom-right corner, so it grows only up and left. The visible size of the handle is unchanged, and the extension never overlaps interactive elements below or to the right of the handle. The focus ring already uses `::before`, so there is no conflict.

### How has this been tested?

- `npx jest -c jest.unit.config.js src/internal/components/drag-handle` — all 147 unit tests pass
- `stylelint` on the changed file passes
- Manually verified the enlarged hit area in the dev pages

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/internal/components/drag-handle/styles.scss
+++ b/src/internal/components/drag-handle/styles.scss
@@ -39,6 +39,17 @@
 
   &-resize-area {
     cursor: nwse-resize;
+    position: relative;
+
+    // Extend the pointer target to at least 24x24px (WCAG 2.5.8)
+    &::after {
+      content: '';
+      position: absolute;
+      inset-block-end: 0;
+      inset-inline-end: 0;
+      inline-size: max(24px, 100%);
+      block-size: max(24px, 100%);
+    }
 
     @include styles.with-direction('rtl') {
       cursor: nesw-resize;


YueyingLu:
Is it expected to change the size regardless one-theme or not?

src/internal/components/drag-handle/styles.scss
```diff
@@ -39,6 +39,17 @@
 
   &-resize-area {
     cursor: nwse-resize;
+    position: relative;
+
+    // Extend the pointer target to at least 24x24px (WCAG 2.5.8)
```
mxschll, line 44: Does this overlap with other elements on a page?

Harsh-Anand-Singh:
> Is it expected to change the size regardless one-theme or not?

Yes, that's intentional. The audit happened to surface it under one theme, but it's a pre-existing gap in all themes, so the fix applies everywhere. Since the extension is invisible there's no visual change in any theme either.

src/internal/components/drag-handle/styles.scss
Harsh-Anand-Singh: No. The pseudo-element is anchored to the handle's bottom-right corner and only grows up and left, into the component's own surface, it can never extend below or to the right of the handle.

src/internal/components/drag-handle/styles.scss
```diff
@@ -39,6 +39,17 @@
...
+    &::after {
+      content: '';
+      position: absolute;
+      inset-block-end: 0;
+      inset-inline-end: 0;
+      inline-size: max(24px, 100%);
```
mxschll, line 50: The icon is 16x16, but now we have a clickable area of 24x24. Should we increase the icon too? Otherwise I can start clicking and dragging at an invisible space.

Harsh-Anand-Singh: The invisible area isn't mystery space. Hovering anywhere in the extended area shows the resize cursor, like a pointer landing 4px off the icon was almost certainly aiming at it; starting the resize there is the expected outcome, not a surprise.

mxschll: It was pretty surprising to me.

mxschll (approved):

PR #4901, opened 2026-08-13
chore: Internal style api v2 for alert, badge, button

pan-kot:
### Description

API proposal: Exk1S1y3Ww5p

Note: the changes are added internally, the public API of the components stays unchanged just yet.

### How has this been tested?

* Unit tests
* New test pages to be later covered by screenshot tests

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/alert/style-api-v2.scss
+++ b/pages/alert/style-api-v2.scss
@@ -1,0 +1,49 @@
+/*
+ Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ SPDX-License-Identifier: Apache-2.0
+*/
+
+@use '../style-api/palette' as palette;
+
+.alert-genai {
+  --awsui-style-background-color: #{palette.$genai-surface};
+  --awsui-style-color: #{palette.$genai-text};
+  --awsui-style-border-color: #{palette.$genai};
+  --awsui-style-border-radius: 4px;
+  --awsui-style-padding-block: 12px;
+  --awsui-style-padding-inline: 24px;
+  --awsui-style-focus-ring-color: #{palette.$genai};
+}
+
+.alert-genai-icon {
+  --awsui-style-color: #{palette.$genai};
+}
+
+.alert-genai-action {
+  --awsui-style-background-color: #{palette.$genai};
+  --awsui-style-color: #{palette.$text-on-accent};
+  --awsui-style-border-color: #{palette.$genai};
+  --awsui-style-border-radius: 4px;
+  --awsui-style-focus-ring-color: #{palette.$genai};
+
+  &:hover,
+  &:focus {
+    --awsui-style-background-color: #{palette.$genai-hover};
+    --awsui-style-border-color: #{palette.$genai-hover};
+  }
+
+  &:active {
+    --awsui-style-background-color: #{palette.$genai-active};
+    --awsui-style-border-color: #{palette.$genai-active};
+  }
+}
+
+.alert-genai-dismiss {
+  --awsui-style-color: #{palette.$genai-text};
+  --awsui-style-focus-ring-color: #{palette.$genai};
+
+  &:hover,
+  &:focus {
+    --awsui-style-color: #{palette.$genai};
+  }
+}
--- a/pages/button/style-api-v2.page.tsx
+++ b/pages/button/style-api-v2.page.tsx
@@ -1,0 +1,109 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import React, { useContext } from 'react';
+import clsx from 'clsx';
+
+import { Button, Checkbox, KeyValuePairs, SpaceBetween } from '~components';
+
+import AppContext, { AppContextType } from '../app/app-context';
+import { SimplePage } from '../app/templates';
+
+import styles from './style-api-v2.scss';
+
+type PageContext = React.Context<
+  AppContextType<{
+    loading?: boolean;
+    disabled?: boolean;
+    disabledReason?: boolean;
+  }>
+>;
+
+export default function () {
+  const {
+    urlParams: { loading = false, disabled = false, disabledReason = false },
+    setUrlParams,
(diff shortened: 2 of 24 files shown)


pages/alert/style-api-v2.scss
```diff
@@ -0,0 +1,49 @@
+/*
+ Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ SPDX-License-Identifier: Apache-2.0
+*/
+
+@use '../style-api/palette' as palette;
+
+.alert-genai {
+  --awsui-style-background-color: #{palette.$genai-surface};
```
mxschll, line 9: Nit: since this is a test page it would make it easier to cross reference the rendered page and the colors if they were explicitly defined here in one place. But I can see why `[palette.scss‎](https://github.com/cloudscape-design/components/pull/4901/changes#diff-25e9b45620fcab75963e99910fec894db13be84ca60225b302ff0808c14ab1bc)` was added.

src/button/__tests__/style-api.test.tsx
```diff
@@ -0,0 +1,32 @@
...
+    expect(link).not.toHaveClass('unknown');
+
+    const button = renderButton(<Button {...({ styleClassNames: { root: 'b', unknown: 'c' } } as any)} />);
+    expect(button).toHaveClass('b');
+    expect(button).not.toHaveClass('unknown');
+  });
+
+  test('does not leak the styleClassNames prop to the DOM', () => {
+    const link = renderButton(<Button href="#" {...({ styleClassNames: { root: 'a' } } as any)} />);
+    expect(link).not.toHaveAttribute('styleClassNames');
```
mxschll, line 27: Is there a case when that can happen?

pages/button/style-api-v2.page.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+export default function () {
+  const {
+    urlParams: { loading = false, disabled = false, disabledReason = false },
+    setUrlParams,
+  } = useContext(AppContext as PageContext);
+  const shared = { loading, disabled, disabledReason: disabledReason ? 'Disabled reason' : undefined };
+  const disabledStyle = disabled || loading ? styles.disabled : undefined;
+  return (
+    <SimplePage
```
mxschll, line 30: Do we want to set `disableAnimations={true}` for screenshot tests?

pages/button/style-api-v2.scss
```diff
@@ -0,0 +1,81 @@
...
+    --awsui-style-border-color: #{palette.$surface};
+  }
+
+  &:active {
+    --awsui-style-background-color: #{palette.$surface-alt};
+    --awsui-style-border-color: #{palette.$surface-alt};
+  }
+}
+
+.disabled.disabled {
```
mxschll, line 77: Will consumers need to use specificity hack?

mxschll: I guess you could also swap out the target class, so its the builders choice how to handle this

src/button/__tests__/style-api.test.tsx
mxschll: Further down we test:
```
expect(container.querySelector('[styleClassNames]')).toBeNull();
```

I wonder if this needed at all?

src/button/styles.scss
```diff
@@ -10,40 +10,86 @@
...
+      var(#{custom-props.$styleBorderColorActive}, map.get($variant, 'active-border-color'))
+    );
+    color: var(
+      #{style-api.read($tokens, color)},
+      var(#{custom-props.$styleColorActive}, map.get($variant, 'active-color'))
+    );
+    box-shadow: var(#{style-api.read($tokens, box-shadow)}, var(#{custom-props.$styleBoxShadowActive}));
   }
 
   &.disabled {
```
mxschll, line 79: As a builder I'd expect to be able to select for `:disabled`, but we are not supporting that pseudo class. Is there a reason?

mxschll: Input also uses `:disabled`. This might be an opportunity to make it consistent.

pages/alert/style-api-v2.scss
pan-kot: The actual token values are of little importance - so I decided to move most of them to a shared place primarily to not re-invent colour combinations for every components that comes after

src/button/__tests__/style-api.test.tsx
pan-kot: This is to ensure that we do not pass all props to the underlying DOM nodes, like this:

```
<button 
   ...
   className={clsx(styles.button, props.styleClassNames.root, ...)}
   {...props} // styleClassNames is passed as an attribute and causes a warning
/>
```

pages/button/style-api-v2.page.tsx
pan-kot: These do not have animations, so it is fine to leave as is. I will create a follow up to use disableAnimations=true by default when screenshotArea is used.

pages/button/style-api-v2.scss
pan-kot: This does not conflict with the internal specificity of the component itself - so yes, it is on the builders to decide

src/button/styles.scss
pan-kot: The `:disabled` is technically supported and works for disabled buttons, but:
1. It does not target loading buttons (those are aria-disabled)
2. It does not target buttons with disabled reasons (those are aria-disabled, too)

The button's style page in the docs will have a guidance section to set the correct expectations. The builders can then choose to follow the original component behaviour or choose to diverge and e.g. style loading buttons differently.

pages/alert/style-api-v2.scss
mxschll: Yes, just made cross-referencing it harder. Maybe I should've categorized as "Super Nit" :)

pages/button/style-api-v2.page.tsx
mxschll: The loading spinner was spinning when I opened the page

mxschll (approved):

PR #4904, opened 2026-08-13
feat: Add dialog component

amanabiy:
### Description
Adds an inline, non-modal Dialog component for in-context prompts, authorizations, and feedback. Dialog supports header actions, content, and footer slots, keeps the surrounding page interactive, manages initial and restored focus, and supports dismissal through the close button and Escape.

Related links, issue #, if available: 6f6tUHjL27wB - doc

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/pages/dialog/permutations.page.tsx
+++ b/pages/dialog/permutations.page.tsx
@@ -25,6 +25,7 @@ const permutations = createPermutations<DialogProps>([
   // Header + content + footer
   {
     initialFocus: ['none'],
+    open: [true],
     header: ['What’s your main goal?'],
     children: ['Choose the option that best matches your intent.'],
     footer: [footer],
@@ -32,12 +33,14 @@ const permutations = createPermutations<DialogProps>([
   // Header + content only (no footer)
   {
     initialFocus: ['none'],
+    open: [true],
     header: ['Authorize account access'],
     children: ['Read and write access will be granted to the deployment agent.'],
   },
   // Long content to check wrapping/scroll behavior
   {
     initialFocus: ['none'],
+    open: [true],
     header: ['Tell us what went wrong'],
     children: [
       'This is a longer body of content used to verify that the dialog grows with its content and that the header, content, and footer regions remain visually distinct across a taller layout.',
--- a/pages/dialog/permutations.page.tsx
+++ b/pages/dialog/permutations.page.tsx
@@ -1,0 +1,60 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import React from 'react';
+
+import Button from '~components/button';
+import Dialog from '~components/dialog';
+import { DialogProps } from '~components/dialog/interfaces';
+import SpaceBetween from '~components/space-between';
+
+import createPermutations from '../utils/permutations';
+import PermutationsView from '../utils/permutations-view';
+import ScreenshotArea from '../utils/screenshot-area';
+
+const footer = (
+  <SpaceBetween direction="horizontal" size="xs">
+    <Button>Skip</Button>
+    <Button variant="primary">Continue</Button>
+  </SpaceBetween>
+);
+
+// initialFocus is forced to 'none' so the many rendered dialogs on the page do
+// not fight over focus during screenshot capture. The close button is always
+// present, so it is not a permutation axis.
+const permutations = createPermutations<DialogProps>([
+  // Header + content + footer
+  {
+    initialFocus: ['none'],
+    header: ['What’s your main goal?'],
+    children: ['Choose the option that best matches your intent.'],
+    footer: [footer],
+  },
+  // Header + content only (no footer)
+  {
+    initialFocus: ['none'],
+    header: ['Authorize account access'],
+    children: ['Read and write access will be granted to the deployment agent.'],
+  },
+  // Long content to check wrapping/scroll behavior
+  {
+    initialFocus: ['none'],
+    header: ['Tell us what went wrong'],
+    children: [
+      'This is a longer body of content used to verify that the dialog grows with its content and that the header, content, and footer regions remain visually distinct across a taller layout.',
+    ],
+    footer: [footer],
+  },
+]);
+
+export default function DialogPermutationsPage() {
+  return (
+    <>
+      <h1>Dialog permutations</h1>
+      <ScreenshotArea disableAnimations={true}>
+        <div style={{ maxInlineSize: 520 }}>
+          <PermutationsView permutations={permutations} render={permutation => <Dialog {...permutation} />} />
+        </div>
+      </ScreenshotArea>
+    </>
+  );
+}
--- a/pages/dialog/simple.page.tsx
+++ b/pages/dialog/simple.page.tsx
@@ -1,0 +1,69 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import React, { useState } from 'react';
+
+import Button from '~components/button';
+import Dialog from '~components/dialog';
+import FormField from '~components/form-field';
+import PromptInput from '~components/prompt-input';
+import RadioGroup from '~components/radio-group';
+import SpaceBetween from '~components/space-between';
+
+import ScreenshotArea from '../utils/screenshot-area';
+import { DialogPage, FakeTranscript } from './common';
+
+// Basic dialog: the minimal in-flow, non-modal shape. The fake transcript above
+// and the PromptInput below exist to prove reflow (the dialog pushes content,
+// it does not overlay) and transcript -> dialog -> prompt tab order.
+export default function DialogSimplePage() {
+  const [open, setOpen] = useState(true);
+  const [goal, setGoal] = useState('increase');
+  const [prompt, setPrompt] = useState('');
+
+  return (
+    <DialogPage title="Dialog — basic (in-flow, non-modal)">
+      <ScreenshotArea>
+        <SpaceBetween size="m">
+          <FakeTranscript />
+
+          {open ? (
+            <Dialog
+              header="What's your main goal?"
+              i18nStrings={{ dismissAriaLabel: 'Close' }}
+              onDismiss={() => setOpen(false)}
+              footer={
+                <SpaceBetween direction="horizontal" size="xs">
+                  <Button onClick={() => setOpen(false)}>Skip</Button>
+                  <Button variant="primary" onClick={() => setOpen(false)}>
+                    Continue
+                  </Button>
+                </SpaceBetween>
+              }
+            >
+              <FormField label="Select the option that best matches your intent">
+                <RadioGroup
+                  value={goal}
+                  onChange={({ detail }) => setGoal(detail.value)}
+                  items={[
+                    { value: 'increase', label: 'Increase volume size' },
+                    { value: 'change-type', label: 'Change volume type' },
+                    { value: 'something-else', label: 'Something else' },
+                  ]}
+                />
+              </FormField>
+            </Dialog>
+          ) : (
+            <Button onClick={() => setOpen(true)}>Re-open dialog</Button>
+          )}
+
+          <PromptInput
+            value={prompt}
+            onChange={({ detail }) => setPrompt(detail.value)}
+            placeholder="Ask a question"
+            ariaLabel="Chat prompt"
+          />
+        </SpaceBetween>
+      </ScreenshotArea>
+    </DialogPage>
+  );
+}
--- a/pages/dialog/simple.page.tsx
+++ b/pages/dialog/simple.page.tsx
@@ -26,35 +26,33 @@ export default function DialogSimplePage() {
         <SpaceBetween size="m">
           <FakeTranscript />
 
-          {open ? (
-            <Dialog
-              header="What's your main goal?"
-              i18nStrings={{ dismissAriaLabel: 'Close' }}
-              onDismiss={() => setOpen(false)}
-              footer={
-                <SpaceBetween direction="horizontal" size="xs">
-                  <Button onClick={() => setOpen(false)}>Skip</Button>
-                  <Button variant="primary" onClick={() => setOpen(false)}>
-                    Continue
-                  </Button>
-                </SpaceBetween>
-              }
-            >
-              <FormField label="Select the option that best matches your intent">
-                <RadioGroup
-                  value={goal}
-                  onChange={({ detail }) => setGoal(detail.value)}
-                  items={[
-                    { value: 'increase', label: 'Increase volume size' },
-                    { value: 'change-type', label: 'Change volume type' },
-                    { value: 'something-else', label: 'Something else' },
-                  ]}
-                />
-              </FormField>
-            </Dialog>
-          ) : (
-            <Button onClick={() => setOpen(true)}>Re-open dialog</Button>
-          )}
+          <Dialog
+            open={open}
+            header="What's your main goal?"
+            i18nStrings={{ dismissAriaLabel: 'Close' }}
+            onDismiss={() => setOpen(false)}
+            footer={
+              <SpaceBetween direction="horizontal" size="xs">
+                <Button onClick={() => setOpen(false)}>Skip</Button>
+                <Button variant="primary" onClick={() => setOpen(false)}>
+                  Continue
+                </Button>
+              </SpaceBetween>
+            }
+          >
+            <FormField label="Select the option that best matches your intent">
+              <RadioGroup
+                value={goal}
+                onChange={({ detail }) => setGoal(detail.value)}
+                items={[
+                  { value: 'increase', label: 'Increase volume size' },
+                  { value: 'change-type', label: 'Change volume type' },
+                  { value: 'something-else', label: 'Something else' },
+                ]}
+              />
+            </FormField>
+          </Dialog>
+          {!open && <Button onClick={() => setOpen(true)}>Re-open dialog</Button>}
 
           <PromptInput
             value={prompt}
--- a/src/dialog/__tests__/dialog.test.tsx
+++ b/src/dialog/__tests__/dialog.test.tsx
@@ -5,6 +5,8 @@ import { fireEvent, render } from '@testing-library/react';
 
 import Dialog from '../../../lib/components/dialog';
 
+import styles from '../../../lib/components/dialog/styles.css.js';
+
 function renderDialog(jsx: React.ReactElement) {
   const { container } = render(jsx);
   const dialog = container.querySelector<HTMLElement>('[role="dialog"]')!;
@@ -68,4 +70,39 @@ test('does not move focus when initialFocus is none', () => {
   );
   const labelledBy = dialog.getAttribute('aria-labelledby')!;
   expect(document.activeElement).not.toBe(document.getElementById(labelledBy));
+});
+
+test('applies the hidden class and does not move focus when open is false', () => {
+  const { dialog } = renderDialog(
+    <Dialog header="Title" open={false} onDismiss={() => {}}>
+      content
+    </Dialog>
+  );
+  expect(dialog.className).toContain(styles.hidden);
+  const labelledBy = dialog.getAttribute('aria-labelledby')!;
+  expect(document.activeElement).not.toBe(document.getElementById(labelledBy));
+});
+
+test('when controlled, moves focus in on open and restores it to the prior element on close', () => {
+  const view = (open: boolean) => (
+    <>
+      <button data-testid="trigger">Open</button>
+      <Dialog header="Title" open={open} onDismiss={() => {}}>
+        content
+      </Dialog>
+    </>
+  );
+  const { container, rerender } = render(view(false));
+  const trigger = container.querySelector<HTMLButtonElement>('[data-testid="trigger"]')!;
+  trigger.focus();
+  expect(document.activeElement).toBe(trigger);
+
+  // open: focus moves into the dialog header
+  rerender(view(true));
+  const labelledBy = container.querySelector('[role="dialog"]')!.getAttribute('aria-labelledby')!;
+  expect(document.activeElement).toBe(document.getElementById(labelledBy));
+
+  // close: focus is restored to the trigger
+  rerender(view(false));
+  expect(document.activeElement).toBe(trigger);
 });
--- a/src/dialog/__tests__/dialog.test.tsx
+++ b/src/dialog/__tests__/dialog.test.tsx
@@ -14,13 +14,21 @@ function renderDialog(jsx: React.ReactElement) {
 }
 
 test('renders a role="dialog" element', () => {
-  const { dialog } = renderDialog(<Dialog header="Title">content</Dialog>);
+  const { dialog } = renderDialog(
+    <Dialog header="Title" open={true} onDismiss={() => {}}>
+      content
+    </Dialog>
+  );
   expect(dialog).not.toBeNull();
   expect(dialog).toHaveTextContent('content');
 });
 
 test('uses the header as the accessible name via aria-labelledby', () => {
-  const { dialog } = renderDialog(<Dialog header="What's your goal?">body</Dialog>);
+  const { dialog } = renderDialog(
+    <Dialog header="What's your goal?" open={true} onDismiss={() => {}}>
+      body
+    </Dialog>
+  );
   const labelledBy = dialog.getAttribute('aria-labelledby');
   expect(labelledBy).toBeTruthy();
   const header = document.getElementById(labelledBy!);
@@ -28,14 +36,18 @@ test('uses the header as the accessible name via aria-labelledby', () => {
 });
 
 test('does not set aria-modal', () => {
-  const { dialog } = renderDialog(<Dialog header="Title">content</Dialog>);
+  const { dialog } = renderDialog(
+    <Dialog header="Title" open={true} onDismiss={() => {}}>
+      content
+    </Dialog>
+  );
   expect(dialog.getAttribute('aria-modal')).toBeNull();
 });
 
 test('always renders a dismiss button and fires onDismiss when clicked', () => {
   const onDismiss = jest.fn();
   const { dialog } = renderDialog(
-    <Dialog header="Title" i18nStrings={{ dismissAriaLabel: 'Close' }} onDismiss={onDismiss}>
+    <Dialog header="Title" open={true} i18nStrings={{ dismissAriaLabel: 'Close' }} onDismiss={onDismiss}>
       content
     </Dialog>
   );
@@ -48,7 +60,7 @@ test('always renders a dismiss button and fires onDismiss when clicked', () => {
 test('fires onDismiss when Escape is pressed inside the dialog', () => {
   const onDismiss = jest.fn();
   const { dialog } = renderDialog(
-    <Dialog header="Title" onDismiss={onDismiss}>
+    <Dialog header="Title" open={true} onDismiss={onDismiss}>
       content
     </Dialog>
   );
@@ -56,15 +68,19 @@ test('fires onDismiss when Escape is pressed inside the dialog', () => {
   expect(onDismiss).toHaveBeenCalledTimes(1);
 });
 
-test('moves focus to the header on mount by default', () => {
-  const { dialog } = renderDialog(<Dialog header="Title">content</Dialog>);
+test('moves focus to the header when opened by default', () => {
+  const { dialog } = renderDialog(
+    <Dialog header="Title" open={true} onDismiss={() => {}}>
+      content
+    </Dialog>
+  );
   const labelledBy = dialog.getAttribute('aria-labelledby')!;
   expect(document.activeElement).toBe(document.getElementById(labelledBy));
 });
 
 test('does not move focus when initialFocus is none', () => {
   const { dialog } = renderDialog(
-    <Dialog header="Title" initialFocus="none">
+    <Dialog header="Title" open={true} initialFocus="none" onDismiss={() => {}}>
       content
     </Dialog>
   );
--- a/src/dialog/__tests__/dialog.test.tsx
+++ b/src/dialog/__tests__/dialog.test.tsx
@@ -1,0 +1,71 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import React from 'react';
+import { fireEvent, render } from '@testing-library/react';
+
+import Dialog from '../../../lib/components/dialog';
+
+function renderDialog(jsx: React.ReactElement) {
+  const { container } = render(jsx);
+  const dialog = container.querySelector<HTMLElement>('[role="dialog"]')!;
+  return { container, dialog };
+}
+
+test('renders a role="dialog" element', () => {
+  const { dialog } = renderDialog(<Dialog header="Title">content</Dialog>);
+  expect(dialog).not.toBeNull();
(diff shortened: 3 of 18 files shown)


src/dialog/interfaces.ts
```diff
@@ -0,0 +1,63 @@
...
+   * Remove the dialog from the render tree in response. When the dialog unmounts, it returns
+   * focus to the element that was focused before it mounted, as long as that element still exists.
+   */
+  onDismiss: NonCancelableEventHandler<DialogProps.DismissDetail>;
+}
+
+export namespace DialogProps {
+  export interface DismissDetail {
+    reason: string;
+  }
```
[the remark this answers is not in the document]
amanabiy: This is similar to how we have it in modal https://github.com/cloudscape-design/components/blob/74e4f3d0f6f500eee81b14e4ae7488ca04d27798/src/modal/interfaces.ts#L114.

pages/dialog/permutations.page.tsx
```diff
@@ -0,0 +1,89 @@
...
+  <SpaceBetween size="m">
+    {content}
+    <Box variant="p">
+      This longer content verifies wrapping, responsive width, section spacing, and focus movement between Dialog and
+      the surrounding page.
+    </Box>
+  </SpaceBetween>
+);
+
+// The close button is always present. The closed playground state renders no Dialog,
```
pan-kot: nit: Let's remove this comment - this looks to be an agentic ai response to a prompt.

pages/dialog/simple.page.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+  return (
+    <SimplePage title="Dialog: Simple" screenshotArea={{}}>
+      <div style={{ maxInlineSize: 520 }}>
+        <SpaceBetween size="m">
+          <Button onClick={() => setOpen(true)}>Open dialog</Button>
+          {open && (
+            <Dialog
+              header="What's your main goal?"
+              i18nStrings={{ dismissAriaLabel: 'Close' }}
```
pan-kot: nit: passing `i18n={}` to the SimplePage helper injects IntlProvider - which should populate all i18nStrings.

pages/dialog/simple.page.tsx
```diff
@@ -0,0 +1,57 @@
...
+import Dialog from '~components/dialog';
+import FormField from '~components/form-field';
+import RadioGroup from '~components/radio-group';
+import SpaceBetween from '~components/space-between';
+
+import { SimplePage } from '../app/templates';
+
+// Minimal conditional-rendering example. Opening Dialog moves focus from the
+// trigger into Dialog. Closing Dialog restores focus to the trigger.
+export default function DialogSimplePage() {
```
pan-kot, line 24: I'd recommend to include a setting or a separate page to imitate a case when the trigger is no longer available after dialog opens (or it opens programmatically, so there is no trigger in the first place). The page should show how to handle focus manually. I'd be nice to secure both cases with an integration test.

src/dialog/__tests__/dialog.test.tsx
```diff
@@ -0,0 +1,283 @@
...
+import TestI18nProvider from '../../../lib/components/i18n/testing';
+import createWrapper from '../../../lib/components/test-utils/dom';
+
+function renderDialog(props: Partial<DialogProps> = {}) {
+  const renderResult = render(<Dialog header="Title" onDismiss={() => {}} {...props} />);
+  return createWrapper(renderResult.container).findDialog()!;
+}
+
+describe('Dialog', () => {
+  describe('root element', () => {
```
pan-kot: nit: I'd recommend to make the structure of tests more flat: are those little groups like "root element" or "header" really needed?

src/dialog/__tests__/dialog.test.tsx
```diff
@@ -0,0 +1,283 @@
...
+    test('renders with role="dialog"', () => {
+      const wrapper = renderDialog({ children: 'content' });
+      expect(wrapper.getElement()).toHaveAttribute('role', 'dialog');
+      expect(wrapper.getElement()).toHaveTextContent('content');
+    });
+
+    test('does not set aria-modal', () => {
```
pan-kot: Why do we need this test? It seems random: if we were to capture everything that the dialog components does not set - it'd be a long list.

src/dialog/__tests__/dialog.test.tsx
```diff
@@ -0,0 +1,283 @@
...
+    test('renders children as ReactNode', () => {
+      const wrapper = renderDialog({
+        children: <div data-testid="custom">Custom content</div>,
+      });
+      expect(wrapper.findContent()!.getElement().querySelector('[data-testid="custom"]')).toHaveTextContent(
+        'Custom content'
+      );
+    });
+
+    test('does not render content when children are not provided', () => {
```
pan-kot: nit: I'd question the practicality of this test, as it is hard to imagine a real use case when the dialog without content is needed.

src/dialog/interfaces.ts
```diff
@@ -0,0 +1,63 @@
...
+   *
+   * Remove the dialog from the render tree in response. When the dialog unmounts, it returns
+   * focus to the element that was focused before it mounted, as long as that element still exists.
+   */
+  onDismiss: NonCancelableEventHandler<DialogProps.DismissDetail>;
+}
+
+export namespace DialogProps {
+  export interface DismissDetail {
+    reason: string;
```
pan-kot: Do we call this field "reason" in some other components? E.g. in the Drawer a similar prop is called "method".

Why does it have a string type and not a union of possible values?

pan-kot: Also take a look at how we name the values in other components - e.g. should we use "closeButton" or "close-button" (I'd argue we also need to call it "dismiss-button" to match the terminology used in the component).

src/dialog/__tests__/dialog.test.tsx
```diff
@@ -0,0 +1,283 @@
...
+        children: 'content',
+        i18nStrings: { dismissAriaLabel: 'Close' },
+        onDismiss,
+      });
+      expect(wrapper.findDismissButton()).not.toBeNull();
+      wrapper.findDismissButton()!.click();
+      expect(onDismiss).toHaveBeenCalledWith(expect.objectContaining({ detail: { reason: 'closeButton' } }));
+    });
+
+    test('fires onDismiss with the keyboard reason and restores focus when Escape is pressed', async () => {
```
pan-kot: Why do we use reason="keyboard" and not reason="escape"? This can be misleading, as one might think it means that the user activated the dismiss button by focusing it with the keyboard and then pressing Space or Enter.

pan-kot: Why do we combine focus management and event details check in one test? The focus management should not depend on the dismiss method - we should ensure it happens in both cases: when the user presses Escape or clicks on the dismiss button. We should either modify the prev test to also include focus check, or create separate tests to capture all four behaviours.

src/dialog/__tests__/dialog.test.tsx
```diff
@@ -0,0 +1,283 @@
...
+      );
+      const useMergeRefsSpy = jest.spyOn(componentToolkit, 'useMergeRefs').mockReturnValueOnce(() => {});
+
+      const activeElement = document.activeElement;
+      const wrapper = renderDialog({
+        i18nStrings: { dismissAriaLabel: 'Close' },
+      });
+
+      expect(document.activeElement).toBe(activeElement);
+      expect(wrapper.findDismissButton()!.getElement()).not.toHaveFocus();
```
pan-kot: why do we need this second assertion?

pan-kot: What is the purpose of this test, was it created for test coverage?

When the dialog is rendered - it is expected to obtain focus. When it is not rendered - then there is no concern. In this test we render the dialog while breaking one of its internal dependencies - this is not a good proxy for any real use case.

src/dialog/__tests__/dialog.test.tsx
```diff
@@ -0,0 +1,283 @@
...
+
+    test('moves focus to the close button when it mounts', () => {
+      const wrapper = renderDialog({
+        children: 'content',
+        i18nStrings: { dismissAriaLabel: 'Close' },
+      });
+      expect(document.activeElement).toBe(wrapper.findDismissButton()!.getElement());
+    });
+
+    test('moves focus to the first headerActions control when present', () => {
```
pan-kot: Why are we doing that?

Can we instead consistently move focus e.g. to the dialog body - so that it announces its role and the header, instead of announcing either the dismiss or first action label, depending on its configuration?

We use this approach for standalone drawers: https://cloudscape.design/components/drawer/?tabId=playground&example=fixed-position-with-backdrop-and-close-action

src/dialog/__tests__/dialog.test.tsx
```diff
@@ -0,0 +1,283 @@
...
+
+    test('does not restore focus when focus was already inside the dialog on mount', async () => {
+      const trigger = document.createElement('button');
+      document.body.appendChild(trigger);
+      trigger.focus();
+
+      const { container, unmount } = render(
+        <Dialog
+          header="Title"
+          headerActions={<button autoFocus={true}>Action</button>}
```
pan-kot: From the previous test it is clear that the first header action is getting focused by the dialog in either case - so I'd recommend experimenting with a second one.

I would also be nice to see how this behaviour works in different browsers, as there can be some race conditions at play.

src/dialog/__tests__/dialog.test.tsx
```diff
@@ -0,0 +1,283 @@
...
+          i18nStrings={{ dismissAriaLabel: 'Close' }}
+          onDismiss={() => {}}
+        />
+      );
+      const wrapper = createWrapper(container).findDialog()!;
+      expect(wrapper.findHeaderActions()!.getElement().querySelector('button')).toHaveFocus();
+
+      unmount();
+      await new Promise(resolve => setTimeout(resolve, 0));
+      expect(trigger).not.toHaveFocus();
```
pan-kot: What is the motivation for not restoring focus? As the dialog closes - the element to restore focus to will no longer be available, but that is also true for cases when the dialog is opened programmatically, or when the trigger is unmounted after click. Should we make the behaviour more predicable by always restoring focus if the prev focused element is still in DOM? When it is not - it is then consumer responsibility to forward it anyways.

src/dialog/__tests__/dialog.test.tsx
```diff
@@ -0,0 +1,283 @@
...
+      trigger.remove();
+    });
+
+    test('moves focus in on mount and restores focus on unmount', async () => {
```
pan-kot: There are already tests that assert initial focus transition and recovery - what is unique about this one?

src/dialog/__tests__/dialog.test.tsx
```diff
@@ -0,0 +1,283 @@
...
+            </Dialog>
+          )}
+        </>
+      );
+      const { container, rerender } = render(view(false));
+      const trigger = container.querySelector<HTMLButtonElement>('[data-testid="trigger"]')!;
+      trigger.focus();
+      expect(document.activeElement).toBe(trigger);
+
+      rerender(view(true));
```
pan-kot: Should we instead do the following check:
1. Open the dialog
2. Ensure the focus is on the dismiss button
3. Move focus to another element
4. Re-render the dialog and ensure the focus stays on another element (the dialog should not steal focus at that point)

src/dialog/__tests__/dialog.test.tsx
```diff
@@ -0,0 +1,283 @@
...
+      const wrapper = createWrapper(container).findDialog()!;
+      expect(document.activeElement).toBe(wrapper.findDismissButton()!.getElement());
+
+      rerender(view(false));
+      await waitFor(() => expect(document.activeElement).toBe(trigger));
+    });
+
+    test('restores the original trigger when the consumer moves focus while dismissing', async () => {
+      function Harness() {
```
pan-kot: nit: can we create some renderStatefulDialog function to include the trigger and additional focus target - so that we don't have to create custom components in tests?

Like this:

```
function renderStatefulDialog({ open, ...dialogProps }) {
  const [isOpen, setIsOpen] = useState(open);
  const { container } = render(
    <>
      <button data-testid="trigger" />
      <button data-testid="focus-target" />
      {isOpen ? <Dialog {...dialogProps} /> : null}
    <>
  );
  return { container, dialog: createWrapper().findDialog() }
}
```

(note that we don't have to use `focusTargetRef` as the required element can be found by testid on the fly).

src/dialog/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+import InternalDialog from './internal';
+
+export { DialogProps };
+
+export default function Dialog(props: DialogProps) {
+  const baseComponentProps = useBaseComponent('Dialog', {
+    props: {},
+    metadata: {
+      hasHeaderActions: Boolean(props.headerActions),
+      hasContent: Boolean(props.children),
```
pan-kot: Why do we need `hasContent` - do we anticipate use cases where content is not present?

src/dialog/interfaces.ts
```diff
@@ -0,0 +1,63 @@
...
+
+export interface DialogProps extends BaseComponentProps {
+  /**
+   * Specifies the heading of the dialog and labels it as its accessible name through `aria-labelledby`.
+   */
+  header: React.ReactNode;
+
+  /**
+   * Actions displayed in the header row, before the always-present close button.
+   * Use this for controls related to the dialog, such as a compact `Pagination`.
```
pan-kot: nit: Do we really need to mention the compact pagination here? It is one of our use cases, which will be surfaced in the component guidelines anyways. This line in the api docs feels random.

src/dialog/interfaces.ts
```diff
@@ -0,0 +1,63 @@
...
+   */
+  headerActions?: React.ReactNode;
+
+  /**
+   * Content of the dialog. The dialog is a shell: the interactive content it hosts
```
pan-kot: "The dialog is a shell" - what does it mean? Do we use this terminology in any other similar components, e.g. modal or container?

src/dialog/interfaces.ts
pan-kot: Thinking some more about dialog dismissal - is there any good reason to prevent dialog from closing on-escape, while keeping the dismiss button? If we don't have any use cases - let's remove the property, too.

If in the future we allow hiding the dismiss button - then we should disable the on-escape dismissal, too - relying on consumers to provide it if needed.

src/dialog/interfaces.ts
```diff
@@ -0,0 +1,63 @@
...
+}
+
+export namespace DialogProps {
+  export interface DismissDetail {
+    reason: string;
+  }
+
+  export interface I18nStrings {
+    /**
+     * Specifies the ARIA label for the dismiss button.
```
pan-kot: This description repeats the one we gave for i18nStrings, and it is not parsed by the Documenter. Let's remove it.

src/dialog/internal.tsx
```diff
@@ -0,0 +1,112 @@
...
+  children,
+  footer,
+  headerActions,
+  i18nStrings,
+  onDismiss,
+  __internalRootRef,
+  ...restProps
+}: InternalDialogProps) {
+  const baseProps = getBaseProps(restProps);
+  const isRefresh = useVisualRefresh();
```
pan-kot: The visual refresh is our default theme - we should not use this util for new components or features.

src/dialog/internal.tsx
```diff
@@ -0,0 +1,112 @@
...
+  useEffect(() => {
+    const dialog = dialogRef.current;
+    if (!dialog) {
+      return;
+    }
+
+    const activeElement = dialog.ownerDocument.activeElement;
+    const HTMLElementConstructor = dialog.ownerDocument.defaultView?.HTMLElement;
+    const restoreTarget =
+      HTMLElementConstructor && activeElement instanceof HTMLElementConstructor && !dialog.contains(activeElement)
```
pan-kot: What is HTMLElementConstructor and why are we doing this?

Can we use the same logic to restore focus as we use in the modal? If we cannot use the focus-lock component - we can at least use the same code, by either copying or extracting the relevant parts.

src/dialog/internal.tsx
```diff
@@ -0,0 +1,112 @@
...
+        ? activeElement
+        : null;
+
+    // Focus the first headerActions control when present, otherwise the
+    // always-present close button. The heading is not a focus target.
```
pan-kot: This description is technically incorrect: the dialog's header type is ReactNode - which means it can include interactive elements, too. I don't think this line requires a description - but most importantly I don't think that this way of handling focus is good (see my suggestion above on focusing the dialog body instead).

src/dialog/internal.tsx
```diff
@@ -0,0 +1,112 @@
...
+    getFirstFocusable(dialog)?.focus();
+
+    return () => {
+      if (restoreTarget?.isConnected) {
+        // Cleanup can run synchronously while onDismiss is still executing. Defer
+        // restoration so focus reliably returns to the original trigger after the handler finishes.
+        dialog.ownerDocument.defaultView?.setTimeout(() => {
```
pan-kot: why do we call setTimeout from `dialog.ownerDocument.defaultView`?

style-dictionary/visual-refresh/metadata/borders.ts
```diff
@@ -27,6 +27,21 @@ const metadata: StyleDictionary.MetadataIndex = {
...
+    description: 'The border radius of the Dialog surface.',
+    public: false,
+    themeable: true,
+  },
+  borderWidthDialog: {
+    description: 'The border width of the Dialog surface.',
+    public: false,
+    themeable: true,
+  },
+  borderDialogTopWidth: {
```
pan-kot: There is no outside visual refresh. We should remove this var.

style-dictionary/visual-refresh/colors.ts
```diff
@@ -124,7 +124,7 @@ const tokens: StyleDictionary.ColorsDictionary = {
   colorBackgroundAlertInfo: '{colorBackgroundStatusInfo}',
   colorBackgroundAlertSuccess: '{colorBackgroundStatusSuccess}',
   colorBackgroundAlertWarning: '{colorBackgroundStatusWarning}',
-  colorBackgroundDialog: '{colorBackgroundStatusInfo}',
+  colorBackgroundDialog: '{colorBackgroundItemCard}',
```
pan-kot: Why do we refer to the item card here?

src/dialog/interfaces.ts
amanabiy: The DismissDetails with reason as a strign is similar to the API in modal https://github.com/cloudscape-design/components/blob/e404a456f43dd1a02fc9aff8897a8becd355cdbe/src/modal/interfaces.ts#L114 The documentation that says `closeButton` and `keyboard` is also similar to the modal. I think adding the reason is beneficial if customers need to control the closing behavior when it is triggered by escape.

src/dialog/__tests__/dialog.test.tsx
amanabiy: I agree, it's better to change name for the reason to `escape` instead of `reason`. The initial reasoning was to keep it inline with the Modal api surface, but the `escape` reason is much better and cleaner.

https://github.com/cloudscape-design/components/blob/e404a456f43dd1a02fc9aff8897a8becd355cdbe/src/modal/interfaces.ts#L89

src/dialog/__tests__/dialog.test.tsx
amanabiy: Yes, it was created for test coverage. I think it's best to remove it as well.

src/dialog/__tests__/dialog.test.tsx
amanabiy: Dialog follows the same first-focusable-element behavior as [Modal](https://cloudscape.design/components/modal/). Modal focuses the close button because it is the first focusable element. In Dialog, headerActions appear before the close button, so focus moves to the first header action when present, and otherwise to the close button.

In both cases, role="dialog" and aria-labelledby cause screen readers to announce the dialog role and heading together with the focused control. I think it's better to keep the behavior consistent with the modal, unless this causes problem in screen-reader.

amanabiy: doc: 3iCP0Z0Rd8Q6

src/dialog/__tests__/dialog.test.tsx
amanabiy: Yes, we always restore focus, when the prev focused element is still in DOM. This test was added for coverage. I will remove it.

style-dictionary/visual-refresh/colors.ts
amanabiy: Good point. I added it because I wanted to have the same colors and border as Item card, but I don't think changing item card should change the dialog. so it's best to isolate it.

src/dialog/__tests__/dialog.test.tsx
amanabiy: I separated this into four tests: two verify the onDismiss event for the dismiss button and Escape, and two independently verify focus restoration for each dismissal method. Focus restoration no longer shares assertions with event details.

src/dialog/__tests__/dialog.test.tsx
amanabiy: Agreed. This test duplicated the existing initial-focus and restoration coverage, so I removed it.

src/dialog/__tests__/dialog.test.tsx
amanabiy: Yes, I agree. I have added the test for this.

src/dialog/__tests__/dialog.test.tsx
amanabiy: Thank you, yes i have added a renderStatefulDialog function.

src/dialog/index.tsx
amanabiy: No, I don't have a use-case, I included it because it is optional similar to the other props. I agree, we can remove it.

src/dialog/interfaces.ts
amanabiy: Agreed, I have simplified the api doc to be `Actions displayed in the header row, before the close button.`

src/dialog/interfaces.ts
amanabiy: Agreed. I have simplified the doc to be `Content displayed in the dialog.`

src/dialog/internal.tsx
amanabiy: Thanks, removed HTMLElementConstructor and aligned focus restoration with Modal’s FocusLock logic.

src/dialog/internal.tsx
amanabiy: Removed the timeout entirely.

style-dictionary/visual-refresh/metadata/borders.ts
amanabiy: Removed.

src/dialog/internal.tsx
amanabiy: Removed useVisualRefresh and its conditional class.

src/dialog/__tests__/dialog.test.tsx
amanabiy: I removed the race-based test and kept coverage of Dialog’s supported behavior: focusing the first focusable element on mount. For reference, I observed the same behavior in Safari, Chrome, and Firefox the second action with autoFocus receives focus first, then Dialog moves focus to the first header action

style-dictionary/visual-refresh/colors.ts
```diff
@@ -125,6 +125,7 @@ const tokens: StyleDictionary.ColorsDictionary = {
   colorBackgroundAlertSuccess: '{colorBackgroundStatusSuccess}',
   colorBackgroundAlertWarning: '{colorBackgroundStatusWarning}',
   colorBackgroundDialog: '{colorBackgroundStatusInfo}',
```
amanabiy, line 127: Kept the existing public colorBackgroundDialog unchanged and added a private colorBackgroundDialogDefault token for the new Dialog component.

src/dialog/interfaces.ts
amanabiy: The reason is a string (like Modal) instead of a union. The detail is an output type we can't widen later, so a union would lock the members and we couldn't add a future dismiss reason without a breaking change. string keeps it forward compatible. New reasons can be added later, with the current values (closeButton, escape) documented in the JSDoc if needed.

That said, I'm also fine dropping reason for now since we don't have a consumer for it yet. Adding a detail property later is non breaking, so we lose nothing by waiting and adding it when a real use case defines the shape. Happy to go either way.

pan-kot: Thanks, I did not realise we use the same API in the modal already - that is a good call out. We can do the same for consistency in this case, but I'd still wait for the use cases. When dismiss button is not optional - the reason prop should not be needed.

src/dialog/__tests__/dialog.test.tsx
pan-kot: Ok, I agree on the consistency point.

pages/dialog/focus-restoration.page.tsx
```diff
@@ -0,0 +1,56 @@
...
+import Dialog from '~components/dialog';
+import SpaceBetween from '~components/space-between';
+
+import { SimplePage } from '../app/templates';
+
+const params = new URLSearchParams(window.location.hash.split('?')[1] || '');
+const programmatic = params.get('mode') === 'programmatic';
+
+export default function DialogFocusRestorationPage() {
+  const [open, setOpen] = useState(programmatic);
```
pan-kot: Take a look at how we use PageContext in other test pages. We can use it to simplify the state management in this page:

```
function DialogFocusRestorationPage() {
  const {
    urlParams: { dialogOpen = false },
    setUrlParams,
  } = useContext(AppContext as PageContext);
  // ...
}

// in test
await browser.url('#/light/dialog/focus-restoration?dialogOpen=true');
```

pages/dialog/focus-restoration.page.tsx
```diff
@@ -0,0 +1,56 @@
...
+  const [showTrigger, setShowTrigger] = useState(!programmatic);
+  const focusTargetRef = useRef<ButtonProps.Ref>(null);
+
+  const dismissDialog = () => {
+    setOpen(false);
+    window.setTimeout(() => focusTargetRef.current?.focus(), 0);
```
pan-kot: Can we avoid having setTimeout here?

What if the original target element is available, but the consumer intent is to move the focus elsewhere - will then both elements be focused in a close sequence? Can this trigger extra announcement by the screen reader?

pages/dialog/focus-restoration.page.tsx
pan-kot: nit: you can also consider combining this page with simple.page - by giving it more configuration settings, so that the same page can be used for manual and automated testing

src/dialog/internal.tsx
```diff
@@ -0,0 +1,112 @@
...
+  ...restProps
+}: InternalDialogProps) {
+  const baseProps = getBaseProps(restProps);
+  const i18n = useInternalI18n('dialog');
+  const dismissAriaLabel = i18n('i18nStrings.dismissAriaLabel', i18nStrings?.dismissAriaLabel);
+  const headerId = useUniqueId('dialog-header-');
+  const dialogRef = useRef<HTMLDivElement>(null);
+  const restoreFocusTargetRef = useRef<HTMLOrSVGElement | null>(null);
+  const [breakpoint, breakpointsRef] = useContainerBreakpoints(['xxs']);
+  const isNarrow = breakpoint === 'default';
```
pan-kot: What problem does isNarrow solve?

pan-kot: I also noticed that dismiss button alignment in narrow screens looks off.

Image

amanabiy: Thanks I have fixed this. It was a residual from my experiments.

pages/dialog/focus-restoration.page.tsx
amanabiy: Yes, we don’t need setTimeout. I updated the example to move focus when Dialog transitions from open to closed. Our [Focus management principles](https://cloudscape.design/foundation/core-principles/accessibility/focus-management-principles/) recommend returning focus to the trigger when it is still available. If it no longer exists, the builder should move focus programmatically to a logical fallback, usually the next focusable element in the DOM order. This example covers the latter case and does not override an available trigger. If a customer intentionally overrides restoration while the original trigger still exists, both elements could receive focus in sequence and cause an additional announcement as you mentioned.

pages/dialog/simple.page.tsx
```diff
@@ -0,0 +1,103 @@
...
+  const {
+    urlParams: { dialogOpen = false, removeTrigger = false },
+    setUrlParams,
+  } = useContext(AppContext as PageContext);
+  const [goal, setGoal] = useState('increase');
+  const focusTargetRef = useRef<ButtonProps.Ref>(null);
+  const wasOpenRef = useRef(dialogOpen);
+
+  useEffect(() => {
+    if (wasOpenRef.current && !dialogOpen && removeTrigger) {
```
pan-kot: Why do we need this? Can the focus be triggered from onDismiss?

amanabiy: Yes, I have updated it to be used onDismiss.

src/dialog/interfaces.ts
```diff
@@ -0,0 +1,50 @@
...
+   * An object containing all the necessary localized strings required by the component.
+   *
+   * - `dismissAriaLabel` (string) - Adds an `aria-label` to the close button.
+   * @i18n
+   */
+  i18nStrings?: DialogProps.I18nStrings;
+
+  /**
+   * Called when the user dismisses the dialog using the close button or the Escape key.
+   * Handle it by removing the dialog from the page, for example with conditional rendering.
```
pan-kot: nit: what does "for example with conditional rendering" mean in this context? Isn't it the only way to remove the dialog?

src/test-utils/dom/dialog/index.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  /**
+   * Finds the header actions slot of the dialog.
+   */
+  findHeaderActions(): ElementWrapper | null {
+    return this.findByClassName(testStyles['header-actions']);
+  }
+
+  /**
+   * Finds the header actions slot of the dialog.
```
pan-kot: This description is not accurate. I think we can also omit the descriptions as not giving more info than it is already conveyed with the method's name.

src/internal/components/focus-lock/use-focus-restore.ts
```diff
@@ -0,0 +1,47 @@
...
+          }
+          getFirstFocusable(container)?.focus();
+        }
+      } else if (restoreFocus) {
+        restoreFocusTargetRef.current?.focus();
+        restoreFocusTargetRef.current = null;
+      }
+    },
+    [autoFocus, restoreFocus]
+  );
```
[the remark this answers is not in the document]
pan-kot: That is a good find. I think for the intended purpose - we don't actually need autoFocus to be optional. That also makes the util less abstract - so I'd recommend moving it under src/dialog for now.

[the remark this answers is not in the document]
amanabiy: Agreed. I have moved it to the src/dialog and removed the autoFocus and restoreFocus, as there are no use-cases for them right now.

src/test-utils/dom/dialog/index.ts
amanabiy: Fixed. But I would prefer to keep it to keep it more consistent with the new components like item-card.

pan-kot (approved):

PR #4906, opened 2026-08-14
fix: Stop repeating the license header in every compiled stylesheet

### LINKED ISSUE #4905: [Bug]: Published CSS contains 30-50 duplicate license headers per file (26% of CSS bytes)
### Browser

_No response_

### Package version

v3.0.1345

### React version

_No response_

### Description

Every stylesheet published in `lib/components` contains the Apache license comment repeated 30-50 times.  Measured on a local `npm run build` of `c4816f9f1`:

| ---------------- | -------------------------------------------------------- |
| stylesheets      | 227                                                      |
| total CSS        | 2269 KB                                                  |
| license comments | 598 KB (**26.4%**)                                       |
| blocks per file  | median 35, max 51 (`file-token-group/styles.scoped.css`) |

Example: `popover/styles.scoped.css` is 27,483 bytes and holds 43 copies of the same 4-line comment.

### Cause

318 of the 319 `.scss` files in `src/` open with a loud CSS comment (`/* ... */`) rather than a Sass silent comment (`// ...`). Sass preserves loud comments, and `theming-build` compiles with `style: 'expanded'`, so every module a `styles.scss` entry point transitively `@use`s contributes its header to that entry's output. `src/popover/styles.scss` loads 25+ modules, hence 43 headers. Most of those modules emit no CSS rules at all: for files like `src/internal/styles/typography/constants.scss` the header is the module's entire contribution to the compiled output.

### Impact

Minifiers drop the comments, so the impact on end users is zero for apps that have a proper build pipeline.

Where this really hurts is the build and dev experience. I hit this with a Vite dev server: Each stylesheet is served as a separate request, so the dev server is blocked on waiting for 200+ assets to load. The assets are tiny, but the overhead of those requests adds up.

### Proposed fix

Keep the loud header on the 229 `styles.scss` entry points and switch the 90 partials to `//`. Sass hoists an entry file's leading loud comment above its modules' CSS (verified), so each output file keeps one header at the top and the artifacts stay compliant. The one exception is `internal/components/chart-series-details`, which uses another component's entry point and would keep two.

### Source code

_No response_

### Reproduction

_No response_

### Code of Conduct

TrevorBurnham:
### Description

Fixes #4905.

Sass copies loud comments into the compiled CSS, so each partial's license header was emitted once per entry point that `@use`s it. Every published stylesheet carried a median of 34 copies, and **580 KB of the 2305 KB of CSS was the same four lines repeated**.

The 91 partials under `src/` now declare the header with Sass line comments, which the compiler strips. The `styles.scss` entry points keep the block comment, so every output file still carries one. `.stylelintrc` enforces the right form for each using the `commentType` option added in cloudscape-design/build-tools#75.

| | before | after |
| --- | --- | --- |
| CSS, raw | 2305 KB | **1745 KB** (-24.3%) |
| CSS, gzipped | 303 KB | 298 KB (-1.8%) |
| license header bytes | 580 KB (25.2%) | 32 KB (1.8%) |
| headers, total | 5400 | 297 |
| headers per stylesheet, median | 34 | 1 |
| headers per stylesheet, max | 51 | 5 |
| stylesheets with more than one | 152 | 48 |

The gzipped win is small because gzip already deduped the identical blocks. The win is in unminified bytes, which is what dev servers serve and what the internal package ships. Minified production bundles are unaffected either way, since minifiers strip the comments already.

Two things reviewers should know:

**1. Every generated scoped class name changes.** The name is `awsui_{name}_{fileHash}_{contentHash}_{lineNumber}`, and removing comments moves both the content hash and the line numbers. No snapshot needs updating, though: the `test-utils` selectors snapshot truncates to the path-derived hash, and since #4936, keyframe names derive from stable per-keyframe hashes rather than from a hash of `src/internal/styles/global.scss`, so the design tokens and `themes` snapshots no longer move with this change either. The compiled CSS is otherwise identical, verified as described below.

**2. This does not remove every duplicate.** 48 stylesheets still carry more than one header, up to 5. That remainder is a Dart Sass behaviour, not something this repo controls: an entry point's leading comments are re-emitted once per module when the entry point and its partials share a dependency. Minimal repro on Sass 1.89.2, no Cloudscape code involved:

```scss
// _shared.scss
.shared { color: gray; }
@mixin m { color: gray; }

// _a.scss     -> @use "shared";  .a { color: red; }
// _b.scss     -> @use "shared";  .b { color: blue; }

// entry.scss
/* ENTRY HEADER */
@use "shared";
@use "a";
@use "b";
```

`ENTRY HEADER` is emitted three times. This is the shape of every Cloudscape component that has sub-partials, which is why `popover` still ends up with several headers rather than one. It accounts for roughly 6 KB and is worth a follow-up upstream; it does not block this change.

The other source named in #4905, `@cloudscape-design/component-toolkit`, is already resolved: cloudscape-design/component-toolkit#248 converted its two SCSS files and shipped in `1.0.0-beta.182`, which this branch builds against. That is why the median here reaches 1 rather than 2.

### How has this been tested?

A normalised comparison of all 229 compiled stylesheets against a build of `main` from the same `node_modules`. After neutralising license blocks and the `contentHash`/`lineNumber` segments of scoped names, **all 229 stylesheets are byte-identical**. No file was added, removed, or left empty.

<details>

#### Correctness

#### Testing

</details>

---

### DIFF
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -58,7 +58,27 @@
       "rules": {
         "property-disallowed-list": ["border", "border-radius", "border-style", "margin", "padding"],
         "csstools/use-logical": "always",
-        "@cloudscape-design/no-motion-outside-of-mixin": [true]
+        "@cloudscape-design/no-motion-outside-of-mixin": [true],
+        "@cloudscape-design/license-headers": [
+          true,
+          {
+            "header": "\n Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.\n SPDX-License-Identifier: Apache-2.0\n",
+            "commentType": "line"
+          }
+        ]
+      }
+    },
+    {
+      "files": [
+        "./src/**/styles.scss"
+      ],
+      "rules": {
+        "@cloudscape-design/license-headers": [
+          true,
+          {
+            "header": "\n Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.\n SPDX-License-Identifier: Apache-2.0\n"
+          }
+        ]
       }
     }
   ],
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -58,7 +58,28 @@
       "rules": {
         "property-disallowed-list": ["border", "border-radius", "border-style", "margin", "padding"],
         "csstools/use-logical": "always",
-        "@cloudscape-design/no-motion-outside-of-mixin": [true]
+        "@cloudscape-design/no-motion-outside-of-mixin": [true],
+        "@cloudscape-design/license-headers": [
+          true,
+          {
+            "header": "\n Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.\n SPDX-License-Identifier: Apache-2.0\n",
+            "commentType": "line"
+          }
+        ]
+      }
+    },
+    {
+      "files": [
+        "./src/**/styles.scss"
+      ],
+      "rules": {
+        "@cloudscape-design/license-headers": [
+          true,
+          {
+            "header": "\n Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.\n SPDX-License-Identifier: Apache-2.0\n",
+            "commentType": "block"
+          }
+        ]
       }
     }
   ],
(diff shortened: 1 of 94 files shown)


.stylelintrc
```diff
@@ -58,7 +58,27 @@
...
+    },
+    {
+      "files": [
+        "./src/**/styles.scss"
+      ],
+      "rules": {
+        "@cloudscape-design/license-headers": [
+          true,
+          {
+            "header": "\n Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.\n SPDX-License-Identifier: Apache-2.0\n"
```
pan-kot: nit: Should we add `"commentType": "block"` for the distinction between two usages of the rule is more explicit?

TrevorBurnham: Done in 5627044. It's behaviour-neutral since `block` is the rule's default, but I agree that it reads better.

pan-kot (dismissed):

jperals (dismissed):

pan-kot (dismissed):

jperals (approved):

pan-kot (approved):

PR #4909, opened 2026-08-17
fix: Avoid mixed-unit max() in drag handle resize area styles

Harsh-Anand-Singh:
### Description
The drag handle resize area used `inline-size: max(24px, 100%)` / `block-size: max(24px, 100%)`, introduced in a1b16ac. The CSS `max()` function with mixed units (px and %) can be misinterpreted by Sass in downstream builds, causing build failures. Same class of issue as #4882 (fixed in 2041ec8).

This replaces it with an exact CSS equivalent that uses no `max()` at all: `inline-size`/`block-size: 100%` with `min-inline-size`/`min-block-size: 24px`. Min constraints always win when the computed size is smaller, so the pointer target still expands to at least 24x24px (WCAG 2.5.8) and otherwise fills the handle. No functional change.

Related links, issue #, if available: #4882

### How has this been tested?
- `npm run quick-build` passes; verified the compiled `lib/components/internal/components/drag-handle/styles.scoped.css` contains the plain declarations with no mixed-unit `max()`
- `npx stylelint` passes on the file

<details>
#### Correctness
#### Security
#### Testing
</details>
---

### DIFF
--- a/src/internal/components/drag-handle/styles.scss
+++ b/src/internal/components/drag-handle/styles.scss
@@ -47,8 +47,11 @@
       position: absolute;
       inset-block-end: 0;
       inset-inline-end: 0;
-      inline-size: max(24px, 100%);
-      block-size: max(24px, 100%);
+      // min-*-size instead of `max(24px, 100%)`, which Sass can misinterpret due to mixed units
+      inline-size: 100%;
+      block-size: 100%;
+      min-inline-size: 24px;
+      min-block-size: 24px;
     }
 
     @include styles.with-direction('rtl') {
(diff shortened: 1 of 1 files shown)


src/internal/components/drag-handle/styles.scss
```diff
@@ -47,8 +47,11 @@
       position: absolute;
       inset-block-end: 0;
       inset-inline-end: 0;
-      inline-size: max(24px, 100%);
-      block-size: max(24px, 100%);
+      // min-*-size instead of `max(24px, 100%)`, which Sass can misinterpret due to mixed units
```
amanabiy: I suggest to remove the comment as it is contextual only to the bug and it's already fixed. I believe the code is self explanatory.

Harsh-Anand-Singh: Done.

YueyingLu (approved):

amanabiy (approved):

PR #4911, opened 2026-08-18
chore: Add skipBrowsers option to screenshot test definitions

amanabiy:
### Description
This test was originally Safari-skipped in the integration tests. The skip was dropped when it migrated to the test-definitions format, because `ScreenshotTestConfiguration` had no field to express it, and it now fails on the internal Safari runner. Adding the field back lets us restore the skip.

**Changes**
- Add optional `skipBrowsers?: string[]` to `ScreenshotTestConfiguration` (`test/definitions/types.ts`).
- Restore `skipBrowsers: ['Safari']` on the area-chart "selects correct series when navigated back from legend" test (`test/definitions/visual/area-chart.ts`).

Related links, issue #, if available: (will share link for reviewer internally)

### How has this been tested?
build passes (run before commit). Field is ignored on the Chrome-only GitHub runner; internal pipeline skips it on Safari.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/test/definitions/types.ts
+++ b/test/definitions/types.ts
@@ -10,6 +10,7 @@ export type Wrapper = ReturnType<typeof createWrapper>;
 export interface ScreenshotTestConfiguration {
   width?: number;
   height?: number;
+  skipBrowsers?: string[];
 }
 
 // 'screenshotArea' — captures the .screenshot-area element on the page.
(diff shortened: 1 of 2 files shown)


NathanZlion (approved):
Good start, let's not forget the other tests whose skipBrowser haven't been brought to Github too as a follow up.

amanabiy:
> Good start, let's not forget the other tests whose skipBrowser haven't been brought to Github too as a follow up.

The other tests are working as expected, in both internal and external. I would rather not skip running them if they are working.

NathanZlion:
> The other tests are working as expected, in both internal and external...

great, In that case no need to skip it,

PR #4927, opened 2026-08-24
fix: Contain inline overflow in expandable section boxes in one theme

jkuelz:
### Description

In one theme, expandable section content is revealed by animating `grid-template-rows` from `0fr` to `1fr`, which makes `.content-inner` a grid item. A grid item's automatic minimum size is content-based on both axes, but only `min-block-size` was zeroed. Content that manages its own inline overflow (e.g. a table's horizontal scroll area) therefore resolved `min-inline-size: auto` to its min-content width and stretched the wrapper past its grid track and outside the section's container, producing a page-level horizontal scrollbar.

This adds `min-inline-size: 0` alongside the existing `min-block-size: 0`, leaving the `overflow-x: visible` / `overflow-y: clip` pairing untouched so the block axis can still collapse while negative-margin backgrounds (side navigation hover/active/focus states) keep bleeding past the inline edges.

The same pairing is applied to .header-content, which side navigation also makes a grid item via `.section >  :first-child { display: grid }`. `.header-actions-wrapper` and `.header-icon-end > .header-content` already zero the inline axis; this leaves the start-icon layout consistent with them rather than being the one header box that can outgrow its container.

Related links, issue #, if available: AWSUI-62232

### How has this been tested?
Manually, in a local one-theme dev build.

Can verify with `INCLUDE_ONE_THEME=true npm start`, then open an expandable section containing a wide table in one theme and confirm no horizontal page scrollbar appears during or after expansion, and the table inside expandable section does not overflow container.

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/expandable-section/styles.scss
+++ b/src/expandable-section/styles.scss
@@ -352,12 +352,14 @@ $nav-caret-pad-one-theme: 12px; // 24px minus the one-theme x-small (12px) icon
 .content-inner {
   /* this comment keeps the class in the style sheet */
   @include theming.one-theme-only {
-    // overflow: visible lets negative-margin backgrounds (e.g. side-navigation
-    // active/hover/focus states) bleed past the inline edges.
-    overflow-x: visible;
-    // overflow: clip (not hidden, which would coerce x-axis to auto) still lets the grid-row track collapse.
+    // Grid item of .content above. `clip` rather than `hidden`: `hidden` coerces the inline axis
+    // to `auto`, making this a scroll container on both axes, where `clip` leaves the inline axis
+    // `visible` so content that paints outside the box (focus rings, inline popovers) survives.
     overflow-y: clip;
     min-block-size: 0;
+    // A grid item's automatic minimum size is content-based on this axis too, so content that
+    // manages its own inline overflow (e.g. a table) would stretch the box out of the container.
+    min-inline-size: 0;
     opacity: 0;
 
     // HIDE (collapsed state): quick opacity collapse, no delay.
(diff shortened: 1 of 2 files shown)


src/expandable-section/styles.scss
```diff
@@ -545,8 +549,13 @@ $nav-caret-pad-one-theme: 12px; // 24px minus the one-theme x-small (12px) icon
 // uses grid-template-rows: 0fr (side-navigation section header collapse).
 // Must remain global (not scoped to .header-icon-end) because the side-nav
 // collapse targets the header wrapper regardless of icon position.
+// min-inline-size: 0 zeroes the other axis of that automatic minimum size, so wide
+// content cannot stretch this box past the track. The end-icon layout below and
+// .header-actions-wrapper already do this; without it the start-icon layout is the
+// only header box that can outgrow its container.
 .header-content {
   min-block-size: 0;
+  min-inline-size: 0;
```
mxschll, line 553: This is a fix for one theme, should we scope this rule to one theme?

src/expandable-section/styles.scss
mxschll: Would it be possible to get rid of that line and instead move the fix to the sidenav itself?

mxschll, line 1: @jkuelz can we extend our dev pages with this scenario and add a screenshot test?

mxschll: I also wonder now if custom content is negatively impact by that logic in one theme? Is it always expected that the overflow is visible?

src/expandable-section/styles.scss
jkuelz: This part is needed because side nav uses `display: grid`, which is not gated to one-theme; side-nav only gates the transitions. It is not gated because the `collapsed` prop is available in VR as well, and rather than having two divergent implementations for the same behavior, this allows for one for both.

src/expandable-section/styles.scss
jkuelz: Good news on the first half: This line can actually be removed completely, the real fix is adding `min-inline-size: 0`.

On moving it to side-nav: this box belongs to the expandable section, not the side-nav. `.content` sets `display: grid;` in the section above (line 332), `.content-inner` is expandable section's grid item, and the overflow/min-size rules belong next to the track that creates the constraint. 

The actual cause of AWSUI-62232 was the missing minimum size reset. For a grid item, `min-*-size: auto` is content-based on both axes, unlike for a flex item, where it's main-axis only, so content with a large min-content width (e.g. a table) stretched this box out of its container. That's why the `min-inline-size: 0` addition fixes the issue.

Regarding whether overflow visible is expected: I would think so, otherwise things like focus outlines, dropdowns, popovers, tooltips, etc. could be clipped. In the settled expanded state, both axes compute to `visible`, same as VR.

mxschll (approved):

PR #4933, opened 2026-08-25
fix: Align loading spinner with status icons in Steps

ywyyu17:
### Description

In the Steps component (vertical orientation, One Theme), the loading spinner and the header text didn't align with the other steps' status icons (e.g., the ✓ success icon). The spinner sat slightly off and its header text started at a different x-position.

Root cause: the status indicator renders its icon inline with the text, so alignment depends on the icon glyph occupying a consistent size. In One Theme, status icons render at x-small (12px), but the spinner had no x-small size — its smallest was normal (16px). So next to 12px icons the spinner was oversized and misaligned. Additionally, the loading entry never passed size to the spinner at all (<InternalSpinner />), so it always defaulted to normal.

### How has this been tested?

Screenshot tests

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/spinner/mixins.scss
+++ b/src/spinner/mixins.scss
@@ -10,7 +10,7 @@
 $_spinner-sizes: (
   (
     name: 'x-small',
-    size: 0.9 * styles.$base-size,
+    size: styles.$base-size,
     iconSizeMap: 12px,
     supportedLineHeight: awsui.$line-height-body-s,
   ),
--- a/src/spinner/mixins.scss
+++ b/src/spinner/mixins.scss
@@ -9,6 +9,12 @@
 
 $_spinner-sizes: (
   (
+    name: 'x-small',
+    size: 0.9 * styles.$base-size,
+    iconSizeMap: 12px,
+    supportedLineHeight: awsui.$line-height-body-s,
+  ),
+  (
     name: 'normal',
     size: 1.2 * styles.$base-size,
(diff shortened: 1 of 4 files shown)


mxschll (dismissed):

src/spinner/mixins.scss
```diff
@@ -8,6 +8,12 @@
 @use '../internal/styles/tokens' as awsui;
 
 $_spinner-sizes: (
+  (
+    name: 'x-small',
+    size: 0.9 * styles.$base-size,
```
at-susie: Why 0.9? I just want to know the rationale

ywyyu17: The rationale is visual alignment. The 0.9 * $base-size is chosen to make the spinner render at the same visual size as the 12px x-small status icons

mxschll: @at-susie should we have a better rule for the spinner or its own tokens so its understandable what is happening here?

ywyyu17: Thanks @at-susie @mxschll :woman-bowing:  I removed 0.9, make it visually more aligned with x-small icon size.

mxschll (approved):

SpyZzey (approved):

src/spinner/mixins.scss
at-susie: Looks great! Approving

PR #4941, opened 2026-08-27
feat: wire motion trigger into components

mxschll:
Adds motion trigger and target attributes to components.

Context: `b9VcPdMJ5zRn` / `NVoiJ1X1df9a`.

### Description

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/top-navigation/parts/overflow-menu/menu-item.tsx
```diff
@@ -109,7 +111,7 @@ const NavigationItem = forwardRef(
       <ButtonItem
         ref={ref}
         startIcon={startIcon}
-        endIcon={<InternalIcon name="angle-right" />}
+        endIcon={<InternalIcon name="angle-right" nativeAttributes={{ 'data-awsui-motion-target': '' }} />}
```
mxschll, line 114: We only want to play animation for the "angle-right" and not custom provided icons, as hovering + focus is coupled with the angle icon action.

src/top-navigation/parts/overflow-menu/menu-item.tsx
```diff
@@ -139,11 +141,12 @@ const ExpandableItem: React.FC<
         className={clsx(styles['overflow-menu-control'], styles['overflow-menu-control-expandable-menu-trigger'])}
         onClick={() => setExpanded(value => !value)}
         aria-expanded={expanded}
+        data-awsui-motion-trigger="hover"
```
pan-kot: the expand toggle already has animation - why do we need these?

src/table/header-cell/index.tsx
```diff
@@ -297,6 +298,7 @@ export function TableHeaderCell<ItemType>({
                     'columnDefinitions.editConfig.editIconAriaLabel',
                     column.editConfig?.editIconAriaLabel
                   )}
+                  nativeAttributes={{ 'data-awsui-motion-target': '' }}
```
pan-kot: The table header is clickable when it is sortable. Clicking on the header causes its sorting direction to change. I believe we should not make the edit icon animate - as it creates confusion.

src/table/body-cell/index.tsx
```diff
@@ -126,9 +126,10 @@ function TableCellEditable<ItemType>({
               ref={editActivateRef}
               onClick={!isEditing && isExpandableColumn ? onEditStart : undefined}
               tabIndex={editActivateTabIndex}
+              data-awsui-motion-trigger="hover"
             >
               <span className={styles['body-cell-editor-icon']}>
-                <Icon name="edit" />
+                <Icon name="edit" nativeAttributes={{ 'data-awsui-motion-target': '' }} />
```
pan-kot: When a cell is editable - then the entire cell is hover-able and clickable. The semantic button in this case should not act as trigger. Instead - the cell should. Alternatively - we can ignore this completely, as edit button is only visible on hover anyways.

src/internal/components/drag-handle/button.tsx
```diff
@@ -80,6 +80,9 @@ const DragHandleButton = forwardRef(
         onPointerDown={onPointerDown}
         onClick={onClick}
         onKeyDown={onKeyDown}
+        {...(variant === 'drag-indicator'
```
pan-kot, line 83: Why do we only support one variant here?

src/table/header-cell/index.tsx
mxschll: I think it's valid that the icon has the animation here as its an interactive element, similar like a button. But I agree that when I tried it out, it felt a bit odd. Removing it as it can be added at a later stage again if necessary.

src/button-dropdown/category-elements/expandable-category-element.tsx
```diff
@@ -139,6 +140,7 @@ const ExpandableCategoryElement = ({
             <InternalIcon
               name={isOneTheme ? 'angle-down' : 'caret-down-filled'}
               size={isOneTheme ? 'x-small' : 'normal'}
+              nativeAttributes={{ 'data-awsui-motion-target': '' }}
```
pan-kot, line 143: Should we also animate consumer-provided icons and external link icon in dropdown items?

src/internal/components/drag-handle/button.tsx
mxschll: The other ones are just lines and they would fall back to the default shrinking animation which is almost not perceptible for these icons which didn't make sense to animate. The drag-indicator is also the only one with it's custom animation.

src/button-dropdown/category-elements/expandable-category-element.tsx
```diff
@@ -110,6 +110,7 @@ const ExpandableCategoryElement = ({
       // tabindex=-1 so we can focus them when necessary.
       tabIndex={filteringEnabled ? -1 : highlighted ? 0 : -1}
       ref={triggerRef}
+      data-awsui-motion-trigger="hover"
```
pan-kot, line 113: I noticed that expand toggles in button dropdown category items also animate when the item is highlighted with the keyboard. How does this work?

src/button/internal.tsx
```diff
@@ -236,6 +236,8 @@ export const InternalButton = React.forwardRef(
       title: __title ?? ariaLabel,
       className: buttonClass,
       onClick: handleClick,
+      'data-awsui-motion-trigger': 'hover',
```
pan-kot, line 239: What these are for: is it for internal buttons only? Why only hover trigger?

src/button-dropdown/category-elements/expandable-category-element.tsx
mxschll: Semantically, focus can never cover more area than hover. Hovering any part of the region hovers the region, while focus lands on exactly one element, which is either the region itself or a descendant of it. Thats why hover is implicitly hover + focus-visible. But we can move the  focus trigger to the children if necessary like in the action card.

We could write it explicit everywhere: `data-awsui-motion-trigger="hover focus" but I decided not to because "hover focus" implies "hover" alone is a valid state, but it never is.

src/button/internal.tsx
mxschll: As per comment above: "hover" implicitly is "hover + focus" because it enforces focus parity. The focus token only exists for designating the focus owner when it's nested inside the hover region.

src/button-dropdown/category-elements/expandable-category-element.tsx
mxschll: Good point 👍

src/button-dropdown/category-elements/expandable-category-element.tsx
pan-kot: Ok, it is clear now - thanks 👍

pan-kot (approved):

PR #4942, opened 2026-08-27
refactor: apply usePortalModeClasses to remaining portaled components

Harsh-Anand-Singh:
## Description

This draft PR applies `usePortalModeClasses` to the remaining portaled components so portal-mode styling follows the same shared pattern across the codebase.

Some components already preserve portal-mode classes correctly when rendering through portals, while others still rely on component-specific handling. This change aligns those implementations and closes the remaining propagation gaps.

## Changes

- apply `usePortalModeClasses` to remaining portaled components
- replace inconsistent component-level portal-mode handling with the shared hook
- add or update tests covering portal-mode class propagation

## Why

This keeps portal behavior consistent across components and reduces the chance of styling mismatches when content is rendered outside the normal DOM tree.

## Testing

- updated relevant tests for affected portaled components
- verified portal-mode classes propagate as expected in portal-rendered content

### DIFF
--- a/src/internal/components/sortable-area/index.tsx
+++ b/src/internal/components/sortable-area/index.tsx
@@ -10,6 +10,7 @@ import clsx from 'clsx';
 import { Portal } from '@cloudscape-design/component-toolkit/internal';
 
 import { fireNonCancelableEvent } from '../../events';
+import { usePortalModeClasses } from '../../hooks/use-portal-mode-classes';
 import { joinStrings } from '../../utils/strings';
 import { SortableAreaProps } from './interfaces';
 import { EventName } from './keyboard-sensor/utilities/events';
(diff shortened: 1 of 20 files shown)


src/internal/components/sortable-area/index.tsx
```diff
@@ -76,6 +79,8 @@ export default function SortableArea<Item>({
             renderItem={renderItem}
             onKeyDown={handleKeyDown}
             dragHandleAriaLabel={i18nStrings?.dragHandleAriaLabel}
+            // All items in one sortable list share the same visual context, so the first item is enough as the anchor.
+            portalModeRef={index === 0 ? portalModeRef : undefined}
```
avinashbot, line 83: A little janky, would a wrapper not work here?

avinashbot (changes requested):
Just one comment, but the hidden spans do make me a little nervous (specifically because of situations like `:empty`). Have you run it through visual tests in the pipeline?

PR #4962, opened 2026-09-02
feat: public consumer channel for global breadcrumbs [RFC]

ajsuvarna6:
**Not for merge.** This is a prototype to make a design question concrete, and I would value maintainer direction before it goes any further.

## The problem

A chrome surface that owns the page header can live in a separate React root and bundle from App Layout, so breadcrumbs cannot be passed as props — the data has to travel at runtime.

`BreadcrumbsController` already has the right shape: `<BreadcrumbGroup>` publishes its own props to a shared registry. But it never publishes outside App Layout, takes a single subscriber, and is internal. Note App Layout itself cannot publish, since it receives a React element it cannot read the items out of.

## What this does

- `BreadcrumbsController` gains an external-consumer channel with last-value replay (`onBreadcrumbsChange`), a `hasExternalConsumer()` signal, and `installPublic()`.
- App Layout auto-yields — null-driven while an external consumer is subscribed, so breadcrumbs are never drawn twice.
- `slots.tsx` moves `ownBreadcrumbs` outside `BreadcrumbsSlotContext` so a slot-passed `BreadcrumbGroup` self-registers. `discoveredBreadcrumbs` stays inside the provider to keep the re-registration loop guard. `isInToolbar` gates registration only, so rendered DOM is unchanged.

Three demo pages cover one consumer, several App Layouts in one React root, and several React roots via iframes.

## The question I am actually asking

**Ownership is inferred from subscriber presence.** That is the weak part. It cannot express "I own the header but have nothing to draw yet", and any subscriber suppresses App Layout — including one that only wants to observe.

I also built an explicit `claimRendering()` / `release()` variant and reverted it in favour of this simpler shape. Happy to bring it back if you prefer that direction. One finding from it: a zero-argument claim cannot self-enforce a version gate, because the plugin installer fills each key only if absent, so the first-loaded bundle owns the method and the check compares a controller against itself. A gate needs a caller-side comparison against an exposed version.

**The change most needing your input** is `slots.tsx`: letting a slot-passed `BreadcrumbGroup` register loosens a deliberate framework guard. A passing test suite is not agreement that this should be allowed.

## Also not covered

Re-registration is suppressed with the existing internal `__disableGlobalization` flag rather than a published marker, so it is unverified across two separately bundled copies. Registrations are an array rather than a stack, so an outgoing group unmounting after its replacement can still clear a live entry. Contention between multiple consumers is last-writer-wins.

## Verification

`tsc` clean. App Layout and breadcrumbs suites pass — 1730 tests, 55 suites (`TZ=UTC` required). The three demos were exercised in a browser, including the cross-iframe case.


src/internal/plugins/api.ts
georgylobko, line 1: Use widget plugin API instead: src/internal/plugins/widget

pages/app-layout/global-nav-breadcrumbs-multi-instance.page.tsx
```diff
@@ -0,0 +1,158 @@
...
+                  </Button>
+                  <Button data-testid="toggle-iframe-beta" onClick={() => setBetaMounted(mounted => !mounted)}>
+                    {betaMounted ? 'Unmount' : 'Mount'} Beta iframe
+                  </Button>
+                </SpaceBetween>
+              </SpaceBetween>
+            </Container>
+
+            {alphaMounted && <IframeInstance id="iframe-alpha" label="React root #2" AppComponent={AlphaApp} />}
+            {betaMounted && <IframeInstance id="iframe-beta" label="React root #3" AppComponent={BetaApp} />}
```
georgylobko, line 152: It could also be useful to check how it behaves with multiple hidden iframe instances, as in this example: `pages/app-layout/multi-layout-with-hidden-instances-iframe.page.tsx`

src/app-layout/visual-refresh-toolbar/skeleton/slots.tsx
```diff
@@ -55,17 +55,22 @@ export function BreadcrumbsSlot({ ownBreadcrumbs, discoveredBreadcrumbs }: Bread
   const isSSR = typeof window === 'undefined';
 
   return (
-    <BreadcrumbsSlotContext.Provider value={breadcrumbsSlotContextValue}>
+    <>
+      {/* Render slot-provided breadcrumbs outside the toolbar context so a passed <BreadcrumbGroup>
+          self-registers with the global breadcrumbs channel. This lets an external host (e.g. the
+          Global Navigation header) consume them; the component hides itself once a consumer takes over. */}
       <div className={styles['breadcrumbs-own']}>{ownBreadcrumbs}</div>
```
georgylobko, line 59: This change will break existing experience. The breadcrumb group component needs to have access to this context here: https://github.com/cloudscape-design/components/blob/main/src/breadcrumb-group/index.tsx#L19

PR #4971, opened 2026-09-03
feat: allow skipping secondary themes in themed builds

Harsh-Anand-Singh:
## Summary
- add an optional `includeSecondaryThemes` parameter to `src-themeable/buildThemedComponents`
- default it to `true`, so existing consumers are unchanged
- when set to `false`, the wrapper forwards the preset with `secondary: []`
- this lets consumers build a themed package without generating secondary theme outputs

### DIFF
--- a/src-themeable/theming.ts
+++ b/src-themeable/theming.ts
@@ -16,12 +16,21 @@ export interface BuildThemedComponentsParams {
   theme: Theme;
   outputDir: string;
   baseThemeId?: string;
+  includeSecondaryThemes?: boolean;
 }
 
-export function buildThemedComponents({ theme, outputDir, baseThemeId }: BuildThemedComponentsParams): Promise<void> {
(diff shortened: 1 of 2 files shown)


src-themeable/theming.ts
```diff
@@ -16,12 +16,21 @@ export interface BuildThemedComponentsParams {
...
   baseThemeId?: string;
+  includeSecondaryThemes?: boolean;
 }
 
-export function buildThemedComponents({ theme, outputDir, baseThemeId }: BuildThemedComponentsParams): Promise<void> {
+export function buildThemedComponents({
+  theme,
+  outputDir,
+  baseThemeId,
+  includeSecondaryThemes = true,
```
pan-kot, line 26: We should not make public API changes like that without good justification.

pan-kot (changes requested):

PR #4980, opened 2026-09-08
fix: icon-only split-trigger inline padding to space-static-xs in one…

ywyyu17:
…-theme

### Description

Adds a one-theme-only override for the icon-only trigger button in the button-dropdown **split trigger**, setting its left/right padding to the`$space-static-xs` token (8px).

### How has this been tested?

- Unit test
- Screenshot test

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/button-dropdown/styles.scss
```diff
@@ -88,6 +88,10 @@ $dropdown-trigger-icon-offset: 2px;
         &.visual-refresh {
           padding-inline-start: calc(#{awsui.$space-s} - #{$dropdown-trigger-icon-offset});
         }
+
+        @include theming.one-theme-only {
```
jperals, line 92: Is the same value to be used for the button component, [here](https://github.com/cloudscape-design/components/blob/main/src/button/styles.scss#L161-L162)?

If so, it might be better to define a custom value for the `spaceButtonIconOnlyHorizontal` in the One Theme style dictionary.

ywyyu17: Thank you so much, @jperals!  These aren't the same value.  After discussing with @at-susie, we 
agree spaceButtonIconOnlyHorizontal is the better choice and will keep it at 6px, This change is intentionally scoped to the button-dropdown main action only, with no impact on other components.

at-susie (approved):

PR #4986, opened 2026-09-09
feat: Add selector-based full base theme generation

avinashbot:
### Description

Developer-facing integration of https://github.com/cloudscape-design/theming-core/pull/208. ~~Tests will fail until that PR is merged, but this is still reviewable.~~ Now passing, ready ro review!

### How has this been tested?

Tested locally. This also adds the new theme to the dev pages (add `?theme=core-update` to the url).

<details>

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/theming/index.ts
+++ b/src/theming/index.ts
@@ -8,34 +8,40 @@ import {
 
 import { preset, TypedOverride } from '../internal/generated/theming';
 
+const VISUAL_REFRESH_THEME_ID = 'visual-refresh';
+
 export type Theme = TypedOverride;
 export interface ApplyThemeParams {
   theme: Theme;
   baseThemeId?: string;
+  selector?: string;
 }
 
 export interface ApplyThemeResult {
   reset: () => void;
 }
 
-export function applyTheme({ theme, baseThemeId }: ApplyThemeParams): ApplyThemeResult {
+export function applyTheme({ theme, baseThemeId, selector }: ApplyThemeParams): ApplyThemeResult {
   return coreApplyTheme({
     override: theme,
     preset,
-    baseThemeId,
+    selector,
+    baseThemeId: baseThemeId ?? (selector ? VISUAL_REFRESH_THEME_ID : undefined),
   });
 }
 
 export interface GenerateThemeStylesheetParams {
   theme: Theme;
   baseThemeId?: string;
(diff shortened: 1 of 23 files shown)


src/theming/index.ts
```diff
@@ -8,34 +8,40 @@ import {
...
 }
 
-export function applyTheme({ theme, baseThemeId }: ApplyThemeParams): ApplyThemeResult {
+export function applyTheme({ theme, baseThemeId, selector }: ApplyThemeParams): ApplyThemeResult {
   return coreApplyTheme({
     override: theme,
     preset,
-    baseThemeId,
+    selector,
+    baseThemeId: baseThemeId ?? (selector ? VISUAL_REFRESH_THEME_ID : undefined),
```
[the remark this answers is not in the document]
avinashbot: This is fine. The check is specifically around whether an argument for this was passed at all.

style-dictionary/core-update/index.ts
```diff
@@ -0,0 +1,72 @@
...
+  createColorMode('.awsui-dark-mode'),
+  createDensityMode('.awsui-compact-mode'),
+  createMotionMode('.awsui-motion-disabled'),
+];
+
+// core-update is an override on top of visual-refresh. Each core-update category module
+// is self-contained: it merges the core-update overrides onto the visual-refresh base
+// and exports the complete category. Categories that core-update does not customize
+// (color-charts, color-severity, motion, shadows) are taken from visual-refresh
+// directly. color-palette comes first so the seeded primary reference tokens are
```
jperals, line 28: > color-palette comes first so the seeded primary reference tokens are registered before any token resolves against them.

Is this the reason why we use async imports here, to guarantee the order? Could you expand on what the problem would be if using static imports instead?

avinashbot: No idea, I'm just following the pattern we define in all the other style-dictionary themes :)

avinashbot: Ah, specifically referring to that comment, it's not about the imports, it's about the array order of tokens. I don't know if anything would change if we changed it to be static imports, other than adding about ten lines of static import code.

jperals (approved):

PR #4995, opened 2026-09-10
fix: Disabled toggle button hover state

SpyZzey:
#4621 introduced an issue, which caused the toggle button to have a hover state when disabled. This PR fixes it by gating the normal variant's `:hover` styles behind `:not(:disabled):not([aria-disabled='true'])`, so the hover background and border are no longer applied when the button is disabled. This matches the base button behavior, where the disabled styling takes precedence over hover.

Related: [AWSUI-62272]

https://github.com/user-attachments/assets/d5e4ffa5-f630-4d1e-ac21-ef61ac5cd4d9

### DIFF
--- a/pages/toggle-button/test-hover.page.tsx
+++ b/pages/toggle-button/test-hover.page.tsx
@@ -17,13 +17,45 @@ const permutations = createPermutations<ToggleButtonProps>([
     pressed: [false, true],
   },
   {
+    variant: ['normal'],
+    children: ['Favorite'],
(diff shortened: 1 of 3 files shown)


pan-kot (approved):

mxschll:
Can we add a screenshot test for this?

SpyZzey:
> Can we add a screenshot test for this?

We have an integration test, which should be sufficient to test this bug. But I guess in general, we could have a screenshot test for disable button hover.

PR #4996, opened 2026-09-10
feat: Add support for disabledReason for inline link buttons

vikaskale2722:
### Description

`disabledReason` only worked on the `normal`, `primary` and `icon` button variants. If we passed it to `link`, `inline-link` or `inline-icon`, the tooltip did not work, nor any reason/warnings exposed to screen readers. This code change makes it work on all the six variants.

The whole fix is one condition in `src/button/internal.tsx`, adding the three missing variants to the gate. Everything downstream already keyed off the `isDisabledWithReason` flag rather than the variant, so nothing else needed touching.

**One thing reviewers should know:** when `disabledReason` is set, the button drops the native `disabled` attribute and uses `aria-disabled` instead. That's not new, it's been the behavior for `normal`/`primary`/`icon` all along, because a natively disabled button can't be focused or hovered, so the tooltip could never be reached. This PR just extends it to three more variants. Plain `<Button disabled />` is untouched.

Worth calling out in the release notes, since `toBeDisabled()` will stop matching for these variants (test-utils `isDisabled()` handles both).

Two components have a side-effect as they render Button as their trigger and forward the prop -
1. `ButtonDropdown` with `variant="inline-icon"`
2. `CopyToClipboard` with `variant="inline"`
Both combinations were silently inert before. I added tests for them so it doesn't quietly regress later.

Also fixed the docs: `Button`'s JSDoc claimed only `link` was unsupported, which was already wrong, and `CopyToClipboard`'s correctly excluded `inline` and would have become wrong. Both are fixed, and the documenter snapshot is regenerated since that text feeds the published API docs.

I deleted the old `disabled with reason ignored for %s variant (not applicable)` test rather than inverting it, it asserted `tabIndex="-1"` for the exact case that now yields `"0"`. Both behaviors it covered still have tests elsewhere.

Open question: the [disabled and read-only states](https://cloudscape.design/patterns/general/disabled-and-read-only-states/) pattern still lists only primary, normal and icon as supporting this tooltip. So this needs a design call as well as a code review, and the pattern page should be updated if we go ahead.

Related links: Issue #60908

### How has this been tested?

Full build and the complete unit test.

On the unit side I widened the existing `describe.each` in `button.test.tsx` from 3 variants to 6, which runs its 14 assertions across both the `<button>` and `<a>` branches. Added `inline-icon` to ButtonDropdown's variant list, and turned CopyToClipboard's disabled-reason test into a `test.each` over all three of its variants.

To Test :
1. `npm start`
2. `#/light/button/disabled-reason`, `#/light/button-dropdown/disabled-reason` and `#/light/copy-to-clipboard/disabled-reason` (new page)

The following items are to be evaluated by the author(s) and the reviewer(s).

#### Correctness

#### Security

#### Testing

</details>

---

### DIFF
--- a/src/button/internal.tsx
+++ b/src/button/internal.tsx
@@ -123,7 +123,14 @@ export const InternalButton = React.forwardRef(
     const target = targetOverride ?? (external ? '_blank' : undefined);
     const isNotInteractive = loading || disabled;
     const isDisabledWithReason =
-      (variant === 'normal' || variant === 'primary' || variant === 'icon') && !!disabledReason && disabled;
+      (variant === 'normal' ||
+        variant === 'primary' ||
+        variant === 'icon' ||
+        variant === 'inline-icon' ||
+        variant === 'inline-link' ||
+        variant === 'link') &&
+      !!disabledReason &&
+      disabled;
 
     const hasAriaDisabled = (loading && !disabled) || (disabled && __focusable) || isDisabledWithReason;
     const shouldHaveContent =
--- a/pages/button-dropdown/disabled-reason.page.tsx
+++ b/pages/button-dropdown/disabled-reason.page.tsx
@@ -176,6 +176,15 @@ export default function DescriptionPage() {
           />
         </div>
         <div style={{ float: isRightAligned ? 'right' : undefined, marginBottom: '100px' }}>
+          <ButtonDropdown
+            items={actionsItems}
+            ariaLabel="Instance actions"
+            variant="inline-icon"
+            disabled={true}
+            disabledReason="disabled reason"
+          />
+        </div>
+        <div style={{ float: isRightAligned ? 'right' : undefined, marginBottom: '100px' }}>
           <ButtonDropdown items={selectableGroupItems} data-testid="buttonDropdownSelectableItems">
             Selectable example
           </ButtonDropdown>
--- a/pages/button/disabled-reason.page.tsx
+++ b/pages/button/disabled-reason.page.tsx
@@ -33,6 +33,19 @@ export default function ButtonsScenario() {
           iconName="star"
           ariaLabel="Disabled reason icon button"
         />
+        <Button variant="link" disabled={true} disabledReason="disabled reason">
+          Link
+        </Button>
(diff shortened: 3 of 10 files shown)


vikaskale2722:
Dry run build executed successfully (ID - 8021940763)

src/button/internal.tsx
```diff
@@ -123,7 +123,14 @@ export const InternalButton = React.forwardRef(
...
     const isDisabledWithReason =
-      (variant === 'normal' || variant === 'primary' || variant === 'icon') && !!disabledReason && disabled;
+      (variant === 'normal' ||
+        variant === 'primary' ||
+        variant === 'icon' ||
+        variant === 'inline-icon' ||
+        variant === 'inline-link' ||
+        variant === 'link') &&
+      !!disabledReason &&
+      disabled;
```
avinashbot, line 133: If we're covering all variants, can't we just get rid of this section instead?

```suggestion
    const isDisabledWithReason = !!disabledReason && disabled;
```

vikaskale2722: but those six are only the public variants. InternalButtonProps adds five more: flashbar-icon, breadcrumb-group, menu-trigger, modal-dismiss, inline-icon-pointer-target. The check keeps disabledReason away from those.
It matters because isDisabledWithReason drops the native disabled attribute in favor of aria-disabled, so the button stays focusable to show its tooltip. None of those five passes disabledReason. But it's an allowlist, so anything new stays excluded until we add it deliberately. Input already passes disabled to its inline-icon-pointer-target button; if it ever passes a reason the same way, that button silently becomes focusable with a tooltip.

We can add a bit more context and use matching the array style used by shouldHaveContent below:

```
// disabledReason is only supported on the public Button variants. Internal-only
// variants stay natively disabled instead of becoming aria-disabled and focusable.
const isDisabledWithReason =
  ['normal', 'primary', 'icon', 'inline-icon', 'inline-link', 'link'].indexOf(variant) !== -1 &&
  !!disabledReason &&
  disabled;
```

avinashbot (approved):

PR #5006, opened 2026-09-14
chore: Fix border radii bug for one-theme

at-susie:
### Description
This PR fixes a border radii visual bug for the hover state of dropdown item in one theme.
Before this change the 4 corners are visually cut off when hovering over / keyboard-focusing on the top or bottom item.
image (24)

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/button-dropdown/category-elements/styles.scss
```diff
@@ -60,12 +60,13 @@
...
       border-block: awsui.$border-item-width solid awsui.$color-border-dropdown-item-hover;
       border-inline: awsui.$border-item-width solid awsui.$color-border-dropdown-item-hover;
-      border-start-start-radius: awsui.$border-radius-item;
-      border-start-end-radius: awsui.$border-radius-item;
-      border-end-start-radius: awsui.$border-radius-item;
-      border-end-end-radius: awsui.$border-radius-item;
+      @include styles.logical-radius(awsui.$border-radius-item);
       z-index: 2;
 
+      @include theming.one-theme-only {
```
mxschll: Why do we need to wrap this in the one theme gate? We should fix this issue globally, so if we change the border radius in a future theme (lets say core?) we don't have that issue again.

at-susie: That makes sense. I changed it to be based on the `border-radius-dropdown` which is exactly what we want for the border radius size for the hovered items

src/internal/styles/utils/mixins.scss
```diff
@@ -42,6 +42,13 @@
   text-overflow: ellipsis;
 }
 
+@mixin logical-radius($radius) {
```
mxschll, line 45: Nice! We should've added this way earlier :)

mxschll (approved):

PR #5007, opened 2026-09-14
chore: Internal style api v2 for Checkbox

vvaliyev:
### Description

API proposal: dvPFIMQepOVO

Note: the changes are added internally, the public API of the component stays unchanged just yet.

### How has this been tested?

- Unit tests
- Manual testing

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/internal/components/checkbox-icon/styles.scss
```diff
@@ -5,6 +5,43 @@
 
 @use '../../styles' as styles;
 @use '../../styles/tokens' as awsui;
+@use '../../styles/style-api' as style-api;
+
+$_control-tokens: style-api.resolve((color, background-color, border-color), carrier);
+
+@function _value($tokens, $token, $fallback) {
```
pan-kot: Can we modify the existing value() util to also check for $tokens presence - so that the _value() one is not needed?

pan-kot: Also applicable to https://github.com/cloudscape-design/components/pull/5008/changes#diff-760fbe24e0a466cd462448a2ef6f7dfdfd8083f66d0e23046669ab11476ac03cR14

src/internal/components/checkbox-icon/styles.scss
```diff
@@ -14,35 +51,31 @@
...
-      stroke: awsui.$color-foreground-control-disabled;
+    @include line-colors {
+      stroke-width: 2;
+      fill: none;
     }
-    &-readonly {
-      stroke: awsui.$color-foreground-control-read-only;
+  }
+
+  @if style-api.enabled() {
```
pan-kot: what is the purpose of this block?

src/internal/components/checkbox-icon/index.tsx
```diff
@@ -14,6 +14,7 @@ interface CheckboxIconProps extends BaseComponentProps {
   indeterminate?: boolean;
   disabled?: boolean;
   readOnly?: boolean;
+  styleApiEnabled?: boolean;
```
pan-kot: Why do we need this prop?

pages/checkbox/style-api-v2.scss
```diff
@@ -0,0 +1,52 @@
...
+}
+
+.checkbox-control.checkbox-control-readonly {
+  --awsui-style-background-color: #{palette.$surface-alt};
+  --awsui-style-border-color: #{palette.$border};
+  --awsui-style-color: #{palette.$text-muted};
+}
+
+.checkbox-label {
+  --awsui-style-color: #{palette.$text};
```
pan-kot, line 39: Can we also imitate label/description style that is conditional from input's state (checked, indeterminate)? I think this might warrant a `root` selector or some other changes. I think this is not a requirement for the initial version, but it is worth probing to make sure the current api holds.

src/internal/components/checkbox-icon/styles.scss
vvaliyev: Sure, done.

pages/checkbox/style-api-v2.scss
vvaliyev: Since checked is controlled, the consumer can just switch the label/description class from props, same as we do for disabled/readOnly. Adding that to the test page.

src/internal/components/checkbox-icon/index.tsx
vvaliyev: Dropped the prop and the block as discussed offline, the icon reads the carriers directly now.

src/internal/components/checkbox-icon/styles.scss
vvaliyev: Dropped the prop and the block as discussed offline, the icon reads the carriers directly now.

src/checkbox/internal.tsx
```diff
@@ -87,13 +96,19 @@ const InternalCheckbox = React.forwardRef<CheckboxProps.Ref, InternalProps>(
       <AbstractSwitch
         {...baseProps}
         className={clsx(styles.root, baseProps.className)}
-        controlClassName={styles['checkbox-control']}
+        controlClassName={clsx(styleClassNames?.control, styles['checkbox-control'])}
```
pan-kot, line 99: nit: should we refactor abstract switch to take `classNames={root,control,label,description}`?

pan-kot (approved):

PR #5010, opened 2026-09-14
feat: Add travelling active-tab indicator for Tabs in One Theme

ywyyu17:
### Behavior
A single indicator travels from the origin tab to the destination, morphing width to fit the target label along the way. Duration is distance-scaled: short hops are snappy, longer jumps take slightly more time. Light weight + high frequency = fast and responsive with personality.
### Description
Today the component renders the underline as a separate ::after on each tab header, sitting at opacity: 0, and selection cross-fades opacity between the old and new one. Those underlines are pinned one-per-tab, so none of them can travel, and the native motion can only ever be a cross-fade.

We replace the per-tab underline cross-fade with a single shared indicator that slides and morphs width between tabs, scoped to expressive-motion themes (One Theme). Other themes are unchanged.

### How has this been tested?

Unit test
Manual test 

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/tabs/tab-header-bar.tsx
```diff
@@ -364,6 +405,9 @@ export function TabHeaderBar({
             onFocus={onFocus}
             onBlur={onBlur}
           >
+            {isOneTheme && (
+              <span className={styles['tabs-active-indicator']} ref={activeIndicatorRef} aria-hidden="true" />
```
mxschll, line 409: Can we use the existing tab indicator? Or remove the old one and make it look like before in visual refresh, so we don't have to have 2 tab indicators?

src/tabs/tab-header-bar.scss
```diff
@@ -161,6 +162,52 @@ $label-horizontal-spacing: awsui.$space-xs;
...
+
+  @include theming.expressive-motion-only {
+    &:after {
+      display: none;
+    }
+  }
+}
+
+.tabs-active-indicator {
+  @include theming.expressive-motion-only {
```
mxschll, line 174: Let's try to only add actual animation properties inside of this mixin. The rest should be outside and shared between visual refresh and one theme.

src/tabs/tab-header-bar.tsx
ywyyu17: what is existing tab indicator? if you mean `::after` on each tab header, we need to keep it when its not one-theme

mxschll: Should we keep it? Or should we use this new span for both VR and One Theme?

PR #4917, opened 2026-08-19
feat: add theme opt-in gate for expressive motion

mxschll:
### Description

Adds scss function and mixin for gating expressive motion rules on a theme basis. Adds tests for both.

Related links, issue #, if available: `b9VcPdMJ5zRn`

### How has this been tested?

<details>
   <summary>Review checklist</summary>

_The following items are to be evaluated by the author(s) and the reviewer(s)._

#### Correctness

- _Changes include appropriate documentation updates._
- _Changes are backward-compatible if not indicated, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#public-apis)._
- _Changes do not include unsupported browser features, see [`CONTRIBUTING.md`](https://github.com/cloudscape-design/components/blob/main/CONTRIBUTING.md#browsers-support)._
- _Changes were manually tested for accessibility, see [accessibility guidelines](https://cloudscape.design/foundation/core-principles/accessibility/)._

#### Security

- _If the code handles URLs: all URLs are validated through [the `checkSafeUrl` function](https://github.com/cloudscape-design/components/blob/main/src/internal/utils/check-safe-url.ts)._

#### Testing

- _Changes are covered with new/existing unit tests?_
- _Changes are covered with new/existing integration tests?_
</details>

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #5011, opened 2026-09-14
feat: checkbox component motion

mxschll:
### Description

Adds checkmark animation for One Theme. Does not animate checkmark on mount by adding `data-awsui-motion-ready` after component has mounted and state changes.

https://github.com/user-attachments/assets/2c87907a-80e0-4f9c-b6a1-9001fb7a7848

Related links, issue #, if available: `CnHg2Iq3u1vL` and `MBMj2hmsUmZz`.

### How has this been tested?

<details>

#### Correctness

#### Security

#### Testing

</details>

---


src/internal/components/checkbox-icon/__tests__/checkbox-icon-motion.test.ts
```diff
@@ -0,0 +1,84 @@
...
+const ICON_SOURCE = fs.readFileSync(path.join(SRC_ROOT, 'internal/components/checkbox-icon/styles.scss'), 'utf8');
+const THEMING_SOURCE = fs.readFileSync(path.join(SRC_ROOT, 'internal/styles/utils/theming.scss'), 'utf8');
+const MOTION_MIXINS_SOURCE = fs.readFileSync(path.join(SRC_ROOT, 'internal/styles/motion/mixins.scss'), 'utf8');
+
+const THEME = '.awsui-one-theme';
+
+/**
+ * Compiles the checkbox-icon stylesheet against the real theming gate.
+ */
+function compile(optedInThemes: string[]): string {
```
mxschll, line 19: Same scaffolding as tests for icon motion, see https://github.com/cloudscape-design/components/pull/4917/changes

I'll open a follow-up PR to move duplicated scaffolding into a helper.

src/internal/components/checkbox-icon/index.tsx
```diff
@@ -74,6 +74,13 @@ const CheckboxIcon = ({
   const theme = useVisualRefresh() ? (oneTheme ? 'one' : 'refresh') : 'default';
   const dimensions = dimensionsByTheme[theme];
 
+  // The draw-in animation only plays once the state has changed since mount.
+  const [initialState] = useState({ checked: !!checked, indeterminate: !!indeterminate });
+  const [hasChanged, setHasChanged] = useState(false);
+  if (!hasChanged && (!!checked !== initialState.checked || !!indeterminate !== initialState.indeterminate)) {
+    setHasChanged(true);
```
mxschll, line 81: Only after component has mounted we want to play animations on state change. That's why we listen for state change since mounting.

src/internal/components/checkbox-icon/__tests__/checkbox-icon-motion.test.ts
mxschll: Tests whether the compiled animation css is included only in themes that are opted in to expressive motion here: https://github.com/cloudscape-design/components/blob/4b4d1f4b53ecdacdf40c9f49a6ba477d6f6a39aa/src/internal/styles/utils/theming.scss#L30

src/internal/components/checkbox-icon/motion.scss
```diff
@@ -0,0 +1,35 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+@use '../../styles/utils/theming' as theming;
+
+// Scopes @content to themes that opted in to $expressive-motion-themes in theming.scss.
+@mixin checkbox-expressive-with-motion {
+  @media (prefers-reduced-motion: no-preference) {
+    @include theming.expressive-motion-only(':not(.awsui-motion-disabled):not(.awsui-mode-entering)') {
```
mxschll, line 9: Same comment as for testing: I'll follow up with a PR wrapping everything in `expressive-motion-only` automatically with these `:not` selectors (or create another mixin) so we don't need to repeat this. Not part of this PR.

src/internal/components/checkbox-icon/motion.scss
```diff
@@ -0,0 +1,35 @@
...
+      @content;
+    }
+  }
+}
+
+// Include inside the checkmark line rule. Draws the checkmark in by sliding the dash offset to zero.
+@mixin draw-in {
+  &[data-awsui-motion-ready] {
+    @include checkbox-expressive-with-motion {
+      animation: awsui-checkbox-draw 250ms linear;
```
NathanZlion, line 19: Do we want to hardcode these values like the 250ms? We have tokens for other animations.
https://github.com/cloudscape-design/components/blob/e26ae03cdd10794f372db80b1a61f40b697487b6/src/attribute-editor/motion.scss#L9-L10

NathanZlion: nit btw

src/internal/components/checkbox-icon/motion.scss
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    @if theming.has-expressive-motion() {
+      @keyframes awsui-checkbox-draw {
+        from {
+          stroke-dasharray: 20;
+          stroke-dashoffset: 20px;
+        }
+        to {
+          stroke-dasharray: 20;
```
NathanZlion, line 29: nit: why repeat it if it's the same value? why include it in the animation keyframe if it doesn't change?

src/internal/components/checkbox-icon/motion.scss
mxschll: Yes, for now. We're working on clear definitions for motion tokens and timings. 

The 250ms tokens  description says it's for "expressive" motion and it's not clear what that means. Motion tokens will get an overhaul.

NathanZlion (approved):

src/internal/components/checkbox-icon/motion.scss
mxschll: This is intentional. `stroke-dasharray` is what makes the checkmark stroke behave like a dash that can be offset. We set it to a value large enough to cover the checkmark path, then animate `stroke-dashoffset` from that value to 0 to reveal the stroke progressively. 

Without the dash array, changing `stroke-dashoffset` would not create the draw-in effect.

PR #5024, opened 2026-09-17
feat: Add programmatic focus support to container

gethinwebster:
## Description

Prototype for **AWSUI-62301** — *Focusable container*.

Adds the ability to **programmatically focus a `Container`**, which:

1. moves keyboard focus to the container,
2. scrolls it into view, and
3. shows a visible **focus ring** at the destination.

This supports "jump to item" and deep-linking patterns (e.g. focusing the container a URL points to on load).

## Usage

`Container` now forwards a ref exposing `ContainerProps.Ref`:

```tsx
const ref = useRef<ContainerProps.Ref>(null);
// ...
ref.current?.focus(); // focuses + scrolls into view + shows the focus ring
<Container ref={ref} header={<Header variant="h2">Security</Header>}>...</Container>
```

## Implementation

- `Container` → internal container converted to `forwardRef`; `focus()` is wired via `useImperativeHandle`.
- `tabindex="-1"` is applied **transiently, only while the container holds focus** — it is not added to the tab order, and is removed on blur.
- `focus()` scrolls explicitly (`focus({ preventScroll: true })` + `scrollElementIntoView`) rather than relying on the implicit scroll of `HTMLElement.focus()`, which is a no-op when the container is already the active element (a repeated jump).
- `role="group"` is added to the container root (public `Container` only — internal composite usages by Table/Cards/Tabs/ExpandableSection are unchanged). When a `header` is present the group is labelled by it via `aria-labelledby`, so a focused container is announced with its heading.
- The focus ring reuses the `container-focus` mixin but is drawn on the root `::after` pseudo-element rather than the root's own `box-shadow`. This layers it above a consumer-supplied `style.root.boxShadow` (an inline box-shadow on the root) so `focus()` always shows the ring. In visual refresh `::after` is otherwise unused; in the classic theme it carries the container drop shadow, which the ring replaces for the duration of the focus.
- The container's funnel sub-step `onBlur` is composed, not replaced.

## Focus-ring visibility — comparison with existing components

The one deliberate design choice here is that the ring is **forced on for programmatic focus**, regardless of input modality. I surveyed how the rest of the library handles this:

| Component | Focus target | Ring visibility |
|---|---|---|
| Flashbar (`flash-focus-container`) | inner wrapper, `tabIndex=-1`, `role="group"` | `when-visible` (keyboard only) |
| Alert (`.focus()` ref) | inner wrapper, `tabIndex=-1`, `role="group"` | `when-visible` (keyboard only) |
| file-input (`force-focus-outline-*`) | relocates ring to visible button | `when-visible-unfocused` (still keyboard-gated) |
| Table / Tabs / date-picker | the element itself | `when-visible` (keyboard only) |
| **This PR** | container root, `tabIndex=-1`, `role="group"` | **forced on** (drawn on `::after`) |

`when-visible` gates the ring on `body[data-awsui-focus-visible="true"]`, which the global tracker sets only after a **keydown** and clears on **mousedown**. So **no existing component shows a ring for programmatic focus triggered by a click or on page load** — which is exactly the jump-to-item / deep-linking case. Matching that convention would mean the ring never appears for this feature's primary use case, so this PR forces it on. Flashbar/Alert remain the structural precedent (transient `tabIndex=-1` + `role="group"`); the only divergence is intentionally showing the ring regardless of modality.

Open questions for the team:
- Should forcing the ring be the default, keyboard-only (`when-visible`) the default, or a prop?
- Headerless containers currently expose `role="group"` without a label — should `role="group"` instead be gated on the presence of a header?

## Demo

New dev page `pages/container/focusable.page.tsx` — "jump to" links that focus/scroll to each container, plus deep-linking on load via a `?focus=<id>` query parameter (e.g. `?focus=security`). A real application would typically use the URL hash for this, but the dev pages router already owns the hash, so the demo uses a query parameter instead.

## Testing

- Unit tests: `ref.focus()` focuses the root and applies the transient `tabindex` + `role="group"` + `aria-labelledby` (to the header) + focus-ring; blur resets the transient state; a headerless container exposes `role="group"` with no label.
- `gulp build` green; container + consumer (Cards/Tabs/ExpandableSection) unit tests pass; documenter and style-docs snapshots unchanged (`role`/`aria` are not public API; `focus()` is captured); eslint + stylelint clean.
- Verified in a headless browser: clicking a jump link scrolls the target container into view and renders the focus ring.

> Draft — opening for team discussion of the proposal (API shape and the focus-ring / `role` questions above).


src/container/internal.tsx
```diff
@@ -34,6 +35,12 @@ export interface InternalContainerProps extends Omit<ContainerProps, 'variant'>,
   __headerRef?: React.RefObject<HTMLDivElement>;
   __fullPage?: boolean;
   __disableStickyMobile?: boolean;
+  /**
+   * Adds role="group" to the container root. Set by the public Container so that a
+   * programmatically focused container is announced as a group. Not applied to internal
```
georgylobko: > programmatically focused container is announced as a group

Why do we need this? I tested it locally by removing role="group" while leaving aria-labelledby unchanged, and nothing changed

gethinwebster: Adding `role="group"` is generally an a11y best-practice for this sort of situation (we do similar in focusable alerts and flashbar messages too). However, the blast radius was a bit large, I've now updated to only add it dynamically when actually focusing.

georgylobko (approved):

